PR #203, opened 2020-09-29
Spark delta client 2/3

rymurr:
Notes:

* this contains the patch from #199 to make sure the build works. Please merge #199 before review
* the deltalake package is source only. its sources are copied to the other deltalake modules for compilation. I have added a profile which makes debugging in an IDE easier but its only activated by default in intellij. I can't find an `OR` condition for 2 properties in maven and i didnt want to add a whole other profile for eclipse. Any ideas?
* as per #199 we don't use dependency management for spark classes
* this utilises packages fro `projectnessie/delta` to make delta work correctly

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/203)


jacques-n:
I believe this was merged separately.

PR #207, opened 2020-09-29
Full support for branching in Delta 

rymurr:
This commit should be merged after #203 

It contains logic to read/write multiple branches in Delta with checkpoint and delta files maintained correctly.

This is done by leveraging the `LogStore.resolvePathOnPhysicalStorage` to give unique directories to Delta's checkpoint logic. A post compact operation is done to record these files in a separate commit to nessie.

This commit does not deactivate the `VACUUM` command. This command could cause a lot of damage to a Delta table and should be deactivated. Similarly metadata is automatically cleaned after 7 days, this check should also be deactivated for Nessie

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/207)


rymurr:
re-raising after repo change

PR #218, opened 2020-09-30
Avoid codecov reporting until two statuses have been sent.

jacques-n:
Also, remove the line by line annotations in GitHub

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/218)

### DIFF
--- a/.github/.codecov.yaml
+++ b/.github/.codecov.yaml
@@ -1,0 +1,3 @@
+codecov.notify.after_n_builds: 2
+github_checks:
+  annotations: false


.github/.codecov.yaml
```diff
@@ -0,0 +1,3 @@
+codecov.notify.after_n_builds: 2
```
laurentgo, line 1: shouldn't it be 4 (3 builds for python, 1 for java?)

jacques-n: Hmm... maybe? Not a lot of docs. I was thinking it was only builds that created jacoco coverage reports. But I guess that is 4 too, isn't it. For some reason I thought we saw all the python reports in a single operation.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/218)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @jacques-n)

___
*[.github/.codecov.yaml, line 1 at r1](https://reviewable.io/reviews/projectnessie/nessie/218#-MIWMsE8IY7i3tyPn-6a-r1-1:-MI_GwSF2XZtDPiw0BJ6:bst1im0) ([raw file](https://github.com/projectnessie/nessie/blob/015bfe52c41609049e52ea29f9a314c94ff10047/.github/.codecov.yaml#L1)):*

Hmm... maybe? Not a lot of docs. I was thinking it was only builds that created jacoco coverage reports. But I guess that is 4 too, isn't it. For some reason I thought we saw all the python reports in a single operation.
</blockquote></details>

The python builds run in parallel and complete almost at the same time (also, coverage should be the same roughly, so we would not notice difference between uploads)?

jacques-n:
outdated

PR #231, opened 2020-10-01
add workflow for pypi

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/231)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,46 @@
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  python:
+    name: Python
+    runs-on: ubuntu-latest
+    env:
+      working-directory: ./python
+    strategy:
+      max-parallel: 4
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
...
+    tags:
+      - '*'
+
+jobs:
+  python:
+    name: Python
+    runs-on: ubuntu-latest
+    env:
+      working-directory: ./python
+    strategy:
```
laurentgo, line 12: does it mean this is pushed 3 times?

.github/workflows/release.yml
```diff
@@ -0,0 +1,46 @@
+on:
+  push:
+    tags:
+      - '*'
```
laurentgo, line 4: shouldn't it be based on release vs tags and the release/tag being used for pypi version too?

rymurr:
makes more sense to revisit this once we have a more complete release process

PR #243, opened 2020-10-02
Update sample CLI commands/outputs

ryantse:
Updates the sample CLI output to match current released version commands. Additionally, fixes the example using jq with a command that exists.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/243)


rymurr (approved):
thanks @ryantse !

PR #258, opened 2020-10-02
Automate deployment to OSSRH/Docker Hub

laurentgo:
Automate deployment to OSSRH/Docker Hub instead of Github packages which are only accessible with a token. Fullfill several Maven central requirements:
- add sources and javadoc packages
- sign artifacts using gpg

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/258)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,12 +15,18 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+        server-id: ossrh
+        server-username: MAVEN_USERNAME
+        server-password: MAVEN_OSSRH_TOKEN
+        gpg-private-key: ${{ secrets.MAVEN_GPG_PRIVATE_KEY }} # Value of the GPG private key to import
+        gpg-passphrase: MAVEN_GPG_PASSPHRASE # env variable for GPG private key passphrase
+
     - name: Cache local Maven repository
       uses: actions/cache@v2
       with:
         path: |
           ~/.m2/repository
-          !~/.m2/repository/com/nessie
+          !~/.m2/repository/org/projectnessie
         key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
         restore-keys: |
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -34,14 +40,16 @@ jobs:
...
       run: mvn -B deploy --file pom.xml -Pcode-coverage,native,release -DdeployAtEnd=true
       env:
-        GITHUB_TOKEN: ${{ github.token }} 
+        MAVEN_USERNAME: ${{ secrets.OSSRH_ACCESS_ID }}
+        MAVEN_OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
+        MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
     - name: Push Docker images
       run: |
-          echo '${{ secrets.GITHUB_TOKEN }}' | docker login https://docker.pkg.github.com -u '${{ github.actor }}' --password-stdin
+          echo '${{ secrets.DOCKER_TOKEN }}' | docker login https://docker.pkg.github.com -u 'nessiebuilder' --password-stdin
```
jacques-n, line 48: will this show up in the build log?

pom.xml
```diff
@@ -1031,5 +1091,48 @@ limitations under the License.
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>release</id>
```
jacques-n, line 1095: I think we should do the javadoc build pre-commit if we need for release. Otherwise, I think we'll have to chase it at release time if javadoc fails (for example because of not using &lt; as in this changeset.

laurentgo: I was wondering about this too

.github/workflows/main.yml
laurentgo: that's straight from github documentation. secrets are supposed to be masked out in job output

jacques-n (approved):

Issue #213, opened 2020-09-30
Move HMS Bridge Service tests to use Maven's toolchains to run tests with JDK8


jacques-n:
It looks like with Maven 3.6.2 you can grab the envrionment variables in the toolchain file. We need to figure out how to configure the toolchain with an action used in the build workflow and then trigger the tests based on that.

PR #260, opened 2020-10-03
Use Maven toolchains for JDK8 tests

laurentgo:
Configure toolchains.xml for both JDK 11 and 8 versions and configure
failsafe plugin for hmsclient tests to use JDK 8 toolchain as Hive tests
are not compatible with JDK11

Fixes projectnessie/nessie#213

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/260)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -64,7 +64,7 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-node-
     - name: Build with Maven
-      run: mvn -B install javadoc:javadoc-no-fork --file pom.xml -Pcode-coverage,jdk8-tests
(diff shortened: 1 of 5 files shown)


jacques-n (approved):
Awesome, thanks for putting this together. 

It looks like you didn't revert the profile for setting the compiler version. We should do a follow-up on that.

jacques-n:
Looks like this has a merge conflict and I can't figure out how to resolve in the UI without a merge (based on ui messaging). Can you fix and re-push?

laurentgo:
I addressed the conflict and also cleaned up maven compiler configuration

jacques-n (approved):

PR #270, opened 2020-10-05
Fix delta in spark3

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/270)


jacques-n (approved):
how did this work before?

rymurr:
> how did this work before?

it doesn't copy any sources or tests. so it passes silently.

I just added a flag to have it fail if no tests at all are present. Prevent this type of thing in the future

jacques-n (approved):
sorry, thought I approved this morning...

PR #305, opened 2020-10-13
 Rework REST API

laurentgo:
Rework REST API to be closer to RESTful design principles:
* Move Content Multi-GET operation to the Content API
* Use query parameters instead of path arguments for expected hash
* Replace multi endpoints with actions resources associated with branch path
* Add default Application with OpenAPI annotations
    
Update both Java and Python clients (but not JS/UI).

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)

### DIFF
--- a/clients/client/src/main/java/com/dremio/nessie/client/ClientContentsApi.java
+++ b/clients/client/src/main/java/com/dremio/nessie/client/ClientContentsApi.java
@@ -25,6 +25,8 @@ import com.dremio.nessie.error.NessieConflictException;
 import com.dremio.nessie.error.NessieNotFoundException;
 import com.dremio.nessie.model.Contents;
 import com.dremio.nessie.model.ContentsKey;
+import com.dremio.nessie.model.MultiGetContentsRequest;
+import com.dremio.nessie.model.MultiGetContentsResponse;
 
 class ClientContentsApi implements ContentsApi {
 
@@ -34,18 +36,11 @@ class ClientContentsApi implements ContentsApi {
     this.target = target;
   }
 
+
   @Override
   public Contents getContents(@NotNull ContentsKey key, String ref) throws NessieNotFoundException {
-    return target.path("contents").path(key.toPathString()).path(ref)
-                 .request()
-                 .accept(MediaType.APPLICATION_JSON_TYPE)
-                 .get()
-                 .readEntity(Contents.class);
-  }
-
-  @Override
-  public Contents getContents(ContentsKey key) throws NessieNotFoundException {
     return target.path("contents").path(key.toPathString())
+                 .queryParam("ref", ref)
                  .request()
                  .accept(MediaType.APPLICATION_JSON_TYPE)
                  .get()
@@ -53,37 +48,34 @@ class ClientContentsApi implements ContentsApi {
   }
 
   @Override
-  public void setContents(@NotNull ContentsKey key, @NotNull String hash, String message,
-                          @NotNull Contents contents) throws NessieNotFoundException, NessieConflictException {
-    target.path("contents").path(key.toPathString()).path(hash)
-          .queryParam("message", message)
-          .request()
-          .post(Entity.entity(contents, MediaType.APPLICATION_JSON_TYPE));
+  public MultiGetContentsResponse getMultipleContents(@NotNull String ref, @NotNull MultiGetContentsRequest request)
+      throws NessieNotFoundException {
+    return target.path("contents")
+        .queryParam("ref", ref)
+        .request()
+        .accept(MediaType.APPLICATION_JSON_TYPE)
+        .post(Entity.entity(request, MediaType.APPLICATION_JSON_TYPE))
+        .readEntity(MultiGetContentsResponse.class);
   }
 
 
   @Override
   public void setContents(@NotNull ContentsKey key, String branch, @NotNull String hash, String message,
                           @NotNull Contents contents) throws NessieNotFoundException, NessieConflictException {
-    target.path("contents").path(key.toPathString()).path(branch).path(hash)
+    target.path("contents").path(key.toPathString())
+          .queryParam("branch", branch)
+          .queryParam("hash", hash)
           .queryParam("message", message)
           .request()
           .post(Entity.entity(contents, MediaType.APPLICATION_JSON_TYPE));
   }
 
   @Override
-  public void deleteContents(ContentsKey key, String hash, String message)
-      throws NessieNotFoundException, NessieConflictException {
-    target.path("contents").path(key.toPathString()).path(hash)
-          .queryParam("message", message)
-          .request()
-          .delete();
-  }
-
-  @Override
   public void deleteContents(ContentsKey key, String branch, String hash, String message)
       throws NessieNotFoundException, NessieConflictException {
-    target.path("contents").path(key.toPathString()).path(branch).path(hash)
+    target.path("contents").path(key.toPathString())
+          .queryParam("branch", branch)
+          .queryParam("hash", hash)
           .queryParam("message", message)
           .request()
           .delete();
--- a/model/src/main/java/com/dremio/nessie/api/ContentsApi.java
+++ b/model/src/main/java/com/dremio/nessie/api/ContentsApi.java
@@ -36,6 +36,8 @@ import com.dremio.nessie.error.NessieConflictException;
 import com.dremio.nessie.error.NessieNotFoundException;
 import com.dremio.nessie.model.Contents;
 import com.dremio.nessie.model.ContentsKey;
+import com.dremio.nessie.model.MultiGetContentsRequest;
+import com.dremio.nessie.model.MultiGetContentsResponse;
 
 @Consumes(value = MediaType.APPLICATION_JSON)
 @Path("contents")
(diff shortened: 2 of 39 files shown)


jacques-n (commented):
Summarized feedback: 
- multi contents change (commit) seems like it should move to the ContentsApi (none of the other methods update or return contents data on the tree api
- I'm not a fan of create and mutate branch/tag. I think those should be distinct.
- Having a merge object feels heavyweight since it is only a single field.
- I don't like the list of hashes in the transplant object. I think a hash range probably makes more sense and is consistent with GIT: https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_commit_ranges. I'd actually make the range optional and allow a single commit. Maybe we should call this cherry-pick in the API, not transplant.

Rest seems good/fine.

laurentgo:
> Summarized feedback:
> 
>     * multi contents change (commit) seems like it should move to the ContentsApi (none of the other methods update or return contents data on the tree api

But an operation is not the same as content. It seems like the case because Put and Delete operation directly operate over the content, but what about unchanged?
As for no other methods return content data, I don't think any should return raw content data anyway, but there should probably be a method to get the list of operations associated with a commit.

>       - I'm not a fan of create and mutate branch/tag. I think those should be distinct.
I'm also not 100% satisfied with it...

> 
>     * Having a merge object feels heavyweight since it is only a single field.

Need to look more for this method

> 
>     * I don't like the list of hashes in the transplant object. I think a hash range probably makes more sense and is consistent with GIT: https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_commit_ranges. I'd actually make the range optional and allow a single commit. Maybe we should call this cherry-pick in the API, not transplant.
> 

Same as for merge (although I haven't changed it in this patch I believe)

> 
> Rest seems good/fine.

laurentgo:
>  - I'm not a fan of create and mutate branch/tag. I think those should be distinct.

The issue with the current design is that those are 3 endpoints (x2 since branch and tag). 
Also there's some ambiguity: is `tag/{tagName}/{hash}` an endpoint to create a tag assigned to `hash`, or is it to (force) assign hash to `tagName`. One can refer to documentation, but just the endpoint doesn't allow for clarity.

Another issue is the use of POST (vs PUT) to create the resource. POST is used for creation when `id` is generated, but here, the id is provided by the user, so PUT might actually be a better verb although it might not help regarding clarity

jacques-n:
> > * I'm not a fan of create and mutate branch/tag. I think those should be distinct.
> 
> The issue with the current design is that those are 3 endpoints (x2 since branch and tag).
> Also there's some ambiguity: is `tag/{tagName}/{hash}` an endpoint to create a tag assigned to `hash`, or is it to (force) assign hash to `tagName`. One can refer to documentation, but just the endpoint doesn't allow for clarity.

I'm fine with most of the changes here, just not the collapsing. I've always been one that thinks insert is distinct from update. For example, it is painful in Dynamo api design that if I want to do an insert I have to a conditional upsert where the condition is nothing is there. That is counterintuitive to me.

rymurr:
on the topic of list of hashes: I don't like the list of hashes either. For cherry-pick in git it actually does take a list of hashes as opposed to a range but its v awkward to handle imho

model/src/main/java/com/dremio/nessie/api/ContentsApi.java
```diff
@@ -46,54 +48,33 @@
...
-      throws NessieNotFoundException;
-
-  /**
-   * create/update a table on default branch.
-   */
   @POST
-  @Path("{key}/{hash}")
   @Consumes(MediaType.APPLICATION_JSON)
-  @Operation(summary = "update contents for path")
+  @Operation(summary = "Get multiple contents on default branch")
```
jacques-n, line 64: shouldn't be "on default branch"

jacques-n (commented):
looking good. one minor comment.

jacques-n:
I'm assuming you will ultimately fix python, [WIP] in subject and rest of tests.

laurentgo:
> I'm assuming you will ultimately fix python, [WIP] in subject and rest of tests.

Yes, waiting on #309 to address python api, and cleaning up the patch

rymurr (commented):
> Reviewed 2 of 28 files at r1, 5 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 7 of 32 files reviewed, 5 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 197 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZhph_0_EVg1xw0Zjr:-MKZhpha5kBWUu0MfOld:b-556ts2) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L197)):*
> ```Python
>     url = "/trees/branch/{}".format(branch)
>     params = None
>     if old_hash:
> ```

I think it would be easier to say:
``` python
params = dict()
if old_hash:
  params['expectedHash'] = old_hash
```
Here and in most places below. 

Alternatively `params = {'expectedHash':old_hash} if old_hash else {}` is slightly more concise
___
*[python/pynessie/_endpoints.py, line 303 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZh8JU441EZ-_NmwAK:-MKZh8JU441EZ-_NmwAL:b-k4d3vg) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L303)):*
> ```Python
>     params = { "message": reason }
>     if expected_hash:
>         params = { "expectedHash": expected_hash }
> ```

this will overwrite the `message` dict. 
___
*[python/pynessie/_endpoints.py, line 304 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZh3F-0vggKQMClBLo:-MKZh3F-0vggKQMClBLp:b-a5fugv) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L304)):*
> ```Python
>     if expected_hash:
>         params = { "expectedHash": expected_hash }
>     _post(base_url + url, json=operations, ssl_verify=ssl_verify, params={"message": message})
> ```

`params=params`?
___
*[servers/quarkus-server/src/main/resources/application.properties, line 131 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZj1eH9RMBvzix4Wel:-MKZj1eH9RMBvzix4Wem:b6ijq8f) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/servers/quarkus-server/src/main/resources/application.properties#L131)):*
> ```INI
> 
> %dev.mp.openapi.extensions.smallrye.info.title=Project Nessie API (development)
> %test.mp.openapi.extensions.smallrye.info.title=Project Nessie API (test)
> ```

why leave these 3?

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 7 of 32 files reviewed, 5 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 197 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZhph_0_EVg1xw0Zjr:-MKa2l9fDx2b896uqgyx:bwqvatc) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L197)):*

I think it would be easier to say:
``` python
params = dict()
if old_hash:
  params['expectedHash'] = old_hash
```
Here and in most places below. 

Alternatively `params = {'expectedHash':old_hash} if old_hash else {}` is slightly more concise
</blockquote></details>

replacing all occurences with your top pattern. The ternary syntax is kind of confusing to me
___
*[python/pynessie/_endpoints.py, line 303 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZh8JU441EZ-_NmwAK:-MKa30Mu15GKPECNRff3:b-886gsx) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L303)):*

this will overwrite the `message` dict. 
</blockquote></details>

Fixed
___
*[python/pynessie/_endpoints.py, line 304 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZh3F-0vggKQMClBLo:-MKa33YS7epzSb_xypYp:bllw08o) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/python/pynessie/_endpoints.py#L304)):*

`params=params`?
</blockquote></details>

Fixed too (I actually found it while fixing tests)
___
*[servers/quarkus-server/src/main/resources/application.properties, line 131 at r2](https://reviewable.io/reviews/projectnessie/nessie/305#-MKZj1eH9RMBvzix4Wel:-MKa380M-wGyTs9wcMfv:b-bi5c7q) ([raw file](https://github.com/projectnessie/nessie/blob/0e958c935d7768e932cc359d11ce3f3521874bcc/servers/quarkus-server/src/main/resources/application.properties#L131)):*

why leave these 3?
</blockquote></details>

The 3rd one is not necessary, I was wondering if we were okay having simply Nessie App for both dev and test (probably yes)

clients/client/src/main/java/com/dremio/nessie/client/ClientContentsApi.java
```diff
@@ -34,56 +36,46 @@ public ClientContentsApi(WebTarget target) {
...
-                 .accept(MediaType.APPLICATION_JSON_TYPE)
-                 .get()
-                 .readEntity(Contents.class);
-  }
 
   @Override
-  public Contents getContents(ContentsKey key) throws NessieNotFoundException {
+  public Contents getContents(@NotNull ContentsKey key, String ref) throws NessieNotFoundException {
     return target.path("contents").path(key.toPathString())
+                 .queryParam("ref", ref)
```
jacques-n, line 43: How does this show up in the construction? It seems like we should reject the ref parameter if it is provided but has no value. Thoughts?

model/src/main/java/com/dremio/nessie/model/Transplant.java
```diff
@@ -35,5 +35,4 @@
 
   List<String> getHashesToTransplant();
```
jacques-n, line 36: Looks like you decide to punt on range for this patch?

jacques-n (commented):
Java generally looks good. Will leave Python review to @rymurr .

model/src/main/java/com/dremio/nessie/model/Transplant.java
laurentgo: I wanted to wrap python tests first

clients/client/src/main/java/com/dremio/nessie/client/ClientContentsApi.java
laurentgo: So the client does not add `ref` query parameter if `ref` (the variable) is null

jacques-n: Are you saying that `foo/?ref=&baz=bar` would be rejected? and that if you passed `(ref, null)` that it would just eliminate the ref parameter entirely? Just confirming that you were stating, not asking.

Thanks

laurentgo:
While working/playing on the API and the Python client, it seems we might be missing specialization of some `/trees/tree` api for both tag and branches, to avoid ambiguity. DynamoDB does not allow tag and branch to have the same name? but inmemory and jgit do.

Conceptually TreeAPI represents 3 endpoints, one for ref (generic), one for tag (which is same as ref except ref should pnly be interpreted as a tag) and one for branch (which has extra write methods)

clients/client/src/main/java/com/dremio/nessie/client/ClientContentsApi.java
laurentgo: I guess there's the server behavior, and the client behavior.
* For the client, `queryParam("ref", null)` would remove `ref` from the list of parameters to be sent as part of the URL. To produce the URL you suggested, one would need to use `queryParam("ref", "")`
* For the server, since the argument is annotated with `@NotNull`, absence of the `ref` query parameter would throw a 400 Bad Request without invoking the method. An empty `ref` will invoke the method, with the implementation likely to send also 400 Bad Request as an empty reference is not valid. I guess we could constrain a bit further and use `@NotEmpty` or better.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 3 of 37 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 246 at r4](https://reviewable.io/reviews/projectnessie/nessie/305#-MKdtMXu6OPVdahMi_3Q:-MKdtMXu6OPVdahMi_3R:b-cazf28) ([raw file](https://github.com/projectnessie/nessie/blob/52bf479c2568bcf3c663ee8eb827898cc7c20cdb/python/pynessie/_endpoints.py#L246)):*
> ```Python
>     url = "/trees/branch/{}/transplant".format(branch)
>     params = {"expectedHash": expected_hash}
>     _post(base_url + url, json=transplant_json, ssl_verify=ssl_verify, params=params)
> ```

minor note: the `params` definition for merge and cherry_pick differ from elsewhere. I believe that requests will ignore query params with `None` as the value, is that the intention here?

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 3 of 37 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 246 at r4](https://reviewable.io/reviews/projectnessie/nessie/305#-MKdtMXu6OPVdahMi_3Q:-MKebe4w7F96Z8Rzr1yl:bsy5q6l) ([raw file](https://github.com/projectnessie/nessie/blob/52bf479c2568bcf3c663ee8eb827898cc7c20cdb/python/pynessie/_endpoints.py#L246)):*

minor note: the `params` definition for merge and cherry_pick differ from elsewhere. I believe that requests will ignore query params with `None` as the value, is that the intention here?
</blockquote></details>

I changed based on the type of `expectedHash`: when required, it is added directly to the dictionnary, otherwise, there's a check. But I guess you are now saying it doesn't matter because requests will ignore key with `None` value?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 3 of 37 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 246 at r4](https://reviewable.io/reviews/projectnessie/nessie/305#-MKdtMXu6OPVdahMi_3Q:-MKedTCsEjp7ntYiasUX:b-i53eyb) ([raw file](https://github.com/projectnessie/nessie/blob/52bf479c2568bcf3c663ee8eb827898cc7c20cdb/python/pynessie/_endpoints.py#L246)):*

I changed based on the type of `expectedHash`: when required, it is added directly to the dictionnary, otherwise, there's a check. But I guess you are now saying it doesn't matter because requests will ignore key with `None` value?
</blockquote></details>

yeah, I didn't realise requests behaves that way. So the `if old_hash:` is redundant in the other examples. Similar to your previous comment on the java client to send an empty param you would have to have `{'expectedHash':''} `

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 3 of 37 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[python/pynessie/_endpoints.py, line 246 at r4](https://reviewable.io/reviews/projectnessie/nessie/305#-MKdtMXu6OPVdahMi_3Q:-MKejWZtD7mpqJrgQdl0:bcvk8ev) ([raw file](https://github.com/projectnessie/nessie/blob/52bf479c2568bcf3c663ee8eb827898cc7c20cdb/python/pynessie/_endpoints.py#L246)):*

yeah, I didn't realise requests behaves that way. So the `if old_hash:` is redundant in the other examples. Similar to your previous comment on the java client to send an empty param you would have to have `{'expectedHash':''} `
</blockquote></details>

should I change to the latter format to make it consistent again (although with the new pattern?)

rymurr (approved):

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/305)* status: 3 of 39 files reviewed, 8 unresolved discussions (waiting on @jacques-n and @rymurr)

___
*[model/src/main/java/com/dremio/nessie/api/ContentsApi.java, line 64 at r1](https://reviewable.io/reviews/projectnessie/nessie/305#-MJVX1IbwGRlNSUYJF79-r1-64:-MKf7MFeDpuMeAHXK7Ck:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/f93e25cb2acb87c83aca63bd3aff4884e63f84c8/model/src/main/java/com/dremio/nessie/api/ContentsApi.java#L64)):*

shouldn't be "on default branch"
</blockquote></details>

Done.
___
*[model/src/main/java/com/dremio/nessie/model/Transplant.java, line 36 at r3](https://reviewable.io/reviews/projectnessie/nessie/305#-MJVX1IbwGRlNSUYJF7C-r3-36:-MKf7Nct2nhPVynHFBhz:b-cn8kn7) ([raw file](https://github.com/projectnessie/nessie/blob/b0a747561d7fbac7b751ea1596358684bf707766/model/src/main/java/com/dremio/nessie/model/Transplant.java#L36)):*

I wanted to wrap python tests first
</blockquote></details>

Will do it post 0.2

Issue #304, opened 2020-10-13
Configuration override for Maven apprunner


laurentgo:
Allow Maven apprunner mojo to override default application configuration so that concurrent executions can use different ports for example.

PR #306, opened 2020-10-13
Allow apprunner mojo to override application config

laurentgo:
Add `applicationProperties` configuration to Apprunner mojo to override
application configuration for a given execution.

Fixes projectnessie/nessie#304

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/306)

### DIFF
--- a/tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusApp.java
+++ b/tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusApp.java
@@ -56,8 +56,7 @@ public class QuarkusApp implements AutoCloseable {
   /**
    * Instantiate and start a quarkus application.
    *
-   * <p>
-   * Instantiates and start a quarkus application using Quarkus bootstrap
+   * <p>Instantiates and start a quarkus application using Quarkus bootstrap
    * framework. Only one application can be started at a time in the same
    * classloader.
    *
@@ -97,8 +96,7 @@ public class QuarkusApp implements AutoCloseable {
   /**
    * Instantiate and start a quarkus application.
    *
-   * <p>
-   * Instantiates and start a quarkus application using Quarkus bootstrap
+   * <p>Instantiates and start a quarkus application using Quarkus bootstrap
    * framework. Only one application can be started at a time in the same
    * classloader.
    *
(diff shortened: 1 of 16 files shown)


tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusAppStartMojo.java
```diff
@@ -70,6 +73,12 @@
   @Parameter(property = "nessie.apprunner.appArtifactId", required = true)
   private String appArtifactId;
 
+  /**
+   * Application configuration properties.
+   */
+  @Parameter
```
rymurr, line 79: This is used via a `<properties>` block in the config of the plugin then? And its a user error if the properties are invalid or contracticting

tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusApp.java
```diff
@@ -62,7 +67,8 @@ private QuarkusApp(RunningQuarkusApplication runningApp) {
    * @throws MojoExecutionException if an error occurs during execution
    */
   public static QuarkusApp newApplication(MavenProject project, RepositorySystem repoSystem,
-      RepositorySystemSession repoSession, String appArtifactId)
+      RepositorySystemSession repoSession, String appArtifactId, Properties applicationProperties,
+      Collection<Path> additionalDependencyPaths)
```
rymurr, line 71: what are the additional dependencies for?

rymurr (commented):
looks good to me. I am not sure I understand yet how to pass these configs to the underlying tests

tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusAppStartMojo.java
laurentgo: the configuration key is derived from the field name, so in this case, it's `applicationProperties`. They take precedence over values defined in `application.properties`

tools/apprunner-maven-plugin/src/main/java/com/dremio/nessie/quarkus/maven/QuarkusApp.java
laurentgo: The `MojoSourceConfig` needs to be part of the application runtime in order to be able to set the properties. Should probably find a better/smarter way of (it's still a bit experimental)

jacques-n (approved):

rymurr (approved):

PR #311, opened 2020-10-16
build and activate swagger ui in docker

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/311)

### DIFF
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -35,8 +35,8 @@ nessie.backends.type=DYNAMO
 nessie.version.store.dynamo.initialize=false
 ### table names for ref, tree and value tables respectively
 nessie.version.store.dynamo.refTableName=nessie_refs
-nessie.version.store.dynamo.treeTableName=nessie_trees
-nessie.version.store.dynamo.valueTableName=nessie_values
+nessie.version.store.dynamo.treeTableName=nessie_objects
+nessie.version.store.dynamo.valueTableName=nessie_objects
 
 ## Dynamo
 quarkus.dynamodb.aws.region=us-west-2
@@ -69,7 +69,7 @@ quarkus.oidc.enabled=false
 ## Quarkus swagger settings
 # fixed at buildtime
 quarkus.swagger-ui.always-include=true
-quarkus.swagger-ui.enable=false
+quarkus.swagger-ui.enable=true
 
 ## Quarkus monitoring and tracing settings
 ## jaeger specific settings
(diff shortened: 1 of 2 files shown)


servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -69,7 +69,7 @@ quarkus.oidc.enabled=false
 ## Quarkus swagger settings
 # fixed at buildtime
 quarkus.swagger-ui.always-include=true
-quarkus.swagger-ui.enable=false
+quarkus.swagger-ui.enable=true
```
jacques-n, line 72: Does this mean it will be on by default? I'd like to include but have off by default.

laurentgo: I'm a bit confused by the configuration keys at https://quarkus.io/guides/openapi-swaggerui#quarkus-smallrye-openapi_configuration

Should we enable, but only include for dev mode?

rymurr: `quarkus.swagger-ui.always-include` adds teh swagger UI to the package
`quarkus.swagger-ui.enable` actually adds a route to the running application.

However both are 'build-time' configs so they are respected only at build time and can't be turned on in the graalvm image. 

This change enables the UI at build time so it will be available in the docker image. Lately I have been using the docker image for testing UI/CLI/Spark so I find the swagger ui activated to be rather useful but as it becomes more mature I think we would want to take it out of the docker image.

So we can:
1) take this change and leave swagger ui always on
2) change this so that the swagger ui is only included and available when running `quarkus:dev`

I prefer to leave the swagger ui in the docker image for at least a few months but am not that bothered if someone else feels diffferently.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -132,15 +127,14 @@ quarkus.container-image.name=nessie
...
-mp.openapi.extensions.smallrye.info.title=Example API
-%dev.mp.openapi.extensions.smallrye.info.title=Example API (development)
-%test.mp.openapi.extensions.smallrye.info.title=Example API (test)
-mp.openapi.extensions.smallrye.info.version=1.0.1
-mp.openapi.extensions.smallrye.info.description=Just an example service
-mp.openapi.extensions.smallrye.info.termsOfService=Your terms here
-mp.openapi.extensions.smallrye.info.contact.email=techsupport@example.com
-mp.openapi.extensions.smallrye.info.contact.name=Example API Support
-mp.openapi.extensions.smallrye.info.contact.url=http://exampleurl.com/contact
+mp.openapi.extensions.smallrye.info.title=Project Nessie API
```
laurentgo, line 130: I was thinking of adding those in model using the annotations. Thoughts?

rymurr: yes, that makes a lot more sense imho

rymurr: I've removed these lines. Want to add your change to this patch or separately?

jacques-n (approved):

PR #361, opened 2020-10-16
Move backend api/model classes to their own module

laurentgo:
Move backend api/model classes from model to backends/spi module as
those are not required in order to use/implement the Nessie REST api.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/361)

### DIFF
--- a/backends/dynamodb/pom.xml
+++ b/backends/dynamodb/pom.xml
@@ -32,7 +32,7 @@
   <dependencies>
     <dependency>
       <groupId>org.projectnessie</groupId>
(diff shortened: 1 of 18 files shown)


jacques-n (approved):
Seems fine for now. It isn't clear how much of this we will use ultimately.

laurentgo:
It's mostly for clarity. While working on the REST API, I was confused by what some of the model were for originally before realizing those were only used by backend api.

rymurr:
im still keen on deleting it :-) its all used for the jgit / dynamo impl

PR #428, opened 2020-11-03
Add validation check for commit and transplant

laurentgo:
Add a validation check for commit and transplant methods to confirm that
the hash used as reference belongs to the branch the changes are
pushed into.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/428)

### DIFF
--- a/versioned/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoDBVersionStore.java
+++ b/versioned/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoDBVersionStore.java
@@ -18,6 +18,8 @@ package com.dremio.nessie.versioned.impl;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.extension.ExtendWith;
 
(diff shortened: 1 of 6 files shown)


jacques-n:
I think there may be an inconsistency in how we're thinking about the empty commit. In Dynamo, I consider all trees to share the same ultimate empty commit. That means that all commits are guaranteed to share a distant ancestor. I think of this as compatible/consistent with how git works (that you can merge a branch into another branch even if the two only share empty state). It seems like this is inconsistent with what you're presenting here.

laurentgo:
I would not phrase it like this? My understanding of the expected hash was it was the branch's hash when fetched before applying commit, and so a sanity check would be to confirm that the hash still belongs to the branch. On the other hand, the `merge` operation definitively finds a common ancestor up to root.

The comparison with git is quite inaccurate because all operations are done on top of the branch (there's no reference hash) and git will refuse pushing changes if the history is changed (unless you force assign)

rymurr:
I suppopse that one issue i have is transplant throws `IllegalArgumentException` when the transplant list is out of order or invalid and the expectedHash check throws `ReferenceNotFound`. To me they are equivalent in the sense they are both determining ancestry and should fail in the same way.

I do however think these tests are relevant and reflect my understanding of how the `VersionStore` works. If expected hash isn't a direct ancestor of HEAD then I feel like that is user error and should be prevented

rymurr (approved):
> Reviewed 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/428)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #430, opened 2020-11-03
Add merge and transplant to jgit VersionStore

rymurr:
Also added basic merge tests 

closes #135

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/430)

PR #434, opened 2020-11-05
include last parent in fetch list of L1s for transplant

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/434)

### DIFF
--- a/versioned/dynamodb/src/main/java/com/dremio/nessie/versioned/impl/TieredVersionStore.java
+++ b/versioned/dynamodb/src/main/java/com/dremio/nessie/versioned/impl/TieredVersionStore.java
@@ -20,10 +20,14 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Spliterator;
+import java.util.Spliterators.AbstractSpliterator;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


versioned/jgit/src/test/java/com/dremio/nessie/versioned/impl/ITJGitInMemoryVersionStore.java
```diff
@@ -78,4 +78,7 @@ void tearDown() {
   protected void transplant() {
   }
 
+  @Disabled
+  protected void transplantBasic() {
+  }
```
rymurr, line 83: temporary till #430 is merged

versioned/jgit/src/test/java/com/dremio/nessie/versioned/impl/ITJGitOnDiskVersionStore.java
```diff
@@ -82,4 +82,7 @@ void tearDown() {
   protected void transplant() {
   }
 
+  @Disabled
```
rymurr, line 85: temporary till #430 is merged

versioned/dynamodb/src/main/java/com/dremio/nessie/versioned/impl/TieredVersionStore.java
```diff
@@ -465,10 +469,9 @@ public void transplant(BranchName targetBranch, Optional<Hash> currentBranchHash
         true,
         (from, commonParent) -> {
           // first we need to validate that the actual history matches the provided sequence.
-          List<L1> l1s = Lists.reverse(
-              new HistoryRetriever(store, from, endTarget, true, false, true).getStream().map(HistoryItem::getL1)
-              .collect(ImmutableList.toImmutableList()));
-          List<Hash> hashes = l1s.stream().map(L1::getId).map(Id::toHash).collect(Collectors.toList());
+          Stream<L1> historyStream = new HistoryRetriever(store, from, null, true, false, true).getStream().map(HistoryItem::getL1);
+          List<L1> l1s = Lists.reverse(takeUntilNext(historyStream, endTarget).collect(ImmutableList.toImmutableList()));
+          List<Hash> hashes = l1s.stream().map(L1::getId).map(Id::toHash).skip(1).collect(Collectors.toList());
```
jacques-n, line 474: why are you skipping 1?

rymurr:
1 is always going to be the commit prior to the first cherry pick. So 2..n
will match the requested array.

jacques-n:
I'm not following why you would skip the first commit. If I say grab c1, you'd get a history starting at c1 and ending at c1 - 1. Why would you drop c1?

rymurr:
> I'm not following why you would skip the first (most recent) commit. If I say grab c1, you'd get a history starting at c1 and ending at c1 - 1. Why would you drop c1?

Correct, we ask for `c1` and the list comes back as `{c1, c1-1}` which is then reversed and we drop `c1-1` so the list is just `c1` to match the assertion. however the list `{c1-1, c1}` is returned back to the `internalTransplant` method.

So we return N+1 `L1`s (where N is the number of commits to transplant) and we also ensure that the last N commits match the commits we wanted to transplant.

jacques-n (approved):
Got it. My skim didn't clarify that the skip is on the l1 list, not the history retriever.

jacques-n (approved):

PR #450, opened 2020-11-11
[fix] bad region variable name in example

rdettai:
The region config comes from the `quarkus.dynamodb.aws.region` property in `BackendProducer.java`, so the environment variable is `QUARKUS_DYNAMODB_AWS_REGION`

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/450)


rymurr (approved):
Thanks for the fix @rdettai !!

rdettai:
Very modest contribution! but it took me quite some time to figure out why nothing was showing up in my local dynamo instance 😄

rymurr:
> Very modest contribution! but it took me quite some time to figure out why nothing was showing up in my local dynamo instance

The more the merrier :-) please shout if you find anymore. I was surprised I didn't pick this up earlier, but it looks like its defaulted in my `~/.aws/config`

PR #480, opened 2020-11-20
Add ErrorProne checker for SLF4J

laurentgo:
Add sl4j error prone checkers to prevent use of string concatenation,
non-constant strings and use of invalid placeholders.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/480)


jacques-n (approved):
nice, thanks @laurentgo !

PR #494, opened 2020-11-24
enable python 3.9 in testing

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/494)


jacques-n:
I think you need to update the value here:
https://github.com/projectnessie/nessie/blob/main/.github/.codecov.yaml

Also, I think it is wrong right now, we seem to have extra builds and thus it is reporting early again. Maybe it should be 6 now and 7 with this patch?

rymurr:
> I think you need to update the value here:
> https://github.com/projectnessie/nessie/blob/main/.github/.codecov.yaml
> 
> Also, I think it is wrong right now, we seem to have extra builds and thus it is reporting early again. Maybe it should be 6 now and 7 with this patch?

Looking at the build for this PR there were 5 separate instances where we sent data to codecov. So I think 5 is correct. Looking at [this](https://docs.codecov.io/docs/notifications#comment) it appears there are separate counts for comments and general notifications. I have added a config specific for comments now too.

.codecov.yml
```diff
@@ -0,0 +1,20 @@
+#
```
rymurr, line 1: note the file move. The github action seems to look for this file in the root rather than in `.github`

jacques-n (approved):

PR #510, opened 2020-11-30
Constrain the NUMBER Entity to be a long.

kylep-dremio:
Previously it was a String which could allow for
a large range of possible values, when a much
smaller range of whole values was used by Nessie.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/510)


jacques-n (approved):

jacques-n:
Coverage error is bad confusion on codecov (confusing python with entire project I think)

PR #557, opened 2020-12-14
put jersey client onto the hive shaded classpath

rymurr:
Hive 2.3.x has only jax-rs 1.1 by default. This PR adds jax-rs 2.0 to the shaded jar. Because 1.1 and 2.0 have overlapping classes we have to relocate most of the shaded dependencies.

Bad news: shaded jar went from 500kb to 5MB :-(

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/557)


rymurr:
Moving to draft status as we look at different options for http client

PR #562, opened 2020-12-15
Keep Python and Java code samples in sync

andrioni:
Small fix to set a Spark 3.0 Catalog to Nessie in both code samples, as originally it was only present in the Python one.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/562)


rymurr (approved):
Good catch @andrioni ! thanks a lot for the contribution

andrioni:
Thanks for the quick review!

PR #566, opened 2020-12-15
Separate generic store tests to tiered-tests.

kylep-dremio:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/566)


versioned/tiered-tests/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie-versioned</artifactId>
+    <version>0.2.2-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nessie-versioned-tiered-tests</artifactId>
+
+  <name>Nessie - Versioned - TieredVersionStore Tests</name>
```
jacques-n, line 30: aren't these really store tests, not TieredVersionStore tests?

I'm also wondering if we should just move these into the spi module... 

Thoughts @laurentgo ?

laurentgo: Do you mean under `src/test/java`? Maven does not have different set of dependencies for jars produced by the same module, which might cause issue for other projects which want to reuse the tests, hence why the versioned tests were made a separate module for example

kylep-dremio: They are store tests, so we can put it in nessie-versioned-store-tests. I had patterned it after the spi-tests as a separate package.

laurentgo (approved):

kylepbit:
I have signed the CLA, unsure why the request is still pending.

kylep-dremio:
Unsure why the license/cla check is still pending, I've signed it and previously contributed to this project.

PR #575, opened 2020-12-16
Replace JAX-RS 2.0 client with basic client which wraps URLConnection

rymurr:
This PR does the following:

* remove dependencies on JAX-RS 2.0 from client 
* Add a simple `URLConnection` wrapper for the client
* clean up dependencies for clients. Remove Jersey and JAX-RS classes from iceberg, hive, delta
* shaded iceberg and hive dependencies are now only ~200kb

Todo:
* ~add unit tests!~
* potentially extend feature set
* efficiency improvements

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)


rymurr:
fixed the build error! Adding more test coverage now.

laurentgo:
FYI did some very basic tests to create a package with nessie client + jax-rs implementation (with jackson provider), and check the jar size:
- jersey version is at 3.7MB
- resteasy version is at 7.1MB

rymurr:
> FYI did some very basic tests to create a package with nessie client + jax-rs implementation (with jackson provider), and check the jar size:

Some other options I looked into:
* https://square.github.io/retrofit/ this adds a ton: okhttp etc. anything that adds okhttp adds kotlin and a bunch of other deps
* http://kong.github.io/unirest-java/ this adds about 5mb if you shade everything. You can take it down to less than a mb (as little as 600kb if you are strict) if you strip out async, gson and assume that apache httpclient & jackson work across versions
* https://github.com/sendgrid/java-http-client this is rather small on its own. It uses httpclient. The best alternative to this patch as it is as thin as this patch but wraps apache httpclient rather than vanilla jdk. Its usability hinges on how much httpclient changes between versions

A thought I just had:
This client is inflexible and limiting (although I like to think rather nice). But the driver for this (hive2 and to some extent spark2.4) are just as limiting: we don't have much room to improve the rest api as long as we are using hive2 (eg adding http2 or async would be potentially troublesome in hive2). However JDK11 has a rather nice http client built in: support for http2, async among others. So as soon as we can remove java 8 we are mostly unblocked and can replace this with a wrapper around java11s http client. From that perspective I like this approach as a stopgap until we are able to move past java8

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)* status: 0 of 38 files reviewed, 6 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java, line 73 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1n-kL8CGZS1CtLNQ-:-MP1n-kL8CGZS1CtLNQ0:b-ivofr7) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java#L73)):*
> ```Java
>     NessieError error;
>     if (inputStream == null) {
>       error = new NessieError(statusCode, null, new RuntimeException("Could not parse error object in response."));
> ```

Can you add similar reason phrases? I think that is helpful.
___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java, line 76 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1mv4h5_PP_uBrK4vT:-MP1mv4iE9Twvn8lx4Ye:bu8g8xl) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java#L76)):*
> ```Java
>     } else {
>       try {
>         error = reader.readValue(inputStream);
> ```

Why did you move the reader from static?
___
*[clients/deltalake/spark2/pom.xml, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nCP80nbT1d7BxAJO:-MP1nCP80nbT1d7BxAJP:b-bu7w39) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/deltalake/spark2/pom.xml#L89)):*
> ```XML
>       <groupId>com.fasterxml.jackson.module</groupId>
>       <artifactId>jackson-module-scala_2.12</artifactId>
>       <version>2.8.8</version>
> ```

why did you move the version out of the property?
___
*[clients/deltalake/spark3/pom.xml, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nFMa8n4_F2VfSWVV:-MP1nFMa8n4_F2VfSWVW:b-bu7w39) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/deltalake/spark3/pom.xml#L89)):*
> ```XML
>       <groupId>com.fasterxml.jackson.module</groupId>
>       <artifactId>jackson-module-scala_2.12</artifactId>
>       <version>2.10.2</version>
> ```

why did you move the version out of the property?
___
*[model/src/main/java/com/dremio/nessie/error/BaseNessieClientServerException.java, line 67 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nNd_4myh6UynJ4U1:-MP1nNd_4myh6UynJ4U2:bw6fdae) ([raw file](https://github.com/projectnessie/nessie/blob/9ce6d6e610bff29a2cfb27d577063c9ca9e84a7a/model/src/main/java/com/dremio/nessie/error/BaseNessieClientServerException.java#L67)):*
> ```Java
>   }
> 
>   public String getReason() {
> ```

Per elsewhere, it is useful to add this information somehow (otherwise when you're tracing you have to go and lookup).
___
*[model/src/main/java/com/dremio/nessie/error/NessieError.java, line 72 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1ncjW4acjE9qDohjr:-MP1ncjW4acjE9qDohjs:blciofj) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/model/src/main/java/com/dremio/nessie/error/NessieError.java#L72)):*
> ```Java
>   private String standardMessage(int status) {
>     switch (status) {
>       case 200:
> ```

Rather than declare the numbers in multiple places, let's add our own basic enum with names, reasons and status codes and use it in place of the current Status class.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)* status: 0 of 38 files reviewed, 6 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java, line 73 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1n-kL8CGZS1CtLNQ-:-MP3vTJ4642LUMF935yA:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java#L73)):*

Can you add similar reason phrases? I think that is helpful.
</blockquote></details>

Done.
___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java, line 76 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1mv4h5_PP_uBrK4vT:-MP3vUN63gLUDkbjhkEQ:bgtc37n) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/client/src/main/java/com/dremio/nessie/client/rest/ResponseCheckFilter.java#L76)):*

Why did you move the reader from static?
</blockquote></details>

Wanted it to be configured the same way as the other object mappers
___
*[clients/deltalake/spark2/pom.xml, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nCP80nbT1d7BxAJO:-MP3wHew4CiEiD6RucEO:b-fzhfw4) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/deltalake/spark2/pom.xml#L89)):*

why did you move the version out of the property?
</blockquote></details>

It was only being used in delta and wanted to clean up the main pom. Doesnt bother me tho, moved it back
___
*[clients/deltalake/spark3/pom.xml, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nFMa8n4_F2VfSWVV:-MP3wR9gBeXThU0i5NVb:b-fzhfw4) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/clients/deltalake/spark3/pom.xml#L89)):*

why did you move the version out of the property?
</blockquote></details>

It was only being used in delta and wanted to clean up the main pom. Doesnt bother me tho, moved it back
___
*[model/src/main/java/com/dremio/nessie/error/BaseNessieClientServerException.java, line 67 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1nNd_4myh6UynJ4U1:-MP47MuJBZPDz2OxtFt2:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/9ce6d6e610bff29a2cfb27d577063c9ca9e84a7a/model/src/main/java/com/dremio/nessie/error/BaseNessieClientServerException.java#L67)):*

Per elsewhere, it is useful to add this information somehow (otherwise when you're tracing you have to go and lookup).
</blockquote></details>

Done.
___
*[model/src/main/java/com/dremio/nessie/error/NessieError.java, line 72 at r2](https://reviewable.io/reviews/projectnessie/nessie/575#-MP1ncjW4acjE9qDohjr:-MP47Nrz1F-nT2AIH1Ys:bmjfbuj) ([raw file](https://github.com/projectnessie/nessie/blob/50fcdc1875ad299b38fcefd5dc63a8d87c37746f/model/src/main/java/com/dremio/nessie/error/NessieError.java#L72)):*

Rather than declare the numbers in multiple places, let's add our own basic enum with names, reasons and status codes and use it in place of the current Status class.
</blockquote></details>

Done. I have created the class in the http package and rather than expose it pass an int and string to the Error classes

rymurr:
I don't love the resource management approach I am taking. The network resources are released when the `InputStream` is closed so I am closing the `InputStream` right after reading the body. This appears to be similar to JAX-RS libraries which don't let you read the input stream multiple times. Still feels a bit floppy tho

snazy (changes requested):
Added a few nit-ish notes to some files.

I've drafted a PR against your branch to address a couple more things that bothered me:

* [De-duplicate http-client configuration constants](https://github.com/rymurr/nessie/commit/5011de565c9500c6b02601de17bac936c94d605a): we have a bunch of places where `CONF_NESSIE_*` constants are defined with the same values. I've moved those constants into a new class and let the existing code use those.
* `HttpClient` already had some static factory-ish methods (`none`, `basic`), but those were unused. Using the (only) constructor of `HttpClient` from everywhere and peeling and validating the config parameters everywhere seems to be duplicated code. [Unified this](https://github.com/rymurr/nessie/commit/280186d8262a17c465f9cb520a71d4dbdc53a303).
* Also simplified the [handling of auth-stuff](https://github.com/rymurr/nessie/commit/924624e6cdba8f6984a1c1751ac92b66b3d2e973)
* Nits: [renaming and such](https://github.com/rymurr/nessie/commit/8c89050172677c8b5f41bfcf65fd458cee7636f2)
* The new `TestHttpClient` hangs infinitely, when an assertion fails. [fixed here](https://github.com/rymurr/nessie/commit/63b22cf8d271bc6576ff3d533a33395aa0a7fd56)
* I wasn't particularly happy with all the lists and such in `UriBuilder`, so I've [refactored that a bit and added a unit test for that class](https://github.com/rymurr/nessie/commit/5ffef6e326c4bcd10661d4546a99402a09bb5829)
* The `Status` class would cause some nasty trouble, when a server returns a status code that's not mentioned in the enum. [Refactored `Status` to be a simple class](https://github.com/rymurr/nessie/commit/2c087016ed8965a95459948501ccf2981d8552bc).

Opened the very first PR against your repo here, that contains all the commits above: https://github.com/rymurr/nessie/pull/1

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575#-:-MP9MfIEDGMWbVZpWRBQ:bq9op8i)* status: 0 of 40 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 36 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9M705CPqtzC_lO8ni:-MP9M705CPqtzC_lO8nj:bd5j6sc) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L36)):*
> ```Java
>  * </p>
>  */
> public class HttpClient {
> ```

This class should do some basic parameter validation.
Proposal here: https://github.com/rymurr/nessie/commit/ff9a765cc021f8da88162b556eaf64666c056a2a
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LrdwEPD_S_5YECr-:-MP9LrdwEPD_S_5YECr0:b-1o1bip) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L31)):*
> ```Java
>  * Construct a URI from base and paths. Adds query parameters, supports templates and handles url encoding of path.
>  */
> class UriBuilder {
> ```

The class should do some basic parameter validattion.
Proposal here: https://github.com/rymurr/nessie/commit/a2059ed9ec39122eb3a195a3ff55eb9f1dff5b43
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 39 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LZ2z5wMzBcF-HoZW:-MP9LZ2z5wMzBcF-HoZX:b8hdg2e) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L39)):*
> ```Java
> 
>   private static final Executor EXEC = Executors.newSingleThreadExecutor();
>   private static final ObjectMapper MAPPER = new ObjectMapper();
> ```

Not a fan of statically instantiating executor-services and let threads linger around.
Added a fix in my PR for your branch: https://github.com/rymurr/nessie/commit/cdc30ddb51136bb0f4581b7f6b516ca7fde65f92
___
*[clients/deltalake/spark2/pom.xml, line 95 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LJ7vAWYZDM5fFKmb:-MP9LJ7vAWYZDM5fFKmc:borxqj5) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/deltalake/spark2/pom.xml#L95)):*
> ```XML
>       <groupId>com.fasterxml.jackson.core</groupId>
>       <artifactId>jackson-databind</artifactId>
>       <version>2.8.8</version>
> ```

Can you move the version in to the top-parent-pom as a property?
___
*[clients/deltalake/spark3/pom.xml, line 95 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LBkZETaiPp-7y36m:-MP9LBk_0HmjUsGEiVvE:borxqj5) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/deltalake/spark3/pom.xml#L95)):*
> ```XML
>       <groupId>com.fasterxml.jackson.core</groupId>
>       <artifactId>jackson-databind</artifactId>
>       <version>2.10.2</version>
> ```

Can you move the version in to the top-parent-pom as a property?

rymurr (commented):
Have fixed your comments below and added some of your fixes in my most recent update. Also responded to rymurr#1

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575#-:-MP9nlQO2p7pnZm4wAkH:b-fxfy4w)* status: 0 of 40 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 36 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9M705CPqtzC_lO8ni:-MP9hf2R7oRCeqob1zVw:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L36)):*

This class should do some basic parameter validation.
Proposal here: https://github.com/rymurr/nessie/commit/ff9a765cc021f8da88162b556eaf64666c056a2a
</blockquote></details>

Done.
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LrdwEPD_S_5YECr-:-MP9iLC40cX_FZ1ZAfJ0:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L31)):*

The class should do some basic parameter validattion.
Proposal here: https://github.com/rymurr/nessie/commit/a2059ed9ec39122eb3a195a3ff55eb9f1dff5b43
</blockquote></details>

Done.
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 39 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LZ2z5wMzBcF-HoZW:-MP9i_dZ0JwWmQwoKNM3:b-e4okhf) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L39)):*

Not a fan of statically instantiating executor-services and let threads linger around.
Added a fix in my PR for your branch: https://github.com/rymurr/nessie/commit/cdc30ddb51136bb0f4581b7f6b516ca7fde65f92
</blockquote></details>

removed the executor. The server should use its internal executor
___
*[clients/deltalake/spark2/pom.xml, line 95 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LJ7vAWYZDM5fFKmb:-MP9iwKyDvw50JK9M9Bc:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/deltalake/spark2/pom.xml#L95)):*

Can you move the version in to the top-parent-pom as a property?
</blockquote></details>

Done.
___
*[clients/deltalake/spark3/pom.xml, line 95 at r4](https://reviewable.io/reviews/projectnessie/nessie/575#-MP9LBkZETaiPp-7y36m:-MP9ixqZEO4_W3FZ4sE0:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/af1ea157dbd9dcc60d8f5101a388e5e26aaa6bf4/clients/deltalake/spark3/pom.xml#L95)):*

Can you move the version in to the top-parent-pom as a property?
</blockquote></details>

Done.

snazy (approved):
> Reviewed 11 of 34 files at r1, 5 of 15 files at r2, 10 of 21 files at r3, 3 of 4 files at r4, 10 of 15 files at r5, 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)* status: all files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

laurentgo (changes requested):
> Reviewed 3 of 34 files at r1, 2 of 15 files at r2, 3 of 21 files at r3, 8 of 15 files at r5.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)* status: all files reviewed, 29 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 68 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUY8i9azHRPeHPoWh:-MPAUY8j2cXQ7Ca32yT9:bqkw7qm) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L68)):*
> ```Java
>   }
> 
>   public void register(RequestFilter filter) {
> ```

javadoc?
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 78 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUK90A1ROlVq5ZQra:-MPAUK90A1ROlVq5ZQrb:bsc4i99) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L78)):*
> ```Java
>   }
> 
>   public HttpRequest create() {
> ```

(style) maybe name the method `newRequest` vs `create` (at first, I was wondering if it was creating a new client...)
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 69 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBDJbJ1EZNTQUKInoa:-MPBDJbJ1EZNTQUKInob:b-rwlw0c) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L69)):*
> ```Java
> 
>   public void register(RequestFilter filter) {
>     filter.init(mapper);
> ```

If not doing automatic DI, I would still do injection before doing registration (which means the client could probably get the objectmapper injected too)
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClientException.java, line 24 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUdW14_U4cw64ZGFr:-MPAUdW14_U4cw64ZGFs:bweve1h) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClientException.java#L24)):*
> ```Java
>   }
> 
>   public HttpClientException(String message) {
> ```

shouldn't status code be part of the exception? or is the exception not designed to return server errors?
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java, line 33 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAVOApDh37As8C4hmZ:-MPAVOApDh37As8C4hm_:bb68l94) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java#L33)):*
> ```Java
>   private final ObjectMapper mapper;
> 
>   HttpResponse(ResponseContext context, ObjectMapper mapper) throws IOException {
> ```

use of exceptions doesn't seem consistent. Sometimes `IOException` is thrown, sometimes it is wrapped...
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java, line 36 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAVd0Q-ZKv5dWarm-Z:-MPAVd0Q-ZKv5dWarm-_:b5se6fu) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java#L36)):*
> ```Java
>     this.responseContext = context;
>     this.mapper = mapper;
>     context.getResponseCode();
> ```

maybe a comment to explain why the method is invoked but result is not stored?
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/RequestContext.java, line 48 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAW96tEyZU42pcEGD6:-MPAW96tEyZU42pcEGD7:bi4vpz8) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/RequestContext.java#L48)):*
> ```Java
>   }
> 
>   public Map<String, String> getHeaders() {
> ```

Technically HTTP headers are multi-valued...
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/RequestFilter.java, line 28 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBD1iX1Cey7fro4pfk:-MPBD1iX1Cey7fro4pfl:b-o8t5r3) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/RequestFilter.java#L28)):*
> ```Java
>   void filter(RequestContext context);
> 
>   default void init(ObjectMapper mapper) {
> ```

it feels like injection should happen at construction time (which should not be a big deal)
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 43 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBFCQB96vSd7oWN0EB:-MPBFCQB96vSd7oWN0EC:b-9a3dc3) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L43)):*
> ```Java
>       uri.append('/');
>     }
>     uri.append(HttpUtils.checkNonNullTrim(path));
> ```

should it check for empty values too?
also, what happens if path contains special characters like "?"?
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 72 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGH3C3y6DLN4nlL0V:-MPBGH3DCl04uL6cVvoM:b-s6qjoc) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L72)):*
> ```Java
>     uriBuilder.append(baseUri);
> 
>     if ('/' != uriBuilder.charAt(uriBuilder.length() - 1)) {
> ```

assumes that baseUri is not empty...
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 78 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGTIqBN2ZUrgdiXXg:-MPBGTIrBb19RWo2fjTn:b-j53kqn) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L78)):*
> ```Java
>     Map<String, String> templates = new HashMap<>(templateValues);
>     String replaced = Arrays.stream(uri.toString().split("/"))
>                             .map(p -> encode((templates.containsKey(p)) ? templates.remove(p) : p))
> ```

so I guess it means that the whole path component should be a template?
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 79 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGaeJD_tI3ojuArsO:-MPBGaeJD_tI3ojuArsP:b-v1mmim) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L79)):*
> ```Java
>     String replaced = Arrays.stream(uri.toString().split("/"))
>                             .map(p -> encode((templates.containsKey(p)) ? templates.remove(p) : p))
>                             .collect(Collectors.joining("/"));
> ```

would be nice if it could be added directly to uriBuilder instead of creating some temporary string (maybe with a custom collector)
___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/NessieHttpResponseFilter.java, line 37 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGxFcBwN92h9ZtKYC:-MPBGxFcBwN92h9ZtKYD:bl6csva) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/rest/NessieHttpResponseFilter.java#L37)):*
> ```Java
>   public void filter(ResponseContext con) {
>     try {
>       ResponseCheckFilter.checkResponse(con, mapper);
> ```

Maybe we should consider doing at NessieClient level instead of using some filter
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 32 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBHa1h8KZJEopuBXpj:-MPBHa1h8KZJEopuBXpk:bo3sqn) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L32)):*
> ```Java
> 
> import com.fasterxml.jackson.databind.ObjectMapper;
> import com.sun.net.httpserver.HttpHandler;
> ```

using com.sun packages doesn't help with portability... (there are probably warnings popping up about use of private packages)
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 238 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBHz5Q-YD7ZbCYQfPH:-MPBHz5Q-YD7ZbCYQfPI:b-t0e79q) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L238)):*
> ```Java
>           handler.handle(exchange);
>         } catch (RuntimeException | Error e) {
>           e.printStackTrace();
> ```

use a logger?
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 76 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBI8VZ0sXhe3b9VO_n:-MPBI8VZ0sXhe3b9VO_o:b-p9l9fg) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L76)):*
> ```Java
>       os.close();
>     };
>     HttpsServer server = HttpsServer.create(new InetSocketAddress("localhost", 0), 0);
> ```

why not using a try-with-resources pattern like the tests for http?
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 84 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBIQVs1ZNnJg0cGYfa:-MPBIQVs1ZNnJg0cGYfb:b-x3k2cn) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L84)):*
> ```Java
>       SSLContext sc = SSLContext.getInstance("SSL");
>       sc.init(null, trustManager, new java.security.SecureRandom());
>       HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
> ```

That's not something we should do (even in a test) as it changes the whole context. The right thing to do is to allow the ssl context to be configured per client...
and it would be interesting to verify that an unconfigured client fails to connect because of a certificate validation issue...
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 88 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBI_jm1rd9Ug9MV3rH:-MPBI_jm1rd9Ug9MV3rI:bcd13zk) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L88)):*
> ```Java
>       client.get();
>     } catch (Exception e) {
>       Assertions.fail();
> ```

we should let the exception be propagated/logged...

rymurr (commented):
Thanks for the review @laurentgo hopefully I have addressed everything in the most recent patch!

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575#-:-MPEkaq08q51da8mYO9j:b-5i1th)* status: all files reviewed, 28 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 68 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUY8i9azHRPeHPoWh:-MPEMXBgDQtE-8yOsgtL:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L68)):*

javadoc?
</blockquote></details>

Done.
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 78 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUK90A1ROlVq5ZQra:-MPENWIy9estDdeH-_NW:bb7f2mf) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L78)):*

(style) maybe name the method `newRequest` vs `create` (at first, I was wondering if it was creating a new client...)
</blockquote></details>

done
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java, line 69 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBDJbJ1EZNTQUKInoa:-MPENGhW7eUTtE-hRmSV:bg31ptp) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClient.java#L69)):*

If not doing automatic DI, I would still do injection before doing registration (which means the client could probably get the objectmapper injected too)
</blockquote></details>

Done, i chose not to do any automatic DI to avoid extra dependencies on the client side. Have moved the creation of the `ObjectMapper` to the `NessieClient`
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpClientException.java, line 24 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAUdW14_U4cw64ZGFr:-MPEPTBr-1jyErGPhvGK:ba0up55) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpClientException.java#L24)):*

shouldn't status code be part of the exception? or is the exception not designed to return server errors?
</blockquote></details>

This exception type is only client side exception. Eg misformed URLs or connection exceptions. Have cleaned up the usage of it to be a bit more clear.
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java, line 33 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAVOApDh37As8C4hmZ:-MPEQ48R2DyjZa2VClBo:bowaz0l) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java#L33)):*

use of exceptions doesn't seem consistent. Sometimes `IOException` is thrown, sometimes it is wrapped...
</blockquote></details>

agreed, have cleaned up the usage and most `IOException` s are now wrapped with a reason. IOExceptions that aren't are subclasses which designate specific server response codes
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java, line 36 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAVd0Q-ZKv5dWarm-Z:-MPEQE7s8GqRtj8pTn8q:bmq7pl7) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/HttpResponse.java#L36)):*

maybe a comment to explain why the method is invoked but result is not stored?
</blockquote></details>

have removed. Isn't needed.
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/RequestContext.java, line 48 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPAW96tEyZU42pcEGD6:-MPESOV01L29XZ0z71MX:b-xj4wg7) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/RequestContext.java#L48)):*

Technically HTTP headers are multi-valued...
</blockquote></details>

yes i know, I was being lazy. Fixed now and added a test
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/RequestFilter.java, line 28 at r5](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBD1iX1Cey7fro4pfk:-MPESTHy2bQmwHKleIbm:bbb2fs3) ([raw file](https://github.com/projectnessie/nessie/blob/7827aba2f5638c2945d45715722a55d3342964f1/clients/client/src/main/java/com/dremio/nessie/client/http/RequestFilter.java#L28)):*

it feels like injection should happen at construction time (which should not be a big deal)
</blockquote></details>

agreed, fixed
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 43 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBFCQB96vSd7oWN0EB:-MPETV1O0xy5heorTvtH:b-vkgfva) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L43)):*

should it check for empty values too?
also, what happens if path contains special characters like "?"?
</blockquote></details>

doing a 0 length check now. I've delegated invalid characters check to when the URL is created. A bit lazy but would rather delegate that behaviour
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 72 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGH3C3y6DLN4nlL0V:-MPETFWN9kbICbPvYtV1:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L72)):*

assumes that baseUri is not empty...
</blockquote></details>

Done.
___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 78 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGTIqBN2ZUrgdiXXg:-MPEU0Jf30ac0NpnzGLd:b-sa1nve) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L78)):*

so I guess it means that the whole path component should be a template?
</blockquote></details>

Yes, I think we can extend if we want to but for now I don't see a need. Have documented the limitation though
___
*[clients/client/src/main/java/com/dremio/nessie/client/rest/NessieHttpResponseFilter.java, line 37 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGxFcBwN92h9ZtKYC:-MPEbiS-ECJVu100_iVT:bfd6vbz) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/rest/NessieHttpResponseFilter.java#L37)):*

Maybe we should consider doing at NessieClient level instead of using some filter
</blockquote></details>

I thought about it, it creates a bunch of boilerplate and mixes concerns. I thought this way was cleaner and closer to the jax-rs impl if we ever migrate and keeps the diff noise down
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 32 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBHa1h8KZJEopuBXpj:-MPEcbB8-5RzTLbI7U9X:b28t7z1) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L32)):*

using com.sun packages doesn't help with portability... (there are probably warnings popping up about use of private packages)
</blockquote></details>

my understanding is that `sun.*` is private and `com.sun.*` is mostly ok. One of several comments: https://stackoverflow.com/questions/58764710/is-package-com-sun-net-httpserver-standard

I can switch if you prefer. I like this for its simplicity
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java, line 238 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBHz5Q-YD7ZbCYQfPH:-MPEeA5iA4LvnW5g-5az:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpClient.java#L238)):*

use a logger?
</blockquote></details>

Done.
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 76 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBI8VZ0sXhe3b9VO_n:-MPEkBUL8SZMcX3bWZxp:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L76)):*

why not using a try-with-resources pattern like the tests for http?
</blockquote></details>

Done.
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 84 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBIQVs1ZNnJg0cGYfa:-MPEkC3t6kiYQf_bMRUt:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L84)):*

That's not something we should do (even in a test) as it changes the whole context. The right thing to do is to allow the ssl context to be configured per client...
and it would be interesting to verify that an unconfigured client fails to connect because of a certificate validation issue...
</blockquote></details>

Done.
___
*[clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java, line 88 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBI_jm1rd9Ug9MV3rH:-MPEkCbuBymKcPUYA40v:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/test/java/com/dremio/nessie/client/http/TestHttpsClient.java#L88)):*

we should let the exception be propagated/logged...
</blockquote></details>

Done.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/575)* status: all files reviewed, 28 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java, line 79 at r6](https://reviewable.io/reviews/projectnessie/nessie/575#-MPBGaeJD_tI3ojuArsO:-MPEkh9B2WxItPH88qZd:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/684991cd009769cd4db97faf8d75048ac00da306/clients/client/src/main/java/com/dremio/nessie/client/http/UriBuilder.java#L79)):*

would be nice if it could be added directly to uriBuilder instead of creating some temporary string (maybe with a custom collector)
</blockquote></details>

Done.

jacques-n (approved):

PR #576, opened 2020-12-16
Fix handling of tag/branch names in the REST API paths containging ch…

snazy:
…aracters that need to be escaped.

Characters like `%`, `/`, etc need to be URI-escaped in the HTTP request URI to work properly.

The fix is rather simple: just use `javax.ws.rs.client.WebTarget#resolveTemplate(java.lang.String, java.lang.Object, boolean)` where necessary.

We probably also need some functionality to check whether a tag/branch names contains valid characters to prevent that "bad" characters (e.g. control characters) get into the systems.

Fixes #395

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/576)

PR #589, opened 2020-12-18
Introduce `resetStoreUnsafe` operation on new `AdminApi` interface.

snazy:
PR #576 introduced a function to clear the version-store for tests before each test runs.
The the only test class using this was `TestRest` by using `@Inject` to get the version-store implementation.
That however fails with `@NativeImageTest`, because native-tests don't support `@Inject`.

So the idea here is to introduce a REST functionality to programmatically reset version-store
for test and development purposes. As none of the existing interfaces (`ConfigApi`, `TreeApi`, `ContentsApi`)
are good fits for such an "reset-unsafe" operation, this PR introduces a new `AdminApi` interface.

The `AdminApi.resetStoreUnsafe()` functionality is disabled by default and needs to be explicitly
enabled in the configuration. Later on, it should probably also get some special permissions, but that's
out of scope here.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/589)

### DIFF
--- a/clients/client-test/pom.xml
+++ b/clients/client-test/pom.xml
@@ -21,19 +21,30 @@
 
   <parent>
     <groupId>org.projectnessie</groupId>
-    <artifactId>nessie</artifactId>
+    <artifactId>nessie-clients</artifactId>
     <version>0.2.2-SNAPSHOT</version>
   </parent>
 
-  <artifactId>nessie-server-parent</artifactId>
-  <packaging>pom</packaging>
+  <artifactId>nessie-client-test</artifactId>
 
-  <name>Nessie - Servers - Parent</name>
+  <name>Nessie - Client</name>
 
-  <modules>
-    <module>services</module>
-    <module>services-test</module>
-    <module>quarkus-server</module>
-    <module>lambda</module>
-  </modules>
+  <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-client</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-model-test</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+      <version>2.1.1</version>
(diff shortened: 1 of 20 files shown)


rymurr:
I don't love this approach. It seems heavy handed to add a new public REST endpoint and a potentially confusing client api call just for 1 test.

My thinking:
* what are the potential pitfalls for not testing the special character limitations in the native image? I think not much, the native image shouldn't be messing w/ encodings or anything
* if we don't need to run it in the native image I think just create another test class and don't add a native test for it
* if we do can we clear the store of just what was created in the test: just delete the refs we created rather than blowing up the whole store?

snazy:
It's actually not just for one single test. I think, that every test should start with a well-defined state (aka empty store) without the "mess" that another test left behind.

Such a functionality can also be quite handy for Nessie users that want to test their own stuff and reset the store from their test code.

(In another OSS project we've had a "base test class" that was responsible to generate unique names, keep track of changes, clean-up stuff and so on. But after some time, that class became messy, hard to maintain, difficult to adopt to new requirements and it unnecessarily slowed down all related tests.)

rymurr:
Is there any way we can put it in a test jar? So its not distributed w/ the docker images etc.

A big red 'delete everything' button on the rest api makes me squirmy even if its only implemented for 1 store

laurentgo:
I agree with @rymurr that the approach seems to heavy handed: adding a dangerous method to the public API seems too risky.

Even the current approach is not that great because it assumes using the inmemory store for tests, but this is controlled by the configuration, not by the test itself (note also that java quarkus uses the test profile by default, but the native tests uses the prod profile)

For Java tests, I look how to configure CDI during tests but looks like the use of `VersionStoreFactory` prevents this to happen, and so a bit more work would be necessary. Also, this approach would not be working for native tests (as injection is precomputed during image generation).

snazy:
> native tests uses the prod profile ... not be working for native tests (as injection is precomputed during image generation)

Ah, damn - that's right. I'll try and see if this can work with a separate jar and a bean that implements the unsafe behavior only there, so it's not even in the production code.

snazy:
(Force) pushed the current state - but at this point it's nothing that can be reviewed, not even working yet.

snazy:
At the moment I tend to skip store-clearing between individual tests of the affected tests and just close this PR. If we really need something like this, we can revisit later.

PR #600, opened 2020-12-21
Design doc for Nessie authorization

laurentgo:
Open for feedback

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/600)

### DIFF
--- a/design/authorization.md
+++ b/design/authorization.md
@@ -1,0 +1,85 @@
+﻿Authorization
+===========
+This document aims at providing some ideas regarding Nessie and authorization, namely:
+* what Nessie authorization should generally cover
+* a Java interface to perform access control, to be used by the JAX-RS server implementation
+* a proposal for an implementation to be bundled with Nessie service
+
+
+## Authorization scope
+It is important to remember that Nessie does not store data directly but only data location and other metadata. As a consequence Nessie authorization layer can only really control access to metadata, but might not prevent data itself to be accessed directly without interacting with Nessie. It is then expected that another system can control access to data itself to make sure no authorized access is made.
+The same is true for access to historical data which is one of Nessie main features. For example while it might seem safe when detecting some undesired sensitive data to commit a change to remove it and to restrict access to only the latest version of the dataset, the truth is that the sensitive data may still exist on the datalake and be accessed by other means (similar to how redacting a PDF by adding black boxes on top of sensitive information does not prevent people to read what is written beneath in most cases). The only safe method to remove this data is to remove it from the datalake, and to add commits to Nessie to point to the new data location.
+
+## Stories
+Here's a list of common authorization scenarios:
+
+* Alice attempts to execute a query against the table `Foo` on branch `prod`. As she has read access to the table on this branch, Nessie allows the execution engine to get the table details so it can locate the data and try to access it.
+* Bob attempts to execute a query against the table `Foo` on branch `prod`. However Bob does not have read access to the table. Nessie returns an authorization error, and the execution engine refuses to execute the query.
+* Carol has access to the branch `prod` content, but not to the table `Foo` on this branch. Carol creates a new reference named `carol-branch`with the same hash as `prod`, and attempts to change permission on table `Foo`. However request is denied, and Carol cannot access the content of `Foo`.
+* Dave has access to the branch `prod` content, and wants to update the content of the table `Foo`. He creates a new reference named `dave-experiment`, and executes several queries against this branch to modify table `Foo`. Each modification is a commit done against `dave-experiment` branch which is approved by Nessie server. When all the desired modifications are done, Dave attempts to merge the changes back to the `prod` branch. However Dave doesn't have the rights to modify the `prod` branch, causing Nessie to deny the request.
+
+## Access control model
+Any object in Nessie can be designated by a pair of coordinates (reference, path). It seems logical that access control is also designed around those two concepts
+
+### Reference access control
+References can be designated by name or by hash. However the practicality of using hashes for access control seems very low as those hashes are supposedly random and not known in advance. So focus will be on named references (branches and tags).
+
+Several operations can be exercised against a reference:
+* create a new named reference
+* assign a hash to a reference
+* delete a reference
+* list objects present in the tree
+* read objects content in the tree
+* commit a change against the reference
+
+### Path access control
+For a specific reference, an entity is designated by its path which is why a simple way of performing access control can be done by applying restrictions on path.
+
+Several operations can be exercised against a entity:
+* create a new entity
+* delete an entity
+* update entity's content
+
+Note that those operations combine themselves with the reference operations. For example to actually be able to update the content of an entity, user needs both permission to do the update AND to commit the change against the reference where the change will be stored
+
+### Other operations
+
+Some operations like performing garbage collection or changing access control do apply to the whole Nessie repository, and as such are control by their own specific set of permissions.
+
+
+## Server interface
+Although multiple server implementations may exist and conversely multiple access control interface may exist, this section aims to describe an interface to be used by the `JAX-RS` reference implementation of Nessie present under `servers/services`. The interface is composed of various methods for each operation to validate, which accept some access control context providing user identity and some arguments regarding the object to be accessed.
+
+More concretely, interface would look like this:
+
+    interface AccessChecker {
+      void canCreateReference(AccessContext context, NamedRef ref) throws AccessControlException;
+      void canDeleteReference(AccessContext context, NamedRef ref) throws AccessControlException;
+      [...]
+      void canReadValue(AccessContext context, Reference ref, ContentsKey key) throws AccessControlException;
+    }
+
+The `AccessContext` object passed as argument contains information regarding the overall context of the operation and will be created by the server itself:
+
+    interface AccessContext {
+      /**
+       * Provide a unique id for the operation being validated (for correlation purposes).
+       */
+      String operationId();
+
+      /**
+       * Provide the user identity.
+       */
+      Principal user();
+      
+      [...]
+    }
+
+## Server reference implementation
+
+An implementation of the `AccessChecker` interface could be written with the following characteristics:
+* Rules are written for either references or paths. Each rule contains a list of operations, and for each operation, a list of users (or group of users) and if the operation is granted or denied
+* To designate references or path, the use of wildcard is allowed. Rule ordering is based on how specific the match is. For example, `foo/bar/baz` matches both`foo/*` and `foo/bar/*`, but `foo/bar/*` is more specific, so rules under `foo/bar/*` would take precedence over `foo/*`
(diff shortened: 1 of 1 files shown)


laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/600)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @sandhyasun)

___
*[design/authorization.md, line 10 at r1](https://reviewable.io/reviews/projectnessie/nessie/600#-MP55vIV6kjwvHN_mgss-r1-10:-MP6EO919UDpK1SFMMCZ:b5cx2yw) ([raw file](https://github.com/projectnessie/nessie/blob/38d43cd8ad6ed40bf892396b1634933fba166f78/design/authorization.md#L10)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

I think you meant "no **un**authorized access is made"  ?
</blockquote></details>

yes, will change
___
*[design/authorization.md, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/600#-MP55vIV6kjwvHN_mgss-r1-43:-MP6DeOx3wOg2zuRDEj9:b9avsbk) ([raw file](https://github.com/projectnessie/nessie/blob/38d43cd8ad6ed40bf892396b1634933fba166f78/design/authorization.md#L43)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

If user1 and user2 on branch 'prod' has access to table 'foo' and user2 branches of a new branch, 'dev'  to do some work of 'foo',  user1, could change/revoke  the access on 'foo' on all users on 'prod'  while user1 is busy working on it  on 'dev'  ? Some someone who has access to 'dev' can still access 'foo' ? 
</blockquote></details>

I guess it's something I need to clarify but `user1` and `user2` do not necessary have privileges to change permission on object `foo`, although they might have access to the object. Also rules on path would apply across references to prevent also someone creating a fork to change permission (see Carol's story)

design/authorization.md
```diff
@@ -0,0 +1,85 @@
...
+Some operations like performing garbage collection or changing access control do apply to the whole Nessie repository, and as such are control by their own specific set of permissions.
+
+
+## Server interface
+Although multiple server implementations may exist and conversely multiple access control interface may exist, this section aims to describe an interface to be used by the `JAX-RS` reference implementation of Nessie present under `servers/services`. The interface is composed of various methods for each operation to validate, which accept some access control context providing user identity and some arguments regarding the object to be accessed.
+
+More concretely, interface would look like this:
+
+    interface AccessChecker {
+      void canCreateReference(AccessContext context, NamedRef ref) throws AccessControlException;
```
jacques-n, line 56: I might call these 'check' instead of 'can' since they don't return a boolean. Can sounds like a question thus expected a return value.

jacques-n (commented):
This looks like a very solid approach. Nice!

rymurr:
Nice one!

Few thoughts:
1) what do you envision the data structures for users/groups/rules that live in Nessie to look like. Do you see new rest APIs for administering users/groups or is it a utility on the client side to translate user/group ops into commits?
2) Would you use any claims from the auth token? To map users to groups for example?
3) If the authorization were to be done by a gateway (eg API Gateway for a FaaS funciton) do you see any of this being reused? i could see the FaaS gateway:
   * storing authz in another service 
   * storing authz in Nessie via this method 
   * just doing auth and passing required user info to Nessie for authz
   * or would the FaaS gateway be w/o authorization and pass everythign down to Nessie

snazy:
(Still trying to get my head around all-the-things authz in Nessie.)

---

I like the rule-matching approach. However, I'm not sure how we could prevent privilege escalation like in the following scenario:
1. Have a branch containing confidential stuff (e.g. `confidential/financials`) with a quite restricted rule-set
2. "Alice" (bad guy) knows a commit-sha in that `confidential/financials` branch and creates a new branch against that commit-sha

Does that mean, that "Alice" could get around the restrictive rules?

---

We should also define how rules inherit permissions. What I mean is for example the permission to commit to a branch. Assuming these rule sets covering different permission sets:
* `dev-branch/alice/*` allows Alice to commit to her dev-branches
* `dev-branch/*` allows all devs to create new tags

Would the above allow all devs to create tags in a branch `dev-branch/alice/foo-branch`?

---

Another (early) thought: we should have a common understanding about _how_ we uniquely identify roles and users.
Just using the user/role name can be ambiguous across different identity-providers but using identity-provider-agnostic IDs like "github/user-id" can become hard to remember. Also: usernames and email-addresses can change or be re-assigned to other people.

design/authorization.md
```diff
@@ -0,0 +1,85 @@
...
+* Bob attempts to execute a query against the table `Foo` on branch `prod`. However Bob does not have read access to the table. Nessie returns an authorization error, and the execution engine refuses to execute the query.
+* Carol has access to the branch `prod` content, but not to the table `Foo` on this branch. Carol creates a new reference named `carol-branch`with the same hash as `prod`, and attempts to change permission on table `Foo`. However request is denied, and Carol cannot access the content of `Foo`.
+* Dave has access to the branch `prod` content, and wants to update the content of the table `Foo`. He creates a new reference named `dave-experiment`, and executes several queries against this branch to modify table `Foo`. Each modification is a commit done against `dave-experiment` branch which is approved by Nessie server. When all the desired modifications are done, Dave attempts to merge the changes back to the `prod` branch. However Dave doesn't have the rights to modify the `prod` branch, causing Nessie to deny the request.
```
kylep-dremio, line 19: What happens if Dave changes the perms on `dave-experiment` to be more restrictive, and then merges back to `prod`? Which set of perms would then "win"?

design/authorization.md
```diff
@@ -0,0 +1,85 @@
...
+      Principal user();
+      
+      [...]
+    }
+
+## Server reference implementation
+
+An implementation of the `AccessChecker` interface could be written with the following characteristics:
+* Rules are written for either references or paths. Each rule contains a list of operations, and for each operation, a list of users (or group of users) and if the operation is granted or denied
+* To designate references or path, the use of wildcard is allowed. Rule ordering is based on how specific the match is. For example, `foo/bar/baz` matches both`foo/*` and `foo/bar/*`, but `foo/bar/*` is more specific, so rules under `foo/bar/*` would take precedence over `foo/*`
```
doronrosenberg, line 82: I assume that the access checks are additive - so if foo/* has COMMIT for bob and foo/bar/baz has READ for bob, bob has COMMIT readon foo/bar/baz.

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/600)* status: 0 of 1 files reviewed, 5 unresolved discussions (waiting on @doronrosenberg, @kylep-dremio, @laurentgo, and @sandhyasun)

___
*[design/authorization.md, line 19 at r1](https://reviewable.io/reviews/projectnessie/nessie/600#-MP55vIV6kjwvHN_mgss-r1-19:-MQIHXTqCJSuKDrYrUaC:b-fug8z4) ([raw file](https://github.com/projectnessie/nessie/blob/38d43cd8ad6ed40bf892396b1634933fba166f78/design/authorization.md#L19)):*
<details><summary><i>Previously, kylep-dremio wrote…</i></summary><blockquote>

What happens if Dave changes the perms on `dave-experiment` to be more restrictive, and then merges back to `prod`? Which set of perms would then "win"? Would it be that if Dave has permissions to change perms at all, and has permissions to merge, then the permissions of the merged object would be the new permissions on `prod`?
</blockquote></details>

If using the model I detailed below (permissions on references + permissions on paths), permissions on prod would win. As in this model, there's no per-object permission, merging branches would not not change permissions.
___
*[design/authorization.md, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/600#-MP55vIV6kjwvHN_mgss-r1-82:-MQIIg5c5UvFpuhe059J:bpke4pd) ([raw file](https://github.com/projectnessie/nessie/blob/38d43cd8ad6ed40bf892396b1634933fba166f78/design/authorization.md#L82)):*
<details><summary><i>Previously, doronrosenberg (Doron Rosenberg) wrote…</i></summary><blockquote>

I assume that the access checks are additive - so if foo/* has COMMIT for bob and foo/bar/baz has READ for bob, bob has COMMIT readon foo/bar/baz.
</blockquote></details>

yes, READ and COMMIT would be independent privileges (meaning COMMIT would not imply READ for example)

design/authorization.md
```diff
@@ -0,0 +1,85 @@
...
+* Dave has access to the branch `prod` content, and wants to update the content of the table `Foo`. He creates a new reference named `dave-experiment`, and executes several queries against this branch to modify table `Foo`. Each modification is a commit done against `dave-experiment` branch which is approved by Nessie server. When all the desired modifications are done, Dave attempts to merge the changes back to the `prod` branch. However Dave doesn't have the rights to modify the `prod` branch, causing Nessie to deny the request.
+
+## Access control model
+Any object in Nessie can be designated by a pair of coordinates (reference, path). It seems logical that access control is also designed around those two concepts
+
+### Reference access control
+References can be designated by name or by hash. However the practicality of using hashes for access control seems very low as those hashes are supposedly random and not known in advance. So focus will be on named references (branches and tags).
```
rymurr, line 25: what happens if we ask for (key, hash) would it always accept/always deny? Eg the Iceberg client does a minimum of isolation control by only asking for hashes (as opposed to branch/tag). Is it possible to have a hash which is a parent of two branches where the branches have different permissions for a key?

design/authorization.md
```diff
@@ -0,0 +1,85 @@
...
+      
+      [...]
+    }
+
+## Server reference implementation
+
+An implementation of the `AccessChecker` interface could be written with the following characteristics:
+* Rules are written for either references or paths. Each rule contains a list of operations, and for each operation, a list of users (or group of users) and if the operation is granted or denied
+* To designate references or path, the use of wildcard is allowed. Rule ordering is based on how specific the match is. For example, `foo/bar/baz` matches both`foo/*` and `foo/bar/*`, but `foo/bar/*` is more specific, so rules under `foo/bar/*` would take precedence over `foo/*`
+* Rules would be stored under a special branch inside Nessie itself (meaning access control over rules would be done in the same manner as regular access control), under two different entities, one for the reference rules and one for the path rules (allowing for example to do an initial reference rule check at the edge).
```
rymurr, line 83: I like storing the rules in Nessie. eat our dogfood :smile: 

I wonder if rather than storing on a separate branch (presumably on a completely empty one to start) we store in a hidden directory. This has the effect of carrying the security permissions around with the whole db. So referencing a hash will have the rules at when the hash was committed and we could stage permissions with tables/views before merging as well. I suppose its not clear to me how we would revoke permissions to a particular table on a particular hash though....

seems like there are a handful of quirks/paradoxes listed above and part of the problem is the different histories (or lack of history) between the actual contents and the acls.

nastra:
@rymurr should we go ahead and merge the doc as is? I'll update it later then at the required places

rymurr (approved):
This design for auth has now moved into the implementation stage. A few minor changes (eg interface and variable names in the access checker) will be added once the implementation is merged

Issue #602, opened 2020-12-22
Reduce the size of l1 to make InternalBranch, etc typically fit under 1K


jacques-n:
This requires adjusting the amount of items in the l1 tree as well as reducing the parents count so that we hit this number. We should do this before 0.3

jacques-n:
@rymurr let me know if you have some time to look at this today. If not, I'll take a look at it.

jacques-n:
Note, this is a non-backwards compatible metadata storage change. I think we should skip building a migration operation at this point.

rymurr:
looking now

rymurr:
The breakdown of the L1 object:
id = 20 bytes
metadata = 20 bytes
type = 2 bytes
id map = n * 20 bytes
parent list = m * 20 bytes
key list = potentially large but only  when adding/changing many tables in 1 transaction or single tables w/ large names.

for now I am ignoring the key list.

Current L1 has n=151 and m=50 giving a size of 4056 or  38 bytes short of the 4k limit
Reducing L1 to n=36 and m=10 gives a size of  956 bytes or 68 bytes short of the 1k limit

The minimum size of the key list is 25 bytes which leaves 37 bytes for a table name. "testing.nation" is already 16 bytes and I expect in practice much larger table names so while n=36 seems small I think it gives enough overhead to support the KeyList in L1.

However, we are really trying to ensure that the refs table is able to mutate at a high rate. The Ref table is sized rather differently:
id = 20 bytes
metadata = 20 bytes
type = 2 bytes
name = name length * 1 byte (assuming lower end of UTF-8 encoding)
tree = n * 20 bytes
commits = 3*20 per commit with unknown number of commits (depending on when L1 last commited)

Size = 120+20*n (assuming 20 character branch name for this purpose)
This could support n=45 and the table size would be 1020 bytes or using 36 from above 840 bytes

I have raised #607 to implement n=36 and to reduce the parent list. We can discuss if that is the right number.

PR #607, opened 2020-12-22
reduce default L1 size

rymurr:
This PR reduces default L1 size from approx 4k to approx 1k. This is to fit inside the potentially limiting 1000 WCUs per partition.

See the discussion on #602

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/607)

### DIFF
--- a/versioned/tiered/src/main/java/com/dremio/nessie/versioned/impl/L1.java
+++ b/versioned/tiered/src/main/java/com/dremio/nessie/versioned/impl/L1.java
@@ -29,7 +29,7 @@ public class L1 extends MemoizedId {
 
   private static final long HASH_SEED = 3506039963025592061L;
 
-  public static final int SIZE = 151;
+  public static final int SIZE = 43;
   public static L1 EMPTY = new L1(Id.EMPTY, new IdMap(SIZE, L2.EMPTY_ID), null, KeyList.EMPTY, ParentList.EMPTY);
   public static Id EMPTY_ID = EMPTY.getId();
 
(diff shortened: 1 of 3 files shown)


jacques-n:
Two quick thoughts: did you focus on InternalBranch size or L1 size? Our goal is for InternalBranch to hit 1k.

Also, it looks like you changed the l1 fanout size to 36. I was preferring a prime for these values for highest quality hash distribution.

rymurr:
as per discussion moved L1 id map to 43 and reverted parent list

rymurr:
@jacques-n this is failing at `ITDynamoVersionStore.mergeNoConflict` line 117. It fails when `L1.SIZE=43` but not with any of 36,41,42,44-53 or 151

rymurr:
The problem described above is because for 43 L2s the two commits in the failing test both have changes for L2 number 20. So L2 number 20 has changed on both branches in the merge.The number 43 isn't special on its own its just the only number of bins that unluckily hashes a table into the same L2 tree. 

I think that the underlying problem is in the way we handle the commits array in a Ref...will now look at what could be the problem there. Working theory is that the commits array should be cleared before a merge...

rymurr:
Well I think have narrowed it down to https://github.com/projectnessie/nessie/blob/main/versioned/tiered/src/main/java/com/dremio/nessie/versioned/impl/TieredVersionStore.java#L659

The intentions state how to get from (in the failing test) the empty commit to the merged state. They don't say how to get from the current HEAD to the merged state. I think the commit objects need to be re-written based on the new parent. Not yet sure how to do that. Probably have to create a Diff from current HEAD to now and use that as the intention.

rymurr:
@jacques-n this should be ready to go now. The intention log is now written with the correct base commit.

jacques-n:
The patch looks good. Any ideas for a test?

jacques-n (approved):

jacques-n:
Merging for 0.3. Can you open follow-on issue for test?

PR #614, opened 2020-12-23
Add IntelliJ-IDEA specific settings for .editorconfig

snazy:
Adds the IntelliJ-IDEA specific settings to `.editorconfig`.
We could add stuff into `.idea/codeStyles/` though, but the editorconfig-file feels nicer.
It's not great, as all these options are specific to IntelliJ, but better than nothing and constantly running into checkstyle-issues.

@laurentgo WDYT?

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/614)


rymurr (commented):
I personally don't love adding IDE specific configs. Not so strongly that I care enough to block. @jacques-n or @laurentgo any opinoin?

snazy:
Let's defer this one - maybe #624 is the better way

laurentgo:
tbh, I would prefer a solution working for most IDEs if possible, and more agnostic too

PR #624, opened 2020-12-28
Replace checkstyle with Spotless & Google Java Style

snazy:
https://github.com/diffplug/spotless/tree/main/plugin-maven

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/624)


pom.xml
```diff
@@ -755,9 +723,13 @@ limitations under the License.
           <useDefaultExcludes>false</useDefaultExcludes>
 
           <includes>
-            <include>src/**</include>
-            <include>*</include>
-            <include>**/.mvn/**</include>
+            <include>*.xml</include>
```
laurentgo, line 726: why is this plugin configuration changed?

laurentgo:
I'm not sure I'm a big fan of Google style (especially the 100 characters limit): it seems to force every statement to be wrapped. Are they other alternatives with good support for IDE (I checked the Eclipse plugin for google formatter, which seems to install easily although manually. I've no idea if it also works with  version of the google plugin.

snazy:
If everyone fine with this PR, we should probably merge it manually to keep the separate commits for 1. the text file + pom changes and 2. the "automatic spotless:apply".

pom.xml
snazy: TBH, I don't recall. I *think* I ran into some differences wrt copyright headers at that time, but that's obviously no longer the case, so I reverted this change.

pom.xml
```diff
@@ -607,16 +607,9 @@
...
-          <dependencies>
-            <dependency>
-              <groupId>com.puppycrawl.tools</groupId>
-              <artifactId>checkstyle</artifactId>
-              <version>8.42</version>
-            </dependency>
-          </dependencies>
+          <groupId>com.diffplug.spotless</groupId>
+          <artifactId>spotless-maven-plugin</artifactId>
+          <version>2.6.1</version>
```
nastra, line 612: nit: should we use the latest version `2.11.1` here?

nastra (approved):
LGTM (I only reviewed the first commit). I agree it would make sense to merge those 2 commits without squashing them

pom.xml
snazy: nice catch! fixed

nastra (dismissed):

nastra:
lgtm

nastra (approved):

rymurr (dismissed):
While I don't like the big change I am happy to move to spotless. @laurengo any thoughts?

@snazy is there any instructions for how to switch from checkstyle to spotless as a dev? Or is it pretty straighforward

laurentgo:
I noticed this morning that the plugin complained about some files found under the `site` directory (where I put my python virtualenv). Maybe we need some exclusions?

snazy:
> I noticed this morning that the plugin complained about some files found under the `site` directory (where I put my python virtualenv). Maybe we need some exclusions?

Huh? Haven't seen it complain about anything in `venv` or so. Do you have some details?

laurentgo:
While trying spotless more, I found a couple of issues:
- No skip flag (surprisingly): there's an open issue for it at diffplug/spotless#491
- spotless (google formatter) only looks at formatting issues, but not all issues detected by checkstyle are detected by spotless/google formatter. For example, using a camel case package name is flagged by checkstyle but not by spotless

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/624)* status: 0 of 616 files reviewed, 2 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[pom.xml, line 994 at r2](https://reviewable.io/reviews/projectnessie/nessie/624#-MaeeLfb5eWdoF5foXK4:-MaeeLfb5eWdoF5foXK5:b-dtq500) ([raw file](https://github.com/projectnessie/nessie/blob/54e9a579cc087237043bbab75573f7c0990dc67f/pom.xml#L994)):*
> ```XML
>             <format>
>               <includes>
>                 <include>**/*.xml</include>
> ```

This causes all the XML files across all modules to be checked once by the top level module and then latter by the owning module. Tbh, it feels too broad, and maybe we should only check things under `${project.basedir}/src`

snazy:
> While trying spotless more, I found a couple of issues:
> 
> * No skip flag (surprisingly): there's an open issue for it at [diffplug/spotless#491](https://github.com/diffplug/spotless/issues/491)
> * spotless (google formatter) only looks at formatting issues, but not all issues detected by checkstyle are detected by spotless/google formatter. For example, using a camel case package name is flagged by checkstyle but not by spotless
> 
> Maybe we should keep checkstyle for now? The lack of skip is kind of annoying to me (sometimes while I'm developing, I have temporary files which like logs which are often flagged by tools, but usually I use skip until I'm ready to clean up things before pushing my changes).

Moved the spotless-plugin into a new profile, which can be deactivated with `-DskipSpotless`.

Not sure whether the name-style checks are that much of a problem. It's something that can be spotted in reviews.

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/624)* status: 0 of 616 files reviewed, 2 unresolved discussions (waiting on @laurentgo)

___
*[pom.xml, line 994 at r2](https://reviewable.io/reviews/projectnessie/nessie/624#-MaeeLfb5eWdoF5foXK4:-MaiKHfw6_Kon0LTcmCD:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/54e9a579cc087237043bbab75573f7c0990dc67f/pom.xml#L994)):*

This causes all the XML files across all modules to be checked once by the top level module and then latter by the owning module. Tbh, it feels too broad, and maybe we should only check things under `${project.basedir}/src`
</blockquote></details>

Done.

snazy:
As discussed in the stand-up, I've re-added the Checkstyle plugin but removed the "pure code formatting" checks from the checkstyle-configs.

Nits:
* moved all code-style related configs to a new `codestyle/` folder (for both spotless + checkstyle)
* also added spotless to the Gradle plugin build

.idea/copyright/copy.xml
```diff
@@ -1,6 +1,6 @@
...
+    <option name="notice" value="Copyright (C) &amp;#36;today.year Dremio&#10;&#10;Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);&#10;you may not use this file except in compliance with the License.&#10;You may obtain a copy of the License at&#10;&#10;http://www.apache.org/licenses/LICENSE-2.0&#10;&#10;Unless required by applicable law or agreed to in writing, software&#10;distributed under the License is distributed on an &quot;AS IS&quot; BASIS,&#10;WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.&#10;See the License for the specific language governing permissions and&#10;limitations under the License." />
```
rymurr, line 3: does this mean once a year we will have a big diff which re-writes all headers or we will have a mixed copywrite year depending on when the source was created

snazy: Nope. It just inserts the current year, when the license is not present.
A "new year" has no effect. However, if you change the copyright text, it will complain.

rymurr: so we will have a mix of 2020, 2021 etc as the copyright year right? That sounds odd to me. Most repos have `Copyright <year>` or `Copyright <year1>-<year2>` right? And they are always uniform across the repo.

rymurr (approved):
LGTM. @laurentgo wdyt

nastra (approved):
LGTM

PR #640, opened 2020-12-31
View definition proposals

sandhyasun:
Including 2 proposals to add view related info to Nessie.  Looking for comments, suggestions and changes.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640)


jacques-n (commented):
Thanks for the first draft. I like where this is going. 

Some general comments:

- Each field needs a lot more detail about how it is defined. It should be clear to anyone interacting with the field about what it means and what are the rules around updating it when making other kinds of changes.
- We need these storage definitions to be defined in protobuf since that is the serialized version. Once we get those finalized, we can explore how to expose in various apis.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640#-:-MPqoOPp3StFWH21Mu9y:bklxr26)* status: 0 of 2 files reviewed, 14 unresolved discussions (waiting on @sandhyasun)

___
*[design/Proposal For View Definition.md, line 4 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPql3ai9BtMqLVfn4MN:-MPql3ai9BtMqLVfn4MO:byn2cut) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L4)):*
> ```Markdown
> 
> 
> > Written with [StackEdit](https://stackedit.io/).
> ```

please remove
___
*[design/Proposal For View Definition.md, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlGQq6QTMQ7RL30Bf:-MPqlGQq6QTMQ7RL30Bg:brg5bv0) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L33)):*
> ```Markdown
>    
> 
>     public class DremioView  implements SqlView {
> ```

Can you please rewrite your proposal focused on the serialized protobuf definition, not the java interface?

Note, that protobuf does not support inheritance. Let's also break this into two separate categories: things we think apply to all views and things that are very special to Dremio internal details. Then we can decide how (and if) the second category should be stored in Dremio. As such, I think this would be better if the proposal was specifically modifying the associated proto classes to see what the updated definition would be.
___
*[design/Proposal For View Definition.md, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqljLv2H4qPdyf-f7t:-MPqljLv2H4qPdyf-f7u:bjtnbp2) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L35)):*
> ```Markdown
>     public class DremioView  implements SqlView {
>     ...    
>       String datasetPath; // distinct path to the view 
> ```

I don't see why this would be necessary. It is already a part of the key that is used to store a view. Or maybe this is an artifact of being defined in java and not protobuf?
___
*[design/Proposal For View Definition.md, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlxMGD7g1mesbrmXL:-MPqlxMGD7g1mesbrmXM:b-6syhcd) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L37)):*
> ```Markdown
>       String datasetPath; // distinct path to the view 
>     
>       long datasetId; //unique ID *generic*
> ```

How is this id unique? What are the expectations of systems that are updating this view around this value? Is it okay for it to be empty?
___
*[design/Proposal For View Definition.md, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqm6Zz-6npOgYD5x9Q:-MPqm6Zz-6npOgYD5x9R:b2ok6cu) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L43)):*
> ```Markdown
>       ViewSchema schema; //Schema describing the view columns.
>     
>       long previousDatasetId; // pointer to the previous definition  used to link previous edits to the view 
> ```

Remove. Nessie already provides history interfaces. Let's not introduce a second way to figure out history.
___
*[design/Proposal For View Definition.md, line 45 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmBO0CcBECPKbiCGZ:-MPqmBO0CcBECPKbiCG_:b-94nf5d) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L45)):*
> ```Markdown
>       long previousDatasetId; // pointer to the previous definition  used to link previous edits to the view 
>     
>       List<Origin> fieldOrigins; //the original table or view columns this view is defined on (useful for join recommendations during view compilation)
> ```

This seems very Dremio specific. It also seems like it is very expensive to keep consistent. Let's move to a "Dremio" section and then reevaluate there. Need a lot more definition to how exactly this is defined so that tools can update it if they need to.
___
*[design/Proposal For View Definition.md, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmRjB-pgTxEuUpE_2:-MPqmRjB-pgTxEuUpE_3:b-9p0hmk) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L47)):*
> ```Markdown
>       List<Origin> fieldOrigins; //the original table or view columns this view is defined on (useful for join recommendations during view compilation)
>     
>       String context; // the context this view was defined in - i.e the home space, folder.
> ```

It sounds like you're talking about a relative-defined view. I support having this as a generic field but we need a lot more definition of how it is defined. I think we should also have a "resolved sql" field in addition to the "user entered sql" field defined above. Additionally, this looks like it needs to probably be similar to a Nessie key and thus should be a list of strings, not a single string.
___
*[design/Proposal For View Definition.md, line 49 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmhmH81iPvYWnv4j1:-MPqmhmH81iPvYWnv4j2:b-5qzal8) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L49)):*
> ```Markdown
>       String context; // the context this view was defined in - i.e the home space, folder.
>     
>       long lastModTimestamp; // Last time this view was modified   *generic*
> ```

I'm inclined to say this shouldn't be included. Nessie already timestamps commits. I think we should come up with a generic way to expose this information (or at least that is first thought).
___
*[design/Proposal For View Definition.md, line 51 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmpiS9sS8YqtVybfO:-MPqmpiS9sS8YqtVybfP:b-wvdlhe) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L51)):*
> ```Markdown
>       long lastModTimestamp; // Last time this view was modified   *generic*
>      
>       long userId ; // the user that last modified this view definition *generic*
> ```

It doesn't make any sense to me that this is a long. Also, need more definition to how we would fill this feel out. When it would be included versus not. What is right values versus wrong.
___
*[design/Proposal For View Definition.md, line 53 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqn0R11wqpXT0BCqw6:-MPqn0R11wqpXT0BCqw7:b-2en78d) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L53)):*
> ```Markdown
>       long userId ; // the user that last modified this view definition *generic*
>     
>       List<String> dependentHash ; // a list of dependent hashes this view was defined on to avoid GC on those hashes.
> ```

I don't understand this. I think you might mean other Nessie refs that were absolutely referenced. That seems like a fairly advanced feature. Let's tackle it as a follow on enhancement, not part of the initial work around expanding view definitions.
___
*[design/Proposal For View Definition.md, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnGhhAQGgB2OoSacL:-MPqnGhhAQGgB2OoSacM:btuz49x) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L55)):*
> ```Markdown
>       List<String> dependentHash ; // a list of dependent hashes this view was defined on to avoid GC on those hashes.
>     
>       List<string> tags; // List of tag strings of a fixed size (100) used to describe the view
> ```

Again, need more detail on who should define this, how it should be updated related to other fields, etc. 
___
*[design/Proposal For View Definition.md, line 59 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnM400YUvurQFiMQO:-MPqnM400YUvurQFiMQP:b-v6zket) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L59)):*
> ```Markdown
>     }
> 
>     class Origin {
> ```

This is very poorly specified. Please add a _lot_ more clarity about what this means.
___
*[design/Proposal For View Definition.md, line 66 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnS9a1Qk7shYkjthL:-MPqnS9a1Qk7shYkjthM:b-uyyzyr) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L66)):*
> ```Markdown
> 
> **Extending the interface**
> Since several of these fields are generic to other views, we could possibly move some of them into an abstract View class (marked with `*generic*` in the definition above)  - a middle layer that would be common to most view definitions. Other option is to leave SqlView as simple as it is now and  leave each "Dialect" to define it's own extended definition of it's view . 
> ```

As mentioned above, there is no such thing as inheritance in protobuf and that is the core specification of what nessie stores so we should focus there first.
___
*[design/Proposal for View Schema.md, line 11 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnbdc93pqIs4oedEu:-MPqnbdc93pqIs4oedEv:b-6dy33v) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal for View Schema.md#L11)):*
> ```Markdown
> (This schema is a data member of the DremioView definition )
> 
> A ViewSchema is a list of ViewFields and maintains a versionId to handle future version changes if any.
> ```

I think this is far to focused on internals of Calcite. I'd like to see something much more well-defined, in protobuf, that isn't tied to the internal concepts of a particular system. Arrow(flatbuf and c-text versions), Iceberg(avro) and Parquet(thrift) all have serializable schemas definitions. Let's start with one of those and then talk about what specifically is missing. Since Arrow is generic, I suggest we start with one of it's two definitions and then look to enhance as necessary. The data community doesn't need another completely unique serialized form of schema.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640)* status: 0 of 2 files reviewed, 14 unresolved discussions (waiting on @jacques-n and @sandhyasun)

___
*[design/Proposal For View Definition.md, line 51 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmpiS9sS8YqtVybfO:-MPs7EUgA4mqMhQnuLM4:b-63xest) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L51)):*

It doesn't make any sense to me that this is a long. Also, need more definition to how we would fill this feel out. When it would be included versus not. What is right values versus wrong.
</blockquote></details>

like the timestamp i am inclined to remove this. The user id should in theory be taken from the commit with the last modified timestamp. 
___
*[design/Proposal for View Schema.md, line 11 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnbdc93pqIs4oedEu:-MPs82pb7KSpLTtSjQf8:bfjs3ay) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal for View Schema.md#L11)):*

I think this is far to focused on internals of Calcite. I'd like to see something much more well-defined, in protobuf, that isn't tied to the internal concepts of a particular system. Arrow(flatbuf and c-text versions), Iceberg(avro) and Parquet(thrift) all have serializable schemas definitions. Let's start with one of those and then talk about what specifically is missing. Since Arrow is generic, I suggest we start with one of it's two definitions and then look to enhance as necessary. The data community doesn't need another completely unique serialized form of schema.
</blockquote></details>

+1 I am concerned about non-java clients being able to manipulate this object so something like Arrow would be my preference as well

sandhyasun (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640)* status: 0 of 2 files reviewed, 14 unresolved discussions (waiting on @jacques-n)

___
*[design/Proposal For View Definition.md, line 4 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPql3ai9BtMqLVfn4MN:-MQ4DL810tD8DaWSjjYj:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L4)):*

please remove
</blockquote></details>

Done.
___
*[design/Proposal For View Definition.md, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlGQq6QTMQ7RL30Bf:-MQ4DO8LA-tRKS6a8Ysx:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L33)):*

Can you please rewrite your proposal focused on the serialized protobuf definition, not the java interface?

Note, that protobuf does not support inheritance. Let's also break this into two separate categories: things we think apply to all views and things that are very special to Dremio internal details. Then we can decide how (and if) the second category should be stored in Dremio. As such, I think this would be better if the proposal was specifically modifying the associated proto classes to see what the updated definition would be.
</blockquote></details>

Done.
___
*[design/Proposal For View Definition.md, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqljLv2H4qPdyf-f7t:-MQ4DZF52xoL-uHlcviI:b-8l7447) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L35)):*

I don't see why this would be necessary. It is already a part of the key that is used to store a view. Or maybe this is an artifact of being defined in java and not protobuf?
</blockquote></details>

Yes it's part of the key
___
*[design/Proposal For View Definition.md, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlxMGD7g1mesbrmXL:-MQ4Db4-1HNxgJbfZ88E:b-wc4odh) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L37)):*

How is this id unique? What are the expectations of systems that are updating this view around this value? Is it okay for it to be empty?
</blockquote></details>

It is a dremio specific key. Moved it to the speicifc definition and describe howwhen it's generated. 
___
*[design/Proposal For View Definition.md, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqm6Zz-6npOgYD5x9Q:-MQ4Dkrp2_T7ltQdNqj8:b-9bw3v2) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L43)):*

Remove. Nessie already provides history interfaces. Let's not introduce a second way to figure out history.
</blockquote></details>

This isn't for tracking history as such. When a view is created and "run" via the client, before it is saved there can be an internal "system commit".  These could be removed later once the View is saved.  But there needs to be a way of distinguishing  these temp versions that are actively being edited Vs ones that are ancestors of a  saved version  which we know are inactive and can be GC'ed. Hence I saved the parent ID to link them. If a system commit belongs to a set of hashes that are in the ancestor set of  saved versions, those can be removed. Also the  client/UI may need to keep track of these versions . I know this isn't ideal but I couldn't think of another way to mark/link the ephemeral versions.   Welcome any other suggestions for a better way to mark the ephemeral versions.  
___
*[design/Proposal For View Definition.md, line 45 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmBO0CcBECPKbiCGZ:-MQ4GhMC-UbtarfYI5ib:b9th5b8) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L45)):*

This seems very Dremio specific. It also seems like it is very expensive to keep consistent. Let's move to a "Dremio" section and then reevaluate there. Need a lot more definition to how exactly this is defined so that tools can update it if they need to.
</blockquote></details>

Ok moving it to the Dremio section in this revision.  Won't the  field origin be relevant for a view definition in general  - i.e which table/column each view field resolves  to ? 
___
*[design/Proposal For View Definition.md, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmRjB-pgTxEuUpE_2:-MQ4H-Y716ss49c_DMHV:b-8pbx7d) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L47)):*

It sounds like you're talking about a relative-defined view. I support having this as a generic field but we need a lot more definition of how it is defined. I think we should also have a "resolved sql" field in addition to the "user entered sql" field defined above. Additionally, this looks like it needs to probably be similar to a Nessie key and thus should be a list of strings, not a single string.
</blockquote></details>

Right  - it's a list of keys so made it a list of strings. for now  You're suggesting a fully resolved SQL string text field instead ?  
___
*[design/Proposal For View Definition.md, line 49 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmhmH81iPvYWnv4j1:-MQ4H7DY7Dl-o9AOIPx-:bch52f9) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L49)):*

I'm inclined to say this shouldn't be included. Nessie already timestamps commits. I think we should come up with a generic way to expose this information (or at least that is first thought).
</blockquote></details>

Agree timestamp and userid can be retrieved from the commit record itself .That would be available only after the view has been committed and the commit hash is available but I suppose  that should be ok. 
___
*[design/Proposal For View Definition.md, line 51 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmpiS9sS8YqtVybfO:-MQ4Pnwi1_Kvcb7tMFRE:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L51)):*

like the timestamp i am inclined to remove this. The user id should in theory be taken from the commit with the last modified timestamp. 
</blockquote></details>

Done.
___
*[design/Proposal For View Definition.md, line 53 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqn0R11wqpXT0BCqw6:-MQ4HIii4erWNcneZZXQ:b-b3ipky) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L53)):*

I don't understand this. I think you might mean other Nessie refs that were absolutely referenced. That seems like a fairly advanced feature. Let's tackle it as a follow on enhancement, not part of the initial work around expanding view definitions.
</blockquote></details>

Yes - meant it to keep track of absolute references. We could tackle it later and enforce that views not be defined on specific versions initially. 
___
*[design/Proposal For View Definition.md, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnGhhAQGgB2OoSacL:-MQ4Hiu_C-SNoi1YWgRW:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L55)):*

Again, need more detail on who should define this, how it should be updated related to other fields, etc. 
</blockquote></details>

Done.
___
*[design/Proposal For View Definition.md, line 59 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnM400YUvurQFiMQO:-MQ4HlRqD-xmwH2-dgLS:byclk70) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L59)):*

This is very poorly specified. Please add a _lot_ more clarity about what this means.
</blockquote></details>

Added more details
___
*[design/Proposal For View Definition.md, line 66 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnS9a1Qk7shYkjthL:-MQ4HofO8PepnEYjKBzr:b-bbep6m) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L66)):*

As mentioned above, there is no such thing as inheritance in protobuf and that is the core specification of what nessie stores so we should focus there first.
</blockquote></details>

Ok. So separated the generic from specific. But we would need a  generic field in the SqlView generic definition to point to the specific implementation object  ?  
___
*[design/Proposal for View Schema.md, line 11 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnbdc93pqIs4oedEu:-MQ4IA3O2p6lyWoeEgFG:b-kpc16r) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal for View Schema.md#L11)):*

+1 I am concerned about non-java clients being able to manipulate this object so something like Arrow would be my preference as well
</blockquote></details>

For Dremio, I already tried saving the Arrow Schema alone and was able to break it with a simple view that has a cast of a varchar with a precision like 25 . Arrow defines it as a utf8 -no precision. So when the VDS is run later the Calcite planner validator complains that the precision of the field is 65526(max precision Dremio defines) and  the SQL text specifies a varchar(25). So made a choice to go with Calcite (didn't check if the difference was only this or there are more) . Let me see if the other schema formats like Avro that allow for a char precision. Would be neat to have a well defined serialized schema that already can fit our needs rather than define a whole new  schema.

jacques-n (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640)* status: 0 of 2 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @sandhyasun)

___
*[design/Proposal For View Definition.md, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlGQq6QTMQ7RL30Bf:-MQ9J7qlBu4xSCLlY7aY:b-lbvjs6) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L33)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

I was asking you to actually post them as proto files so we can iterate on actual "code" rather than embedded in a markdown.
___
*[design/Proposal For View Definition.md, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlxMGD7g1mesbrmXL:-MQ9JK5gDw5oyBVoyujY:brujduk) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L37)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

It is a dremio specific key. Moved it to the speicifc definition and describe howwhen it's generated. 
</blockquote></details>

The update doesn't answer my questions. If I update the view how do I set this correctly? What are valid and invalid values? Why couldn't this be a derived field?
___
*[design/Proposal For View Definition.md, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqm6Zz-6npOgYD5x9Q:-MQ9JUEeEzdcH0KneWCF:b-e8nyal) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L43)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

This isn't for tracking history as such. When a view is created and "run" via the client, before it is saved there can be an internal "system commit".  These could be removed later once the View is saved.  But there needs to be a way of distinguishing  these temp versions that are actively being edited Vs ones that are ancestors of a  saved version  which we know are inactive and can be GC'ed. Hence I saved the parent ID to link them. If a system commit belongs to a set of hashes that are in the ancestor set of  saved versions, those can be removed. Also the  client/UI may need to keep track of these versions . I know this isn't ideal but I couldn't think of another way to mark/link the ephemeral versions.   Welcome any other suggestions for a better way to mark the ephemeral versions.  
</blockquote></details>

This continues to feel like it does not belong. This whole concept of an internal system commit and heritage between them seems like it already exists as branches + merges in Nessie.  Let's not rebuild what we already have.
___
*[design/Proposal For View Definition.md, line 45 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmBO0CcBECPKbiCGZ:-MQ9Kh309UUgBEeb7aD1:by0in6w) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L45)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Ok moving it to the Dremio section in this revision.  Won't the  field origin be relevant for a view definition in general  - i.e which table/column each view field resolves  to ? 
</blockquote></details>

The definition still feels very ambiguous.
___
*[design/Proposal For View Definition.md, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmRjB-pgTxEuUpE_2:-MQ9JyKO4IZ0XMsIcIoU:b-1e28an) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L47)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Right  - it's a list of keys so made it a list of strings. for now  You're suggesting a fully resolved SQL string text field instead ?  
</blockquote></details>

I think it makes sense that resolved sql is a top-level field in addition to "context". A tool that doesn't have a parser could then still submit the sql to a system.
___
*[design/Proposal For View Definition.md, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnGhhAQGgB2OoSacL:-MQ9KFEU-yVLa2ufzh9r:b1ebf5h) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L55)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

This seems like a generic concept, not specific to any one tool.
___
*[design/Proposal For View Definition.md, line 66 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnS9a1Qk7shYkjthL:-MQ9LGw32pkoVcVz8zNN:b-6d1ri0) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L66)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Ok. So separated the generic from specific. But we would need a  generic field in the SqlView generic definition to point to the specific implementation object  ?  
</blockquote></details>

Let's replace the current dialect enum with a oneof {
DremioProperties, HiveProperties, etc.
}

Thoughts @laurentgo @rymurr ?
___
*[design/Proposal For View Definition.md, line 67 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MQ9L7zc7I0uMyTbwleb:-MQ9L7zc7I0uMyTbwlec:b-ocepfi) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L67)):*
> ```Markdown
>     message FieldOrigin {
>     
>     required string name = 1; // Name of the field . Can be a rename of an underlying column of the table  or an expression on the underlying base table columns  like : basetable.column1+basetable.column2.
> ```

Please proto3 syntax like the rest of Nessie. Required is not available in proto3.
___
*[design/Proposal for View Schema.md, line 11 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnbdc93pqIs4oedEu:-MQ9LVak6g0W5Bz8EEXI:b-iojwim) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal for View Schema.md#L11)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

For Dremio, I already tried saving the Arrow Schema alone and was able to break it with a simple view that has a cast of a varchar with a precision like 25 . Arrow defines it as a utf8 -no precision. So when the VDS is run later the Calcite planner validator complains that the precision of the field is 65526(max precision Dremio defines) and  the SQL text specifies a varchar(25). So made a choice to go with Calcite (didn't check if the difference was only this or there are more) . Let me see if the other schema formats like Avro that allow for a char precision. Would be neat to have a well defined serialized schema that already can fit our needs rather than define a whole new  schema. 
</blockquote></details>

If the only thing missing from Arrow is varchar precision, let's add that to Arrow somehow, not create a whole new system...

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+
+## Proposal to extend the View Definition
+**Background**
+The current  interface for a SqlView  which is a first class object in Nessie specifies a Sql text and a Dialect  and schema :
+
+    public interface SqlView extends Contents {  
+      
+      public static enum Dialect {
```
laurentgo, line 11: - What criteria to use to accept a dialect to be added? I guess the goal is interoperability but if lots of dialects are supported, it's getting harder and harder to reach that goal, with only some dialects to be supported by some tools in practice? For example how a view using a Dremio/Presto dialect be used from a spark context?
- Should dialect be versioned too? Like Hive 2.x and Hive 3.x dialects (assuming they are different)?

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+    public interface SqlView extends Contents {  
+      
+      public static enum Dialect {  
+        HIVE,  
+        SPARK,  
+        DREMIO,  
+        PRESTO  
+      }  
+      
+      String getSqlText();
```
laurentgo, line 18: I guess it's the query? any restrictions to be put on it? For example, all table reference should be absolute (and not relative to a default schema?)

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+      
+      Dialect getDialect();  
+      
+      // Schema getSchema();  
+    }
+
+While thinking of integrating Dremio with Nessie, we found we needed several more fields to fully define a view along with  a schema  (written up in a separate proposal) 
+
+**Proposal for additional fields**
+While a View is ultimately describe by just the SQL query , these  additional fields needed to describe a view (or what we refer to as a Virtual Dataset in the context of Dremio) for the DREMIO dialect .
```
laurentgo, line 28: What makes the DREMIO dialect special in that regard?

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+
+    message PSqlView {
+    
+    required string sqlText = 1 ; // SQL string  that defines the view
+    
+    string dialect = 2 ; // Describes the type of system this view is defined on - eg hive, presto, dremio etc. 
+    
+    optional ViewSchema schema = 3; // Details on the fields/column of the view, the datatypes, precision and scale. (See definition of ViewSchema)
+    
+    repeated string context = 4; // Defines the folder, space the view is defined on. Represented as list of strings which are keys  to container objects in Nessie.Eg ["@myspace","myfolder","mysubfolder"] or ["mycatalog", "myschema"]
```
laurentgo, line 44: I guess this would be the default Schema from a SQL point of view. Let's try to not use Dremio specific terminology

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+       
+    }
+    
+
+**Dremio specific  fields**
```
laurentgo, line 49: Nessie trying to be an interoperable system, it doesn't seem like a good idea to have fields designed for a specific system in mind while other systems might know how to interpret them, or how to deal with them for certain operations. For example, let's say you have some tool allowing to copy an object under a different name, I guess reusing the same datasetId might become an issue?

design/Proposal for View Schema.md
```diff
@@ -0,0 +1,57 @@
...
+## Schema definition for a View
+This is proposed  specification for a  schema definition of a view.  This is based on what a Dremio view schema would need to save along with the SQL text string describing the view. The goal is this schema could be used generally as well. 
+Described is an outline of the classes for the types and attributes for the view fields. 
+
+(This schema is a data member of the DremioView definition )
+
+A ViewSchema is a list of ViewFields and maintains a versionId to handle future version changes if any.
+
+    public final class ViewSchema implements Externalizable, Message<ViewSchema>, Schema<ViewSchema>  {      
+        private Integer versionId;
```
laurentgo, line 14: why a version for the schema? isn't that handled by Nessie git-like structure already?

design/Proposal For View Definition.md
vvellanki: One difference may be how a specific version of a table is referred to. For e.g. a View can refer to a specific version of a table in a specific branch. This particular syntax could be different in different dialects

design/Proposal For View Definition.md
```diff
@@ -0,0 +1,84 @@
...
+**Generic Vs Implementation specific View Definition**
+Some fields are  generic enough that  we could create a  definition common to view definitions of any dialect. Each "Dialect" can  then define its own extended fields needed for maintaining metadata for its internal implementation of a view (shown below for Dremio) .
+
+**Generic fields**
+
+    message PSqlView {
+    
+    required string sqlText = 1 ; // SQL string  that defines the view
+    
+    string dialect = 2 ; // Describes the type of system this view is defined on - eg hive, presto, dremio etc.
```
vvellanki, line 40: Shouldn't dialect be an enum rather than a string?

design/Proposal for View Schema.md
```diff
@@ -0,0 +1,57 @@
...
+
+A ViewSchema is a list of ViewFields and maintains a versionId to handle future version changes if any.
+
+    public final class ViewSchema implements Externalizable, Message<ViewSchema>, Schema<ViewSchema>  {      
+        private Integer versionId;  
+        private List<ViewField> field;
+     }
+A ViewField describes the attributes of the field - Simple, Struct, List, Map types are covered.
+
+    public final class ViewField implements  Externalizable, Message<ViewField>, Schema<ViewField> {
```
rymurr, line 19: As previously mentioned can we copy someone elses format? Arrow is probably the most complete but Parquet, Iceberg and potentially Hive are geared towards SQL.

Hive has a simple thrift format for its schemas (though I can't find a list of data types)
Iceberg has a well defined schema as well: http://iceberg.apache.org/spec/#schemas-and-data-types

rymurr (changes requested):
Getting there @sandhyasun. Nice one!

sandhyasun (commented):
Thanks so much  for  all  of your feedback on  the draft version. The goals of Nessie are clearer to me now so made changes accordingly.

Still need input for  resolution : 

resolvedSQL - the definition and what we should save
Proposed a new markdown/text object type.
Need to choose between hive schema and arrow schema. Hence deleted the separate proposal for a new schema definition and added a  new section in the View Definition Proposal . 

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640#-:-MQUOnD85cCExY28rI3O:b-s4sshx)* status: 0 of 1 files reviewed, 23 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @vvellanki)

___
*[design/Proposal For View Definition.md, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlGQq6QTMQ7RL30Bf:-MQUF8EZ0OzRiSE292N4:bszotjf) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L33)):*

I was asking you to actually post them as proto files so we can iterate on actual "code" rather than embedded in a markdown.
</blockquote></details>

Sorry misunderstood. Since there are a lot of comments and still some unknowns for eg which flavor of  schema we should pick, I will wait until they are resolved and then create the actual code definitions.
___
*[design/Proposal For View Definition.md, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlxMGD7g1mesbrmXL:-MQUFMSrDRbw_27IvdT0:by67l2a) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L37)):*

The update doesn't answer my questions. If I update the view how do I set this correctly? What are valid and invalid values? Why couldn't this be a derived field?
</blockquote></details>

I  have removed datasetId. Given the comments it's clear to me it does not fit saving it in Nessie. We  can instead derive it as a hash of the view schema or view contents instead.
___
*[design/Proposal For View Definition.md, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqm6Zz-6npOgYD5x9Q:-MQUG1CBEoPRv6bgkNwE:b-rh4aft) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L43)):*

This continues to feel like it does not belong. This whole concept of an internal system commit and heritage between them seems like it already exists as branches + merges in Nessie.  Let's not rebuild what we already have.
</blockquote></details>

Agree . It is a bit kludgy as well. We can handle it either internally in the UI or via a temp branch mechanism
___
*[design/Proposal For View Definition.md, line 45 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmBO0CcBECPKbiCGZ:-MQUGfzo6cNPGhfwG8nK:bzcrnmn) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L45)):*

The definition still feels very ambiguous.
</blockquote></details>

FieldOrigins is used for a Dremio specific feature that the UI provides for recommending joins based on jobs run against a view .  It gets generated whenever the View schema is updated so as long as the view is consistent the FieldOrigins shoudl also be consistent. Since  it's such a specialized  field , I am removing it now and will revisit  when that feature needs to be supported. 
___
*[design/Proposal For View Definition.md, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmRjB-pgTxEuUpE_2:-MQUHd_68fIoyX_lNyWJ:b-wt12ey) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L47)):*

I think it makes sense that resolved sql is a top-level field in addition to "context". A tool that doesn't have a parser could then still submit the sql to a system.
</blockquote></details>

I added a string for "resolvedSQL" but was unsure what exactly you had in mind. Did you mean sql text with all the underlying table and view names fully resolved ?  I didn't get the part about a system not having a parser submitting the sql. Once there is clarity I'll add the details of what you had in mind for this  top level field. 
___
*[design/Proposal For View Definition.md, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnGhhAQGgB2OoSacL:-MQUIhxe28evPp35EXTp:b-2v2q98) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L55)):*

This seems like a generic concept, not specific to any one tool.
</blockquote></details>

OK. tags seems to be more like "labels". So I added a comment field in the general definition instead.  And kept tags specific to dremio.  Comments ? 
___
*[design/Proposal For View Definition.md, line 66 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnS9a1Qk7shYkjthL:-MQUJBLB1OAou5-LVFoC:bun9bc6) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L66)):*

Let's replace the current dialect enum with a oneof {
DremioProperties, HiveProperties, etc.
}

Thoughts @laurentgo @rymurr ?
</blockquote></details>

I made this change in this version. 
___
*[design/Proposal For View Definition.md, line 7 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-7:-MQUC3IDA808EloOcNMv:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L7)):*

(style) missing new line and extra space
</blockquote></details>

Done.
___
*[design/Proposal For View Definition.md, line 11 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-11:-MQUC5qF31qrx-phllwz:b-nqtnt7) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L11)):*

- What criteria to use to accept a dialect to be added? I guess the goal is interoperability but if lots of dialects are supported, it's getting harder and harder to reach that goal, with only some dialects to be supported by some tools in practice? For example how a view using a Dremio/Presto dialect be used from a spark context?
- Should dialect be versioned too? Like Hive 2.x and Hive 3.x dialects (assuming they are different)?
</blockquote></details>

If a n engine uses very specific non-ansi SQL syntax , it will be difficult to support unless Nessie provides a translation layer to convert the view text to a standard format. I tried to cover that point in "Future enhancements"  Also made dialects a Oneof type as  one of the comments suggested. 
___
*[design/Proposal For View Definition.md, line 18 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-18:-MQUCiJV7r4coR6VfqTk:b-lbhjr) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L18)):*

I personally like the idea of absolute references here. relative references presuppose a lot (potentially) about the hierarchy of a set of views and tables and strengthens the meaning of Dialect more than we probably want. 
</blockquote></details>

Ok mentioned that specifically as a requirement. Did not say  how we will enforce it though - assume it will be some validatio on the ObjectType before we allow a commit.
___
*[design/Proposal For View Definition.md, line 28 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-28:-MQUE3Ns34M1TuD63rmH:b-ji5ghl) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L28)):*

imho we want to keep dialect strictly related to the SQL syntax. There may be a need for a `sourceSystem` or a subclass of View for a specific system but any system can read any View if they have the tools to understand a dialect.
</blockquote></details>

Agree the workd "dialect" is a bit confusing since it applies grammar or language. But I  guess view  dialect is not just about syntax but also determines special extensions or handling of views. I suppose ? So for Dremio views for example, I have retained the "tags" which are more Dremio specific.  Syntax  wise the SQl text would be the same.
___
*[design/Proposal For View Definition.md, line 40 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-40:-MQUGIeG8VH3s43n2Py0:bowmvsd) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L40)):*
<details><summary><i>Previously, vvellanki (Vivekanand Vellanki) wrote…</i></summary><blockquote>

Shouldn't dialect be an enum rather than a string?
</blockquote></details>

Yes made it a oneof  in this version. 
___
*[design/Proposal For View Definition.md, line 42 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-42:-MQUGNS-81vuAyEJOijJ:b-cfyglx) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L42)):*

under what conditions can this be included and when not? What is the end purpose of it, just an optimisation to start the query quickly? 
</blockquote></details>

I have added some comments about schema . If it is not present it can be generated and not considered an error.  It  is useful to save it for a few reasons - tools that use the schema to render details on the view;  if other objects like materializations need to know if a  view has changed the schema is useful to have. It  would be expensive to compile each time to check if the view is valid and uptodate. 
___
*[design/Proposal For View Definition.md, line 44 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-44:-MQUG_Q28AH-OXO8Edtn:b-dxw91n) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L44)):*

I guess this would be the default Schema from a SQL point of view. Let's try to not use Dremio specific terminology
</blockquote></details>

Ok. I changed the name and example - hope it's more generic  now. 
___
*[design/Proposal For View Definition.md, line 49 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-49:-MQUI79KBMfrwClH0Rw1:b92g792) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L49)):*

Not much to add other than what @laurentgo already said but...I am curious about why:

* the datasetId: its fairly ephemeral within Dremio already, I am wondering why it is required at all and if it would ever be used.
* previousDatasetId: this should imho be gotten via Nessie history (more comments about edits below)
* FieldOrigin feels like an optimisation for somethign that can be derived by the compilation of the view in Dremio...and really hard to keep up to date.
* Tags are very specific to Dremio and hard for other systems to use/update/etc. What about another object type or in another system?

Having said this I am not 100% sure how much you want to jump between Dremio's metastore and Nessie to build a table and how it will be synced across nessie hashes/branches. 
</blockquote></details>

I understand the goal better now.  The current update stores only tags for now.  If we need to add anything "custom" specific to a system, it needs to be a field that could get wiped out when there is a copy or a move - so something like a field needed only for performance. I have specified this requirement . 
___
*[design/Proposal For View Definition.md, line 55 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLQ-r2-55:-MQUIZB4AVSFzhdwvAN1:b-ify8q2) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L55)):*

I am not sure I understand the ephemeral edits. An edit prior to saving should probably be stored internally right? This is just a mechanism so the user doesn't lose work if their browser crashes isn't it? That sounds like a UI concern to me.

If they are going to be stored in Nessie doesn't it make more sense to create a short lived branch where the ephemeral commits are added. When saving the short lived branch is deleted and the 'saved' view is then transplanted onto the working branch (effectively a cheap squash)?
</blockquote></details>

I did think about maintaining the edits int he UI memory as well. Will go with that or a temp branch option. 
___
*[design/Proposal For View Definition.md, line 67 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MQ9L7zc7I0uMyTbwleb:-MQUNOHKAV1lGczbsl2O:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal For View Definition.md#L67)):*

Please proto3 syntax like the rest of Nessie. Required is not available in proto3.
</blockquote></details>

Done.
___
*[design/Proposal for View Schema.md, line 11 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqnbdc93pqIs4oedEu:-MQUKVtfEi_iRp5KLVuE:b-6rt7ft) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal for View Schema.md#L11)):*

If the only thing missing from Arrow is varchar precision, let's add that to Arrow somehow, not create a whole new system...
</blockquote></details>

Hive or Arrow are choices that would work - added the details in a separate section in the View definition proposal.   Looking for inputs on what fits in better with Nessie and we can decide. 
___
*[design/Proposal for View Schema.md, line 6 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLR-r2-6:-MQUJHWfBpkhmbunoFrX:bjnrsfg) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal for View Schema.md#L6)):*

What is the purpose of storing the Schema? who could use it for what (apologies for the repeated comments)
</blockquote></details>

Replied about this above. 
___
*[design/Proposal for View Schema.md, line 14 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLR-r2-14:-MQUKyk5CYFO-gtyMibJ:b-pj76xv) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal for View Schema.md#L14)):*

why a version for the schema? isn't that handled by Nessie git-like structure already?
</blockquote></details>

The version was introduced to handle any changes in  the schema  definitions intsef - eg if we added a new type or attribute.  An underlying system would have different handling based on what version schema a view was defined on. But in anycase, we have decided to use one of the available open source schemas so abandoning this idea for now. 
___
*[design/Proposal for View Schema.md, line 19 at r2](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqi8Wx0YeApGjedgLR-r2-19:-MQULW_qBJXMfwdJSAsY:b-ebmm9q) ([raw file](https://github.com/projectnessie/nessie/blob/fc040d9b349892fbce65be88044858d761fd6ecd/design/Proposal for View Schema.md#L19)):*

As previously mentioned can we copy someone elses format? Arrow is probably the most complete but Parquet, Iceberg and potentially Hive are geared towards SQL.

Hive has a simple thrift format for its schemas (though I can't find a list of data types)
Iceberg has a well defined schema as well: http://iceberg.apache.org/spec/#schemas-and-data-types
</blockquote></details>

Yes. Proposed hive and arrow. Iceberg is very similar to Arrow with the same issue (no precision for varchar/char types) This causes issues with calcite compiler.

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/640)* status: 0 of 1 files reviewed, 26 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, @sandhyasun, @snazy, and @vvellanki)

___
*[design/Proposal For View Definition.md, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqlGQq6QTMQ7RL30Bf:-MRL6OOo4I4GXg5_m6cT:b-v0c8w6) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L33)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

Sorry misunderstood. Since there are a lot of comments and still some unknowns for eg which flavor of  schema we should pick, I will wait until they are resolved and then create the actual code definitions.
</blockquote></details>

I think it would be easier to discuss the View definition at this point if it was in a `proto`file.
___
*[design/Proposal For View Definition.md, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/640#-MPqmRjB-pgTxEuUpE_2:-MRL8I04-GYMlBy4inWz:b64naln) ([raw file](https://github.com/projectnessie/nessie/blob/f375d7509da009c4d40d96c7948152e1e275cbea/design/Proposal For View Definition.md#L47)):*
<details><summary><i>Previously, sandhyasun (Sandhya Sundaresan) wrote…</i></summary><blockquote>

I added a string for "resolvedSQL" but was unsure what exactly you had in mind. Did you mean sql text with all the underlying table and view names fully resolved ?  I didn't get the part about a system not having a parser submitting the sql. Once there is clarity I'll add the details of what you had in mind for this  top level field. 
</blockquote></details>

I understand where the other fields above come from and my earlier comments are prob duplicated now. Regarding `context` and `resolvedSQL` I feel like we should either be fully defining the absolute sql or using `context`, `resolvedSQL` etc. but not allow both.

Given the (apparent) lack of support in other tools for context I am inclined to just use absolute SQL everywhere in Nessie and have tools handle context on their own. I think it might be confusing to other implementers about when to use which field and what needs updating when. Context could then be moved to Dremio specific properties

I would be curious to hear other opinions on context vs absolute @laurentgo @jacques-n @snazy 
___
*[design/Proposal For View Definition.md, line 53 at r3](https://reviewable.io/reviews/projectnessie/nessie/640#-MRL6XJZ-plB1nclv2jY:-MRL6XJZ-plB1nclv2jZ:b-ttmk9z) ([raw file](https://github.com/projectnessie/nessie/blob/5731d8170237ea5a37d23f385ebaa40413c8d6d5/design/Proposal For View Definition.md#L53)):*
> ```Markdown
>        string sqlText = 1 ; // SQL string  that defines the view. All table and view references must be fully qualified
>        
>        string resolvedSql = 2; //  ? 
> ```

what is this for?
___
*[design/Proposal For View Definition.md, line 68 at r3](https://reviewable.io/reviews/projectnessie/nessie/640#-MRL6k763tcI2rz9YTmN:-MRL6k77CmQJY-O1foE9:b-ehwz6i) ([raw file](https://github.com/projectnessie/nessie/blob/5731d8170237ea5a37d23f385ebaa40413c8d6d5/design/Proposal For View Definition.md#L68)):*
> ```Markdown
>        Schema viewSchema = 4; // Optional. Details on the fields/column of the view, the datatypes, precision and scale. (See definition of ViewSchema) This field gets generated when the view sql string is compiled . The output of the compilation is saved for future execution of the view.
>     
>        repeated string viewContext = 5; // Defines the namspace or schema where this view is contained as a list of strings. Systems that have a hierarchical namespace  can define the path. Some examples : ["mycatalog", "myschema"] or ["@myspace","myfolder","mysubfolder"]  
> ```

`viewContext` is odd here, its confusing as we have specifically stated the SQL tables are absolutely referenced
___
*[design/Proposal For View Definition.md, line 70 at r3](https://reviewable.io/reviews/projectnessie/nessie/640#-MRL6zg82CuWENsJnfBX:-MRL6zg82CuWENsJnfBY:b-pvdxeo) ([raw file](https://github.com/projectnessie/nessie/blob/5731d8170237ea5a37d23f385ebaa40413c8d6d5/design/Proposal For View Definition.md#L70)):*
> ```Markdown
>        repeated string viewContext = 5; // Defines the namspace or schema where this view is contained as a list of strings. Systems that have a hierarchical namespace  can define the path. Some examples : ["mycatalog", "myschema"] or ["@myspace","myfolder","mysubfolder"]  
>         
>        PWiki comment = 6; // Comment text field  that describes the view in plain text or markdown
> ```

I feel like this should probably be a different object:
1) it could be large
2) changing it typically wont happen at the same time as the view changes
3) it could contain links to system specific things like images etc which will be hard to update consistently

I am not opposed to a `PDatasetDocumentation` (or something similar) object but i think storing it outside of the view would be better. This also allows systems to store docs for tables or namespaces using the same object.
___
*[design/Proposal For View Definition.md, line 98 at r3](https://reviewable.io/reviews/projectnessie/nessie/640#-MRLEYr98bcUOPkqe2Xa:-MRLEYr98bcUOPkqe2Xb:b-5bskl2) ([raw file](https://github.com/projectnessie/nessie/blob/5731d8170237ea5a37d23f385ebaa40413c8d6d5/design/Proposal For View Definition.md#L98)):*
> ```Markdown
>     }
> 
> For the format of FieldSchema, below are 2 existing options that can be leveraged. Both are widely used and would suit the  purpose of storing the view schema. 
> ```

Which option do you prefer @sandhyasun ? Do you have any more analysis about which might be more suitable? I have more faith in the Arrow option but the Hive option appears a bit easier. Also see https://twitter.com/intjesus/status/1349117151387422721 on the topic of precision on SQL types

rymurr:
Thanks @sandhyasun I think we are getting closer. Looking forward to the next iteration!

rymurr:
ignore my comments on resolved/context. As discussed we do need all those fields.

rymurr:
Hey @sandhyasun can we close this in favour of #1046 ?

sandhyasun:
Yes I think so . Lets iterate on #1046

PR #659, opened 2021-01-07
Ignore files+folders that get generated by venv + mkdocs

snazy:
When following the instructions in the `README.md`, the Python `venv` creates a `.gitignore` in `site/` covering all files (contains a single `*` entry), which is not good.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/659)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ gradle/
 out
 gradle/wrapper/gradle-wrapper.jar
 version.txt
+
+# Python venv
+venv/
--- a/site/README.md
+++ b/site/README.md
@@ -30,9 +30,9 @@ dependencies.
   # cd to the site/ directory
   cd site/
   # setup the virtual environment (only needed once)
-  virtualenv -p $(which python3) .
+  virtualenv -p $(which python3) venv
   # activate the virtual environment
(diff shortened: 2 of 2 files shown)


site/.gitignore
```diff
@@ -0,0 +1,7 @@
+## Python venv
+/pyvenv.cfg
```
laurentgo, line 2: is the '/' prefix needed? `.gitignore` under `python` directory doesn't seem to use it.

snazy: Not strictly required, but this way it's just explicit (otherwise it would match any `pyvenv.cfg`, not just the one for "our" venv).

rymurr:
good catch @snazy 

This is actually an artifact of the way the venv is created. The more standard/pythonic way to create a venv would be:

``` bash
virtualenv -p $(which python3) venv
venv/bin/activate
```
And I believe the `venv` is already in `.gitignore` and would be ignored correctly.

I would rather we respect python best practices so would you mind updating the `README` instead?

snazy:
Yup, updated the README, but also added `venv/` to `.gitignore`

rymurr (approved):

PR #671, opened 2021-01-11
Exclude pentaho-aggdesigner-algorithm

snazy:
---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/671)


rymurr:
Does hive3 contain the dependency too?

snazy:
> Does hive3 contain the dependency too?

Don't see it with `dependency:tree` there

rymurr (approved):

PR #699, opened 2021-01-18
Upper-case + ImmutableMap

### LINKED ISSUE #691: Clarify the call order for types on new valuetypes
Trying to work with BranchCommitConsumer or com.dremio.nessie.tiered.builder.Ref are difficult for similar reasons. For BranchCommitConsumer, the unsaved/saved nature of the value can only be determined at completion. Ref is better as it has a RefType but there is no declared ordering of when that is called. I think we may want to declare stronger guarantees and always explicit subtypes so code implement a BranchCommitConsumer can do so without having to jump through hoops.

### LINKED ISSUE #692: Update Mongo PROPERTY_PRODUCERS to use ImmutableMap
Right now, the static maps for the MongoBaseValue objects use a mutable hashmap. We should build an immutable map within a single static invocation.

snazy:
Fixes #692

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/699)


rymurr (approved):
LGTM

PR #700, opened 2021-01-18
API call-order constraints for `Ref`-tag/branch + `BranchCommit`-(un)saved

### LINKED ISSUE #691: Clarify the call order for types on new valuetypes

snazy:
Updates the API of `com.dremio.nessie.tiered.builder.Ref` and `com.dremio.nessie.tiered.builder.Ref.BranchCommit` to ensure that the order of API calls is enforced by the API and not "verbally" in the Javadocs.

The implementations were not actually, only a few methods that return the "right" interface-types were added. Currently, all these new methods just return `this`, but in case that does change, the call-sites of the consumer-methods now handle the returned instances.

Fixes #691

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)


rymurr (changes requested):
> Reviewed 16 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 17 files reviewed, 2 unresolved discussions (waiting on @jacques-n and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java, line 147 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKwSuO9y8uk89oixJl:-MRKwSuO9y8uk89oixJm:b-iyt5e4) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java#L147)):*
> ```Java
>       // See com.dremio.nessie.versioned.store.dynamo.DynamoL1Consumer.addKeyMutation about a
>       // proposal to simplify this one.
>       consumer.keyMutations(attributeValue(keys, MUTATIONS).l().stream()
> ```

`forEach` instead since the output of map is being ignored?
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 254 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRL2hE37YzYU6qEbamG:-MRL2hE37YzYU6qEbamH:b-desx59) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L254)):*
> ```Java
>      * @return this consumer
>      */
>     UnsavedCommitMutations keyMutation(Key.Mutation keyMutation);
> ```

Just a general comment: the strict ordering of method calls has a bit of a code smell to me. Some doc on why would be good. Alternatively creating a utility to help enforce or relaxing would be nice.

rymurr:
lgtm, a few minor comments to improve readability

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 17 files reviewed, 2 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java, line 147 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKwSuO9y8uk89oixJl:-MRLcbBp7HJFpgEXejwA:b-4hcllm) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java#L147)):*

`forEach` instead since the output of map is being ignored?
</blockquote></details>

Ehm, the `map()`ed stream is passed to `keyMutations`

versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/versioned/impl/InternalRef.java
```diff
@@ -98,16 +98,17 @@ InternalTag getTag() {
    * Implement {@link Ref} to build an {@link InternalRef} object.
    */
   // Needs to be a package private class, otherwise class-initialization of ValueType fails with j.l.IllegalAccessError
-  static final class Builder extends EntityBuilder<InternalRef> implements Ref {
+  static final class Builder extends EntityBuilder<InternalRef> implements Ref, Ref.Tag, Ref.Branch {
```
jacques-n, line 101: I think it would be cleaner for all of these implements of Ref to actually be nested structures as opposed to implementing all the interfaces.

jacques-n (commented):
I like the updated Ref interface. Seems much safer, cleaner. I'm not a fan of the multiple inheritance of the implementations. Why I can see that you can do that, I think it would be cleaner to use subobjects.

I don't understand @rymurr's comments on the code smell. It actually feels like the interface of Ref is now much cleaner/safer because it ensures that mistakes should be avoided without depending on any reading of javadocs. 

One other thought I had on all of these, we should probably implement default implementations for all of them that just returns this. That way, a consumer that only wants a few fields doesn't have to keep getting updated if other fields are added to the BaseValue subclass interfaces.

rymurr:
> I don't understand @rymurr's comments on the code smell. It actually feels like the interface of Ref is now much cleaner/safer because it ensures that mistakes should be avoided without depending on any reading of javadocs.

Apologies, I should have been more clear. The idea of having an order that methods have to be called is squishy to me. Before it would only work if one read the javadoc and now it forces you to read the javadoc (put glibly). If the order is unavoidable then such is life, but a good javadoc would make up for it. Explaining why the order is necessary and a clear def'n of the order. I am less close to this part of the code and i have a hard time understanding just from the existing code/javadoc

snazy:
> Explaining why the order is necessary and a clear def'n of the order.

The order is mentioned in `Ref` and `Ref.BranchCommit`.

Some store implementations (actually: Mongo, which only has a `BsonWriter`/`BsonReader`) work most efficient (think: not buffering information during (de)serialization) this way. The other things at play here is that `Ref` can be either a tag or a branch, and that `BranchCommit` can be saved or unsaved. All these have different characteristics/attributes/constraints, but all are handled as "an opaque `Ref`".

versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/versioned/impl/InternalRef.java
snazy: Implementations are actually free to either return `this` or a new instance.
E.g. while all implementations of `Ref.tag()` or `Ref.branch()` currently return `this`, others may choose to return some specialized `TagBuilder` or `BranchBuilder`.

rymurr:
Lol my bad. I totally missed the 75 lines of javadoc in https://github.com/projectnessie/nessie/blob/main/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 17 files reviewed, 3 unresolved discussions (waiting on @jacques-n and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java, line 147 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKwSuO9y8uk89oixJl:-MRP6hrG6HX1cAwiVRwd:brf3huc) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java#L147)):*

Ehm, the `map()`ed stream is passed to `keyMutations`
</blockquote></details>

:facepalm:  thats what I get for reading diffs on github

rymurr (dismissed):
LGTM

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 17 files reviewed, 4 unresolved discussions (waiting on @jacques-n and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MRR-Tx90Hd3qGTcOVNH:b-4whk12) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*
> ```Java
>    * @return consumer instance for tags.
>    */
>   Ref.Tag tag();
> ```

I wondering about specializing the ref vs directly going for the tag or the branch. For example in `InternalTag` there's this method: `Ref applyToConsumer(Ref consumer)` but how does it guard against passing `Ref.Branch` for example?

jacques-n (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 17 files reviewed, 4 unresolved discussions (waiting on @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/versioned/impl/InternalRef.java, line 101 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKZrbfro1HfpRqymeN-r1-101:-MRRIUVhA7XjpF8RTkp0:b1fqnz7) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/versioned/impl/InternalRef.java#L101)):*

Implementations are actually free to either return `this` or a new instance.
E.g. while all implementations of `Ref.tag()` or `Ref.branch()` currently return `this`, others may choose to return some specialized `TagBuilder` or `BranchBuilder`.
</blockquote></details>

@snazy I understood that they are free to do either. I see the pattern used here as an anti-pattern and think our reference implementations shouldn't use it. We should have separate classes that implement tag and branch rather than a single object.

laurentgo (changes requested):
> Reviewed 4 of 29 files at r2.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 5 of 30 files reviewed, 4 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MRv_vZY2gbYHCRZTGJ1:b-7l3ba0) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

I wondering about specializing the ref vs directly going for the tag or the branch. For example in `InternalTag` there's this method: `Ref applyToConsumer(Ref consumer)` but how does it guard against passing `Ref.Branch` for example?
</blockquote></details>

ping?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 5 of 30 files reviewed, 4 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MRvfmMx28Eeekl0jPkV:be8vsu6) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

ping?
</blockquote></details>

Well, you can't pass a `Ref.Branch` there, because `Ref.Branch` doesn't implement `Ref`. (And similar for `Tag`.)
Whether it (the `Ref` in the argument) constructs a branch or tag, depends on whether you call `Ref.branch()` or `Ref.tag()`.

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 5 of 30 files reviewed, 4 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MRx0FjcBV9GFMt91VzG:b-r4qbrr) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

Well, you can't pass a `Ref.Branch` there, because `Ref.Branch` doesn't implement `Ref`. (And similar for `Tag`.)
Whether it (the `Ref` in the argument) constructs a branch or tag, depends on whether you call `Ref.branch()` or `Ref.tag()`.
</blockquote></details>

I guess that changed since my last review. That said, the pattern is kind of weird to me because Tag and Branch have very different behaviors but they are both classified as Ref. Have we look into making Tag/Branch top-level value instead of some kind of view around Ref?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 5 of 30 files reviewed, 4 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MRyBUj206FLgwRFvwD1:b-6m93r) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

I guess that changed since my last review. That said, the pattern is kind of weird to me because Tag and Branch have very different behaviors but they are both classified as Ref. Have we look into making Tag/Branch top-level value instead of some kind of view around Ref?
</blockquote></details>

That's not really doable IMO, because all you know when you read a `Ref` is only its name (via the name's hash) - and not whether that's actually a tag or branch. Also the public API often just refers to a reference name (e.g. `TreeApi.getCommitLog`or `VersionStore.toRef/toHash`). The origin of the `Ref`-abstraction is probably git itself.

versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java
```diff
@@ -45,31 +49,29 @@
 @SuppressWarnings({"unchecked", "rawtypes"})
 final class MongoSerDe {
   private static final Map<ValueType<?>, Function<BsonWriter, MongoBaseValue>> CONSUMERS;
-  private static final Map<ValueType<?>, Map<String, BiConsumer<BaseValue, BsonReader>>> DESERIALIZERS;
-
-  private static final String MONGO_ID_NAME = "_id";
+  private static final Map<ValueType<?>, BiConsumer<Document, BaseValue>> DESERIALIZERS;
```
jacques-n, line 52: A common pattern for this type of thing is just put the construction in the declaration as opposed to having a separate static block. Makes things cleaner for simple constructions like those here.

private static final Map<A, B> FOO = ImmutableMap.builder()
  .put(...,...)
  .put(...,...)
  .put(...,...)
  .build();

versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java
```diff
@@ -118,38 +117,35 @@ private void addKeysSafe(String key, AttributeValue value) {
    * Deserialize a DynamoDB entity into the given consumer.
    */
   static void toConsumer(Map<String, AttributeValue> entity, L1 consumer) {
-    consumer.id(deserializeId(entity, ID)).dt(AttributeValueUtil.getDt(entity));
+    consumer = consumer.id(deserializeId(entity, ID)).dt(AttributeValueUtil.getDt(entity));
```
jacques-n, line 120: rather than force this pattern, please update the javadoc on BaseValue to clarify that the consumer should hold shared state e.g. the user can call either the returned value or the original value and the same thing happens.

versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java
```diff
@@ -60,125 +60,173 @@
   static final String METADATA = "metadata";
   static final String KEY_LIST = "keys";
 
+  private boolean tag;
```
jacques-n, line 63: please avoid using multiple parallel booleans. Best in these situations to use an enum such as INIT, TAG, BRANCH

versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java
```diff
@@ -123,106 +111,164 @@
...
-   *
-   * @param commits The branch's commit-log receiver.
-   * @return This consumer.
-   */
-  Ref commits(Consumer<BranchCommitConsumer> commits);
+    /**
+     * Convenience method to get back to the {@link Ref}.
+     * @return owning {@link Ref}
+     */
+    Ref toRef();
```
jacques-n, line 141: I think we should maybe call this backToRef, returnToRef, parentRef or similar. "To" functions usually are verb/conversion functions. E.g. toBinary() or toJson(), not returns to parent pointers.

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 2 of 42 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java, line 120 at r3](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKZrbfro1HfpRqymeA-r3-120:-MSRn3zV7txgsEJhFNNx:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/d05273fc7b9f60a1037651f46c9381332ea75496/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL1.java#L120)):*

rather than force this pattern, please update the javadoc on BaseValue to clarify that the consumer should hold shared state e.g. the user can call either the returned value or the original value and the same thing happens.
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java, line 63 at r3](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKZrbfro1HfpRqymeD-r3-63:-MSRn5bF3kJPG_WbH9Q1:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/d05273fc7b9f60a1037651f46c9381332ea75496/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java#L63)):*

please avoid using multiple parallel booleans. Best in these situations to use an enum such as INIT, TAG, BRANCH
</blockquote></details>

Done.
___
*[versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java, line 52 at r3](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKZrbfro1HfpRqymeJ-r3-52:-MSRn8OK4QkK3qlpX9gO:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/d05273fc7b9f60a1037651f46c9381332ea75496/versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java#L52)):*

A common pattern for this type of thing is just put the construction in the declaration as opposed to having a separate static block. Makes things cleaner for simple constructions like those here.

private static final Map<A, B> FOO = ImmutableMap.builder()
  .put(...,...)
  .put(...,...)
  .put(...,...)
  .build();
</blockquote></details>

Done.
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 141 at r3](https://reviewable.io/reviews/projectnessie/nessie/700#-MRKZrbfro1HfpRqymeL-r3-141:-MSRnAbp8VKmELsUJm5H:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/d05273fc7b9f60a1037651f46c9381332ea75496/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L141)):*

I think we should maybe call this backToRef, returnToRef, parentRef or similar. "To" functions usually are verb/conversion functions. E.g. toBinary() or toJson(), not returns to parent pointers.
</blockquote></details>

Done.

laurentgo (changes requested):
> Reviewed 15 of 29 files at r2, 1 of 8 files at r3, 24 of 24 files at r4.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: all files reviewed, 14 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java, line 69 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWEy9e285NFd90ATlo:-MSWEy9e285NFd90ATlp:b7h44l9) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java#L69)):*
> ```Java
>    */
>   static void toConsumer(Map<String, AttributeValue> entity, L3 consumer) {
>     DynamoBaseValue.toConsumer(entity, consumer);
> ```

(after seeing that pattern several times, where it looks like we should have been using some inheritance, I wonder if this pattern is optimal)
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java, line 267 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFQ_D80Czba2h3ON1:-MSWFQ_D80Czba2h3ON2:bxw8jmy) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java#L267)):*
> ```Java
>   private static void deserializeCommits(Map<String, AttributeValue> map, BranchCommit cc) {
>     for (AttributeValue raw : attributeValue(map, COMMITS).l()) {
>       cc = deserializeCommit(raw.m(), cc);
> ```

what's the point of reassigning cc? (same below)
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSWFtiU1Pz4PNWJ6tBV:b-10pxbn) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*
> ```Java
>       @Override
>       public Branch branch() {
>         return new AbstractBranch(this) {
> ```

Let's avoid the anonymous class
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 153 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWG3fjCJAq8tsfCmKE:-MSWG3fjCJAq8tsfCmKF:b-vrue3h) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L153)):*
> ```Java
>                   @Override
>                   public UnsavedCommitMutations mutations() {
>                     return new AbstractUnsavedCommitMutations(branchCommit) {
> ```

that's a lot of anonymous classes intertwined. I think we could do without that many nesting levels
___
*[versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoL1.java, line 52 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWGKlZEXtUQXOHxF61:-MSWGKlZEXtUQXOHxF62:bt74jvg) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoL1.java#L52)):*
> ```Java
> 
>     Document keyList = (Document) document.get(KEY_LIST);
>     boolean chk = keyList.getBoolean(IS_CHECKPOINT);
> ```

(style) let's avoid abbreviations
___
*[versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java, line 187 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWHJhd5v40blWNEj-c:-MSWHJhd5v40blWNEj-d:b8eke85) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java#L187)):*
> ```Java
> 
>   static Stream<Key.Mutation> deserializeKeyMutations(Document d, String param) {
>     List<Document> kms = (List<Document>) d.get(param);
> ```

(style) let's avoid acronyms. I guess it stands for `keyMutations`
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MSWHmokEtzmBQpRxhSy:b-pq4c84) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

That's not really doable IMO, because all you know when you read a `Ref` is only its name (via the name's hash) - and not whether that's actually a tag or branch. Also the public API often just refers to a reference name (e.g. `TreeApi.getCommitLog`or `VersionStore.toRef/toHash`). The origin of the `Ref`-abstraction is probably git itself.
</blockquote></details>

From an external API point of view maybe, but at least for the tiered store, Branch and Tag are clearly two distinct entities...
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 254 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRL2hE37YzYU6qEbamG:-MSWIK6I-rnPCapQG73V:b-7jqkpx) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L254)):*

Just a general comment: the strict ordering of method calls has a bit of a code smell to me. Some doc on why would be good. Alternatively creating a utility to help enforce or relaxing would be nice.
</blockquote></details>

I agree this is a code smell
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSWIpA4C6j6L57icK4a:b-u1kv4e) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*
> ```Java
>   @SuppressWarnings("unchecked")
>   @Override
>   public C id(Id id) {
> ```

what's the point of adding those methods which do nothing, except the unchecked cast?

I think I would rather getting rid of this class and add the methods directly to AbstractL1, AbstractL2, etc.... At the same time, are they subclasses of AbstractL1, ... where some methods are not overridden by the final implementation?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: all files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java, line 69 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWEy9e285NFd90ATlo:-MSXJQzrA_KQdU1tXlO2:b-gbej3o) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java#L69)):*

(after seeing that pattern several times, where it looks like we should have been using some inheritance, I wonder if this pattern is optimal)
</blockquote></details>

Well, it's a `static` method. The pattern helps to not forget to handle both `id` + `dt`.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java, line 267 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFQ_D80Czba2h3ON1:-MSXJbFs0fKFb-gLM5gL:b-l1dd6g) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoRef.java#L267)):*

what's the point of reassigning cc? (same below)
</blockquote></details>

oversight, removed
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSXJeQk-hDaXT7HseTN:b-4w85eo) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*

Let's avoid the anonymous class
</blockquote></details>

That's not been added by this PR, just the addition to use abstract base classes instead of the raw interfaces was added upon request.
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 153 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWG3fjCJAq8tsfCmKE:-MSXK2LdDGzGb80gxC0w:b-4w85eo) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L153)):*

that's a lot of anonymous classes intertwined. I think we could do without that many nesting levels
</blockquote></details>

That's not been added by this PR, just the addition to use abstract base classes instead of the raw interfaces was added upon request.
___
*[versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoL1.java, line 52 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWGKlZEXtUQXOHxF61:-MSXK4BaC0JDCVlcJ462:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoL1.java#L52)):*

(style) let's avoid abbreviations
</blockquote></details>

Done.
___
*[versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java, line 187 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWHJhd5v40blWNEj-c:-MSXK5wU54koTVqDm3W-:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/mongodb/src/main/java/com/dremio/nessie/versioned/store/mongodb/MongoSerDe.java#L187)):*

(style) let's avoid acronyms. I guess it stands for `keyMutations`
</blockquote></details>

Done.
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MSXKOi0-_eSSm9leVGg:bqelqbh) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

From an external API point of view maybe, but at least for the tiered store, Branch and Tag are clearly two distinct entities...
</blockquote></details>

This would result in another huge and major refactoring. I'd rather defer this to a separate change.
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 254 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRL2hE37YzYU6qEbamG:-MSXK8MI3UZ1dLhnw8U0:bbrjvkp) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L254)):*

I agree this is a code smell
</blockquote></details>

Added some docs. It is a requirement by some store implementations to have a strict order of invocations to avoid unnecessary buffering of data during serialization.
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSXKbBg5TU0u-t48-Ye:b236hbe) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*

what's the point of adding those methods which do nothing, except the unchecked cast?

I think I would rather getting rid of this class and add the methods directly to AbstractL1, AbstractL2, etc.... At the same time, are they subclasses of AbstractL1, ... where some methods are not overridden by the final implementation?
</blockquote></details>

These classes were added upon request and those do (now) implement all methods. Those `Abstract*` classes are meant for users like the Nessie-GC - users that do not need all properties of a value.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 22 of 42 files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java, line 69 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWEy9e285NFd90ATlo:-MSYWtR_8w683YFz7WEQ:b-67sduv) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoL3.java#L69)):*

Well, it's a `static` method. The pattern helps to not forget to handle both `id` + `dt`.
</blockquote></details>

Does it? What prevents someone to forget about the call (except for careful review?)
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSYWf9CBNdObnHrS0CO:bvh10v1) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*

That's not been added by this PR, just the addition to use abstract base classes instead of the raw interfaces was added upon request.
</blockquote></details>

Sorry, not sure about what you mean
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/700#-MRR-Tx90Hd3qGTcOVNG:-MSYVjpVDLSbfnpn6qE2:b-7xnci8) ([raw file](https://github.com/projectnessie/nessie/blob/cba22f7d69748c9bcb32851869748eb4a163b88b/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/Ref.java#L118)):*

This would result in another huge and major refactoring. I'd rather defer this to a separate change.
</blockquote></details>

I'm fine doing it later
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSYXFJ6BS13PwBdOgza:bn88fyk) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*

These classes were added upon request and those do (now) implement all methods. Those `Abstract*` classes are meant for users like the Nessie-GC - users that do not need all properties of a value.
</blockquote></details>

I don't think the GC code uses AbstractBaseValue but AbstractBranch directly, so the first part of my comment seems to be still relevant?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 22 of 42 files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSYYeZ99NtnSCWE4KWt:b-xb6i00) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*

Sorry, not sure about what you mean
</blockquote></details>

There were anonymous classes before, no?
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSYYo-C8gRyH7e4EJ52:b-swppz2) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*

I don't think the GC code uses AbstractBaseValue but AbstractBranch directly, so the first part of my comment seems to be still relevant?
</blockquote></details>

Those are just helper classes for use cases that do not need all the methods.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 22 of 42 files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSYZ_ej2RTih4crRORo:b-nvd8ok) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*

There were anonymous classes before, no?
</blockquote></details>

but somehow the readability is a bit worse with the extra levels of nesting introduced with this patch?
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSYZqDSDOMyYT0cHZgE:bly9rfc) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*

Those are just helper classes for use cases that do not need all the methods.
</blockquote></details>

But the first part of my original comment was about `AbstracBaseValue` which does unchecked cast, while the same methods could be added to `AbstractL1`, where the casts would be unnecessary...

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 22 of 41 files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java, line 103 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWFtiU1Pz4PNWJ6tBU:-MSZ4q8a-0wBaZK67wYH:b-ekb3k5) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefFrame.java#L103)):*

but somehow the readability is a bit worse with the extra levels of nesting introduced with this patch?
</blockquote></details>

flattened it a bit
___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java, line 30 at r4](https://reviewable.io/reviews/projectnessie/nessie/700#-MSWIpA4C6j6L57icK4_:-MSZ-a8xAZ2TzfAotccw:b-84rtum) ([raw file](https://github.com/projectnessie/nessie/blob/da2fa1424fd3a4569c110ad0a74e683df61ba226/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/base/AbstractBaseValue.java#L30)):*

But the first part of my original comment was about `AbstracBaseValue` which does unchecked cast, while the same methods could be added to `AbstractL1`, where the casts would be unnecessary...
</blockquote></details>

`AbstracBaseValue` is gone now

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 34 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/L1.java, line 51 at r5](https://reviewable.io/reviews/projectnessie/nessie/700#-MSeI3tF5i0JiBhHYKMq:-MSeI3tF5i0JiBhHYKMr:b4fkwp0) ([raw file](https://github.com/projectnessie/nessie/blob/78ffbda896c0ca5f8e9a024c575e1428a1e50cad/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/L1.java#L51)):*
> ```Java
>    */
>   default L1 ancestors(Stream<Id> ids) {
>     ids.forEach(ignored -> {});
> ```

is there a specific reason why the stream is consumed with a terminal operation?

laurentgo (approved):
I'm okay with the changes as-is, with improvements in future patches

snazy:
Thanks for the reviews!

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/700)* status: 16 of 34 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/L1.java, line 51 at r5](https://reviewable.io/reviews/projectnessie/nessie/700#-MSeI3tF5i0JiBhHYKMq:-MSg_bza4w0lEzcql7en:bpzxtte) ([raw file](https://github.com/projectnessie/nessie/blob/78ffbda896c0ca5f8e9a024c575e1428a1e50cad/versioned/tiered/tiered-impl/src/main/java/com/dremio/nessie/tiered/builder/L1.java#L51)):*

is there a specific reason why the stream is consumed with a terminal operation?
</blockquote></details>

Yea, but maybe just paranoid. It a store implementation deserializes the properties via e.g. some BSON/JSON-reader, it may rely on the streams being consumed eagerly.

PR #545, opened 2020-12-10
Mongo load support

stevelorddremio:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/545)

PR #660, opened 2021-01-07
Mongo: Add support for Condition/UpdateExpressions

kylep-dremio:
Remove UpdateExpression ADD as it was unused.
Add visitors to allow generation of backend appropriate queries from Condition/UpdateExpressions.
Add tests for both (note that array removal relies on MongoDB 4.2 functionality, so they are disabled currently as flapdoodle only supports 4.0).

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/660)

PR #742, opened 2021-01-30
Mongo condition expression tests

stevelorddremio:
Builds off https://github.com/projectnessie/nessie/pull/545 to add tests specific to generating BSON strings.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/742)

### DIFF
--- a/versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/condition/TestExpressions.java
+++ b/versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/condition/TestExpressions.java
@@ -115,14 +115,13 @@ class TestExpressions {
   void updateSetClause() {
     UpdateExpression e0 = ImmutableUpdateExpression.builder().addClauses(
         SetClause.equals(p0, av0),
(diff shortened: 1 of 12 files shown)


snazy:
It's a bit difficult to evaluate whether the visitor-pattern works for all use-cases by just looking at these changes in isolation, although I think that the visitor-pattern could work - but maybe it needs more or maybe it needs less engineering effort. Can you provide a prototype with implementations of the `AbstractITTieredVersionStore` and `AbstractITVersionStore` to see how this fits together?

stevelorddremio (commented):
The visitors are extracted from https://github.com/projectnessie/nessie/pull/660 if that would serve as an example of them being used. We envisage using them for the RocksDBStore work too. I'm not sure they would be relevant at the `AbstractITTieredVersionStore` and `AbstractITVersionStore` level since they are used in the Store implementations.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/742#-:-MSU5ad83YMUtc8cxsBX:bdi7nax)* status: 0 of 12 files reviewed, all discussions resolved

PR #751, opened 2021-02-01
Make build+test use en/US

snazy:
Integration-tests like `TestRest` and derived native-image tests use the user's language. This lets the tests fail, when the user's language is not `en`. This PR changes the language to `en`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/751)


rymurr (approved):
LGTM!

PR #755, opened 2021-02-02
Add metrics and tracing to dynamo store

rymurr:
This does the following:

* adds a metrics publisher for [aws sdk2](https://aws.amazon.com/blogs/developer/using-the-new-client-side-metrics-feature-in-the-aws-sdk-for-java-v2/)
* adds an option to each SDK request to return consumed capacity values
* adds an interceptor to capture Dynamo consumed capacity and convert to metrics
* adds an opentracing interceptor for dynamo

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)

### DIFF
--- a/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java
+++ b/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java
@@ -16,13 +16,22 @@
 package com.dremio.nessie.versioned.store.dynamo.metrics;
 
 import java.time.Duration;
-import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.ToDoubleFunction;
+import java.util.function.ToLongFunction;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableMap;
 
 import io.micrometer.core.instrument.DistributionSummary;
 import io.micrometer.core.instrument.Metrics;
@@ -31,21 +40,30 @@ import io.micrometer.core.instrument.Timer;
 import software.amazon.awssdk.core.SdkRequest;
 import software.amazon.awssdk.core.SdkResponse;
 import software.amazon.awssdk.core.interceptor.Context.ModifyHttpRequest;
+import software.amazon.awssdk.core.interceptor.Context.ModifyRequest;
 import software.amazon.awssdk.core.interceptor.Context.ModifyResponse;
 import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
 import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.core.metrics.CoreMetric;
 import software.amazon.awssdk.http.SdkHttpRequest;
 import software.amazon.awssdk.metrics.MetricCollection;
 import software.amazon.awssdk.metrics.MetricPublisher;
+import software.amazon.awssdk.metrics.MetricRecord;
+import software.amazon.awssdk.services.dynamodb.model.BatchGetItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.BatchGetItemResponse;
 import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.BatchWriteItemResponse;
 import software.amazon.awssdk.services.dynamodb.model.ConsumedCapacity;
 import software.amazon.awssdk.services.dynamodb.model.DeleteItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.DeleteItemResponse;
+import software.amazon.awssdk.services.dynamodb.model.GetItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.GetItemResponse;
 import software.amazon.awssdk.services.dynamodb.model.PutItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.PutItemResponse;
+import software.amazon.awssdk.services.dynamodb.model.QueryRequest;
+import software.amazon.awssdk.services.dynamodb.model.QueryResponse;
+import software.amazon.awssdk.services.dynamodb.model.ReturnConsumedCapacity;
+import software.amazon.awssdk.services.dynamodb.model.ScanRequest;
 import software.amazon.awssdk.services.dynamodb.model.ScanResponse;
 import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest;
 import software.amazon.awssdk.services.dynamodb.model.UpdateItemResponse;
@@ -57,23 +75,93 @@ import software.amazon.awssdk.services.dynamodb.model.UpdateItemResponse;
  * todo add a filter on level https://micrometer.io/docs/concepts#_naming_meters
  */
 public class DynamoMetricsPublisher implements MetricPublisher {
-  private static final Map<String, Timer> TIMER_CACHE = new HashMap<>();
-  private static final Map<String, AtomicInteger> GAUGE_CACHE = new HashMap<>();
-  private static final Map<String, DistributionSummary> COUNTER_CACHE = new HashMap<>();
+  private static final Logger LOG = LoggerFactory.getLogger(DynamoMetricsPublisher.class);
+  private static final Set<Class<?>> LOG_CACHE = new HashSet<>();
+  private static final Map<String, Timer> TIMER_CACHE = new ConcurrentHashMap<>();
+  private static final Map<String, AtomicInteger> GAUGE_CACHE = new ConcurrentHashMap<>();
+  private static final Map<String, DistributionSummary> COUNTER_CACHE = new ConcurrentHashMap<>();
+  private static final String UNKNOWN = "unknown";
+
+  private static final Map<Class<? extends SdkResponse>, ToDoubleFunction<SdkResponse>> CONSUMED_CAPACITY;
+  private static final Map<Class<? extends SdkResponse>, ToLongFunction<SdkResponse>> ITEM_COUNT_RESPONSE;
+  private static final Map<Class<? extends SdkRequest>, ToLongFunction<SdkRequest>> ITEM_COUNT_REQUEST;
+  private static final Map<Class<? extends SdkRequest>, Function<SdkRequest, SdkRequest>> ADD_RETURN_CAPACITY;
+
+
+  static {
+    CONSUMED_CAPACITY = ImmutableMap.<Class<? extends SdkResponse>, ToDoubleFunction<SdkResponse>>builder()
+      .put(PutItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((PutItemResponse) r).consumedCapacity()))
+      .put(BatchWriteItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((BatchWriteItemResponse) r).consumedCapacity()))
+      .put(GetItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((GetItemResponse) r).consumedCapacity()))
+      .put(BatchGetItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((BatchGetItemResponse) r).consumedCapacity()))
+      .put(DeleteItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((DeleteItemResponse) r).consumedCapacity()))
+      .put(ScanResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((ScanResponse) r).consumedCapacity()))
+      .put(UpdateItemResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((UpdateItemResponse) r).consumedCapacity()))
+      .put(QueryResponse.class,
+        r -> DynamoMetricsPublisher.getConsumedCapacity(((QueryResponse) r).consumedCapacity()))
+      .build();
+    ADD_RETURN_CAPACITY = ImmutableMap.<Class<? extends SdkRequest>, Function<SdkRequest, SdkRequest>>builder()
+      .put(PutItemRequest.class,
+        r -> ((PutItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(BatchWriteItemRequest.class,
+        r -> ((BatchWriteItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(GetItemRequest.class,
+        r -> ((GetItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(BatchGetItemRequest.class,
+        r -> ((BatchGetItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(DeleteItemRequest.class,
+        r -> ((DeleteItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(ScanRequest.class,
+        r -> ((ScanRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(UpdateItemRequest.class,
+        r -> ((UpdateItemRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .put(QueryRequest.class,
+        r -> ((QueryRequest) r).toBuilder().returnConsumedCapacity(ReturnConsumedCapacity.TOTAL).build())
+      .build();
+    ITEM_COUNT_REQUEST = ImmutableMap.<Class<? extends SdkRequest>, ToLongFunction<SdkRequest>>builder()
+      .put(PutItemRequest.class, r -> ((PutItemRequest) r).hasItem() ? 1 : 0)
+      .put(BatchWriteItemRequest.class,
+        r -> ((BatchWriteItemRequest) r).requestItems().values().stream().map(List::size).mapToInt(x -> x).count())
+      .put(DeleteItemRequest.class, r -> ((DeleteItemRequest) r).hasKey() ? 1 : 0)
+      .put(UpdateItemRequest.class, r -> ((UpdateItemRequest) r).hasKey() ? 1 : 0)
+      .build();
+    ITEM_COUNT_RESPONSE = ImmutableMap.<Class<? extends SdkResponse>, ToLongFunction<SdkResponse>>builder()
+      .put(GetItemResponse.class, r -> ((GetItemResponse) r).hasItem() ? 1 : 0)
+      .put(BatchGetItemResponse.class,
+        r -> ((BatchGetItemResponse) r).responses()
+          .entrySet()
+          .stream()
+          .flatMap(x -> x.getValue().stream())
+          .map(Map::size)
+          .mapToInt(x -> x).count())
+      .put(ScanResponse.class, r -> ((ScanResponse) r).count())
+      .put(QueryResponse.class, r -> ((QueryResponse) r).count())
+      .build();
+  }
+
+  public DynamoMetricsPublisher() {
+  }
 
   @Override
   public void publish(MetricCollection metricCollection) {
-    AtomicReference<String> operationName = new AtomicReference<>("unknown");
-    AtomicReference<String> serviceId = new AtomicReference<>("unknown");
-    metricCollection.stream().forEach(r -> {
-      String name = r.metric().name();
-      if (name.equals("OperationName")) {
-        operationName.set(r.value().toString());
-      } else if (name.equals("ServiceId")) {
-        serviceId.set(r.value().toString());
+    String operationName = UNKNOWN;
+    String serviceId = UNKNOWN;
+    for (MetricRecord<?> metricRecord : metricCollection) {
+      String name = metricRecord.metric().name();
+      if (name.equals(CoreMetric.OPERATION_NAME.name())) {
+        operationName = metricRecord.value().toString();
+      } else if (name.equals(CoreMetric.SERVICE_ID.name())) {
+        serviceId = metricRecord.value().toString();
       }
-    });
-    publishInternal(metricCollection, operationName.get(), serviceId.get());
+    }
+    publishInternal(metricCollection, operationName, serviceId);
   }
 
   private void publishInternal(MetricCollection metricCollection, String operationName, String serviceId) {
@@ -84,19 +172,15 @@ public class DynamoMetricsPublisher implements MetricPublisher {
       String level = m.metric().level().name();
       String key = String.format("%s/%s/%s", metricName, level, operationName);
       Tags tags = Tags.of("level", level, "operation", operationName);
-      if (Duration.class.equals(m.metric().valueClass())) {
+      Class<?> valueClass = m.metric().valueClass();
+      if (Duration.class == valueClass) {
         Timer timer = TIMER_CACHE.computeIfAbsent(key, n -> getTimer(metricName, tags));
         Duration duration = (Duration) value;
         timer.record(duration.toNanos(), TimeUnit.NANOSECONDS);
-      } else if (Integer.class.equals(m.metric().valueClass()) || Boolean.class.equals(m.metric().valueClass())) {
-        Integer intValue;
-        if (Integer.class.equals(m.metric().valueClass())) {
-          intValue = (Integer) value;
-        } else {
-          Boolean boolValue = (Boolean) value;
-          intValue = boolValue ? 1 : 0;
-        }
-        measure(key, metricName, tags, (int) intValue);
+      } else if (Integer.class == valueClass) {
+        measure(key, metricName, tags, (Integer) value);
+      } else if (Boolean.class == valueClass) {
+        measure(key, metricName, tags, ((Boolean) value) ? 1 : 0);
       }
     });
     metricCollection.children().forEach(m -> publishInternal(m, operationName, serviceId));
@@ -113,11 +197,7 @@ public class DynamoMetricsPublisher implements MetricPublisher {
   private static Timer getTimer(String name, Tags tags) {
     return Timer.builder(name)
                 .tags(tags)
-                //.publishPercentiles(0.5, 0.95) // median and 95th percentile
                 .publishPercentileHistogram()
-                //.sla(Duration.ofMillis(100))
-                //.minimumExpectedValue(Duration.ofMillis(1))
-                //.maximumExpectedValue(Duration.ofSeconds(10))
                 .register(Metrics.globalRegistry);
   }
 
@@ -135,7 +215,7 @@ public class DynamoMetricsPublisher implements MetricPublisher {
                                                    .mapToDouble(Double::doubleValue).sum()).orElse((double) 0);
   }
 
-  private final void measure(String key, String metricName, Tags tags, int consumedValue) {
+  private void measure(String key, String metricName, Tags tags, int consumedValue) {
     AtomicInteger integer = GAUGE_CACHE.computeIfAbsent(key, x -> getGauge(metricName + ".last", tags));
     DistributionSummary counter = COUNTER_CACHE.computeIfAbsent(key, x -> getSummary(metricName + ".summary", tags));
     counter.record(consumedValue);
@@ -150,29 +230,29 @@ public class DynamoMetricsPublisher implements MetricPublisher {
     return new ExecutionInterceptor() {
 
       @Override
+      public SdkRequest modifyRequest(ModifyRequest context,
+                                      ExecutionAttributes executionAttributes) {
+        SdkRequest request = context.request();
+        return ADD_RETURN_CAPACITY.getOrDefault(request.getClass(), r -> r).apply(request);
+      }
+
+      @Override
       public SdkHttpRequest modifyHttpRequest(
           ModifyHttpRequest context,
           ExecutionAttributes executionAttributes) {
         SdkRequest request = context.request();
-        long count;
         String name = request.getClass().getSimpleName().replace("Request", "");
-        if (request instanceof PutItemRequest) {
-          count = ((PutItemRequest) request).hasItem() ? 1 : 0;
-        } else if (request instanceof BatchWriteItemRequest) {
-          count = ((BatchWriteItemRequest) request).requestItems().values().stream().map(List::size).mapToInt(x -> x).count();
-        } else if (request instanceof DeleteItemRequest) {
-          count = ((DeleteItemRequest) request).hasKey() ? 1 : 0;
-        } else if (request instanceof UpdateItemRequest) {
-          count = ((UpdateItemRequest) request).hasKey() ? 1 : 0;
-        } else {
-          count = 0;
-          name = null;
-        }
-        if (name != null) {
+        ToLongFunction<SdkRequest> count = ITEM_COUNT_REQUEST.getOrDefault(request.getClass(), r -> {
+          if (LOG_CACHE.add(r.getClass())) {
+            LOG.info("Could not count number of entries for request of type {}", r.getClass());
+          }
+          return -1;
+        });
+        if (count != null) {
           String metricName = "DynamoDb.RecordCount";
           String key = String.format("%s/%s", metricName, name);
           Tags tags = Tags.of("operation", name);
-          measure(key, metricName, tags, (int) count);
+          measure(key, metricName, tags, (int) count.applyAsLong(request));
         }
         return context.httpRequest();
       }
@@ -180,53 +260,32 @@ public class DynamoMetricsPublisher implements MetricPublisher {
       @Override
       public SdkResponse modifyResponse(ModifyResponse context,
                                         ExecutionAttributes executionAttributes) {
-        double consumedValue;
-        long count = -1;
+
         SdkResponse response = context.response();
         String name = response.getClass().getSimpleName().replace("Response", "");
         Tags tags = Tags.of("operation", name);
-        if (response instanceof PutItemResponse) {
-          ConsumedCapacity consumedCapacity = ((PutItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-        } else if (response instanceof BatchWriteItemResponse) {
-          List<ConsumedCapacity> consumedCapacity = ((BatchWriteItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-        } else if (response instanceof GetItemResponse) {
-          ConsumedCapacity consumedCapacity = ((GetItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-          count = ((GetItemResponse) response).hasItem() ? 1 : 0;
-        } else if (response instanceof BatchGetItemResponse) {
-          List<ConsumedCapacity> consumedCapacity = ((BatchGetItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-          count = ((BatchGetItemResponse) response).responses()
-                                                   .entrySet()
-                                                   .stream()
-                                                   .flatMap(x -> x.getValue().stream())
-                                                   .map(Map::size)
-                                                   .mapToInt(x -> x).count();
-        } else if (response instanceof DeleteItemResponse) {
-          ConsumedCapacity consumedCapacity = ((DeleteItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-        } else if (response instanceof ScanResponse) {
-          ConsumedCapacity consumedCapacity = ((ScanResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-          count = ((ScanResponse) response).count();
-        } else if (response instanceof UpdateItemResponse) {
-          ConsumedCapacity consumedCapacity = ((UpdateItemResponse) response).consumedCapacity();
-          consumedValue = getConsumedCapacity(consumedCapacity);
-        } else {
-          consumedValue = 0;
-          name = null;
-        }
-        if (name != null) {
+
+        ToLongFunction<SdkResponse> count = ITEM_COUNT_RESPONSE.getOrDefault(response.getClass(), r -> {
+          if (LOG_CACHE.add(r.getClass())) {
+            LOG.info("Could not count number of entries for response of type {}", r.getClass());
+          }
+          return -1;
+        });
+        ToDoubleFunction<SdkResponse> consumedValue = CONSUMED_CAPACITY.getOrDefault(response.getClass(), r -> {
+          if (LOG_CACHE.add(r.getClass())) {
+            LOG.info("Could not get returned used capacity for response of type {}", r.getClass());
+          }
+          return -1;
+        });
+        if (consumedValue != null) {
           String metricName = "DynamoDb.ConsumedCapacity";
           String key = String.format("%s/%s", metricName, name);
-          measure(key, metricName, tags, (int) consumedValue);
+          measure(key, metricName, tags, (int) consumedValue.applyAsDouble(response));
         }
-        if (count > -1) {
+        if (count != null) {
           String metricName = "DynamoDb.RecordCount";
           String key = String.format("%s/%s", metricName, name);
-          measure(key, metricName, tags, (int) count);
+          measure(key, metricName, tags, (int) count.applyAsLong(response));
         }
         return response;
       }
--- a/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java
+++ b/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.dremio.nessie.versioned.store.dynamo.metrics;
+
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.micrometer.core.instrument.DistributionSummary;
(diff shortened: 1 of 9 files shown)


laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on @rymurr and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 138 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSde9kd84KO4eRABfNI:-MSde9kd84KO4eRABfNJ:bjgy0zf) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L138)):*
> ```Java
>     try {
>       DynamoMetricsPublisher publisher = new DynamoMetricsPublisher();
>       TracingExecutionInterceptor tracing = new TracingExecutionInterceptor(GlobalTracer.get());
> ```

I wonder if this should be resolved when the plugin is started vs at the time it's being accessed. Being a global state, it means it could be changed later on, but the store would not pick up the change.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 205 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdehb35RozVKJnIc8t:-MSdehb35RozVKJnIc8u:b2d4pd2) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L205)):*
> ```Java
>         }));
> 
>         BatchGetItemResponse response = client.batchGetItem(BatchGetItemRequest.builder().requestItems(loads).returnConsumedCapacity(
> ```

shouldn't the interceptor do it?
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdjqi11HQT7iuNvvp8:-MSdjqi11HQT7iuNvvp9:b-d0patt) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L60)):*
> ```Java
>  */
> public class DynamoMetricsPublisher implements MetricPublisher {
>   private static final Map<String, Timer> TIMER_CACHE = new HashMap<>();
> ```

this is definitively not thread safe...
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 67 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdgvKY9XXJACGmlKwo:-MSdgvKZBCa-uqR6X8ss:b-9mqzm8) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L67)):*
> ```
>     AtomicReference<String> operationName = new AtomicReference<>("unknown");
>     AtomicReference<String> serviceId = new AtomicReference<>("unknown");
> ```
trick: you can use an single element array too

That said, it looks like you could do `String operationName = Iterables.getFirst(metricCollection.metricValue(CoreMetric.OperationName), "unknown")`
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 70 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdh9bI5hc90_vUfvTP:-MSdh9bI5hc90_vUfvTQ:ba7llfq) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L70)):*
> ```Java
>     metricCollection.stream().forEach(r -> {
>       String name = r.metric().name();
>       if (name.equals("OperationName")) {
> ```

(style) let's use a switch + string constants (doesn't AWS have some already)?

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 4 files reviewed, 19 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 144 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgpCVr-5SbJRJv3G5W:-MSgpCVr-5SbJRJv3G5X:bcx0d9p) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L144)):*
> ```Java
>       b1.httpClient(UrlConnectionHttpClient.create());
>       DynamoDbAsyncClientBuilder b2 =
>           DynamoDbAsyncClient.builder().overrideConfiguration(x -> x.addExecutionInterceptor(publisher.interceptor())
> ```

Would be nicer to have a method that applies this here and above.

Having a configuration to opt-out would be nice. I.e. tracing+metrics are enabled by default, but people can turn it off, just in case.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdjqi11HQT7iuNvvp8:-MSgi2sq4UEi7cVwtHiB:b-iy8otk) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L60)):*

this is definitively not thread safe...
</blockquote></details>

yup, maybe just use a `ConcurrentHashMap`
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkx1yDZTdUWgez5CQ:-MSgkx1yDZTdUWgez5CR:b-xmy6st) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L63)):*
> ```Java
>   private static final Map<String, AtomicInteger> GAUGE_CACHE = new HashMap<>();
>   private static final Map<String, DistributionSummary> COUNTER_CACHE = new HashMap<>();
> ```

Hint: Better add a default, no-arg constructor. Future Java versions will require that.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 68 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgiX_l8FZ8KQGt9TqB:-MSgiX_l8FZ8KQGt9TqC:b-xvteoq) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L68)):*
> ```Java
>     AtomicReference<String> operationName = new AtomicReference<>("unknown");
>     AtomicReference<String> serviceId = new AtomicReference<>("unknown");
>     metricCollection.stream().forEach(r -> {
> ```

You can change it to an "old school" `for (MetricRecord<?> r : metricCollecton)`, so you save both the two `AtomicReference` instances, the unnecessary "volatile" read-writes and the spliterator/stream-instances. `software.amazon.awssdk.metrics.MetricCollection#stream` always delegates to a `Spliterator`.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 87 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgjg8n0xPWPTfe5o7d:-MSgjg8n0xPWPTfe5o7e:bc6694u) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L87)):*
> ```Java
>       String key = String.format("%s/%s/%s", metricName, level, operationName);
>       Tags tags = Tags.of("level", level, "operation", operationName);
>       if (Duration.class.equals(m.metric().valueClass())) {
> ```

It's a `Class`, mind changing this to `==`?
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 87 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkC5fBIOYq6oH3Usw:-MSgkC5fBIOYq6oH3Usx:b-jcbzdn) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L87)):*
> ```Java
>       String key = String.format("%s/%s/%s", metricName, level, operationName);
>       Tags tags = Tags.of("level", level, "operation", operationName);
>       if (Duration.class.equals(m.metric().valueClass())) {
> ```

Can you get `m.metric().valueClass()` into a separate variable and use that all the way along below?
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 91 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgjkw37_GxP_8MLaZ1:-MSgjkw4Dz2S1L8moe1u:b5c3ob6) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L91)):*
> ```Java
>         Duration duration = (Duration) value;
>         timer.record(duration.toNanos(), TimeUnit.NANOSECONDS);
>       } else if (Integer.class.equals(m.metric().valueClass()) || Boolean.class.equals(m.metric().valueClass())) {
> ```

It's a `Class`, mind changing these to `==`?
I'd prefer to split this into two distinct branches instead of ("if (c == Integer || c == Boolean) { ... if (c == Integer) { ... } else { ... } ... }")
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 92 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkcMz31o6BmcirOr4:-MSgkcMz31o6BmcirOr5:bum9esl) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L92)):*
> ```Java
>         timer.record(duration.toNanos(), TimeUnit.NANOSECONDS);
>       } else if (Integer.class.equals(m.metric().valueClass()) || Boolean.class.equals(m.metric().valueClass())) {
>         Integer intValue;
> ```

use the primitive `int` here
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 97 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkiKACacMZM3PPjOR:-MSgkiKACacMZM3PPjOS:b-tqn9sn) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L97)):*
> ```Java
>         } else {
>           Boolean boolValue = (Boolean) value;
>           intValue = boolValue ? 1 : 0;
> ```

use the primitive `boolean` here and inline into a single line
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 99 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSghzfx2n9xPxj9ujPN:-MSghzfx2n9xPxj9ujPO:bxof3vn) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L99)):*
> ```Java
>           intValue = boolValue ? 1 : 0;
>         }
>         measure(key, metricName, tags, (int) intValue);
> ```

nit: unnecessary cast
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSghTrp8J0PGg7f6WMC:-MSghTrp8J0PGg7f6WMD:b-hbexfk) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L118)):*
> ```Java
>                 //.publishPercentiles(0.5, 0.95) // median and 95th percentile
>                 .publishPercentileHistogram()
>                 //.sla(Duration.ofMillis(100))
> ```

Does it make sense to use `serviceLevelObjectives()` here with a configurable list of `Duration`s?
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 120 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSghCxB17V53SyLnyic:-MSghCxC5Wqb19ITti38:b-dofoph) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L120)):*
> ```Java
>                 //.sla(Duration.ofMillis(100))
>                 //.minimumExpectedValue(Duration.ofMillis(1))
>                 //.maximumExpectedValue(Duration.ofSeconds(10))
> ```

Nit: remove the commented options
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 138 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgl89DDj3pUrMca3vg:-MSgl89DDj3pUrMca3vh:b-6ysmey) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L138)):*
> ```Java
>   }
> 
>   private final void measure(String key, String metricName, Tags tags, int consumedValue) {
> ```

Nit: unnecessary `final`
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 167 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgle094iopj53TycXx:-MSgle0A35PgXXM4QWQz:b-1kahpm) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L167)):*
> ```Java
>         } else if (request instanceof UpdateItemRequest) {
>           count = ((UpdateItemRequest) request).hasKey() ? 1 : 0;
>         } else {
> ```

Can you also add all potential request types (e.g. `BatchGetRequest` here as well.
Given the large-ish number of `instanceof` here, maybe a statically initialized `Map<Class<? extends DynamoDbRequest>, ToIntFunction<? extends DynamoDbRequest>>` could be more more efficient. Since all the `instanceof` refer to `final` classes, that should work. That way you can also get rid of the `if (name != null)` below.

And maybe add a "non-spamming" log that logs all unhandled request types (once per class should be okay).
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 218 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgnR9e-GG31mzDjlkG:-MSgnR9e-GG31mzDjlkH:b-u7aqzq) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L218)):*
> ```Java
>           consumedValue = getConsumedCapacity(consumedCapacity);
>         } else {
>           consumedValue = 0;
> ```

Given the large-ish number of `instanceof` here, maybe a statically initialized `Map<Class<? extends DynamoDbRequest>, Consumer<? extends DynamoDbRequest>>` could be more efficient. Since all the `instanceof` refer to `final` classes, that should work. That way you can also get rid of the `if (name != null)` and `if (count > -1)`below.

And maybe add a "non-spamming" log that logs all unhandled request types (once per class should be okay).
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/TracingExecutionInterceptor.java, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSggE8r-7D6uPFhOU5T:-MSggE8r-7D6uPFhOU5U:bmcp7r) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/TracingExecutionInterceptor.java#L35)):*
> ```Java
> 
> /**
>  * Taken from https://github.com/opentracing-contrib/java-aws-sdk as its version of opentracing and Quarkus' collide.
> ```

Can you add the original copyright-header below ours?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 4 files reviewed, 17 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 138 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSde9kd84KO4eRABfNI:-MSgAdYe32n9uvSqHpOG:b-k5k4l1) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L138)):*

I wonder if this should be resolved when the plugin is started vs at the time it's being accessed. Being a global state, it means it could be changed later on, but the store would not pick up the change.
</blockquote></details>

not sure what you mean, sorry. The `Tracer` is needed to add the interceptor which is needed at construction time of the client.

We could theoretically override every dynamo call individually to add the interceptors but that adds a lot of noise.

Finally, I think once the tracer is started i doubt it will be changing, its certainly outside the scope of what Quarkus does
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 144 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgpCVr-5SbJRJv3G5W:-MSgxOQ1EV1PyZF2llKy:bwco60l) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L144)):*

Would be nicer to have a method that applies this here and above.

Having a configuration to opt-out would be nice. I.e. tracing+metrics are enabled by default, but people can turn it off, just in case.
</blockquote></details>

added the method, it looks kinda goofy because the two builders dont have a base interface so its not easy to make it nicely reusable. 

Added a config
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 205 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdehb35RozVKJnIc8t:-MSh11UZ43SJBD-6wTh4:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L205)):*

shouldn't the interceptor do it?
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdjqi11HQT7iuNvvp8:-MSh18x-EGwAiHbsEk2o:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L60)):*

yup, maybe just use a `ConcurrentHashMap`
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkx1yDZTdUWgez5CQ:-MSh1C4a3b_Ck5ZlQqni:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L63)):*

Hint: Better add a default, no-arg constructor. Future Java versions will require that.
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 67 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdgvKY9XXJACGmlKwo:-MSh2ZJZAst8MIL4mcul:b4u4qyy) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L67)):*

> ```
>     AtomicReference<String> operationName = new AtomicReference<>("unknown");
>     AtomicReference<String> serviceId = new AtomicReference<>("unknown");
> ```
trick: you can use an single element array too

That said, it looks like you could do `String operationName = Iterables.getFirst(metricCollection.metricValue(CoreMetric.OperationName), "unknown")`
</blockquote></details>

switched to `for` loop
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 68 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgiX_l8FZ8KQGt9TqB:-MSh2a6m5xD9hVbQ4Bq4:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L68)):*

You can change it to an "old school" `for (MetricRecord<?> r : metricCollecton)`, so you save both the two `AtomicReference` instances, the unnecessary "volatile" read-writes and the spliterator/stream-instances. `software.amazon.awssdk.metrics.MetricCollection#stream` always delegates to a `Spliterator`.
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 70 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdh9bI5hc90_vUfvTP:-MSh2l8r8Fj6NB3kFGnh:b-p3ruk3) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L70)):*

(style) let's use a switch + string constants (doesn't AWS have some already)?
</blockquote></details>

They don't have string constants that I am aware of. Switched to using the static metrics though
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 87 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgjg8n0xPWPTfe5o7d:-MSh34SM7WcQ4Up2VE9i:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L87)):*

It's a `Class`, mind changing this to `==`?
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 87 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkC5fBIOYq6oH3Usw:-MSh3E3x2RpKjoCyCd-K:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L87)):*

Can you get `m.metric().valueClass()` into a separate variable and use that all the way along below?
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 91 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgjkw37_GxP_8MLaZ1:-MSh3UTpBPf231nbpTvn:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L91)):*

It's a `Class`, mind changing these to `==`?
I'd prefer to split this into two distinct branches instead of ("if (c == Integer || c == Boolean) { ... if (c == Integer) { ... } else { ... } ... }")
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 92 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkcMz31o6BmcirOr4:-MSh3sqc0AjQiIwNYY0D:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L92)):*

use the primitive `int` here
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 97 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgkiKACacMZM3PPjOR:-MSh3sYW3s0TKqxPKYzJ:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L97)):*

use the primitive `boolean` here and inline into a single line
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSghTrp8J0PGg7f6WMC:-MSh6cpq8BhGyQbOVvT-:b-3zd0fb) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L118)):*

Does it make sense to use `serviceLevelObjectives()` here with a configurable list of `Duration`s?
</blockquote></details>

for the time being I have left off...curious what others think? I don't really want to complicate the config and I dont know what our SLOs should be. Have just left the percentile histogram, which should be enough for us now
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 138 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgl89DDj3pUrMca3vg:-MSh6ujSErdy4aW6Qzqa:bhj9wt4) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L138)):*

Nit: unnecessary `final`
</blockquote></details>

fixed, curious about what I was thinking when I put that there ;-)
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 167 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgle094iopj53TycXx:-MShAug532Yo3iA0Gk4L:bskoue) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L167)):*

Can you also add all potential request types (e.g. `BatchGetRequest` here as well.
Given the large-ish number of `instanceof` here, maybe a statically initialized `Map<Class<? extends DynamoDbRequest>, ToIntFunction<? extends DynamoDbRequest>>` could be more more efficient. Since all the `instanceof` refer to `final` classes, that should work. That way you can also get rid of the `if (name != null)` below.

And maybe add a "non-spamming" log that logs all unhandled request types (once per class should be okay).
</blockquote></details>

The only one missing was Query, added. Switched to a map structure also. I prefer that idiom in python but its not very pretty for this use case. However it does clean up the business logic a lot
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 218 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgnR9e-GG31mzDjlkG:-MShI1z75xBitQGiOQVH:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L218)):*

Given the large-ish number of `instanceof` here, maybe a statically initialized `Map<Class<? extends DynamoDbRequest>, Consumer<? extends DynamoDbRequest>>` could be more efficient. Since all the `instanceof` refer to `final` classes, that should work. That way you can also get rid of the `if (name != null)` and `if (count > -1)`below.

And maybe add a "non-spamming" log that logs all unhandled request types (once per class should be okay).
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/TracingExecutionInterceptor.java, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSggE8r-7D6uPFhOU5T:-MShIBG_AuCCNqHRgm3h:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/TracingExecutionInterceptor.java#L35)):*

Can you add the original copyright-header below ours?
</blockquote></details>

Done.

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 144 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgpCVr-5SbJRJv3G5W:-MShjPPo8ph65cgukffi:b-pggz11) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L144)):*

added the method, it looks kinda goofy because the two builders dont have a base interface so its not easy to make it nicely reusable. 

Added a config
</blockquote></details>

ah, bummer. didn't notice that those dynamo-API don't share a common interface for that `overrideConfiguration`. Yea, it looks a bit ugly. I wouldn't mind to revert this then.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 67 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSdgvKY9XXJACGmlKwo:-MShl1w7Da4fn5OGneNV:bgf827c) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L67)):*

switched to `for` loop
</blockquote></details>

Much nicer :)
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSghTrp8J0PGg7f6WMC:-MShlHMb0BXI-uZSAnlJ:bb6vmix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L118)):*

for the time being I have left off...curious what others think? I don't really want to complicate the config and I dont know what our SLOs should be. Have just left the percentile histogram, which should be enough for us now
</blockquote></details>

Fair
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 93 at r2](https://reviewable.io/reviews/projectnessie/nessie/755#-MShkmVY5EsfkqhNh4n6:-MShkmVY5EsfkqhNh4n7:byii8mb) ([raw file](https://github.com/projectnessie/nessie/blob/02f72da991dd41ff120546d80bcdf523e79e0949/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L93)):*
> ```Java
>   static {
>     CONSUMED_CAPACITY = ImmutableMap.<Class<? extends SdkResponse>, ToDoubleFunction<SdkResponse>>builder()
>       .put(PutItemResponse.class,
> ```

Would this look a bit nicer, when the whole `put()` is on a single line?

snazy (commented):
Maybe there's a nicer way to format the `static` init in `DynamoMetricsPublisher`.

It would also be nice to have some rudimentary unit test that verifies that the metrics return something. We cannot really assert on any `Duration` and probably only whether some `count > 0` or so. I'd be okay if there are 2 or 3 existing test-cases overridden for dynamo that do that "metrics-smoke-test".

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755#-:-MShlnPtBFXUt0BrY45q:bi7j501)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @laurentgo and @rymurr)

rymurr (commented):
added a test, can easily add more but this asserts that generally the metrics we want are getting created and have sane values

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755#-:-MSmAlxICybJ6AjCECDm:b-i75mid)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 144 at r1](https://reviewable.io/reviews/projectnessie/nessie/755#-MSgpCVr-5SbJRJv3G5W:-MSlo1k_0YWLvrb6LWIE:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b8e73df3282ec615759987914c97a9999eb43207/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L144)):*

ah, bummer. didn't notice that those dynamo-API don't share a common interface for that `overrideConfiguration`. Yea, it looks a bit ugly. I wouldn't mind to revert this then.
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 93 at r2](https://reviewable.io/reviews/projectnessie/nessie/755#-MShkmVY5EsfkqhNh4n6:-MSlo6g88s5-B_rK73Zu:b-auusig) ([raw file](https://github.com/projectnessie/nessie/blob/02f72da991dd41ff120546d80bcdf523e79e0949/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L93)):*

Would this look a bit nicer, when the whole `put()` is on a single line?
</blockquote></details>

yeah, played around to put them on single lines

versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java
```diff
@@ -0,0 +1,298 @@
...
+      @Override
+      public SdkHttpRequest modifyHttpRequest(
+          ModifyHttpRequest context,
+          ExecutionAttributes executionAttributes) {
+        SdkRequest request = context.request();
+        String name = request.getClass().getSimpleName().replace("Request", "");
+        ToLongFunction<SdkRequest> count = ITEM_COUNT_REQUEST.get(request.getClass());
+
+        if (count != null) {
+          String metricName = "DynamoDD.RequestedRecordCount";
```
snazy, line 253:
```suggestion
          String metricName = "DynamoDB.RequestedRecordCount";
```

snazy (changes requested):

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 253 at r3](https://reviewable.io/reviews/projectnessie/nessie/755#-MSmDqqs5T5RZKTQsDqA:-MSmDqqt6sfWFMPnmFp9:b5z1ryp) ([raw file](https://github.com/projectnessie/nessie/blob/6a6539740bbbf0b8330f243d23716ff2542072b0/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L253)):*
> ```Java
> 
>         if (count != null) {
>           String metricName = "DynamoDD.RequestedRecordCount";
> ```

DynamoDD ??

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/755#-MSmEs5614li-NXjiyFE:-MSmEs5614li-NXjiyFF:b-el4u79) ([raw file](https://github.com/projectnessie/nessie/blob/6a6539740bbbf0b8330f243d23716ff2542072b0/versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java#L75)):*
> ```Java
>     DistributionSummary putCapacity = (DistributionSummary) meters.stream().findFirst().get();
>     Assertions.assertEquals(4, putCapacity.count());
>     Assertions.assertEquals(10.0, putCapacity.totalAmount());
> ```

I guess, it's better to just assert on > 0 or so

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 9 files reviewed, 7 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java, line 253 at r3](https://reviewable.io/reviews/projectnessie/nessie/755#-MSmDqqs5T5RZKTQsDqA:-MSmF49B8Vsf8i47c7rJ:b8h2s3) ([raw file](https://github.com/projectnessie/nessie/blob/6a6539740bbbf0b8330f243d23716ff2542072b0/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/metrics/DynamoMetricsPublisher.java#L253)):*

DynamoDD ??
</blockquote></details>

good eye
___
*[versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/755#-MSmEs5614li-NXjiyFE:-MSmFANN2iRjuP24ecI-:b8vaddh) ([raw file](https://github.com/projectnessie/nessie/blob/6a6539740bbbf0b8330f243d23716ff2542072b0/versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java#L75)):*

I guess, it's better to just assert on > 0 or so
</blockquote></details>

I would rather assert the value...if that changes then the impl changes which is worth knowinng about imho

its isolated from other unit tests so its only if the put impl changes significantly

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/755)* status: 0 of 9 files reviewed, 6 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/755#-MSmEs5614li-NXjiyFE:-MSmWmyNCJJyqGpdfT5t:b7uu04p) ([raw file](https://github.com/projectnessie/nessie/blob/6a6539740bbbf0b8330f243d23716ff2542072b0/versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/impl/ITDynamoMetrics.java#L75)):*

I would rather assert the value...if that changes then the impl changes which is worth knowinng about imho

its isolated from other unit tests so its only if the put impl changes significantly

</blockquote></details>

I ended up switching it back to your suggestion. The code is dependent on which order the tests run and I dont know enough micrometer to decouple it.

snazy (approved):
LGTM, as long as the last commits fix the build ;) ;)

PR #757, opened 2021-02-02
Dynamo: split initializeDatabase + setupTables configs

snazy:
Splits the DynamoDB config `initialize` that actually does two things:
* create any missing tables
* write any missing _initial_ objects (`L1`, `L2`, `L3`) that are needed for an initial setup

A single property doesn't account for "upgrade scenarios": the case when we change the schema or add more tables but want to keep existing content without letting the store re-add initial objects.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/757)

### DIFF
--- a/servers/quarkus-server/src/main/java/com/dremio/nessie/server/config/ApplicationConfig.java
+++ b/servers/quarkus-server/src/main/java/com/dremio/nessie/server/config/ApplicationConfig.java
@@ -117,6 +117,9 @@ public class ApplicationConfig {
     @ConfigProperty(name = "initialize", defaultValue = "false")
     boolean isDynamoInitialize();
 
+    @ConfigProperty(name = "setupTables", defaultValue = "false")
+    boolean isSetupTables();
+
     @ConfigProperty(defaultValue = DynamoStoreConfig.TABLE_PREFIX)
     String getTablePrefix();
 
--- a/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/VersionStoreFactory.java
+++ b/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/VersionStoreFactory.java
@@ -140,6 +140,7 @@ public class VersionStoreFactory {
           }))
           .region(Region.of(region))
           .initializeDatabase(in.isDynamoInitialize())
+          .setupTables(in.isSetupTables())
           .tablePrefix(in.getTablePrefix())
           .build());
     dynamo.start();
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -29,7 +29,9 @@ nessie.version.store.jgit.type=DISK
 nessie.version.store.jgit.directory=/tmp/jgit
 
 ## Dynamo version store specific configuration
-### should Nessie create its own dynamo tables
+### should Nessie create its own DynamoDB tables
+nessie.version.store.dynamo.setupTables=false
+### should Nessie initialize any missing initial objects in DynamoDB
 nessie.version.store.dynamo.initialize=false
 ### table names for ref, tree and value tables respectively
 nessie.version.store.dynamo.tablePrefix=nessie_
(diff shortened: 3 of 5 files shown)


rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/757)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @laurentgo and @snazy)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/757#-MSawLD6E2IiYiE2ArYn:-MSawLD6E2IiYiE2ArYo:b343fne) (no related file):*
The `docker-compose.yml` in `perftest` also needs to change to match this change. Possibly some other places as well.

Overall I don't see this being a huge improvement over existing. This seems to add a hypothetical upgrade scenario when the original intent of this option was to initialize an empty database to help people get moving quickly.

I would be inclined to have a completely separate tool/flag/endpoint to perform database upgrades than to change the meaning of this flag.

---
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 150 at r1](https://reviewable.io/reviews/projectnessie/nessie/757#-MSav2bv57L8hNnDFdX7:-MSav2bv57L8hNnDFdX8:b-iix52a) ([raw file](https://github.com/projectnessie/nessie/blob/36ebf9bed1006e16007cfb832323569d18a6dca1/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L150)):*
> ```Java
>       async = b2.build();
> 
>       if (config.setupTables()) {
> ```

You are changing the meaning of an existing property which I don't like. I would prefer to keep the initialize database option to have the same semantics
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 156 at r1](https://reviewable.io/reviews/projectnessie/nessie/757#-MSavCkL2jA41w2zsTC4:-MSavCkL2jA41w2zsTC5:b-3tkvpj) ([raw file](https://github.com/projectnessie/nessie/blob/36ebf9bed1006e16007cfb832323569d18a6dca1/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L156)):*
> ```Java
>             .forEach(this::createIfMissing);
>       }
>       if (config.initializeDatabase()) {
> ```

this appears to be an idempotent operation, why not always run it if it has no effect. 

If it must be an extra option then the option should be overridden when we create the database. ie if we create the tables but don't put anything in them then the tables wont be very useful

servers/quarkus-server/src/main/java/com/dremio/nessie/server/config/ApplicationConfig.java
```diff
@@ -117,6 +117,9 @@ public VersionStoreDynamoConfig getVersionStoreDynamoConfig() {
     @ConfigProperty(name = "initialize", defaultValue = "false")
     boolean isDynamoInitialize();
 
+    @ConfigProperty(name = "setupTables", defaultValue = "false")
+    boolean isSetupTables();
```
laurentgo, line 121: (style) the property name is not really correct english (but I guess same above). I'm not sure to understand what's the difference with the `initialize` property above...

laurentgo:
I'm actually curious about the `upgrade` scenario mentioned, since it hasn't been discussed. And so by extension, I'm not sure why this change is needed right now.

snazy:
I've originally split the configs to make the tests against the JDBC-store run faster, because checking the tables via database-metadata takes quite long, but re-creating missing L1+L2+L3 is mandatory for most test cases. I.e. it makes a lot of sense to improve the runtime of the store-tests, but probably not to expose it as a user-facing configuration.

Upgrade scenarios, and actually also initial setups, should be handled externally or at least via a more sophisticated mechanism, because having more than one instance doing the create/alter/drop/migrate is usually not a good idea (way more complicated error and retry handling + some databases can actually be corrupted when concurrent DDLs happen).

snazy:
Ugh - just realized that DynamoDB doesn't have a truncate-operation, so the split doesn't make any more sense.

PR #758, opened 2021-02-02
Re-enable `ITNativeRest`

snazy:
As noted in #589, we can continue with the current approach and get around a `resetStoreUnsafe` by just using "unique names" in each test of `TestRest`. Therefore this PR, which re-enables `ITNativeRest`.

Also adds a nit-ish fix in `RestGitTest` to only generate hash-strings up to the length of the hash as supported by the underlying store implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/758)


rymurr (approved):
:+1:

PR #767, opened 2021-02-04
Fix passing file.encoding system property + update maven-failsafe-plugin as well

snazy:
Maven build warning before this patch:
```
[WARNING] file.encoding cannot be set as system property, use <argLine>-Dfile.encoding=...</argLine> instead
```

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/767)


rymurr (approved):
Lgtm

ohbus (approved):

PR #783, opened 2021-02-08
Moved TableCommitMetaStoreWorker outside of nessie-quarkus-server module

brunos-bq:
In this PR, I am moving the class TableCommitMetaStoreWorker from the nessie-quarkus subproject to the nessie-services project.

The reason for this move is to facilitate embedding Nessie within other projects.
Standalone Nessie runs through Quarkus. Embedding Nessie into another project will probably not use Quarkus.

In order to create a VersionStore, a StoreWorker is needed. Prior to that change, projects would have to depend on nessie-quarkus in order to instantiate a StoreWorker.

With this change, projects will now only depend on nessie-model and nessie-services, not importing anything related to Quarkus.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)

### DIFF
--- a/servers/server/pom.xml
+++ b/servers/server/pom.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (C) 2020 Dremio
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie-server-parent</artifactId>
+    <version>0.3.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nessie-server</artifactId>
+
+  <name>Nessie - Server</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-model</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-versioned-spi</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>jakarta.enterprise</groupId>
+      <artifactId>jakarta.enterprise.cdi-api</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.xolstice.maven.plugins</groupId>
+        <artifactId>protobuf-maven-plugin</artifactId>
+        <version>0.6.1</version>
+        <configuration>
+          <protoSourceRoot>${basedir}/src/main/proto</protoSourceRoot>
+          <protocArtifact>com.google.protobuf:protoc:${protobuf.version}:exe:${os.detected.classifier}</protocArtifact>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>compile</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/generated-sources/protobuf</outputDirectory>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/servers/pom.xml
+++ b/servers/pom.xml
@@ -34,5 +34,6 @@
     <module>services</module>
     <module>quarkus-server</module>
     <module>lambda</module>
+    <module>server</module>
   </modules>
 </project>
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -41,6 +41,11 @@
     </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
       <artifactId>nessie-versioned-jgit</artifactId>
       <version>${project.version}</version>
     </dependency>
@@ -196,25 +201,6 @@
               <goal>generate-code-tests</goal>
               <goal>build</goal>
             </goals>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.xolstice.maven.plugins</groupId>
-        <artifactId>protobuf-maven-plugin</artifactId>
-        <version>0.6.1</version>
-        <configuration>
-          <protoSourceRoot>${basedir}/src/main/proto</protoSourceRoot>
-          <protocArtifact>com.google.protobuf:protoc:${protobuf.version}:exe:${os.detected.classifier}</protocArtifact>
-        </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>compile</goal>
-            </goals>
-            <configuration>
-              <outputDirectory>${project.build.directory}/generated-sources/protobuf</outputDirectory>
-            </configuration>
           </execution>
         </executions>
       </plugin>
--- a/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/VersionStoreFactory.java
+++ b/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/VersionStoreFactory.java
@@ -37,6 +37,7 @@ import org.slf4j.LoggerFactory;
 
 import com.dremio.nessie.model.CommitMeta;
 import com.dremio.nessie.model.Contents;
+import com.dremio.nessie.server.TableCommitMetaStoreWorker;
 import com.dremio.nessie.server.config.ApplicationConfig;
 import com.dremio.nessie.server.config.ApplicationConfig.VersionStoreDynamoConfig;
 import com.dremio.nessie.server.config.converters.VersionStoreType;
--- a/servers/server/src/main/java/com/dremio/nessie/server/TableCommitMetaStoreWorker.java
+++ b/servers/server/src/main/java/com/dremio/nessie/server/TableCommitMetaStoreWorker.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.dremio.nessie.server.providers;
+package com.dremio.nessie.server;
 
 import java.io.IOException;
 import java.util.stream.Collectors;
@@ -21,12 +21,6 @@ import java.util.stream.Stream;
 
 import javax.inject.Singleton;
 
-import com.dremio.nessie.jgit.ObjectTypes.PContents;
-import com.dremio.nessie.jgit.ObjectTypes.PDeltaLakeTable;
-import com.dremio.nessie.jgit.ObjectTypes.PHiveDatabase;
-import com.dremio.nessie.jgit.ObjectTypes.PHiveTable;
-import com.dremio.nessie.jgit.ObjectTypes.PIcebergTable;
-import com.dremio.nessie.jgit.ObjectTypes.PSqlView;
 import com.dremio.nessie.model.CommitMeta;
 import com.dremio.nessie.model.Contents;
 import com.dremio.nessie.model.DeltaLakeTable;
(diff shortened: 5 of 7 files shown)


jacques-n (commented):
While I can see the utility of possibly exposing some of this functionality outside the quarkus-server module, it doesn't seem it should be moved to the services module. What do you think @rymurr and @laurentgo ?

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783#-:-MT2rgMW5pLh8TrorVz-:b1zgz9n)* status: 0 of 6 files reviewed, all discussions resolved

laurentgo:
I would agree with @jacques-n : `services` module has been designed to be store-agnostic, so there's no good reason to bring specific code to this module. I'm not saying that the integration between services and the tiered store in quarkus is great either though...

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @brunos-bq)

___
*[servers/services/pom.xml, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT3Vggq6N-JGPJoA8_U:-MT3Vggq6N-JGPJoA8_V:b-50nzx2) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/pom.xml#L40)):*
> ```XML
>     <dependency>
>       <groupId>org.projectnessie</groupId>
>       <artifactId>nessie-versioned-jgit</artifactId>
> ```

it's on purpose that we do not depend on a specific implementation

rymurr:
I am on the fence @laurentgo and @jacques-n, the reason being the `TableMetaStoreWorker` and associated proto def'n are definitely not part of the `quarkus-server`. Any service needs them to function regardless of the underlying server protocol. However the correct place for them to live is up for discussion:
1) in the version store module. I think this is incorrect as the version store module uses only parameterised types
2) services module. This is the first place that the `VerionStore` is specialized to `Contents` and `CommitMeta` so in some ways its not a bad place for them to live
3) new module that services depends on. I am ok with this too but its a whole new module for a class and a protobuf definition
4) in the quarkus-server. As I said this is too high, we have already specialised the `VersionStore` in `services` and the services are useless w/o the store worker and protobuf definition
5) somewhere else?

Having said that the services and quarkus-server modules are still tightly coupled and I am not sure if we should take a crack at decoupling them instead of just shifting 1 class.

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 6 files reviewed, 5 unresolved discussions (waiting on @brunos-bq and @laurentgo)

___
*[servers/services/pom.xml, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT3Vggq6N-JGPJoA8_U:-MT4l3JAA_YE8AYODfFt:btmwgaw) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/pom.xml#L40)):*

it's on purpose that we do not depend on a specific implementation
</blockquote></details>

agreed, jgit shouldn't be a dependency
___
*[servers/services/src/main/java/com/dremio/nessie/services/rest/TreeResource.java, line 72 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lKGS2_pKM1p3Tstu:-MT4lKGS2_pKM1p3Tstv:be9kkbk) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/rest/TreeResource.java#L72)):*
> ```Java
> 
>   @Inject
>   public TreeResource(ServerConfig config, Principal principal,
> ```

why does this need changing?
___
*[servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java, line 16 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lphx8lDA5Md9jzW5:-MT4lphx8lDA5Md9jzW6:b-ytfyaq) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java#L16)):*
> ```Java
>  * limitations under the License.
>  */
> package com.dremio.nessie.services.providers;
> ```

I don't think this is a provider
___
*[servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java, line 56 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4latd7EOxOCnh_jxl:-MT4latd7EOxOCnh_jxm:b-bgh0lg) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java#L56)):*
> ```Java
> import com.google.protobuf.UnsafeByteOperations;
> 
> @Singleton
> ```

I am not 100% convinced that this class should be annotated w/ injection. I think a builder in the specific server modules should take care of it.
___
*[servers/services/src/main/proto/table.proto, line 19 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lBfq4gBAgMJuhruE:-MT4lBfq4gBAgMJuhruF:b1yjt5n) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/proto/table.proto#L19)):*
> ```protobuf
> package com.dremio.plugin.partsyn;
> 
> option java_package = "com.dremio.nessie.jgit";
> ```

I think this package name is wrong, and has been for some time. These don't have any relation to `jgit`

rymurr:
After thinking more I am actually leaning towards point 3 in my list. A new module that contains workers and utils for the server modules. I think a server doesn't necessarily have to use the Store Worker we use with quarkus but its certainly the easiest option.

snazy:
+1 on a new `nessie-server` module and move the non-runtime-environment-specific and non-(version-)store-specific code there. (This is related to #786 .) Such a `nessie-server` module must however not depend on any (version-)store configuration or implementation. To build a working Nessie Server, you need three things:
1. Runtime environment (Quarkus or embedded) + code/config specific for the runtime-environment
2. `nessie-server` module
3. One version-store implementation

brunos-bq:
I have just updated the PR according to all the feedback received.

brunos-bq (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[servers/services/pom.xml, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT3Vggq6N-JGPJoA8_U:-MT7P41-6tde2CjbBKOm:bm9eeu5) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/pom.xml#L40)):*

agreed, jgit shouldn't be a dependency
</blockquote></details>

The dependency was not needed and removed.
___
*[servers/services/src/main/java/com/dremio/nessie/services/rest/TreeResource.java, line 72 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lKGS2_pKM1p3Tstu:-MT7P8Tf6dRqv3NHf4vW:bloiygd) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/rest/TreeResource.java#L72)):*

why does this need changing?
</blockquote></details>

The way we instantiate a VersionStore where Nessie will be embedded, we use build the TreeResource object using this constructor.
___
*[servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java, line 16 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lphx8lDA5Md9jzW5:-MT7OsVu3xlEcBy-Neja:bt2vr1b) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java#L16)):*

I don't think this is a provider
</blockquote></details>

Fixed.
___
*[servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java, line 56 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4latd7EOxOCnh_jxl:-MT7PSrw5mtVYIQTXJvm:bxexult) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/java/com/dremio/nessie/services/providers/TableCommitMetaStoreWorker.java#L56)):*

I am not 100% convinced that this class should be annotated w/ injection. I think a builder in the specific server modules should take care of it.
</blockquote></details>

Should this be done in this PR, or as a TODO?
___
*[servers/services/src/main/proto/table.proto, line 19 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lBfq4gBAgMJuhruE:-MT7P-Vu0FpHZKKDGjqo:byyojxv) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/proto/table.proto#L19)):*

I think this package name is wrong, and has been for some time. These don't have any relation to `jgit`
</blockquote></details>

Not familiar with this part of the code. What would be a good replacement name?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @brunos-bq, @laurentgo, and @rymurr)

___
*[servers/services/src/main/proto/table.proto, line 19 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lBfq4gBAgMJuhruE:-MTAm2p33XbUrHhSDhsc:buxok4u) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/proto/table.proto#L19)):*
<details><summary><i>Previously, brunos-bq (Bruno Paiva Lima da Silva) wrote…</i></summary><blockquote>

Not familiar with this part of the code. What would be a good replacement name?
</blockquote></details>

`com.dremio.nessie.server.model`?

servers/server/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    </dependency>
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
```
rymurr, line 55: these jboss and javax/jakarta dependencies look strange to me. Where are the coming from?

rymurr:
I am happier with this in another module. @snazy  @laurentgo @jacques-n ? Thoughts?

This still isn't ideal and it feels squishy to me but I don't have a better way atm. This module is better than forcing everyone who may want to implement a server to pull in quarkus stuff. Short of other servers creating their own workers, which given the size of the Worker may actually be better. There is no reason that all servers should be forced to use protobuf to store values.

brunos-bq (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[servers/server/pom.xml, line 55 at r2](https://reviewable.io/reviews/projectnessie/nessie/783#-MT7KWr839oPi-e9mZgt-r2-55:-MTCHuQf-QIi5Jgn2v15:bhv09xr) ([raw file](https://github.com/projectnessie/nessie/blob/19289c96e62baf5e3c9ad8f3be3844d0f82eb12f/servers/server/pom.xml#L55)):*

these jboss and javax/jakarta dependencies look strange to me. Where are the coming from?
</blockquote></details>

Some dependencies were not needed. Removed them.
___
*[servers/services/src/main/proto/table.proto, line 19 at r1](https://reviewable.io/reviews/projectnessie/nessie/783#-MT4lBfq4gBAgMJuhruE:-MTCI-yS2AvXg_i_y5_F:b-h1kwns) ([raw file](https://github.com/projectnessie/nessie/blob/fd5ffae247c48adfcf3d24489dac6a9f32c91209/servers/services/src/main/proto/table.proto#L19)):*

`com.dremio.nessie.server.model`?
</blockquote></details>

That works. Updating.

brunos-bq:
r3: Removed unused dependencies. Renamed package hierarchy in table.proto.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @brunos-bq, @laurentgo, and @rymurr)

___
*[servers/server/src/main/proto/table.proto, line 17 at r3](https://reviewable.io/reviews/projectnessie/nessie/783#-MTEV_N10rjR7xnGo2RL:-MTEV_N10rjR7xnGo2RM:b7z1p6g) ([raw file](https://github.com/projectnessie/nessie/blob/34bb8a4fb2d8879ec34382cd35326beec52a6680/servers/server/src/main/proto/table.proto#L17)):*
> ```protobuf
>  */
> syntax = "proto3";
> package com.dremio.plugin.partsyn;
> ```

this is definitely wrong too :)

laurentgo:
> This still isn't ideal and it feels squishy to me but I don't have a better way atm. This module is better than forcing everyone who may want to implement a server to pull in quarkus stuff. Short of other servers creating their own workers, which given the size of the Worker may actually be better. There is no reason that all servers should be forced to use protobuf to store values.

The main question for me seems to be "Do we need a different implementation of StoreWorker?". One area for sure which is not great at the moment is the mapping requirement between a POJO in model and a Protobuf class in the store worker. This doesn't seem something extensible and working well especially if multiple store worker implementions have to be supported.

brunos-bq (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/783)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[servers/server/src/main/proto/table.proto, line 17 at r3](https://reviewable.io/reviews/projectnessie/nessie/783#-MTEV_N10rjR7xnGo2RL:-MTIJz9KApG4DCbeTmpE:bt2vr1b) ([raw file](https://github.com/projectnessie/nessie/blob/34bb8a4fb2d8879ec34382cd35326beec52a6680/servers/server/src/main/proto/table.proto#L17)):*

this is definitely wrong too :)
</blockquote></details>

Fixed.

rymurr:
closing as this was done in #895

PR #792, opened 2021-02-10
Add sharding support for Mongo store tests.

kylep-dremio:
This adds support for a sharded Mongo configuration for tests.

It adds a LocalMongoD and LocalMongoS so both sharded and non-sharded Mongo instances are used for running tests.

Note that the ability to specify Mongo as a Jupiter extension will (hopefully) eventually be merged into flapdoodle itself, at which point this may be vastly simplified and rely on the flapdoodle code instead.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/792)


rymurr (approved):
The structure looks good to me. I like the use of the junit extension. 

I am not a mongo expert so I am trusting that the sharding and stuff is set up correctly. It looks sensible to me.

rymurr (changes requested):
Just was looking at the logs of the build. Currently mongo is using `DEBUG` which adds significant size to the build log output. Can you switch to `INFO`?

Also, when I ran the test locally I added ~1.2GB to my tmp w/o it being cleaned up. Can you double check your cleaning up after yourself?

kylep-dremio (commented):
I've disabled the logs by default for tests, we can re-enable when necessary.

Looking for a solution on the cleanup, the design is meant to reduce the amount of downloads that occur but that also means that there's no clean trigger to clean up artifacts.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/792#-:-MTDenDg-quqsvS4sQ3x:b-c5gs67)* status: 0 of 10 files reviewed, all discussions resolved

rymurr:
I merged the tests PR, looks like this needs a rebase now.

rymurr:
Sorry @kylep-dremio I merged the package name changes ahead of this one. Needs another rebase :-(

kylepbit:
@rymurr done!

rymurr (approved):
Thanks @kylep-dremio

PR #798, opened 2021-02-10
Respect DynamoDB limit of max 25 requests per batch-write

snazy:
While running some perf-test experiments, I ran into situations that DynamoDB refused a Batch-Write for a `save()` due to exceeded limits. The current implementation of `com.dremio.nessie.versioned.store.dynamo.DynamoStore#save` does some already some grouping, but it was incorrect and used a way too high limit (100).

The [DynamoDB BatchWriteItem docs](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html) says, that all batch-writes with more than 25 requests will be rejected.

I'm not sure, why the existing `DynamoStore.save()` implementation used the `paginationSize` (defaults to  `LOAD_SIZE`, defaults to `100`), so I left it there, but limited the number of requests to 25.

I don't _think_ that we need to assert on a single request being > 400kB or the whole batch > 16MB, because IIRC Nessie objects are already "capped" at 400kB and 25*400 is not > 16MB.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/798)

### DIFF
--- a/versioned/tiered/dynamodb/pom.xml
+++ b/versioned/tiered/dynamodb/pom.xml
@@ -116,6 +116,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
--- a/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java
+++ b/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java
@@ -23,7 +23,9 @@ import static org.projectnessie.versioned.dynamodb.DynamoSerDe.deserializeToCons
 import static org.projectnessie.versioned.dynamodb.DynamoSerDe.serializeWithConsumer;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -104,6 +106,12 @@ import software.amazon.awssdk.services.dynamodb.model.WriteRequest;
 public class DynamoStore implements Store {
 
   public static final int LOAD_SIZE = 100;
+
+  /**
+   * The <a href="https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchWriteItem.html">Dynamo docs</a>
+   * state a batch-write be rejected, if it contains more than 25 requests in a batch.
+   */
+  static final int DYNAMO_BATCH_WRITE_LIMIT_COUNT = 25;
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DynamoStore.class);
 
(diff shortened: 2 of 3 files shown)


rymurr (dismissed):

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/798)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 114 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC39kQ3Qhj3GeS81Ad:-MTC39kQ3Qhj3GeS81Ae:bgmf5bv) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L114)):*
> ```Java
>    * state a batch-write be rejected, if it contains more than 25 requests in a batch.
>    */
>   public static final int DYNAMO_BATCH_WRITE_LIMIT_COUNT = 25;
> ```

isn't that some internal concern?
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 357 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC41W4AxSEzHimAvOK:-MTC41W4AxSEzHimAvOL:be988zc) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L357)):*
> ```Java
>     // The following nested forEach construct collects up to `pageSize` `WriteRequest`s in
>     // the `currentBatch` map. The current number of `WriteRequest`s is held in `currentBatchSize[0]`.
>     Map<String, Collection<WriteRequest>> currentBatch = new HashMap<>();
> ```

isn't that a good use case for a collector which could handle the batching and processing?

Or maybe we should look at using reacting stream like we did for Mongo?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/798)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @laurentgo)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 114 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC39kQ3Qhj3GeS81Ad:-MTHL0II8nD9JjenMvmt:bwjcd2l) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L114)):*

isn't that some internal concern?
</blockquote></details>

Made it package-private
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 357 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC41W4AxSEzHimAvOK:-MTHL5TS9TBUNcLHcC48:b-97knep) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L357)):*

isn't that a good use case for a collector which could handle the batching and processing?

Or maybe we should look at using reacting stream like we did for Mongo?
</blockquote></details>

`Collector` are quite difficult to use here due to the, say, "batching characteristics" here. Not so happy with pulling in a reflective-stuff-library to do this.
I opted to extract the "collecting magic" into a separate class, which is package-private, because it's implementation pretty focused on how the Dynamo-Java-API requires the batches to work.

Since that class is actually pretty isolated, I'll add a unit test for it as well in a few.

versioned/tiered/dynamodb/src/test/java/com/dremio/nessie/versioned/store/dynamo/BatchesCollectorTest.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.stream.IntStream;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.dremio.nessie.versioned.store.dynamo.DynamoStore.BatchesCollector;
+import com.google.common.base.Objects;
+
+public class BatchesCollectorTest {
```
rymurr, line 40: all other tests have Test* format

rymurr (commented):
just the class rename and the rebase missing. I think the code is a lot easier to read w/ the change to a separate class

rymurr (approved):

snazy (commented):
> Dismissed @laurentgo from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/798)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 114 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC39kQ3Qhj3GeS81Ad:-MTLA4do2IR_pVOvBVq0:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L114)):*

Made it package-private
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java, line 357 at r1](https://reviewable.io/reviews/projectnessie/nessie/798#-MTC41W4AxSEzHimAvOK:-MTLA1J9-cOMVpO80rNY:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/e95f1d826c402a4c02f0c57ea30e8b288241b023/versioned/tiered/dynamodb/src/main/java/com/dremio/nessie/versioned/store/dynamo/DynamoStore.java#L357)):*

`Collector` are quite difficult to use here due to the, say, "batching characteristics" here. Not so happy with pulling in a reflective-stuff-library to do this.
I opted to extract the "collecting magic" into a separate class, which is package-private, because it's implementation pretty focused on how the Dynamo-Java-API requires the batches to work.

Since that class is actually pretty isolated, I'll add a unit test for it as well in a few.
</blockquote></details>

Done.

Issue #707, opened 2021-01-20
How to call and take action on GC algorithm?


rymurr:
#684 gives us the core GC algorithm. Currently its called only via tests and the results are just a `List`. We should figure out how to 

1. make calling the gc algo ergonomic: via eg `spark-submit` or so. and document it. Maybe make a custom SQL extension so the algo could be called from Spark SQL. This implies all the docs to go along with it: 
2. Do something about the result.
    - result is written to the data lake or to a new Store table
    - check the sanity of the result
    - another spark job to action the result (eg perform deletes)
    - or have a `dry-run` option on the algo and re-run it to perform the delete

rymurr:
some thoughts from #803 

> Hmm.... as I contemplate things. I wonder if we should actually be writing something that can map to a Spark struct type (somehow). That would probably be the most useful. The concern I had with JSON was that it isn't doesn't handle versioning/schema evolution well. Since asset keys may be saved across time and then format changes would be hard to manage.

> Interesting, my thought was that they shouldn't be saved across time. At least across long times/version changes. I like the idea of spark rows though. In the end I want to be able to show this first as a spark table, then via python or js via the UI. Thinking more, writing an iceberg table and storing it in Nessie is interesting too...

PR #803, opened 2021-02-11
GC support for iceberg

rymurr:
This PR adds support for running Iceberg tables with the GC algo. The main changes focus on adding a traversal mechanism for iceberg manifests and asset keys specialised to data files.

Should be rebased once #802  is in.


versioned/tiered/gc/src/test/java/com/dremio/nessie/versioned/gc/MultimapCollector.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.stream.Collector;
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Multimap;
+
+/**
+ * Collector to collect into multimap.
+ */
+class MultimapCollector<T, K, V> implements Collector<T, Multimap<K, V>, Multimap<K, V>> {
```
jacques-n, line 32: Doesn't Guava already provide this? 

Maybe this? https://guava.dev/releases/22.0/api/docs/com/google/common/collect/Multimaps.html#toMultimap-java.util.function.Function-java.util.function.Function-java.util.function.Supplier-

versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java
```diff
@@ -0,0 +1,76 @@
...
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+
+public class IcebergAssetKeyReader<T> implements Supplier<Stream<AssetKey>> {
+
+  private final IcebergTable table;
+  private final SerializableConfiguration hadoopConfig;
+
+  public IcebergAssetKeyReader(IcebergTable table, SerializableConfiguration configuration) {
```
jacques-n, line 43: I think we should be passing a general gc configuration object which this then uses to build a specific configuration.

I guess your current thinking is that the we use the user running the job's credentials for the data retrieval as well. This feels half okay and half squishy... @laurentgo thoughts on best way to handle this?

versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java
```diff
@@ -0,0 +1,114 @@
...
+    } catch (IOException e) {
+      throw new IllegalStateException(String.format("Unable to delete %s", this.path), e);
+    }
+  }
+
+  @Override
+  public List<String> toReportableName() {
+    List<String> name = Lists.newArrayList();
+    name.add(type.name());
+    name.addAll(DOT.splitToList(path));
```
jacques-n, line 73: isn't this path a file path and thus would be split on back slash?

versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      return CompletableFuture.completedFuture(path.getFileSystem(hadoopConfig.value()).delete(path, type.isRecursiveDelete()));
+    } catch (IOException e) {
+      throw new IllegalStateException(String.format("Unable to delete %s", this.path), e);
+    }
+  }
+
+  @Override
+  public List<String> toReportableName() {
+    List<String> name = Lists.newArrayList();
```
jacques-n, line 71: do you think we can add more data here? For example, maybe add the table name and snapshot name?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 15 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java, line 102 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJo-r1-102:-MTL_rPK5O0NGsT1LZiF:br5017x) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java#L102)):*

I'm not a big fan of the use of json here. Can't we use a protobuf oneof for Iceberg?
</blockquote></details>

My thinking on json was its flexible and these objects should only ever be seen by users so human readable and easy to read (from eg javascript or python) is useful. what was your thinking around protobuf? Maybe we have protobuf for internal serialisation and json when it is written out?
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/IdentifyUnreferencedAssets.java, line 241 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJl-r1-241:-MTKx-kGCozg5W-nVBws:bf9s1dn) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/IdentifyUnreferencedAssets.java#L241)):*

It might be good to split out the core algo changes in a separate PR from the iceberg/contents impl.
</blockquote></details>

done see #811
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJm-r1-55:-MTKafUw7k1uzri5u3ju:b-z8bzr7) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java#L55)):*

Why the change? Are you reusing the spark session? I was thinking that we shouldn't support that.
</blockquote></details>

Spark Sessions and catalogs are static so are re-used across tests. What is the downside to recreating? See #811
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java, line 22 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJq-r1-22:-MTLZkFS9HlysKYanUpK:b-t9w5dl) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java#L22)):*

I don't think the iceberg gc stuff should be in the gc package. I think it would be better to have separate modules that are sourced somehow by the general gc package.

Also, wondering whether we need both an enum here or if we can survive with just a lower-level oneof.
</blockquote></details>

Going by this and other comments I am wondering if you are thinking that the `AssetKey` class hierarchy should be deeper? Eg FileSystemKey -> FolderFilesSystemKey -> IcebergTableAssetKey etc and removing the enum maybe? I wanted to keep it shallow for simplicity but am curious to see what you were thinking.
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java, line 71 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJr-r1-71:-MTLb_5z1dDa_zz1B0lG:bvn6m56) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java#L71)):*

do you think we can add more data here? For example, maybe add the table name and snapshot name?
</blockquote></details>

yes definitely. Will wait on some of the other discussions before adding
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java, line 73 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJr-r1-73:-MTLZaY02c-nCUY45R3t:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/FileSystemAssetKey.java#L73)):*

isn't this path a file path and thus would be split on back slash?
</blockquote></details>

Done.
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJs-r1-43:-MTKbOt488GJ-46E6L4d:b-auqk3t) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java#L43)):*

I think we should be passing a general gc configuration object which this then uses to build a specific configuration.

I guess your current thinking is that the we use the user running the job's credentials for the data retrieval as well. This feels half okay and half squishy... @laurentgo thoughts on best way to handle this?
</blockquote></details>

I agree, I don't like the way the Hadoop config has to float around everywhere. Its hard to get away from it though...maybe a filesystem path doesn't have config at all and needs to be handed a config object to do its work?

Not sure where to source the general gc config from or who would own it? Typically in a spark job all hadoop configs are derived from the spark owned one. So if I understand what you were thinking was to overlay w/ an 'admin' like set of settings? 

The config being passed in here only needs permissions to read the metadata so the users creds should be sufficient right?

My thinking is that, at least initially, teh most common GC workflow is for a super user to run the job and inspect the results. Followed by a second run to actually delete if they are happy.  The 2nd run should definitely be done by a user that has the permissions to delete the data so their hadoop config should be ok. What was your thinking there?
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJs-r1-57:-MTKfmbA958u4DTeWIQC:b-ow29ik) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java#L57)):*

We should be considering all snapshots, right? It looks like you're not doing that.

Also, can we avoid the pattern where we load all manifests before we start working with them? It would be preferable to work through manifests lazily.
</blockquote></details>

Yeah, I am frustrated by this. I want to keep hold of all snapshots too as the metadata files are technically invalid for iceberg timetravel if we delete snapshots w/o re-writing metadata files.

However then we will never get rid of anything, from what I can tell. http://iceberg.apache.org/maintenance/ implies that the only way to expire snapshots is to run an action which is destructive and not branch aware. 

Its rather complicated otherwise. I have to think more about it but it probably means re-writing every metadata file in every valid branch using the expire snapshots action and a dummy deleter. Basically re-writing all of history in order to expire a snapshot from all metadata files. Iceberg doesn't read metadata files and only tracks the snapshot list in the most recent manifest file as opposed to nessie reading all metadata files that it has a reference to.

A halfway solution is to run the expire snapshots action with a dummy deleter and just wait for the old metadata to roll off through other expiration techniques. I will investigate this option further now.

This raises other questions that worry me around table maintenance. If any of the table maintenance actions in iceberg run while the table is under nessie we are in a heap of trouble. Our GC algo doens't take into consideration orphan files either, which involves `ls`-ing the entire data/metadata directories looking for unreferenced items on-top of the existing gc algo.

Apologies for the long ramble, I was using Reviewable as a rubber duck.
___
*[versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJt-r1-37:-MTL1kvp-o3NdaxfzE33:b-dhx7s9) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java#L37)):*

Per comments above, can't this take a stream as input?
</blockquote></details>

All manifests is a `List` in the `TableMetadata` (in fact its the manifest list) so its materialized by the Iceberg libs at the time the metadata is read. The reason its a Set here is because I was originally reading all snapshots and collecting all manifests into a set. I will switch to list depending on the result of the other conversations.
___
*[versioned/tiered/gc/src/test/java/com/dremio/nessie/versioned/gc/MultimapCollector.java, line 32 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJv-r1-32:-MTKy7If0rjd5cigaN5x:b-6kq4ci) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/test/java/com/dremio/nessie/versioned/gc/MultimapCollector.java#L32)):*

Doesn't Guava already provide this? 

Maybe this? https://guava.dev/releases/22.0/api/docs/com/google/common/collect/Multimaps.html#toMultimap-java.util.function.Function-java.util.function.Function-java.util.function.Supplier-
</blockquote></details>

looked everywhere for that. Thanks. Removed

rymurr:
Thanks for the feedback @jacques-n I haven't addressed all your comments yet, waiting on some of the design coversations to wrap up before going further.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 13 files reviewed, 11 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java, line 22 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJq-r1-22:-MTLfNzj68nZ5zFWax6k:bnz6z19) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java#L22)):*

Going by this and other comments I am wondering if you are thinking that the `AssetKey` class hierarchy should be deeper? Eg FileSystemKey -> FolderFilesSystemKey -> IcebergTableAssetKey etc and removing the enum maybe? I wanted to keep it shallow for simplicity but am curious to see what you were thinking.
</blockquote></details>

Forgot to comment on the package. Thats not a problem, I don't mind. It seems excessive to me as its only the `IcebergTable` to `Stream<AssetKey>` code that is moving and the iceberg deps are already marked provided. But I think that the outcome of the design discussions would change things. Will wait till the design discussion wraps up before moving.

rymurr:
@jacques-n as requested added another unit test for the `AssetKeyReader` to check the iceberg metadata enumeration in isolation

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 8 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJm-r1-55:-MTc8kZ48DRQ87Imd50J:b-fmgwlt) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java#L55)):*

Spark Sessions and catalogs are static so are re-used across tests. What is the downside to recreating? See #811
</blockquote></details>

My main thought would be that this only needs to be changed for tests. For a real world scenario, having a conflict here should be a assumption failure.
___
*[versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJt-r1-37:-MTc8zFFCrinx5dSQ0dA:bdby9rr) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java#L37)):*

All manifests is a `List` in the `TableMetadata` (in fact its the manifest list) so its materialized by the Iceberg libs at the time the metadata is read. The reason its a Set here is because I was originally reading all snapshots and collecting all manifests into a set. I will switch to list depending on the result of the other conversations.
</blockquote></details>

I suggest we update the api in Iceberg to expose a Stream. Collapsing to Set for multiple snapshots makes sense but I'd be inclined to maintain a set of manifest identifiers as opposed to the entire manifest file object.

It actually begs an interesting question. Right now the GC algo uses equality against the serialized representation of the entire asset key. I think that is actually a problem. If for example the same parquet file is in two different tables, the gc algorithm won't realize that they are equal. We need to update the algo to use equality based on the only the unique part of the asset, not including any metadata that might be provided such as the name of the table for toReportableName purposes. Do you want to open an issue to resolve this?
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java, line 102 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJo-r1-102:-MTc9inr4w1MqIZOlb_4:b8tk2l5) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java#L102)):*

My thinking on json was its flexible and these objects should only ever be seen by users so human readable and easy to read (from eg javascript or python) is useful. what was your thinking around protobuf? Maybe we have protobuf for internal serialisation and json when it is written out?
</blockquote></details>

Hmm.... as I contemplate things. I wonder if we should actually be writing something that can map to a Spark struct type (somehow). That would probably be the most useful. The concern I had with JSON was that it isn't doesn't handle versioning/schema evolution well. Since asset keys may be saved across time and then format changes would be hard to manage.
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java, line 22 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJq-r1-22:-MTcADbF7A8bQEeSd--b:b-wdyzkb) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/AssetKeyType.java#L22)):*

Forgot to comment on the package. Thats not a problem, I don't mind. It seems excessive to me as its only the `IcebergTable` to `Stream<AssetKey>` code that is moving and the iceberg deps are already marked provided. But I think that the outcome of the design discussions would change things. Will wait till the design discussion wraps up before moving.
</blockquote></details>

I think it would be good to have a zoom chat on the design pattern. I initially saw the asset key converters to be pluggable as opposed to being coupled to the gc algo itself (similar to how the versionstore is agnostic to what is stored in it).
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJs-r1-43:-MTcA_ncBoboIC6TY8IZ:b-dn17oc) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java#L43)):*

I agree, I don't like the way the Hadoop config has to float around everywhere. Its hard to get away from it though...maybe a filesystem path doesn't have config at all and needs to be handed a config object to do its work?

Not sure where to source the general gc config from or who would own it? Typically in a spark job all hadoop configs are derived from the spark owned one. So if I understand what you were thinking was to overlay w/ an 'admin' like set of settings? 

The config being passed in here only needs permissions to read the metadata so the users creds should be sufficient right?

My thinking is that, at least initially, teh most common GC workflow is for a super user to run the job and inspect the results. Followed by a second run to actually delete if they are happy.  The 2nd run should definitely be done by a user that has the permissions to delete the data so their hadoop config should be ok. What was your thinking there?
</blockquote></details>

This goes to my thought that the individual asset handlers might have their own configuration properties. Also, as I reflect on this, would it be important to include the table properties that may be applied to the connection properties (does that exist?) so that different tables may use different connection properties to get to the files (?)
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJs-r1-57:-MTcAzNm0J9u5_a1BzsS:bpx1jje) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/assets/IcebergAssetKeyReader.java#L57)):*

Yeah, I am frustrated by this. I want to keep hold of all snapshots too as the metadata files are technically invalid for iceberg timetravel if we delete snapshots w/o re-writing metadata files.

However then we will never get rid of anything, from what I can tell. http://iceberg.apache.org/maintenance/ implies that the only way to expire snapshots is to run an action which is destructive and not branch aware. 

Its rather complicated otherwise. I have to think more about it but it probably means re-writing every metadata file in every valid branch using the expire snapshots action and a dummy deleter. Basically re-writing all of history in order to expire a snapshot from all metadata files. Iceberg doesn't read metadata files and only tracks the snapshot list in the most recent manifest file as opposed to nessie reading all metadata files that it has a reference to.

A halfway solution is to run the expire snapshots action with a dummy deleter and just wait for the old metadata to roll off through other expiration techniques. I will investigate this option further now.

This raises other questions that worry me around table maintenance. If any of the table maintenance actions in iceberg run while the table is under nessie we are in a heap of trouble. Our GC algo doens't take into consideration orphan files either, which involves `ls`-ing the entire data/metadata directories looking for unreferenced items on-top of the existing gc algo.

Apologies for the long ramble, I was using Reviewable as a rubber duck.
</blockquote></details>

This feels like it is outside the scope of a gc concern. GC is just about stating that a particular metadata file has a set of associated assets (including all snapshots). Separately, we should make a snapshot removal operation that is shallow (just creates a new metadata file). The rest of the operation would be handled by the gc algorithm since assets associated with the snapshot would be valid for original metadata but not metadata'.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 9 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/803#-MTef4bs-xyxRoEmKXAe:-MTef4bs-xyxRoEmKXAf:b-rfpnka) (no related file):*
Just added a few replies. I think a call is the next course of action atm

---
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java, line 55 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJm-r1-55:-MTeaQW8Eaq1rVysZTWC:bcy57rc) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/RefToL2Producer.java#L55)):*

My main thought would be that this only needs to be changed for tests. For a real world scenario, having a conflict here should be a assumption failure.
</blockquote></details>

I agree, I will look into the possibility of clearing state in a test. Spark is annoying that way so I am not hopeful.
___
*[versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJt-r1-37:-MTeafks7tj4D6IurBhl:b-bacvjl) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/org/apache/iceberg/DataFileCollector.java#L37)):*

I suggest we update the api in Iceberg to expose a Stream. Collapsing to Set for multiple snapshots makes sense but I'd be inclined to maintain a set of manifest identifiers as opposed to the entire manifest file object.

It actually begs an interesting question. Right now the GC algo uses equality against the serialized representation of the entire asset key. I think that is actually a problem. If for example the same parquet file is in two different tables, the gc algorithm won't realize that they are equal. We need to update the algo to use equality based on the only the unique part of the asset, not including any metadata that might be provided such as the name of the table for toReportableName purposes. Do you want to open an issue to resolve this?
</blockquote></details>

Re the Iceberg api. I think this is a non-trivial change to Iceberg. The list is taken directly from an avro iterator and cached internally by the `BaseSnapshot` https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/BaseSnapshot.java#L139 This includes the manifest list and a copy of it for deletes/data files. I think switching this to a stream could be fairly invasive and at the end it still will need caching. Im currently reading through the iceberg actions to see how they deal w/ the same issue

to your 2nd point I agree and have been playing around with subclassing `AssetKey` to try and figure this out. Can discuss more on the architecture call?
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java, line 102 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJo-r1-102:-MTedVDp0OLAGfcZwwpc:bxv1kjr) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java#L102)):*

Hmm.... as I contemplate things. I wonder if we should actually be writing something that can map to a Spark struct type (somehow). That would probably be the most useful. The concern I had with JSON was that it isn't doesn't handle versioning/schema evolution well. Since asset keys may be saved across time and then format changes would be hard to manage.
</blockquote></details>

Interesting, my thought was that they shouldn't be saved across time. At least across long times/version changes. I like the idea of spark rows though. In the end I want to be able to show this first as a spark table, then via python or js via the UI. Thinking more, writing an iceberg table and storing it in Nessie is interesting too...another one for the call...
___
*[versioned/tiered/gc/src/test/java/com/dremio/nessie/versioned/gc/ITTestIdentifyUnreferencedAssetsIceberg.java, line 89 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJu-r1-89:-MTef1fC0D0i68JcciRb:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/test/java/com/dremio/nessie/versioned/gc/ITTestIdentifyUnreferencedAssetsIceberg.java#L89)):*

it would be nice to have some tests that just unit test your new worker to confirm that you're getting the right set of assets for things like multiple snapshots, etc.
</blockquote></details>

Done.

rymurr:
This has now addressed all the previous comments and is up to date. Will move into open state once #843 is merged

rymurr:
After rebasing w/ the split gc changes this is ready for review

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 14 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java, line 34 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmrR8-IUCXpgUiByc:-MUZmrR8-IUCXpgUiByd:b-w52ggu) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java#L34)):*
> ```Java
> 
> 
> public class IcebergAssetKeyConverter implements AssetKeyConverter<Contents, IcebergAssetKey>, Serializable {
> ```

Can we come up with a way to avoid doing this extension? This feels like bad oo
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java, line 68 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmdrY05b355XYUXvN:-MUZmdrY05b355XYUXvO:bft817g) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java#L68)):*
> ```Java
>   @Override
>   public Stream<IcebergAssetKey> apply(Contents contents) {
>     Preconditions.checkArgument(contents instanceof IcebergTable);
> ```

Maybe add a message here identifying what happened on failure.
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmmfz-yCpkzeerwli:-MUZmmfz-yCpkzeerwlj:b-w52ggu) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java#L31)):*
> ```Java
> import com.google.protobuf.ByteString;
> 
> public class IdentifyUnreferencedIcebergAssets extends IdentifyUnreferencedAssets<Contents, IcebergAssetKey> {
> ```

Can we come up with a way to avoid doing this extension? This feels like bad oo
___
*[versioned/gc/iceberg/src/main/proto/assetkey.proto, line 23 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmuh-6d64YTt58KVG:-MUZmuh-6d64YTt58KVH:b-b2dho4) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/proto/assetkey.proto#L23)):*
> ```protobuf
> option java_generate_equals_and_hash = true;
> 
> message PIcebergAssetKey {
> ```

Can't there be other properties that are necessary for the assets? For example a bunch of name/value pairs from the table's properties. Remember that the job that deletes need to be able work solely from the asset key and won't necessary have access to table metadata from something like hive. (For example, imagine some properties that replace the fileio).
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 35 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmFJS7rxe9atA1OGl:-MUZmFJTCRkDUD1v49gt:bsnzxn5) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L35)):*
> ```Java
>  * Store worker specific to GC ops. Requires the base StoreWorker used by the server.
>  */
> public class GcStoreWorker implements StoreWorker<Contents, CommitMeta> {
> ```

Why is this still necessary?
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 98 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmQbn3KNyWZwFcBwQ:-MUZmQbn3KNyWZwFcBwR:berio2s) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L98)):*
> ```Java
>     @Override
>     public void writeExternal(ObjectOutput objectOutput) throws IOException {
> ```

what purpose does this serve?
___
*[versioned/gc/iceberg/src/test/java/org/projectnessie/versioned/gc/ITIcebergAssetKeyReader.java, line 71 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZnGiKAub505yD6j2q:-MUZnGiKAub505yD6j2r:b-t4kru9) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/test/java/org/projectnessie/versioned/gc/ITIcebergAssetKeyReader.java#L71)):*
> ```Java
> 
>   @TempDir
>   static File ALLEY_LOCAL_DIR;
> ```

alley?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 10 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java, line 34 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmrR8-IUCXpgUiByc:-MUhJnXF81wgonJWjbm9:b-ulomth) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java#L34)):*

Can we come up with a way to avoid doing this extension? This feels like bad oo
</blockquote></details>

Not sure what you mean here. Its an interface that is meant to be extended right?
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java, line 68 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmdrY05b355XYUXvN:-MUhKizC3S8_8RUDb7ft:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java#L68)):*

Maybe add a message here identifying what happened on failure.
</blockquote></details>

Done.
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmmfz-yCpkzeerwli:-MUhKrObDkQsxNfpIqQh:b-1fnlfb) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java#L31)):*

Can we come up with a way to avoid doing this extension? This feels like bad oo
</blockquote></details>

can you be more specific about what you don't like about it? I agree its a bit lazy but as the entry point to a spark script i wasn't overly worried. Happy to change if you had something in mind. Eg I can add an abstract method 'filter' to `IdentifyUnreferencedAssets`?
___
*[versioned/gc/iceberg/src/main/proto/assetkey.proto, line 23 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmuh-6d64YTt58KVG:-MUhMGe4EiQfqY3ggl2r:bugb3qj) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/proto/assetkey.proto#L23)):*

Can't there be other properties that are necessary for the assets? For example a bunch of name/value pairs from the table's properties. Remember that the job that deletes need to be able work solely from the asset key and won't necessary have access to table metadata from something like hive. (For example, imagine some properties that replace the fileio).
</blockquote></details>

To me the iceberg asset key is the path and just the path. Eg you just need the path to view/delete the asset key. The other stuff is extra info that is easy to pull off the metadata while generating the keys and helps the user identify what is what.

Things like eg how to build the `FileIO` to me are the responsibility of the calling application. I feel like we risk leaking potentially sensitive filesystem config information if we start sticking those types of properties into the asset key. At the very least we split up the information required to read/delete those files. Some are in the hadoop config of the spark session others are in the asset key. The spark session that runs the identification jobs or the delete jobs should be configured with the neccessary properties and permissions to perform those actions and we shouldn't carry that information in the object. I think this statement is valid in the context of iceberg keys only and may be revised as other asset key types are built.
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 35 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmFJS7rxe9atA1OGl:-MUhHmPW6bzXU9E3Emja:bo6xcsj) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L35)):*

Why is this still necessary?
</blockquote></details>

it serves up serializable versions of the store Serializers for spark. Externalizable is used to make it properly serializable. Otherwise we add `Serializable` to the `Serializer` interface. Which felt weird to me
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 98 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmQbn3KNyWZwFcBwQ:-MUhJhOk6l9rVfS1wkL1:byr4ffv) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L98)):*

what purpose does this serve?
</blockquote></details>

see above
___
*[versioned/gc/iceberg/src/test/java/org/projectnessie/versioned/gc/ITIcebergAssetKeyReader.java, line 71 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZnGiKAub505yD6j2q:-MUhOIow66MXggzYRkWR:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/test/java/org/projectnessie/versioned/gc/ITIcebergAssetKeyReader.java#L71)):*

alley?
</blockquote></details>

Done.

rymurr:
Thanks a lot for the comments @jacques-n. I have addressed most and asked for a bit of clarification on others. I think the main sticking point is us seeing asset keys differently

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 5 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java, line 34 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmrR8-IUCXpgUiByc:-MUkapNLAIvAeuy_wNOk:b-eqy65x) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IcebergAssetKeyConverter.java#L34)):*

Not sure what you mean here. Its an interface that is meant to be extended right?
</blockquote></details>

yeah ignore this comment. really only applied to below.
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmmfz-yCpkzeerwli:-MUkau4kAAObdF7YiBP6:bf4ujps) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java#L31)):*

can you be more specific about what you don't like about it? I agree its a bit lazy but as the entry point to a spark script i wasn't overly worried. Happy to change if you had something in mind. Eg I can add an abstract method 'filter' to `IdentifyUnreferencedAssets`?
</blockquote></details>

Instead of making IdentifyUnreferencedAssets be something that is extended, I think it would be better to pass something into it's constructor (filter factory or similar). Extension like this just feels squishy. It also makes unit testing less clean (as opposed to just working with a filter factory or whatever)
___
*[versioned/gc/iceberg/src/main/proto/assetkey.proto, line 23 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmuh-6d64YTt58KVG:-MUkbDShBW2CL_yNqQE4:b-vt41g) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/proto/assetkey.proto#L23)):*

To me the iceberg asset key is the path and just the path. Eg you just need the path to view/delete the asset key. The other stuff is extra info that is easy to pull off the metadata while generating the keys and helps the user identify what is what.

Things like eg how to build the `FileIO` to me are the responsibility of the calling application. I feel like we risk leaking potentially sensitive filesystem config information if we start sticking those types of properties into the asset key. At the very least we split up the information required to read/delete those files. Some are in the hadoop config of the spark session others are in the asset key. The spark session that runs the identification jobs or the delete jobs should be configured with the neccessary properties and permissions to perform those actions and we shouldn't carry that information in the object. I think this statement is valid in the context of iceberg keys only and may be revised as other asset key types are built.
</blockquote></details>

I think there are two types of properties: system properties and table properties. Think about using hive tables. Some properties come from config but some come from the hive metastore for the table. The table level properties may be different for different tables. In those cases, I think we need to pass them here in addition to path (as I assume they could influence how this resolution would occur). Isn't it true that Hive table level properties could influence how the iceberg table file is read and must be kept? (And are you sure that none of those properties would be relevant here?)
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 35 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmFJS7rxe9atA1OGl:-MUka9lF8RxfDQBC3DHY:b-ua5kk0) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L35)):*

it serves up serializable versions of the store Serializers for spark. Externalizable is used to make it properly serializable. Otherwise we add `Serializable` to the `Serializer` interface. Which felt weird to me
</blockquote></details>

What do you think about just making Serializer implement Serializable (or externalizable). 

Externalizable shouldn't be required since these are stateless either way. You should be able to either pick Serializable or Externalizable.
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java, line 102 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJo-r1-102:-MUkbhdNCwfBtHCCX5tJ:b94mza) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java#L102)):*

Interesting, my thought was that they shouldn't be saved across time. At least across long times/version changes. I like the idea of spark rows though. In the end I want to be able to show this first as a spark table, then via python or js via the UI. Thinking more, writing an iceberg table and storing it in Nessie is interesting too...another one for the call...
</blockquote></details>

My expectation: run gc every few days. Look at the last 2-3 files and gc anything that is included in all of them. The key here being the output is written and then later consumed. If that is the case, having a consistent serialization is important.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 16 files reviewed, 5 unresolved discussions (waiting on @jacques-n and @laurentgo)

___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmmfz-yCpkzeerwli:-MUmbrKO0m7KvsHu3f_C:b-e6xqe4) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedIcebergAssets.java#L31)):*

Instead of making IdentifyUnreferencedAssets be something that is extended, I think it would be better to pass something into it's constructor (filter factory or similar). Extension like this just feels squishy. It also makes unit testing less clean (as opposed to just working with a filter factory or whatever)
</blockquote></details>

Done. Just added a filter. It feels a bit weird as the value dataset being passed in is immediately filtered by the passed in filter. It could easily be done outside before passing in the filter, we are just enforcing that it happens.
___
*[versioned/gc/iceberg/src/main/proto/assetkey.proto, line 23 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmuh-6d64YTt58KVG:-MUmcLcxBTgWwXS1kyU0:bso84l) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/proto/assetkey.proto#L23)):*

I think there are two types of properties: system properties and table properties. Think about using hive tables. Some properties come from config but some come from the hive metastore for the table. The table level properties may be different for different tables. In those cases, I think we need to pass them here in addition to path (as I assume they could influence how this resolution would occur). Isn't it true that Hive table level properties could influence how the iceberg table file is read and must be kept? (And are you sure that none of those properties would be relevant here?)
</blockquote></details>

With Hive anything is possible. However I am not sure if it applies here. This is for GC-ing iceberg tables owned by nessie. In that case the only thing we know about the table is the metadata path. That path is then transformed into all its known manifests, datafiles etc, all of which are just paths.

I don't see how we would ever see a Hive owned iceberg table (and the menagerie of properties that go along with it) in this algorithm. Now its another story if we are talking about GC-ing nessie owned hive tables. Then its likely to need a new asset key and associated proto. Off the top of my head that asset key will likely be completely different and require executing hive DML rather than manipulating paths.

I think my goal here is to keep the design effort low and relatively decoupled until this is being used in anger
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java, line 35 at r4](https://reviewable.io/reviews/projectnessie/nessie/803#-MUZmFJS7rxe9atA1OGl:-MUmbCoW7M0yb3hm4WbJ:b-x7y1q3) ([raw file](https://github.com/projectnessie/nessie/blob/08ea6ba5e24e1419855cead3f1f1e3cf021e0c6c/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/GcStoreWorker.java#L35)):*

What do you think about just making Serializer implement Serializable (or externalizable). 

Externalizable shouldn't be required since these are stateless either way. You should be able to either pick Serializable or Externalizable.
</blockquote></details>

Done. Externalizable was needed before because Spark is a bit weird. Just added serializable and deleted the GcStoreWorker
___
*[versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java, line 102 at r1](https://reviewable.io/reviews/projectnessie/nessie/803#-MTHMgs4L-_pukRREmJo-r1-102:-MUmsL1N-7HuenNp-9Bv:b-vgdnk3) ([raw file](https://github.com/projectnessie/nessie/blob/93f366f07d1af7bb78a2fe1ce49fb5e53cf7e12a/versioned/tiered/gc/src/main/java/com/dremio/nessie/versioned/gc/ContentsValueWorker.java#L102)):*

My expectation: run gc every few days. Look at the last 2-3 files and gc anything that is included in all of them. The key here being the output is written and then later consumed. If that is the case, having a consistent serialization is important.
</blockquote></details>

agreed, serialization is important. I am going to investigate how to run gc as my next patch so how to store rows and how to operate gc is goign to become very interesting. I've copied this thread over to #707

rymurr:
Just rebased, any more comments @projectnessie/core ?

jacques-n (dismissed):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/803)* status: 0 of 18 files reviewed, 2 unresolved discussions (waiting on @jacques-n and @laurentgo)

snazy (approved):

PR #804, opened 2021-02-11
WIP: Jackson-based Store worker

laurentgo:
The current store worker implementation used with the quarkus server
requires each content type to have some equivalent protobuf schema in
order to be serialized back and forth between JSON <-> POJO <-> Proto.

Replace the handwritten protobuf schemas with jackson protobuf module,
which analyze the POJO object to generate the object on the fly, and
leverage jackson to directly write the protobuf binary representation of
the object.


laurentgo:
The change needs some more comments and tests for sure, but I'm also looking for feedback on the overall approach.

jacques-n:
Have you evaluated using something like modelmapper instead? I didn't really like how this compounds the wire and serialization protocols. As the two move forward I'd expect there to be inconsistentices/migrations/etc that need to be handled that will make this difficult. Keeping them as separate tiers but with less boilerplate for conversion between would be my preference.

rymurr:
I am on the fence. I don't love the current table store worker. Its a bit loose. I agree w/ @jacques-n that tightly coupling the persistent storage to the on the wire format might not be the right answer either. Its not clear to me how we would migrate these objects but that might just be my lack of understanding of jacksons protobuf mapper.

One question I had while looking at this is what place does the `TableCommitMetaStoreWorker` have in the larger code hierarchy for Nessie. This is obviously closely related to #783. I might be trying to ask the question: is serializing values the job of hte store or the REST server? Or neither? I think my answer is neither. This PR seems to favour REST. Current impl is :man_shrugging:

snazy (changes requested):
I think, the `VersionStore` should not have to deal with the specifics of the contents at all, it should be transparent for a `VersionStore` and rather be some arbitrary "blob" (or multiple of those).

Clients of the `VersionStore` _might_ need to know about the specifics of those contents. The GC is one client that needs to know how to deal with an iceberg table or hive table or whatever. "Committers" also need to know that.

Regarding "where" the (de)serialization should happen, I could imagine that something like this _could_ work?
```
+----+
| GC |
+----+
  |
  |
+---------------+       +-----------------------------------+       +-----------------+
| Nessie Client | <---> | Contents (de)serialization module | <---> | Embedded client |
+---------------*       +-----------------------------------+       +-----------------+
       |                                                                     |
       |                                                                     |
+-------------+                                                              |
| REST client |                                                              |
+-------------+                                                              |
       |                                                                     |
       |                                                                     |
+-------------+                                                     +-----------------+
| REST server |                                                     | Embedded server | 
+-------------+                                                     +-----------------+
       |                                                                     |
       | +-------------------------------------------------------------------+
       | |
+--------------+
| VersionStore |
+--------------+
```

IMHO I'd prefer msgpack over protobuf, because it doesn't require any schema up-front and has a pretty compact serialization format.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/804#-:-MTL6HFuAU_9H9J_W__8:b-jbpw0z)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @laurentgo)

___
*[model/src/main/java/com/dremio/nessie/model/DeltaLakeTable.java, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/804#-MTKnsaqD0893CbWWQ_F:-MTKnsaqD0893CbWWQ_G:b-shpjep) ([raw file](https://github.com/projectnessie/nessie/blob/4782a522090a5ead859e92222e8026be21aba369/model/src/main/java/com/dremio/nessie/model/DeltaLakeTable.java#L35)):*
> ```Java
> public interface DeltaLakeTable extends Contents {
> 
>   @JsonProperty(index = 1)
> ```

Mind adding a comment here and in the other classes that it's required to index all properties and start at `1` because of protobuf (and not with `0`, which would be fine for `@JsonProperty`). I.e. documenting all the protobuf restrictions.
___
*[servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/TableCommitMetaStoreWorker.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/804#-MTKowIeBMC3y5q9eAtx:-MTKowIeBMC3y5q9eAty:b6jbktm) ([raw file](https://github.com/projectnessie/nessie/blob/4782a522090a5ead859e92222e8026be21aba369/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/TableCommitMetaStoreWorker.java#L57)):*
> ```Java
>         return PROTOBUF_MAPPER.generateSchemaFor(type);
>       } catch (JsonMappingException e) {
>         throw new IllegalArgumentException("Cannot generate schema for class " + type, e);
> ```

Nit: this is rather some `RuntimeException` due to a programmer's bug, not strictly an "illegal argument".

laurentgo (commented):
My main concern was not so much about having to write some mapping code vs the need of having to map things out between api and storage.
Nessie is mostly content type agnostic as of now: except for the store worker code which does some mapping, and possibly the GC code, everything else does not care of things are stored at all. Now, the `StoreWorker` interface is a weird one. Used by the server, it is only used to serialize `Content` and `CommitMeta`. Used by the GC process, it needs also to provide `AssetsKey` which require somehow to understand the persistent format used + from the content object to be able to extract some additional information. So now you can start arguing that there are 2 different sets of actors using the `VersionStore` API as their main interface, meaning the store representation become the wire representation too.
One other area of concern is the lack of extensibility regarding the kind of content which can be stored in Nessie, and the need to modify some store related code each time a content type is modified or added seems burdensome (on top of the POJO changes). It seems simpler to say that a given content type has a single representation, and that changes to the content type should be backward compatible (newer client should be able to read an older version of the JSON) and possibly forward compatible (older clients might still be able to read/use an newer version of the JSON). Or maybe content types are versioned too, and an incompatible change is basically a different type, and a different POJO?

@snazy I would believe that `protobuf` is more compact than `msgpack` since fields are referenced using index for protobuf and with the field name for `msgpack`, but to be honest, the binary serialization format was not my main concern.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/804#-:-MTN-xAGA7G2280ld_BT:b8yb5vv)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @snazy)

___
*[model/src/main/java/com/dremio/nessie/model/DeltaLakeTable.java, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/804#-MTKnsaqD0893CbWWQ_F:-MTN-bbh-wsKiuIwfLlP:bacud2y) ([raw file](https://github.com/projectnessie/nessie/blob/4782a522090a5ead859e92222e8026be21aba369/model/src/main/java/com/dremio/nessie/model/DeltaLakeTable.java#L35)):*

Mind adding a comment here and in the other classes that it's required to index all properties and start at `1` because of protobuf (and not with `0`, which would be fine for `@JsonProperty`). I.e. documenting all the protobuf restrictions.
</blockquote></details>

Yes, it would need some checks if we decide to go this route
___
*[servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/TableCommitMetaStoreWorker.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/804#-MTKowIeBMC3y5q9eAtx:-MTN-kGH-Qp-XAJfawug:b-xh0f3n) ([raw file](https://github.com/projectnessie/nessie/blob/4782a522090a5ead859e92222e8026be21aba369/servers/quarkus-server/src/main/java/com/dremio/nessie/server/providers/TableCommitMetaStoreWorker.java#L57)):*

Nit: this is rather some `RuntimeException` due to a programmer's bug, not strictly an "illegal argument".
</blockquote></details>

Correct, will change if we decide to move forward

jacques-n (commented):
I think the two consumer question shapes things somewhat differently than I think about them. I think of there being:

* core version store that includes gc operations, etc. For TieredVersionStore, we keep the GC impl separate because it depends on Hadoop/Spark but it is really part of the TieredVersionStore.
* Separately, I think of each content type as being independent and ideally pluggable. It should include how it maps to the general content object and information about things like how it should be serialized, how gc assets are exposed, etc.

We mostly exist like this today. The place where it becomes squishy is we haven't yet made the content types pluggable (and external to core code-- store worker does the gluing today). The cleanest way to resolve this would be to possibly move Contents to be a Any protobuf type and then we could use some kind of plugin module system to add other things. However, I actually like the pattern of having a rigid/well-defined core schema object that uses the one-of pattern. As such, I'd suggest that we create separate modules like "content-iceberg" and "content-hive-view" and each of those should be responsible for all the mapping etc. If we do that, I think we move to a situation where each content module can decide: how it is serialized and how it is exposed at the api level.

One random idea to achieve better modularity/pluggability but still have a well defined schema is to do protobuf preprocessing that allows us to declare each content-type's serialized representation separately and then just merge them before compiling. It feels a bit heavyweight/hacky since I think we should have a clear process and keep all content types in the Nessie project (you shouldn't be able to drop a random jar and get a new content type added-- especially since the native image doesn't like no stinkin jars).

The other sticky point about the modularity is that the inspection operations (currently only used for gc but should probably used in ui etc) require a boat-load more dependencies per content type. We may want to think about having a "core" module and an "inspection" for each content type. Tools like GC and UI would need to use inspection while things like commit operations would only need core.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/804#-:-MTcEWEG3ieyNNGYvQ6G:b-n2l2o1)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @snazy)

rymurr:
Ive spent entirely too much time thinking about this and I have come up with the following.

the idea of splitting all Content types into different modules makes me rather uneasy. Having 4+ more modules with 10 lines of code each seems excessive for the stated goal (if you were so inclined you could invoke DRY and YAGNI at this stage).

I think the stated goal is the following: 'create good separation and modularity' but that can take several forms. 

Form 1 decoupling for serialization and extensibility:
Separate modules to me means each module could in theory have its own on-the-wire impl and storage impl. It also implies pluggable modules. Pluggability is not a goal to me: clients need all types as a client doens't know whats on the server and a server needs all types as it doesnt know what a client might want to do. We end up having to reference everywhere and distribute all these modules anyways so pluggability isn't realistic. Pluggable modules also implies extensibility which I think we don't want (and can't easily have in the native image).

In terms of serialization we have no plans to support anything other than json on the wire and I can't imagine a scenario where different Content types would need different on the wire protocols. Similar for storage format: while its theoretically nice to be able to store ContentA as json and ContentB as protobuf I don't think there is a burning need to do so atm.

So to me having a bunch of modules creates a bunch of overhead for something that in the end has to be loosely coupled anyways (eg some module needs to know about all types).

Form 2 How easy is it to add new Content types
In my mind a decoupled system should make it easier for an implementer to add new Content types. Eg we don't have to force the implementer to touch version store, store, on the wire formats, on disk formats and rest api. That to me is the type of coupling we want to avoid. This patch would make adding a new content type as easy as adding a subclass to `Contents` and (potentially) updating a few jackson annotations (caveat there is a few tricks when defining that object to make it compatible w jackson and protobuf). To me that is already quite good. Especially given that the current count of `Contents` implementers is 5 and as likely to decrease as increase (realistically in the next few years I don't see this number going above 10 and if we do :tada:). Conversely adding a new Contents type if we have several modules means adding a module (and all the pom boilerplate), a protobuf (or equivalent storage format), updating several poms to expose to client and server and likely twiddling some config classes in the server (or at least doing something to make sure service loading works). That speaks to me of tight coupling (of an invisible sort). 

I guess in summary a bunch of small modules is to me a strong code smell and we should have a good reason for doing it given how not bad the current impl is. This patch and #783 combined are becoming more and more sensible to me. With #783 separating the store worker impl from the rest api and version store.

PR #810, opened 2021-02-12
Rename com.dremio.nessie package to org.projectnessie

laurentgo:
To better reflect the name of the project, rename classes from
`com.dremio.nessie` to `org.projectnessie` (same number of characters).

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/810)


rymurr (approved):
Dont love the mixing of package moves w/ the base package rename. Not bad enough to leave this large/disruptive patch open though.

laurentgo:
I did basic package rename so that to match conventions. Since basically every class was modified, that would have no impact on the number of line changed.

Thanks for merging

Issue #812, opened 2021-02-12
REST API change: Move `message` parameter of ContentsApi.setContents to POST body


snazy:
A commit message can be very long, but since it is submitted in the URL, it is implicitly restricted by "practical limits" in the network infrastructure.

Although the HTTP RFCs do not put a limit on the length of a URL, actual HTTP implementations _can_ put a limit and reply with a HTTP/414, if the request URL is too long.

Applies to
* `ContentsApi.setContents`
* `ContentsApi.delete`
* `TreeApi.commitMultipleOperations`
* `TreeApi.transplantCommitsIntoBranch`

rymurr:
I am not 100% sure what to do here. The practical limit for this seems to be upwards of 8k in most places. Old IE being the exception and limiting to 2048 chars. So this puts a realistic upper limit on the commit message however we prob would want to restrict its size anyways at some stage as 8k is a lot of space for a message.

By adding the data to the body we now have to expose a commit metadata like object in the public api rather than assembling it from query params and credentials. And the body then becomes a composite of the commit meta and the actual contents object. Making it harder to implement in other languages and cascading a bunch of changes into js and python.

As part of #779 I am proposing moving to headers for new commit meta related items and leaving the commit message as is. Headers have an 8k limit each in the oldest web servers and are less disruptive to the already published spec. Wondering what other peoples thoughts are?

rymurr:
Having said that the commit meta object is already exposed via the log api so maybe it isn't so bad...

snazy:
The question here is not only whether a specific browser or particular HTTP endpoint can handle 4k or 8k URIs or header-fields. The question is about whether individual HTTP libraries and network components (proxies, firewalls, etc) can handle that, which I frankly doubt.

Implicitly adding a limit by using a HTTP header for e.g. the commit-message makes Nessie dependent on limitations of network infrastructure and setup. Some network-infrastructure might just block unknown headers due to security policies that some people have in place.

Using headers also makes it impossible to use compression for e.g. the commit-message.

It would be very disappointing for Nessie users, if their application completely stops or fails in an unrecoverable way, because some automatically generated commit message is 1 byte too big.

rymurr:
> It would be very disappointing for Nessie users, if their application completely stops or fails in an unrecoverable way, because some automatically generated commit message is 1 byte too big.

[...]

Issue #779, opened 2021-02-08
Add extra changes properties (committer...)


laurentgo:
Add some missing change properties to Nessie changes:
- author
- committer
- date of change
- date of commit

PR #818, opened 2021-02-15
Add more commit info and update REST api

rymurr:
This addresses #779 to add extra commit info to the REST api. It also addresses #812.

I took the approach of adding headers and am looking for early feedback. The 3 approaches I looked at:

1. adding to the body. This was inconsistent as I would have to add a `CommitMeta` object to the `Operations` object to satisfy the multi-update endpoint and add a container to hold the `Contents` and `CommitMeta`  for the single-update enpoint. The delete contents endpoint would need another approach as its `DELETE` and not all servers accept bodys in delete calls. Due to the inconsistency here I thought it wasn't an awesome direction
2. adding query params. As per #812 this only works to some implementation dependent level that is likely in the region of 8k. While its unlikely that all query params + the URL would add up to that high a value I didn't want to risk it, as in theory a commit message could be large
3. adding headers. This approach is outlined in this PR. It also suffers a limit of arbitrary max length but this is ~8k per value as opposed to the entire URL. This effectively puts a limit on the commit message size but 8k is a large limit and I think ok.

Looking for feedback before implementing anything further and seeing if anyone prefers options 1 or 2

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/818)


rymurr:
I've also removed the `message` from `transplant` as it isn't being used anywhere.

snazy (commented):
Not a fan of such super long  repeated `String` parameter lists. Too easy to introduce hard to detect bugs.

HTTP headers are actually also not suitable for such large amounts of request data as that makes Nessie dependent on actual network infrastructure details (and people really have strange network stuff in place).

Another issue with HTTP headers is that it requires many different headers, which makes the HTTP header list very long and there's no validation on HTTP headers per-se - i.e. you could easily use `X-Nessie-Committer-Emall` and the field would just be missing in Nessie due to a simple typo. And it's not detectable later in a debugger or added debug-logging, because the HTTP request information isn't passed down to the `VersionStore`.

I really prefer adding all commit-related information to the payload. If that requires making that single `ContentsApi.deleteContents` a `POST` operation, that's unfortunate, but IMO way better than all the foreseeable and unforeseeable issues with using many/huge HTTP headers.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/818#-:-MTeD5lPCtpvNjNDaFcT:b-ivicqa)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[model/src/main/java/org/projectnessie/api/NessieHeaders.java, line 18 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTe3lKkCWqvngIneZJO:-MTe3lKkCWqvngIneZJP:b-2vg408) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/api/NessieHeaders.java#L18)):*
> ```Java
> package org.projectnessie.api;
> 
> public final class NessieHeaders {
> ```

Custom HTTP headers must start with `X-`

rymurr:
> Custom HTTP headers must start with `X-`

I thought so too. Deprecated since 2012 https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

rymurr:
I also don't love the headers but I still am not convinced the solution is to pollute our model objects. I am happy to deprecate or change to post the `DELETE` method though, its a specialization of the commit.

snazy:
I'd add a `CommitMetadata` to the API level and pass it to the relevant methods/functions and let the implementation deal with it. E.g. on the client side:
```
  @Override
  public void commitMultipleOperations(
      String branch,
      @NotNull String expectedHash,
      // metadata like committer, author, message, etc for this commit
      @NotNull CommitMetadata commitMetadata,
      // the actual operations for this commit
      @NotNull Operations operations) throws NessieNotFoundException, NessieConflictException {
    client.newRequest().path("trees/branch/{branchName}/commit")
          .resolveTemplate("branchName", branch)
          .queryParam("expectedHash", expectedHash)
          // "Merge" commit-metadata and commit-operations
          .post(new CommitOperations(commitMetadata, operations));
  }
```

Just thinking: maybe get rid of `ContentsApi.setContents` + `ContentsApi.deleteContents`, because it's just a specialization of `TreeApi.commitMultipleOperations`? Or let `ClientContentsApi.setContents`+`deleteContents` delegate to `TreeApi.commitMultipleOperations`?

rymurr:
I like the idea of simplifying the broad set of ops too. At the end of the day they are just specialisations of multi-put. Just adding a commit meta to the Operations list is a much easier change to me. 

Since delegating to `commitMultipleOperations` from the `ContentsApi` endpoints would end up adding a commit meta object to those endpoitns I am thinking just getting rid of them rather than delegation. a) its duplicate functionality b) it hides the fact that all ops in Nessie are effectively multi-table transactions. Which we have seen time and time again is an important feature of Nessie so why hide it!

rymurr:
I will update this patch w/ the removal of delete and put endpoints and we can see how things look.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/818)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @jacques-n and @rymurr)

___
*[clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java, line 150 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTfhPi8ATgeBvApddZJ:-MTfhPi8ATgeBvApddZK:b2npx2d) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java#L150)):*
> ```Java
>       @NotNull String expectedHash,
>       String message,
>       String author,
> ```

shouldn't we have a richer object to capture all fields at once instead of having a very large method?
___
*[model/src/main/java/org/projectnessie/api/NessieHeaders.java, line 18 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTe3lKkCWqvngIneZJO:-MTiVJbr5nlGezH_t6PK:bhfucda) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/api/NessieHeaders.java#L18)):*

Custom HTTP headers must start with `X-`
</blockquote></details>

I think @rymurr already mentioned it, but it's actually not true anymore.
___
*[model/src/main/java/org/projectnessie/api/TreeApi.java, line 332 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTiVfED2PwdDdBnAdNc:-MTiVfED2PwdDdBnAdNd:b-no3w8x) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/api/TreeApi.java#L332)):*
> ```Java
>           String hash,
>       @Parameter(description = "Commit message")
>       @HeaderParam(NessieHeaders.MESSAGE_HEADER_NAME)
> ```

From an API point of view, I think this has poor discoverability and usability, and we should consider making it part of the body instead
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 70 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTiW5Fq5qINt6Nav8h8:-MTiW5Fq5qINt6Nav8h9:bt54sws) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/model/CommitMeta.java#L70)):*
> ```Java
>   @Value.Default
>   public Map<String, String> getProperties() {
>     return new HashMap<>();
> ```

shouldn't it be `Collections.emptyMap()`

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/818)* status: 0 of 9 files reviewed, 4 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java, line 150 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTfhPi8ATgeBvApddZJ:-MTjDd431-XdKvb-0uQb:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java#L150)):*

shouldn't we have a richer object to capture all fields at once instead of having a very large method?
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/api/NessieHeaders.java, line 18 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTe3lKkCWqvngIneZJO:-MTjDecq72zJ0xIQyeRJ:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/api/NessieHeaders.java#L18)):*

I think @rymurr already mentioned it, but it's actually not true anymore.
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/api/TreeApi.java, line 332 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTiVfED2PwdDdBnAdNc:-MTjDgFp-fk10A-LnjZ9:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/api/TreeApi.java#L332)):*

From an API point of view, I think this has poor discoverability and usability, and we should consider making it part of the body instead
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 70 at r1](https://reviewable.io/reviews/projectnessie/nessie/818#-MTiW5Fq5qINt6Nav8h8:-MTjDiA42ZiiXDMCglOr:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/0ce39d293d8fadf580c7c54cdd80a2b356ae521f/model/src/main/java/org/projectnessie/model/CommitMeta.java#L70)):*

shouldn't it be `Collections.emptyMap()`
</blockquote></details>

Done.

clients/client/src/main/java/org/projectnessie/client/ClientContentsApi.java
```diff
@@ -18,57 +18,20 @@
...
-    return client.newRequest().path("contents")
-                 .queryParam("ref", ref)
-                 .post(request)
-                 .readEntity(MultiGetContentsResponse.class);
-  }
 
+/**
+ * Extends ContentsApi with some helper methods for clients.
+ */
+public interface ClientContentsApi extends ContentsApi {
```
rymurr, line 30: This extends `ContentsApi` to provide helper functions for updating and deleting single contents. Currently only tests use these methods so not sure if its really required to have them or if extending the interface is the cleanest way

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
```diff
@@ -113,13 +113,11 @@ public LogResponse getCommitLog(@NotNull String ref) throws NessieNotFoundExcept
   public void transplantCommitsIntoBranch(
       @NotNull String branchName,
       @NotNull String expectedHash,
-      String message,
       @Valid Transplant transplant)
       throws NessieNotFoundException, NessieConflictException {
     client.newRequest().path("trees/branch/{branchName}/transplant")
           .resolveTemplate("branchName", branchName)
           .queryParam("expectedHash", expectedHash)
-          .queryParam("message", message)
```
rymurr, line 122: unused parameter

clients/deltalake/core/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
```diff
@@ -150,9 +150,10 @@ class NessieLogStore(sparkConf: SparkConf, hadoopConf: Configuration)
   private def commit(path: Path, ref: String, hash: String, message: String = "delta commit", lastCheckpoint: String = null): Boolean = {
     val targetRef = if (ref == null) reference.getName else ref
     val targetHash = if (hash == null) reference.getHash else hash
-    val messageWithSparkId = s"$message ; spark.app.id=${sparkConf.get("spark.app.id")}"
+    val meta = ImmutableCommitMeta.builder.message(message).putProperties("spark.app.id", sparkConf.get("spark.app.id")).build
```
rymurr, line 153: example usage of properties map

model/src/main/java/org/projectnessie/api/ContentsApi.java
```diff
@@ -84,70 +82,4 @@ public MultiGetContentsResponse getMultipleContents(
           MultiGetContentsRequest request)
       throws NessieNotFoundException;
 
-  /**
-   * create/update an object on a specific ref.
-   */
-  @POST
```
rymurr, line 90: remove two endpoints that are unused by clients (hms, iceberg, delta) and are really just helper methods for multi-commit. Thoughts?

model/src/main/java/org/projectnessie/model/CommitMeta.java
```diff
@@ -31,15 +34,46 @@
   @Nullable
   public abstract String getHash();
 
-  public abstract String getCommiter();
+  @Value.Default
+  public String getCommiter() {
```
rymurr, line 38: defaults are used to help clients quickly construct commit meta. Can be reset for consistency in REST client though. Committer is the system of record/auditable committer as it is set by REST server w/o overrides.

model/src/main/java/org/projectnessie/model/CommitMeta.java
```diff
@@ -31,15 +34,46 @@
...
+
+  /**
+   * according to git the author time == the commit time unless set otherwise.
+   */
   @Nullable
-  public abstract Long getCommitTime();
+  public abstract Long getAuthorTime();
+
+  @Value.Default
+  public Map<String, String> getProperties() {
```
rymurr, line 74: structure used to store kv pairs such as spark app id.

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -90,11 +91,14 @@ protected void doOps(String branch, String hash, String message, List<org.projec
     }
   }
 
-  private static CommitMeta meta(Principal principal, String message) {
-    return ImmutableCommitMeta.builder()
-        .commiter(principal == null ? "" : principal.getName())
-        .message(message == null ? "" : message)
-        .commitTime(System.currentTimeMillis())
-        .build();
+  private static CommitMeta meta(CommitMeta commitMeta, Principal principal) {
```
rymurr, line 94: This may overwrite committer from what the user sets. Thoughts?

rymurr (commented):
Some comments to further explain this PR. Note #831 as a follow up

rymurr:
closing in favour of #949

PR #752, opened 2021-02-02
RocksDB rebase and add ConditionExpression handling

stevelorddremio:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/752)

PR #819, opened 2021-02-15
RocksDB implementation

stevelorddremio:
This contains the Store implementation using RocksDB. It does not include implementation for UpdateExpression, nor is it available as a selectable Nessie backing store.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)


snazy (changes requested):
Did a first pass on the implementation.

As mentioned in #752, the methods on the `RocksBaseValue`+extending classes should (actually: must) interpret the passed-in parameters and put those into the protobuf-builder without buffering the information.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819#-:-MTeeOOMA1s3LRaNxOBM:b-2228l3)* status: 0 of 30 files reviewed, 18 unresolved discussions (waiting on @stevelorddremio)

___
*[versioned/tiered/mongodb/src/main/java/org/projectnessie/versioned/mongodb/MongoBaseValue.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZdtaDft3_ktk_Ap1:-MTeZdtaDft3_ktk_Ap2:b-cjj56y) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/mongodb/src/main/java/org/projectnessie/versioned/mongodb/MongoBaseValue.java#L57)):*
> ```Java
>   public C id(Id id) {
>     if (!properties.contains(ID)) {
>       // MongoSerDe calls this method as the very first one during serialization to satisfy the
> ```

This is an unrelated change.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeYYInE4tDmznoe8g7:-MTeYYInE4tDmznoe8g8:bxistpw) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L118)):*
> ```Java
>         valueTypeToColumnFamily.values().forEach(cf -> {
>           try {
>             rocksDB.flush(options, cf);
> ```

What does a `flush` do exactly? Is it mandatory to make (unflushed) changes persistent?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 261 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_TgW84pnzn2Xe0AP:-MTe_TgW84pnzn2Xe0AQ:bunxc2i) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L261)):*
> ```Java
> 
>     final Iterable<Acceptor<C>> iterable = () -> new AbstractIterator<Acceptor<C>>() {
>       private final RocksIterator itr = rocksDB.newIterator(columnFamilyHandle);
> ```

Nit: rename to rocksIter or something.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 282 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_HOz7EByh4SHNJth:-MTe_HP-A32u6qiWEOue:b-7rq6cl) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L282)):*
> ```Java
>     };
> 
>     return StreamSupport.stream(iterable.spliterator(), false);
> ```

This needs to handle the Stream-close to close the `RocksIterator`
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 342 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZ4f53tCvg6pd6dgJ:-MTeZ4f53tCvg6pd6dgK:b-pxcf4h) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L342)):*
> ```Java
>   private void checkValue(byte[] value, String operation, Id id) {
>     if (null == value) {
>       throw new NotFoundException("Unable to " + operation + " item with ID: " + id);
> ```

String.format
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 380 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTed5RA-FO2GAXD9WKJ:-MTed5RB5jVYtkK_LhGV:bs9mho7) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L380)):*
> ```Java
> 
>   private String verifyPath() {
>     final File dbDirectory = new File(config.getDbDirectory());
> ```

Nit: I'd prefer `java.nio.Path` here.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java, line 26 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeXhNX5KVw19zXI9ie:-MTeXhNY5OQ5j5mxqAID:bmc9gqt) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java#L26)):*
> ```Java
>   @Default
>   public String getDbDirectory() {
>     return "";
> ```

Returning an empty string here as a default value is dangerous, because `new File("")` "somewhat" translates to the current directory.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java, line 54 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZtF2DH04q6u3dNAQ:-MTeZtF2DH04q6u3dNAR:b-wjmkvi) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java#L54)):*
> ```Java
>   @Override
>   public Function visit(Entity entity) {
>     throw new UnsupportedOperationException();
> ```

Add some information to the exception (e.g. `entity.toString()`)
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java, line 86 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZzBoCs3zOkXxFpz9:-MTeZzBoCs3zOkXxFpzA:b-wjmkvi) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java#L86)):*
> ```Java
>   @Override
>   public Function visit(ExpressionPath value) {
>     throw new UnsupportedOperationException();
> ```

Add some information to the exception (e.g. `entity.toString()`)
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 90 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_jTl2mO1O-LTs_B3:-MTe_jTl2mO1O-LTs_B4:bqh85li) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L90)):*
> ```Java
>         }
>       default:
>         // Invalid Condition Function.
> ```

If it's something invalid, why not throw here?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_uiXCnJb3hlQJnZ6:-MTe_uiXCnJb3hlQJnZ7:b-6qb0tx) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L47)):*
> ```Java
> 
>   private Id metadataId; // commitMetadataId
>   private Stream<Id> parentList; // ancestors
> ```

Don't save the `Stream`s and other fields. `Stream`s that come from other store implementations can be based on currently open resources (e.g. a JDBC ResultSet), which are closed, when this object's `build()` method's called.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 140 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeajAK1SuahBO1T942:-MTeajAK1SuahBO1T943:bktb99v) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L140)):*
> ```Java
>     checkPresent(keyMutations, KEY_LIST);
> 
>     final ValueProtos.L1.Builder builder = ValueProtos.L1.newBuilder()
> ```

All `build()` methods should basically only do some post-validation and just the `builder.build().toByteArray()`.
The actual additions (like `builder.addAllTree(buildIds(tree))`) should happen in the "consuming methods (like `children(Stream<Id>`).
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTec-QLA4norXesiOzv:-MTec-QLA4norXesiOzw:bcq0vdo) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java#L37)):*
> ```Java
> 
>   @Override
>   public boolean evaluate(Function function) {
> ```

Looks identical to `RocksCommitMetadata`?
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 38 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTee1Ka8ddkSK_V525c:-MTee1Ka8ddkSK_V525d:brrj6lv) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L38)):*
> ```Java
>   @AfterAll
>   static void tearDown() throws IOException {
>     for (Path path : ImmutableList.of(DB_PATH, getRawPath())) {
> ```

Why the concatenation?
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 39 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTedf1g9sAzvgGdwKCQ:-MTedf1g9sAzvgGdwKCR:b-fl5nwm) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L39)):*
> ```Java
>   static void tearDown() throws IOException {
>     for (Path path : ImmutableList.of(DB_PATH, getRawPath())) {
>       if (path.toFile().exists()) {
> ```

`Files.exists` as mixing `java.io.File` with the `Files` API isn't nice.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTecWRP3i84r_wXFwye:-MTecWRP3i84r_wXFwyf:b-21axj) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L40)):*
> ```Java
>     for (Path path : ImmutableList.of(DB_PATH, getRawPath())) {
>       if (path.toFile().exists()) {
>         Files.walk(path).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
> ```

Why map to a `File` and not use `java.nio.file.Files#delete`? Better also do this "cleanup" in a `@BeforeAll`, so the test starts with a clean state.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 450 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeJAP81qB8m9fcL3q:-MTeeJAP81qB8m9fcL3r:bcgjvre) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L450)):*
> ```Java
>     void putWithFailingConditionExpression() {
>       final InternalRef sample = SampleEntities.createBranch(random);
>       putThenLoad(ValueType.REF, sample);
> ```

Why this change?
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 483 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeKkB6pXKKjM5AdMH:-MTeeKkB6pXKKjM5AdMI:bv3k2ps) ([raw file](https://github.com/projectnessie/nessie/blob/bf8994c202a3aaa4363b8d61ca964b0926df256a/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L483)):*
> ```Java
>           Assertions.fail(cfe);
>         }
> ```

Why this change?

rymurr (changes requested):
> Reviewed 21 of 30 files at r1.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 21 of 30 files reviewed, 25 unresolved discussions (waiting on @stevelorddremio)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf5oXb-Ov1VMFGJy2f:-MTf5oXc4XRrS9bHDL31:b-lsqort) (no related file):*
Few minor comments from me. I think less use of raw `byte[]` would be nice.

---
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksBaseValue.java, line 140 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeydDKBC6bK8u-e_7d:-MTeydDKBC6bK8u-e_7e:b76kd7e) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksBaseValue.java#L140)):*
> ```Java
>    * @return the value serialized as protobuf.
>    */
>   abstract byte[] build();
> ```

Maybe a `ByteString` or something less mutable?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 173 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTezrhP7V6Zrdxxlm7C:-MTezrhP7V6Zrdxxlm7D:b-carmrx) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L173)):*
> ```Java
>   public <C extends BaseValue<C>> boolean putIfAbsent(SaveOp<C> saveOp) {
>     final ColumnFamilyHandle columnFamilyHandle = getColumnFamilyHandle(saveOp.getType());
>     try (final Transaction transaction = rocksDB.beginTransaction(new WriteOptions())) {
> ```

I don't know a lot about rocks but do we have to create the `WriteOptions` for every transaction? Can it be cached?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 184 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTezkwf88SU5pRv7jQs:-MTezkwf88SU5pRv7jQt:b-3ner8g) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L184)):*
> ```Java
>       return false;
>     } catch (RocksDBException e) {
>       throw new RuntimeException(e);
> ```

possible to give any more info on this exception?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 209 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-5uV6iuZ_eRoXSCf:-MTf-5uV6iuZ_eRoXSCg:b-g942a6) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L209)):*
> ```Java
>     try (final Transaction transaction = rocksDB.beginTransaction(new WriteOptions())) {
>       if (!isConditionExpressionValid(transaction, columnFamilyHandle, id, type, condition, "delete")) {
>         LOGGER.debug("Condition failed during delete operation.");
> ```

any more info on this debug log possible? eg what failed to be deleted or what condition was missed etc
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 245 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-JgA8nDBX_fVVFmn:-MTf-JgBBZdAk9oekDFg:bjh9ayd) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L245)):*
> ```Java
>       RocksSerDe.deserializeToConsumer(type, buffer, consumer);
>     } catch (RocksDBException e) {
>       throw new RuntimeException(e);
> ```

Similar to other comments (apologies for teh repetition). Can we get more info on these exceptions
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 289 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-dWw3_RKI5kgJEJ-:-MTf-dWxEaYY6as_Z9nc:b-68424g) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L289)):*
> ```Java
>    */
>   @VisibleForTesting
>   void deleteAllData() {
> ```

this makes me mildly nervous. Is it possible to move the 'destroy the entire db' logic into a test scope? Maybe only expose the rocks connection or instantiate it in the test?

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 21 of 30 files reviewed, 25 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/mongodb/src/main/java/org/projectnessie/versioned/mongodb/MongoBaseValue.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZdtaDft3_ktk_Ap1:-MTh5vYa4AYW-ZEY7gYB:bwx9irs) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/mongodb/src/main/java/org/projectnessie/versioned/mongodb/MongoBaseValue.java#L57)):*

This is an unrelated change.
</blockquote></details>

I reverted this back to main.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeYYInE4tDmznoe8g7:-MThBExs7IupoKLvoqmD:bpzkgqv) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L118)):*

What does a `flush` do exactly? Is it mandatory to make (unflushed) changes persistent?
</blockquote></details>

It flushed all memory table data. Data is held in memory and not persisted until the memory table becomes full. This code was actually taken from the Dremio RocksDBStore code.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 173 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTezrhP7V6Zrdxxlm7C:-MThG8ENBojL3eZfOGey:b3yv6a6) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L173)):*

I don't know a lot about rocks but do we have to create the `WriteOptions` for every transaction? Can it be cached?
</blockquote></details>

Yes, I just made a static final and use that everywhere except on close(), which uses a non vanilla option.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 184 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTezkwf88SU5pRv7jQs:-MThKKji6e-tAMIW1xr1:brdrxah) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L184)):*

possible to give any more info on this exception?
</blockquote></details>

This is catching 3 possible rocksdb methods that may fail. I added a message to at least say which operation is failing. Not sure if this is sufficient?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 209 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-5uV6iuZ_eRoXSCf:-MThT3e721I31ZxOkajz:bwo7nhr) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L209)):*

any more info on this debug log possible? eg what failed to be deleted or what condition was missed etc
</blockquote></details>

I added relevant type and id to the exception message.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 245 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-JgA8nDBX_fVVFmn:-MThTCpXA6lxGH4bUoec:br81by2) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L245)):*

Similar to other comments (apologies for teh repetition). Can we get more info on these exceptions
</blockquote></details>

I added type and id to the exception message.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 21 of 30 files reviewed, 25 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 261 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_TgW84pnzn2Xe0AP:-MThbDvbBiNqoZmq5MeN:b-aa6ofe) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L261)):*

Nit: rename to rocksIter or something.
</blockquote></details>

Updated to your suggestion
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 342 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZ4f53tCvg6pd6dgJ:-MThdHOb685Id8EMY0-L:b-navklt) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L342)):*

String.format
</blockquote></details>

Added String.format
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 380 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTed5RA-FO2GAXD9WKJ:-MTiM9PXE9MS-ijX_31o:b-d85urh) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L380)):*

Nit: I'd prefer `java.nio.Path` here.
</blockquote></details>

I converted to java.nio.file.*
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java, line 26 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeXhNX5KVw19zXI9ie:-MTiPeVG8YCwGayYHoYG:b-s56ved) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java#L26)):*

Returning an empty string here as a default value is dangerous, because `new File("")` "somewhat" translates to the current directory.
</blockquote></details>

What would you suggest should be the default value?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java, line 54 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZtF2DH04q6u3dNAQ:-MTiR6k47fFTelgUfE4b:bazau13) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java#L54)):*

Add some information to the exception (e.g. `entity.toString()`)
</blockquote></details>

Added exception string.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java, line 86 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeZzBoCs3zOkXxFpz9:-MTiS3VZ9j1rP3fgY9ub:b-ndnic7) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBValueVisitor.java#L86)):*

Add some information to the exception (e.g. `entity.toString()`)
</blockquote></details>

Added value to exception message.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 90 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_jTl2mO1O-LTs_B3:-MTiSy54A3SlmKV1azsi:bkvkmke) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L90)):*

If it's something invalid, why not throw here?
</blockquote></details>

The action to take when the ConditionExpression is not met varies depending on which Store method invoked it. Currently it is one of return false or throw a ConditionFailedException. As that decision is not made within the evaluation itself I have passed back a boolean indicating the pass result. If the condition fails the calling Store method then takes the appropriate action.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_uiXCnJb3hlQJnZ6:-MTiTotG358CAOZXxK1D:b-lia07a) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L47)):*

Don't save the `Stream`s and other fields. `Stream`s that come from other store implementations can be based on currently open resources (e.g. a JDBC ResultSet), which are closed, when this object's `build()` method's called.
</blockquote></details>

Nice catch. I will refactor this.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 12 of 29 files reviewed, 25 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTec-QLA4norXesiOzv:-MTkgKCD2Dgmq40K918Y:b-7h3kwn) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java#L37)):*

Looks identical to `RocksCommitMetadata`?
</blockquote></details>

I have moved the common implementation up to RocksWrappedValue
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 38 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTee1Ka8ddkSK_V525c:-MTkjMd-CYEkR-tvvjlJ:b-tw1zrw) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L38)):*

Why the concatenation?
</blockquote></details>

It is a list of paths used in the for loop to close down open stores. It's one way to avoid duplicating code.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 39 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTedf1g9sAzvgGdwKCQ:-MTkl1aE-jhgU0Ag6dGg:bq58std) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L39)):*

`Files.exists` as mixing `java.io.File` with the `Files` API isn't nice.
</blockquote></details>

Updated to use java.nio throughout.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTecWRP3i84r_wXFwye:-MTl-45y9UcUzekyDjXy:b-i0wwzg) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L40)):*

Why map to a `File` and not use `java.nio.file.Files#delete`? Better also do this "cleanup" in a `@BeforeAll`, so the test starts with a clean state.
</blockquote></details>

Refactored to use java.nio throughout.
FYI I broke out the stream to a list as the IOException was not handled in the stream.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 450 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeJAP81qB8m9fcL3q:-MTl-wpUALAT38llmTcV:b328jwb) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L450)):*

Why this change?
</blockquote></details>

The test did not match the test method name. The test expects to execute a ConditionExpression on a Branch that should already be present in the store. Subsequently the ConditionExpression fails.

What the test did was execute a ConditionExpression on a branch that did not exist. So it would fail anyway but for the wrong reason.

I added the missing putThenLoad method so initial conditions were good.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 483 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeKkB6pXKKjM5AdMH:-MTl1MW86DI1cQ3BfKYk:b-m4uqe5) ([raw file](https://github.com/projectnessie/nessie/blob/bf8994c202a3aaa4363b8d61ca964b0926df256a/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L483)):*

Why this change?
</blockquote></details>

put() should not throw a NotFoundException if we follow Store.java API.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 9 of 29 files reviewed, 25 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf5oXb-Ov1VMFGJy2f:-MTmINu78MkHdocGJfgb:bmyfmry) (no related file):*

Few minor comments from me. I think less use of raw `byte[]` would be nice.
</blockquote></details>

This is the format that RocksDB passes data as. We could wrap the data with a ByteString but there is an overhead in SerDe. When we need to serialize ByteString back to a byte[] to write to RocksDB it would have to create a copy as ByteString is immutable. This is an overhead. The other aspect is could a race condition occur. Looking at the code, byte[] variables are local or passed as return parameters, therefore they are not shared or class attributes. So I think race conditions would not occur.
So unless you can see a particular problem I propose leaving as is.

---
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksBaseValue.java, line 140 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeydDKBC6bK8u-e_7d:-MTmEoRV4f-niTBCbBfi:b1xlshn) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksBaseValue.java#L140)):*

Maybe a `ByteString` or something less mutable?
</blockquote></details>

Please see my comment to your general concern about byte[]. I won't change this unless you can see a particular problem.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 9 of 29 files reviewed, 25 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf5oXb-Ov1VMFGJy2f:-MTojlYr0jEu6UHUy94e:bvjin60) (no related file):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

This is the format that RocksDB passes data as. We could wrap the data with a ByteString but there is an overhead in SerDe. When we need to serialize ByteString back to a byte[] to write to RocksDB it would have to create a copy as ByteString is immutable. This is an overhead. The other aspect is could a race condition occur. Looking at the code, byte[] variables are local or passed as return parameters, therefore they are not shared or class attributes. So I think race conditions would not occur.
So unless you can see a particular problem I propose leaving as is.
</blockquote></details>

Understood, thanks for the explanation. It looks weird to my eye but I take your point.

---

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 9 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeYYInE4tDmznoe8g7:-MTpMmzs9X6vfkxlJWgh:bfl22z8) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L118)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

It flushed all memory table data. Data is held in memory and not persisted until the memory table becomes full. This code was actually taken from the Dremio RocksDBStore code.
</blockquote></details>

Do we actually need to flush the memtable on close? Why not let it re-apply the commit-log upon start?
And the other way around: does it need a way to let an admin manually flush the memtable and/or other administrative options?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java, line 26 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeXhNX5KVw19zXI9ie:-MTpO37qA_Nf5S7KY31B:b-muxlq1) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java#L26)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

What would you suggest should be the default value?
</blockquote></details>

Make it mandatory, no default value?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 90 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_jTl2mO1O-LTs_B3:-MTpOWT35LiHszVE35a3:b-8rpz27) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L90)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

The action to take when the ConditionExpression is not met varies depending on which Store method invoked it. Currently it is one of return false or throw a ConditionFailedException. As that decision is not made within the evaluation itself I have passed back a boolean indicating the pass result. If the condition fails the calling Store method then takes the appropriate action.
</blockquote></details>

Can you add a comment there? A bit more than "Invalid ... function", where it actually means "not applying function X to path segment Y".
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/819#-MTpOrlL388uZYk7cL6P:-MTpOrlL388uZYk7cL6Q:b-8evi4k) ([raw file](https://github.com/projectnessie/nessie/blob/a0df308ca9652a799023a4543abad5ede2f819a4/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L89)):*
> ```Java
>             return (keys.size() == function.getValue().getNumber());
>           } else {
>             return false;
> ```

Pls remove the final `else` block and just leave the `return false` there. That's a bit clearer to read (think: case-fall-through).
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 96 at r2](https://reviewable.io/reviews/projectnessie/nessie/819#-MTpOH873cnmb7uG_1KG:-MTpOH873cnmb7uG_1KH:bn329e) ([raw file](https://github.com/projectnessie/nessie/blob/a0df308ca9652a799023a4543abad5ede2f819a4/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L96)):*
> ```Java
>       }
>     } catch (IllegalStateException e) {
>       // Catch exceptions raise due to malformed ConditionExpressions.
> ```

Similar to the above - it says `malformed ConditionExpressions`. "malformed" means something is wrong with the condition and it cannot be evaluated. I'd assume, we should fail here, as that indicates a bug, no?
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_uiXCnJb3hlQJnZ6:-MTpPcNpBUAKWPReQVHx:b-usment) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L47)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

Nice catch. I will refactor this.
</blockquote></details>

Well, I actually meant to apply those values directly to the protobuf-builder (see my comment for the `build()` method.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTec-QLA4norXesiOzv:-MTpRexu-Hnvpg9pHdag:b-70hoy) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java#L37)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

I have moved the common implementation up to RocksWrappedValue
</blockquote></details>

Actually, if `RocksValue` and `RocksCommitMetadata` don't do anything different than `RocksWrappedValue`, you can remove those two classes.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 38 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTee1Ka8ddkSK_V525c:-MTpQ8uz27wuwI8YZwNB:b-yzppll) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L38)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

It is a list of paths used in the for loop to close down open stores. It's one way to avoid duplicating code.
</blockquote></details>

But `getRawPath` is underneath `DB_PATH`, no?
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTecWRP3i84r_wXFwye:-MTpQLQ50bapCrk0IZ3p:bqwktaj) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L40)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

Refactored to use java.nio throughout.
FYI I broke out the stream to a list as the IOException was not handled in the stream.
</blockquote></details>

Nit: use a `forEach` so you don't have an intermediate `collect`.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 483 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeKkB6pXKKjM5AdMH:-MTpREaN-ZH45WxO2xC1:b293r4y) ([raw file](https://github.com/projectnessie/nessie/blob/bf8994c202a3aaa4363b8d61ca964b0926df256a/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L483)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

put() should not throw a NotFoundException if we follow Store.java API.
</blockquote></details>

It's still confusing me - mean: do the DynamoDB tests don't exercise this path in this catch-block? If they do, doesn't the Rocks store have a different behavior?

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 9 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 282 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_HOz7EByh4SHNJth:-MTqaf8S4-6-elhuBiXT:bvzrfrv) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L282)):*

This needs to handle the Stream-close to close the `RocksIterator`
</blockquote></details>

I added a close after the Stream is created as it looks like spliterator does not automatically close the stream.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 9 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java, line 26 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeXhNX5KVw19zXI9ie:-MTrE8vuE9XAV02HomFH:bfovir4) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStoreConfig.java#L26)):*

Make it mandatory, no default value?
</blockquote></details>

Updated as per your suggestion and removed default value to make it mandatory.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 90 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_jTl2mO1O-LTs_B3:-MTv-y0nAfEWghCIy9HG:b-8x7s4w) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L90)):*

Can you add a comment there? A bit more than "Invalid ... function", where it actually means "not applying function X to path segment Y".
</blockquote></details>

I changed the logic around and now throw an exception with more detail in upon failure.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 89 at r2](https://reviewable.io/reviews/projectnessie/nessie/819#-MTpOrlL388uZYk7cL6P:-MTrGhKjAZQiZfyHYN_w:bu7ma37) ([raw file](https://github.com/projectnessie/nessie/blob/a0df308ca9652a799023a4543abad5ede2f819a4/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L89)):*

Pls remove the final `else` block and just leave the `return false` there. That's a bit clearer to read (think: case-fall-through).
</blockquote></details>

I have change the logic around to allow exceptions to be thrown instead. This should address your other comments about not reporting more detail about what has failed in the ConditionExpression.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java, line 96 at r2](https://reviewable.io/reviews/projectnessie/nessie/819#-MTpOH873cnmb7uG_1KG:-MTv0B9h5xBQ367sINJO:b-d5i307) ([raw file](https://github.com/projectnessie/nessie/blob/a0df308ca9652a799023a4543abad5ede2f819a4/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksFragment.java#L96)):*

Similar to the above - it says `malformed ConditionExpressions`. "malformed" means something is wrong with the condition and it cannot be evaluated. I'd assume, we should fail here, as that indicates a bug, no?
</blockquote></details>

As above. I now throw an exception with more detail.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 7 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTec-QLA4norXesiOzv:-MTwHOcR51JjDlwBxSUk:bs5mwp2) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksValue.java#L37)):*

Actually, if `RocksValue` and `RocksCommitMetadata` don't do anything different than `RocksWrappedValue`, you can remove those two classes.
</blockquote></details>

They get used in RocksSerDe.java to build the deserializer list. They provide strong typing in mapping to the ValueType.
You can't specify:
deserializers.put(ValueType.COMMIT_METADATA, (e, c) -> RocksWrappedValue<Value>.toConsumer(e, (CommitMetadata) c));
for instance.

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 7 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 282 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_HOz7EByh4SHNJth:-MUCwNLNAF3NQ3ZQfM6z:b-5eanh) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L282)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

I added a close after the Stream is created as it looks like spliterator does not automatically close the stream.
</blockquote></details>

I think, you have to implement the `Stream`'s `onClose`, no?

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 7 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 289 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTf-dWw3_RKI5kgJEJ-:-MUK-W6MCWsl2aDoZbwK:b-miaeah) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L289)):*

this makes me mildly nervous. Is it possible to move the 'destroy the entire db' logic into a test scope? Maybe only expose the rocks connection or instantiate it in the test? 
</blockquote></details>

I extended RocksDBStore in TestRocksDBStore and moved the deleteAllData() method there.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 47 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_uiXCnJb3hlQJnZ6:-MUK1Lnq8UXvKhXOOW2m:bb6un1j) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L47)):*

Well, I actually meant to apply those values directly to the protobuf-builder (see my comment for the `build()` method.
</blockquote></details>

Done
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java, line 140 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeajAK1SuahBO1T942:-MUK1RpIDo--ot8mEKVz:bb6un1j) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksL1.java#L140)):*

All `build()` methods should basically only do some post-validation and just the `builder.build().toByteArray()`.
The actual additions (like `builder.addAllTree(buildIds(tree))`) should happen in the "consuming methods (like `children(Stream<Id>`).
</blockquote></details>

Done
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 483 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeKkB6pXKKjM5AdMH:-MUK2lzX5phvOWikPizA:b-r0i1a2) ([raw file](https://github.com/projectnessie/nessie/blob/bf8994c202a3aaa4363b8d61ca964b0926df256a/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L483)):*

It's still confusing me - mean: do the DynamoDB tests don't exercise this path in this catch-block? If they do, doesn't the Rocks store have a different behavior?
</blockquote></details>

I'll update the put() implementation to throw a NotFoundException and revert this change.

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: 7 of 29 files reviewed, 19 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 118 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeYYInE4tDmznoe8g7:-MUUo6gUDkLl__mQEiY_:bdox2o6) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L118)):*

Do we actually need to flush the memtable on close? Why not let it re-apply the commit-log upon start?
And the other way around: does it need a way to let an admin manually flush the memtable and/or other administrative options?
</blockquote></details>

Flushing on close has now been removed.
___
*[versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java, line 282 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTe_HOz7EByh4SHNJth:-MUUoVkW9WyhIiZcb_zq:bb6un1j) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/main/java/org/projectnessie/versioned/rocksdb/RocksDBStore.java#L282)):*

I think, you have to implement the `Stream`'s `onClose`, no?
</blockquote></details>

Done
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 38 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTee1Ka8ddkSK_V525c:-MUUpDCREqZmsaCbZZf5:bhidid3) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L38)):*

But `getRawPath` is underneath `DB_PATH`, no?
</blockquote></details>

Now just uses DB_PATH.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTecWRP3i84r_wXFwye:-MUUtgeg5ggWQYpoNtzH:b-hasl7x) ([raw file](https://github.com/projectnessie/nessie/blob/7e1b772c5b0ff731ce5b6433173258e625fca805/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestRocksDBStore.java#L40)):*

Nit: use a `forEach` so you don't have an intermediate `collect`.
</blockquote></details>

I didn't see a way to do this and handle the IOException from Files.delete() in the Stream.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 483 at r1](https://reviewable.io/reviews/projectnessie/nessie/819#-MTeeKkB6pXKKjM5AdMH:-MUUuEQK5LYDxWicGiSV:brkntft) ([raw file](https://github.com/projectnessie/nessie/blob/bf8994c202a3aaa4363b8d61ca964b0926df256a/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L483)):*
<details><summary><i>Previously, stevelorddremio (Steve Lord) wrote…</i></summary><blockquote>

I'll update the put() implementation to throw a NotFoundException and revert this change.
</blockquote></details>

We didn't a bit more refactoring of this.

rymurr (commented):
> Reviewed 5 of 30 files at r1, 2 of 13 files at r2, 3 of 11 files at r4, 1 of 3 files at r8, 11 of 11 files at r9.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: all files reviewed, 22 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 50 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhFcKk0w1IgF5nJC-O:-MUhFcKk0w1IgF5nJC-P:bkb0g5o) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L50)):*
> ```Java
> 
>   protected static long getRandomSeed() {
>     return -2938423452345L;
> ```

why a different seed from above?
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 53 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhF_kx3sIQ4-d2wNhq:-MUhF_kx3sIQ4-d2wNhr:b-e9hxwz) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L53)):*
> ```Java
>   }
> 
>   static final Id ID = createId(new Random(getRandomSeed()));
> ```

why not use the Random from above?
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 675 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhGLsUEPJXEV8rDp4b:-MUhGLsUEPJXEV8rDp4c:bpuj7st) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L675)):*
> ```Java
>    * @param random object to use for randomization of entity creation.
>    * @return sample ID entity.
>    */
> ```

a lot of these static exist in `SampleEntities`. Can you remove the duplicate code?

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: all files reviewed, 13 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/819#-MUiOo9IAn7WjENUYieD:-MUiOo9IAn7WjENUYieE:b-rrw1wo) (no related file):*
Looks like `RocksDBStore.update()` is not implemented, so the rocksdb store implementation in the current state is incomplete and Nessie commits won't work. Also missing the integration tests against `AbstractITTieredVersionStore`and `AbstractITVersionStore`.

---

stevelorddremio (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/819)* status: all files reviewed, 13 unresolved discussions (waiting on @rymurr, @snazy, and @stevelorddremio)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/819#-MUiOo9IAn7WjENUYieD:-MUiYzMFAHJjh0BKSJTq:btp841) (no related file):*

Looks like `RocksDBStore.update()` is not implemented, so the rocksdb store implementation in the current state is incomplete and Nessie commits won't work. Also missing the integration tests against `AbstractITTieredVersionStore`and `AbstractITVersionStore`.
</blockquote></details>

Correct `RocksDBStore.update()` is not implemented in this PR. That will be a later PR.
Integration tests are in the extended test classes;
`TestRocksDBTieredVersionStore` and `TestRocksDBVersionStore`, respectively. Although without `update()` it is not expected that they will pass.

---
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 50 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhFcKk0w1IgF5nJC-O:-MUibaiG7ydvGGFdTQ0k:b48hrk9) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L50)):*

why a different seed from above?
</blockquote></details>

Good point. They now use the same seed.
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 53 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhF_kx3sIQ4-d2wNhq:-MUibjsV383-TZ34ED_T:bb6un1j) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L53)):*

why not use the Random from above?
</blockquote></details>

Done
___
*[versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java, line 675 at r9](https://reviewable.io/reviews/projectnessie/nessie/819#-MUhGLsUEPJXEV8rDp4b:-MUihka30bux5IwThv53:b-ripxvx) ([raw file](https://github.com/projectnessie/nessie/blob/e9bb7506cbc10783194ac0744f9c7d85c89ccdd5/versioned/tiered/rocksdb/src/test/java/org/projectnessie/versioned/rocksdb/TestConditionExecutor.java#L675)):*

a lot of these static exist in `SampleEntities`. Can you remove the duplicate code?
</blockquote></details>

Well spotted. Yes, removed.

snazy:
#1638 has a working RocksDB implementation

Issue #813, opened 2021-02-12
Concurrent commits against same branch fail


snazy:
There's a race condition somewhere, that lets concurrent updates (commits) against the same branch fail.

Some log messages (added those messages): Two concurrent commits - one of these saves the L1, the other one tries to load that L1, but fails, because that L1 is not yet there.

It feels, that the ID of the L1 to be saved is referenced elsewhere. Just letting `TieredVersionStore.ensureValidL1` wait until `ensureAvailable()` has finished doesn't help (that actually increases the number of errors returned to the client).

Generally (not strictly related to this issue), we can save some effort by allowing only one run of `InternalBranch.collapseIntentionLog` at any time per Nessie-Server. This would reduce the number of failed attempts/iteration and gets/puts against the backend database. This would not fix this issue though, it would just reduce the probability to hit this issue. Such an approach would definitely reduce the number of async started theres to run `collapseIntentionLog` though.

```
2021-02-12 18:55:50,563 bear nessie-quarkus-0.3.1-SNAPSHOT-runner.jar[123203] INFO  [io.qua.htt.access-log] (executor-thread-21) 127.0.0.1 - - 12/Feb/2021:18:55:50 +0100 "POST /api/v1/contents/name.space.someTable_7_20?branch=main&hash=5d1e09785da83245caf91d53c8d8c5b753ef21ca&message=test%20commit%207 HTTP/1.1" 204 -

2021-02-12 18:55:50,625 bear nessie-quarkus-0.3.1-SNAPSHOT-runner.jar[123203] INFO  [io.qua.htt.access-log] (executor-thread-20) 127.0.0.1 - - 12/Feb/2021:18:55:50 +0100 "POST /api/v1/contents/name.space.someTable_8_3?branch=main&hash=5d1e09785da83245caf91d53c8d8c5b753ef21ca&message=test%20commit%208 HTTP/1.1" 204 -

[...]

PR #820, opened 2021-02-15
Fix for #813

jacques-n:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/820)


jacques-n:
@snazy I believe this will fix the issue. Do you have a unit test that shows the problem?

snazy (approved):
LGTM! Thanks for the patch!

PR #825, opened 2021-02-16
Handle errors during (Version)Store initialization in Quarkus

snazy:
When the a (Version)Store fails to start (network broken, local Dynamo-instance not started, etc), the current implementation implicitly retries the (Version)Store initialization for every incoming request. Since a failed start can take very long, this implicitly results in many requests piling up and fail sequentially (as Quarkus synchronizes the bean initialization and eventually the call to `DynamoStore.start()`).

The proposed implementation is quite "dumb": if a previous (Version)Store initialization failed, all following initialization attempts within 5 seconds will just fail as well. This gives Quarkus some room to "abandon" the requests that have piled up in the meantime so those do not force a likely-to-fail (Version)Store initialization as well, sequentially.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/825)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java
@@ -130,7 +130,7 @@ public class VersionStoreFactory {
               .build();
           break;
         default:
-          throw new RuntimeException(String.format("unknown jgit repo type %s",
+          throw new RuntimeException(String.format("unknown version-store type %s",
               config.getVersionStoreConfig().getVersionStoreType()));
       }
       lastUnsuccessfulStart = 0L;
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java
@@ -20,6 +20,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 import javax.enterprise.inject.Produces;
@@ -72,6 +73,8 @@ public class VersionStoreFactory {
   @ConfigProperty(name = "quarkus.dynamodb.endpoint-override")
   Optional<String> endpoint;
 
+  private static final long START_RETRY_MIN_INTERVAL_NANOS = TimeUnit.SECONDS.toNanos(2);
+  private volatile long lastUnsuccessfulStart = 0L;
 
   @Produces
   public StoreWorker<Contents, CommitMeta> worker() {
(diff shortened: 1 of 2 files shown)


rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/825)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @snazy)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeX_z36-vlt7O6dDtG:-MTeX_z4EdllTdtqayuw:b5cjg6) (no related file):*
dumb question: Why do we want the server to retry initialisation if there is an issue w/ the backend? As a cloud native/horizontally scalable solution don't we want to just crash and let the managing system deal w/ it. Eg a lambda who can't talk to dynamo should just die and return 500 errors. If I understand this is to keep a single isolated instance up under high load and severe backend faults right?

---
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java, line 133 at r1](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeYZZJ8XVL3SZB29lO:-MTeYZZK7AXWzmh02V5H:bt5shro) ([raw file](https://github.com/projectnessie/nessie/blob/669ded24408e3478c10016d919f52dd8d9b7be2d/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java#L133)):*
> ```Java
>           break;
>         default:
>           throw new RuntimeException(String.format("unknown jgit repo type %s",
> ```

i know this wasn't you (it was likely me!) but I think this log message is incorrect
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 180 at r1](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeYd1P3hInS5X9hOqJ:-MTeYd1P3hInS5X9hOqK:busw28u) ([raw file](https://github.com/projectnessie/nessie/blob/669ded24408e3478c10016d919f52dd8d9b7be2d/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L180)):*
> ```Java
>         EntityStoreHelper.storeMinimumEntities(this::putIfAbsent);
>       }
>     } catch (Exception ex) {
> ```

What other exception types are we expecting? I personally prefer narrow exception scopes

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/825)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @rymurr)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeX_z36-vlt7O6dDtG:-MTeiw3S2uGRZmhYr6p-:b7imkeo) (no related file):*

dumb question: Why do we want the server to retry initialisation if there is an issue w/ the backend? As a cloud native/horizontally scalable solution don't we want to just crash and let the managing system deal w/ it. Eg a lambda who can't talk to dynamo should just die and return 500 errors. If I understand this is to keep a single isolated instance up under high load and severe backend faults right?
</blockquote></details>

That's actually how Quarkus works / how it's implemented. For a lambda, fail-fast is fine (well, it wouldn't be "fast", because it takes a few seconds until the error-exception is thrown). For a long-running server probably not.

We should maybe change the initialization of the version-store to be eager and not lazy as it's currently implemented. But that's a different change.

---
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java, line 133 at r1](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeYZZJ8XVL3SZB29lO:-MTejeCbE_g3japAptZL:b-tkxpnj) ([raw file](https://github.com/projectnessie/nessie/blob/669ded24408e3478c10016d919f52dd8d9b7be2d/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/VersionStoreFactory.java#L133)):*

i know this wasn't you (it was likely me!) but I think this log message is incorrect
</blockquote></details>

Changed to `Unknown Git repo type`. (Kidding ;) )
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 180 at r1](https://reviewable.io/reviews/projectnessie/nessie/825#-MTeYd1P3hInS5X9hOqJ:-MTejk-oDzC2gLpOHGKi:boer2ty) ([raw file](https://github.com/projectnessie/nessie/blob/669ded24408e3478c10016d919f52dd8d9b7be2d/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L180)):*

What other exception types are we expecting? I personally prefer narrow exception scopes
</blockquote></details>

A lot probably, as a lot can go wrong in that try-block. I've changed it to ensure that it will always call `close()` to clean-up potentially allocated resources.

rymurr (approved):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/825)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rymurr)

PR #826, opened 2021-02-16
Report proper exception message, telling the user what actually happened

snazy:
`BaseResource.doOps` unconditionally returns a `new NessieConflictException("Failed to commit data. Provided hash does not match current value.", e);` to the user, even if the root-cause has a meaningful error message. In case of "overloading" the system with commits (i.e. many clients committing to the same branch concurrently), the `TieredVersionStore` throws a `ReferenceConflictException` saying `Unable to complete commit due to conflicting events. Retried %d times before failing.`. The previously hard-coded message implies a user error, where the "right" message indicates an "overload-situation" - so two distinct things.

I.e. the user sees `Provided hash does not match current value.` even if the user did not provide a hash, but just a branch name - the user should see `Unable to complete commit due to conflicting events. Retried %d times before failing.`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/826)

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
@@ -84,7 +84,7 @@ abstract class BaseResource {
     } catch (IllegalArgumentException e) {
       throw new NessieNotFoundException("Invalid hash provided.", e);
     } catch (ReferenceConflictException e) {
-      throw new NessieConflictException("Failed to commit data. Provided hash does not match current value.", e);
+      throw new NessieConflictException("Failed to commit data. " + e.getMessage(), e);
     } catch (ReferenceNotFoundException e) {
       throw new NessieNotFoundException("Failed to commit data. Provided ref was not found.", e);
     }
--- a/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
@@ -82,11 +82,11 @@ abstract class BaseResource {
           operations
       );
     } catch (IllegalArgumentException e) {
-      throw new NessieNotFoundException("Invalid hash provided.", e);
+      throw new NessieNotFoundException("Invalid hash provided. " + e.getMessage(), e);
     } catch (ReferenceConflictException e) {
       throw new NessieConflictException("Failed to commit data. " + e.getMessage(), e);
     } catch (ReferenceNotFoundException e) {
-      throw new NessieNotFoundException("Failed to commit data. Provided ref was not found.", e);
+      throw new NessieNotFoundException("Failed to commit data. " + e.getMessage(), e);
     }
   }
 


servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -84,7 +84,7 @@ protected void doOps(String branch, String hash, String message, List<org.projec
     } catch (IllegalArgumentException e) {
       throw new NessieNotFoundException("Invalid hash provided.", e);
     } catch (ReferenceConflictException e) {
-      throw new NessieConflictException("Failed to commit data. Provided hash does not match current value.", e);
+      throw new NessieConflictException("Failed to commit data. " + e.getMessage(), e);
```
rymurr, line 87: Can we have a richer error? We've gone from 'guessing and wrong in some cases' to hoping that the underlying exception has a useful value. Eg 'Usually this means the provided hash does not match current value, {e.getMessage()} may provide more context'

snazy (commented):
All instantiations/throw sites of `ReferenceConflictException` include a proper message. It was just wrong in `BaseResource` to hide that message from the user. Since the message was actually also hidden for the two other cases in the re-throws in `BaseResource`, I've changed those as well, because: why hide probably useful information from the user.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/826#-:-MTj841C-4NJ-cy3OS7u:b-ktfkxo)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @snazy)

rymurr:
> All instantiations/throw sites of `ReferenceConflictException` include a proper message. It was just wrong in `BaseResource` to hide that message from the user. Since the message was actually also hidden for the two other cases in the re-throws in `BaseResource`, I've changed those as well, because: why hide probably useful information from the user.
> 
> > _[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/826#-:-MTj841C-4NJ-cy3OS7u:b-ktfkxo)_ status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @snazy)

Apologies for being thick, this still looks weird to me. We are assuming that all stores have populated meaningful exceptions. We are also copying the message that would normally be visible in the 'Caused by' part of the stack trace. Are we somehow eating the 'Caused by' section?

snazy (commented):
Yes, it's not present in the "top-level exception", so hidden when e.g. running perf-tests.

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/826#-:-MTjGEqq6E5WQY09Mi3C:biinzq4)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @snazy)

rymurr (approved):

jacques-n:
I'm not a fan of passing unstructured internal error messages back to the user. Especially for something generic like Illegal argument exception. If we want more nuance in the messaging, I think we should introduce subclasses of certain specific exceptions and then be selective about exposing the details.

PR #832, opened 2021-02-17
Fix config application.properties to match name in ApplicationConfig.java

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/832)


rymurr (approved):

snazy:
Ugh - that build failure is strange.
So, the annotation reads `@ConfigProperty(name = "send-stacktrace-to-client"`, but the base(!) interface method's named `shouldSendStackTraceToAPIClient()`, which seems to confuse Quarkus.

rymurr (approved):

PR #834, opened 2021-02-17
Update commit metadata in transplant/merge

### LINKED ISSUE #831: re-set commit meta on transplant/merge
Currently #818 only addresses adding properties to commit metadata. It does not correctly set them when rebasing or merging. This ticket is for the follow up task of setting commit meta on transplant and merge.

Specifically:
1) committer should change to current principal
2) commit time should change to current time
3) author should stay the same or be the original committer (if null)
4) author time should stay the same or be the original commit time (if null)
5) perhaps add an item in properties to record more info about the merge/transplant (eg original branch, was it a merge or transplant etc)

rymurr:
Closes #831

Needs #818 merged and a rebase first

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/834)


rymurr:
The thing I don't like about this is we are passing down opaque functions into the VersionStore to manipulate our generic values. This is going to happen again while working on #780 too. Maybe our VersionStore generics need more speialization?

jacques-n:
I think we should start separating system metadata from user data. Author and commit times feel like system metadata and I'd be inclined to move them to the generic type. I also think the same is true of entity type. I have before proposed that entity type be a single byte value where the versionstore is unaware of the meaning but is additional component of things.

rymurr:
> I think we should start separating system metadata from user data. Author and commit times feel like system metadata and I'd be inclined to move them to the generic type.

Just to be clear, do you mean giving `METADATA` template an `extends` clause and making a few things system metadata? I prefer that approach but didn't think many people would like such a thing. That would simplify this change and make me happeir.

jacques-n:
> Just to be clear, do you mean giving `METADATA` template an `extends` clause and making a few things system metadata? I prefer that approach but didn't think many people would like such a thing. That would simplify this change and make me happeir.

Yes, that is what I am proposing. We should probably also make a decision on whether those properties can be set by consumers of versionstore. I would expect they cannot (expected empty when provided to api, populated by versionstore impl and then returned when read).

rymurr:
> I have before proposed that entity type be a single byte value where the versionstore is unaware of the meaning but is additional component of things.

I am not sure how to support this cleanly. Does the `StoreWorker` control the mapping? do we specialize the `DATA` parameter in `VersionStore` to supply a `byte` type?

In both this case and the statement above I am personally for making the `VersionStore` less generic by putting restrictions on `DATA` and `METADATA`.  I think passing 0 information to the VersionStore about what its storing make is annoying to do a number of things. 

cc @laurentgo

rymurr:
> Yes, that is what I am proposing. We should probably also make a decision on whether those properties can be set by consumers of versionstore. I would expect they cannot (expected empty when provided to api, populated by versionstore impl and then returned when read).

:100:% agree

rymurr:
closing for now until we figure out more what we want from this

Issue #836, opened 2021-02-17
Add entity-type-id to TieredVersionStore's KeyList


rymurr:
Entity type filtering and display has been added to the REST API and non-Tiered version stores. The `TieredVersionStore` should modify its structure to store the entity type along with the key.

PR #837, opened 2021-02-17
Add filtering by entity type in `getEntities`

### LINKED ISSUE #780: Allow entries API to filter by entity type

rymurr:
This requires the following:
1) add query param to getEntities for which types we want (empty list means all)
2) version store now returns a tuple of Key, entity type
3) in memory and jgit store and populate types on return. Tiered store does not and requires extra lookups

For follow-up PR:
* possibly push filtering into version store. Can potentially be more efficient depending on store impl
* add an entity-type-id to the key list in tiered version store to make it more efficient (#836)

Closes #780

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/837)

### DIFF
--- a/model/src/main/java/org/projectnessie/api/TreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/TreeApi.java
@@ -164,7 +164,10 @@ public interface TreeApi {
           Integer maxRecords,
       @Parameter(description = "pagination continuation token, as returned in the previous EntriesResponse.token")
       @QueryParam("pageToken")
-          String pageToken)
+          String pageToken,
+      @Parameter(description = "list of value types to return. Return all if empty")
+      @QueryParam("types")
+        List<String> types)
           throws NessieNotFoundException;
 
   /**
--- a/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
@@ -17,8 +17,10 @@
 package org.projectnessie.services.rest;
 
 import java.security.Principal;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -202,7 +204,15 @@ public class TreeResource extends BaseResource implements TreeApi {
   }
 
   @Override
-  public EntriesResponse getEntries(String refName, Integer maxEntriesHint, String pageToken) throws NessieNotFoundException {
+  public EntriesResponse getEntries(String refName, Integer maxEntriesHint, String pageToken, List<String> types)
+      throws NessieNotFoundException {
+    final Set<Type> payloads;
+    if (types.isEmpty()) {
+      payloads = Arrays.stream(Type.values()).collect(Collectors.toSet());
+    } else {
+      payloads = types.stream().map(Type::valueOf).collect(Collectors.toSet());
+    }
+
     final Hash hash = getHashOrThrow(refName);
     // TODO Implement paging. At the moment, we do not expect that many keys/entries to be returned.
     //  So the size of the whole result is probably reasonable and unlikely to "kill" either the
@@ -210,8 +220,10 @@ public class TreeResource extends BaseResource implements TreeApi {
     //  whether we shall just do the whole computation for a specific hash for every page or have
     //  a more sophisticated approach, potentially with support from the (tiered-)version-store.
     try {
+      // note currently we are filtering types at the REST level. This could in theory be pushed down to the store though
+      // all existing VersionStore implementations have to read all keys anyways so we don't get much
       List<EntriesResponse.Entry> entries;
-      try (Stream<EntriesResponse.Entry> s = getStore().getKeys(hash)
+      try (Stream<EntriesResponse.Entry> s = getStore().getKeys(hash).filter(x -> payloads.contains(x.getType()))
           .map(key -> EntriesResponse.Entry.builder().name(fromKey(key.getValue())).type((Type) key.getType()).build())) {
         entries = s.collect(ImmutableList.toImmutableList());
       }
--- a/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
@@ -219,9 +219,9 @@ public class TreeResource extends BaseResource implements TreeApi {
     //  server or client. We have to figure out _how_ to implement paging for keys/entries, i.e.
     //  whether we shall just do the whole computation for a specific hash for every page or have
     //  a more sophisticated approach, potentially with support from the (tiered-)version-store.
+    //  note currently we are filtering types at the REST level. This could in theory be pushed down to the store though
+    //  all existing VersionStore implementations have to read all keys anyways so we don't get much
     try {
-      // note currently we are filtering types at the REST level. This could in theory be pushed down to the store though
-      // all existing VersionStore implementations have to read all keys anyways so we don't get much
       List<EntriesResponse.Entry> entries;
       try (Stream<EntriesResponse.Entry> s = getStore().getKeys(hash).filter(x -> payloads.contains(x.getType()))
           .map(key -> EntriesResponse.Entry.builder().name(fromKey(key.getValue())).type((Type) key.getType()).build())) {
--- a/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
@@ -23,6 +23,7 @@ import javax.validation.constraints.NotNull;
 
 import org.projectnessie.api.TreeApi;
 import org.projectnessie.client.http.HttpClient;
+import org.projectnessie.client.http.HttpRequest;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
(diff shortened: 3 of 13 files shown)


jacques-n:
I'm uncomfortable introducing this api before we've implemented a good way to support it in tiered version store. 

By solving the storage side before exposing at the api level we know that we aren't exposing things in an awkward or inefficient way at the higher levels.

rymurr:
> I'm uncomfortable introducing this api before we've implemented a good way to support it in tiered version store.

Happy either way, I think the discussion on #834 makes it easier to do this.

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -126,8 +126,11 @@ public EntriesResponse getEntries(
       @Pattern(regexp = Validation.REF_NAME_OR_HASH_REGEX, message = Validation.REF_NAME_OR_HASH_MESSAGE)
       @Parameter(description = "name of ref to fetch from")
       @PathParam("ref")
-          String refName)
-          throws NessieNotFoundException;
+          String refName,
+      @Parameter(description = "list of value types to return. Return all if empty")
```
laurentgo, line 130: It feels like a more generic `filter` query parameter might allow for extensibility

servers/services/pom.xml
```diff
@@ -60,5 +60,11 @@
       <groupId>jakarta.enterprise</groupId>
       <artifactId>jakarta.enterprise.cdi-api</artifactId>
     </dependency>
+      <dependency>
```
laurentgo, line 63: I'm not sure why this dependency is being added as I don't see a code change using it, `datanucleus` which is something for DB persistence should not be required

rymurr: IDE just being dumb and helpfully adding deps for me. Will remove it.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -188,18 +189,44 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge) thro
...
     } catch (ReferenceNotFoundException e) {
       throw new NessieNotFoundException(String.format("Unable to find the reference [%s].", refName), e);
     }
   }
 
+
+  private static Type toType(String valueType) {
+    switch (valueType) {
+      case "IcebergTable":
+      case "ImmutableIcebergTable":
```
laurentgo, line 211: That seems wrong that 2 different value types point to the same Type object

model/src/main/java/org/projectnessie/api/TreeApi.java
rymurr: How do you mean? Just rename the parameter or some other object type?

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java
```diff
@@ -435,7 +436,7 @@ public void assign(NamedRef namedRef, Optional<Hash> currentTarget, Hash targetH
   }
 
   @Override
-  public Stream<Key> getKeys(Ref ref) throws ReferenceNotFoundException {
+  public Stream<WithEntityType<Key>> getKeys(Ref ref) throws ReferenceNotFoundException {
```
laurentgo, line 439: Wouldn't this change back once the filter is pushed down?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: agreed, I think based on conversations on #834 we will likely change this

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java
rymurr: no, as the type is still pushed to the client

model/src/main/java/org/projectnessie/api/TreeApi.java
laurentgo: some object type

rymurr: Sorry @laurentgo, its still not clear to me what you are asking for.

rymurr:
This builds on #866 to add filtering to the REST API. Should be ready to go once that is merged

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/837)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[python/pynessie/_endpoints.py, line 149 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2ANgABHVQ3L9SlatS:-MX2ANgABHVQ3L9SlatT:b-v2s8pj) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/python/pynessie/_endpoints.py#L149)):*
> ```Python
>     max_result_hint: Optional[int] = None,
>     page_token: Optional[str] = None,
>     entity_types: Optional[list] = None,
> ```

Mind renaming it to `types` to make the name consistent w/ `TreeApi`?
___
*[python/pynessie/_endpoints.py, line 162 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2A7A4Ex-b9J0EPnln:-MX2A7A4Ex-b9J0EPnlo:b-lcgdn7) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/python/pynessie/_endpoints.py#L162)):*
> ```Python
>     :return: json list of Nessie table names
>     """
>     if not entity_types:
> ```

I think, we can entirely omit the `types` parameter, when it's not present/empty.
___
*[servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java, line 281 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2BsRgDX8yhtwdHrAx:-MX2BsRgDX8yhtwdHrAy:b4irxib) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java#L281)):*
> ```Java
> 
>     entries = tree.getEntries(branch, null, null, ImmutableList.of(Contents.Type.VIEW.name())).getEntries();
>     assertEquals(1, entries.size());
> ```

Nit: These two asserts can be simplified to `assertEquals(Collections.singletonList(expected.get(1)), entries)`
___
*[servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java, line 223 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2Aq9v3sLJZ_uQ9oMv:-MX2Aq9v3sLJZ_uQ9oMw:b-vot2ol) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java#L223)):*
> ```Java
>     //  a more sophisticated approach, potentially with support from the (tiered-)version-store.
>     try {
>       // note currently we are filtering types at the REST level. This could in theory be pushed down to the store though
> ```

Can you move that to the "big TODO" above, so we don't forget about it?

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/837)* status: 0 of 13 files reviewed, 9 unresolved discussions (waiting on @laurentgo and @rymurr)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2CcM7BIKXU7kp3Lbz:-MX2CcM7BIKXU7kp3Lc-:b-og9arw) (no related file):*
LGTM, few minor comments

---

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/837)* status: 0 of 13 files reviewed, 8 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[python/pynessie/_endpoints.py, line 149 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2ANgABHVQ3L9SlatS:-MX2Dp6y7fe9UwoRSoHl:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/python/pynessie/_endpoints.py#L149)):*

Mind renaming it to `types` to make the name consistent w/ `TreeApi`?
</blockquote></details>

Done.
___
*[python/pynessie/_endpoints.py, line 162 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2A7A4Ex-b9J0EPnln:-MX2DpuoCYTHHBgIGxH_:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/python/pynessie/_endpoints.py#L162)):*

I think, we can entirely omit the `types` parameter, when it's not present/empty.
</blockquote></details>

Done.
___
*[servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java, line 223 at r2](https://reviewable.io/reviews/projectnessie/nessie/837#-MX2Aq9v3sLJZ_uQ9oMv:-MX2Dr2p7fn_7-AIebPx:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/2c5af2d78304772c394a78adb41f82bcec99653e/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java#L223)):*

Can you move that to the "big TODO" above, so we don't forget about it?
</blockquote></details>

Done.

snazy (approved):

PR #843, opened 2021-02-18
Better modularity in GC

rymurr:
* remove StoreWorker and associated Asset Key stuff moving to gc
* separate the gc into two stages: find referenced state of values and convert to unreferenced asset keys
* restrict use of `System.currentTimeMillis()` and favour absolute times over relative

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)


rymurr:
There is potentially an argument to create a new module for the non-tiered part of the gc algo. Thoughts anyone?

rymurr:
> There is potentially an argument to create a new module for the non-tiered part of the gc algo. Thoughts anyone?

this is done.

laurentgo (changes requested):
> Reviewed 23 of 32 files at r1, 8 of 14 files at r2, 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: all files reviewed, 3 unresolved discussions (waiting on @jacques-n and @rymurr)

___
*[versioned/gc/core/pom.xml, line 52 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9F--CBiX5oGo0zOA:-MTv9F--CBiX5oGo0zOB:b-1vd4o) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/pom.xml#L52)):*
> ```XML
>       <groupId>org.apache.spark</groupId>
>       <artifactId>spark-core_2.12</artifactId>
>       <version>3.0.1</version>
> ```

not the same as `${spark3.version}` ?

I also wonder if the spark dependencies should be added into dependencyManagement in the parent module `nessie-versioned-gc` so that all gc modules are on the same spark version with the same exclusions?
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java, line 20 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9gelEo7cXLaSskNm:-MTv9gelEo7cXLaSskNn:b-9z8x8u) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java#L20)):*
> ```Java
> import java.util.stream.Stream;
> 
> @FunctionalInterface
> ```

Javadoc :)

why not extending `Function<T, Stream<A>>`
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java, line 21 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9miFCibmL5gbC9u-:-MTv9miFCibmL5gbC9u0:b5rbyi0) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java#L21)):*
> ```Java
> 
> @FunctionalInterface
> public interface AssetKeyConverter<T> {
> ```

Why not simply `Stream<AssetKey>` or `Stream<A>` where `A extends AssetKey`?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: 33 of 41 files reviewed, 3 unresolved discussions (waiting on @jacques-n and @laurentgo)

___
*[versioned/gc/core/pom.xml, line 52 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9F--CBiX5oGo0zOA:-MTvIf2P3jCChHzkeSvr:bulk3jf) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/pom.xml#L52)):*

not the same as `${spark3.version}` ?

I also wonder if the spark dependencies should be added into dependencyManagement in the parent module `nessie-versioned-gc` so that all gc modules are on the same spark version with the same exclusions?
</blockquote></details>

Done. put them up into the main pom
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java, line 20 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9gelEo7cXLaSskNm:-MTvIjjL3lW36ktfurNV:bwfx9kj) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java#L20)):*

Javadoc :)

why not extending `Function<T, Stream<A>>`
</blockquote></details>

agreed, done
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java, line 21 at r3](https://reviewable.io/reviews/projectnessie/nessie/843#-MTv9miFCibmL5gbC9u-:-MTvUzO39sCufQX63Ucg:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/308f981ef710c991ef981b9bad87d674101b39a6/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java#L21)):*

Why not simply `Stream<AssetKey>` or `Stream<A>` where `A extends AssetKey`?
</blockquote></details>

Done.

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Referenced state of a value, its type and its byte[] representation.
+ */
+public final class CategorizedValue implements Serializable {
+
+  private static final long serialVersionUID = -1466847843373432962L;
+
+  private boolean referenced;
+  private byte[] data;
+  private String valueType;
```
jacques-n, line 31: I'm confused by valueType here. I don't think is a ValueType so what is it? I think maybe you're talking about contents type? What is the thinking on the mapping of this string? Are you proposing canonical strings for contents at the tiered version store level?

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java
```diff
@@ -143,17 +114,21 @@ public boolean call(Row r) throws Exception {
 
     private boolean referenced;
     private byte[] data;
+    private long hashedData;
```
jacques-n, line 117: I think it is a bad idea to do pre-hashing. It seems like it is likely to substantially increase the bloomfilter false positives. Instead, I propose just using the inner contents field of bytes instead of the outer object.

versioned/gc/core/pom.xml
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
jacques-n: I could also see a situation where you put a core module that just exposes the AssetKeyConverter interface and then have the tiered core and non-tiered iceberg modules but it feels like the non-tiered core module will just be that one class, no?

versioned/tiered/gc/src/test/java/org/projectnessie/versioned/tiered/gc/ITTestIdentifyUnreferencedAssets.java
```diff
@@ -129,9 +122,14 @@ public void run() throws Exception {
         .maxAgeMicros(ONE_DAY_OLD_MICROS)
         .timeSlopMicros(ONE_HOUR_OLD_MICROS)
         .build();
-    IdentifyUnreferencedAssets<DummyValue> app = new IdentifyUnreferencedAssets<DummyValue>(helper, new DynamoSupplier(), spark, options);
-    Dataset<UnreferencedItem> items = app.identify();
-    Set<String> unreferencedItems = items.collectAsList().stream().map(UnreferencedItem::getName).collect(Collectors.toSet());
+    IdentifyUnreferencedValues<DummyValue> app = new IdentifyUnreferencedValues<>(helper, new DynamoSupplier(), spark, options);
```
jacques-n, line 125: I think it probably makes sense to decompose the tests into tests for Assets and Values separately.

versioned/tiered/gc/src/test/java/org/projectnessie/versioned/tiered/gc/ITTestIdentifyUnreferencedAssets.java
```diff
@@ -129,9 +122,14 @@ public void run() throws Exception {
...
+    Dataset<CategorizedValue> values = app.identify();
+
+    IdentifyUnreferencedAssets<DummyValue, DummyAsset> ident = new IdentifyUnreferencedAssets<>(helper.getValueSerializer(),
```
jacques-n, line 128: weird variable names here (ident) and above (app)

clients/hmsbridge/core/pom.xml
```diff
@@ -131,6 +131,10 @@
           <groupId>commons-logging</groupId>
           <artifactId>commons-logging</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.glassfish</groupId>
```
jacques-n, line 135: This pom change and the one above are weird.

jacques-n (commented):
Sorry that I forgot to do this review in reviewable... Please respond there.

versioned/gc/core/pom.xml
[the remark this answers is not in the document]
rymurr: What about the classes in this module make them tiered to you? My motivation for this was it takes values and their referenced state and turns them into unreferenced assets. I expect most version store impls will require this as a separate step. The tiered part, to me, is finding the list of values and whether they are referenced

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/AssetKeyConverter.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.gc;
+
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * convert a value into a stream of asset keys.
```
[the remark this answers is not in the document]
rymurr: done

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java
rymurr: The thinking behind this was that the identification of AssetKeys from Values is dependent on the value type: eg the iceberg listing is different from Delta. This helps easy identification of which asset key listing algo should be used for this row. The entity type concept being discussed in other PRs would eventually fill this role so in the meantime this uses the simple class name. Happy to remove it until the entity type PRs are closed if you prefer. However I did rename to entity type to make its use more clear.

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java
rymurr: I dont love the pre-hashing either. I am not sure I understand your suggestion though. Do you suggest to use the byte bloom filter and have two byte representations of the asset key? Or using a better hashing algo?

My thinking: an AssetKey has a set of 'unique' fields and a set of 'informational' fields. Serialising this to bytes and using the byte array bloom filter means all the fields are considered so in the iceberg case the same data file referenced from different snapshots (or tables) is treated as 2 different Assets. I had understood your forcing `equals` and `hashCode` to be implemented by `AssetKey`s as a recognition of this and the most reliable way to compare objects.

versioned/tiered/gc/src/test/java/org/projectnessie/versioned/tiered/gc/ITTestIdentifyUnreferencedAssets.java
rymurr: I agree. I have added extra asserts to test the values. I wasn't sure how to split it up into separate tests as the 2nd half is so reliant on the first. 

I did add a new test to test the Asset search module which is independent of the value search

versioned/tiered/gc/src/test/java/org/projectnessie/versioned/tiered/gc/ITTestIdentifyUnreferencedAssets.java
rymurr: fixed

clients/hmsbridge/core/pom.xml
rymurr: the one above is because we put the spark3 dependency w/ exclusions into the parent pom. This one is already on master, not sure why its showing: https://github.com/projectnessie/nessie/pull/848/files

rymurr (commented):
Sorry, didn't see your comment about reviewable until just now :-( anyways. Just did some updates and asked a few questions. Thanks for the review :-)

snazy:
Can you remove the use of `System.currentTimeMIllis()` and replace it with some `Clock` interface that can be modified in tests as needed? Using wall clock in tests, even with rather huge intervals, is a good source for flaky tests.

rymurr:
> Can you remove the use of `System.currentTimeMIllis()` and replace it with some `Clock` interface that can be modified in tests as needed? Using wall clock in tests, even with rather huge intervals, is a good source for flaky tests.

100% agree. Was reluctant to do this as part of this change (already pulling the clock stuff up like i did was confusing this PR in my mind). Ive added a simple clock interface in the most recent change however I would rather extend its usage in a follow up.

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java
jacques-n: My suggestion was based on thinking this was at the value level, not the asset key level. I think the "right" thing to do is probably to make an asset key be composed of two binary values, the unique parts and the non-unique parts. And/or make it so that the reportable name is well defined e.g.:

AssetKey {
  repeated string name;
  byte[] data;
}

That way we can add additional generic metadata fields while declaring they aren't considered for uniqueness.

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java
jacques-n: Part of the confusion here was specifically the name ValueType, which has another meaning. In general, I'm not a huge fan of using a string as opposed to an enum.

versioned/gc/core/pom.xml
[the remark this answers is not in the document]
jacques-n: I see what you're doing now. The abbreviated paths shown in github made it less than obvious. Agreed that the second phase of gc can be core as long as the first isn't. I think part of this that I find confusing is that this also makes spark a dependency of core gc. That was why I was making a comment about how few classes could actually be in core. I was thinking you wouldn't make core gc depend on Spark. It might even make sense to call this spark-gc or something (as opposed to core)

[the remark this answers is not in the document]
rymurr: Agreed, I think core is misleading. I have removed core and named the directory spark

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java
rymurr: I was going to add an enum but it required knowledge of the `DummyValue` in tests which I didn't like. I see two options: remove entity type from `CategorizedValue` until the entity type changes go into the `VersionStore` or leave it as is and take a second pass after the entity type changes. I am not bothered either way :-)

versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java
rymurr: I have taken the 2nd option by exposing a unique name. The two byte string representation felt like a heavy change. I didn't use the reportable name as that to me contains the extra metadata, instead I exposed a unique name, I don't love this though, perhaps the reportable name should be the unique one and the toString shoudl show metadata?

rymurr:
Thanks a lot for the review @jacques-n , I have updated w/ your suggestions.

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: 29 of 45 files reviewed, 6 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/843#-MTqAQyhXUDgmSQHvoOZ-r4-31:-MUPPe_fE96vy-kCjOUT:b5cn602) ([raw file](https://github.com/projectnessie/nessie/blob/99765a1749ab3c0297302307863bf26f73fa8201/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java#L31)):*

I was going to add an enum but it required knowledge of the `DummyValue` in tests which I didn't like. I see two options: remove entity type from `CategorizedValue` until the entity type changes go into the `VersionStore` or leave it as is and take a second pass after the entity type changes. I am not bothered either way :-)
</blockquote></details>

It feels like this is still squishy given the alternative representation of byte or int or enum at the VersionStore level. Would be inclined to exclude until a consistent model is achieved.
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java, line 117 at r4](https://reviewable.io/reviews/projectnessie/nessie/843#-MTqAQyhXUDgmSQHvoO_-r4-117:-MUPPwFh4sJEZCMUOeP7:b-y3af3l) ([raw file](https://github.com/projectnessie/nessie/blob/99765a1749ab3c0297302307863bf26f73fa8201/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java#L117)):*

I have taken the 2nd option by exposing a unique name. The two byte string representation felt like a heavy change. I didn't use the reportable name as that to me contains the extra metadata, instead I exposed a unique name, I don't love this though, perhaps the reportable name should be the unique one and the toString shoudl show metadata?
</blockquote></details>

How about key and bytes? Name is weird and the information may not be string based.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: 26 of 45 files reviewed, 5 unresolved discussions (waiting on @jacques-n and @laurentgo)

___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java, line 31 at r4](https://reviewable.io/reviews/projectnessie/nessie/843#-MTqAQyhXUDgmSQHvoOZ-r4-31:-MUTocwIEqhOsaNhsqA8:bb7f2mf) ([raw file](https://github.com/projectnessie/nessie/blob/99765a1749ab3c0297302307863bf26f73fa8201/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/CategorizedValue.java#L31)):*

It feels like this is still squishy given the alternative representation of byte or int or enum at the VersionStore level. Would be inclined to exclude until a consistent model is achieved.
</blockquote></details>

done
___
*[versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java, line 117 at r4](https://reviewable.io/reviews/projectnessie/nessie/843#-MTqAQyhXUDgmSQHvoO_-r4-117:-MUToeXG4NjINncuZrxR:bb7f2mf) ([raw file](https://github.com/projectnessie/nessie/blob/99765a1749ab3c0297302307863bf26f73fa8201/versioned/gc/core/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java#L117)):*

How about key and bytes? Name is weird and the information may not be string based.
</blockquote></details>

done

rymurr:
Thanks for the feedback @jacques-n, updated based on your comments.

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: 26 of 45 files reviewed, 6 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[versioned/gc/spark/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java, line 126 at r8](https://reviewable.io/reviews/projectnessie/nessie/843#-MUVH8ah4OWfWcsiHCkm:-MUVH8ah4OWfWcsiHCkn:b-rf9myf) ([raw file](https://github.com/projectnessie/nessie/blob/f98e465d749d9f6d69de9b4fa4e3e1eb5a475b2b/versioned/gc/spark/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java#L126)):*
> ```Java
>      * Construct asset key.
>      */
>     public CategorizedAssetKey(boolean referenced, ByteString data, byte[] uniqueKey, long timestamp) {
> ```

Why not bytestring like data?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: 26 of 45 files reviewed, 6 unresolved discussions (waiting on @jacques-n and @laurentgo)

___
*[versioned/gc/spark/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java, line 126 at r8](https://reviewable.io/reviews/projectnessie/nessie/843#-MUVH8ah4OWfWcsiHCkm:-MUZ6qBp42vZColM8zsM:b-q9xw8z) ([raw file](https://github.com/projectnessie/nessie/blob/f98e465d749d9f6d69de9b4fa4e3e1eb5a475b2b/versioned/gc/spark/src/main/java/org/projectnessie/versioned/gc/IdentifyUnreferencedAssets.java#L126)):*

Why not bytestring like data?
</blockquote></details>

doh! Fixed

jacques-n (approved):
> Reviewed 23 of 32 files at r1, 2 of 14 files at r2, 1 of 7 files at r3, 3 of 8 files at r4, 3 of 4 files at r6, 4 of 9 files at r7, 6 of 9 files at r8, 3 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/843)* status: all files reviewed, 6 unresolved discussions (waiting on @jacques-n and @laurentgo)

jacques-n:
Lgtm +q
1

rymurr:
@jacques-n mind approving again? the build fell at the last (checkstyle shaped) hurdle and fixing it dismissed your review.

jacques-n (approved):

PR #849, opened 2021-02-19
Mongodb dependency

kylepbit:
Add mongodb-driver-core as an explicit dependency since the transitive dependencies of mongodb-driver-reactivestreams is incorrect for 4.2.x and causes exceptions on instantation.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/849)


kylepbit:
Closing in favour of updating dependency directly and creating against main.

PR #850, opened 2021-02-21
Use Quarkus BOM version of MongoDB

kylepbit:
Remove MongoDB from dependencyMangagement in favour of versions in the Quarkus BOM to avoid version mismatches.

Keep flapdoodle as we rely on a later version then is in the Quarkus BOM.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/850)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,7 +137,6 @@
     <logback.version>1.2.3</logback.version>
     <protobuf.version>3.15.0</protobuf.version>
     <mockito.version>3.7.7</mockito.version>
-    <mongodb.version>4.2.1</mongodb.version>
     <prometheus.version>0.9.0</prometheus.version>
     <reactor.version>2020.0.4</reactor.version>
     <serverless.version>1.5</serverless.version>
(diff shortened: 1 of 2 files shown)


jacques-n:
Is there a compatibility reason why reactive streams isn't pointing at the right Mongo. It seems a little weird to upgrade underneath our client library.

Also, isn't Mongo in the Quarkus BOM? Are we comfortable changing from what is there?

laurentgo:
Checking https://repo1.maven.org/maven2/org/mongodb/mongodb-driver-reactivestreams/4.2.1/mongodb-driver-reactivestreams-4.2.1.pom, the dependency seems correct to me, and I would suspect that Quarkus BOM might be the culprit: https://search.maven.org/artifact/io.quarkus/quarkus-bom/1.12.0.Final/pom

rymurr:
the mongo deps in our pom are also available in quarkus bom. I suspect its better to just keep compatibility w/ quarkus and remove mongo references from our `dependenciesManagement` section. We tried to be different from quarkus w/ both opentracing and metrics and it didn't work great.

kylepbit:
Updated to use the Quarkus version, thanks for the pointer on that.

rymurr (approved):
LGTM! Thanks @kylepbit

PR #862, opened 2021-02-23
Add RandomSource utility class

laurentgo:
In order for some unit tests to be repeteable, it is required to be able
to set the seed for the random generator used by the source.

Currently it requires code to accept a Random instance argument with the
test framework to set the seed before hand, but this approach makes code
heavier (the Random instance needs to be passed between function calls)
for what should is usually an internal concern.

Replace this pattern with a new utility class named `RandomSource` which
behave exactly like `ThreadLocalRandom` except that it is possible to
temporarily override the current random generator with one initialized
with a specific seed.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/862)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,6 +80,7 @@
     <module>perftest</module>
     <module>servers</module>
     <module>tools</module>
+    <module>utils</module>
   </modules>
 
   <scm>
--- a/utils/pom.xml
+++ b/utils/pom.xml
@@ -21,24 +21,19 @@
 
   <parent>
     <groupId>org.projectnessie</groupId>
-    <artifactId>nessie-versioned-tiered</artifactId>
+    <artifactId>nessie</artifactId>
     <version>0.3.1-SNAPSHOT</version>
   </parent>
 
-  <artifactId>nessie-versioned-tiered-impl</artifactId>
+  <artifactId>nessie-utils</artifactId>
 
-  <name>Nessie - Versioned - Tiered - Core Implementation</name>
+  <name>Nessie - Utils</name>
 
   <dependencies>
-    <dependency>
-      <groupId>org.projectnessie</groupId>
-      <artifactId>nessie-versioned-spi</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.projectnessie</groupId>
-      <artifactId>nessie-utils</artifactId>
-      <version>${project.version}</version>
+  <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.hamcrest</groupId>
@@ -46,35 +41,13 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>com.google.protobuf</groupId>
-      <artifactId>protobuf-java</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.immutables</groupId>
-      <artifactId>value</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter-engine</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-classic</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>ch.qos.logback</groupId>
-      <artifactId>logback-core</artifactId>
+      <artifactId>junit-jupiter-params</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/utils/src/main/java/org/projectnessie/Closeable.java
+++ b/utils/src/main/java/org/projectnessie/Closeable.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie;
+
+/**
+ * A {@code AutoCloseable} extension which doesn't need
+ * to throw checked exceptions.
+ *
+ */
+@FunctionalInterface
+public interface Closeable extends AutoCloseable {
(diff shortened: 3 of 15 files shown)


rymurr (commented):
> Reviewed 16 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/862)* status: all files reviewed, 3 unresolved discussions (waiting on @laurentgo)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI3Jjg--OW3nyiGCBU:-MUI3Jjg--OW3nyiGCBV:bukaxfd) (no related file):*
LGTM, a few comments for clarification.

---
___
*[utils/src/main/java/org/projectnessie/RandomSource.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI1KHz5jWq2OASSzNA:-MUI1KHz5jWq2OASSzNB:bh3w2ql) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/src/main/java/org/projectnessie/RandomSource.java#L36)):*
> ```Java
>   private static final RandomSource INSTANCE = new RandomSource();
> 
>   private final boolean isInitialized;
> ```

this seems weird. Maybe its just a part of java i dont understand though. To me this field is always true and never used
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 116 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI3-qBD5JRvlJ7cP9J:-MUI3-qBD5JRvlJ7cP9K:b-lv0sb3) ([raw file](https://github.com/projectnessie/nessie/blob/aaccb0ba6799c319759a924c323556ac2317b123/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L116)):*
> ```Java
>   protected abstract S createRawStore();
> 
>   protected abstract long getRandomSeed();
> ```

does this mean we made every impl set a random seed, they all set it the same and it wasn't required?

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/862)* status: all files reviewed, 4 unresolved discussions (waiting on @laurentgo and @rymurr)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI3Jjg--OW3nyiGCBU:-MUIWhcbBcn2mBAPhp8K:b-auopdc) (no related file):*

LGTM, a few comments for clarification.
</blockquote></details>

I'd favor a slightly different annotation/JUnit-extension that auto-magically sets up a `RandomSource` with an optional seed. I.e. instead of having a `@WithSeed` annotation, have an annotation `@WithRandom` that optionally takes a seed value and falls back to a random seed value.

To make tests reproducible, it would be necessary to somehow print the seed value of a failed test run, as you need to know that seed value to reproduce the test failure. To (quickly) re-run a test with a known seed value, it feels easier to specify the seed-value as a system property.

Not sure why some tests already specify a constant seed-value. Why do we specify constant seed-values at all?

Can you add some tests for the JUnit extension?

---
___
*[utils/pom.xml, line 28 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUIND9A-rgRITSUOAU1:-MUIND9A-rgRITSUOAU2:b-hej36y) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/pom.xml#L28)):*
> ```XML
>   </parent>
> 
>   <artifactId>nessie-utils</artifactId>
> ```

Can you rename this module to `nessie-test-utils`?
___
*[utils/src/main/java/org/projectnessie/RandomSource.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI1KHz5jWq2OASSzNA:-MUIOqzb0j7P7JsOkNVx:bpxlnbz) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/src/main/java/org/projectnessie/RandomSource.java#L36)):*

this seems weird. Maybe its just a part of java i dont understand though. To me this field is always true and never used
</blockquote></details>

Def. needs some inline-docs, but the default `Random` constructor is called before `RandomSource`'s constructor's being executed. `Random`'s constructor calls `setSeed`.

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/862)* status: all files reviewed, 4 unresolved discussions (waiting on @rymurr and @snazy)

___
*[utils/pom.xml, line 28 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUIND9A-rgRITSUOAU1:-MUKVfAY0NmDY07NwVAH:b-2bq9mx) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/pom.xml#L28)):*

Can you rename this module to `nessie-test-utils`?
</blockquote></details>

I would prefer not as `RandomSource` is used outside of test contexts...
___
*[utils/src/main/java/org/projectnessie/RandomSource.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI1KHz5jWq2OASSzNA:-MUKVomI84bfSpY0N2lF:b-cycud6) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/src/main/java/org/projectnessie/RandomSource.java#L36)):*

Def. needs some inline-docs, but the default `Random` constructor is called before `RandomSource`'s constructor's being executed. `Random`'s constructor calls `setSeed`.
</blockquote></details>

This is because `Random` base class is weird: base constructor will call `setSeed`, so the call should not fail until the flag is set to `true` which cause later calls to `setSeed` to throw `UOE`.
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 116 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI3-qBD5JRvlJ7cP9J:-MUKWE_cAXzBILlVbCm8:b-jx7kz) ([raw file](https://github.com/projectnessie/nessie/blob/aaccb0ba6799c319759a924c323556ac2317b123/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L116)):*

does this mean we made every impl set a random seed, they all set it the same and it wasn't required?
</blockquote></details>

not sure what you comment is referencing tbh...

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/862)* status: all files reviewed, 4 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[utils/src/main/java/org/projectnessie/RandomSource.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI1KHz5jWq2OASSzNA:-MUNIV9wAA_ypWMPKDHR:btvsxud) ([raw file](https://github.com/projectnessie/nessie/blob/a7fd974fd451bcb7210ef4e1e18ad1967cca9972/utils/src/main/java/org/projectnessie/RandomSource.java#L36)):*

This is because `Random` base class is weird: base constructor will call `setSeed`, so the call should not fail until the flag is set to `true` which cause later calls to `setSeed` to throw `UOE`.
</blockquote></details>

Can we just add that to the javadoc or stick a comment in somewhere?
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java, line 116 at r1](https://reviewable.io/reviews/projectnessie/nessie/862#-MUI3-qBD5JRvlJ7cP9J:-MUNIbVh3e0tAibnkWda:b-7tk5zm) ([raw file](https://github.com/projectnessie/nessie/blob/aaccb0ba6799c319759a924c323556ac2317b123/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractTestStore.java#L116)):*

not sure what you comment is referencing tbh...
</blockquote></details>

nws, will discuss offline

rymurr:
closing as significantly out of date

PR #866, opened 2021-02-24
add WithPayload class to VersionStore interface

### LINKED ISSUE #836: Add entity-type-id to TieredVersionStore's KeyList

rymurr:
* adds entity type to back-end stores
* maps entity type int to Contents type

Areas to think about/change:
* int entity type the best way to go?
* the wrapper method better than adding an `extends` to value parameter
* is entity type the correct name?

Closes #836

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)


servers/quarkus-server/src/main/proto/table.proto
```diff
@@ -32,24 +32,29 @@ message PContents {
 
 message PIcebergTable {
   string metadata_location = 1;
+  int32 entityType = 2;
```
rymurr, line 35: I think the entityType should be first but didn't want to break the proto w/o checking w/ the team

versioned/spi/src/main/java/org/projectnessie/versioned/StoreWorker.java
```diff
@@ -23,20 +28,42 @@
...
+   * Create StoreWorker for provided helpers. Adding single byte data type as first byte in serialized value.
    */
   static <VALUE, COMMIT_METADATA> StoreWorker<VALUE, COMMIT_METADATA>
       of(ValueWorker<VALUE> valueWorker, Serializer<COMMIT_METADATA> commitSerializer) {
     return new StoreWorker<VALUE, COMMIT_METADATA>() {
 
       @Override
-      public ValueWorker<VALUE> getValueWorker() {
-        return valueWorker;
+      public ValueWorker<WithEntityType<VALUE>> getValueWorker() {
```
rymurr, line 43: This is only used in tests currently and I think it shoudl mvoe into a test package. It wraps a string worker to also record the type

versioned/tiered/mongodb/src/main/java/org/projectnessie/versioned/mongodb/MongoFragment.java
```diff
@@ -30,21 +33,24 @@
 
   static void produce(Document document, Fragment v) {
     produceBase(document, v)
-        .keys(deserializeKeys(document, KEY_LIST));
+        .keys(deserializeKeysWithEntityType(document, KEY_LIST));
```
rymurr, line 36: @stevelorddremio @kylep-dremio mongo also changed a bit. I added the entity type field here

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalL3.java
```diff
@@ -36,27 +38,27 @@
 
   private static final long HASH_SEED = 4604180344422375655L;
 
-  private final TreeMap<InternalKey, PositionDelta> map;
+  private final TreeMap<InternalKey, Map.Entry<PositionDelta, Integer>> map;
```
rymurr, line 41: i think this should change to a small container type rather than this generic pair

laurentgo (changes requested):
> Reviewed 6 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 6 of 42 files reviewed, 12 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 75 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPA3sXDq1kosGDTdVj:-MUPA3sYEBCK2fzUs-JB:bkcv7gf) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L75)):*
> ```Java
>   }
> 
>   @JsonIgnore
> ```

Should we change the `@JsonTypeInfo` annotation to use `include = JsonTypeInfo.As.EXISTING_PROPERTY` instead?
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 86 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAE4U0E9haye1gVMi:-MUPAE4U0E9haye1gVMj:b82ng29) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L86)):*
> ```Java
>    * @return byte entity type
>    */
>   static Type contentsTypeFromEntityType(int entityType) {
> ```

Should it be part of the model?
___
*[model/src/main/java/org/projectnessie/model/DeltaLakeTable.java, line 42 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAYazCFN90InpZLdA:-MUPAYazCFN90InpZLdB:bbz84k1) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/DeltaLakeTable.java#L42)):*
> ```Java
> 
>   @Override
>   default Type getEntityType() {
> ```

Maybe there's a compelling use-case to convert this interface (and all other Contents subinterface) to an abstract class and make this field final?
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPBrHh4MJhaiwhVk1Q:-MUPBrHh4MJhaiwhVk1R:b-1o7kxi) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java#L82)):*
> ```Java
>         builder.setIcebergTable(
>             PIcebergTable.newBuilder().setMetadataLocation(((IcebergTable) value).getMetadataLocation())
>                 .setEntityType(wrappedValue.getEntityType()));
> ```

The fact that the message is a PIcebergTable isn't it enough to get the type?
___
*[servers/quarkus-server/src/main/proto/table.proto, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUK-8IcG0JWlLLNcTqU-r1-35:-MUPCC187n0CQpSBXXLQ:b-bts84e) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/proto/table.proto#L35)):*

I think the entityType should be first but didn't want to break the proto w/o checking w/ the team
</blockquote></details>

I don't understand why the field is actually needed to be honest. Isn't the message type enough?
___
*[servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java, line 105 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPGlLfA73XLhUAppVX:-MUPGlLfA73XLhUAppVY:b9h8fx9) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java#L105)):*
> ```
> Delete
> ```
I'm not sure why Delete and Unchanged need the type
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/Key.java, line 109 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPFPi_7K3-w43dUbiI:-MUPFPi_7K3-w43dUbiJ:b-3glezk) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java#L109)):*
> ```Java
>     public abstract Key getKey();
> 
>     public abstract int getEntityType();
> ```

I'm not sure I get why the type needs to be exposed for the mutation
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java, line 186 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPF3af9JWHNFmsPNdY:-MUPF3af9JWHNFmsPNdZ:b-az9c8j) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java#L186)):*
> ```Java
>    * @throws ReferenceNotFoundException if {@code ref} is not present in the store
>    */
>   WithEntityType<VALUE> getValue(Ref ref, Key key) throws ReferenceNotFoundException;
> ```

Why is the entity type needed as it is part of the value anyway?
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MUPDAuQ9NGXO3i1ihr-:b-hcn8e6) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*
> ```Java
>    * Get the byte identifier of the type of this value.
>    */
>   int getEntityType();
> ```

This is a storage implementation leak imho. The return type should be the enum defined in model. Since we are discussing generic, I guess the interface could be defined as
`interface WithEntityType<E, T extends Enum<T>>`

jacques-n (commented):
> Reviewed 6 of 42 files at r1.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 18 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPOV9y03rHRNiKW6pX:-MUPOV9y03rHRNiKW6pY:bq5oa0e) (no related file):*
It feels like WithEntityType is used in far more places than necessary. Introducing a base VALUE interface and then limiting WithEntityType to the keylisting interface seems like a narrower change.

Separately, I think we need to be conscientious about where we store EntityType. My thinking is that Fragment and L3 are the only places we would store this information in the context of TieredVersionStore. I don't see any reason to store in the contents as we already have oneof which should capture the same meaning.

---
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 77 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPLWbm20Y_oRILC9D6:-MUPLWbm20Y_oRILC9D7:b-jwm34u) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L77)):*
> ```Java
>   @JsonIgnore
>   default Type getEntityType() {
>     return Type.UNKNOWN;
> ```

Do we need a default implementation here? It seems undesirable to have one...
___
*[model/src/main/java/org/projectnessie/model/DeltaLakeTable.java, line 42 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAYazCFN90InpZLdA:-MUPMFNJA6ufkl2oUp06:b-xsd5ra) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/DeltaLakeTable.java#L42)):*

Maybe there's a compelling use-case to convert this interface (and all other Contents subinterface) to an abstract class and make this field final?
</blockquote></details>

I'm generally against having a field represent something that the class type already represents (which is the case here).  (I'm ambivalent about switching this to abstract class so we can avoid the weird default definition.)
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPBrHh4MJhaiwhVk1Q:-MUPJmqj9JHfV9wivHna:b-yzw6eo) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java#L82)):*

The fact that the message is a PIcebergTable isn't it enough to get the type?
</blockquote></details>

My comment as well. Don't understand why this is in the proto definitions.
___
*[servers/quarkus-server/src/main/proto/table.proto, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUK-8IcG0JWlLLNcTqU-r1-35:-MUPLfKC634ZLvsmowuo:b6aom1h) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/proto/table.proto#L35)):*

I don't understand why the field is actually needed to be honest. Isn't the message type enough?
</blockquote></details>

Agree with Laurent. Why do we have these?
___
*[servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java, line 105 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPGlLfA73XLhUAppVX:-MUPMf1C-P9PR0HVHERt:b-7ttm3t) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java#L105)):*

> ```
> Delete
> ```
I'm not sure why Delete and Unchanged need the type
</blockquote></details>

ditto. 
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPM2Am7VI6IZLRMgKk:-MUPM2Am7VI6IZLRMgKl:bjezv8b) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java#L36)):*
> ```Java
>  */
> @ThreadSafe
> public interface VersionStore<VALUE, METADATA> {
> ```

Shouldn't we just have VALUE be swapped with T extends ValueInterface where ValueInterface has getEntityType().
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java, line 186 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPF3af9JWHNFmsPNdY:-MUPJvIW2plnrGQHpfZA:bxb279u) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java#L186)):*

Why is the entity type needed as it is part of the value anyway?
</blockquote></details>

Agreed, I don't think anything that returns VALUE should use WithEntityType.
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 29 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPKVbQ-hgcXXmOGGpj:-MUPKVbQ-hgcXXmOGGpk:bw94e71) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L29)):*
> ```Java
> 
>   /**
>    * Get the byte identifier of the type of this value.
> ```

On a separate note, I think the storage capacity here should be byte, not integer (as per your javadoc). Not sure why you're int in the signature.
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MUPK4-Q33wSNx5rMCWE:bfs3427) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*

This is a storage implementation leak imho. The return type should be the enum defined in model. Since we are discussing generic, I guess the interface could be defined as
`interface WithEntityType<E, T extends Enum<T>>`
</blockquote></details>

I'm mixed on this point. I think this as an extension of the key and on that point, a integer type seems fine.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 120 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPNJfeB6gwdbb3SynV:-MUPNJfeB6gwdbb3SynW:b-4q7zpa) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L120)):*
> ```Java
>   static AttributeValue keyElementsWithEntityType(WithEntityType<Key> key) {
>     Map<String, AttributeValue> keyWithEntityType = new HashMap<>();
>     keyWithEntityType.put(ENTITY_TYPE, number(checkNotNull(key).getEntityType()));
> ```

I don't think this is a good way to store this data. I think we probably need to come up with something more compact. Generally speaking, the entity type can be thought of as simply part of the key for this kind of operation so treating as a suffix or prefix on the existing array seems like it would make sense.

We should also consider if we start making this backwards compatible/safe.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java, line 157 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPL-q72pLEhevfmdcY:-MUPL-q72pLEhevfmdcZ:b-d8l65t) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java#L157)):*
> ```Java
> 
>       Set<InternalKey> removals = new HashSet<>();
>       Map<InternalKey, WithEntityType<InternalKey>> adds = new HashMap<>();
> ```

Let's keep this as a set with a different object. Using a map seems like using a generic tuple and is a bit confusing.

I'd additionally say that we should probably avoid using public api classes like WithEntityType for purely internal concerns.

rymurr:
Thanks a lot @laurentgo and @jacques-n for your reviews! Before I start addressing the points in a follow up commit I am wondering what you think about the `WithEntityType` class or if we just do `VersionStore<T extends HasEntityType>` or so. I rejected that approach early in the design of this patch but its starting to feel more desirable (your comments hint thatyou may agree)

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 18 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MURgqcTEFkG3fi7c73K:b-9mwelb) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*

I'm mixed on this point. I think this as an extension of the key and on that point, a integer type seems fine.
</blockquote></details>

See my comment, I definitively do not see type as an extension of the key, but a important part of the value associated with it
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 120 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPNJfeB6gwdbb3SynV:-MURgPS-27tA9mkjXNls:by30n78) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L120)):*

I don't think this is a good way to store this data. I think we probably need to come up with something more compact. Generally speaking, the entity type can be thought of as simply part of the key for this kind of operation so treating as a suffix or prefix on the existing array seems like it would make sense.

We should also consider if we start making this backwards compatible/safe.
</blockquote></details>

I'm not sure it makes sense to me: wouldn't mean that the same name + different types would be 2 different keys? wouldn't that create weird scenarios?

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 18 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 120 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPNJfeB6gwdbb3SynV:-MURmbXvEOSbpchQ_Bmd:bqhiax0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L120)):*

I'm not sure it makes sense to me: wouldn't mean that the same name + different types would be 2 different keys? wouldn't that create weird scenarios?
</blockquote></details>

This is about how it is stored in dynamo and wouldn't change the logical behavior.

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 18 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MURrtVz3jmsVn4w2akG:b-frb4ou) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*

See my comment, I definitively do not see type as an extension of the key, but a important part of the value associated with it
</blockquote></details>

You'll then need to define a "serializer" to get back to an Enum from a byte (as I'd be disinclined to use the default ordinals). It feels excessive but I'm fine with it. The one that this pattern doesn't allow that I was thinking about using is concepts like byte masks and using the individual bits differently to potentially enhance the type of data that we store in this value. For example, maybe we say the first 4 bits are for entity type and the next set of bits are for something else (maybe security related). In those circumstances being able to express a bitmask filter would be helpful and I don't see how to expose that with this structured enum pattern.

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 21 unresolved discussions (waiting on @laurentgo and @rymurr)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSX33N31Z9RxDJSuLx:-MUSX33N31Z9RxDJSuLy:b-dbv1f7) (no related file):*
IIUC the "entity-type" is related to the `Key`, so I wonder why `Key` doesn't expose the entity-type but a `Key`'s always wrapped with `WithEntityType`. I.e. why not just add `getEntityType()` to `Key`?

---
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSR5cQ4tImPXlpbdVO:-MUSR5cQ4tImPXlpbdVP:bqcsuqe) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L82)):*
> ```Java
>       List<ContentsKey> externalKeys = request.getRequestedKeys();
>       List<Key> internalKeys = externalKeys.stream().map(ContentsResource::toKey).collect(Collectors.toList());
>       List<Optional<Contents>> values = getStore().getValues(ref, internalKeys).stream().map(x -> x.map(WithEntityType::getValue))
> ```

No need to stream+collect here, you can do the transformation in the loop below via `Optional`'s `map()`
___
*[versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java, line 446 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSRxiP3_oPldGzz5Ay:-MUSRxiP3_oPldGzz5Az:bm4e2yj) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java#L446)):*
> ```Java
>         })
>         // extract the keys
>         .map(x -> {
> ```

maybe a more descriptive name than `x`
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MUSU6vcCW3BO22yXw9S:b-mcbh5x) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*

See my comment, I definitively do not see type as an extension of the key, but a important part of the value associated with it
</blockquote></details>

Agree with Laurent here. Maybe not an enum, but at least some "enum-ish-class" that can represent the known types but also a generic "unknown" one, which then can expose an ordinal.

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 12 of 42 files reviewed, 21 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalL3.java, line 41 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUK-8IcG0JWlLLNcTqs-r1-41:-MUSXfpX7-Yw6VX21LnV:bg5mymu) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalL3.java#L41)):*

i think this should change to a small container type rather than this generic pair
</blockquote></details>

`TreeMap`s are super expensive structures. Not sure, how big/expensive this whole particular object-tree for this tree-map can become. If this whole thing becomes "too big" and becomes a visible perf/heap issue, then we should think about using less expensive structures (aka optimizations). For now, it's okay-ish IMO.

But having a container type for this InternalL3 isn't a bad idea ;)

rymurr:
Been thinking about this change. I think I see two ways forward:

1. Change `VersionStore<VALUE, METADATA>` to `VersionStore<VALUE extends HasEntityType, METADATA>`
2. Add a `byte getEntityType(VALUE value)` to `ValueWorker`

I am not a huge fan of number one at the moment as it means all of our `Contents` subtypes have to expose a method to tell the world what type they are even though its known from the type. These are public objects so they have to get pushed to the client side and stuff. I think some of the comments alluded to this issue as well.

I like 2. at the moment because the entity type is really only needed in a narrow set of places and this way we narrow the number of places/people that can see entity type. it also decouples the labelling of the types from the actual types.

Wondering what other people think? @jacques-n @laurentgo @snazy

jacques-n (commented):
The getEntityType is insufficient. You also need the inverse: byte => type when returning the key stream.

As mentioned elsewhere, I'm inclined to have the versionstore to expose this as a meaningless single byte payload that is maintained and let higher levels create meaning around it. Much more like a size-constrained version of Lucene's [Payload|https://lucidworks.com/post/getting-started-with-payloads/] from a VersionStore perspective. Applying meaning in the version store api (even seems like excessive coupling of meaning. Maybe call it WithPayload instead of WithEntityType...

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866#-:-MUZq2zh0BbElT0vXq8N:b7pbdvm)* status: 12 of 42 files reviewed, 21 unresolved discussions (waiting on @laurentgo and @rymurr)

rymurr:
@jacques-n @snazy  @laurentgo  I have updated based on your comments. I have not gone through each comment individually as the change is quite significant.

The approach now: Add payload to `Key` object. This was suggested by @snazy and in light of some of @jacques-n comments it makes a lot of sense to me. The code changes are rather small now, hope its more sensible than last iteration!

rymurr:
Offline conversations led to the decision that an entity type isn't really a property of a Key and really messes with key equality. Making some modifications here to address that.

rymurr:
This is ready for review again after removing the payload from Key and putting it next to the value

jacques-n (commented):
> Reviewed 3 of 65 files at r2.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 1 of 49 files reviewed, 30 unresolved discussions (waiting on @jacques-n, @laurentgo, and @rymurr)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 60 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDQgvg9C2_qZD2iG5s:-MVDQgvg9C2_qZD2iG5t:b-f3u0n6) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/model/src/main/java/org/projectnessie/model/Contents.java#L60)):*
> ```Java
> 
>     public static Type fromPayload(Byte payload) {
>       if (payload == null || payload == 0 || payload > 5) {
> ```

you should add the value as a property of the enum.

UNKNOWN(0), ICEBERG_TABLE(1), etc. If you want a function like this, then probably do as index into a static array of values.
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 97 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDR0Yx1Q3S7zsmhnZj:-MVDR0Yx1Q3S7zsmhnZk:b-9lzbh2) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/model/src/main/java/org/projectnessie/model/Contents.java#L97)):*
> ```Java
>    */
>   default byte toPayload() {
>     if (this instanceof IcebergTable) {
> ```

Rather than define a default here, let's define on the actual impls. I'd also define this method as protected so it doesn't show up out of package except for overriders. (move the iface to abstract class)
___
*[servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java, line 175 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDRQHEAnTszSSnha4P:-MVDRQHEAnTszSSnha4Q:bddxx5c) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java#L175)):*
> ```Java
> 
>     @Override
>     public Byte getPayload(CommitMeta value) {
> ```

why is there a payload for commit meta? What is the thinking there?
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/Serializer.java, line 29 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDRyYk769JqZeJT5ZN:-MVDRyYk769JqZeJT5ZO:bjopmll) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/spi/src/main/java/org/projectnessie/versioned/Serializer.java#L29)):*
> ```Java
>   V fromBytes(ByteString bytes);
> 
>   Byte getPayload(V value);
> ```

I wouldn't put this on serializer directly. It means we have to implement in lots of places when we really only want it for VALUE. Either create SerializerWithPayload that extends Serializer or put it some place else.
___
*[versioned/tests/src/main/java/org/projectnessie/versioned/StringSerializer.java, line 61 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDSVAy20A83lZAmTLq:-MVDSVAy20A83lZAmTLr:bq3snim) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tests/src/main/java/org/projectnessie/versioned/StringSerializer.java#L61)):*
> ```Java
>    * set the byte payload for testing purposes.
>    */
>   public static void setPayload(Byte payload) {
> ```

The pattern in this class doesn't make sense to me (code smell). Why not just define separate classes or subclasses or something? Or mock. Or?
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 112 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDT2mx6KxOOF71qOuq:-MVDT2mx6KxOOF71qOur:b-yptjlj) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L112)):*
> ```Java
>   static AttributeValue keyElementsWithPayload(WithPayload<Key> key) {
>     Stream<AttributeValue> keyValue = checkNotNull(key.getValue()).getElements().stream().map(AttributeValueUtil::string);
>     AttributeValue payload = AttributeValueUtil.string(key.getPayload() == null
> ```

(style) Once a ternary goes to multiple lines, inclined to move to bracketed syntax.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoL3.java, line 78 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDTL6qEb20sBuTsKfX:-MVDTL6qEb20sBuTsKfY:b-yxihv1) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoL3.java#L78)):*
> ```Java
>           .map(m -> {
>             WithPayload<Key> key = deserializeKeyWithPayload(attributeValue(m, TREE_KEY));
>             return KeyDelta.of(key.getValue(), deserializeId(m, TREE_ID), key.getPayload());
> ```

Maybe add constructors of KeyDelta that take a WithPayload<Key> rather than doing current pattern?
___
*[versioned/tiered/gc/src/test/java/org/projectnessie/versioned/impl/TestValueRetriever.java, line 161 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDTV1q5AmWfga1ZlNX:-MVDTV1q5AmWfga1ZlNY:b-oqemmc) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/gc/src/test/java/org/projectnessie/versioned/impl/TestValueRetriever.java#L161)):*
> ```Java
>     Id val2 = Id.generateRandom();
>     InternalL3 l3 = InternalL3.EMPTY.set(new InternalKey(Key.of("foo")), val1, (byte) 0)
>         .set(new InternalKey(Key.of("bar")), val2, (byte) 0);
> ```

I suggest you put this in a static constant or at least a separate field declaration. Looks a big magical here (given the cast).
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/tiered/Mutation.java, line 59 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDUZ6P5TEMyncgYSNe:-MVDUZ6P5TEMyncgYSNf:bjv5soy) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/tiered/Mutation.java#L59)):*
> ```Java
> 
>   @Value.Immutable
>   abstract class Modification implements Mutation {
> ```

Same comments as elsewhere wrt modification.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalMutation.java, line 94 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDU0NJ9CJh5zjq7DZm:-MVDU0NJ9CJh5zjq7DZn:b-qcehjl) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalMutation.java#L94)):*
> ```Java
> 
>   @Immutable
>   public abstract static class InternalModification extends InternalMutation {
> ```

I struggle with the value of modification. It seems like an excessive detail. Why not just treat as addition and removal (similar to how iceberg doesn't actually have modifications, only additions and deletions)? I guess clarity of ordering? Maybe we just declare if deletes are applied first. Solves the ordering problem without adding extra complexion here.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 60 files reviewed, 28 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPOV9y03rHRNiKW6pX:-MVH5IFr8kNb6lP6Cupb:b-qqn5s0) (no related file):*

It feels like WithEntityType is used in far more places than necessary. Introducing a base VALUE interface and then limiting WithEntityType to the keylisting interface seems like a narrower change.

Separately, I think we need to be conscientious about where we store EntityType. My thinking is that Fragment and L3 are the only places we would store this information in the context of TieredVersionStore. I don't see any reason to store in the contents as we already have oneof which should capture the same meaning.

</blockquote></details>

Done. Stale

---
*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSX33N31Z9RxDJSuLx:-MVH5IYd99r_SYruzQOj:b-qqn5s0) (no related file):*

IIUC the "entity-type" is related to the `Key`, so I wonder why `Key` doesn't expose the entity-type but a `Key`'s always wrapped with `WithEntityType`. I.e. why not just add `getEntityType()` to `Key`?
</blockquote></details>

Done. Stale

---
*[a discussion](https://reviewable.io/reviews/projectnessie/nessie/866#-MVH5FXrAAQ4kGek-2Hb:-MVH5FXs3xBqeGSXGonW:b-2s9090) (no related file):*
I have dismissed all stale comments. Apologies for the nosie

---
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 75 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPA3sXDq1kosGDTdVj:-MVH5KSaBj7v9dpf-2LY:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L75)):*

Should we change the `@JsonTypeInfo` annotation to use `include = JsonTypeInfo.As.EXISTING_PROPERTY` instead?
</blockquote></details>

Done. Stale
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 77 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPLWbm20Y_oRILC9D6:-MVH5L-c416yEyvJMqRZ:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L77)):*

Do we need a default implementation here? It seems undesirable to have one...
</blockquote></details>

Done. Stale
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 86 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAE4U0E9haye1gVMi:-MVH5LelDVVOGltKw6A3:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L86)):*

Should it be part of the model?
</blockquote></details>

Done. Stale
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 60 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDQgvg9C2_qZD2iG5s:-MVH50820cbW3e5PHzd7:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/model/src/main/java/org/projectnessie/model/Contents.java#L60)):*

you should add the value as a property of the enum.

UNKNOWN(0), ICEBERG_TABLE(1), etc. If you want a function like this, then probably do as index into a static array of values.
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 97 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDR0Yx1Q3S7zsmhnZj:-MVH7QTK-mTPJP5qiBz5:bquu1oe) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/model/src/main/java/org/projectnessie/model/Contents.java#L97)):*

Rather than define a default here, let's define on the actual impls. I'd also define this method as protected so it doesn't show up out of package except for overriders. (move the iface to abstract class)
</blockquote></details>

Done. This is a bit weird as the Value Serializer needs to get the byte value. I have added a static method to get it. It looks a bit weird but it hides the protected method which is desirable.
___
*[model/src/main/java/org/projectnessie/model/DeltaLakeTable.java, line 42 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAYazCFN90InpZLdA:-MVH7mSjDx5Kozrsjuap:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/DeltaLakeTable.java#L42)):*

I'm generally against having a field represent something that the class type already represents (which is the case here).  (I'm ambivalent about switching this to abstract class so we can avoid the weird default definition.)
</blockquote></details>

Done. Stale
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPBrHh4MJhaiwhVk1Q:-MVH7stBChZGyoibe34d:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/java/org/projectnessie/server/providers/TableCommitMetaStoreWorker.java#L82)):*

My comment as well. Don't understand why this is in the proto definitions.
</blockquote></details>

Done. Stale
___
*[servers/quarkus-server/src/main/proto/table.proto, line 35 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUK-8IcG0JWlLLNcTqU-r1-35:-MVH7tkX9tI8oqrN0WT_:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/quarkus-server/src/main/proto/table.proto#L35)):*

Agree with Laurent. Why do we have these?
</blockquote></details>

Done. Stale
___
*[servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java, line 105 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPGlLfA73XLhUAppVX:-MVH7urE8flSo5PzT4Kr:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java#L105)):*

ditto. 
</blockquote></details>

Done. Stale
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 82 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSR5cQ4tImPXlpbdVO:-MVH7vbnCnB4JXqe3MSp:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L82)):*

No need to stream+collect here, you can do the transformation in the loop below via `Optional`'s `map()`
</blockquote></details>

Done. Stale
___
*[servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java, line 175 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDRQHEAnTszSSnha4P:-MVH7xzsD0tYVL1ghGfi:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java#L175)):*

why is there a payload for commit meta? What is the thinking there?
</blockquote></details>

Done.
___
*[versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java, line 446 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUSRxiP3_oPldGzz5Ay:-MVH8dSq8_U9fH4vNQrL:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java#L446)):*

maybe a more descriptive name than `x`
</blockquote></details>

Done. Stale
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/Key.java, line 109 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPFPi_7K3-w43dUbiI:-MVH8f3d2cs3OMRtDmVL:bnfeolm) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java#L109)):*

I'm not sure I get why the type needs to be exposed for the mutation
</blockquote></details>

Done. Stale
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/Serializer.java, line 29 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDRyYk769JqZeJT5ZN:-MVH8gb00C5KLqY4B2OL:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/spi/src/main/java/org/projectnessie/versioned/Serializer.java#L29)):*

I wouldn't put this on serializer directly. It means we have to implement in lots of places when we really only want it for VALUE. Either create SerializerWithPayload that extends Serializer or put it some place else.
</blockquote></details>

Done.
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java, line 36 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPM2Am7VI6IZLRMgKk:-MVH8rE_7bJ6OnT9EUdA:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java#L36)):*

Shouldn't we just have VALUE be swapped with T extends ValueInterface where ValueInterface has getEntityType().
</blockquote></details>

Done. Stale
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java, line 186 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPF3af9JWHNFmsPNdY:-MVH8s7AAKQBi5IYeILS:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java#L186)):*

Agreed, I don't think anything that returns VALUE should use WithEntityType.
</blockquote></details>

Done. Stale
___
*[versioned/tests/src/main/java/org/projectnessie/versioned/StringSerializer.java, line 61 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDSVAy20A83lZAmTLq:-MVH932G6jtNhjKdzyR6:bq47uhq) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tests/src/main/java/org/projectnessie/versioned/StringSerializer.java#L61)):*

The pattern in this class doesn't make sense to me (code smell). Why not just define separate classes or subclasses or something? Or mock. Or?
</blockquote></details>

just being lazy. Fixed
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 120 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPNJfeB6gwdbb3SynV:-MVH9Mr482RGQxNE1dAY:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L120)):*

This is about how it is stored in dynamo and wouldn't change the logical behavior. 
</blockquote></details>

Done. Stale
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java, line 112 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDT2mx6KxOOF71qOuq:-MVH9MG53SebYpTigRPp:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/AttributeValueUtil.java#L112)):*

(style) Once a ternary goes to multiple lines, inclined to move to bracketed syntax.
</blockquote></details>

Done.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoL3.java, line 78 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDTL6qEb20sBuTsKfX:-MVHAOza6bH3BpeWeYKP:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoL3.java#L78)):*

Maybe add constructors of KeyDelta that take a WithPayload<Key> rather than doing current pattern?
</blockquote></details>

Done.
___
*[versioned/tiered/gc/src/test/java/org/projectnessie/versioned/impl/TestValueRetriever.java, line 161 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDTV1q5AmWfga1ZlNX:-MVHBgEiBGh02fajLuad:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/gc/src/test/java/org/projectnessie/versioned/impl/TestValueRetriever.java#L161)):*

I suggest you put this in a static constant or at least a separate field declaration. Looks a big magical here (given the cast).
</blockquote></details>

Done.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalL3.java, line 41 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUK-8IcG0JWlLLNcTqs-r1-41:-MVHBimbCZQN9bDatVqH:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalL3.java#L41)):*

`TreeMap`s are super expensive structures. Not sure, how big/expensive this whole particular object-tree for this tree-map can become. If this whole thing becomes "too big" and becomes a visible perf/heap issue, then we should think about using less expensive structures (aka optimizations). For now, it's okay-ish IMO.

But having a container type for this InternalL3 isn't a bad idea ;)
</blockquote></details>

Done. Stale
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java, line 157 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPL-q72pLEhevfmdcY:-MVHBlmG4mDP5iIXKgLU:bnfeolm) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java#L157)):*

Let's keep this as a set with a different object. Using a map seems like using a generic tuple and is a bit confusing.

I'd additionally say that we should probably avoid using public api classes like WithEntityType for purely internal concerns.
</blockquote></details>

Done. Stale
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/tiered/Mutation.java, line 59 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDUZ6P5TEMyncgYSNe:-MVHCcs85P6LUeeKi2G7:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/tiered/Mutation.java#L59)):*

Same comments as elsewhere wrt modification.
</blockquote></details>

Done.
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 29 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPKVbQ-hgcXXmOGGpj:-MVH8sqk70z5h3W1stg7:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L29)):*

On a separate note, I think the storage capacity here should be byte, not integer (as per your javadoc). Not sure why you're int in the signature.
</blockquote></details>

Done. Stale
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java, line 31 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPDAuQ9NGXO3i1ihqz:-MVH8tha819KywId0VCz:b-qqn5s0) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/versioned/spi/src/main/java/org/projectnessie/versioned/WithEntityType.java#L31)):*

You'll then need to define a "serializer" to get back to an Enum from a byte (as I'd be disinclined to use the default ordinals). It feels excessive but I'm fine with it. The one that this pattern doesn't allow that I was thinking about using is concepts like byte masks and using the individual bits differently to potentially enhance the type of data that we store in this value. For example, maybe we say the first 4 bits are for entity type and the next set of bits are for something else (maybe security related). In those circumstances being able to express a bitmask filter would be helpful and I don't see how to expose that with this structured enum pattern.
</blockquote></details>

Done. Stale
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalMutation.java, line 94 at r2](https://reviewable.io/reviews/projectnessie/nessie/866#-MVDU0NJ9CJh5zjq7DZm:-MVHChD83i26LQF1JHKD:bpx244v) ([raw file](https://github.com/projectnessie/nessie/blob/4e1aea9fb8adaad17816485e9aab2626a858f30e/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalMutation.java#L94)):*

I struggle with the value of modification. It seems like an excessive detail. Why not just treat as addition and removal (similar to how iceberg doesn't actually have modifications, only additions and deletions)? I guess clarity of ordering? Maybe we just declare if deletes are applied first. Solves the ordering problem without adding extra complexion here.
</blockquote></details>

removed modification object to reduce complexion

rymurr:
Updated based on comments. Thanks a lot @jacques-n 

There is a fair bit of noise in reviewable as I dismissed stale comments, sorry for that.

jacques-n (dismissed):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 60 files reviewed, 21 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVP8G3dDV5BqHn9VKNF:-MVP8G3dDV5BqHn9VKNG:b-rtqx0m) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L75)):*
> ```Java
>     public static Type fromPayload(Byte payload) {
>       if (payload == null || payload > 5 || payload < 0) {
>         throw new IllegalArgumentException(String.format("Cannot create type from payload. Payload %s does not exist", payload));
> ```

I don't think %s works for a byte. You probably need %d. I thought findbugs or similar would find this (or am I wrong?).

laurentgo (commented):
> Reviewed 1 of 37 files at r3.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 1 of 60 files reviewed, 22 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 86 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAE4U0E9haye1gVMi:-MVS-7n86y0NrAkwoBFn:bvy7gha) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/Contents.java#L86)):*

Done. Stale
</blockquote></details>

why isn't the method public? is it because we should rely on the class itself? (there are already multiple classes in Nessie which relies on type methods vs class instanceof)
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 57 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVRyvUr7ZbngUPbyRNn:-MVRyvUr7ZbngUPbyRNo:b-41eh7y) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L57)):*
> ```Java
> 
>   public static enum Type {
>     UNKNOWN((byte) 0),
> ```

For now we decided that model and storage would use 2 differents representations, so IMHO the byte id should be encoded into the storeworker, and not on this class. If this is too late to address in this patch, let's address it a new pull request...
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 74 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVRya8j4HgjX-Bc0BW9:-MVRya8j4HgjX-Bc0BWA:bk1n0ex) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L74)):*
> ```Java
>      */
>     public static Type fromPayload(Byte payload) {
>       if (payload == null || payload > 5 || payload < 0) {
> ```

It should use `Type.values().length`
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVP8G3dDV5BqHn9VKNF:-MVRyIdLCd1g5mSTwcsx:bqx6n7b) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L75)):*

I don't think %s works for a byte. You probably need %d. I thought findbugs or similar would find this (or am I wrong?).
</blockquote></details>

I would expect `String.format` to use `String.valueOf(Object)`, so that's why ErrorProne might not flag it.
___
*[model/src/main/java/org/projectnessie/model/DeltaLakeTable.java, line 42 at r1](https://reviewable.io/reviews/projectnessie/nessie/866#-MUPAYazCFN90InpZLdA:-MVS00gLAgQ5__srZmfq:b-2oyfuf) ([raw file](https://github.com/projectnessie/nessie/blob/82e8e1e10528c5db6d381f961e07384f53c04e6c/model/src/main/java/org/projectnessie/model/DeltaLakeTable.java#L42)):*

Done. Stale
</blockquote></details>

tbh I don't see the difference really about having a field exposed by a method in a parent class, and a method being implemented by each subclass returning a constant...

At least the method should be final.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 76 files reviewed, 22 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 57 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVRyvUr7ZbngUPbyRNn:-MVWrjunCzb9L9hqaS-O:b-67d5vg) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L57)):*

For now we decided that model and storage would use 2 differents representations, so IMHO the byte id should be encoded into the storeworker, and not on this class. If this is too late to address in this patch, let's address it a new pull request...
</blockquote></details>

I just pushed a change which attempts to put all the payload related stuff in the store worker. It isn't great but let me know what you think
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 74 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVRya8j4HgjX-Bc0BW9:-MVWnZz-8nSOFtZFJc68:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L74)):*

It should use `Type.values().length`
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 75 at r3](https://reviewable.io/reviews/projectnessie/nessie/866#-MVP8G3dDV5BqHn9VKNF:-MVWncKo5VOxkXz6sWBv:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/098b37c1def761f7b5c7c9354bc894abbbd69298/model/src/main/java/org/projectnessie/model/Contents.java#L75)):*

I would expect `String.format` to use `String.valueOf(Object)`, so that's why ErrorProne might not flag it.
</blockquote></details>

Done.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 76 files reviewed, 18 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java, line 148 at r4](https://reviewable.io/reviews/projectnessie/nessie/866#-MVZmgDw7m-7DapVA5UO:-MVZmgDx0zUc9M2xX4fh:bqrdwht) ([raw file](https://github.com/projectnessie/nessie/blob/266103271887a529e9343bcf49645adecb3386ba/servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java#L148)):*
> ```Java
>     public Byte getPayload(Contents value) {
>       if (value instanceof IcebergTable) {
>         return 1;
> ```

since we are using the ordinals below (which might be an issue if values are reordered, we might as well use them here (better be symbolic)

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 76 files reviewed, 18 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java, line 148 at r4](https://reviewable.io/reviews/projectnessie/nessie/866#-MVZmgDw7m-7DapVA5UO:-MV_WltYEEkbe002M8iW:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/266103271887a529e9343bcf49645adecb3386ba/servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java#L148)):*

since we are using the ordinals below (which might be an issue if values are reordered, we might as well use them here (better be symbolic)
</blockquote></details>

Done.

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 76 files reviewed, 20 unresolved discussions (waiting on @jacques-n, @laurentgo, @rymurr, and @snazy)

___
*[versioned/gc/spark/src/test/java/org/projectnessie/versioned/gc/GcTestUtils.java, line 82 at r5](https://reviewable.io/reviews/projectnessie/nessie/866#-MVvdePS1ppx-YRqUDR2:-MVvdePT-9kU5_aBF9mG:b-kytidg) ([raw file](https://github.com/projectnessie/nessie/blob/bcc0889021d6870e4f03d1f50502260a2b525bf6/versioned/gc/spark/src/test/java/org/projectnessie/versioned/gc/GcTestUtils.java#L82)):*
> ```Java
>       return delegate.fromBytes(bytes);
>     }
> ```

Pls remove this change (it's just adding a newline)
___
*[versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java, line 447 at r5](https://reviewable.io/reviews/projectnessie/nessie/866#-MVvcbEO0i60v2jYjCms:-MVvcbEO0i60v2jYjCmt:b686gwf) ([raw file](https://github.com/projectnessie/nessie/blob/bcc0889021d6870e4f03d1f50502260a2b525bf6/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java#L447)):*
> ```Java
>           return !deleted.contains(key);
>         })
>         .filter(x -> x instanceof Put)
> ```

Nit: `.filter(Put.class:isInstance)` (or like that)

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/866)* status: 0 of 76 files reviewed, 19 unresolved discussions (waiting on @jacques-n, @laurentgo, and @snazy)

___
*[versioned/gc/spark/src/test/java/org/projectnessie/versioned/gc/GcTestUtils.java, line 82 at r5](https://reviewable.io/reviews/projectnessie/nessie/866#-MVvdePS1ppx-YRqUDR2:-MW-4MZp3gflQUTrBCwc:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/bcc0889021d6870e4f03d1f50502260a2b525bf6/versioned/gc/spark/src/test/java/org/projectnessie/versioned/gc/GcTestUtils.java#L82)):*

Pls remove this change (it's just adding a newline)
</blockquote></details>

Done.
___
*[versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java, line 447 at r5](https://reviewable.io/reviews/projectnessie/nessie/866#-MVvcbEO0i60v2jYjCms:-MW-4_Fg1UwnaeN9C5zP:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/bcc0889021d6870e4f03d1f50502260a2b525bf6/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java#L447)):*

Nit: `.filter(Put.class:isInstance)` (or like that)
</blockquote></details>

Done.

rymurr:
Thanks for the comments @snazy and @laurentgo have addressed them in the recent commit.

rymurr:
@snazy and @laurentgo any more comments?

snazy (approved):
LGTM

PR #894, opened 2021-03-03
Update expression visitor

stevelorddremio:
This is a precursor to the RocksDB and MongoDB backing store implementations. The UpdateClauseVisitor is used by both and is similar in concept to the existing ValueVisitor interface.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/894)


rymurr (approved):
LGTM. Will wait for a bit before merging to see if anyone else wants to weigh in

PR #910, opened 2021-03-05
Less log spam during tests

snazy:
Currently a ton of rather trace and debug information is printed to the console for every run of
`./mvnw test integration-test`. This PR removes most of that output, so it's possible to scroll
back in the console to see relevant output.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/910)


rymurr (commented):
How can this be changed at runtime? I personally use these logs quite regularly

snazy:
`-Dtest.log.level=INFO` or so. Defaults to `WARN`.

rymurr:
I think I would prefer to keep `INFO` I use info logs from maven quite a bit. I don't mind putting them down to warn on the CI build though.

rymurr (approved):

PR #913, opened 2021-03-05
Tweak logging and don't log HTTP accesses to the console

snazy:
* Prepares but doesn't enabel file-level logging
* Adds jaeger-tracing IDs to the log file
* Effectively disable HTTP access log to be able to see errors

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/913)


rymurr (commented):
I dont know if the default should write to file. Raises cleanup issues. I also like the `INFO` level on teh command line

snazy:
> I dont know if the default should write to file.
No, not by default (you still have to set `quarkus.log.file.enable=true`). That's why it's not enabled by default, just prepared the pattern. 

> I also like the INFO level on teh command line
Made it default to `INFO`, but configurable via an env var

rymurr (approved):

PR #915, opened 2021-03-05
Fix handling of unsaved commits in getUpdateState()

snazy:
From #813, (this comment](https://github.com/projectnessie/nessie/issues/813#issuecomment-779817450) :

From `TieredVersionStore.commit`, in the `store.load` for current+expected, `TieredVersionStore.ensureValidL1` is called, which calls `InternalBranch.getUpdateState`, which performs a loop over the `unsavedCommits`: it's `lastL1` is initialized to `lastSavedL1` and `lastL1` is "moved iteratively forward in history" via `lastL1.getChildWithTree(c.commit, tree, c.keyMutationList).withCheckpointAsNecessary(store)`.
That `withCheckpointAsNecessary()` occasionally throws a `(Reference)NotFoundException` when it tries to load an unsaved L1 via the `HistoryRetriever`, causing the commit-operation to fail.
I'm not sure whether that should distinguish between saved and unsaved commits or whether it's okay to only call `withCheckpointAsNecessary()` for the `lastSavedL1` in `getUpdateState`:
```
class InternalBranch extends InternalRef {
...
  public UpdateState getUpdateState(Store store)  {
...
    for (Commit c : unsavedCommits) {
      for (UnsavedDelta delta : c.deltas) {
        tree = delta.apply(tree);
      }

      lastL1 = lastL1.getChildWithTree(c.commit, tree, c.keyMutationList);
      if (lastL1 == lastSavedL1) {
        lastL1 = lastL1.withCheckpointAsNecessary(store);
      }

      toSave.add(EntityType.L1.createSaveOpForEntity(lastL1));
...
```

I went with the latter (only call `withCheckpointAsNecessary()` for the `lastSavedL1` in `getUpdateState`) and ran into no more errors during my test.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915)

### DIFF
--- a/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java
+++ b/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.Spliterators;
@@ -44,44 +45,62 @@ class HistoryRetriever {
   private final Store store;
   private final InternalL1 start;
   private final Id end;
+  private final Map<Id, InternalL1> unsavedL1s;
 
-  public HistoryRetriever(Store store, InternalL1 start, Id end, boolean retrieveL1, boolean retrieveCommit, boolean includeEndEmpty) {
-    super();
+  public HistoryRetriever(Store store, InternalL1 start, Id end, boolean retrieveL1, boolean retrieveCommit,
+      boolean includeEndEmpty) {
+    this(store, start, end, retrieveL1, retrieveCommit, includeEndEmpty, Collections.emptyMap());
+  }
+
+  public HistoryRetriever(Store store, InternalL1 start, Id end, boolean retrieveL1, boolean retrieveCommit,
+      boolean includeEndEmpty, Map<Id, InternalL1> unsavedL1s) {
     this.store = store;
     this.start = start;
     this.end = end;
     this.retrieveL1 = retrieveL1;
     this.retrieveCommit = retrieveCommit;
     this.includeEndEmpty = includeEndEmpty;
+    this.unsavedL1s = unsavedL1s;
   }
 
-  class HistoryItem {
+  static class HistoryItem {
 
-    private Id id;
+    private final Id id;
     private InternalL1 l1;
     private InternalCommitMetadata commitMetadata;
 
-    public HistoryItem(Id id) {
+    HistoryItem(Id id) {
       this.id = id;
     }
 
-    public InternalL1 getL1() {
+    InternalL1 getL1() {
       return l1;
     }
 
-    public Id getId() {
+    private void setL1(InternalL1 l1) {
+      this.l1 = l1;
+    }
+
+    private Id getL1MetadataId() {
+      return l1.getMetadataId();
+    }
+
+    Id getId() {
       return id;
     }
 
-    public InternalCommitMetadata getMetadata() {
+    InternalCommitMetadata getMetadata() {
       return commitMetadata;
+    }
+
+    private void setMetadata(InternalCommitMetadata commitMetadata) {
+      this.commitMetadata = commitMetadata;
     }
 
     @Override
     public String toString() {
       return "HistoryItem [id=" + id + ", l1=" + l1 + ", commitMetadata=" + commitMetadata + "]";
     }
-
   }
 
   Stream<HistoryItem> getStream() {
@@ -151,14 +170,19 @@ class HistoryRetriever {
         final HistoryItem item = new HistoryItem(parent);
         items.add(item);
         if (retrieveL1 || retrieveCommit || lastInList) {
-          loadOps.load(EntityType.L1, InternalL1.class, parent, l1 -> item.l1 = l1);
+          InternalL1 unsaved = unsavedL1s.get(parent);
+          if (unsaved != null) {
+            item.setL1(unsaved);
+          } else {
+            loadOps.load(EntityType.L1, InternalL1.class, parent, item::setL1);
+          }
         }
 
         if (retrieveCommit && !parent.equals(InternalL1.EMPTY_ID)) {
           secondOps.loadDeferred(EntityType.COMMIT_METADATA,
               InternalCommitMetadata.class,
-              () -> item.l1.getMetadataId(),
-              cmd -> item.commitMetadata = cmd);
+              item::getL1MetadataId,
+              item::setMetadata);
         }
 
         if (parent.equals(end)) {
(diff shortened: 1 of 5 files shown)


jacques-n (changes requested):
My initial response is: 
* we're addressing this in the wrong way.
* we need some tests for this.

I believe your change will result in a situation where checkpoints may come late. I think the correct result is that the history retriever is constructed using a combination of existing L1s as well as those that are pending save. The process is this:

Imagine we have 

L1.0 (saved)
L1.1 (unsaved)
L1.2 (unsaved)

When we start, we create an updatestate for L1.1. In that we try to retrieve L1.0 which saved and succeed. 

The problem comes when we then step to L1.2. In L1.2 the HistoryRetriever attempts to load L1.1 but we're currently still in the getUpdateState loop and so L1.1 is pending save but hasn't actually been save. This is by design, we want to batch up all the L1s that need to be saved and save them at once. The problem is that the HistoryRetriever is going direct ot the Store rather than also seeing that the current "transaction" does have the L1.1 value. We need to rework this path so that the HistoryRetriever can also access the in-flight L1.1 value and not fail. 

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915#-:-MVDWeftCmG0vxyriabT:b-4144wm)* status: 0 of 1 files reviewed, all discussions resolved

snazy:
Added a commit to this PR that "injects" the unsaved-L1s into the HistoryRetriever. If that approach feels legit, I can add a few tests for it.

jacques-n (dismissed):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java, line 55 at r2](https://reviewable.io/reviews/projectnessie/nessie/915#-MVNHKG-9bbWS-uEeHtu:-MVNHKG-9bbWS-uEeHtv:b1wuek8) ([raw file](https://github.com/projectnessie/nessie/blob/893a80a6cfff01612d43084194d5c26b9ba65aa8/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java#L55)):*
> ```Java
>   }
> 
>   public HistoryRetriever(Store store, InternalL1 start, Id end, boolean retrieveL1, boolean retrieveCommit,
> ```

While I think the spirit is good, I think you should create a new decorating Store that can hold inflight items rather than adding extra complexity here. This would be a precursor to a CachingStore that can cache reads for all known "immutable" objects. This should make the change much less bespoke/cross-cutting.

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @jacques-n)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java, line 55 at r2](https://reviewable.io/reviews/projectnessie/nessie/915#-MVNHKG-9bbWS-uEeHtu:-MVWkSz7DRSMaNXIAMb4:bju08n) ([raw file](https://github.com/projectnessie/nessie/blob/893a80a6cfff01612d43084194d5c26b9ba65aa8/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/HistoryRetriever.java#L55)):*

While I think the spirit is good, I think you should create a new decorating Store that can hold inflight items rather than adding extra complexity here. This would be a precursor to a CachingStore that can cache reads for all known "immutable" objects. This should make the change much less bespoke/cross-cutting.
</blockquote></details>

I've played around with a cache for immutable objects and perf-tested. For "classic JVMs", there's clearly a benefit. But for native-images it results in actually quite worse results, namely: ~1 second stop-the-world GC times (community Graal-native-images only provide the serial-GC). So I don't think that a cache is a good idea, since our primary runtime shall be a Graal native image, preferably a lambda. For this particular issue, I think that adding an layer of a per-version-store-request `Store` implementation between the `TieredVersionStore` and the actual `Store` implementation adds a lot of complexity and more of heap pressure, which is again not great for Serial-GC.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jacques-n and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java, line 260 at r3](https://reviewable.io/reviews/projectnessie/nessie/915#-MWA1kPy4LTFBM5A0pHO:-MWA1kPy4LTFBM5A0pHP:b972gks) ([raw file](https://github.com/projectnessie/nessie/blob/c91f3a810a2679399485614ec04a42ec9dbe51b0/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/KeyList.java#L260)):*
> ```Java
> 
>     @Override
>     public Optional<KeyList> createCheckpointIfNeeded(InternalL1 startingPoint, Store store, Map<Id, InternalL1> unsavedL1s) {
> ```

I don't love passing this into the KeyList everywhere, but i understand there isn't much of a way around it. I can't think of much better. Anything that might be > this approach but < full blown cache that can be done here?
___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractITTieredVersionStore.java, line 331 at r3](https://reviewable.io/reviews/projectnessie/nessie/915#-MWA1UcRBmRq5fPIJ6Ot:-MWA1UcRBmRq5fPIJ6Ou:b-7yxj3p) ([raw file](https://github.com/projectnessie/nessie/blob/c91f3a810a2679399485614ec04a42ec9dbe51b0/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractITTieredVersionStore.java#L331)):*
> ```Java
>   }
> 
>   private InternalBranch simulateCommit(InternalRefId ref, int num) {
> ```

This is effectively a copy of the commit routine in TVS right? maybe a javadoc to say why the copy is here?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/915)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jacques-n, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractITTieredVersionStore.java, line 331 at r3](https://reviewable.io/reviews/projectnessie/nessie/915#-MWA1UcRBmRq5fPIJ6Ot:-MWA3Y4U744n3EeNkNLC:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/c91f3a810a2679399485614ec04a42ec9dbe51b0/versioned/tiered/tiered-tests/src/main/java/org/projectnessie/versioned/impl/AbstractITTieredVersionStore.java#L331)):*

This is effectively a copy of the commit routine in TVS right? maybe a javadoc to say why the copy is here?
</blockquote></details>

Done.

rymurr (approved):

PR #916, opened 2021-03-05
Configurable DynamoDB read/write capacity units + billing-mode

snazy:
Adds configuration options to make the following options for tables created by the `DynamoDBStore` configurable:

* read and write capacity units configurable
* billing-mode configurable
* change default billing-mode to `PAY_PER_REQUEST`

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/916)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/config/DynamoVersionStoreConfig.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/config/DynamoVersionStoreConfig.java
@@ -19,12 +19,19 @@ import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.projectnessie.versioned.dynamodb.DynamoStoreConfig;
 
 import io.quarkus.arc.config.ConfigProperties;
+import io.quarkus.runtime.annotations.RegisterForReflection;
 
 /**
  * DynamoDB version store configuration.
  */
 @ConfigProperties(prefix = "nessie.version.store.dynamo")
 public interface DynamoVersionStoreConfig {
+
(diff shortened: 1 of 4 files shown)


rymurr (commented):
I dont love this. IMHO we shouldn't be creating and managing cloud resources. We create a basic table for testing but in prod we should expect people to have terraform, CFT, etc to configure their own resources.

snazy:
Hm. Expecting all users to setup Terraform/CloudFormation every time is a high bar for people to get started. Sure, you need "production recommendations", but it's easier if people can just start Nessie and it just creates the tables, so users can start using Nessie quicker.
A proper prod setup should have the "check and create tables" option disabled, no disagreement here.

rymurr:
Agreed, which is why we have a basic create function. Past simple create test sized tables I don't think we should give more choice.

jacques-n (commented):
I generally agree with Ryan here. (We actually discussed this when I added the original code here :)

I think there are two different things being discussed here:
* Have optimal defaults
* Have configurable creation

I think we should focus on the first and people can use terraform for the second. If pay per request is our default recommendation, let's set that when we create the table (along with anything else that is appropriate).

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/916#-:-MVDaGVO0vBW771yZYFC:b1arbvh)* status: 0 of 4 files reviewed, all discussions resolved

snazy:
Sounds good to me. Updated the PR to default to `PAY_PER_REQUEST` and removed the rest.

rymurr (approved):

PR #917, opened 2021-03-05
Make TieredVersionStore configurable

snazy:
Adds a configuration for the `TieredVersionStore` that exposes options to:
* configure wait-on-collapse
* configure commit-retry-count
* configure p2 (collapse) retry-count

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/917)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/config/TieredVersionStoreConfig.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/config/TieredVersionStoreConfig.java
@@ -20,9 +20,6 @@ import static org.projectnessie.versioned.impl.TieredVersionStoreConfig.DEFAULT_
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 
-import io.quarkus.arc.config.ConfigProperties;
-
-@ConfigProperties(prefix = "nessie.version.store.tiered")
 public interface TieredVersionStoreConfig {
 
   @ConfigProperty(name = "commit-retry-count", defaultValue = DEFAULT_COMMIT_RETRY_COUNT)
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/config/TieredVersionStoreConfig.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/config/TieredVersionStoreConfig.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server.config;
+
+import static org.projectnessie.versioned.impl.TieredVersionStoreConfig.DEFAULT_COMMIT_RETRY_COUNT;
+import static org.projectnessie.versioned.impl.TieredVersionStoreConfig.DEFAULT_P2_COMMIT_RETRY_COUNT;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import io.quarkus.arc.config.ConfigProperties;
+
+@ConfigProperties(prefix = "nessie.version.store.tiered")
+public interface TieredVersionStoreConfig {
+
+  @ConfigProperty(name = "commit-retry-count", defaultValue = DEFAULT_COMMIT_RETRY_COUNT)
+  int commitRetryCount();
+
+  @ConfigProperty(name = "p2-commit-retry-count", defaultValue = DEFAULT_P2_COMMIT_RETRY_COUNT)
+  int p2CommitRetryCount();
+}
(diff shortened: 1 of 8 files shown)


servers/quarkus-server/src/main/java/org/projectnessie/server/config/TieredVersionStoreConfig.java
```diff
@@ -0,0 +1,33 @@
...
+package org.projectnessie.server.config;
+
+import static org.projectnessie.versioned.impl.TieredVersionStoreConfig.DEFAULT_COMMIT_RETRY_COUNT;
+import static org.projectnessie.versioned.impl.TieredVersionStoreConfig.DEFAULT_P2_COMMIT_RETRY_COUNT;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import io.quarkus.arc.config.ConfigProperties;
+
+@ConfigProperties(prefix = "nessie.version.store.tiered")
```
laurentgo, line 25: why not making this a super interface of each tiered store backend instead so that you could use different values for different implementations

snazy: You mean like in the commit I've just pushed?

laurentgo: Yes, and I guess that `ConfigProperties` is not needed anymore?

snazy: yea, true, removed that

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStoreConfig.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.impl;
+
+import org.immutables.value.Value.Default;
+import org.immutables.value.Value.Immutable;
+import org.projectnessie.versioned.impl.ImmutableTieredVersionStoreConfig.Builder;
+
+@Immutable
+public interface TieredVersionStoreConfig {
```
jacques-n, line 23: I generally don't think these should be end user configurable. I suggest we come up with things that should be configured versus those that generally shouldn't be and namespace them separately. Additionally, this exposes some internal detail. For example, I think we should actually adjust so we only collapse on save on every N commits or time, which makes the first option here a bit misleading.

snazy: Yea, concurrent commits and especially the kicked-off collapses are tricky. The collapses can be synchronized (i.e. only one collapse per branch), which already works fine. I implemented a mechanism to ensure that only one collapse per Nessie instance is running for each branch for perf-testing (the `Serialize collapse-branch-commits per branch` commit in my `trace-commit` branch + a follow-up that eliminates some more read operations in the commit-code-path). The retry-mechanism(s) used by commit + collapse should have reasonable defaults, but IMO should be tunable, maybe with a configurable retry/back-off strategy - but it's a complicated thing to tune and requires a lot of in-depth knowledge. I'd like to have those at least for now, but maybe prefix those with `internal` or so. WDYT?

jacques-n (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/917)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStoreConfig.java, line 23 at r2](https://reviewable.io/reviews/projectnessie/nessie/917#-MV1KIgxdkFTFf5PxRWb-r2-23:-MVP5vpiAfhO7xOX95a5:b-aybu6c) ([raw file](https://github.com/projectnessie/nessie/blob/c24ab807f90a31f8bc14a08c967af192c5ca6eab/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStoreConfig.java#L23)):*

Yea, concurrent commits and especially the kicked-off collapses are tricky. The collapses can be synchronized (i.e. only one collapse per branch), which already works fine. I implemented a mechanism to ensure that only one collapse per Nessie instance is running for each branch for perf-testing (the `Serialize collapse-branch-commits per branch` commit in my `trace-commit` branch + a follow-up that eliminates some more read operations in the commit-code-path). The retry-mechanism(s) used by commit + collapse should have reasonable defaults, but IMO should be tunable, maybe with a configurable retry/back-off strategy - but it's a complicated thing to tune and requires a lot of in-depth knowledge. I'd like to have those at least for now, but maybe prefix those with `internal` or so. WDYT?
</blockquote></details>

I'm fine with internal. What do other people think?

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStoreConfig.java
rymurr: > but it's a complicated thing to tune and requires a lot of in-depth knowledge

This to me means they shouldn't be exposed as parameters almost by definition. if I had a dollar for every time I have seen exposing internal details as config result in poor performance. bugs, correctness errors and generally pain...

an `internal` prefix won't discourage or even slow most people down. The best speed bump i can think of is to make people recompile. Which given the likelihood of people needing to change these parameters and the type of people that would change them isn't to me such a big deal.

snazy:
After playing around with the perf-test-related stuff, it turns out that it's not worth to expose these configurables anymore.

Initially, I had ideas to add some configurable retry-implementations/backoff-strategies for commit + collapse. But the combination made it actually much harder to reason about, especially compared to the potential "win", which turned out to be quite low, rather "not measurable" but easy to get wrong. The spark that turned into all that was the question how to reduce the number of retries (or: "produce" less update-collisions).

Therefore I'm closing this

PR #928, opened 2021-03-09
Pimp gradle plugin

snazy:
Three tiny commits:

* Pull the project version from the pom in the parent directory, if `version.txt` is missing
* Pull the Quarkus version from the Nessie top-level pom
* Set the properties as system properties (hello Quarkus 1.10....)

### DIFF
--- a/tools/apprunner-gradle-plugin/build.gradle
+++ b/tools/apprunner-gradle-plugin/build.gradle
@@ -4,7 +4,11 @@ plugins {
   id 'com.gradle.plugin-publish' version '0.12.0'
 }
 
-version = file('version.txt').text.trim()
+if (!file('version.txt').exists()) {
+  version = file('../pom.xml').readLines().find {l -> l.contains('<version>') }.replaceFirst(~/.*<version>(.*)<\/version>.*/, '$1' )
+} else {
+  version = file('version.txt').text.trim()
+}
 
 allprojects {
   group = "org.projectnessie"
--- a/tools/apprunner-gradle-plugin/build.gradle
+++ b/tools/apprunner-gradle-plugin/build.gradle
@@ -4,6 +4,7 @@ plugins {
   id 'com.gradle.plugin-publish' version '0.12.0'
 }
 
+def quarkusVersion = file('../../pom.xml').readLines().find {l -> l.contains('<quarkus.version>') }.replaceFirst(~/.*<quarkus.version>(.*)<\/quarkus.version>.*/, '$1' )
 if (!file('version.txt').exists()) {
   version = file('../pom.xml').readLines().find {l -> l.contains('<version>') }.replaceFirst(~/.*<version>(.*)<\/version>.*/, '$1' )
 } else {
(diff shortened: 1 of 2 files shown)


tools/apprunner-gradle-plugin/build.gradle
```diff
@@ -4,7 +4,12 @@ plugins {
   id 'com.gradle.plugin-publish' version '0.12.0'
 }
 
-version = file('version.txt').text.trim()
+def quarkusVersion = file('../../pom.xml').readLines().find {l -> l.contains('<quarkus.version>') }.replaceFirst(~/.*<quarkus.version>(.*)<\/quarkus.version>.*/, '$1' )
+if (!file('version.txt').exists()) {
+  version = file('../pom.xml').readLines().find {l -> l.contains('<version>') }.replaceFirst(~/.*<version>(.*)<\/version>.*/, '$1' )
```
laurentgo, line 9: regex for parsing xml make me sad :(

tools/apprunner-gradle-plugin/build.gradle
```diff
@@ -4,7 +4,12 @@ plugins {
   id 'com.gradle.plugin-publish' version '0.12.0'
 }
 
-version = file('version.txt').text.trim()
+def quarkusVersion = file('../../pom.xml').readLines().find {l -> l.contains('<quarkus.version>') }.replaceFirst(~/.*<quarkus.version>(.*)<\/quarkus.version>.*/, '$1' )
```
laurentgo, line 7: How does this work for plugin users?

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
```diff
@@ -30,6 +30,7 @@
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collections;
```
laurentgo, line 33: I'm not sure I see where this is used...

tools/apprunner-gradle-plugin/build.gradle
snazy: Users don't use this build.gradle (it's just for building and publishing the plugin), they just "apply" the published plugin.

tools/apprunner-gradle-plugin/build.gradle
snazy: TIL there's `mvn help:evaluate` - added a commit to this PR

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
snazy: oops

tools/apprunner-gradle-plugin/build.gradle
```diff
@@ -4,7 +4,22 @@ plugins {
...
 }
 
-version = file('version.txt').text.trim()
+def quarkusVersionBuffer = new ByteArrayOutputStream()
+exec {
+  commandLine = ['../../mvnw', 'help:evaluate', '-Dexpression=quarkus.version', '-q', '-DforceStdout', '-f', '../../pom.xml']
+  standardOutput = quarkusVersionBuffer
+}.assertNormalExitValue()
+
+if (!file('version.txt').exists()) {
```
rymurr, line 13: I would be tempted to just remove the `version.txt` business. I think the module below is supposed to do it but if we are already doing it for quarkus we might as well do it for the nessie version

PR #935, opened 2021-03-11
Paging for get-commit-log

snazy:
Implements paging for the `TreeApi.getCommitLog()`, also updates only the API for `TreeApi.getEntries()`.

Paging for the commit log is implemented purely on the REST endpoint. Since `VersionStore.getCommits()`
already returns a `Stream`, we can just `Stream.limit()` the result. The implementation is quite dumb,
it requests "one more" entry from the version-store and if it sees that "one more" entry, it sets
`hasMore=true` and `token=hash-of-that-one-more-thing`, so the next page starts at the given token.

Generally speaking, clients and users of the API _must not_ interpret the token in any way, but just
treat it as some opaque thing. The API docs now include a sentence that the structure/format of the
token can change without prior notice.

`TreeApi.getEntries()` got the same two new parameters that `TreeApi.getCommitLog()` got, just to make
the API consistent. There is no paging implementation for the entries/keys yet.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/935)


model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -144,9 +218,46 @@ LogResponse getCommitLog(
...
+   * Default implementation to return a stream of objects for a ref, functionally equivalent to
+   * calling {@link #getEntries(String, Integer, String)} with manual paging.
+   * <p>The {@link Stream} returned by {@code getEntriesStream(ref, OptionalInt.empty())},
+   * if not limited, returns all commit-log entries.</p>
+   *
+   * @param ref a named reference (branch or tag name) or a commit-hash
+   * @param pageSizeHint page-size hint for the backend
+   * @return stream of {@link Entry} objects
+   */
+  default Stream<Entry> getEntriesStream(@NotNull String ref, OptionalInt pageSizeHint) throws NessieNotFoundException {
```
rymurr, line 241: I am not sure why these are in the model. To me they are client only and should be there

snazy: It's a double-edged sword IMO. The model defines the paging properties via `PaginatedResponse` but doesn't define whether paging is only "legit" via HTTP or whether another tier can actually enforce paging. `NessieClient` "only" exposes the `TreeApi` interface, so we'd either have to add a new `TreeWithConvenienceApi` in `nessie-client` or some class with static utility methods. Both approaches feel a bit odd to me.

rymurr: I actually, personally, don't mind the convenience api. Primarily because the contents tree has a few 'wasteful' apis: delete and put single contents. I thikn there is an argument that the direct jax-rs api shouldn't be exposed, rather act as a delegate so the client can continue to grow convenience methods.

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/935)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @snazy)

___
*[model/src/main/java/org/projectnessie/api/StreamingUtil.java, line 32 at r2](https://reviewable.io/reviews/projectnessie/nessie/935#-MW-9A7jCsJo29OenarN:-MW-9A7jCsJo29OenarO:b2trvwa) ([raw file](https://github.com/projectnessie/nessie/blob/091357e7c813edbbf5d728769757d39ac0e3ddaf/model/src/main/java/org/projectnessie/api/StreamingUtil.java#L32)):*
> ```Java
>  * Helper and utility methods around streaming of {@link TreeApi} et al.
>  */
> public final class StreamingUtil {
> ```

Shoudln't this move to the client?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/935)* status: 0 of 14 files reviewed, 2 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/api/StreamingUtil.java, line 32 at r2](https://reviewable.io/reviews/projectnessie/nessie/935#-MW-9A7jCsJo29OenarN:-MW-nT9t9kzOPIN-GljE:bbicati) ([raw file](https://github.com/projectnessie/nessie/blob/091357e7c813edbbf5d728769757d39ac0e3ddaf/model/src/main/java/org/projectnessie/api/StreamingUtil.java#L32)):*

Shoudln't this move to the client?
</blockquote></details>

Moved - now the build complains that TreeApi + ContentsApi do not have enough javadoc.......

rymurr (approved):

PR #938, opened 2021-03-11
Handle RuntimeException + DynamoDbException in DynamoStore

snazy:
* Handles the `*LimitExceeded`-exceptions from DynamoDB and wraps those in a `BackendLimitExceededException`
* Log unhandled exceptions in `TieredVersionStore`
* Does not log unhandled exceptions in `DynamoStore`, as that's done by the `TieredVersionStore`

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)


rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 145 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsJVk1jxoJGDyU3iP:-MVpsJVk1jxoJGDyU3iQ:bkts62y) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L145)):*
> ```Java
>   @Nonnull
>   @VisibleForTesting
>   static RuntimeException unhandledException(String operation, Throwable e) {
> ```

what is the difference between these exceptions (practically speaking) and why are we removing the extra context? eg what happens if we want torespond differently to limit exceeded vs provisioned throughput exceeded?
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 155 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsxHm5Yho0Ox7veV3:-MVpsxHm5Yho0Ox7veV4:b-e45k12) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L155)):*
> ```Java
>       return (StoreException) e;
>     } else {
>       return new StoreOperationException(String.format("Failure during %s", operation), e);
> ```

its not clear to me that eveyrthing should be wrapped in a store operation exception. Eg `IndexOutOfBoundException` or any of the many `java.lang` runtime exceptions
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 281 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVptn253ltKPVZn44of:-MVptn253ltKPVZn44og:b-22qlo6) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L281)):*
> ```Java
>       }
>     } catch (RuntimeException e) {
>       throw unhandledException("load", e);
> ```

why not follow the `guava` style of returning `void` and throwing from the method? 
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MVpuTXD9LaZPnofRt7Q:bqp3cdn) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*
> ```Java
>   @Nonnull
>   @VisibleForTesting
>   static RuntimeException unhandledException(String operation, Throwable e) {
> ```

I am not sure I understand what this is doing. We only call this method when catching runtime exceptions. So these are all goign to be just rethrown right? The last line will never get called? Or is everything getting caught and we are invalidating all the checked exception annotations on these methods? I think we want to keep the checked exceptions and I don't see the point this method if we are only catching unchecked.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @rymurr and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MVsKJuu8daXRSo6AsD4:bju5phg) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*

I am not sure I understand what this is doing. We only call this method when catching runtime exceptions. So these are all goign to be just rethrown right? The last line will never get called? Or is everything getting caught and we are invalidating all the checked exception annotations on these methods? I think we want to keep the checked exceptions and I don't see the point this method if we are only catching unchecked.
</blockquote></details>

Design-wise, I'm not convinced by that approach of mutualizing all the exception code, and how it is an improvement. Can you go over and explain why you chose this approach over doing the exception management in the specific code instead? For example, since the method only handles a Throwable, it is not possible to leverage the context of the call as maybe one exception might be a fatal in one case but retryable in a second case.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 157 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMA0-1Lf7Vm-Gmu4M:-MVsMA009ODUzHBKhdYQ:b4e84f8) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L157)):*
> ```Java
>         throw new ReferenceAlreadyExistsException("A branch or tag already exists with that name.");
>       }
>     } catch (IllegalArgumentException e) {
> ```

why is it okay to throw back `IAE` but not `RuntimException`? What's the logic?
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java, line 24 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMOKvC3JI8hoD4HrT:-MVsMOKvC3JI8hoD4HrU:bhiuj4j) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java#L24)):*
> ```Java
>  */
> @SuppressWarnings("serial")
> public class BackendLimitExceededException extends StoreException {
> ```

is it that productive to expose that exception to the upper layer? Wouldn't it be better for the store to leverage and throttle itself, or throw/exit if it is not doable?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 6 files reviewed, 6 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 145 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsJVk1jxoJGDyU3iP:-MVugw16ArYHNMtrVku5:b-blncfp) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L145)):*

what is the difference between these exceptions (practically speaking) and why are we removing the extra context? eg what happens if we want torespond differently to limit exceeded vs provisioned throughput exceeded?
</blockquote></details>

Not sure whether you actually can react differently and in the "right way", because the _how_ you want to react requires a lot more knowledge about the whole system (all clients, all Nessie servers, DynamoDB configuration, DynamoDB throttling state, etc). Backpressure is hard to implement right. IMO knowing that a request failed, because DynamoDB limits were hit is good enough for now.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 155 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsxHm5Yho0Ox7veV3:-MVubZCF68sHyS9LAHaq:bb4qf4h) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L155)):*

its not clear to me that eveyrthing should be wrapped in a store operation exception. Eg `IndexOutOfBoundException` or any of the many `java.lang` runtime exceptions
</blockquote></details>

It's only there to add at least some context (the name of the operation). Otherwise you'd be completely blind and only see the exception.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 281 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVptn253ltKPVZn44of:-MVubuaN6z5OzvT2Hef-:b-fz09t1) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L281)):*

why not follow the `guava` style of returning `void` and throwing from the method? 
</blockquote></details>

Not having to add a boilerplate `return something;` at the call site of `unhandledException()`
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MVugVhE9NsFIy6X0TSo:b-v3imod) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*

Design-wise, I'm not convinced by that approach of mutualizing all the exception code, and how it is an improvement. Can you go over and explain why you chose this approach over doing the exception management in the specific code instead? For example, since the method only handles a Throwable, it is not possible to leverage the context of the call as maybe one exception might be a fatal in one case but retryable in a second case.
</blockquote></details>

The idea here is to just log _that_ something bad happened, because currently nothing is logged at all, which makes it extremely hard to investigate issues. Moved the logging and re-throw in TVS to the call-sites of this method and removed this one.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 157 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMA0-1Lf7Vm-Gmu4M:-MVuesiOAs3jeJOyP3T3:bt8ppyj) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L157)):*

why is it okay to throw back `IAE` but not `RuntimException`? What's the logic?
</blockquote></details>

Not sure why an `IllegalArgumentException` is thrown there. It's been there before and needs to be re-throws.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java, line 24 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMOKvC3JI8hoD4HrT:-MVuf9mT1iTzn5MXjBA7:bkjnxff) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java#L24)):*

is it that productive to expose that exception to the upper layer? Wouldn't it be better for the store to leverage and throttle itself, or throw/exit if it is not doable?
</blockquote></details>

That would require an in-depth knowledge of the actual DynamoDB configuration (provisioned reads/writes), its throttling state, the load caused by other users against other Nessie instances (lamdbas) sharing the same backend, etc.
I do not think that a (Version)Store implementation should do any throttling, because that usually leads to even more load and very likely more retries against an already overloaded backend.
Informing the client that Nessie cannot fulfill the request, because the backend store is overloaded is IMO all we can do. Backpressure is hard to implement right.

rymurr (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 145 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsJVk1jxoJGDyU3iP:-MVzyZQxCa8niFTZr17i:b-jivvan) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L145)):*

Not sure whether you actually can react differently and in the "right way", because the _how_ you want to react requires a lot more knowledge about the whole system (all clients, all Nessie servers, DynamoDB configuration, DynamoDB throttling state, etc). Backpressure is hard to implement right. IMO knowing that a request failed, because DynamoDB limits were hit is good enough for now.
</blockquote></details>

Im not sure I agree. Currently we do nothing to respond to throttling. This change just wraps aws throttle exceptions with our own exception. stores still ignore it and users still see 500 errors. I don't see the point in adding a wrapper if we aren't going to react to it.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 155 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsxHm5Yho0Ox7veV3:-MVzzCOVAYbZB2bFxWJQ:bpzau6c) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L155)):*

It's only there to add at least some context (the name of the operation). Otherwise you'd be completely blind and only see the exception.
</blockquote></details>

I don't understand why the context isn't in the stacktrace. I would expect the stacktrace to show the callsite that caused the operation and the exception msg taht caused it. Now I have to scroll to the `caused by` section of the stactrace.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 281 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVptn253ltKPVZn44of:-MVzzWQz4_jJurDI0j5v:b-ybj0ee) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L281)):*

Not having to add a boilerplate `return something;` at the call site of `unhandledException()`
</blockquote></details>

hmmmm, understood. this still makes me uncomfortable and feels fragile
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MW--cvk8jBF8HPBAZVv:bb4hgkh) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*

The idea here is to just log _that_ something bad happened, because currently nothing is logged at all, which makes it extremely hard to investigate issues. Moved the logging and re-throw in TVS to the call-sites of this method and removed this one.
</blockquote></details>

Im not 100% sure this is better. In my experience this is going to create a lot of noise in the log, lots of operations fail w/ the checked exceptions as part of a regular functioning Nessie. Its going to be hard to tell the difference. Also, imho the version store thorwing an exception should be logged by the layer above. It is already isn't it? That is where I would prefer to fix this. Or maybe the log level shoudl be debug?

If we log something as a 'warning' it should be clear and actionable by the user/operator. I don't think these are.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 157 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMA0-1Lf7Vm-Gmu4M:-MW-0wM26LkA6DectzAb:bbnvq7x) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L157)):*

Not sure why an `IllegalArgumentException` is thrown there. It's been there before and needs to be re-throws.
</blockquote></details>

IMHO i see no reason to treat `IAE` differently here.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java, line 24 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMOKvC3JI8hoD4HrT:-MW-1P6xEu45SIYpAnCl:b-ilkr9w) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java#L24)):*

That would require an in-depth knowledge of the actual DynamoDB configuration (provisioned reads/writes), its throttling state, the load caused by other users against other Nessie instances (lamdbas) sharing the same backend, etc.
I do not think that a (Version)Store implementation should do any throttling, because that usually leads to even more load and very likely more retries against an already overloaded backend.
Informing the client that Nessie cannot fulfill the request, because the backend store is overloaded is IMO all we can do. Backpressure is hard to implement right.
</blockquote></details>

I agree that backpressure is hard to get right but we should at some point make an effort do to so, not in this PR and hopefully above the level of Dynamo. 

However, this to me doesn't really alert the user. It wraps an AWS throttle exception. We still get a 500 error back to the client and the client still swallows it.

My concern w/ throttle messages from AWS is that excessive throttling may lead to inconsistent or unrecoverable states. I think there is something between a wrapper exception and a full blown backpressure algo that this PR could do: options may include simple throttle at the http level, client exception for throttling, different return codes etc

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 145 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsJVk1jxoJGDyU3iP:-MW3sqDKDgjN1teC6mx2:b-koljc0) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L145)):*

Im not sure I agree. Currently we do nothing to respond to throttling. This change just wraps aws throttle exceptions with our own exception. stores still ignore it and users still see 500 errors. I don't see the point in adding a wrapper if we aren't going to react to it.
</blockquote></details>

I'd leave that for a follow-up PR that updates the client-API and therefore also the rest interfaces. It also heavily depends on whether and how support for backpressure will be in Nessie.
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 155 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsxHm5Yho0Ox7veV3:-MW3t28g6TGcQ1MC5uMw:b-9uucp3) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L155)):*

I don't understand why the context isn't in the stacktrace. I would expect the stacktrace to show the callsite that caused the operation and the exception msg taht caused it. Now I have to scroll to the `caused by` section of the stactrace.
</blockquote></details>

Because you should not send a stacktrace to clients in a production environment, so you lose all context information. BTW: unchecked exceptions are not logged at all.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MW3tl-x2BBRAL1ZiY5D:bwu0ju7) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*

Im not 100% sure this is better. In my experience this is going to create a lot of noise in the log, lots of operations fail w/ the checked exceptions as part of a regular functioning Nessie. Its going to be hard to tell the difference. Also, imho the version store thorwing an exception should be logged by the layer above. It is already isn't it? That is where I would prefer to fix this. Or maybe the log level shoudl be debug?

If we log something as a 'warning' it should be clear and actionable by the user/operator. I don't think these are.
</blockquote></details>

It is not logged at all.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 157 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMA0-1Lf7Vm-Gmu4M:-MW3ttAW9_mgQRNIQCJ3:b-len3hd) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L157)):*

IMHO i see no reason to treat `IAE` differently here.
</blockquote></details>

Well, I didn't want to change existing behavior, so IAE had to be treated differently here.
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java, line 24 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMOKvC3JI8hoD4HrT:-MW3u-DO86y3BsgjFccF:b-amxhzr) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java#L24)):*

I agree that backpressure is hard to get right but we should at some point make an effort do to so, not in this PR and hopefully above the level of Dynamo. 

However, this to me doesn't really alert the user. It wraps an AWS throttle exception. We still get a 500 error back to the client and the client still swallows it.

My concern w/ throttle messages from AWS is that excessive throttling may lead to inconsistent or unrecoverable states. I think there is something between a wrapper exception and a full blown backpressure algo that this PR could do: options may include simple throttle at the http level, client exception for throttling, different return codes etc
</blockquote></details>

I think, you only look at how AWS "communicates" throttling/request-shedding. Other backends do it differently and may report something useful or not. But implementing the specific behavior of DynamoDB and then forcing all other (Version)Store implementations to behave like Dynamo is not great IMO. From experience I can tell, that some distributed database just responds with some "timeout exception", that can mean everything and nothing.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 145 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsJVk1jxoJGDyU3iP:-MW4Pwge4NeXpYDkD4iY:bczbpja) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L145)):*

I'd leave that for a follow-up PR that updates the client-API and therefore also the rest interfaces. It also heavily depends on whether and how support for backpressure will be in Nessie.
</blockquote></details>

This is PR is then two separate changes: the start of a way to rate limit and some changes to exception handling. I would prefer to handle rate limits in one PR, likely after a design discussion and make this PR just about exception handling changes . What do others think?
___
*[versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java, line 155 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpsxHm5Yho0Ox7veV3:-MW4Qjk0BeYcvlX4bmNR:bfptgom) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java#L155)):*

Because you should not send a stacktrace to clients in a production environment, so you lose all context information. BTW: unchecked exceptions are not logged at all.
</blockquote></details>

If this is about giving more context to a client w/ limited visibility to the internals then I would say we move this up to the REST api. Add some richer info to the `NessieError` for example. With possible causes, ways to fix etc. 

I think at this level just throwing the full excpetion w/ cause is the best way to give that information to the REST api
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 118 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVpuTXD9LaZPnofRt7P:-MW4RjprDh4kEcPUAZ5a:bf8l5pp) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L118)):*

It is not logged at all.
</blockquote></details>

As stated above I think its is the responsibility of the REST api layer to choose how to respond to these exceptions. Preferably with succinct and actionable logs/exceptions. My feeling is this will add a lot of unclear warning level stack traces. 
___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java, line 24 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMOKvC3JI8hoD4HrT:-MW4TVLP6OrPMUCuGrEu:b-oepxhe) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/BackendLimitExceededException.java#L24)):*

I think, you only look at how AWS "communicates" throttling/request-shedding. Other backends do it differently and may report something useful or not. But implementing the specific behavior of DynamoDB and then forcing all other (Version)Store implementations to behave like Dynamo is not great IMO. From experience I can tell, that some distributed database just responds with some "timeout exception", that can mean everything and nothing.
</blockquote></details>

I think my above comment was written to be backend store agnostic, and I totally agree we don't want to design somethig to be too dynamo-y. We really don't know how the backend is going to react to high load or what messages it will send but we do know that continuing to hammer it at full pace could lead to undefined behaviour and ultimately data corruption. And at the simplest level if a system tells you to slow down and you drop those messages and continue at full pace you are asking for data corruption or worse. Apologies for being stubborn but I feel like either a) we investigate and design a simple response to overwhelming backends with the goal of avoiding data loss or severely degraded performance or b) we wait until we know enough to fix the problem. I am leaning towards (b) as right now we know when we hit specifically the dynamo backend very hard it acts goofy. But we dont really see 'very hard' in practice and I don't have a clear picture of exactly what 'goofy' is. Thoughts?

laurentgo (dismissed):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/938)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java, line 157 at r2](https://reviewable.io/reviews/projectnessie/nessie/938#-MVsMA0-1Lf7Vm-Gmu4M:-MW5Amb45t8VXaYiH_nS:baai8rj) ([raw file](https://github.com/projectnessie/nessie/blob/2f4e8441332e1267208a70283ea286345f0f8b08/versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/TieredVersionStore.java#L157)):*

Well, I didn't want to change existing behavior, so IAE had to be treated differently here.
</blockquote></details>

Sorry, I got confused myself. The code itself throws IAE because from spec, it's illegal to not provide a hash if you want to create a tag.
But because we wrap the whole method body, instead of just the store access (which should not throw btw), that's why we are rethrowing IAE directly. But for me, it's just a statement that this approach of rewrapping everything is a code smell.

It is also a *bad* practice of logging an exception (especially at warn level which is likely to be enabled) and throw the exception back. Let's say the exception is recoverable according to the caller. The exception would still be logged although not impacting the program behavior, but that might confuse the person reading the logs into thinking there's an actual problem.

versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java
```diff
@@ -201,59 +219,63 @@ public void close() {
...
+            List<Map<String, AttributeValue>> values = responses.get(table);
+            int missingResponses = loadList.size() - values.size();
+            if (missingResponses != 0) {
+              ValueType<?> loadType = loadList.get(0).getValueType();
+              if (loadType == ValueType.REF || loadType == ValueType.L1) {
+                throw new NotFoundException("Unable to find requested ref.");
+              }
+
+              throw new NotFoundException(
+                  String.format("[%d] object(s) missing in table read [%s]. \n\nObjects expected: %s\n\nObjects Received: %s",
```
lillieMaiBauer, line 251: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [FS: Format string should use %n rather than n (VA_FORMAT_STRING_USES_NEWLINE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#fs-format-string-should-use-n-rather-than-n-va-format-string-uses-newline).
This format string includes a newline character (\n). In format strings, it is generally  preferable to use %n, which will produce the platform-specific line separator.

servers/services/src/main/java/org/projectnessie/services/rest/NessieExceptionMapper.java
```diff
@@ -58,28 +62,39 @@ public NessieExceptionMapper(ServerConfig config) {
...
       reason = e.getReason();
+      message = exception.getMessage();
     } else if (exception instanceof JsonParseException
         || exception instanceof JsonMappingException
         || exception instanceof IllegalArgumentException) {
       status = Status.BAD_REQUEST.getStatusCode();
       reason = Status.BAD_REQUEST.getReasonPhrase();
+      message = exception.getMessage();
+    } else if (exception instanceof BackendLimitExceededException) {
+      LOGGER.warn("Backend throttled/refused the request: {}", exception.toString());
```
rymurr, line 87: what is the  reasoning behind `debug` in some branches, `warn` in others and no log for `WebApplicationException`?

snazy: TBH, I don't really know. For the `BackendLimitExceededException` the reason was to give an operator at least some visibility that some requests failed, because the backend's overloaded. We have debug-logging for all code-paths via `BaseExceptionMapper#buildExceptionResponse()`, but that one covers really all exceptions.

snazy: But pushed another commit to prevent logging at debug twice

rymurr (approved):

PR #939, opened 2021-03-11
In-memory `Store` implementation

snazy:
Adds a CHM base in-memory `Store` implementation for local testing and perf-testing.

Changes that abstract the factory and config (`TieredVersionStoreFactory` + `TieredVersionStoreConfig`) are precursors for a follow-up PR that adds more information to Jaeger tracing around the version-store and store (those changes look pretty useless at the moment though).

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/939)

### DIFF
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -53,6 +53,11 @@
     </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-versioned-tiered-inmem</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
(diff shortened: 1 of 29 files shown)


rymurr:
Before reading this too thoroughly I had a few questions:

1) why are we adding a 2nd in memory db to our prod configs? Isn't this a test-only thing?
2) what is the advantage of this over eg dynamo or rocks for testing. I am reluctant to add a few thousand lines of code to upkeep without a clear purpose.

snazy:
This is a test thing for both unit/integration tests and also to see the "best possible performance (without a database backend)" in a perf-test. It makes integration testing easier (think; the Nessie-GC stuff), because you don't need to spin up an external DynamoDB.

An in-mem `Store` can also help to later test resilience by introducing various kinds of failures (random exceptions, pauses, etc).

snazy:
Closing in favor of #1638

PR #942, opened 2021-03-11
Remove `verify`

snazy:
Last main-branch CI failed with
`2021-03-11T18:51:08.7827881Z [ERROR] Failed to execute goal org.apache.maven.plugins:maven-deploy-plugin:3.0.0-M1:deploy (default-deploy) on project nessie-lambda: NoFileAssignedException: The packaging plugin for this project did not assign a main file to the project but it has attachments. Change packaging to 'pom'. -> [Help 1]`

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/942)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,7 +76,7 @@ jobs:
         key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
         restore-keys: ${{ runner.os }}-gradle
     - name: Build with Maven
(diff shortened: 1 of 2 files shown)


laurentgo:
I don't think that's enough and that the failure is caused by the fact the jar is not generated for the `nessie-deltalake-core` module, and that deploying fails because there's no main artifact.

snazy:
This PR is just to fix the main-build, which fails since #941

snazy:
Replaced with #948

PR #949, opened 2021-03-12
add more options to commit meta

### LINKED ISSUE #779: Add extra changes properties (committer...)

rymurr:
* update REST apis to accept full commit meta
* deprecate some api methods
* set commit metadata on server side

closes #779

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)


snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 0 of 28 files reviewed, 12 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-yPZ6M61bQ75373z:-MVb-yPZ6M61bQ75374-:b-g7mmyy) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L37)):*
> ```Java
> 
>   @Nullable
>   public abstract String getCommiter();
> ```

One thing I'm a bit unsure is: who (respectively "what") is the commiter wrt REST endpoints?
Say, I ("Robert") kick off a Spark job that commits to Nessie. The "user" that performs the commits (executes the REST requests) is probably some technical user (e.g. spark-job-runner@some-company.foo), but actually I am the committer. Maybe we need a "4th persona" (technical committer so something like that).
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawgrD7xwAbBYKRtwV:-MVawgrD7xwAbBYKRtwW:bxhewq5) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L40)):*
> ```Java
> 
>   @Nullable
>   public abstract String getEmail();
> ```

Should be `getCommitterEmail()` IMO
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVax13g33bpgr-MM3C6:-MVax13g33bpgr-MM3C7:bbai8jt) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L43)):*
> ```Java
> 
>   /**
>    * according to git the author == the committer unless set otherwise.
> ```

Should have some notes about all properties somewhere, this class' javadoc is probably a good start.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 46 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVax3xJ8-RHkqlFkK7D:-MVax3xJ8-RHkqlFkK7E:b-g2rb56) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L46)):*
> ```Java
>    */
>   @Nullable
>   public abstract String getAuthor();
> ```

Wonder if we should group "name + email" as a separate object, because there are already three personas: committer, author, signed-off-by.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxaDCEaeouVK4a6YC:-MVaxaDCEaeouVK4a6YD:b-woth9d) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L57)):*
> ```Java
>   public abstract String getSignedOffByEmail();
> 
>   public abstract String getMessage();
> ```

Shall we declare/recommend that commit-messages are/should be markdown-formatted, so we don't get a mix of different styles? It's rather cosmetic though.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MVaxHGZ9_MCn_jE5PS0:b552btu) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*
> ```Java
> 
>   @Nullable
>   public abstract Long getCommitTime();
> ```

Can we store the time-zone as well?
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MVawoXz-pV-4UyMWABw:bbai8jt) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*
> ```Java
> 
>   /**
>    * according to git the author time == the commit time unless set otherwise.
> ```

Should have some notes about all properties somewhere, this class' javadoc is probably a good start.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 68 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxwRp8kXirKocfiIU:-MVaxwRp8kXirKocfiIV:b-tvqmmc) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L68)):*
> ```Java
>   public abstract Long getAuthorTime();
> 
>   public abstract Map<String, String> getProperties();
> ```

I could imagine, that knowing which client (Iceberg, DL, Spark version) and Nessie-client-version could be useful. Something like a "User-Agent" or so. And maybe a user-provided "application name" or "job name". So it would be easier to figure out issues, could blame that a specific job or client-version caused issues.
___
*[servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java, line 259 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-64V6PYsTR276Imk:-MVb-64V6PYsTR276Iml:b-12l7m0) ([raw file](https://github.com/projectnessie/nessie/blob/5c48864764a246615ff72af1e0522370abae7134/servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java#L259)):*
> ```Java
>             assertThrows(NessieBadRequestException.class,
>                 () -> tree.transplantCommitsIntoBranch(invalidBranchName, validHash, null, null)).getMessage()),
>         () -> assertEquals("Bad Request (HTTP/400): setContents.branch: " + REF_NAME_MESSAGE,
> ```

Why this removal?
___
*[servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java, line 319 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-AbSDsLNCe-iNW9k:-MVb-AbTCHYWg4SEQYZa:b-12l7m0) ([raw file](https://github.com/projectnessie/nessie/blob/5c48864764a246615ff72af1e0522370abae7134/servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java#L319)):*
> ```Java
>         () -> assertEquals("Bad Request (HTTP/400): setContents.hash: " + HASH_MESSAGE,
>             assertThrows(NessieBadRequestException.class,
>                 () -> contents.setContents(key, validBranchName, invalidHash, null, cts)).getMessage()),
> ```

Why this removal?
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 98 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazdRD8MPfmremG_vV:-MVazdRD8MPfmremG_vW:bqbcwyq) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L98)):*
> ```Java
>   public void setContents(ContentsKey key, String branch, String hash, String message, Contents contents)
>       throws NessieNotFoundException, NessieConflictException {
>     CommitMeta meta = CommitMeta.builder().message(message).build();
> ```

I'd prefer to let the caller provide the `CommitMeta` and not limit it to a single `message`.
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 105 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazlLw7OywYCbd3klm:-MVazlLw7OywYCbd3kln:bytuugb) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L105)):*
> ```Java
>   public void deleteContents(ContentsKey key, String branch, String hash, String message)
>       throws NessieNotFoundException, NessieConflictException {
>     CommitMeta meta = CommitMeta.builder().message(message).build();
> ```

(same for `CommitMeta` here)

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 0 of 28 files reviewed, 12 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 98 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazdRD8MPfmremG_vV:-MVb1djb8lRGeYxIyLXs:bams17c) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L98)):*

I'd prefer to let the caller provide the `CommitMeta` and not limit it to a single `message`.
</blockquote></details>

Actually, I wouldn't mind to remove this method + `deleteContents` on the REST API level, and only have those methods as a "convenience API". WDYT?

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 0 of 28 files reviewed, 12 unresolved discussions (waiting on @laurentgo and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 37 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-yPZ6M61bQ75373z:-MVbNizhCVyMxYWnXjcf:b-173nsq) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L37)):*

One thing I'm a bit unsure is: who (respectively "what") is the commiter wrt REST endpoints?
Say, I ("Robert") kick off a Spark job that commits to Nessie. The "user" that performs the commits (executes the REST requests) is probably some technical user (e.g. spark-job-runner@some-company.foo), but actually I am the committer. Maybe we need a "4th persona" (technical committer so something like that).
</blockquote></details>

I think signed-off by works fine for that right? The committer is whoever is authorised (whether bot or human)
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawgrD7xwAbBYKRtwV:-MVbO-ZEBz9frfSUX638:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L40)):*

Should be `getCommitterEmail()` IMO
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 43 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVax13g33bpgr-MM3C6:-MVbOBWK3KpzyuhJUfYB:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L43)):*

Should have some notes about all properties somewhere, this class' javadoc is probably a good start.
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 46 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVax3xJ8-RHkqlFkK7D:-MVbOfxA9hknj3UpmB5N:bfpql6l) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L46)):*

Wonder if we should group "name + email" as a separate object, because there are already three personas: committer, author, signed-off-by.
</blockquote></details>

I have removed email. I don't think it makes a ton of sense. We don't know it at the server level and can never guarantee to be populated. If you dislike I can return it.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 57 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxaDCEaeouVK4a6YC:-MVbPL8_3b5naoUVFpro:b9cyifn) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L57)):*

Shall we declare/recommend that commit-messages are/should be markdown-formatted, so we don't get a mix of different styles? It's rather cosmetic though.
</blockquote></details>

I think its up to the org to specify a message style. I have noted that if in markdown the UI may display cleanly.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MVbPmR451o8iJStUtgz:b-518kim) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

Can we store the time-zone as well?
</blockquote></details>

I have a phobia of storing timezones. I have made a note this is in UTC. Up to the UI etc to display in a local timezone. I think since Nessie isnt a distributed vcs its less relevant to know the tz.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MVbQCJf0MUgKnOv72LX:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

Should have some notes about all properties somewhere, this class' javadoc is probably a good start.
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 68 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxwRp8kXirKocfiIU:-MVbQop5BUy0CW0zqyX0:b-cq0t0u) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L68)):*

I could imagine, that knowing which client (Iceberg, DL, Spark version) and Nessie-client-version could be useful. Something like a "User-Agent" or so. And maybe a user-provided "application name" or "job name". So it would be easier to figure out issues, could blame that a specific job or client-version caused issues.
</blockquote></details>

Have added more javadoc specifying the types of data one should add there
___
*[servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java, line 259 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-64V6PYsTR276Imk:-MVbQx9R5NMvCBEIHdjt:byk80my) ([raw file](https://github.com/projectnessie/nessie/blob/5c48864764a246615ff72af1e0522370abae7134/servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java#L259)):*

Why this removal?
</blockquote></details>

reverted. 
___
*[servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java, line 319 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVb-AbSDsLNCe-iNW9k:-MVbRJVZB_uXiGHhBMdx:byk80my) ([raw file](https://github.com/projectnessie/nessie/blob/5c48864764a246615ff72af1e0522370abae7134/servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java#L319)):*

Why this removal?
</blockquote></details>

reverted. 
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 98 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazdRD8MPfmremG_vV:-MVbRMwE-0U2LoQBmlCK:b28096z) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L98)):*

Actually, I wouldn't mind to remove this method + `deleteContents` on the REST API level, and only have those methods as a "convenience API". WDYT?
</blockquote></details>

I agree. I think `setContents` and `deleteContents` should go to util methods. I didn't do it here because the iceberg stuff still depends on the existence of the server methods. I think it should be done as part of a separate change though, given the discussion around similar in the tree api. I am thinking probably a slightly different interface for the client tbh.

Note we can't remove the server version of these methods till we've propagated to iceberg. It uses `setContents`
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 105 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazlLw7OywYCbd3klm:-MVbRu6-CuO4lMqAZUQg:b-xgwur6) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L105)):*

(same for `CommitMeta` here)
</blockquote></details>

We can't do this one at all till we mvoe to util/helper methods. Delete can't accept a body.

laurentgo (commented):
> Reviewed 2 of 14 files at r1, 2 of 24 files at r2.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 12 unresolved discussions (waiting on @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawgrD7xwAbBYKRtwV:-MVsI9JpAAMGOBxeOIrD:b-wrgili) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L40)):*
> ```
> CommitMeta
> ```
Maybe rely on the mailbox definition from https://tools.ietf.org/html/rfc5322 (or something like the generic description from git: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---authorltauthorgt)
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MVri4ib3Z1XzIgiMKRR:b-j3lxjm) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

I have a phobia of storing timezones. I have made a note this is in UTC. Up to the UI etc to display in a local timezone. I think since Nessie isnt a distributed vcs its less relevant to know the tz.
</blockquote></details>

(tbh) I would be happier with using `Instant` and Jackson jdk8 datetime module which should serialize it using the ISO-8601 instant format

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 12 unresolved discussions (waiting on @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 40 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawgrD7xwAbBYKRtwV:-MVz_NES2IUVyB6U3Mdk:b-za3qn) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L40)):*

> ```
> CommitMeta
> ```
Maybe rely on the mailbox definition from https://tools.ietf.org/html/rfc5322 (or something like the generic description from git: https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---authorltauthorgt)
</blockquote></details>

Have added some language that the commiter/author/signed-off name should follow the git format. I haven't enforced it b/c its typically going to be hard to relyably get an email address from a spark job (or equivalent)
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MVzbBzB4TnvY2KKmCu7:b-k7yf48) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

(tbh) I would be happier with using `Instant` and Jackson jdk8 datetime module which should serialize it using the ISO-8601 instant format
</blockquote></details>

I just moved to Instant but all the jackson related tests fail. Spark2/Hive2 don't include the correct modules to decode `Instant`.  Annoyingly hive2 has the joda module, spark2 has neither, spark3 has the correct module, hive3 has neither...sigh

I am tempted to leave as long ms since epoch for compatibility sake

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 12 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 46 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVax3xJ8-RHkqlFkK7D:-MW3xjmw6KGtbwcYhA9r:b-28ybcv) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L46)):*

I have removed email. I don't think it makes a ton of sense. We don't know it at the server level and can never guarantee to be populated. If you dislike I can return it.
</blockquote></details>

Let's leave it. If it's not always present, that's not great, but better than not having the field at all.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MW3xt1zCy3s3j9QWCdA:b1sxhwm) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

I just moved to Instant but all the jackson related tests fail. Spark2/Hive2 don't include the correct modules to decode `Instant`.  Annoyingly hive2 has the joda module, spark2 has neither, spark3 has the correct module, hive3 has neither...sigh

I am tempted to leave as long ms since epoch for compatibility sake
</blockquote></details>

You could add the string representation of `java.util.TimeZone` or `java.time.ZoneId` 
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MW3zFHV5CQlCkzLG7tg:bojdfjf) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

Done.
</blockquote></details>

Is this Unix-like (seconds since epoch) or Java-like (milliseconds since epoch)? It's very easy to interpret a pure `long` wrong due to an accidentally added ` * 1000L` or whatever. Can you make all the timestamps here wrapped in another object that has both millis-since-epoch and a time-zone along with explicit methods like `asMillisSinceEpoch()` and `asSecondsSinceEpoch()` 
and `asCalendar()` (what's needed).

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 12 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MW4WcUYCOp90vqUWvfD:b-lof1a6) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

You could add the string representation of `java.util.TimeZone` or `java.time.ZoneId` 
</blockquote></details>

What is the reasoning behind adding timezone info? I have found that its almost always a bad idea to store timezones in the backend. eg `Instant` is just a long millis since epoch and extra nanos.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MW4X2bOCDvoDvsIOjlb:bwdtkft) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

Is this Unix-like (seconds since epoch) or Java-like (milliseconds since epoch)? It's very easy to interpret a pure `long` wrong due to an accidentally added ` * 1000L` or whatever. Can you make all the timestamps here wrapped in another object that has both millis-since-epoch and a time-zone along with explicit methods like `asMillisSinceEpoch()` and `asSecondsSinceEpoch()` 
and `asCalendar()` (what's needed).
</blockquote></details>

Agreed, a `long` too error prone. Will try and find a way around it w/o relying on jackson libraries.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 12 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MW4sjHL0DeTyMMGTQu_:bplynne) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

Agreed, a `long` too error prone. Will try and find a way around it w/o relying on jackson libraries.
</blockquote></details>

Just wrote a custom Jackson serde. Now we don't need the jackson module

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 16 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 60 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVaxHGZ9_MCn_jE5PS-:-MW52UtGCSS6MIzkewXc:buwhm8m) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L60)):*

What is the reasoning behind adding timezone info? I have found that its almost always a bad idea to store timezones in the backend. eg `Instant` is just a long millis since epoch and extra nanos.
</blockquote></details>

ah good point about the module likely not being present :( 
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MW53USLDTND4Qsv8ySt:b-pd6vu7) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

Just wrote a custom Jackson serde. Now we don't need the jackson module
</blockquote></details>

I was wondering if the jackson API has been stable enough that compiling with 2.11 would still work with whatever version hive2 is using....
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 124 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW53jIu9w7vblCh1HEg:-MW53jIu9w7vblCh1HEh:b-vip9nl) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L124)):*
> ```Java
>    * Used to serialize an instant to ISO-8601 format. Required because not all platforms we work with support jackson's jdk8 modules.
>    */
>   public static class CustomInstantSerializer extends StdSerializer<Instant> {
> ```

(style) Maybe `Custom` is not necessary?
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 124 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW55iAc45gYVz-7yO9C:-MW55iAc45gYVz-7yO9D:b-eau8x2) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L124)):*
> ```Java
>    * Used to serialize an instant to ISO-8601 format. Required because not all platforms we work with support jackson's jdk8 modules.
>    */
>   public static class CustomInstantSerializer extends StdSerializer<Instant> {
> ```

unit tests for the serde?
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 125 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW56ySH5c1OqkOptAiV:-MW56ySH5c1OqkOptAiW:b1yfqus) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L125)):*
> ```
> DateTimeFormatter
> ```
isn't the formatter already set to UTC (according to javadoc?)
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 128 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW53wXTB_F7qoRMoJjt:-MW53wXTB_F7qoRMoJju:b-wngq0f) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L128)):*
> ```Java
> 
>     public CustomInstantSerializer() {
>       this(null);
> ```

why not `this(Instant.class)`?
___
*[servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java, line 98 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVazdRD8MPfmremG_vV:-MW550E-842COuu1MzIm:bh9z02) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java#L98)):*

I agree. I think `setContents` and `deleteContents` should go to util methods. I didn't do it here because the iceberg stuff still depends on the existence of the server methods. I think it should be done as part of a separate change though, given the discussion around similar in the tree api. I am thinking probably a slightly different interface for the client tbh.

Note we can't remove the server version of these methods till we've propagated to iceberg. It uses `setContents`
</blockquote></details>

I think we got confused when we looked at the github equivalent API. I checked it recently and although github has those API, what they do is create an alternate tree representation, which is then committed using the tree api.

rymurr (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 27 files reviewed, 16 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 63 at r1](https://reviewable.io/reviews/projectnessie/nessie/949#-MVawoXz-pV-4UyMWABv:-MW5AUTB0LVv9UagJbm1:b-u52214) ([raw file](https://github.com/projectnessie/nessie/blob/b06f2cb22197e225e5ab5181cc54e6628c70e32e/model/src/main/java/org/projectnessie/model/CommitMeta.java#L63)):*

I was wondering if the jackson API has been stable enough that compiling with 2.11 would still work with whatever version hive2 is using....
</blockquote></details>

I've added an Instant to the jackson unit tests and re-ran them. So far so good.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 124 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW53jIu9w7vblCh1HEg:-MW5BCHB5o1sa4w6vfbU:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L124)):*

(style) Maybe `Custom` is not necessary?
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 124 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW55iAc45gYVz-7yO9C:-MW5BCbr3lpT3qrqAeEI:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L124)):*

unit tests for the serde?
</blockquote></details>

Done.
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 125 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW56ySH5c1OqkOptAiV:-MW5BDaH4_9ajY6M70ZQ:b-lriem1) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L125)):*

> ```
> DateTimeFormatter
> ```
isn't the formatter already set to UTC (according to javadoc?)
</blockquote></details>

where did you see that? I couldn't find it. It looks like it gets initialized with `null`
___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 128 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW53wXTB_F7qoRMoJjt:-MW5BHjK-EUaZZcc01VP:b-dlcupy) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L128)):*

why not `this(Instant.class)`?
</blockquote></details>

copying from the interwebs. fixed.

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 29 files reviewed, 16 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 125 at r4](https://reviewable.io/reviews/projectnessie/nessie/949#-MW56ySH5c1OqkOptAiV:-MWA52nZ3TuzWBFfoiPQ:b-vdkg26) ([raw file](https://github.com/projectnessie/nessie/blob/cad2f38bd00c3bd8e97f3f4045d123f484e66f3d/model/src/main/java/org/projectnessie/model/CommitMeta.java#L125)):*
From javadoc, unless I misinterpreted it? https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_INSTANT
> This is a special case formatter intended to allow a human readable form of an Instant. The Instant class is designed to only represent a point in time and internally stores a value in nanoseconds from a fixed epoch of 1970-01-01Z. As such, an Instant cannot be formatted as a date or time without providing some form of time-zone. This formatter allows the Instant to be formatted, by providing a suitable conversion using ZoneOffset.UTC.

laurentgo (approved):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/949)* status: 4 of 29 files reviewed, 15 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

rymurr:
@laurentgo could i get a re-approve post rebase?

snazy (approved):

PR #960, opened 2021-03-16
Fix issue when loading the Gradle plugin into IntelliJ ("missing JAVA_HOME")

snazy:
... and some code style cosmitic changes

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/960)


tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/StartTask.java
```diff
@@ -32,52 +32,48 @@
 import org.gradle.api.tasks.testing.Test;
 
 public class StartTask extends DefaultTask {
-  private static final Object lock = new Object();
```
snazy, line 35: Task-actions are not executed concurrently.

laurentgo (approved):

PR #961, opened 2021-03-16
Updating access to classes of the in-memory VersionStore.

brunos-bq:
This PR updates the classes Commit and CommitIterator to be publicly accessible, instead of only accessible from within the package, as it is now.

The reason for the change is to facilitate writing VersionStore implementations that follow the same logic using different storage structures.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/961)

### DIFF
--- a/versioned/memory/src/main/java/org/projectnessie/versioned/memory/Commit.java
+++ b/versioned/memory/src/main/java/org/projectnessie/versioned/memory/Commit.java
@@ -43,7 +43,7 @@ import com.google.protobuf.UnsafeByteOperations;
  * @param <ValueT> commit value
  * @param <MetadataT> commit metadata
  */
-class Commit<ValueT, MetadataT> {
+public class Commit<ValueT, MetadataT> {
   private static final HashFunction COMMIT_HASH_FUNCTION = Hashing.sha256();
 
   public static final Hash NO_ANCESTOR = Hash.of(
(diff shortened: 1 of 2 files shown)


snazy:
I would not make a `VersionStore` implementation depend on another `VersionStore` implementation because of a single class.
However, sharing common code makes sense to me. Moving the class to `nessie-versioned-spi` could work. @laurentgo WDYT ?

laurentgo:
I'm torn on this one request:
- `Commit` exists because the inmemory store needs something to save commit data, but doesn't require serialization. Moving it to spi (or allowing access to it) sounds equivalent to saying Dynamo internal objects should be exposed too so that they can be reused
- Kind of the same for `CommitsIterator`. It was designed to be used in conjunction with the store implementation

laurentgo:
To clarify, it could make more sense if the in-memory store itself would serve as a basis for a new family of stores, but considering the store was designed for tests and helping quick iterations on API design (meaning it can be refactored to accomodate changes), and since our recommended solution is to use the tiered store spi as the basis for creating a new store implementation, I don't see a long term objective of maintaining those classes for usages other than internal.

rymurr:
Hey @brunos-bq will you just copy the classes then?

brunos-bq:
Yes, I will just copy the classes. I will close this PR.

PR #965, opened 2021-03-17
Fix parameters in python demo

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/965)


python/demo/nessie-delta-demo.ipynb
```diff
@@ -29,7 +29,7 @@
     "                    .config(\"spark.jars\", \"../../clients/deltalake/spark3/target/nessie-deltalake-spark3-0.4.1-SNAPSHOT.jar\")\n",
     "                    .config(\"spark.sql.execution.pyarrow.enabled\", \"true\") \\\n",
     "                    .config(\"spark.hadoop.fs.defaultFS\", 'file://' + os.getcwd() + '/spark_warehouse') \\\n",
-    "                    .config(\"spark.hadoop.nessie.url\", \"http://localhost:19120/api/v1\") \\\n",
```
rymurr, line 32: Was this change tested? I don't recall making the associated change on the Delta side

snazy: The same change was applied to `org.projectnessie.deltalake.NessieLogStore` [here](https://github.com/projectnessie/nessie/blob/main/clients/deltalake/core/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala#L51)

rymurr (approved):

PR #986, opened 2021-03-23
Add `TracingStore` + `TracingVersionStore` facades

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/986)

### DIFF
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
@@ -198,7 +198,7 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     return delegateStream1Ex("VersionStore.getDiffs", b -> b
         .withTag("nessie.version-store.operation", "GetDiffs")
         .withTag("nessie.version-store.from", safeToString(from))
-        .withTag("to", safeToString(to)),
+        .withTag("nessie.version-store.to", safeToString(to)),
         () -> delegate.getDiffs(from, to));
   }
 
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
@@ -223,6 +223,9 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     try {
       result = delegate.handle().onClose(scope::close);
       return result;
+    } catch (IllegalArgumentException e) {
+      // IllegalArgumentException is a special kind of exception that indicates a user-error.
+      throw e;
     } catch (RuntimeException e) {
       throw traceRuntimeException(scope, e);
     } finally {
@@ -240,6 +243,9 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     try {
       result = delegate.handle().onClose(scope::close);
       return result;
+    } catch (IllegalArgumentException e) {
+      // IllegalArgumentException is a special kind of exception that indicates a user-error.
+      throw e;
     } catch (RuntimeException e) {
       throw traceRuntimeException(scope, e);
     } finally {
@@ -255,6 +261,9 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     try (Scope scope = createActiveScope(spanName, spanBuilder)) {
       try {
         return delegate.handle();
+      } catch (IllegalArgumentException e) {
+        // IllegalArgumentException is a special kind of exception that indicates a user-error.
+        throw e;
       } catch (RuntimeException e) {
         throw traceRuntimeException(scope, e);
       }
@@ -266,6 +275,9 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     try (Scope scope = createActiveScope(spanName, spanBuilder)) {
       try {
         return delegate.handle();
+      } catch (IllegalArgumentException e) {
+        // IllegalArgumentException is a special kind of exception that indicates a user-error.
+        throw e;
       } catch (RuntimeException e) {
         throw traceRuntimeException(scope, e);
       }
@@ -277,6 +289,9 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
     try (Scope scope = createActiveScope(spanName, spanBuilder)) {
       try {
         delegate.handle();
+      } catch (IllegalArgumentException e) {
+        // IllegalArgumentException is a special kind of exception that indicates a user-error.
+        throw e;
       } catch (RuntimeException e) {
         throw traceRuntimeException(scope, e);
       }
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
@@ -40,9 +40,10 @@ import io.opentracing.util.GlobalTracer;
  * @param <VALUE> see {@link VersionStore}
  * @param <METADATA> see {@link VersionStore}
  */
-public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE, METADATA> {
+public class TracingVersionStore<VALUE, METADATA, VALUE_TYPE extends Enum<VALUE_TYPE>>
+    implements VersionStore<VALUE, METADATA, VALUE_TYPE> {
 
-  private final VersionStore<VALUE, METADATA> delegate;
+  private final VersionStore<VALUE, METADATA, VALUE_TYPE> delegate;
   private final Supplier<Tracer> tracerSupplier;
 
   /**
@@ -50,7 +51,7 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
    * {@link GlobalTracer#get() GlobalTracer}.
    * @param delegate backing/delegate {@link VersionStore}
    */
-  public TracingVersionStore(VersionStore<VALUE, METADATA> delegate) {
+  public TracingVersionStore(VersionStore<VALUE, METADATA, VALUE_TYPE> delegate) {
     this(delegate, GlobalTracer::get);
   }
 
@@ -61,7 +62,7 @@ public class TracingVersionStore<VALUE, METADATA> implements VersionStore<VALUE,
    * @param tracerSupplier {@link Supplier} for the {@link Tracer} to use
    */
   @VisibleForTesting
-  TracingVersionStore(VersionStore<VALUE, METADATA> delegate, Supplier<Tracer> tracerSupplier) {
+  TracingVersionStore(VersionStore<VALUE, METADATA, VALUE_TYPE> delegate, Supplier<Tracer> tracerSupplier) {
     this.tracerSupplier = tracerSupplier;
     this.delegate = delegate;
   }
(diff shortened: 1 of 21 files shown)


laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/986)* status: 0 of 17 files reviewed, 2 unresolved discussions (waiting on @snazy)

___
*[versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java, line 64 at r1](https://reviewable.io/reviews/projectnessie/nessie/986#-MWfGBJaBn-rK103OCNz:-MWfGBJaBn-rK103OCO-:b-axp298) ([raw file](https://github.com/projectnessie/nessie/blob/6aaa3c3c885fb57a75ddc156b8eab5c821c7d66f/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java#L64)):*
> ```Java
>    */
>   @VisibleForTesting
>   public TracingVersionStore(VersionStore<VALUE, METADATA> delegate, Supplier<Tracer> tracerSupplier) {
> ```

does it need to be public or would package protected visibility be enough?
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java, line 326 at r1](https://reviewable.io/reviews/projectnessie/nessie/986#-MWfHnn30504t89NyW3h:-MWfHnn30504t89NyW3i:bmy87i8) ([raw file](https://github.com/projectnessie/nessie/blob/6aaa3c3c885fb57a75ddc156b8eab5c821c7d66f/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java#L326)):*
> ```Java
> 
>   private static RuntimeException traceRuntimeException(Scope scope, RuntimeException e) {
>     Tags.ERROR.set(scope.span().log(ImmutableMap.of(Fields.EVENT, Tags.ERROR.getKey(),
> ```

is it a good pattern or is it similar to how we should not log an error and throw back the exception which might not have been an error after all if the caller knows how to handle it?

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/986)* status: 0 of 18 files reviewed, 2 unresolved discussions (waiting on @laurentgo)

___
*[versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java, line 64 at r1](https://reviewable.io/reviews/projectnessie/nessie/986#-MWfGBJaBn-rK103OCNz:-MWi2Oxw0JcnQHHSct4a:b-177il) ([raw file](https://github.com/projectnessie/nessie/blob/6aaa3c3c885fb57a75ddc156b8eab5c821c7d66f/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java#L64)):*

does it need to be public or would package protected visibility be enough?
</blockquote></details>

not really, no. changed it.
___
*[versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java, line 326 at r1](https://reviewable.io/reviews/projectnessie/nessie/986#-MWfHnn30504t89NyW3h:-MWi2TLjBAARmqAk00Sd:b-k09cgz) ([raw file](https://github.com/projectnessie/nessie/blob/6aaa3c3c885fb57a75ddc156b8eab5c821c7d66f/versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java#L326)):*

is it a good pattern or is it similar to how we should not log an error and throw back the exception which might not have been an error after all if the caller knows how to handle it?
</blockquote></details>

It's a bit tricky IMO. On the one hand we have unchecked exceptions that indicate a real technical failure but also usages of e.g. `IllegalArgumentException` that indicate a user-error, but are returned as a "HTTP/500 server error" to clients. This particular one is meant to mark the trace as "error" because of some technical error. TL;DR it's hard to distinguish a real technical error from a user-error here. It feels like it's worth to have a discussion how to properly differentiate between "real technical runtime-exceptions" and actual "parameter validation exceptions"?

snazy:
Rebased + force-pushed due to conflicts

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -0,0 +1,334 @@
...
+  TracingVersionStore(VersionStore<VALUE, METADATA, VALUE_TYPE> delegate, Supplier<Tracer> tracerSupplier) {
+    this.tracerSupplier = tracerSupplier;
+    this.delegate = delegate;
+  }
+
+  @Override
+  @Nonnull
+  public Hash toHash(@Nonnull NamedRef ref) throws ReferenceNotFoundException {
+    return delegate1Ex("VersionStore.toHash", b -> b
+        .withTag("nessie.version-store.operation", "ToHash")
```
rymurr, line 74: I thin the tag names should be `static final String` s

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -0,0 +1,334 @@
...
+        // IllegalArgumentException is a special kind of exception that indicates a user-error.
+        throw e;
+      } catch (RuntimeException e) {
+        throw traceRuntimeException(scope, e);
+      }
+    }
+  }
+
+  @FunctionalInterface
+  interface Delegate<R> {
```
rymurr, line 303: The overloading of `Delegate`, `delegate()` and `delegate` is slightly confusing. Could we rename a few to create more distinction?

versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java
```diff
@@ -570,4 +598,14 @@ private static void verifyKeySchema(TableDescription description) {
     throw new IllegalStateException(String.format("Invalid key schema for table: %s. Key schema should be a hash partitioned "
         + "attribute with the name 'id'.", description.tableName()));
   }
+
+  private Tracer getTracer() {
```
rymurr, line 602: we can only use the global tracer here but above we have the option to specify a tracer. Can we do one or the other? I prefer just using the global tracer.

versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/metrics/TracingExecutionInterceptor.java
```diff
@@ -76,11 +77,11 @@ public void afterMarshalling(final AfterMarshalling context,
     final Span span = executionAttributes.getAttribute(SPAN_ATTRIBUTE);
     final SdkHttpRequest httpRequest = context.httpRequest();
```
rymurr, line 79: nit: This class only exists while we are watiing for opentracing to update in Quarkus. Wonder if its worth maintaining cosmetic changes?

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalBranch.java
```diff
@@ -403,57 +410,71 @@ private void save(Store store) {
...
-            LOGGER.debug("Completed collapse update on attempt {}.", attempt);
-            return producer.build().getBranch();
+      try (Scope outerScope = createSpan("InternalBranch.collapseIntentionLog")
+          .withTag("nessie.operation", "CollapseIntentionLog")
+          .withTag("nessie.branch", branch.getName())
+          .startActive(true)) {
+        try {
+          UpdateState updateState = initialState;
+          for (int attempt = 0; attempt < attempts; attempt++) {
+            try (Scope innerScope = createSpan("Attempt-" + attempt).startActive(true)) {
```
rymurr, line 420: again minor concern over cost of this op vs payoff

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/TracingStore.java
```diff
@@ -0,0 +1,248 @@
...
+    try {
+      return store.getValues(type).onClose(scope::close);
+    } catch (RuntimeException e) {
+      e = traceRuntimeException(scope, e);
+      scope.close();
+      throw e;
+    }
+  }
+
+  private static <C extends BaseValue<C>> String safeName(ValueType<C> type) {
```
rymurr, line 221: some dup code here and in version store. Possible to create a util?

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalBranch.java
```diff
@@ -403,57 +410,71 @@ private void save(Store store) {
...
-              .and(SetClause.equals(last.toBuilder().name(Commit.PARENT).build(), updateState.finalL1.getParentId().toEntity()))
-              .and(SetClause.equals(last.toBuilder().name(Commit.ID).build(), updateState.finalL1.getId().toEntity()));
-
-          InternalRef.Builder<?> producer = EntityType.REF.newEntityProducer();
-          boolean updated = store.update(ValueType.REF, branch.getId(), update, Optional.of(condition), Optional.of(producer));
-          if (updated) {
-            LOGGER.debug("Completed collapse update on attempt {}.", attempt);
-            return producer.build().getBranch();
+      try (Scope outerScope = createSpan("InternalBranch.collapseIntentionLog")
+          .withTag("nessie.operation", "CollapseIntentionLog")
```
rymurr, line 414: here and elsewhere: tag names could move to static final Strings

rymurr (changes requested):

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
snazy: Done

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
snazy: Renamed some of those.

versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/DynamoStore.java
snazy: It's been due to `GlobalTracer` not allowíng us to register the "global tracer" once, but never change it. I've updated the tests to use a proxy-instance between the `Tracer` for the tests and `GlobalTracer`

versioned/tiered/dynamodb/src/main/java/org/projectnessie/versioned/dynamodb/metrics/TracingExecutionInterceptor.java
snazy: ah, interesting. i've reverted the changes to this file.

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalBranch.java
snazy: Yea, that particular one is tricky. The code is "too far away" from knowing whether the user wants things to be traced, but if you trace, the information is pretty useful.

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/store/TracingStore.java
snazy: Yup, done. Also tried to "unify" the test-code stuff to reuse common things and to make the test-code a more readable.

versioned/tiered/tiered-impl/src/main/java/org/projectnessie/versioned/impl/InternalBranch.java
snazy: done

rymurr (commented):
LGTM. The only comment is if we can try and abstract the tracing in `InternalBranch` and `DynamoStore`. I feel like its a bit harder to read and I don't fully understand the perf  implications

snazy:
Added a commit to address this. Opted to introduce an _internal_ `TieredVersionStoreConfig` object that carries the wait-for-collapse, enable-tracing and the two commit-attempts values, as adding another param to TVS's ctor felt too much.
Refactored the InternalBranch.UpdateState.collapse... method a bit to make it more readable.

rymurr (dismissed):

rymurr (approved):

PR #1003, opened 2021-03-29
Fix handling of server-side IllegalArgumentException

snazy:
When the server (`VersionStore`) throws an `IllegalArgumentException` indicating a parameter error, so a user-mistake, it is returned as an HTTP/500 (server-error).

This PR maps `IllegalArgumentException` to HTTP/400.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1003)


rymurr (approved):
LGTM

PR #1029, opened 2021-04-05
Backwards compatible payload parsing in Dynamo.

rymurr:
If payload doesn't exist then ignore and assume key is the whole key and not a byte prefixed key.

This fix is slightly annoying but not harmful. This happens mostly in testing of GC packages because of version mismatches between the iceberg version of Nessie and the version of nessie on HEAD. As the format stabilises and the iceberg version of Nessie catches up we should be able to remove this shim.

The goal of the fix is:
1) if the first element of the key is `\u0000` then the payload byte is present but the payload is null
2) if the first element of the key is not `\u0000` and parseable as a Byte then its a payload
3) otherwise the first element of the key isn't a payload and is the first part of the key

Case 1 or 2 should skip the first element when building the key and 3 should use the first element to build the key

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1029)


snazy (dismissed):

rymurr:
> Can you open an issue to basically revert this change, once Iceberg has a "recent enough" Nessie version?

Done, see #1047

PR #1031, opened 2021-04-05
Add `key` to gc data set.

rymurr:
The `key` is the `Key` or table name and is carried through the GC process to provide more context to the end user.

Currently the Tiered Version Store GC algorithm scans the entire list of values to determine the last access time. While doing so it marks whether the value is reachable but the key that points to the value is lost. This PR retains the key information when scanning for referencable values and carries that key through the entire GC algorithm. This allows for the key to be referenced as part of the GC reporting of AssetKeys that can be cleaned up.

This isn't an improvement or functional change of the tiered gc algorithm, rather a cosmetic change to how a GC report should look.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1031)

### DIFF
--- a/versioned/gc/iceberg/pom.xml
+++ b/versioned/gc/iceberg/pom.xml
@@ -44,12 +44,7 @@
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-server-store</artifactId>
       <version>${project.version}</version>
-      <exclusions>
-        <exclusion>
-          <groupId>commons-logging</groupId>
-          <artifactId>commons-logging</artifactId>
-        </exclusion>
-      </exclusions>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.iceberg</groupId>
--- a/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/AssetKeySerializer.java
+++ b/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/AssetKeySerializer.java
@@ -30,7 +30,7 @@ import com.google.protobuf.InvalidProtocolBufferException;
  * <p>We use a specialized object mapper to persist with classname w/o annotations. Ideally this data should
  * not be persisted and only used to move around Spark. If it is the ObjectMapper needs to be configured the same way.
  */
-public class AssetKeySerializer implements Serializer<AssetKey>, Serializable {
+class AssetKeySerializer implements Serializer<AssetKey>, Serializable {
 
   private final SerializableConfiguration configuration;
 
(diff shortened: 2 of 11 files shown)


snazy:
@rymurr can you merge main?

rymurr:
> @rymurr can you merge main?

done!

snazy (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1031)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @laurentgo and @rymurr)

___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/AssetKeySerializer.java, line 33 at r1](https://reviewable.io/reviews/projectnessie/nessie/1031#-MXkwfVu9-pV4_xv5Yi4:-MXkwfVu9-pV4_xv5Yi5:bcgjvre) ([raw file](https://github.com/projectnessie/nessie/blob/206a2318f792aabf99d02f0b8eac42969614781e/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/AssetKeySerializer.java#L33)):*
> ```Java
>  * not be persisted and only used to move around Spark. If it is the ObjectMapper needs to be configured the same way.
>  */
> public class AssetKeySerializer implements Serializer<AssetKey>, Serializable {
> ```

Why this change?
___
*[versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/ValueTypeFilter.java, line 27 at r1](https://reviewable.io/reviews/projectnessie/nessie/1031#-MXkwh8C1AUkW3jtnhsX:-MXkwh8C1AUkW3jtnhsY:bcgjvre) ([raw file](https://github.com/projectnessie/nessie/blob/206a2318f792aabf99d02f0b8eac42969614781e/versioned/gc/iceberg/src/main/java/org/projectnessie/versioned/gc/ValueTypeFilter.java#L27)):*
> ```Java
> import com.google.protobuf.ByteString;
> 
> public class ValueTypeFilter implements FilterFunction<CategorizedValue>, Serializable {
> ```

Why this change?

snazy (commented):
Err - didn't want to "block"/"request changes" (fat finger bug)
Actually the opposite... happy with this one, if @laurentgo is.

rymurr:
those package-private to public changes should be part of #1032 removed them from this PR

rymurr:
@snazy and @laurentgo any thoughts on this one?

snazy:
See [here](https://github.com/projectnessie/nessie/pull/1031#pullrequestreview-631215156) ;)

snazy (approved):

PR #1042, opened 2021-04-06
Support lazy start of Nessie-Quarkus-App via Gradle plugin

### LINKED ISSUE #1022: Proper Gradle build-cache interaction w/ the Gradle Plugin
Currently the gradle plugin runs at every gradle invocation even if the underlying code hasn't changed. The gradle plugin should be sensitive to its inputs and only run tests/start nessie service if its inputs have changed.

snazy:
Move most of the boilerplate code from the Iceberg (a Gradle project) test-build-script-code into the plugin.

Two relevant changes:

* The `quarkus.test.http-port` system property must not be in the Gradle `Test` tasks system property, because it's a rather random port, and system properties are part of the Gradle build-cache-key. This has been moved to a `CommandLineArgumentProvider` that's added to `Test.getJvmArgumentProviders()`.
* Gradle has no explicit support to "run a task only if another task would run". I.e. the `quarkus-start` task would always run, even if the `test` task is cached. The workaround here is to let the Nessie-Quarkus-App start via a `Test.doFirst()`-closure.

Minor changes:
* Cleanup user-side (Iceberg) Gradle build-script (will be done via a "bump Nessie to 0.5.0" Iceberg-PR)
* Add the relevant inputs for the Gradle build-cache to the `Test` task as well
* Change nomenclature to make clear that configs/tasks are not general Quarkus tasks, but Nessie-specific Quarkus tasks
* Add a log4j2 configuration to allow logging (no log4j2 config means no logging at all) + an env var to override the default log level of "warn"

WIP for the Bump-Nessie-to-0.5-in-Iceberg-PR is here: https://github.com/snazy/iceberg/tree/nessie-0.5 (it uses a locally mvn-install'd 0.4.1-SNAPSHOT)

The "nessie-quarkus" configuration in Iceberg's `build.gralde` *must* have a `nessieQuarkusRunner(enforcedPlatform("io.quarkus:quarkus-bom:1.13.0.Final"))` dependency, otherwise dependency hell will break lose --> latest 0.4.1-SNAPSHOT: all HTTP/POST requests have a completely empty response. HTTP/GET requests work fine - also the web-UI works fine. I suspect, that some dependency-version bump in latest Nessie conflicts with some libraries that come in via Iceberg or Gradle or whatever and causes this weird behavior. Netty had two conflicting versions (4.1.49 + 4.1.61), some javax-APIs and jackson deps in conflicting versions.

Fixes #1022

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1042)

### DIFF
--- a/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
+++ b/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
@@ -68,7 +68,7 @@ public class QuarkusApp extends org.projectnessie.quarkus.maven.QuarkusApp {
 
   public static AutoCloseable newApplication(Configuration configuration, Project project, Properties props) {
 
-    Configuration deploy = project.getConfigurations().create("quarkusAppDeploy");
+    Configuration deploy = project.getConfigurations().create("nessieQuarkusDeploy");
     final AppModel appModel;
 
     appModel = convert(configuration, deploy, props);
@@ -197,9 +197,8 @@ public class QuarkusApp extends org.projectnessie.quarkus.maven.QuarkusApp {
   }
 
   private static AppArtifact toDependency(Dependency dependency) {
-    AppArtifact artifact = new AppArtifact(dependency.getGroup(), dependency.getName(), null,
+    return new AppArtifact(dependency.getGroup(), dependency.getName(), null,
       "jar", dependency.getVersion());
-    return artifact;
   }
 
   private static AppDependency toDependency(ResolvedArtifact dependency) {
--- a/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppExtension.java
+++ b/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppExtension.java
@@ -15,16 +15,41 @@
  */
 package org.projectnessie.quarkus.gradle;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.gradle.api.Project;
 import org.gradle.api.provider.MapProperty;
+import org.gradle.api.provider.Property;
 
+@SuppressWarnings("UnstableApiUsage") // omit warning about `Property`+`MapProperty`
 public class QuarkusAppExtension {
   private final MapProperty<String, Object> props;
-  public QuarkusAppExtension(Project project) {
-    props = project.getObjects().mapProperty(String.class, Object.class);
+  private final Property<String> nativeBuilderImage;
+
+  private static Map<String, Object> defaultProps() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("quarkus.http.test-port", 0);
+    return map;
   }
 
-  public MapProperty<String, Object> getProps() {
+  public QuarkusAppExtension(Project project) {
+    props = project.getObjects().mapProperty(String.class, Object.class).convention(defaultProps());
+
+    // This is not doing anything with Docker or building a native image, just a quirk of Quarkus since 1.10.
+    nativeBuilderImage = project.getObjects().property(String.class).convention("quay.io/quarkus/ubi-quarkus-native-image:21.0.0-java11");
+  }
+
+  public MapProperty<String, Object> getPropsProperty() {
     return props;
+  }
+
+  public Property<String> getNativeBuilderImageProperty() {
+    return nativeBuilderImage;
+  }
+
+  @SuppressWarnings("unused") // convenience method
+  public void setNativeBuilderImage(String nativeBuilderImage) {
+    this.nativeBuilderImage.set(nativeBuilderImage);
   }
 }
--- a/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
+++ b/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
@@ -18,27 +18,58 @@ package org.projectnessie.quarkus.gradle;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.tasks.testing.Test;
 
+@SuppressWarnings("Convert2Lambda") // Gradle complains when using lambdas (build-cache won't wonk)
 public class QuarkusAppPlugin implements Plugin<Project> {
+
+  static final String START_TASK_NAME = "nessie-quarkus-start";
+  static final String STOP_TASK_NAME = "nessie-quarkus-stop";
+  static final String EXTENSION_NAME = "nessieQuarkusApp";
+  static final String CONFIG_NAME = "nessieQuarkusRunner";
 
   @Override
   public void apply(Project target) {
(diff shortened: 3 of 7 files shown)


snazy:
Review notes:
* mvn-clean-install this PR
* checkout https://github.com/snazy/iceberg/tree/nessie-0.5 + run `./gradlew :iceberg-nessie:test` to verify

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
```diff
@@ -68,7 +68,7 @@ protected QuarkusApp(RunningQuarkusApplication runningApp) {
 
   public static AutoCloseable newApplication(Configuration configuration, Project project, Properties props) {
 
-    Configuration deploy = project.getConfigurations().create("quarkusAppDeploy");
+    Configuration deploy = project.getConfigurations().create("nessieQuarkusDeploy");
```
rymurr, line 71: nit: the original `quarkusAppDeploy` was there as this could theoretically be reusable and pushed to Quarkus in the future. Is the change signifying that this code is un-reusable?

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppExtension.java
```diff
@@ -15,16 +15,41 @@
...
+    map.put("quarkus.http.test-port", 0);
+    return map;
+  }
+
   public QuarkusAppExtension(Project project) {
-    props = project.getObjects().mapProperty(String.class, Object.class);
+    props = project.getObjects().mapProperty(String.class, Object.class).convention(defaultProps());
+
+    // This is not doing anything with Docker or building a native image, just a quirk of Quarkus since 1.10.
+    nativeBuilderImage = project.getObjects().property(String.class).convention("quay.io/quarkus/ubi-quarkus-native-image:21.0.0-java11");
```
rymurr, line 40: does this need to be updated on every quarkus release?

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
```diff
@@ -18,27 +18,58 @@
...
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.tasks.testing.Test;
 
+@SuppressWarnings("Convert2Lambda") // Gradle complains when using lambdas (build-cache won't wonk)
 public class QuarkusAppPlugin implements Plugin<Project> {
 
+  static final String START_TASK_NAME = "nessie-quarkus-start";
+  static final String STOP_TASK_NAME = "nessie-quarkus-stop";
+  static final String EXTENSION_NAME = "nessieQuarkusApp";
+  static final String CONFIG_NAME = "nessieQuarkusRunner";
```
rymurr, line 31: similar as comment above

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
```diff
@@ -18,27 +18,58 @@
...
+    QuarkusAppExtension extension = target.getExtensions().create(EXTENSION_NAME, QuarkusAppExtension.class, target);
 
-    final Configuration config = target.getConfigurations().create("quarkusAppRunnerConfig")
-      .setVisible(false)
-      .setDescription("The config for the Quarkus Runner.");
+    Configuration config = target.getConfigurations().create(CONFIG_NAME).setDescription("The config for the Nessie-Quarkus Runner.");
 
-    target.getTasks().register("quarkus-start", StartTask.class, new Action<StartTask>() {
+    // Cannot use the task name "test" here, because the "test" task might not have been registered yet.
+    // This `withType(Test.class...)` construct will configure any current and future task of type `Test`.
```
rymurr, line 40: current and future for the module or the whole project?

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/StopTask.java
```diff
@@ -29,7 +29,8 @@ public StopTask() {
   public void stop() {
 
     if (application == null) {
-      getLogger().warn("No application found.");
+      getLogger().debug("No application found.");
```
rymurr, line 32: Shouldn't this be higher priority? If stop is being run w/o an application doesn't that indicate an error?

snazy: It would always log a warning, if the tests don't run, because we cannot make the start+stop tasks cacheable, because you cannot make a task's execution depend on another task execution.

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
snazy: TBH, this configuration isn't actually be accessible, because you create it when the task executes. Not sure whether that's still possible in newer Gradle versions.

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppExtension.java
snazy: Don't know - but it's configurable. It's the quirk from the last Quarkus version bump.

rymurr: Thinking of the deploy doc. I think there are a handful of versions to update by hand in the gradle plugin. Might as well add this one to the list.

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
snazy: It's the initialization order of plugins (the Gradle java-(base-)plugin isn't guaranteed to be configured when our plugin configures).

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
snazy: I've renamed those to make clear that this comes from the nessie-quarkus-gradle plugin and not from the "original" Quarkus plugin, which seems to have changed quite a bit recently. The Iceberg build script has no more references to any of these.

tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppPlugin.java
rymurr: does this mean quarkus-start binds to all modules in the project? Sorry...this is my gradle naivete showing

snazy: Don't know what you're referring to ;)

Gradle plugins add their tasks - and the order of when plugins are initialized isn't "that deterministic".

snazy: IIRC There are ways to add "plugin A depends on plugin B", but I ran out of time for the release. :(

rymurr (approved):

PR #1043, opened 2021-04-06
Correct OpenAPI and add examples

### LINKED ISSUE #1004: OpenAPI spec isn't entirely correct
Need another pass of the OpenApi spec, noticed the following:

* missing return error codes
* undocumented return error codes
* add clearer and more complete docs

rymurr:
fixes #1004

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1043)


snazy (approved):
Haven't verified every little aspect of the examples, but the change LGTM overall, so +1

PR #1049, opened 2021-04-07
Add changeId to commit metadata.

rymurr:
This very simple change adds a `change id` to commit metadata.

This is meant to be distinct from the commit id in that it will never change even over rebases and merges. This is primarily used to uniquely identify an object in the store. The tuple (CommitMeta.changeId,Contents.uuid) is immutable for a commit and unique across the database.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1049)


rymurr:
The only not-great thing about this change is that the change-id isn't searchable. To find a unique (changeId,uuid) pair you have to first find teh change id in the log then search that specific commit for the uuid. A potentially expensive operation

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1049)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 108 at r1](https://reviewable.io/reviews/projectnessie/nessie/1049#-MXj_-ujAZ6sr3duhjpE:-MXj_-ujAZ6sr3duhjpF:b-cs68xj) ([raw file](https://github.com/projectnessie/nessie/blob/858f9d52c2ba6c96ed27882e1a4f6b928a28f0fe/model/src/main/java/org/projectnessie/model/CommitMeta.java#L108)):*
> ```Java
>    */
>   @Nullable
>   public abstract String getChangeId();
> ```

I would have assumed it to be part of the `Content` object, not of the commit meta. I'm not sure how it could help 3rd party link a object back as they would not access the commit meta.

model/src/main/java/org/projectnessie/model/CommitMeta.java
```diff
@@ -97,6 +97,16 @@
   @JsonDeserialize(using = InstantDeserializer.class)
   public abstract Instant getAuthorTime();
 
+
```
snazy, line 100: Nit: superfluous newline

snazy (changes requested):
The change LGTM, but can you add a test?

rymurr (commented):
sure. Is there anything specific you wanted to see in the test? That changeId doesn't change through commits or?

> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1049#-:-MXl73eRBCCP9VabbSg0:b-tue2no)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @laurentgo)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 108 at r1](https://reviewable.io/reviews/projectnessie/nessie/1049#-MXj_-ujAZ6sr3duhjpE:-MXl7Bar6DGATdt7JWGn:b-dq2osm) ([raw file](https://github.com/projectnessie/nessie/blob/858f9d52c2ba6c96ed27882e1a4f6b928a28f0fe/model/src/main/java/org/projectnessie/model/CommitMeta.java#L108)):*

I would have assumed it to be part of the `Content` object, not of the commit meta. I'm not sure how it could help 3rd party link a object back as they would not access the commit meta.
</blockquote></details>

Making it part of the contents feels weird. So far nothing in `Contents` relates to the VCS nature of Nessie and it feels like it shouldn't. Putting it on the `Contents` makes it feel like `lastChangeTime` or `lastChangeId` or something. Maybe I misunderstood the original intention?

snazy:
> sure. Is there anything specific you wanted to see in the test? That changeId doesn't change through commits or?

Just a plain, stupid test that asserts that the validation works.

rymurr:
@snazy @laurentgo any thoughts on this iteration?

rymurr:
Hey @laurentgo and @snazy  I have updated `site/docs/develop/spec.md` to clearly define the goal of the code. Can we focus the review on that document and when the specification is decided I will adjust the code accordingly

servers/quarkus-server/src/test/java/org/projectnessie/server/ContentsCustomMatcher.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.util.Objects;
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+import org.projectnessie.model.Contents;
+import org.projectnessie.model.IcebergTable;
+
+public class ContentsCustomMatcher  extends TypeSafeMatcher<Contents> {
```
snazy, line 26:
```suggestion
public class ContentsCustomMatcher extends TypeSafeMatcher<Contents> {
```

servers/quarkus-server/src/test/java/org/projectnessie/server/RestGitTest.java
```diff
@@ -192,4 +191,14 @@ public void testOptimisticLocking() {
     Assertions.assertNotEquals(b3.getHash(), newHash);
   }
 
+  static void assertEquals(Contents expected, Contents actual) {
```
snazy, line 194: Mind renaming it to assertContentsEquals so it doesn't conflict with Assertions.assertEquals?

servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java
```diff
@@ -546,4 +551,9 @@ void unwrap(Executable exec) throws Throwable {
       throw targetException;
     }
   }
+
+  private static void assertEqualsContentsKey(ContentsWithKey expected, ContentsWithKey actual) {
```
snazy, line 555: Super nit: maybe assertContentWithKeyEquals (sounds "better" in my German ears)

site/docs/develop/spec.md
```diff
@@ -26,6 +28,35 @@ There are several expectations on this field:
...
 
+#### PermanentId
+
+All contents object must have a `permanentId` field. This field is set at commit time and is unique for the commit. All objects in
+the commit will have the same `permanentId`. This field would not be updated on merges or transplants therefore the id is unique for
+that commit permanently. By convention it is a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier) though
+this is not enforced by the Specification. There are several expectations on this field:
+
+1. `permanentId`  will always be updated at commit time.
+1. The `permanentId` will not be updated during any transplant or merge operations.
```
snazy, line 39:
```suggestion
1. `permanentId` will not be updated during any transplant or merge operations.
```

site/docs/develop/spec.md
```diff
@@ -26,6 +28,35 @@ There are several expectations on this field:
...
+1. All objects in the same original commit will have the same `permanentId`
+1. This has **nothing** to do with the commit hash.
```
snazy, line 41:
```suggestion
1. `permanentId` is *not* related or derived from the commit-hash.
```

site/docs/develop/spec.md
```diff
@@ -26,6 +28,35 @@ There are several expectations on this field:
...
+
+There is no API to look up an object by `permanentId` and the intention of an `permanentId` is not to serve in that capacity.
+An example usage of the `permanentId` field might be storing auxiliary data on an object in a local cache and using
```
snazy, line 44:
```suggestion
An example usage of the `permanentId` field is storing auxiliary data on an object in a local cache and using
```

site/docs/develop/spec.md
```diff
@@ -26,6 +28,35 @@ There are several expectations on this field:
...
+`permanentId` to look up that auxiliary data.
+
+!!! note
+    A note about caching. The `Contents` objects are likely to be cached locally by services using Nessie. There is also likely
+    to be auxiliary data (eg schema, partitions etc) stored by services which refer to a specific `Contents` at a specific
+    commit or time. The tuple `(id, permanentId)` uniquely reference an object in the Nessie history and is a suitable key
```
snazy, line 50: Can you elaborate where "id" comes from?

model/src/main/java/org/projectnessie/model/CommitMeta.java
```diff
@@ -97,6 +97,15 @@
   @JsonDeserialize(using = InstantDeserializer.class)
   public abstract Instant getAuthorTime();
 
+  /**
+   * Unique id for a change. This is distinct from a hash in that it doesn't change if rebased/merged etc.
+   *
+   * <p>Primary use is to uniquely identify objects in the Nessie Database. This id + the {@code Contents} id is a unique
+   * object in the database. This value is set by the store and will be overwritten if set at commit time by the client.
```
snazy, line 104:
```suggestion
   * This value is set by the store and will be overwritten if set at commit time by the client.
   * Although the {@code Contents.id} + this permanent-id is unique, there is intentionally no way to lookup this tuple, see Nessie project documentation for details.</p>
```

model/src/main/java/org/projectnessie/model/Contents.java
```diff
@@ -76,6 +76,17 @@ public String getId() {
     return UUID.randomUUID().toString();
   }
 
+  /**
+   * Unique change id for this object.
+   *
+   * <p>This id is uniquely refers to a specific commit. It is an id generated at commit time and is unique across all objects in that
+   * commit. It has nothing to do with the actual commit hash and is only used to find a unique object in the Nessie history.
```
snazy, line 83:
```suggestion
   * commit. It has nothing to do with the actual commit hash and is only used to find a unique object in the Nessie history.</p>
```

servers/quarkus-server/src/test/java/org/projectnessie/server/RestGitTest.java
rymurr: fixed

servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java
rymurr: fixed

site/docs/develop/spec.md
rymurr: fixed

rymurr:
rebased against main and all of @snazy suggestions taken into account.

snazy (dismissed):

laurentgo (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1049)* status: 0 of 9 files reviewed, 13 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/CommitMeta.java, line 108 at r1](https://reviewable.io/reviews/projectnessie/nessie/1049#-MXj_-ujAZ6sr3duhjpE:-MZBbtAD-bvKe2Iq5ELc:b-jzmef3) ([raw file](https://github.com/projectnessie/nessie/blob/858f9d52c2ba6c96ed27882e1a4f6b928a28f0fe/model/src/main/java/org/projectnessie/model/CommitMeta.java#L108)):*

Making it part of the contents feels weird. So far nothing in `Contents` relates to the VCS nature of Nessie and it feels like it shouldn't. Putting it on the `Contents` makes it feel like `lastChangeTime` or `lastChangeId` or something. Maybe I misunderstood the original intention?
</blockquote></details>

There are possibly 2 different concepts I guess:
- one to represent the change intent
- one to represent the result of the change at the dataset level

This would take care of the first point and this id would stay identical during a rebase or a cherry pick. But the data resulting of the change may not stay identical though.

I was more interested into the second point as a way for external system like a caching layer to identify changes to the data itself. The hash of the commit could be used for it, but one would need ideally the hash of the commit where the change to the data was made which is not an information Nessie has handy I guess. Hence my suggestion to have some id at the Content object itself...
___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 83 at r2](https://reviewable.io/reviews/projectnessie/nessie/1049#-MYLIVS0x2UIO35NKx9Y-r2-83:-MZBeET_6EJI8YLSe-yW:bqa53w6) ([raw file](https://github.com/projectnessie/nessie/blob/5ad7409db13fbb4376d62a9f2ca71667d6d0d3aa/model/src/main/java/org/projectnessie/model/Contents.java#L83)):*

```suggestion
   * commit. It has nothing to do with the actual commit hash and is only used to find a unique object in the Nessie history.</p>
```
</blockquote></details>

The javadoc is kind of confusing me to be honest because cannot we use a commit hash to find a unique object in the Nessie history?
Since the name of this field is the same as for `CommitMeta`, I would imagine there is a relationship between those 2, but it is not explained. I think we should also expand on what are the rules for this id for the various Nessie operations (something I started elaborating on in my comment above)

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1049)* status: 0 of 9 files reviewed, 14 unresolved discussions (waiting on @rymurr and @snazy)

___
*[model/src/main/java/org/projectnessie/model/Contents.java, line 83 at r2](https://reviewable.io/reviews/projectnessie/nessie/1049#-MYLIVS0x2UIO35NKx9Y-r2-83:-MZBfMpw0e6eCJTlH4gC:b-muutec) ([raw file](https://github.com/projectnessie/nessie/blob/5ad7409db13fbb4376d62a9f2ca71667d6d0d3aa/model/src/main/java/org/projectnessie/model/Contents.java#L83)):*

The javadoc is kind of confusing me to be honest because cannot we use a commit hash to find a unique object in the Nessie history?
Since the name of this field is the same as for `CommitMeta`, I would imagine there is a relationship between those 2, but it is not explained. I think we should also expand on what are the rules for this id for the various Nessie operations (something I started elaborating on in my comment above)
</blockquote></details>

Got some issue with Reviewable and so didn't see the extra changes after this file. 
___
*[site/docs/develop/spec.md, line 31 at r6](https://reviewable.io/reviews/projectnessie/nessie/1049#-MZBgOh67l88waVrLkOI:-MZBgOh67l88waVrLkOJ:b-3d7mcs) ([raw file](https://github.com/projectnessie/nessie/blob/6e89bfb1b25710ae7a0642e7f82100afa6dac84a/site/docs/develop/spec.md#L31)):*
> ```Markdown
> of the `id` field might be storing auxiliary data on an object in a local cache and using `id` to look up that auxiliary data.
> 
> #### PermanentId
> ```

What I believe is missing in that definition is what this id conceptually represents, and without it it is hard to explain/justify the rules about merge and transplant

snazy (approved):
Just make sure that the commit message+summary get adjusted.

LGTM

PR #1064, opened 2021-04-09
Improve dependency ordering in the Gradle plugin

snazy:
Since we have to add the quarkus-bom as an `enforcedPlatform()`-dependency in Gradle build script,
there were two dependencies instead of one handled by the apprunner-gradle-plugin. The Gradle plugin
accidentally used a `j.u.HashSet`, which has a non-deterministic iteration order (can change from
one JVM launch to another). Since all tests were run against the same Gradle Daemon instance, and
that JVM instance caused the "right" iteration order --> returning the nessie-quarkus-server dependency
before the quarkus-bom dependency - everything was fine.

This PR changes `j.u.HashSet` to `j.u.LinkedHashSet` but also splits the Gradle configuration into two:
one only for the application artifact (nessie-quarkus-server) and one for the runtime dependencies
(quarkus-bom), so it's easier for users to reason about the difference there.

Also adds some logging at info and debug levels.

Removed unnecessary "input-parameter" from `StartTask`, which isn't cacheable at all.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1064)

### DIFF
--- a/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
+++ b/tools/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusApp.java
@@ -16,6 +16,9 @@
 package org.projectnessie.quarkus.gradle;
 
 import static io.quarkus.bootstrap.resolver.maven.DeploymentInjectingDependencyVisitor.toArtifact;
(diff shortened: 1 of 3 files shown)


snazy:
Ran a test locally, 20 iterations against "fresh" Gradle Daemons and it works fine.

rymurr (approved):

PR #1105, opened 2021-04-21
fix noise from jacoco

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1105)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1093,6 +1093,7 @@ limitations under the License.
                 <configuration>
                   <dataFileIncludes>
                     <dataFileInclude>**/jacoco.exec</dataFileInclude>
+                    <dataFileInclude>**/jacoco-quarkus.exec</dataFileInclude>
                   </dataFileIncludes>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -1093,6 +1093,7 @@ limitations under the License.
                 <configuration>
                   <dataFileIncludes>
                     <dataFileInclude>**/jacoco.exec</dataFileInclude>
+                    <dataFileInclude>**/jacoco-quarkus.exec</dataFileInclude>
```
snazy, line 1096: The Quarkus extension just writes `jacoco.exec`, do we need this one?

rymurr: https://quarkus.io/guides/tests-with-coverage#running-the-tests-with-coverage it actually doesn't. We can change it to write `jacoco.exec` if this test works though, and remove this line.

rymurr:
Increased coverage and reduced the logs by 5000 lines (80%)

pom.xml
snazy: Ah, right.

snazy (approved):

PR #1113, opened 2021-04-22
add unit tests for SerDe of commit messages and contents

### LINKED ISSUE #1093: More complete unit tests of TableCommitMetaStoreWorker
As per #1023 we have to add a more comlete set of tests on `TableCommitMetaStoreWorker`. Particularly the serde ops and any edge cases w/ the serde

rymurr:
fixes #1093

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1113)

### DIFF
--- a/model/src/main/java/org/projectnessie/model/HiveTable.java
+++ b/model/src/main/java/org/projectnessie/model/HiveTable.java
@@ -24,7 +24,6 @@ import org.immutables.value.Value;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import com.google.errorprone.annotations.Var;
 
 @Value.Immutable(prehash = true)
 @JsonSerialize(as = ImmutableHiveTable.class)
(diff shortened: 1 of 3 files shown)


model/src/main/java/org/projectnessie/model/HiveTable.java
```diff
@@ -33,4 +36,50 @@
...
   public abstract List<byte[]> getPartitions();
 
+  /**
+   * Because of the List of byte arrays Immutables doesn't generate a correct hashcode.
+   *
+   * <p> We hand compute the List elements to ensure the byte arrays hash codes are consistent.
+   */
+  @Override
+  public int hashCode() {
+    int h = 5381;
```
snazy, line 46: I'd prefer the "standard" way common in the JDK. I.e. initialize with `1` and use `h = (31 * h) + xyz`, like `java.util.Arrays#hashCode(java.lang.Object[])` does.

snazy (changes requested):

model/src/main/java/org/projectnessie/model/HiveTable.java
rymurr: cool, fixed. That was a direct copy from immutables but I agree, we should stick to the standard

snazy (approved):

laurentgo (approved):

PR #1139, opened 2021-04-26
fix jacoco for test classes

rymurr:
By default all abstract test classes that were packaged w/o the `tests` classifier get added to our total coverage count. Since these are Test classes they do not get coverage info. This PR excludes them from our total test coverage stats to give a better estimation of our actual code coverage on active code.

This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1139)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1099,6 +1099,12 @@ limitations under the License.
                   <dataFileIncludes>
                     <dataFileInclude>**/jacoco.exec</dataFileInclude>
                   </dataFileIncludes>
+                  <excludes>
+                    <exclude>org.projectnessie.versioned.tests.AbstractITVersionStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractITTieredVersionStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractTestStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractTieredStoreFixture</exclude>
+                  </excludes>
                   <outputDirectory>${project.reporting.outputDirectory}/jacoco-aggregate</outputDirectory>
                 </configuration>
               </execution>


laurentgo:
Can you detail what's the issue?

pom.xml
```diff
@@ -1099,6 +1099,12 @@ limitations under the License.
                   <dataFileIncludes>
                     <dataFileInclude>**/jacoco.exec</dataFileInclude>
                   </dataFileIncludes>
+                  <excludes>
+                    <exclude>org.projectnessie.versioned.tests.AbstractITVersionStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractITTieredVersionStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractTestStore</exclude>
+                    <exclude>org.projectnessie.versioned.impl.AbstractTieredStoreFixture</exclude>
```
snazy, line 1106: Should `org.projectnessie.versioned.impl.SampleEntities` be added as well (b/c this is basically a list of all classes in versioned-tests and versioned-tiered-tests)?

snazy (commented):
Looks like the exclusion did fix the "artificial code-coverage reduction".
I'm not a fan of manually mentioning those exclusions, but I don't see a better way at the moment.

pom.xml
rymurr: Since they are not actual test classes there is sensible code coberage for those classes (bumps our coverage ;-)

snazy (approved):
OKay, then go for it.
Can you open an issue that we re-visit this once Quarkus 1.14 (or whatever comes next) is there, so we can maybe remove this? Not sure if having that other Quarkus-jacoco-PR in would help, but who knows.

rymurr:
I don't think this is a quarkus thing. This is a `Test` class in `main` jar

PR #1243, opened 2021-05-13
Gradle plugin: proper fix to run multiple gradle test tasks + support "reducing the log level"

snazy:
The previous fix #1221 to support multiple Gradle `Test` tasks unfortunately did only fix it for the Gradle configuration-phase, but actually running both a `:iceberg-nessie:test` plus `:iceberg-nessie:integrationTest` tasks failed. The fix is to "just" not start Quarkus twice.

This PR also adds a new `systemProperties` property to the `QuarkusAppExtension`, to set additional system-properties, which is necessary to reduce the log level as requested by https://github.com/apache/iceberg/issues/2557

On top, it restores the previous system-properties when the `StopTask` is executed.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1243)


snazy:
Note: a draft for a Iceberg-Spark-Nessie integration test is here: https://github.com/snazy/iceberg/tree/nessie-spark-integration-test

nastra (approved):

PR #1249, opened 2021-05-14
Handle `NessieConflictException` in Delta clients + handling of multiple-branches

snazy:
The Delta clients can run into an unrecoverable (dead) state, when the reference changed in the mean time.
This commit adds a single retry in case the client throws a `NessieConflictException`.

The Delta-Client kept a single `Reference` that was initialized with the named-reference given
in the hadoop-config. However, if the configuration changed to switch to another Nessie branch,
that `Reference` was never refreshed and resulted into `NessieNotFoundException`s.

The change is to replace that single `Reference` field with a map of ref-name to `Reference`
to support multiple branches.

Also adds tests to verify that the Delta client works with multiple branches and the retry
when hitting a `NessieConflictExcepiton`.

The PR also changes the three delta-modules to include the resources from the -core module for tests.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1249)


nastra (approved):

clients/deltalake/core/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
```diff
@@ -51,11 +52,17 @@ class NessieLogStore(sparkConf: SparkConf, hadoopConf: Configuration)
     NessieClient.builder().fromConfig(c => hadoopConf.get(c)).build()
   }
 
-  private def getOrCreate(): Reference = {
+  /**
+   * Keeps a mapping of reference name to current hash.
+   */
+  private val referenceMap: util.Map[String, Reference] = {
```
rymurr, line 58: why a `java` map and not a `scala` map?

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
```diff
@@ -51,6 +57,91 @@ protected static void createDelta() {
...
+
+    Dataset<Row> count3 = spark.sql("SELECT COUNT(*) FROM test_multiple_branches");
+    Assertions.assertEquals(35L, count3.collectAsList().get(0).getLong(0));
+  }
+
+  @Test
+  // Delta < 0.8 w/ Spark 2.x doesn't support multiple branches well (warnings when changing the configuration)
+  @DisabledIfSystemProperty(named = "skip-multi-branch-tests", matches = "true")
+  void testCommitRetry() throws Exception {
+    String csvSalaries1 = ITDeltaLog.class.getResource("/salaries1.csv").getPath();
```
rymurr, line 104: Any reason to use csvs when everywhere else in spark tests we create simple few row in memory datasets?

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
```diff
@@ -51,6 +57,91 @@ protected static void createDelta() {
         .set("spark.sql.catalog.spark_catalog", "org.apache.spark.sql.delta.catalog.DeltaCatalog");
   }
 
+  @Test
+  // Delta < 0.8 w/ Spark 2.x doesn't support multiple branches well (warnings when changing the configuration)
```
rymurr, line 61: for my benefit can you explain a bit more? If i recall correctly `0.6.0` is Spark2 only and `>0.7.0` is Spark3 only.

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
```diff
@@ -61,6 +64,8 @@ protected void create() throws IOException {
                         .config(conf)
                         .getOrCreate();
     spark.sparkContext().setLogLevel("WARN");
+
+    nessieClient = NessieClient.builder().withUri(url).build();
```
rymurr, line 68: nit: can we rename the `url` parameter

rymurr (approved):

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
snazy: that's delta/spark2 - it complains with a warning that the configuration has changed and "might not have been applied completely".

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
snazy: A mixture of "laziness and reproduction of the original issue". I wanted this test to be close to what the demo in `nessie-demos` looks like.

rymurr (approved):

clients/deltalake/core/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
snazy: Just need a mutable map here - and using a Java `HashMap` felt to be the easiest way to go.

PR #1256, opened 2021-05-17
Remove Nullable on Contents#getId()

### LINKED ISSUE #1054: Remove nullable annotation on Contents.getId

nastra:
Fixes #1054

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1256)

### DIFF
--- a/model/src/main/java/org/projectnessie/model/Contents.java
+++ b/model/src/main/java/org/projectnessie/model/Contents.java
@@ -20,8 +20,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 
-import javax.annotation.Nullable;
-
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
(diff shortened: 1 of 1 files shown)


model/src/main/java/org/projectnessie/model/Contents.java
```diff
@@ -66,11 +64,7 @@
    *
    * <p>This id is unique for the entire lifetime of this Contents object and persists across renames. Two contents with the same key
    * will have different id.
-   *
```
rymurr, line 69: What happens if we run a `>0.6.0` server w/ the `0.11.1` iceberg? Does this need to wait for an iceberg release?

snazy: It already does not work, because the server sends the `id` field, that the Nessie client in Iceberg does not understand and fails.

rymurr: Ok raised #1310 to address more rigorously.

rymurr (approved):

PR #1260, opened 2021-05-18
Ignore unknown fields in JSON objects

snazy:
JSON object deserialization fails when it tries to deserialize an unknown field.
New fields can, and have been, added in newer Nessie versions, which causes older clients
to fail. For example, the `Contents.id` field added in Nessie 0.6.0 made Nessie 0.5.1
clients incompatible.

This change consists of the Java related changes, which basically lets all model-types
extend/implement `Base`, which got the `@JsonIgnoreProperties(ignoreUnknown=true)` annotation.
A unit test that covers hopefully all types verifies the behavior by artificually injecting
additional fields into the JSON objects.

The Python changes are a less nice, as `unknown=EXCLUDE` had to be added to a lot of places.
The generated JSON from the Java tests has been used to add new Python unit tests. A few
functional changes have been made to fix existing issues in the `pynessie` library:
* Fix `data_key` for attributes in `DeltaLakeTable`
* Add missing `HiveTable`

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1260)

### DIFF
--- a/model/src/main/java/org/projectnessie/model/Base.java
+++ b/model/src/main/java/org/projectnessie/model/Base.java
@@ -15,5 +15,8 @@
  */
 package org.projectnessie.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
 /** base interface for all model/api classes. */
+@JsonIgnoreProperties(ignoreUnknown = true)
 public interface Base {}
--- a/model/src/main/java/org/projectnessie/model/CommitMeta.java
+++ b/model/src/main/java/org/projectnessie/model/CommitMeta.java
@@ -34,7 +34,7 @@ import org.immutables.value.Value;
 @Value.Immutable(prehash = true)
 @JsonSerialize(as = ImmutableCommitMeta.class)
 @JsonDeserialize(as = ImmutableCommitMeta.class)
-public abstract class CommitMeta {
+public abstract class CommitMeta implements Base {
 
   /**
    * Hash of this commit.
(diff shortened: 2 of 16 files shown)


snazy:
As discussed offline, there are pros and cons about this approach.
On the plus side, this approach makes REST-API-model changes technically less prone to incompatibilities. This is also the downside, because ignoring fields on either the client or server side can lead to functionally wrong/unexpected results.
There's probably more to the whole "backwards compatibility spiel" and all that requires more thoughts. Maybe we also need some way to let clients & servers announce the API version they're running (just a thought).

nastra (dismissed):

snazy:
related: #1310

rymurr:
This feels a bit loose. 

* if a field isn't needed by a client and the server sends it this will ensure everythign works
* if a field is needed by the server and the cclient doesn't send it we get weird errors or silent corruption
* if a field isn't needed by a client but is important for further work w/ the server then it will still fail further into the process

Before we add a handful of minor fixes (that we are unlikely to remove) for compatibility issues I would like discuss how we will deal w/ compatibility between server and client going forward. Will raise on todays standup.

snazy:
Yep - that's my impression as well. Neither ignoring fields nor version-checks (#1310) are great solution here atm.
Maybe we bite the bullet until we have a "Nessie 1.0 GA". After that, we can decide which route we want to go. Could also be that we stick with "classic URL versioning" (https://host/api/v1/... - https://host/api/v2/...) or so.

PR #1271, opened 2021-05-20
Introduce AssertJ for better test code readability

nastra:
I've had really good experiences in the past with [AssertJ](https://assertj.github.io/doc/) when it comes to testing as it's very expressive and makes the code easy to read.

I figured that I'll introduce it and rewrite a bunch of Test classes to show how it makes certain things easier to test.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1271)


laurentgo:
How does it compare with hamcrest (which we already have as a dependency?)

clients/client/pom.xml
```diff
@@ -102,6 +102,12 @@
       <version>1.68</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>${assertj.version}</version>
```
snazy, line 108: can `<version>` be removed here?

servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java
```diff
@@ -167,23 +159,25 @@ void referenceNames(String refNamePart) throws NessieNotFoundException, NessieCo
...
     Map<String, Reference> references = tree.getAllReferences().stream()
         .filter(r -> "main".equals(r.getName()) || r.getName().endsWith(refNamePart))
         .collect(Collectors.toMap(Reference::getName, Function.identity()));
-    assertThat(references.values(), containsInAnyOrder(
-        referenceWithNameAndType("main", Branch.class),
-        referenceWithNameAndType(tagName, Tag.class),
-        referenceWithNameAndType(branchName, Branch.class),
-        referenceWithNameAndType(branchName2, Branch.class)));
+
+    assertThat(references).containsAllEntriesOf(
```
snazy, line 176: Looks like this one removes the assertions on the type ("instanceof")

snazy (changes requested):
I think, we can then completely remove the hamcrest-dependency from the top-level `pom.xml`, if rest-assured is fine with that change.

servers/quarkus-server/src/test/java/org/projectnessie/server/TestRest.java
nastra: fixed

clients/client/pom.xml
nastra: yep, fixed

nastra:
> How does it compare with hamcrest (which we already have as a dependency?)

@laurentgo Writing AssertJ checks seems to be just easier and more readable when comparing to Hamcrest, as those require nested methods. Also checking exceptional code seems to be much easier with AssertJ than with Junit/Hamcrest.
AssertJ Dev seems to be quite active when looking at https://github.com/assertj/assertj-core/releases, whereas Hamcrest's last release was in Oct 2019.

snazy (dismissed):
LGTM

snazy (approved):

laurentgo (approved):

PR #1276, opened 2021-05-20
Attach quarkus-server-uber-runner-jar to github releases

snazy:
Also build a standalone uber-jar from nessie-quarkus and attach that to the GitHub release as well.

Add some notes to the GitHub release notes about the native binary and the uber-jar.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1276)

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -167,7 +167,6 @@ jobs:
 
     # Deploys Maven artifacts. Build and test steps were already ran in previous steps.
     # Not running tests, because the environment contains secrets.
-    # TODO switch autoReleaseAfterClose to true - see https://github.com/sonatype/nexus-maven-plugins/tree/master/staging/maven-plugin
     - name: Publish Maven artifacts for release
       id: build_maven
       env:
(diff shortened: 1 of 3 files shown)


laurentgo (commented):
Why not posting on maven central?

snazy:
> Why not posting on maven central?

Can't let the Quarkus-Maven-Plugin build two different package types at once (native image + uber-jar). It might work, with an additional execution of the Quarkus-Maven-Plugin, but I opted for the easier way for now, which feels consistent with the native image, which isn't published to Maven Central as well.

snazy:
Meh - the quarkus-maven-plugin neither supports building multiple package types at once nor can the package type specified in the plugin configuration (but only as a project-property).

However, attaching the uber-jar could work when we build the uber-jar first and then the native-image in a separate step.

rymurr:
Why do we want to deploy the uber jar at all? I agree w/ Laurent, to me this should be published to maven rather than the release page. After all its a maven artifact, we could add this to a pom and start a quarkus server in a test right?

snazy:
The uber-jar will be published to Maven Central with this PR.
Adding the uber-jar as a release-asset is a convenience thing for people that cannot run the native image, because their desktop OS is MacOS or their Linux distro is incompatible (glibc versions).

rymurr (approved):

PR #1325, opened 2021-05-27
Add server-side filters to /api/v1/trees/tree/{ref}/log

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1325)


python/pynessie/_log.py
```diff
@@ -31,22 +29,8 @@ def show_log(
 
     def generator() -> Generator[CommitMeta, Any, None]:
         for i in raw_log:
-            if committer and i.committer != committer:
```
nastra, line 34: note that this isn't currently tested in the existing `test_nessie_cli.py` tests because the `author` field can't be set when committing things and therefore is always empty. Once we have https://github.com/projectnessie/nessie/issues/1326 we can adjust the tests and make sure we can filter based on the `author` field

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -66,8 +66,8 @@ private StreamingUtil() {
    */
   public static Stream<CommitMeta> getCommitLogStream(@NotNull TreeApi treeApi, @NotNull String ref,
```
snazy, line 67: This method should get the new parameters as well.

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -233,7 +233,19 @@ LogResponse getCommitLog(
           Integer maxRecords,
       @Parameter(description = "pagination continuation token, as returned in the previous LogResponse.token")
       @QueryParam("pageToken")
-          String pageToken)
+          String pageToken,
+      @Parameter(description = "The author of a commit. This is the original committer")
+      @QueryParam("author")
+      String author,
+      @Parameter(description = "The entity that performed the commit")
```
snazy, line 240: Nit: "entity" feels a bit too abstract. Maybe "person or system"?

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -233,7 +233,19 @@ LogResponse getCommitLog(
...
+      @QueryParam("committer")
+      String committer,
+      @Parameter(description = "Commits newer than the specified date in ISO-8601 format")
```
snazy, line 243: Nit: "Only include commits ..."

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -233,7 +233,19 @@ LogResponse getCommitLog(
...
+      @QueryParam("after")
+          String after,
+      @Parameter(description = "Commits older than the specified date in ISO-8601 format")
```
snazy, line 246: Nit: "Only include commits ..."

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -233,7 +233,19 @@ LogResponse getCommitLog(
           Integer maxRecords,
       @Parameter(description = "pagination continuation token, as returned in the previous LogResponse.token")
       @QueryParam("pageToken")
-          String pageToken)
+          String pageToken,
+      @Parameter(description = "The author of a commit. This is the original committer")
```
snazy, line 237: Should mention that `null` means no filtering on the author.
(And similar for the other new parameters)

python/pynessie/_endpoints.py
```diff
@@ -180,13 +188,25 @@ def list_logs(
     :param max_result_hint: hint for the server, maximum number of results to return
     :param page_token: the token retrieved from a previous page returned for the same ref
     :param ssl_verify: ignore ssl errors if False
+    :param author: The author of a commit. This is the original committer
```
snazy, line 191: The comments on `TreeApi` about the parameter docs apply here as well.

python/pynessie/_log.py
```diff
@@ -18,6 +16,7 @@ def show_log(
     number: Optional[int] = -1,
     after: Optional[str] = None,
     before: Optional[str] = None,
+    author: Optional[str] = None,
```
snazy, line 19: Wonder whether we should remove the filtering parameter and use "Python kwargs" for the filtering parameters.

python/pynessie/_log.py
```diff
@@ -27,26 +26,12 @@ def show_log(
     Note:
         limiting by path is not yet supported.
     """
-    raw_log = nessie.get_log(start)
+    raw_log = nessie.get_log(start_ref=start, author=author, committer=committer, before=before, after=after)
 
     def generator() -> Generator[CommitMeta, Any, None]:
         for i in raw_log:
-            if committer and i.committer != committer:
```
snazy, line 34: Think, it's maybe better to keep this part and remove it in a future release. Not a big deal and I won't fight for not removing this.

python/pynessie/nessie_client.py
```diff
@@ -183,7 +183,14 @@ def cherry_pick(self: "NessieClient", branch: str, old_hash: Optional[str] = Non
         cherry_pick(self._base_url, branch, transplant_json, old_hash, self._ssl_verify)
 
     def get_log(
-        self: "NessieClient", start_ref: str, max_result_hint: Optional[int] = None, page_token: Optional[str] = None
+        self: "NessieClient",
+        start_ref: str,
+        max_result_hint: Optional[int] = None,
+        page_token: Optional[str] = None,
```
snazy, line 189: Similiar to the above comment: maybe use "Python kwargs" here as well?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -317,11 +337,23 @@ private static Reference makeRef(WithHash<? extends Ref> refWithHash) {
...
     } else if (o instanceof Operation.Unchanged) {
       return Unchanged.of(key);
     } else {
       throw new IllegalStateException("Unknown operation " + o);
     }
   }
+
+  @Nullable
+  private Instant parseToInstant(String instant) {
+    try {
```
snazy, line 350: Nit-picky:
Add the `if (instant == null) return null;` at the top.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -155,12 +158,29 @@ public void deleteBranch(String branchName, String hash) throws NessieConflictEx
...
+      @Nullable String after, @Nullable String before) throws NessieNotFoundException {
     int max = Math.min(maxEntriesHint != null ? maxEntriesHint : MAX_COMMIT_LOG_ENTRIES, MAX_COMMIT_LOG_ENTRIES);
     Hash startRef = getHashOrThrow(pageToken != null ? pageToken : ref);
+    final Instant afterInstant = parseToInstant(after);
+    final Instant beforeInstant = parseToInstant(before);
+
     try (Stream<ImmutableCommitMeta> s = getStore().getCommits(startRef).limit(max + 1)
         .map(cwh -> cwh.getValue().toBuilder().hash(cwh.getHash().asString()).build())) {
-      List<CommitMeta> items = s.collect(Collectors.toList());
+      Stream<ImmutableCommitMeta> commits = s;
```
snazy, line 170: I think, the algorithm here is wrong now. For example, if `max==10` and you filter out one or more commits below, the `if (items.size() == max + 1) {` will *not* trigger and return the current page as the *last* page so the client will not ask for more pages.

The filtering probably requires the removal of the `.limit(max + 1)` above and a slightly changed logic below.

snazy (commented):
Not a full review, but a first pass.

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: I "stole" those descriptions from `CommitMeta` but I agree we should improve them. Will follow up shortly

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
```diff
@@ -106,10 +106,15 @@ public Branch getDefaultBranch() {
 
   @Override
   public LogResponse getCommitLog(@NotNull String ref, @Nullable Integer maxEntriesHint,
-      @Nullable String pageToken) throws NessieNotFoundException {
+      @Nullable String pageToken, @Nullable String author, @Nullable String committer,
```
rymurr, line 109: can we use an object here rather than a `method(String, String, String, String, String)`? I thought there was somethign in JAX-RS that would even convert the object into query params. I think the method signature is real easy to get messed up.

Also do we want to support these as Lists? to fetch all commits by eg author1,author2,author3

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
rymurr: another reason to use a wrapper object

model/src/main/java/org/projectnessie/api/TreeApi.java
rymurr: +1 I believe the `Parameter` openapi arg has a default value flag?

python/pynessie/_log.py
rymurr: +1

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
laurentgo: I would actually more inclined to use some query language (something like https://opensource.google/projects/cel) in order to give some flexibility in term of querying

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
laurentgo: It might be more obvious if filtering is extracted out in some separate method/class and the logic can be truly unit tested?

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: fixed

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: fixed

python/pynessie/_endpoints.py
nastra: fixed

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: fixed

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
nastra: unfortunately there's no Java implementation for CEL :(

rymurr: I think we want to follow the git semantics as much as possible here. So looking at `git log --help` we should be able to do:

* before a date
* after a date
* multiple authors/committers and regex support for author/committer

We will eventually want to add regex to filter on log messages, file paths etc so my thinking is we need something more than adding `String` parameters to the function signature. Whether we need a full fledged query language for the query params at this stage is up for debate but if there is an easy to use lib it would be nice.

rymurr: nit: before and after should likely be date/instant etc objects

model/src/main/java/org/projectnessie/api/TreeApi.java
rymurr, line 240: I think the parameter descriptions should be a bit more descriptive. 

eg Filter the log on the author that specified the commit. Can take multiple authors and a regular expression

Similarily there should be a few words in the `@Operation` above to discuss the filters

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: I updated the description to mention that `null` means that we're not filtering based in this value. @rymurr what did you have in mind with a default value?

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
nastra: I agree having X parameters on a method is super annoying. I'll switch stuff to using `@BeanParam` and put all parameters into a Wrapper class.

> Also do we want to support these as Lists? to fetch all commits by eg author1,author2,author3

Not sure, maybe as a follow-up after this PR?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
nastra: agreed. I'll switch stuff to using `@BeanParam` and putting all parameters into a `CommitLogParams` class

model/src/main/java/org/projectnessie/api/TreeApi.java
rymurr: Ahh, I meant that _I think_ you can do `@Parameter(description="blah", default="empty, not filtering")`

I think this should be switched to a `List<String>` though. Empty list means return all author/committers (ie no filtering) and each entry in the list is a author in the set of authors to filter and return.We should make the entries in the List interpretable as regexs but maybe not in this patch

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
rymurr: > Not sure, maybe as a follow-up after this PR?

don't see a reason to not do it now.

model/src/main/java/org/projectnessie/api/TreeApi.java
nastra: fixed

python/pynessie/_log.py
nastra: fixed

python/pynessie/nessie_client.py
nastra: fixed

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: you're totally right :man_facepalming:

nastra: should be fixed now

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -0,0 +1,188 @@
...
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+
+import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.projectnessie.model.Validation;
+
+public class CommitLogParams {
```
snazy, line 35: Can't immutables help here?

nastra: Unfortunately not as I tried it already. It doesn't play nice with how this class is instantiated during a request

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
nastra: added support for filtering by multiple authors/committers

python/pynessie/_log.py
nastra: @rymurr wdyt? Should we keep this code for now?

rymurr: We should prob keep it w/ a note that it can be removed once the server has compatiblity guarantees. A new client on an old server wouldn't do any filtering if we remove it

python/pynessie/_endpoints.py
```diff
@@ -170,24 +170,15 @@ def list_tables(
...
+    :param filtering_args: All of the args used to filter the log
     :return: json dict of Nessie logs
     """
-    params = {}
-    if max_result_hint:
-        params["max"] = str(max_result_hint)
-    if page_token:
-        params["pageToken"] = page_token
-    return cast(dict, _get(base_url + "/trees/tree/{}/log".format(ref), ssl_verify=ssl_verify, params=params))
+    return cast(dict, _get(base_url + "/trees/tree/{}/log".format(filtering_args.pop("ref")), ssl_verify=ssl_verify, params=filtering_args))
```
rymurr, line 181: I think we should probably leave `ref` as a named arg here. If it isn't included (ie `filtering_args.pop("ref") == None`) then we would get an error from the server. We can prevent the error on this side by ensuring its present

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -0,0 +1,192 @@
...
+      + "No filtering will happen if this is set to null", examples = {@ExampleObject(ref = "javaInstant")})
+  @QueryParam("before")
+  private Instant before;
+
+  public CommitLogParams() {
+  }
+
+  private CommitLogParams(String ref, Integer maxRecords, String pageToken, List<String> authors, List<String> committers, Instant after,
+      Instant before) {
+    this.ref = ref;
```
rymurr, line 76: Why is ref part of the commit log Params? In my mind its part of the URL and is a key component of the call. all the other params in commit log params are optional/defaultable.

python/pynessie/_endpoints.py
nastra: sgtm

python/pynessie/_log.py
nastra: sgtm. fixed

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
nastra: the idea was that `CommitLogParams` just wraps all the params that exist in order to **query** & **filter** the commit log and that would include everything that is required or optional. The validation process of all parameters is still happening when the endpoint is being called. 
Or were you thinking that we might potentially forget to set `ref` via the `Builder` for some tests because we don't check if `ref` ends up being non-null? I guess for that case we could add some checking logic in the builder.

nastra: I think it makes sense to have some validation logic in the `Builder` itself, which I just added with some tests

rymurr: Im thinking of this from an API user standpoint. If the ref is in the object and I don't want any filtering then I am doing

``` java
CommitLogParams params = new CommitLogParams("ref");
handleResults(api.getLog(params));
```

Compared to:

``` java
handleResults(api.getLog("ref", null)); //or I add a helper for the no filtering case and drop the null argument
```

To me its clearer and more direct this way for a user. The `ref` is very different semantically from the optional args. @laurentgo  and @snazy you feel strongly either way?

PS I like the valildation logic though :-)

nastra: I'm with you and I totally get your point. I had the same thoughts at the beginning, meaning I was hesitant whether `ref` should go into `CommitLogParams` or not. 
At the end I decided it's probably better to have `ref` in `CommitLogParams`, since `CommitLogParams` isn't just for filtering, but rather for all params.
Therefore `handleResults(api.getLog(CommitLogParams.builder().ref("ref").build()))` seemed to be the better choice.

I guess we could also just add a helper where we can call `api.getLog("ref")`, which builds the `CommitLogParams` object and forwards the call to `getLog(CommitLogParams)`.

Thoughts on this?

rymurr: > At the end I decided it's probably better to have `ref` in `CommitLogParams`, since `CommitLogParams` isn't just for filtering, but rather for all params.

Not sure I understand what you mean. Maybe Im just not thinking all the way through though. All the others are for filtering right?

> I guess we could also just add a helper where we can call `api.getLog("ref")`, which builds the `CommitLogParams` object and forwards the call to `getLog(CommitLogParams)`.

I would rather keep it consistent. Either `getLog("ref")` and `getLog("ref", commitLogParams)` or `getLog(comitLogParams)`

nastra: What I mean is that `CommitLogParams` just wraps all the arguments to fetch the commit log and that includes mandatory and optional parameters. Does that make sense?

Just assume we have some endpoint `getXyz(...)` where we already have 2-3 mandatory params. And now we add a few more fields (mandatory & optional). We would have to adjust the respective method parameters everywhere where this method is being called/used. Using a wrapper class for **all** of those parameters seems like the better choice imo, as otherwise you would have a mix of mandatory parameters on the respective `getXyz(...)` method that would need adjusting, but also the optional params in the `XyzParams` class.

I don't feel strongly about having `ref` in `CommitLogParams`, so let's see what other people think.

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1325)* status: 0 of 16 files reviewed, 16 unresolved discussions (waiting on @laurentgo, @nastra, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/api/params/CommitLogParams.java, line 36 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMcXsz0XG3Jspj5qQ6:-MbMcXt-2gwmK149ZumE:b5fubzl) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java#L36)):*
> ```Java
> import org.projectnessie.model.Validation;
> 
> public class CommitLogParams {
> ```

javadoc
___
*[model/src/main/java/org/projectnessie/api/params/CommitLogParams.java, line 42 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMcbIq9RBd2PMU0XZw:-MbMcbIq9RBd2PMU0XZx:b-xtcjzz) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java#L42)):*
> ```Java
>   @Parameter(description = "ref to show log from", examples = {@ExampleObject(ref = "ref")})
>   @PathParam("ref")
>   private String ref;
> ```

Couldn't we use immutables here like we do for all the other objects?
___
*[servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java, line 28 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMczBI-BBizCWV7oEA:-MbMczBI-BBizCWV7oEB:b-eryfat) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java#L28)):*
> ```Java
> import org.junit.jupiter.api.Test;
> 
> public class InstantParamConverterProviderTest {
> ```

javadoc (I thought there was some checkstyle rule for this tbh)

nastra (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1325)* status: 0 of 16 files reviewed, 16 unresolved discussions (waiting on @laurentgo, @nastra, @rymurr, and @snazy)

___
*[model/src/main/java/org/projectnessie/api/params/CommitLogParams.java, line 76 at r5](https://reviewable.io/reviews/projectnessie/nessie/1325#-ManDoU_L_TzBkRhfR8x-r5-76:-Mb_JkY82HnDJASokbnL:bwn7pfj) ([raw file](https://github.com/projectnessie/nessie/blob/ec555efabd93ea8cd443746417ae7f9b457b595d/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java#L76)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

What I mean is that `CommitLogParams` just wraps all the arguments to fetch the commit log and that includes mandatory and optional parameters. Does that make sense?

Just assume we have some endpoint `getXyz(...)` where we already have 2-3 mandatory params. And now we add a few more fields (mandatory & optional). We would have to adjust the respective method parameters everywhere where this method is being called/used. Using a wrapper class for **all** of those parameters seems like the better choice imo, as otherwise you would have a mix of mandatory parameters on the respective `getXyz(...)` method that would need adjusting, but also the optional params in the `XyzParams` class.

I don't feel strongly about having `ref` in `CommitLogParams`, so let's see what other people think.
</blockquote></details>

I moved `ref` out of `CommitLogParams` and made it part of the original method signature of `getCommitLog(...)`.
___
*[model/src/main/java/org/projectnessie/api/params/CommitLogParams.java, line 36 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMcXsz0XG3Jspj5qQ6:-Mb_Jf2_0g2uDCgqdhOQ:b-7ql1td) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java#L36)):*

javadoc
</blockquote></details>

fixed
___
*[model/src/main/java/org/projectnessie/api/params/CommitLogParams.java, line 42 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMcbIq9RBd2PMU0XZw:-Mb_JO1k3McyafbJHCvB:b-1pd5md) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java#L42)):*

Couldn't we use immutables here like we do for all the other objects?
</blockquote></details>

I wish it would work. Unfortunately it doesn't play nice with the way the object is being instantiated (requiring a default constructor) during a request

nastra (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1325)* status: 0 of 16 files reviewed, 16 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java, line 28 at r7](https://reviewable.io/reviews/projectnessie/nessie/1325#-MbMczBI-BBizCWV7oEA:-Mb_KKbq--iXG2b8R7Ym:bxx3inx) ([raw file](https://github.com/projectnessie/nessie/blob/52fad2e6423ad4ee1829c2c2fcac5bb76c5c68a9/servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java#L28)):*

javadoc (I thought there was some checkstyle rule for this tbh)
</blockquote></details>

This is a test class. Do we require Javadoc on those?

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
nastra: I moved `ref` out of `CommitLogParams` back to the original `getCommitLog(...)` method. This way, when we add/remove mandatory params, clients will know immediately by getting a compile error (rather than a runtime error when such a mandatory param is in a wrapper class)

servers/quarkus-server/src/test/java/org/projectnessie/server/RestGitTest.java
```diff
@@ -65,9 +67,9 @@ public void testBasic() {
     Assertions.assertEquals(preSize + 1, references.length);
 
     Reference reference = rest().get("trees/tree/mainx").then()
-                 .statusCode(200)
-                 .extract()
-                 .as(Reference.class);
+        .statusCode(200)
```
rymurr, line 70: nit: lots of whitespace changes in this test.

servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Instant;
+
+import javax.ws.rs.ext.ParamConverter;
+
+import org.assertj.core.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class InstantParamConverterProviderTest {
```
rymurr, line 28: nit: why isn't the javadoc checkstyle arg picking these up? Most of the nessie codebase uses package private tests and test methods or adds javadocs for the tests

rymurr (commented):
the new tests are public which:
a) without javadoc im surprised its getting past checkstyle
b) is different from most unit tests in the codebase which are package-private

servers/services/src/test/java/org/projectnessie/services/rest/InstantParamConverterProviderTest.java
nastra: I haven't checked, but my assumption was that Checkstyle isn't complaining on test classes. I was comparing with a few other test classes, such as `public class RestGitTest`

servers/quarkus-server/src/test/java/org/projectnessie/server/RestGitTest.java
nastra: this is coming from the formatter unfortunately :(

rymurr (approved):

PR #1332, opened 2021-05-28
loosen pins for client install on python cli

### LINKED ISSUE #1137: investigate relaxing the pinned deps on python libs
It is fairly hard to install pynessie in some envs because the version pins are so strong. End up hacking to get it installed.

rymurr:
`setup.py` has changed to list all dependencies w/o explicit pins to enable flexibility in client installs.

we continue to use `requirements.txt` in the `python` directory when testing to ensure a stable test in CI.

fixes #1137

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1332)

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -11,8 +11,21 @@ with open("README.rst") as readme_file:
 with open("HISTORY.rst") as history_file:
     history = history_file.read()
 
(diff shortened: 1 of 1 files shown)


nastra (approved):
LGTM

PR #1245, opened 2021-05-14
Enable Java11 for Quarkus modules

laurentgo:
As Quarkus is designed for Java11 and Quarkus 2.0 will remove Java 8
support, allow quarkus server and lambda modules to be build with the
Java 11 release.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1245)

PR #1344, opened 2021-05-28
Update Maven+Gradle plugins to launch the Nessie-Server via `java -jar`

### LINKED ISSUE #1333: Let apprunner-gradle-plugin fork Nessie process
Currently the apprunner-gradle-plugin uses Quarkus mechanisms to resolve artifacts, setup the configuration, etc. - this reuses functionality from the apprunner-maven-plugin.

Since #1245 (new in Nessie 0.6.1), the Nessie-Server is built with Java 11.

The Iceberg-Java-8 build (naturally) fails with Nessie 0.6.1 ([comment](https://github.com/apache/iceberg/pull/2588#issuecomment-849912309)). It's not a solution to revert #1245, because Quarkus 2.0 will drop Java 8 support.

The idea is to launch the Nessie server as a separate process using the uber-jar, which is published to Maven Central since Nessie 0.6.1.

Removing all the Quarkus-bootstrap code from the plugin should also simplify the plugin code a lot and reduce the amount of dependencies.

It might be good to use the same approach in the Maven plugin as well, and continue to share the code base for both the Maven and Gradle plugins.

The new code that launches the Nessie server as a separate process just needs a bit more logic (rather configuration) to distinguish whether it shall use an "external dependency to the uber-jar" or whether it shall (or can) look for the `servers/quarkus-server/target/quarkus-app/quarkus-run.jar` (the non-uber variant).

snazy:
Before this PR, the apprunner-gradle-plugin uses Quarkus mechanisms to resolve artifacts, setup the configuration, etc. - this
reuses functionality from the apprunner-maven-plugin. Since #1245 (new in Nessie 0.6.1), the Nessie-Server is built with Java 11.
The Iceberg-Java-8 build (naturally) fails with Nessie 0.6.1 (comment). It's not a solution to revert #1245, because Quarkus 2.0
will drop Java 8 support.

The idea is to launch the Nessie server as a separate process using the uber-jar, which is published to Maven Central since Nessie 0.6.1.

Removing all the Quarkus-bootstrap code from the plugin simplifies the plugin code a lot and removes the amount of required dependencies.

There are 2 "official" and 1 "private" way to specify the Nessie-Quarkus jar to run:
1. In Maven by specifying the coordinates via the `appArtifactId` config options, in Gradle by declaring the
  dependency via the `nessieQuarkusServer` configuration).
2. In Maven or Gradle by providing the path to the jar via the `executableJar` config option.
3. Only in Maven and only for Nessie-source-tree builds: neither 1 nor 2, the Maven plugin figures out the path on its own.

A new module `nessie-apprunner-common` contains the underlying code to find the JVM via environment variables or system properties.
The Maven tries to find a matching JDK via the Maven toolkits API and falls back to the `JavaVM` implementation. The Gradle plugin 
always uses the `JavaVM` implementation (Gradle 5.4 doesn't have that functionality, but newer Gradle version do IIRC).
That module also contains code to start the Java process running the Nessie-Server-Quarkus-jar, capture the http listen URL and provide
both the full URL and port via project properties to associated test tasks.

The overall implementation makes it much easier to reason about system properties and environment variables passed to the executable
Nessie-Quarkus-jar. With the Quarkus-bootstrap code, it was hard to reason about and debug issues with properties passed to Quarkus and/or
Nessie.

The Maven plugin also got some integration tests.

Fixes #1333

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1344)


tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/JavaVM.java
```diff
@@ -0,0 +1,203 @@
...
+public final class JavaVM {
+
+  public static final int MAX_JAVA_VERSION_TO_CHECK = 19;
+
+  private static final Pattern MAJOR_VERSION_PATTERN = Pattern.compile("^(1[.])?([0-9]+)([.].+)?$");
+
+  private static final AtomicReference<JavaVM> CURRENT_JVM = new AtomicReference<>();
+  private final Path javaHome;
+
+  //@VisibleForTesting
```
rymurr, line 44: nit: commented annotation

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/JavaVM.java
```diff
@@ -0,0 +1,203 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.IntFunction;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Helper class to locate a JDK by Java major version and return the path to the {@code java} executable.
+ */
+public final class JavaVM {
```
rymurr, line 35: sad face that you had to write this :-( no one has packaged this as a library before?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
```diff
@@ -0,0 +1,113 @@
...
+  private static final Pattern HTTP_PORT_LOG_PATTERN = Pattern.compile("^.*Listening on: (http[s]?://[^ ]*)$");
+  static final String TIMEOUT_MESSAGE = "Did not get the http(s) listen URL from the console output.";
+
+  private final LongSupplier clock;
+  private final Consumer<String> stdoutTarget;
+  private final long deadlineListenUrl;
+
+  private final CompletableFuture<String> listenUrl = new CompletableFuture<>();
+
+  ListenUrlWaiter(LongSupplier clock, long timeToListenUrlMillis, Consumer<String> stdoutTarget) {
```
rymurr, line 43: Why not use `Clock` interface?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
```diff
@@ -0,0 +1,113 @@
...
+    if (!listenUrl.isDone()) {
+      Matcher m = HTTP_PORT_LOG_PATTERN.matcher(line);
+      if (m.matches()) {
+        listenUrl.complete(m.group(1));
+      }
+    }
+    stdoutTarget.accept(line);
+  }
+
+  //@VisibleForTesting
```
rymurr, line 59: nit

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.function.Consumer;
+import java.util.function.LongSupplier;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Accepts {@link String}s via it's {@link #accept(String)} method and checks for the
+ * {@code Listening on: http...} pattern.
+ */
+final class ListenUrlWaiter implements Consumer<String> {
```
rymurr, line 32: Presumably this means if the Quarkus log line changes or the user swiches logging levels we would never pick up the started prcoess? Will the started process still be killed or oprhaned?

rymurr (commented):
A handful of nits but mostly looks good. I am a bit worried about all the potentail edge cases: are we able to test taht the external app gets stopped if mvn dies or there are other crashes. It looks like if maven fails the app gets cleaned up, is there any way for that to fail?

I just want to make sure we fail smoothly in as many edge cases as possible and don't leave a bunch of nessie servers running after a build

pom.xml
```diff
@@ -161,6 +161,7 @@
     <quarkus.version>1.13.6.Final</quarkus.version>
     <openapi.version>1.1.2</openapi.version>
     <jsr305.version>3.0.2</jsr305.version>
+    <soebes.itf.version>0.10.0</soebes.itf.version>
```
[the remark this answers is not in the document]
snazy: Let's fix it in another PR?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
snazy: Yea - `Clock` is an interface for wall-clock-time, i.e. affected by NTP and such.

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
snazy: If it doesn't find the pattern, it will kill the process and error out.

snazy:
> I am a bit worried about all the potentail edge cases: are we able to test taht the external app gets stopped if mvn dies or there are other crashes. It looks like if maven fails the app gets cleaned up, is there any way for that to fail?
> 
> I just want to make sure we fail smoothly in as many edge cases as possible and don't leave a bunch of nessie servers running after a build

If Maven/Gradle run properly, the external process will get killed. If the build tools get killed (SIGTERM), a shutdown-hook will take care of killing the process. If the build tool gets killed with SIGKILL, well, there's nothing we can do.

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
```diff
@@ -0,0 +1,273 @@
...
+  }
+
+  private Object watchdog() throws IOException {
+    try (InputStream out = process.getInputStream(); InputStream err = process.getErrorStream()) {
+      InputBuffer stdout = new InputBuffer(out, listenUrlWaiter);
+      InputBuffer stderr = new InputBuffer(err, stderrTarget);
+      try {
+
+        while (true) {
+          boolean anyIo = stdout.io();
```
rymurr, line 233: I am not sure I get the point of checking for text logged to the screen, especially below where we enter a warm loop if no io and a hot loop if io. This while loop is unclear to me in general. can you say a few words?

tools/apprunner-common/src/test/java/org/projectnessie/quarkus/runner/TestProcessHandler.java
```diff
@@ -0,0 +1,315 @@
...
+    ProcessHandlerMock phMock = new ProcessHandlerMock();
+
+    phMock.ph.started(phMock.proc);
+
+    assertThatThrownBy(() -> phMock.ph.started(phMock.proc))
+        .isInstanceOf(IllegalStateException.class)
+        .hasMessage("Process already started");
+  }
+
+  @RepeatedTest(20) // repeat, risk of flakiness
```
rymurr, line 77: why are they flaky?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
```diff
@@ -0,0 +1,273 @@
...
+   * @return {@code this}
+   */
+  ProcessHandler started(Process process) {
+    if (this.process != null) {
+      throw new IllegalStateException("Process already started");
+    }
+
+    listenUrlWaiter = new ListenUrlWaiter(clock, timeToListenUrlMillis, stdoutTarget);
+
+    this.process = process;
```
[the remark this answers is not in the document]
snazy: It's not a thread-safe thing at all.

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
```diff
@@ -0,0 +1,273 @@
...
+          try {
+            int ec = process.exitValue();
+            exitCode.set(ec);
+            if (!anyIo) {
+              if (!stopped.get()) {
+                stderrTarget.accept(String.format("Watched process exited with exit-code %d", ec));
+              }
+              break;
+            }
+          } catch (IllegalThreadStateException e) {
```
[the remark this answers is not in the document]
snazy: It's a copy of the `Thread.getExitCode()` API.

tools/apprunner-common/src/test/java/org/projectnessie/quarkus/runner/TestProcessHandler.java
snazy: They're not. They could be though, because concurrency.

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
snazy: added some comment

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
```diff
@@ -0,0 +1,273 @@
...
+  private Consumer<String> stdoutTarget = System.out::println;
+  private Consumer<String> stderrTarget = System.err::println;
+  private ListenUrlWaiter listenUrlWaiter;
+
+  private volatile ExecutorService watchdogExecutor;
+  private volatile Future<?> watchdogFuture;
+  private volatile Thread shutdownHook;
+
+  /**
+   * The name for this process handler.
```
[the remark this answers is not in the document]
snazy: argh - left-over. removed.

rymurr:
As per our discussion yesterday I think we should discuss this w/ the Iceberg community. @snazy could you ask on the mailing list? I think our options are: 

1. have both java8 and java11 available to gradle and go with this patch
2. use testcontainers
3. skip nessie completely for java8
4. build a non-quarkus test server

I think I prefer in order option 4, 1, 2, 3. Option 4 could be a thin, in memory impl only jersey server that we could easily start and stop from java code. That would remove the need for the gradle and maven plugins completely.

pom.xml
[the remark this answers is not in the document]
rymurr: can you fix the one you added

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/InputBuffer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.Charset;
+import java.util.function.Consumer;
+
+/** Captures input from an {@link InputStream} and emits full lines terminated with a {@code LF}. */
+final class InputBuffer {
```
rymurr, line 27: Unit tests for this? I can't read this to know its right. Is there really nothing in the standard library to read inputs and split by line breaks?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/JavaVM.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.function.Function;
+import java.util.function.IntFunction;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Helper class to locate a JDK by Java major version and return the path to the {@code java}
+ * executable.
+ */
+public final class JavaVM {
```
rymurr, line 36: Given the way we are changing the plugin can we remove this code and just fail if we are <JDK11?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ListenUrlWaiter.java
```diff
@@ -0,0 +1,126 @@
...
+
+  String peekListenUrl() {
+    try {
+      return listenUrl.isDone() ? listenUrl.get() : null;
+    } catch (Exception e) {
+      throw new RuntimeException();
+    }
+  }
+
+  private static final long MAX_ITER_WAIT_NANOS = TimeUnit.MILLISECONDS.toNanos(50);
```
rymurr, line 78: this should be at the top of the class

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
```diff
@@ -0,0 +1,297 @@
...
+ *
+ * <p>This implementation is not thread-safe.
+ */
+public class ProcessHandler {
+
+  // intentionally long timeouts - think: slow CI systems
+  public static final long MILLIS_TO_HTTP_PORT = 30_000L;
+  public static final long MILLIS_TO_STOP = 15_000L;
+
+  private LongSupplier clock = System::nanoTime;
```
rymurr, line 49: I still don't like calling this a clock. Its a ticker or counter

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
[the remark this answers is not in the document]
rymurr: Why do we need it if its a copy? I would rather not have this hard to read code duplicated

tools/apprunner-common/src/test/java/org/projectnessie/quarkus/runner/TestInputBuffer.java
```diff
@@ -0,0 +1,102 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.Reader;
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import org.junit.jupiter.api.Test;
+
+class TestInputBuffer {
```
rymurr, line 27: Found the test, ignore my comment above.

rymurr (changes requested):
Im really torn about this change tbh. It gets rid of a lot of the complicated quarkus classpath code but adds in a bunch of complicated process handling and string stream handling.

Complexity wise it seems like a wash. What are the advantages otherwise?

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/ProcessHandler.java
snazy: renamed to `ticker`

snazy:
The approach to run the Nessie-Server as a separate process removes a lot of issues from the project (e.g. Iceberg) that runs Nessie integration tests. For example:
* Quarkus dependency added as a bom to the whole test-project - injecting dependencies of the nessie-server to the tested project
* Quarkus / Nessie-Server configuration (think: reducing spam from the http-access-log) is overly complex - rather impossible
* Logger infra between Nessie-Server and the tested project conflicts due to classloader / static-initialization / service-stuff

tools/apprunner-common/src/main/java/org/projectnessie/quarkus/runner/JavaVM.java
snazy: The integration tests do fail when run with Java < 11.
I'd rather keep this one so we can test against newer Java versions, when we want to.
The code that actually checks for a different JDK/JRE is only run, if you're not running a Java 11 compatible JVM.

rymurr: That sounds to me like its dead code then? Or am i missing something? Realistically we will support only jdk11 until quarkus stops support. So the build will fail before we reach this point on jdk8 and on jdk11 we will not use this code.

please correct me if I don't get it.

snazy: Not really dead. In case of Iceberg yes, but other projects that are strictly tied to a Java 8 build have to use this functionality.

nastra (approved):

PR #1357, opened 2021-05-31
fix naming in nessie client merge operation

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1357)

### DIFF
--- a/python/pynessie/cli.py
+++ b/python/pynessie/cli.py
@@ -316,8 +316,8 @@ def tag(ctx: ContextObject, list: bool, force: bool, delete: bool, tag_name: str
 
 
 @cli.command("merge")
-@click.option("-b", "--branch", help="branch to merge onto. If not supplied the default branch from config is used")
-@click.argument("merge_branch", nargs=1, required=False)
+@click.option("-b", "--branch", "from_branch", help="branch to merge onto. If not supplied the default branch from config is used")
+@click.argument("onto_branch", nargs=1, required=False)
 @click.option(
     "-f",
     "--force",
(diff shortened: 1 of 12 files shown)


python/pynessie/nessie_client.py
```diff
@@ -164,14 +164,14 @@ def assign_tag(self: "NessieClient", tag: str, to_ref: str, old_hash: Optional[s
         tag_json = ReferenceSchema().dumps(Tag(tag, self.get_reference(to_ref).hash_))
         assign_tag(self._base_url, tag, tag_json, old_hash, self._ssl_verify)
 
-    def merge(self: "NessieClient", branch: str, to_branch: str, old_hash: Optional[str] = None) -> None:
+    def merge(self: "NessieClient", branch: str, from_branch: str, old_hash: Optional[str] = None) -> None:
```
nastra, line 167: should this be named `onto_branch` maybe? It seems we're using `onto` at a bunch of other code places as well

rymurr: agreed, fixed

python/pynessie/nessie_client.py
```diff
@@ -164,15 +164,15 @@ def assign_tag(self: "NessieClient", tag: str, to_ref: str, old_hash: Optional[s
         tag_json = ReferenceSchema().dumps(Tag(tag, self.get_reference(to_ref).hash_))
         assign_tag(self._base_url, tag, tag_json, old_hash, self._ssl_verify)
 
-    def merge(self: "NessieClient", branch: str, to_branch: str, old_hash: Optional[str] = None) -> None:
+    def merge(self: "NessieClient", onto_branch: str, from_branch: str, old_hash: Optional[str] = None) -> None:
```
snazy, line 167: Can you adopt the order of "from" and "onto" to the same as in cli.py (or vice versa)?

rymurr: done

python/pynessie/cli.py
```diff
@@ -316,8 +316,8 @@ def tag(ctx: ContextObject, list: bool, force: bool, delete: bool, tag_name: str
 
 
 @cli.command("merge")
-@click.option("-b", "--branch", help="branch to merge onto. If not supplied the default branch from config is used")
-@click.argument("merge_branch", nargs=1, required=False)
+@click.option("-b", "--branch", "from_branch", help="branch to merge onto. If not supplied the default branch from config is used")
```
nastra, line 319: isn't this rather the `onto_branch` here? At least that's what the `help` description suggests.

It seems `from_branch` always defaults to `ctx.nessie.get_default_branch()` as there is no `--from-branch` flag (unless I'm missing something?

rymurr:
Yeah, hence why the tests are failing still. Will update it soon....one of
those mornings :-P

rymurr:
this should be ok now.

nastra (approved):

PR #1384, opened 2021-06-07
Allow setting author when committing via Python CLI

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1384)

### DIFF
--- a/python/pynessie/cli.py
+++ b/python/pynessie/cli.py
@@ -181,12 +181,12 @@ def set_head(ctx: ContextObject, head: str, delete: bool) -> None:
 @click.option("-n", "--number", help="number of log entries to return", type=int)
 @click.option("--since", "--after", help="Commits more recent than specific date")
 @click.option("--until", "--before", help="Commits older than specific date")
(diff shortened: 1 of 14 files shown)


python/pynessie/cli.py
```diff
@@ -181,12 +181,12 @@ def set_head(ctx: ContextObject, head: str, delete: bool) -> None:
 @click.option("-n", "--number", help="number of log entries to return", type=int)
 @click.option("--since", "--after", help="Commits more recent than specific date")
 @click.option("--until", "--before", help="Commits older than specific date")
-@click.option("--author", "--committer", is_flag=True, help="limit commits to specific committer")
```
rymurr, line 184: this makes the author and commiter the same flag. As in we cant filter on author and commiter

nastra: this will be fixed in a follow up PR that I already have in the pipeline

rymurr (approved):

rymurr (approved):

PR #1401, opened 2021-06-09
DRAFT: VersionStore global tables

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1401)

### DIFF
--- a/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java
+++ b/versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java
@@ -37,11 +37,13 @@ import java.util.stream.Stream;
 import org.projectnessie.versioned.BranchName;
 import org.projectnessie.versioned.Delete;
 import org.projectnessie.versioned.Diff;
+import org.projectnessie.versioned.GlobalOperation;
 import org.projectnessie.versioned.Hash;
 import org.projectnessie.versioned.Key;
 import org.projectnessie.versioned.NamedRef;
 import org.projectnessie.versioned.Operation;
 import org.projectnessie.versioned.Put;
+import org.projectnessie.versioned.PutGlobal;
 import org.projectnessie.versioned.Ref;
 import org.projectnessie.versioned.ReferenceAlreadyExistsException;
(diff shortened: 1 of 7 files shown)


versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java
```diff
@@ -515,13 +560,40 @@ public Collector collectGarbage() {
     return InactiveCollector.of();
   }
 
+  @Override
+  public Stream<WithTypeAndId<Key, EnumT>> getGlobalKeys() {
+    synchronized (lock) {
```
nastra, line 565: I think it might be good to replace `ConcurrentHashMap` with a `NonBlockingHashMap` (https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/maps/NonBlockingHashMap.java#L83) and remove the `lock`

nastra (commented):
given my limited knowledge, the first approach seems to be fine, although I would probably get rid of using `locks`

versioned/memory/src/main/java/org/projectnessie/versioned/memory/InMemoryVersionStore.java
rymurr: right now the lock is there to ensure that this hashmap isn't used to read during a commit. rather than using a nice collection I chose the lock because it has to synchronise across 3 different hashmaps during the commit. I haven't decided if we really need to lock this method at all (it might be that concurrently reading form this map while a commit is happenig is ok). But the lock is definitely needed (imho) for the commit method. Do you disagree?

snazy: Better not mix CHM + synchronized/locks. It may give people a false sense of safety to access the CHM outside the lock. But that's rather irrelevant as long it's a prototype ;)

rymurr: agreed, the `ConcurrentHashMap` is only there because i am lazy. These shoudl all be a `HashMap`s

PR #1415, opened 2021-06-11
Python Client: Allow creating a branch/tag with ref pointing to a branch name

nastra:


rymurr (approved):

python/pynessie/nessie_client.py
```diff
@@ -85,10 +85,20 @@ def create_branch(self: "NessieClient", branch: str, ref: str = None) -> Branch:
         :param ref: ref to fork from
         :return: Nessie branch object
         """
-        ref_json = ReferenceSchema().dump(Branch(branch, ref))
+        ref_json = ReferenceSchema().dump(Branch(branch, self.get_hash_from_ref(ref)))
         ref_obj = create_reference(self._base_url, ref_json, self._ssl_verify)
         return cast(Branch, ReferenceSchema().load(ref_obj))
 
+    def get_hash_from_ref(self: "NessieClient", ref: Optional[str]) -> Optional[str]:
```
rymurr, line 92: Why does this return optional? It should always return a string right

nastra: if the original `ref` was `None`, then we just return `None`

rymurr (dismissed):

PR #1416, opened 2021-06-11
Throw exception when conflicting keys are modified during merge without a common ancestor

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1416)


rymurr (approved):
lgtm, @laurentgo do you want to double check the in memory changes?

PR #1432, opened 2021-06-15
DRAFT: Validate client/server version on the REST API level

snazy:
Nessie client+server HTTP requests+responses include a new `Nessie-Version` HTTP header indicating the release version.

Clients validate the Nessie server version and the server validates the Nessie client version.

This is still a draft but illustrates the technical possibility.

The Nessie-Server part and the Java+Python+TS clients are implemented.

Open questions:
* I'm not yet sure how to exactly check whether "the other side" is compatible. Either use a "minimum version" that "the other side" must match or consider a version-range as compatible or assume that major+minor is compatible. I tend to the version-range with major+minor as the default.

Not nice:
* The implementation uses a `javax.ws.rs.container.ContainerResponseFilter` to add the `Nessie-Version` header on the server side. However, that filter is *not* called when the response has no content. So for example a HTTP `DELETE` w/o a response doesn't return that header. This behavior is the same when implementing `javax.ws.rs.ext.WriterInterceptor`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1432)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ venv/
 
 # Maven flatten plugin
 .flattened-pom.xml
+
+# UI / generated from gen-src/
+ui/src/utils/version-numbers.tsx
--- a/clients/client/src/main/java/org/projectnessie/client/NessieHttpClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieHttpClient.java
@@ -34,6 +34,7 @@ import java.lang.reflect.Proxy;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
+import javax.ws.rs.WebApplicationException;
 import org.projectnessie.api.ConfigApi;
 import org.projectnessie.api.ContentsApi;
 import org.projectnessie.api.TreeApi;
@@ -41,8 +42,11 @@ import org.projectnessie.client.auth.AwsAuth;
 import org.projectnessie.client.auth.BasicAuthFilter;
 import org.projectnessie.client.http.HttpClient;
 import org.projectnessie.client.http.HttpClientException;
+import org.projectnessie.client.http.RequestContext;
 import org.projectnessie.client.http.RequestFilter;
+import org.projectnessie.client.http.ResponseContext;
 import org.projectnessie.client.rest.NessieHttpResponseFilter;
+import org.projectnessie.common.NessieVersion;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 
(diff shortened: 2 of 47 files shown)


snazy:
Summarizing the TL;DR from our stand-up here.

This approach, although it looks like a "quick and easy win", probably causes more issues than it actually solves.
* Every client (pynessie, type-script, nessie-client/java and even `curl` and generic openapi-frontends) would have to send that `Nessie-Version` header. If not, then the server would reject the request with a HTTP/400.
* Determining whether the Nessie-version on the "other side" is compatible is hard. All we can effectively do is to say on both sides that we expect a minimum version, so that both sides have to agree. But that doesn't protect us and is quite hard to maintain.
* This header-approach would also break the 1st build after a version-bump, at least the tests in pynessie, because the recorded cassettes contain the old version. So this approach complicates the release-process/workflow.

For now we could not come up with something that's particularly awesome. The current idea is to 
* use versioning in the URL (`https://foo.bar/api/v2` , `.../v3` etc) and to
* version the payload classes (e.g. `EntriesResponse` -> `EntriesResponseV1`) and interfaces (e.g. `TreeApi` -> `TreeApiV1`) and probably put them in a separate Java package.

Especially the 2nd point is a lot of work, but it can soft-protect us from accidentally changing the Open/REST-API definition for an older version.

We need to have something for Nessie 1.0 GA, but can probably live with breaking changes until then. To mitigate that, we use the approach in #1260 but revert it before 1.0 GA.

/cc @laurentgo @rymurr @nastra

rymurr:
I would prefer something less complicated. My research indicates that doing too much is going to get complicated quickly. I thought https://apisyouwonthate.com/blog/api-versioning-has-no-right-way was a nice summary of my research.

> version the payload classes (e.g. EntriesResponse -> EntriesResponseV1) and interfaces (e.g. TreeApi -> TreeApiV1) and probably put them in a separate Java package.

I think this isn't so simple as it seems as `TreeApi` may not do breaking changes everywhere. We could end up with TreeApiVn which all have 90% common code or weird chained delegations. In the same vein I think its time we create a Client version of the API in Java. This wouldn't expose eg `TreeApi` directly and definitely wouldn't allow new clients to call old methods. Eg V3 of `ClientNessieApi` would only be allowed to call the most recent versions of each API endpoing rather than `{v1,V2,V3}` APIs. This would also allow us to pull in the `StreamUtils` to the actual API.

This made me think its also important to record a metric for API usage. It would be good to add a counter of how many clients are a particular version so we can plan deprecations accordingly.

From the article above I think I prefer the Mime-based resource versioning. I think this potentially will lead to less code duplication over the long term and fewer package moves now. It will keep the URIs simple also. Given we expect the Nessie endpoints may be used in non REST contexts this may be simpler too, as opposed to duplicating versioned endpoints in other contexts (eg gRPC). What do others think?

My opinion in summary:
1) go with #1260 for now and revert prior to GA
In this PR (or follow ons):
2) prepare mime based versioning and add `Content-Type` and `Accept` handling. IMHO w/o `Accept` you get the most up to date endpoint. Also modifying `Consumes` and `Produces` accordingly so requests are routed correctly.
3) Add metrics to count client types (this may already be covered by standard metrics, we should double check)
4) create a client api that hides direct usage of underlying rest apis.

PR #940, opened 2021-03-11
Add Gatling simulations

snazy:
Make `perftest/` a Maven parent pom, remove the JMeter test, add Gatling-test

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/940)

PR #1441, opened 2021-06-16
Add GH workflow to work with PR#940

snazy:
The PR for #940 shall upload perf-test results as HTML+YAML to present to the developer as a simple HTML page exposed in a commit status.

The idea is to push the HTML+YAML to the `docs/` directory of a separate github repo for now (it's easy and doesn't require us to have a publicly accessible S3 or something like that).

The change adds a new workflow that only runs for #940 (guarded with an `if: ` ), so other builds won't be affected.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1441)

### DIFF
--- a/.github/workflows/pull-request-upload.yml
+++ b/.github/workflows/pull-request-upload.yml
@@ -1,0 +1,95 @@
+name: PR Report Upload
+
+# This workflow runs after a successful run of the "PR Build Check" workflow.
+# It generates HTML and YAML reports from the Gatling simulation-logs and publishes those via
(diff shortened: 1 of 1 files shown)


snazy:
As an example, such results would be uploaded like here https://github.com/snazy/some-test/tree/master/docs/pr/foo-bar-baz/gatling and be linked to a commit status as https://snazy.github.io/some-test/pr/foo-bar-baz/gatling/report.html

rymurr (approved):

PR #1451, opened 2021-06-17
Allow using CEL expression when filtering commit log / entries

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1451)


servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -186,6 +212,33 @@ public LogResponse getCommitLog(String ref, CommitLogParams params)
     }
   }
 
+  private Stream<ImmutableCommitMeta> filterWithCelExpression(
+      Stream<ImmutableCommitMeta> items, CommitLogParams params) {
+    return items.filter(
+        commit -> {
+          try {
+            Script script =
```
snazy, line 220: Move this out of the Stream.
No need to parse+check the same expression for every CommitMeta.
That also allows to properly propagate parse+check errors back to the caller (`ScriptCreateException`).

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -186,6 +212,33 @@ public LogResponse getCommitLog(String ref, CommitLogParams params)
...
+                        "author",
+                        commit.getAuthor(),
+                        "committer",
+                        commit.getCommitter(),
+                        "commitTime",
+                        commit.getCommitTime()));
+
+            return script.execute(Boolean.class, arguments);
+          } catch (Exception e) {
+            e.printStackTrace();
```
snazy, line 236: It's probably necessary to wrap it into a `RuntimeException`.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -430,4 +483,58 @@ private static Reference makeRef(WithHash<? extends Ref> refWithHash) {
       throw new IllegalStateException("Unknown operation " + o);
     }
   }
+
+  private static class CommitMetaT extends BaseVal {
```
snazy, line 487: Remove?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: yeah it was rather a quick hack to get something running. Will move it out in the next comit

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: yep

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class CELTranslator {
+
+  public static String from(CommitLogParams params) {
+    Preconditions.checkArgument(null != params, "Commit log filtering params must be non-null");
+
+    String authorExpr =
+        or(
+            params.getAuthors().stream()
+                .map(x -> String.format("commit.author=='%s'", x))
```
snazy, line 44: Better put the author into a separate variable, so you get around the "classic SQL injection" pattern.

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+                .collect(Collectors.toList()));
+    authorExpr = params.getAuthors().size() > 1 ? wrap(authorExpr) : authorExpr;
+
+    String committerExpr =
+        or(
+            params.getCommitters().stream()
+                .map(c -> String.format("commit.committer=='%s'", c))
```
snazy, line 51: Better put the author into a separate variable, so you get around the "classic SQL injection" pattern.

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+    StringJoiner and = new StringJoiner(" && ");
+    if (!Strings.isNullOrEmpty(authorExpr)) {
+      and.add(authorExpr);
+    }
+    if (!Strings.isNullOrEmpty(committerExpr)) {
+      and.add(committerExpr);
+    }
+
+    if (null != params.getAfter()) {
+      and.add(String.format("timestamp(commit.commitTime) > timestamp('%s')", params.getAfter()));
```
snazy, line 64: Better put the author into a separate variable, so you get around the "classic SQL injection" pattern.

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    if (null != params.getBefore()) {
+      and.add(String.format("timestamp(commit.commitTime) < timestamp('%s')", params.getBefore()));
```
snazy, line 67: Better put the author into a separate variable, so you get around the "classic SQL injection" pattern.

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
nastra: done

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
nastra: done

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
nastra: done

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
nastra: done

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
```diff
@@ -134,6 +134,7 @@ public LogResponse getCommitLog(String ref, CommitLogParams params)
         .queryParam("pageToken", params.getPageToken())
         .queryParam("before", null != params.getBefore() ? params.getBefore().toString() : null)
         .queryParam("after", null != params.getAfter() ? params.getAfter().toString() : null)
+        .queryParam("cel_expr", params.getCelExpr())
```
rymurr, line 137: Here and elsewhere I don't like the name `cel_expr`. I would prefer `query_expression` or `expression` or something. I don't think it is relevant that cel is being used. We want to advertise to clients that there is an expression parameter available only.

I think this applies to everythign below (eg cli, python etc)  I will just mention it here to keep the review clean

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -74,6 +74,10 @@
   @QueryParam("before")
   private Instant before;
 
+  @Parameter(description = "A Common Expression Language (CEL) expression")
```
rymurr, line 77: This should prob be more descriptive. Somewhere in the docs or OpenAPI spec there should be more detail and this should link to it.

python/pynessie/cli.py
```diff
@@ -236,6 +245,8 @@ def log(
         filtering_args["before"] = until
     if end:
         filtering_args["end"] = end
+    if cel_expr:
```
rymurr, line 248: I thought we were going to remove all query params but the expression?

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+ * The purpose of the {@link CELTranslator} is to translate given filtering parameters, such as
+ * {@link CommitLogParams} / {@link EntriesParams} into valid CEL expressions as defined in the <a
+ * href="https://github.com/google/cel-spec/blob/master/doc/langdef.md">CEL language definition</a>.
+ *
+ * @see <a href="https://github.com/google/cel-spec/blob/master/doc/langdef.md">CEL language
+ *     definition</a>
+ */
+public class CELTranslator {
+
+  public static String from(CommitLogParams params) {
```
rymurr, line 38: I wonder if this is cleaner as a `Visitor` pattern or something? This will get big and hairy w/ parameter additions.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -69,6 +78,18 @@
 public class TreeResource extends BaseResource implements TreeApi {
 
   private static final int MAX_COMMIT_LOG_ENTRIES = 250;
+  private static final ScriptHost SCRIPT_HOST = ScriptHost.newBuilder().build();
```
rymurr, line 81: It feels to me like the CEL specific stuff should be in a separate class. Keeps versioning of API simpler if CEL isn't inside imho.

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
nastra: updated to `query_expression`

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
nastra: totally agreeing with you. I just didn't have time yet to add docs

python/pynessie/cli.py
nastra: I think it's probably easier to still send the same query params to the server as before and let the server deal with the translation of those parameters to CEL rather than having to translate this on the client-side

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: +1 on that

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
nastra: so far I tried it to keep it as simple as possible for the prototype. I think once we support more and more params that we translate to CEL, we'll eventually rewrite this and make it more flexible

python/pynessie/cli.py
rymurr: I don't agree. I think there should be one way of doing things w/ the REST API. An implementer shouldn't guess what will happen to their query if they send query params and an expression

servers/services/src/main/java/org/projectnessie/services/cel/CELTranslator.java
rymurr: makes sense to me :+1:

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: moved stuff out

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
nastra: updated the docs with some examples on the CLI & on the REST API side

python/pynessie/cli.py
nastra: I believe we still have to support the same query params on the REST API (`authors` / `committers`, `before`, `after`, ...) as otherwise we're breaking the API for older clients. Even though we're not on 1.0 yet, it feels wrong to break this.

Also having the `CELTranslator` on the server makes it much easier imo than moving this to the client-side. 

If a user ends up using the `query_expression` param with other parameters, then we'll immediately let them know by throwing an error.

Overall I believe we don't want to make people have to use & learn CEL on the client-side if they don't want to. Users should still be able to say `nessie log --author X` rather than having to learn CEL and being forced to use `nessie log --expression commit.author = 'X'`. Using `--expression` / `query_expression` should be an additional option for users that want to use **advanced filtering** and are willing to look at CEL.

rymurr: >Overall I believe we don't want to make people have to use & learn CEL on the client-side if they don't want to. Users should still be able to say nessie log --author X rather than having to learn CEL and being forced to use nessie log --expression commit.author = 'X'. Using --expression / query_expression should be an additional option for users that want to use advanced filtering and are willing to look at CEL.

I agree. I am not suggesting we make users learn expressions just to do simple queries on the CLI. I am suggesting doing the translation to a CEL expression on the client side.

>If a user ends up using the query_expression param with other parameters, then we'll immediately let them know by throwing an error.

This is a runtime error rather than something they understand while reading docs. I think we should make the REST API as clear and well defined as possible. The ambiguity comes while reading the API docs rather than at implementation time. It is not clear to me why a design would support two ways of achieving the same thing via the REST API.

>  Even though we're not on 1.0 yet, it feels wrong to break this.

I disagree, the query params have been available for ~1 week in a pre-GA release. No one will notice/complain that they disappear in another week.

nastra: @laurentgo / @snazy pinging you here so that you guys have some context as I wanted to bring this topic up in our standup so that we can decide on the optimal way going forward

nastra: We had a good discussion in the standup and the summary of the discussion is:
* we don't care much about backwards compatility at this point. Also Iceberg and won't notice these breaking changes
* REST API will only support `query_expression` and get rid of parameters, such as `authors` / `committers` / `before` / `after`
* Python and Java client will send only the `query_expression` parameter
* Python CLI can still have `--author` / `--committer` / `--before`/ `--after` flags. Those will be **shortcuts** only and will be translated directly to a CEL expression in the CLI, thus only sending a `query_expression` to the server. Those **shortcus** should probably not be used in combination with the `--query-expression` flag.

I'll go ahead and implement those changes tomorrow.

nastra: fixed as described in the last comment

python/pynessie/expression_util.py
```diff
@@ -0,0 +1,95 @@
...
+        return query_expression
+
+    if types:
+        return _expression_for_entries_by_types(types)
+
+    return None
+
+
+def _expression_for_commit_log_by_author(authors: List[str]) -> str:
+    return __generate_expression(authors, "commit.author=='{}'")
```
snazy, line 57: This one and the other `'{}'` formatted strings need to add CEL-string escaping.
It's unlikely at the moment, but still possible to "ruin" an expression without proper escaping - and the error message would be rather confusing for users.
Maybe also add some lose pattern-checks for the different types (timestamp: `[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}([.][0-9]{1-9})?Z` , committer/author: `[a-zA-Z0-9_.+$@-]+` (or whatever makes sense here)).

Maybe some Python library (`dateutil`?) allows some more flexible way for users to specify timestamps and provide a string format that CEL supports.

[celpy](https://github.com/cloud-custodian/cel-python/blob/master/src/celpy/evaluation.py#L2345) has something to adopt here.

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -15,11 +15,6 @@
  */
 package org.projectnessie.api.params;
```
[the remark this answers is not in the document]
nastra: I would prefer to keep it as it is for now until we have more `Param` classes, where it gets clearer how the Base class could look. Even though the 3 fields are identical, only two have the same parameter description. Not sure it's worth adding inheritance for 2 fields atm.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -187,36 +195,50 @@ public LogResponse getCommitLog(String ref, CommitLogParams params)
...
+          SCRIPT_HOST.getOrCreateScript(
+              params.getQueryExpression(), COMMIT_LOG_DECLARATIONS, Collections.emptyList());
+    } catch (ScriptException e) {
+      throw new IllegalArgumentException(e);
     }
-    return commits;
+    return commits.filter(
+        commit -> {
+          // currently this is just a workaround where we put CommitMeta into a hash structure.
+          // Eventually we should have a CommitMetaT that we register to avoid unnecessary object
```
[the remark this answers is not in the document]
nastra: yes Jackson-ized objects would also be good. I think in the long run, users of Cel-Java still might want to provide their own types

python/pynessie/expression_util.py
rymurr: agreed. For parsing the `dateutil.parser.parse` method should be able to do all the heavy lifting for dates. If it can't parse the date we assume and error. If it can we write the `datetime` back out as a formatted str

pom.xml
```diff
@@ -166,6 +166,11 @@
 
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>org.projectnessie.cel</groupId>
+        <artifactId>cel-tools</artifactId>
+        <version>0.1.0</version>
```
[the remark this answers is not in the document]
nastra: sgtm

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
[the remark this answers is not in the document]
rymurr: can we not filter on the other parts of the commit meta object? Eg, what if I wanted all commits from a particular spark app (via the parameters map)?

python/pynessie/expression_util.py
nastra: yes we definitely need to better check for dates and the CEL strings in general. I was planning to do a bunch of those things in a follow-up PR

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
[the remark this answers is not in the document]
nastra: We definitely can. I just didn't add it in the first implementation. Is this something you'd like to be implemented as part of this PR or a follow-up?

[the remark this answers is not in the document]
nastra: Also once we switch to Jackson-ized objects the filtering on other parts of `CommitMeta` should work ouf of the box

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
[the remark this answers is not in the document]
rymurr: I agree w/ @nastra. I think if anythign we would get rid of these objects and return to 3 parameters in the actual jax-rs method. We only intruduced these objects to make it easy/clean to add multiple filters, which we have now folded into CEL. 

Just food for thought, not needed for this PR

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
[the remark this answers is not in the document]
rymurr: sounds good. Later is fine by me. Can we just file an issue so it doesn't slip?

python/pynessie/expression_util.py
rymurr: I think that string escaping is an important issue to fix asap. I would like to see clean date handling too. Neither HAVE to be done in this PR but a quick follow up would be good.

nastra: I was planning to work on the follow-up right after this PR, but didn't want to make stuff here even more difficult to review. I'll open an issue for that in a bit

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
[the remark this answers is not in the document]
nastra: indeed it would probably make more sense to move those 3 params to the actual method again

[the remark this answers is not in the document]
nastra: opened https://github.com/projectnessie/nessie/issues/1479 to address that outside of this PR as the required changes would otherwise just pollute everything

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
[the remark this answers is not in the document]
nastra: opened https://github.com/projectnessie/nessie/issues/1480

python/pynessie/expression_util.py
nastra: opened https://github.com/projectnessie/nessie/issues/1481

rymurr (approved):

snazy (approved):

PR #1468, opened 2021-06-21
Experiment w/ jacoco-reports to fix code-coverage issues

snazy:
We generate two aggregate reports: `jacoco-aggregate` and `jacoco-aggregate-it`.
Both contain information for the same classes, but with _different_ values for
hits/misses.

This experiment generates another report `jacoco-aggregate-all` and only pushes
that one to codecov.

Crossing fingers, that this helps...

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1468)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -127,7 +127,7 @@ jobs:
       with:
         flags: java
         files: |
-          **/target/site/jacoco-aggregate-all/jacoco.xml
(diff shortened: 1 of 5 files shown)


snazy:
Um, so adding another aggregated report didn't work.
But it looks like adding a project that aggregates all other projects and generates the code-coverage report there seems to work.
Let's see, what codecov says...

rymurr (approved):

PR #1477, opened 2021-06-23
Do not let CI fail on forks

snazy:
Github workflow runs currently fail on forks, because the configured relevant
workflows require secrets to deploy e.g. snapshot artifacts or push a docker image.

This change adds some `if: ${{ github.repository == 'projectnessie/nessie' }}`
conditions to the relevant workflows and steps so the workflows are happy on
Nessie-forks as well.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1477)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,10 +91,15 @@ jobs:
         restore-keys: ${{ runner.os }}-gradle
     - name: Build with Maven
       run: ./mvnw -B deploy --file pom.xml -Pcode-coverage,jdk8-tests,native,release -DdeployAtEnd=true -Dtest.log.level=WARN
+      if: ${{ github.repository == 'projectnessie/nessie' }}
       env:
         MAVEN_USERNAME: ${{ secrets.OSSRH_ACCESS_ID }}
         MAVEN_OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
         MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
+        SPARK_LOCAL_IP: localhost
+    - name: Build with Maven (Fork)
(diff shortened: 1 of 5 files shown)


snazy:
Note: this change's tested on my fork: https://github.com/snazy/nessie/runs/2892411232?check_suite_focus=true

rymurr:
Do you want it to run on your fork? I just disabled actions on my fork

snazy:
> Do you want it to run on your fork? I just disabled actions on my fork

It's polite to people that just fork the Nessie repo. Doesn't look good IMO if CI on the main branch fails and people have to reason about "why".

rymurr:
I would say we don't even want it to run at all on forks. I would prefer to turn it off rather than skip. I find it bizarre not that I got an error that it failed rather that it ran at all on my fork.

.github/workflows/main.yml
```diff
@@ -91,11 +91,16 @@ jobs:
         restore-keys: ${{ runner.os }}-gradle
     - name: Build with Maven
       run: ./mvnw -B deploy --file pom.xml -Pcode-coverage,jdk8-tests,native,release -DdeployAtEnd=true -Dtest.log.level=WARN
+      if: ${{ github.repository == 'projectnessie/nessie' }}
```
laurentgo, line 94: how about checking if MAVEN_USERNAME is defined?

.github/workflows/release-create.yml
```diff
@@ -46,6 +46,8 @@ jobs:
   create-release:
     name: Create release
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'projectnessie/nessie' }}
```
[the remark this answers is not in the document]
snazy: It's the only manually run workflow. Removed the condition here.

.github/workflows/main.yml
snazy: Yup, changed to that in the whole PR.

rymurr (approved):

PR #1513, opened 2021-07-01
Support filtering on all fields of CommitMeta

### LINKED ISSUE #1480: Filtering should be possible on all fields of CommitMeta
Once we can just pass in the `CommitMeta` object as in `return script.execute(Boolean.class, Collections.singletonMap("commit", commit));`, we should be able to filter on all fields (including Map properties). 
We should also add a bunch of tests and mention the usable variables in REST API & CLI docs

nastra:
fixes #1480


nastra:
Note that https://github.com/nastra/nessie/blob/59046a0075cb6ad4b93a0377bd185636113107da/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java#L313-L346 is currently untouched by this PR because there are a few quirks that I'd like to talk about in the standup.

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
```diff
@@ -17,18 +17,22 @@
...
 
-  public static final ScriptHost SCRIPT_HOST = ScriptHost.newBuilder().build();
+  public static final ScriptHost SCRIPT_HOST =
+      ScriptHost.newBuilder().registry(JacksonRegistry.newRegistry()).build();
 
   public static final List<Decl> COMMIT_LOG_DECLARATIONS =
       Arrays.asList(
-          Decls.newVar("commit", Decls.newMapType(Decls.String, Decls.String)),
+          Decls.newVar("commit", Decls.newObjectType(CommitMeta.class.getName())),
           Decls.newVar("author", Decls.String),
```
snazy, line 36: Do we still need to declare `author` + `committer` + ` commitTime` ?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -207,28 +208,18 @@ public LogResponse getCommitLog(
     final Script script;
     try {
       script =
-          SCRIPT_HOST.getOrCreateScript(
-              queryExpression, COMMIT_LOG_DECLARATIONS, Collections.emptyList());
+          SCRIPT_HOST
```
snazy, line 211: Can you use the `buildScript` method for entries filtering as well, so it doesn't raise a deprecation-warning.

snazy (commented):
I think, it makes sense to add `withContainer("org.projectnessie.model")` to the `Script`-creation.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: updated

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
nastra: probably not. Updated the code and pushed a new version

snazy (approved):

PR #1521, opened 2021-07-02
Prepare multi-tenancy in the API

snazy:
This change adds support for multi-tenancy to the (REST) API.

There is no change in functionality, except that the Java and Python clients allows configuration of these two parameters when building/configuring the clients, at the same time/place when all other basic parameters (base-uri, auth, etc) are configured/applied.

Multi-tenant URIs for `TreeApi` + `ContentsApi` have the prefix `{owner}/{repo}/` as path parameters. "Non multi-tenant requests" work like before (no change to the REST API URIs).

Multi-tenant URIs look like this:
```
   http://localhost:19120/api/v1/<owner>/<repo>/trees...
```
It would be nicer to have the owner+repo parts before `api/v1/`, but `ContainerRequestFilter` implementations don't get called in that case.

The `ContainerRequestFilter` implementation `MultiTenantFilter` matches "multi-tenant URIs" and rewrites those as non-MT URIs, so Quarkus (or Jersey) treat those requests like non-MT, which has the advantage that we do not need to have the owner+repo information throughout all `TreeApi`/`ContentsApi` methods. The MT-information is available via `MultiTenant`, which has either non-null owner+repo information or `null` for both values.

The words `config` + `settings` (maybe more?) are not allowed for `owner` + `repo`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1521)


snazy:
Note: this isn't ready for review yet. I'm going to split out a few commits into separate PRs and try to find a solution to get the GC tests against iceberg working again (those don't anymore work with the REST-API change).

servers/services/pom.xml
```diff
@@ -76,6 +76,10 @@
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
     </dependency>
+    <dependency>
```
snazy, line 79: This will go away in a later PR.

servers/services/src/main/java/org/projectnessie/services/rest/MultiTenantFilter.java
```diff
@@ -0,0 +1,92 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.inject.Inject;
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import javax.ws.rs.container.PreMatching;
+import javax.ws.rs.ext.Provider;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
```
snazy, line 28: Will go away / be changed in a later PR.

servers/services/src/main/java/org/projectnessie/services/rest/MultiTenantFilter.java
```diff
@@ -0,0 +1,92 @@
...
+  private final Pattern multiTenantPathPattern;
+  private final String restPath;
+
+  @Inject MultiTenant multiTenant;
+
+  // TODO @ConfigProperty is the way in Quarkus to get the REST-API base path, but it's probabyl
+  //  not the right thing in Jersey. It would also be much nicer to have the owner+repo part
+  //  before the /api/v1 URI part.
+  @Inject
+  public MultiTenantFilter(@ConfigProperty(name = "quarkus.resteasy.path") String restPath) {
```
rymurr, line 53: I don't think this is compatible w/ all impls so we should prob take care of it now given where jersey PR is at

rymurr (commented):
looks good. I guess this is only dealing w/ the existence of multi tenancy in the endpoints. We are not yet taking on creation and modification of repos? Are you thinking this is an immediate follow-up or dependent on version store refactors?

servers/services/pom.xml
snazy: Superflous w/ the next commit

servers/services/src/main/java/org/projectnessie/services/rest/MultiTenantFilter.java
snazy: Will update

python/pynessie/cli.py
```diff
@@ -101,15 +101,25 @@ def parse_args(self: "DefaultHelp", ctx: click.Context, args: List) -> List:
...
 @click.pass_context
-def cli(ctx: click.core.Context, json: bool, verbose: bool, endpoint: str) -> None:
+def cli(ctx: click.core.Context, json: bool, verbose: bool, endpoint: str, owner: str, repo: str) -> None:
     """Nessie cli tool.
 
     Interact with Nessie branches and tables via the command line
     """
     try:
-        config = build_config({"endpoint": endpoint} if endpoint else None)
+        cfg_map = {}
```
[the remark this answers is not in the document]
snazy: They are (or better: can be) in the config file.
I've added some more tests for that case. It turned out, that eagerly creating the `NessieClient` and therefore validating the owner+repo parameters then, doesn't work. For example, if you'd do a `nessie-cli config --add owner snazy`, it would error out saying that the "repo" option is missing. So I'd changed the context-object to build the `NessieClient` lazily.

python/pynessie/cli.py
```diff
@@ -46,10 +46,22 @@
 class ContextObject(object):
     """Click context object."""
 
-    nessie: NessieClient
+    config: confuse.Configuration
     verbose: bool
     json: bool
 
+    __nessie: Optional[NessieClient] = None
```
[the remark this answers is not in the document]
snazy: It's private ;)

python/pynessie/cli.py
```diff
@@ -46,10 +46,22 @@
...
+
+    def nessie(self: "ContextObject") -> NessieClient:
+        """Get the ``NessieClient`` instance, create it, if necessary."""
+        if not self.__nessie:
```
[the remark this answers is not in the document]
snazy: IIRC, double-underscore is "private" and single-underscore is "protected"

python/pynessie/cli.py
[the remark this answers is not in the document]
rymurr: its still static here. Not sure why

python/pynessie/cli.py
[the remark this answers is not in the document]
rymurr: it means that the name will be mangled by the interpreter, not that its private. `__` is used to avoid name clashes w/ subclasses. `_` is used for non-public members. We prefer `_` over `__` unless the variable really shouldn't be used from subclasses. https://www.python.org/dev/peps/pep-0008/#id47

python/pynessie/cli.py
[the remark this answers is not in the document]
snazy: Python's a mystery ...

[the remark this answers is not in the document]
rymurr: Its just not Java. I am at fault for calling an instance variable static, which it isn't in the Java sense.  This may clarify things: https://dev.to/ogwurujohnson/distinguishing-instance-variables-from-class-variables-in-python-81

python/pynessie/cli.py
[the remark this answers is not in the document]
snazy: Updated to `_`

python/pynessie/cli.py
```diff
@@ -46,10 +46,22 @@
...
     """Click context object."""
 
-    nessie: NessieClient
+    config: confuse.Configuration
     verbose: bool
     json: bool
 
+    _nessie: Optional[NessieClient] = None
+
+    def nessie(self: "ContextObject") -> NessieClient:
```
rymurr, line 55: Now I am not sure why we have to lazy initialize the nessie object. We can't change the config after the ctx is created.

This change also means we should no longer use `attr.s` on the `ContextObject`. `attrs` library is meant for describing simple data objects and not really meant for stateful objects. At least imho. 

So we either have to:
1) remove the mutable state `_nessie`
2) remove the `attr.s` decorator and make this a normal class (w/ `__init__` and properties)

snazy:
Updated the PR to unify owner+repo parameters for Python.

python/pynessie/cli.py
```diff
@@ -101,15 +101,23 @@ def parse_args(self: "DefaultHelp", ctx: click.Context, args: List) -> List:
 @click.option("--json", is_flag=True, help="write output in json format.")
 @click.option("-v", "--verbose", is_flag=True, help="Verbose output.")
 @click.option("--endpoint", help="Optional endpoint, if different from config file.")
+@click.option("--repo", help="Optional repository owner + name, if different from config file, separated by a slash..")
```
rymurr, line 104: nit: double '.' at end of sentence

python/pynessie/nessie_client.py
```diff
@@ -51,7 +51,12 @@ class NessieClient(object):
 
     def __init__(self: "NessieClient", config: confuse.Configuration) -> None:
         """Create a Nessie Client from known config."""
-        self._base_url = config["endpoint"].get()
+        url = config["endpoint"].get()
+        repo = config["repo"].get() if "repo" in config else None
+        if repo:
+            sep = "" if url[-1:] == "/" else "/"
```
rymurr, line 57: what does this mean?

snazy: It's different whether your `endpoint` config value ends with a `/` or not.

snazy: To produce a proper endpoint-URL and always separate the "raw base" URI and owner/repo part.

python/pynessie/cli.py
snazy: fixed

python/pynessie/nessie_client.py
rymurr: got it, misread the first time

python/docs/main.rst
```diff
@@ -10,6 +10,8 @@
 	  --json           write output in json format.
 	  -v, --verbose    Verbose output.
 	  --endpoint TEXT  Optional endpoint, if different from config file.
+	  --repo TEXT      Optional repository owner + name, if different from config
+	                   file, separated by a slash..
```
rymurr, line 14: The double `.` is still in the readme

python/pynessie/cli.py
```diff
@@ -101,15 +101,23 @@ def parse_args(self: "DefaultHelp", ctx: click.Context, args: List) -> List:
...
     """Nessie cli tool.
 
     Interact with Nessie branches and tables via the command line
     """
     try:
-        config = build_config({"endpoint": endpoint} if endpoint else None)
+        cfg_map = {}
+        if endpoint:
+            cfg_map["endpoint"] = endpoint
+        # Handle owner+repo parameters individually to tackle the case when for example the owner is specified in
```
rymurr, line 116: is the comment still relevant?

python/docs/main.rst
snazy: Damn code generation... fixed it

python/pynessie/cli.py
snazy: nope. removed.

rymurr (approved):

PR #1534, opened 2021-07-07
Fix Quarkus ConfigProperties deprecation

laurentgo:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1534)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/config/DynamoVersionStoreConfig.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/config/DynamoVersionStoreConfig.java
@@ -15,12 +15,13 @@
  */
 package org.projectnessie.server.config;
 
-import io.quarkus.arc.config.ConfigProperties;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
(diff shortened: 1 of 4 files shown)


nastra (approved):

servers/quarkus-server/src/main/java/org/projectnessie/server/config/DynamoVersionStoreConfig.java
```diff
@@ -15,12 +15,13 @@
  */
 package org.projectnessie.server.config;
 
-import io.quarkus.arc.config.ConfigProperties;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.projectnessie.versioned.dynamodb.DynamoStoreConfig;
```
snazy, line 20: Think that'll cause a spotless-check error

nastra: I thought this would have been picked up https://github.com/projectnessie/nessie/blob/b784d4d0290536644c4ed27163520a4dbe8d0f0d/pom.xml#L1325-L1330 but that doesn't seem to be the case

rymurr:
superceded by #1563

PR #1547, opened 2021-07-07
Do spotless:check before building project

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1547)


nastra:
I missed that this is already covered by https://github.com/projectnessie/nessie/blob/b784d4d0290536644c4ed27163520a4dbe8d0f0d/pom.xml#L1325-L1329

PR #1583, opened 2021-07-12
Only accept NamedRefs in REST API

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1583)


servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -180,7 +180,14 @@ public LogResponse getCommitLog(
...
     int max =
         Math.min(maxRecords != null ? maxRecords : MAX_COMMIT_LOG_ENTRIES, MAX_COMMIT_LOG_ENTRIES);
-    Hash startRef = getHashOrThrow(pageToken != null ? pageToken : ref);
+
+    Ref startRef;
+    if (null == pageToken) {
+      // we should only allow named references when no paging is defined
+      startRef = namedRefWithHashOrThrow(ref).getValue();
+    } else {
+      startRef = getHashOrThrow(pageToken);
```
snazy, line 189: I think, it's fine to accept the "insecure design" (aka: you can pass in any hash and it will give you all the commits) for now until we have both authz and tvs2 in place.
Mind adding a `TODO` here?

servers/jax-rs/src/test/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -920,6 +920,7 @@ void invalidBranchNames(String invalidBranchName, String validHash) {
     Contents cts = IcebergTable.of("moo");
     MultiGetContentsRequest mgReq = MultiGetContentsRequest.of(key);
     Tag tag = Tag.of("valid", validHash);
+    String ref = "1234567890123456";
```
snazy, line 923: Where's `ref` being used?

python/pynessie/cli.py
```diff
@@ -186,6 +186,7 @@ def set_head(ctx: ContextObject, head: str, delete: bool) -> None:
 
 
 @cli.command("log")
+@click.option("-r", "--ref", help="branch to list from. If not supplied the default branch from config is used")
```
snazy, line 189: I'd personally prefer to have it work like `git log branch-name HASH..HASH` (so two args instead of one arg plus an option) - not that new option.

python/pynessie/cli.py
```diff
@@ -232,32 +234,34 @@ def log(  # noqa: C901
...
         if ".." in revision_range:
             start, end = revision_range.split("..")
         else:
             start = revision_range
-            end = None
 
     filtering_args: Any = {}
     if number:
         filtering_args["max"] = str(number)
+    if start:
```
snazy, line 255: I think, we should have a follow-up later to push all these down to the server. Especially "start" + "end" + "max" are relatively easy to implement and "max" is relatively easy to push down via the paging functionality.

snazy (commented):
Not sure whether I'm too early with my review, but am I missing changes to `TreeApi` here?
If this is not yet ready, ignore me

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -58,6 +60,18 @@ protected BaseResource(
     }
   }
 
+  WithHash<Ref> namedRefWithHashOrThrow(String ref) throws NessieNotFoundException {
+    try {
+      if (null != ref) {
+        Preconditions.checkArgument(
```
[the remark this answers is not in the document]
nastra: good point, didn't know that existed

servers/jax-rs/src/test/java/org/projectnessie/jaxrs/AbstractTestRest.java
nastra: it's a leftover that needs to be removed

python/pynessie/cli.py
nastra: agreed

python/pynessie/cli.py
nastra: I was trying to make it consistent with other commands that use `--ref` but I don't have a personal preference. @rymurr thoughts on this?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
nastra: done

nastra:
> but am I missing changes to `TreeApi` here

you were indeed quite early to the review because I didn't have time to review my own code yet ;)

I don't think we need any changes to the `TreeApi`. Those passed `refs` will stay as `Strings` and we're just validating (via a regex) that a `ref` isn't a hash.

python/pynessie/cli.py
rymurr: I don't have a strong opinion either. In doubt keep consistent so prob marginally lean towards the new ref option

nastra: ok then let's move forward with the proposed approach with `--ref`

rymurr (commented):
How do we support time travel? eg how do I read from a past hash?

rymurr (approved):

nastra:
> How do we support time travel? eg how do I read from a past hash?

Just for completeness, we talked about this on Slack and decided that we probably need to have an additional `hashOnRef` query parameter where a user can specify a particular hash on the given Branch/Tag.

I'm working on a follow-up PR that addresses this.

PR #1589, opened 2021-07-13
Add hashOnRef query param to support time travel on a named ref

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1589)

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
@@ -117,7 +117,7 @@ class ClientTreeApi implements TreeApi {
 
   @Override
   public LogResponse getCommitLog(
-      String ref, Integer maxRecords, String pageToken, String queryExpression)
+      String ref, String hashOnRef, Integer maxRecords, String pageToken, String queryExpression)
       throws NessieNotFoundException {
     HttpRequest builder =
         client.newRequest().path("trees/tree/{ref}/log").resolveTemplate("ref", ref);
@@ -125,6 +125,7 @@ class ClientTreeApi implements TreeApi {
         .queryParam("max", maxRecords != null ? maxRecords.toString() : null)
         .queryParam("pageToken", pageToken)
         .queryParam("query_expression", queryExpression)
+        .queryParam("hashOnRef", hashOnRef)
         .get()
         .readEntity(LogResponse.class);
   }
@@ -159,7 +160,11 @@ class ClientTreeApi implements TreeApi {
 
   @Override
   public EntriesResponse getEntries(
-      @NotNull String refName, Integer maxRecords, String pageToken, String queryExpression)
+      @NotNull String refName,
+      String hashOnRef,
+      Integer maxRecords,
+      String pageToken,
+      String queryExpression)
       throws NessieNotFoundException {
     HttpRequest builder =
         client.newRequest().path("trees/tree/{ref}/entries").resolveTemplate("ref", refName);
@@ -167,6 +172,7 @@ class ClientTreeApi implements TreeApi {
         .queryParam("max", maxRecords != null ? maxRecords.toString() : null)
         .queryParam("pageToken", pageToken)
         .queryParam("query_expression", queryExpression)
+        .queryParam("hashOnRef", hashOnRef)
         .get()
         .readEntity(EntriesResponse.class);
   }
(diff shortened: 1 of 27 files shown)


clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
```diff
@@ -117,14 +117,15 @@ public Branch getDefaultBranch() {
 
   @Override
   public LogResponse getCommitLog(
-      String ref, Integer maxRecords, String pageToken, String queryExpression)
+      String ref, String hashOnRef, Integer maxRecords, String pageToken, String queryExpression)
```
snazy, line 120: Guess this should be `startHash` instead of `hashOnRef`. Since we're already changing this, can you add `endHash` as welll?:

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -34,51 +34,57 @@ private StreamingUtil() {
...
+   * @param ref a named reference (branch or tag name)
+   * @param hashOnRef the hash on the given ref
    * @param pageSizeHint page-size hint for the backend
    * @param queryExpression The query expression to filter by
    * @return stream of {@link CommitMeta} objects
    */
   public static Stream<CommitMeta> getCommitLogStream(
       @NotNull TreeApi treeApi,
       @NotNull String ref,
+      String hashOnRef,
```
snazy, line 80: (similar to above about `startHash` instead of `hashOnRef` + `endHash`)

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -290,14 +296,19 @@ public EntriesResponse getEntries(
...
-          @Pattern(
-              regexp = Validation.REF_NAME_OR_HASH_REGEX,
-              message = Validation.REF_NAME_OR_HASH_MESSAGE)
+          @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
           @Parameter(
               description = "ref to show log from",
               examples = {@ExampleObject(ref = "ref")})
           @PathParam("ref")
           String ref,
+      @Nullable
```
snazy, line 305: (similar to above about `startHash` instead of `hashOnRef` + `endHash`)

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -176,15 +176,19 @@ public void deleteBranch(String branchName, String hash)
 
   @Override
   public LogResponse getCommitLog(
-      String ref, Integer maxRecords, String pageToken, String queryExpression)
+      String namedRef,
+      String hashOnRef,
```
snazy, line 180: (similar to above about `startHash` instead of `hashOnRef` + `endHash`)

ui/src/utils/api/apis/DefaultApi.ts
```diff
@@ -90,24 +90,28 @@ export interface DeleteTagRequest {
 
 export interface GetCommitLogRequest {
     ref: string;
+    hashOnRef?: string;
```
snazy, line 93: (similar to above about `startHash` instead of `hashOnRef` + `endHash`)

snazy (commented):
I just have one comment (but I was so mean to annotate every occurrence ;) ) - it's about the commit-log: since we effectively add the "start hash", it should be named "start hash" and since we're already there, it wouldn't be much of an effort to add the "end hash" parameter as well (maybe need to implement a `Spliterator` for the "end hash" though).

clients/client/src/main/java/org/projectnessie/client/ClientTreeApi.java
nastra: I would rather prefer to have this in a separate commit/PR since it's a slightly different feature

snazy (approved):

rymurr (approved):

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -59,14 +62,44 @@ protected BaseResource(
     }
   }
 
-  WithHash<Ref> namedRefWithHashOrThrow(String ref) throws NessieNotFoundException {
+  WithHash<NamedRef> namedRefWithHashOrThrow(String namedRef, @Nullable String hashOnRef)
+      throws NessieNotFoundException {
+    List<WithHash<NamedRef>> collect =
+        store
```
rymurr, line 68: how often is this called? This could be expensive and I don't quite understand the purpose

nastra: it's called in 4 places (getCommitLog / getEntries / getContents / getMultipleContents) and we will eventually have a faster alternative in in the VersionStoreV2

PR #1595, opened 2021-07-14
Re-introduce wrapper classes for query params of CommitLog/Entries

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1595)

### DIFF
--- a/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
+++ b/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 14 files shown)


rymurr (approved):

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -0,0 +1,157 @@
...
+import org.projectnessie.model.Validation;
+
+/**
+ * The purpose of this class is to include optional parameters that can be passed to {@link
+ * org.projectnessie.api.TreeApi#getCommitLog(String, CommitLogParams)}.
+ *
+ * <p>For easier usage of this class, there is {@link CommitLogParams#builder()}, which allows
+ * configuring/setting the different parameters.
+ */
+public class CommitLogParams {
```
snazy, line 33: Why isn't this an "Immutable"?

model/src/main/java/org/projectnessie/api/params/EntriesParams.java
```diff
@@ -0,0 +1,157 @@
...
+import org.projectnessie.model.Validation;
+
+/**
+ * The purpose of this class is to include optional parameters that can be passed to {@link
+ * org.projectnessie.api.TreeApi#getEntries(String, EntriesParams)}.
+ *
+ * <p>For easier usage of this class, there is {@link EntriesParams#builder()}, which allows
+ * configuring/setting the different parameters.
+ */
+public class EntriesParams {
```
snazy, line 33: Why isn't this an "Immutable"?

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
nastra: bean instantiation and immutables don't play well together. See also my comment in https://github.com/projectnessie/nessie/pull/1325#discussion_r642307366

PR #1607, opened 2021-07-15
Fix for missing scala javadoc+source artifacts

snazy:
Also adds a `skipGpg` property, so you can run with `-Prelease` without having a gpg key.

The [issue](https://github.com/davidB/scala-maven-plugin/issues/46) that the `scala-maven-plugin` does not pick up _generated_ sources for docs requires the hack in `clients/spark-extensions/pom.xml`.

Since the nessie-deltalake-core module doesn't actually compile anything, the doc-part needs to be skipped there.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1607)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -116,7 +116,7 @@ jobs:
         MAVEN_USERNAME: ${{ secrets.OSSRH_ACCESS_ID }}
         SPARK_LOCAL_IP: localhost
       if: ${{ ! env.MAVEN_USERNAME }}
-      run: ./mvnw -B install --file pom.xml -Pcode-coverage,jdk8-tests,native -Dtest.log.level=WARN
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         MAVEN_USERNAME: ${{ secrets.OSSRH_ACCESS_ID }}
         SPARK_LOCAL_IP: localhost
       if: ${{ ! env.MAVEN_USERNAME }}
-      run: ./mvnw -B install --file pom.xml -Pcode-coverage,jdk8-tests,native -Dtest.log.level=WARN
+      run: ./mvnw -B install --file pom.xml -Pcode-coverage,jdk8-tests,native,release -DskipGpg -Dtest.log.level=WARN
```
rymurr, line 119: Why skipGpg here?

snazy: it's the fork-run

rymurr (approved):

PR #1608, opened 2021-07-15
Add authorization rules API / check for authorization Rules

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server.authz;
+
+import com.google.common.collect.ImmutableMap;
+import java.security.AccessControlException;
+import java.util.Map;
+import java.util.function.Supplier;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import org.projectnessie.cel.tools.ScriptException;
+import org.projectnessie.model.ContentsKey;
+import org.projectnessie.server.config.QuarkusNessieAuthorizationConfig;
+import org.projectnessie.services.authz.AccessChecker;
+import org.projectnessie.services.authz.AccessContext;
+import org.projectnessie.versioned.NamedRef;
+
+/**
+ * A reference implementation of the {@link AccessChecker} that performs access checks using CEL
+ * expressions.
+ */
+@ApplicationScoped
+public class CelAccessChecker implements AccessChecker {
+  private final QuarkusNessieAuthorizationConfig config;
+  private final CompiledAuthorizationRules compiledRules;
+
+  public enum AuthorizationRuleType {
+    VIEW_REFERENCE,
+    CREATE_REFERENCE,
+    DELETE_REFERENCE,
+    LIST_COMMIT_LOG,
+    READ_ENTRIES,
+    ASSIGN_REFERENCE_TO_HASH,
+    COMMIT_CHANGE_AGAINST_REFERENCE,
+    READ_ENTITY_VALUE,
+    UPDATE_ENTITY,
+    DELETE_ENTITY;
+  }
+
+  @Inject
+  public CelAccessChecker(
+      QuarkusNessieAuthorizationConfig config, CompiledAuthorizationRules compiledRules) {
+    this.config = config;
+    this.compiledRules = compiledRules;
+  }
+
+  @Override
+  public void canViewReference(AccessContext context, NamedRef ref) throws AccessControlException {
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.VIEW_REFERENCE);
+  }
+
+  @Override
+  public void canCreateReference(AccessContext context, NamedRef ref)
+      throws AccessControlException {
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.CREATE_REFERENCE);
+  }
+
+  @Override
+  public void canAssignRefToHash(AccessContext context, NamedRef ref)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.ASSIGN_REFERENCE_TO_HASH);
+  }
+
+  @Override
+  public void canDeleteReference(AccessContext context, NamedRef ref)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.DELETE_REFERENCE);
+  }
+
+  @Override
+  public void canReadEntries(AccessContext context, NamedRef ref) throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.READ_ENTRIES);
+  }
+
+  @Override
+  public void canListCommitLog(AccessContext context, NamedRef ref) throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.LIST_COMMIT_LOG);
+  }
+
+  @Override
+  public void canCommitChangeAgainstReference(AccessContext context, NamedRef ref)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnReference(context, ref, AuthorizationRuleType.COMMIT_CHANGE_AGAINST_REFERENCE);
+  }
+
+  @Override
+  public void canReadEntityValue(
+      AccessContext context, NamedRef ref, ContentsKey key, String contentsId)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnPath(context, key, AuthorizationRuleType.READ_ENTITY_VALUE);
+  }
+
+  @Override
+  public void canUpdateEntity(
+      AccessContext context, NamedRef ref, ContentsKey key, String contentsId)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnPath(context, key, AuthorizationRuleType.UPDATE_ENTITY);
+  }
+
+  @Override
+  public void canDeleteEntity(
+      AccessContext context, NamedRef ref, ContentsKey key, String contentsId)
+      throws AccessControlException {
+    canViewReference(context, ref);
+    canPerformOpOnPath(context, key, AuthorizationRuleType.DELETE_ENTITY);
+  }
+
+  private String getRoleName(AccessContext context) {
+    return null != context.user() ? context.user().getName() : "";
+  }
+
+  private void canPerformOpOnReference(
+      AccessContext context, NamedRef ref, AuthorizationRuleType type) {
+    if (!config.enabled()) {
+      return;
+    }
+    String roleName = getRoleName(context);
+    ImmutableMap<String, Object> arguments =
+        ImmutableMap.of("ref", ref.getName(), "role", roleName, "op", type.name());
+
+    Supplier<String> errorMsgSupplier =
+        () ->
+            String.format(
+                "'%s' is not allowed for role '%s' on reference '%s'",
+                type, roleName, ref.getName());
+    canPerformOp(arguments, errorMsgSupplier);
+  }
+
+  private void canPerformOpOnPath(
+      AccessContext context, ContentsKey contentsKey, AuthorizationRuleType type) {
+    if (!config.enabled()) {
+      return;
+    }
+    String roleName = getRoleName(context);
+    ImmutableMap<String, Object> arguments =
+        ImmutableMap.of("path", contentsKey.toPathString(), "role", roleName, "op", type.name());
+
+    Supplier<String> errorMsgSupplier =
+        () ->
+            String.format(
+                "'%s' is not allowed for role '%s' on content '%s'",
+                type, roleName, contentsKey.toPathString());
+
+    canPerformOp(arguments, errorMsgSupplier);
+  }
+
+  private void canPerformOp(Map<String, Object> arguments, Supplier<String> errorMessageSupplier) {
+    boolean allowed =
+        compiledRules.getRules().entrySet().stream()
+            .anyMatch(
+                entry -> {
+                  try {
+                    return entry.getValue().execute(Boolean.class, arguments);
+                  } catch (ScriptException e) {
+                    throw new RuntimeException(
+                        String.format(
+                            "Failed to execute authorization rule with id '%s' and expression '%s' due to: %s",
+                            entry.getKey(), entry.getValue(), e.getMessage()),
+                        e);
+                  }
+                });
+    if (!allowed) {
+      throw new AccessControlException(errorMessageSupplier.get());
+    }
+  }
+}
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server.authz;
+
+import static org.projectnessie.server.authz.CelAccessChecker.AuthorizationRuleType.VIEW_REFERENCE;
+
+import com.google.common.collect.ImmutableMap;
+import io.quarkus.runtime.Startup;
+import java.util.HashMap;
+import java.util.Map;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.projectnessie.cel.tools.Script;
+import org.projectnessie.cel.tools.ScriptException;
+import org.projectnessie.server.config.QuarkusNessieAuthorizationConfig;
+import org.projectnessie.services.cel.CELUtil;
+
+/**
+ * Compiles the authorization rules from {@link QuarkusNessieAuthorizationConfig} at startup and
+ * provides access to them via {@link CompiledAuthorizationRules#getRules()}.
+ */
+@Singleton
+@Startup
+public class CompiledAuthorizationRules {
+  private final QuarkusNessieAuthorizationConfig config;
+  private final Map<String, Script> compiledRules;
+  private static final String ALLOW_VIEWING_ALL_REFS_ID = "__ALLOW_VIEWING_REF_ID";
+  private static final String ALLOW_VIEWING_ALL_REFS =
+      String.format("op=='%s' && ref.matches('.*')", VIEW_REFERENCE);
+
+  @Inject
+  public CompiledAuthorizationRules(QuarkusNessieAuthorizationConfig config) {
+    this.config = config;
+    this.compiledRules = compileAuthorizationRules();
+  }
+
+  /**
+   * Compiles all authorization rules and returns them.
+   *
+   * @return A map of compiled authorization rules
+   */
+  private Map<String, Script> compileAuthorizationRules() {
+    Map<String, String> rules = new HashMap<>(config.rules());
+    // by default we allow viewing all references until there's a user-defined VIEW_REFERENCE rule
+    if (rules.entrySet().stream().noneMatch(r -> r.getValue().contains(VIEW_REFERENCE.name()))) {
+      rules.put(ALLOW_VIEWING_ALL_REFS_ID, ALLOW_VIEWING_ALL_REFS);
+    }
+    Map<String, Script> scripts = new HashMap<>();
+    rules.forEach(
+        (key, value) ->
+            scripts.computeIfAbsent(
+                value,
+                (k) -> {
+                  try {
+                    return CELUtil.SCRIPT_HOST
+                        .buildScript(value)
+                        .withContainer(CELUtil.CONTAINER)
+                        .withDeclarations(CELUtil.AUTHORIZATION_RULE_DECLARATIONS)
+                        .build();
+                  } catch (ScriptException e) {
+                    throw new RuntimeException(
+                        String.format(
+                            "Failed to compile authorization rule with id '%s' and expression '%s' due to: %s",
+                            key, value, e.getMessage()),
+                        e);
+                  }
+                }));
+    return ImmutableMap.copyOf(scripts);
+  }
+
+  /**
+   * Returns a map of compiled authorization rules.
+   *
+   * @return A map of compiled authorization rules
+   */
+  public Map<String, Script> getRules() {
+    return compiledRules;
+  }
+}
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -19,6 +19,47 @@
 nessie.server.default-branch=main
 nessie.server.send-stacktrace-to-client=false
 
+## Nessie authorization settings
+### This will perform authorization on branches/tags and contents where rule definitions are
+### using a Common Expression Language (CEL) expression (an intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md).
+### Rule definitions are of the form nessie.server.authorization.rules.<ruleId>="<rule_expression>"
+### Available variables within the <rule_expression> are: 'op' / 'role' / 'ref' / 'path'
+### The 'op' variable in the <rule_expression> can be any of:
+### 'VIEW_REFERENCE', 'CREATE_REFERENCE', 'DELETE_REFERENCE', 'READ_ENTRIES', 'LIST_COMMIT_LOG',
+### 'COMMIT_CHANGE_AGAINST_REFERENCE', 'ASSIGN_REFERENCE_TO_HASH', 'UPDATE_ENTITY', 'READ_ENTITY_VALUE', 'DELETE_ENTITY'
+### The 'role' refers to the user's role and can be any string
+### The 'ref' refers to a string representing a branch/tag name
+### The 'path' refers to the Key for the contents of an object and can be any string
+### Some "use-case-based" example rules are shown below (in practice you might rather create a single rule that allows e.g. branch creation/deletion/commits/...):
+# nessie.server.authorization.enabled=false
+# nessie.server.authorization.rules.allow_branch_listing=\
+#   op=='VIEW_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_branch_creation=\
+#   op=='CREATE_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_branch_deletion=\
+#   op=='DELETE_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_listing_commitlog=\
+#   op=='LIST_COMMIT_LOG' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_entries_reading=\
+#   op=='READ_ENTRIES' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_assigning_ref_to_hash=\
+#   op=='ASSIGN_REFERENCE_TO_HASH' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_commits=\
+#   op=='COMMIT_CHANGE_AGAINST_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_reading_entity_value=\
+#   op=='READ_ENTITY_VALUE' && role=='test_user' && path.startsWith('allowed.')
+# nessie.server.authorization.rules.allow_updating_entity=\
+#   op=='UPDATE_ENTITY' && role=='test_user' && path.startsWith('allowed.')
+# nessie.server.authorization.rules.allow_deleting_entity=\
+#   op=='DELETE_ENTITY' && role=='test_user' && path.startsWith('allowed.')
+# nessie.server.authorization.rules.allow_commits_without_entity_changes=\
+#   op=='COMMIT_CHANGE_AGAINST_REFERENCE' && role=='test_user2' && ref.startsWith('allowedBranch')
+# nessie.server.authorization.rules.allow_all=\
+#   op in ['VIEW_REFERENCE','CREATE_REFERENCE','DELETE_REFERENCE','LIST_COMMITLOG','READ_ENTRIES','LIST_COMMIT_LOG',\
+#   'COMMIT_CHANGE_AGAINST_REFERENCE','ASSIGN_REFERENCE_TO_HASH','UPDATE_ENTITY','READ_ENTITY_VALUE','DELETE_ENTITY'] \
+#   && role=='admin_user'
+
+
 ### which type of version store to use: JGIT, INMEMORY, DYNAMO. JGIT is best for local testing, DYNAMO preferred for production
 nessie.version.store.type=INMEMORY
 
@@ -131,5 +172,33 @@ quarkus.container-image.name=nessie
 %test.quarkus.security.users.embedded.roles.admin_user=admin,user
 %test.quarkus.security.users.embedded.roles.test_user=test123
 %test.nessie.version.store.jgit.type=INMEMORY
+
+## test overrides for authorization rule tests
+%test.nessie.server.authorization.rules.allow_all=\
+  op in ['VIEW_REFERENCE','CREATE_REFERENCE','DELETE_REFERENCE','LIST_COMMITLOG','READ_ENTRIES','LIST_COMMIT_LOG',\
+  'COMMIT_CHANGE_AGAINST_REFERENCE','ASSIGN_REFERENCE_TO_HASH','UPDATE_ENTITY','READ_ENTITY_VALUE','DELETE_ENTITY'] \
+  && role=='admin_user'
+%test.nessie.server.authorization.rules.allow_branch_listing=\
+  op=='VIEW_REFERENCE' && role.startsWith('test_user') && ref.matches('.*')
+%test.nessie.server.authorization.rules.allow_branch_creation=\
+  op=='CREATE_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_branch_deletion=\
+  op=='DELETE_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_listing_commitlog=\
+  op=='LIST_COMMIT_LOG' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_entries_reading=\
+  op=='READ_ENTRIES' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_assigning_ref_to_hash=\
+  op=='ASSIGN_REFERENCE_TO_HASH' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_commits=\
+  op=='COMMIT_CHANGE_AGAINST_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')
+%test.nessie.server.authorization.rules.allow_reading_entity_value=\
+  op=='READ_ENTITY_VALUE' && role=='test_user' && path.startsWith('allowed.')
+%test.nessie.server.authorization.rules.allow_updating_entity=\
+  op=='UPDATE_ENTITY' && role=='test_user' && path.startsWith('allowed.')
+%test.nessie.server.authorization.rules.allow_deleting_entity=\
+  op=='DELETE_ENTITY' && role=='test_user' && path.startsWith('allowed.')
+%test.nessie.server.authorization.rules.allow_commits_without_entity_changes=\
+  op=='COMMIT_CHANGE_AGAINST_REFERENCE' && role=='test_user2' && ref.startsWith('allowedBranch')
 
 mp.openapi.extensions.smallrye.operationIdStrategy=METHOD
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
@@ -1,0 +1,304 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.quarkus.test.security.TestSecurity;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.projectnessie.api.ContentsApi;
+import org.projectnessie.api.TreeApi;
+import org.projectnessie.api.params.CommitLogParams;
+import org.projectnessie.api.params.EntriesParams;
+import org.projectnessie.client.NessieClient;
+import org.projectnessie.client.rest.NessieForbiddenException;
+import org.projectnessie.error.NessieConflictException;
+import org.projectnessie.error.NessieNotFoundException;
+import org.projectnessie.model.Branch;
+import org.projectnessie.model.CommitMeta;
+import org.projectnessie.model.ContentsKey;
+import org.projectnessie.model.EntriesResponse.Entry;
+import org.projectnessie.model.IcebergTable;
+import org.projectnessie.model.ImmutableDelete;
+import org.projectnessie.model.ImmutableOperations;
+import org.projectnessie.model.ImmutablePut;
+import org.projectnessie.model.Operations;
+import org.projectnessie.model.Reference;
+import org.projectnessie.server.authz.NessieAuthorizationTestProfile;
+
+@QuarkusTest
+@TestProfile(value = NessieAuthorizationTestProfile.class)
+class TestAuthorizationRules {
+
+  private static NessieClient client;
+  private static TreeApi tree;
+  private static ContentsApi contents;
+
+  @BeforeEach
+  void setupClient() {
+    client = NessieClient.builder().withUri("http://localhost:19121/api/v1").build();
+    tree = client.getTreeApi();
+    contents = client.getContentsApi();
+  }
+
+  @AfterEach
+  void closeClient() {
+    if (client != null) {
+      client.close();
+      client = null;
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(booleans = {true, false})
+  @TestSecurity(user = "test_user")
+  void testAllOpsWithTestUser(boolean shouldFail)
+      throws NessieNotFoundException, NessieConflictException {
+    testAllOps("allowedBranchForTestUser", "test_user", shouldFail);
+  }
+
+  @Test
+  @TestSecurity(user = "admin_user")
+  void testAdminUserIsAllowedEverything() throws NessieNotFoundException, NessieConflictException {
+    testAllOps("testAdminUserIsAllowedAllBranch", "admin_user", false);
+  }
+
+  private void testAllOps(String branchName, String role, boolean shouldFail)
+      throws NessieConflictException, NessieNotFoundException {
+    ContentsKey key = ContentsKey.of("allowed", "x");
+    if (shouldFail) {
+      branchName = "disallowedBranchForTestUser";
+      key = ContentsKey.of("disallowed", "x");
+    }
+
+    createBranch(Branch.of(branchName, null), role, shouldFail);
+
+    Branch branchWithInvalidHash = Branch.of(branchName, "1234567890123456");
+    Branch branch =
+        shouldFail ? branchWithInvalidHash : retrieveBranch(branchName, role, shouldFail);
+
+    listAllReferences(branchName, shouldFail);
+
+    ImmutableOperations createOps =
+        ImmutableOperations.builder()
+            .addOperations(ImmutablePut.builder().key(key).contents(IcebergTable.of("foo")).build())
+            .commitMeta(CommitMeta.fromMessage("add stuff"))
+            .build();
+    addContent(branch, createOps, role, shouldFail);
+
+    getCommitLog(branchName, role, shouldFail);
+    getEntriesFor(branchName, role, shouldFail);
+    readContent(branchName, key, role, shouldFail);
+
+    branch = shouldFail ? branchWithInvalidHash : retrieveBranch(branchName, role, shouldFail);
+
+    ImmutableOperations deleteOps =
+        ImmutableOperations.builder()
+            .addOperations(ImmutableDelete.builder().key(key).build())
+            .commitMeta(CommitMeta.fromMessage("delete stuff"))
+            .build();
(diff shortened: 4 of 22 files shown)


servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * A reference implementation of the {@link AccessChecker} that performs access checks using CEL
+ * expressions.
+ */
+@ApplicationScoped
+public class CelAccessChecker implements AccessChecker {
+
+  private final QuarkusNessieAuthorizationConfig config;
+
+  public enum AuthorizationRuleType {
```
rymurr, line 49: The static import above of this enum looks strange to my eye. Can it be simplified?

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,197 @@
...
+                            Boolean.class,
+                            ImmutableMap.of(
+                                "ref",
+                                ref.getName(),
+                                "role",
+                                getRoleName(context),
+                                "op",
+                                type.name()));
+                  } catch (ScriptException e) {
+                    throw new ForbiddenException(
```
rymurr, line 141: Is this really a forbidden exception? Or is it a 500 error?

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * Checks whether the given role/principal is allowed to read an entity value as defined by the
+   * {@link ContentsKey} for the given Branch/Tag. It is up to the implementor to define which
+   * exception should being thrown in case of denial.
+   *
+   * @param context The context carrying the principal information.
+   * @param ref The {@link NamedRef} to check
+   * @param key The {@link ContentsKey} to check
+   */
+  void canReadEntityValue(AccessContext context, NamedRef ref, ContentsKey key);
```
rymurr, line 94: we may have to provide the contents object or contents.uuid here and in other places contentskey is used. That complicates how/when this is called though

servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.security.Principal;
+
+/** Provides some context about a role/principal that accesses Nessie resources. */
+public interface AccessContext {
+  /** Provide a unique id for the operation being validated (for correlation purposes). */
+  String operationId();
+
+  /** Provide the user identity. */
+  Principal user();
```
rymurr, line 26: perhaps instead of adding contents.id to the interface it could be added here as Optional<String>?

servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java
```diff
@@ -60,9 +63,10 @@ protected SecurityContext getSecurityContext() {
   @Override
   public Contents getContents(ContentsKey key, String namedRef, String hashOnRef)
       throws NessieNotFoundException {
-    Hash ref = namedRefWithHashOrThrow(namedRef, hashOnRef).getHash();
+    WithHash<NamedRef> ref = namedRefWithHashOrThrow(namedRef, hashOnRef);
+    getAccessChecker().canReadEntityValue(createAccessContext(), ref.getValue(), key);
```
rymurr, line 67: Is it not possible to add this as a filter or something?

rymurr (commented):
Few minor points in comments.

Also:
1) I guess this is inert till a reference impl of authn is added?
2) we didn't change the config endpoint. I guess the assumption is anyone logged in can read server config?
3) We should probably add to the openapi spec that throwing 403 is possible for these endpoints

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: imo it should be a `ForbiddenException` since it indicates that access to a resource has been forbidden (thus throwing a HTTP 403). Throwing a 500 error would mean something went wrong on the server-side , which the client usually can't do anything about

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
nastra: agreed that this makes it a bit more complicated

servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java
nastra: that means that non-related stuff "leaks" into this, which I'm not really happy with

servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java
nastra: you mean like a HTTP filter?

rymurr: yes, a jax-rs filter. I guess not easily as each  call to the access checker is slightly different. It just feels a bit icky to have the authz logic embedded in the operational logic

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
rymurr: This call happens when something goes wrong w/ calculating permissions though right? It means that the server rules contain a syntax error or some other server side error. This doesn't mean the user is unauthorised, rather that the server is having an issue.

nastra: oh you're absolutely right. I think I copy/pasted the wrong exception during refactoring, so this definitely should be a 5xx

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,186 @@
...
+    boolean allowed =
+        config.rules().entrySet().stream()
+            .anyMatch(
+                entry -> {
+                  try {
+                    return CELUtil.SCRIPT_HOST
+                        .buildScript(entry.getValue().replace("\"", ""))
+                        .withContainer(CELUtil.CONTAINER)
+                        .withDeclarations(CELUtil.AUTHORIZATION_RULE_DECLARATIONS)
+                        .build()
```
snazy, line 162: The built scripts should really be held in a "global" map (like a ConcurrentHashMap using computeIfAbsent), because parsing and checking the script every time an access needs to be checked it too expensive.

Might be worth a PR against cel-java, which could then use for example caffeine instead of a "stupid CHM" ;)

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,186 @@
...
```
snazy, line 119: (same comment as below re: parsing+checking)

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,186 @@
...
+    if (!config.enabled()) {
+      return;
+    }
+    boolean allowed =
+        config.rules().entrySet().stream()
+            .anyMatch(
+                entry -> {
+                  try {
+                    return CELUtil.SCRIPT_HOST
+                        .buildScript(entry.getValue().replace("\"", ""))
```
snazy, line 159: Why does it need to replace here at all? The map should contain the just script IMO.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,186 @@
...
```
snazy, line 116: (same as below comment re: replacing)

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
snazy: If (really, *if*) the CEL expressions are rather constant (like defined in the configuration at Quarkus-image-build-time), the actual CEL parse+check operations should ideally be part of the Quarkus-image-build and not slow down the startup.

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
snazy: Checking against the value of a `Contents` object should be a separate operation applied on the returned values IMO, because otherwise it requires a additional reads before the access check.

Not sure whether it makes a lot of sense to couple the contents with authz rules.

(Checking for the contents-key is fine IMO.)

servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java
snazy: It should be possible to abstract authz checking from the actual implementation.
Either by delegation or "naive `public class AuthzContentsResource extends ContentsResource`".

It feels, that it should also be possible to have a separate module `nessie-authz` next to `servers/services` with the "right" CDI annotations, so a `AuthzContentsResource` + `AuthzTreeResource` are preferred by the container over the "raw" `ContentsResource` + `TreeResource`.

snazy (changes requested):

servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java
nastra: I agree it should definitely be outside of the operational logic. My first goal was to have a prototype so that we can decide whether we want to configure authz rules via a quarkus config

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: I didn't focus too much on perf in this iteration, but I totally agree that this should be cached as there's no need to parse/check scripts all the time, even though they don't change during runtime

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: the rule string is wrapped in `"` when we define it in the config `nessie.server.authorization.rules.allow_branch_creation="<rule>"`. Technically I think it would work without wrapping it in `"` but defining`nessie.server.authorization.rules.allow_branch_creation=<rule>` looks a bit weird

laurentgo (changes requested):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)* status: 0 of 32 files reviewed, 10 unresolved discussions (waiting on @nastra, @rymurr, and @snazy)

___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 94 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeeDx72xn_J02Ju3D_K-r2-94:-MfDkORA3NS-zuJxQkyV:blephjk) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L94)):*

Checking against the value of a `Contents` object should be a separate operation applied on the returned values IMO, because otherwise it requires a additional reads before the access check.

Not sure whether it makes a lot of sense to couple the contents with authz rules.

(Checking for the contents-key is fine IMO.)
</blockquote></details>

Having the object uuid might be important though as an implementor might track different permissions for different objects but under the same name...
But note that to provide the content or part of the content, I guess Nessie cannot just provide namedref but would actually need the actual hash?
___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 30 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MfDjwCU453H4lz7bCl-:-MfDjwCU453H4lz7bCl0:b-b5k1dx) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L30)):*
> ```
>    * Checks whether the given role/principal is allowed to create a Branch/Tag. It is up to the
>    * implementor to define which exception should being thrown in case of denial.
> ```
I'm not sure why it should be up to the implementor to define which exception to throw instead of the interface to be somewhat prescriptive. Otherwise how can server know how to handle with exception being thrown and doing the mapping correctly?
___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java, line 26 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeeDx72xn_J02Ju3D_L-r2-26:-MfIvcZI6V8o66I6ZVdG:bevbiyv) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java#L26)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

that means that non-related stuff "leaks" into this, which I'm not really happy with
</blockquote></details>

I would agree with Eduard: we should keep separate the access context (who's accessing Nessie) from the query context (what is accessed)

servers/services/src/main/java/org/projectnessie/services/rest/ContentsResource.java
nastra: I experimented a bit with Jax-RS filters, but this really makes things quite difficult as for some checks (e.g. when creating a branch/tag, which is a `POST`) as we need to access the `POST` body and such in order to perform our checks. Thus I figured it's easier to have a `TreeResourceWithAuthorizationChecks` class that performs those checks and then just delegates to `TreeResource` for the normal functionality.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: added `CompiledAuthorizationRules` that compiles the rules during startup

nastra (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)* status: 0 of 35 files reviewed, 10 unresolved discussions (waiting on @laurentgo, @nastra, @rymurr, and @snazy)

___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 141 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r2-141:-MfbGvst7QcSOtFtcuG0:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L141)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

oh you're absolutely right. I think I copy/pasted the wrong exception during refactoring, so this definitely should be a 5xx
</blockquote></details>

Done.
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 162 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r3-162:-MfbGxcJ1ApomyS4GMnY:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L162)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

added `CompiledAuthorizationRules` that compiles the rules during startup
</blockquote></details>

Done.
___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 30 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MfDjwCU453H4lz7bCl-:-MfbHdDOC8VpPylrKmzW:b-7ql1td) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L30)):*

> ```
>    * Checks whether the given role/principal is allowed to create a Branch/Tag. It is up to the
>    * implementor to define which exception should being thrown in case of denial.
> ```
I'm not sure why it should be up to the implementor to define which exception to throw instead of the interface to be somewhat prescriptive. Otherwise how can server know how to handle with exception being thrown and doing the mapping correctly?
</blockquote></details>

fixed

nastra:
> 2. we didn't change the config endpoint. I guess the assumption is anyone logged in can read server config?

that's my assumption as well, therefore I didn't add anything for the config stuff.


> 3. We should probably add to the openapi spec that throwing 403 is possible for these endpoints

done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Compiles the authorization rules from {@link QuarkusNessieAuthorizationConfig} at startup and
+ * provides access to them via {@link CompiledAuthorizationRules#getRules()}.
+ */
+@Singleton
+@Startup
+public class CompiledAuthorizationRules {
+  private final QuarkusNessieAuthorizationConfig config;
+  private final Map<String, Script> compiledRules;
```
rymurr, line 36: can you copy this into an immutable map so we aren't returning a mutable map in teh getter

rymurr (commented):
lgmt other than 1 minor comment

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
nastra: thanks for bringing this up, this is definitely needed. I pushed a fix.

nastra (commented):
> Reviewed 3 of 29 files at r1, 23 of 28 files at r2, 1 of 2 files at r3, 4 of 6 files at r4, 4 of 4 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)* status: all files reviewed, 11 unresolved discussions (waiting on @laurentgo, @rymurr, and @snazy)

___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 141 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r2-141:-MfboUqm9Ia6DiAy_hb8:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L141)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Done.
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 116 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r3-116:-MfboTZoDQQB86rvcCJf:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L116)):*

(same as below comment re: replacing)
</blockquote></details>

Done.
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 119 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r3-119:-MfboTzi-7YdHDZTJfuM:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L119)):*

(same comment as below re: parsing+checking)
</blockquote></details>

Done.
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java, line 159 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeyZ_sbOz-LQZYLy-UV-r3-159:-MfboVfQBkLnM_POy2j-:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java#L159)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

the rule string is wrapped in `"` when we define it in the config `nessie.server.authorization.rules.allow_branch_creation="<rule>"`. Technically I think it would work without wrapping it in `"` but defining`nessie.server.authorization.rules.allow_branch_creation=<rule>` looks a bit weird
</blockquote></details>

Done.
___
*[servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java, line 36 at r7](https://reviewable.io/reviews/projectnessie/nessie/1608#-MfazSDGx0KosIHspjmM-r7-36:-MfboZEP9M2WPs2KiVqW:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/851dfe9db13a4d07a3f3a27fec48aff9cb3f4ef9/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java#L36)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

thanks for bringing this up, this is definitely needed. I pushed a fix.
</blockquote></details>

Done.
___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 30 at r3](https://reviewable.io/reviews/projectnessie/nessie/1608#-MfDjwCU453H4lz7bCl-:-Mfboa30-oZ0pShTUGjt:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/669b6db685102fe3d0605a609262ae993a4e8c59/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L30)):*
<details><summary><i>Previously, nastra (Eduard Tudenhöfner) wrote…</i></summary><blockquote>

fixed
</blockquote></details>

Done.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -0,0 +1,143 @@
...
+import org.projectnessie.services.config.ServerConfig;
+import org.projectnessie.versioned.BranchName;
+import org.projectnessie.versioned.NamedRef;
+import org.projectnessie.versioned.TagName;
+import org.projectnessie.versioned.VersionStore;
+
+/** Does authorization checks (if enabled) on the {@link TreeResource} endpoints. */
+@RequestScoped
+@Default
+public class TreeResourceWithAuthorizationChecks extends TreeResource {
```
snazy, line 49: This is missing checks for `TreeApi.getAllReferences` and `getReferenceByName`

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
```diff
@@ -0,0 +1,83 @@
...
+                        return CELUtil.SCRIPT_HOST
+                            .buildScript(value.replace("\"", ""))
+                            .withContainer(CELUtil.CONTAINER)
+                            .withDeclarations(CELUtil.AUTHORIZATION_RULE_DECLARATIONS)
+                            .build();
+                      } catch (ScriptException e) {
+                        throw new RuntimeException(
+                            String.format(
+                                "Failed to compile query expression with id '%s' and expression '%s' due to: %s",
+                                key, value, e.getMessage()));
```
snazy, line 69: Let's add `e` as the cause to the thrown RE (but keep `e.getMessage()` in the message

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * Compiles all authorization rules and returns them.
+   *
+   * @return A map of compiled authorization rules
+   */
+  private Map<String, Script> compileAuthorizationRules() {
+    Map<String, Script> scripts = new HashMap<>();
+    config
```
snazy, line 52: Can simplify with `return config.rules().collect(e -> e.getValue(), e -> (...build-script...));`

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -19,6 +19,29 @@
...
+### Rule definitions are of the form nessie.server.authorization.rules.<ruleId>="<rule_expression>"
+### Available variables within the <rule_expression> are: 'op' / 'role' / 'ref' / 'path'
+### The 'op' variable in the <rule_expression> can be any of:
+### 'CREATE_REFERENCE', 'DELETE_REFERENCE', 'READ_ENTRIES', 'LIST_COMMIT_LOG', 'COMMIT_CHANGE_AGAINST_REFERENCE', 'ASSIGN_REFERENCE_TO_HASH', 'UPDATE_ENTITY', 'READ_ENTITY_VALUE', 'DELETE_ENTITY'
+### The 'role' refers to the user's role and can be any string
+### The 'ref' refers to a string representing a branch/tag name
+### The 'path' refers to the Key for the contents of an object and can be any string
+### Some example rules are shown below:
+# nessie.server.authorization.enabled=false
+# nessie.server.authorization.rules.allow_branch_creation="op=='CREATE_REFERENCE' && role.startsWith('test_user') && ref.startsWith('allowedBranch')"
```
snazy, line 34: Can you change these examples to some say "use case based" examples - like "allow all too all users named 'admin'" or "allow only reads to all users names 'foo'"?
Having one expression per access-check is probably a bad example, because it's not necessary to do so.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+import org.projectnessie.services.authz.AccessChecker;
+import org.projectnessie.services.authz.AccessContext;
+import org.projectnessie.versioned.NamedRef;
+
+/**
+ * A reference implementation of the {@link AccessChecker} that performs access checks using CEL
+ * expressions.
+ */
+@ApplicationScoped
+public class CelAccessChecker implements AccessChecker {
```
snazy, line 34: Better make this class abstract (+ remove the `@ApplicationScoped` + move it to nessie-services) and have concrete implementations for the individual server implementations. That would give server-specific implementations freedom to extract the principal-name and role/group information.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+                entry -> {
+                  try {
+                    return entry
+                        .getValue()
+                        .execute(
+                            Boolean.class,
+                            ImmutableMap.of(
+                                "ref",
+                                ref.getName(),
+                                "role",
```
snazy, line 126: Think, this should be named `principal` or even `user`?

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+    boolean allowed =
+        compiledRules.getRules().entrySet().stream()
+            .anyMatch(
+                entry -> {
+                  try {
+                    return entry
+                        .getValue()
+                        .execute(
+                            Boolean.class,
+                            ImmutableMap.of(
```
snazy, line 123: How can we check role or direct group memberships? Mean, updating all security policies when users come and go isn't great. The default implementation in this class may return an empty set, but implementations of this class for concrete servers can easily do extract role/group memberships, which makes security-policy-checks easier.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+                    throw new RuntimeException(
+                        String.format(
+                            "Failed to execute query expression with id '%s' and expression '%s' due to: %s",
+                            entry.getKey(), entry.getValue(), e.getMessage()));
+                  }
+                });
+    if (!allowed) {
+      throw new AccessControlException(
+          String.format(
+              "'%s' is not allowed for Role '%s' on Reference '%s'",
```
snazy, line 140: typo: subjects start with a lower char in engish ;)

snazy, line 140: Would also be nice to include the op-name in the message

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+              type, getRoleName(context), ref.getName()));
+    }
+  }
+
+  private String getRoleName(AccessContext context) {
+    return null != context.user() ? context.user().getName() : "";
```
snazy, line 146: Nit: extract `context.user()` into a variable (you don't know here how expensive that call actually is)

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  private void canPerformOpOnPath(
```
snazy, line 149: Think, you can unify this method with `canPerformOpOnReference` by preparding the map passed to `execute` in advance, which would also be cheaper for the runtime, because all script-args are the same.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,184 @@
...
+                                "ref",
+                                ref.getName(),
+                                "role",
+                                getRoleName(context),
+                                "op",
+                                type.name()));
+                  } catch (ScriptException e) {
+                    throw new RuntimeException(
+                        String.format(
+                            "Failed to execute query expression with id '%s' and expression '%s' due to: %s",
```
snazy, line 133: typo: better: `Security policy script execution ...`

servers/quarkus-server/src/main/resources/application.properties
snazy, line 34: Would be easier to read the rules if those were slip over multiple lines (this is rather a standard java properties file, so i doubt you need to quote the value and can use `\` to wrap lines.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -19,6 +19,29 @@
...
 nessie.server.send-stacktrace-to-client=false
 
+## Nessie authorization settings
+### This will perform authorization on branches/tags and contents where rule definitions are
+### using a Common Expression Language (CEL) expression (an intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md).
+### Rule definitions are of the form nessie.server.authorization.rules.<ruleId>="<rule_expression>"
+### Available variables within the <rule_expression> are: 'op' / 'role' / 'ref' / 'path'
+### The 'op' variable in the <rule_expression> can be any of:
+### 'CREATE_REFERENCE', 'DELETE_REFERENCE', 'READ_ENTRIES', 'LIST_COMMIT_LOG', 'COMMIT_CHANGE_AGAINST_REFERENCE', 'ASSIGN_REFERENCE_TO_HASH', 'UPDATE_ENTITY', 'READ_ENTITY_VALUE', 'DELETE_ENTITY'
+### The 'role' refers to the user's role and can be any string
```
snazy, line 29: I think, this is missing another `roles` parameter, which contains the names of the roles the logged in principal is assigned to.

snazy, line 29: I assume, it's actually okay to rename `role` to `user` in this case - `user` is also probably easier to grasp.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -130,4 +153,17 @@ quarkus.container-image.name=nessie
 %test.quarkus.security.users.embedded.roles.test_user=test123
 %test.nessie.version.store.jgit.type=INMEMORY
 
+## test overrides for authorization rule tests
```
snazy, line 156: Move these into `NessieAuthorizationTestProfile` ?

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
```diff
@@ -0,0 +1,276 @@
...
+import org.projectnessie.model.IcebergTable;
+import org.projectnessie.model.ImmutableDelete;
+import org.projectnessie.model.ImmutableOperations;
+import org.projectnessie.model.ImmutablePut;
+import org.projectnessie.model.Operations;
+import org.projectnessie.server.authz.NessieAuthorizationTestProfile;
+
+@QuarkusTest
+@TestProfile(value = NessieAuthorizationTestProfile.class)
+class TestAuthorizationRules {
```
snazy, line 51: I think, this test can perfectly live in `nessie-services`, with a mocked `VersionStore` store implementation, so we do not have the "only" test for rule only in `nessie-quarkus`.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
nastra: @laurentgo were those left out from the original design doc on purpose? My assumption was that those were left out on purpose as you still want users to be able to retrieve all available references / a single reference.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
nastra: done

servers/services/src/main/java/org/projectnessie/services/authz/ServerAccessContext.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.services.authz;
+
+import java.security.Principal;
+
+public class ServerAccessContext implements AccessContext {
```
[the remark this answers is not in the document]
nastra: done

servers/quarkus-server/src/main/resources/application.properties
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: not sure why we would want to do that? The `CelAccessChecker` is a concrete implementation that depends on Quarkus-specific things (like the config) that just shows a way of how it can be done.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: already done as can be seen in the tests, such as https://github.com/projectnessie/nessie/pull/1608/files#diff-92a3006c29e0a5ca0159acc7713e3347bf66cc35f63ee3f60169f1821e3eed77R146

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: adjusted to `Failed to execute authorization rule with id..` as we're talking about authorization rules everywhere else

servers/quarkus-server/src/main/resources/application.properties
nastra: for me `user` seems to be more narrow in describing what it is, therefore I thought keeping `role` gives us slightly more flexibility in describing what we deal with here

nastra: in the reference implementation I'm not sure we want to distinguish directly between a user and its assigned roles. I think dealing with `roles` should be fine for now. We can always re-visit if necessary

servers/quarkus-server/src/main/resources/application.properties
nastra: the idea for keeping them here is to also have some kind of documentation for rules that we use in testing.

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
nastra: I think it needs to live in `nessie-quarkus`, otherwise we'll have to pull in quarkus-related dependencies into `nessie-services` (for `QuarkusTest / `TestProfile` / `TestSecurity`)

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: as mentioned in the other comments, I think `role` gives us slightly more flexibility.

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: I agree that this isn't great but for this implementation I don't think we were planning to do group membership checks and such

servers/quarkus-server/src/main/resources/application.properties
nastra: agreed and fixed

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
```diff
@@ -0,0 +1,123 @@
...
+import java.security.AccessControlException;
+import org.projectnessie.model.ContentsKey;
+import org.projectnessie.versioned.NamedRef;
+
+/**
+ * The purpose of the {@link AccessChecker} is to make sure that a particular role is allowed to
+ * perform an action (such as creation, deletion) on a {@link NamedRef} (Branch/Tag). Additionally,
+ * this interface also provides checks based on a given {@link ContentsKey}.
+ */
+public interface AccessChecker {
```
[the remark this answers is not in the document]
nastra: `canMerge` & `canTransplant` are both handled via the `canCommitChangeAgainstReference` check

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * Checks whether the given role/principal is allowed to assign the given Branch/Tag to a Hash.
+   *
+   * @param context The context carrying the principal information.
+   * @param ref The {@link NamedRef} to check
+   * @throws AccessControlException When the permission to assign the given Branch/Tag to a Hash is
+   *     not granted.
+   */
+  void canAssignRefToHash(AccessContext context, NamedRef ref) throws AccessControlException;
```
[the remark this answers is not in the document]
nastra: I would prefer to have `canAssignRefToHash` as that seems to be a bit clearer to me

snazy (commented):
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)* status: 27 of 38 files reviewed, 26 unresolved discussions (waiting on @laurentgo, @nastra, @rymurr, and @snazy)

___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 94 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeeDx72xn_J02Ju3D_K-r2-94:-Mfl3pCY9_rCXE3SXvPz:bq20fae) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L94)):*

Having the object uuid might be important though as an implementor might track different permissions for different objects but under the same name...
But note that to provide the content or part of the content, I guess Nessie cannot just provide namedref but would actually need the actual hash?
</blockquote></details>

Not sure whether "the" UUID would be helpful here. In case of Iceberg there are also two UUIDs - both are unique, but different: one is Contents.id the other is Iceberg's TableMetadata.id. The Contents.id is only available to the check, after the read actually happened.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
nastra: talked to @laurentgo yesterday and not having checks for `getAllReferences` and `getReferenceByName` was an oversight and didn't make it into the design doc. I updated the design doc and the SPI and all the other related stuff

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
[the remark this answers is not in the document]
nastra: also added `canViewReference(AccessContext context, NamedRef ref)`

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
rymurr: Does this mean we are explicitly saying the reference impl of authz only works in Quarkus? Is that documented anywhere?

servers/services/src/main/java/org/projectnessie/services/authz/ServerAccessContext.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.security.Principal;
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+
+@Value.Immutable(prehash = true)
+public abstract class ServerAccessContext implements AccessContext {
+
+  @Override
+  public abstract String operationId();
```
rymurr, line 26: nit: we have previously been keeping with the `getOperationId` syntax for getters

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
nastra: it's not explicitly documented, but rather implicit since all of the authorization rules are defined in a quarkus config.

servers/services/src/main/java/org/projectnessie/services/authz/ServerAccessContext.java
nastra: that's just how it was named in the original design doc in https://github.com/nastra/nessie/blob/be631849f0c0a75b6f36ffd728b461e7586b1e1f/design/authorization.md#L76. I don't have a preference of whether it should have a `get` prefix or not

rymurr:
> Not sure whether "the" UUID would be helpful here. In case of Iceberg there are also two UUIDs - both are unique, but different: one is Contents.id the other is Iceberg's TableMetadata.id. The Contents.id is only available to the check, after the read actually happened.

I don't think the iceberg uuid is relevant here, its an opaque value to Nessie. When we say uuid we mean the `Contents.id`.  Same goes for the Delta table uuid.

>> Having the object uuid might be important though as an implementor might track different permissions for different objects but under the same name...

The Nessie spec says both `uuid` and the hash of a contents object can be a reference to an object externally. We can either force the implementer to look them up or we can provide them. The downside to providing them is that we can't fail fast, the upside is that the implementer doesn't have to have a version store. I think the upside is stronger: the implementer of the authz spec may not want/need a version store and its potentially annoying to set one up. From a safety standpoint I would prefer the authz plugin can't/doesn't instantiate a version store as the authz plugin shoudln't have the ability to modify the database. The interface in its current form doesn't make that clear and increases the likelyhood that an implementer of the authz interface would do that. I think the interface is more informative/easier to use if it covers all aspects of the spec wrt object references.

>> But note that to provide the content or part of the content, I guess Nessie cannot just provide namedref but would actually need the actual hash? 

I am not sure I undersant this comment @laurentgp

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAuthorizationRules.java
rymurr: It should go in the public facing documentation (eg in `site`) for authz. Will the public documetnation be a follow up PR?

nastra: yep, I was planning to add some docs once the PR is finished/merged

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -0,0 +1,175 @@
...
+  private final CompiledAuthorizationRules compiledRules;
+
+  public enum AuthorizationRuleType {
+    VIEW_REFERENCE,
+    CREATE_REFERENCE,
+    DELETE_REFERENCE,
+    LIST_COMMIT_LOG,
+    READ_ENTRIES,
+    ASSIGN_REFERENCE_TO_HASH,
+    COMMIT_CHANGE_AGAINST_REFERENCE,
```
sandhyasun, line 47: Is  the rule for COMMIT_CHANGE_AGAINST_REFERENCE the one that covers the check for a user allowed to do the "merge" operation ? i.e commit changes from one branch to another ?

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Checks whether the given role/principal is allowed to read an entity value as defined by the
+   * {@link ContentsKey} for the given Branch/Tag.
+   *
+   * @param context The context carrying the principal information.
+   * @param ref The {@link NamedRef} to check
+   * @param key The {@link ContentsKey} to check
+   * @throws AccessControlException When the permission to read an entity value is not granted.
+   */
+  void canReadEntityValue(AccessContext context, NamedRef ref, ContentsKey key)
```
sandhyasun, line 106: What is the difference between canReadEntityValue and canReadEntries ? Also is canViewReferences a more controlled access - i.e only view the list of references but not necessarily read the contents ?

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
nastra: `COMMIT_CHANGE_AGAINST_REFERENCE` covers merge / transplant / commits

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
nastra: `canReadEntries` refers to https://github.com/nastra/nessie/blob/40bc1672eab1295c1eaa76ed65d50f145df91d4e/model/src/main/java/org/projectnessie/model/EntriesResponse.java#L27-L49 (which holds the `ContentsKey`) and `canReadEntityValue` refers to the actual `Contents` (https://github.com/nastra/nessie/blob/52a6141fb3d3b538ecaf1c39006e6160832ec166/model/src/main/java/org/projectnessie/model/Contents.java#L56) when looking it up via `ContentsKey`

servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java
```diff
@@ -0,0 +1,151 @@
...
+  /**
+   * Checks whether the given role/principal is allowed to update an entity value as defined by the
+   * {@link ContentsKey} for the given Branch/Tag.
+   *
+   * @param context The context carrying the principal information.
+   * @param ref The {@link NamedRef} to check
+   * @param key The {@link ContentsKey} to check
+   * @throws AccessControlException When the permission to update an entity value is not granted.
+   */
+  void canUpdateEntity(AccessContext context, NamedRef ref, ContentsKey key)
```
rymurr, line 137: In my mind we should prob put contents id on these methods too. I am not certain of a situation where a user would not be able to read a `ContentsKey` (b/c of the underlying Id) but still be able to update/delete it.

nastra: sgtm, pushed a fix

rymurr (approved):

nastra (commented):
> Reviewed 3 of 7 files at r10, 11 of 11 files at r13.
> *[Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1608)* status: 32 of 38 files reviewed, 30 unresolved discussions (waiting on @laurentgo, @nastra, @rymurr, @sandhyasun, and @snazy)

___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java, line 94 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeeDx72xn_J02Ju3D_K-r2-94:-Mg5GXTn0uRA3KHLQ9lf:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/services/src/main/java/org/projectnessie/services/authz/AccessChecker.java#L94)):*

Not sure whether "the" UUID would be helpful here. In case of Iceberg there are also two UUIDs - both are unique, but different: one is Contents.id the other is Iceberg's TableMetadata.id. The Contents.id is only available to the check, after the read actually happened.
</blockquote></details>

Done.
___
*[servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java, line 26 at r2](https://reviewable.io/reviews/projectnessie/nessie/1608#-MeeDx72xn_J02Ju3D_L-r2-26:-Mg5GaiL7_GmTZDldF55:b-896fix) ([raw file](https://github.com/projectnessie/nessie/blob/923bf4f962e6e045670eced19ef74089f0c0474c/servers/services/src/main/java/org/projectnessie/services/authz/AccessContext.java#L26)):*

I would agree with Eduard: we should keep separate the access context (who's accessing Nessie) from the query context (what is accessed)
</blockquote></details>

Done.

PR #1660, opened 2021-07-27
More flexible CI for PRs

### LINKED ISSUE #1575: More flexible CI for PRs
Once #1574 is merged, we can further simplify and extend the CI workflows for the main branch and PRs.

Merge `.github/workflows/pull-request.yml` into `.github/workflows/main.yml` so there's only one file left to maintain.

Add more flexibility to PR CI runs by controlling how CI runs against a PR using different labels.

For most PRs, we don't need to run the "whole spiel" (Maven `-Pjdk8-tests`, all Python versions). So I'm proposing to default to a "quick" and "not-that-resource-hungry" CI for PRs: no `-Pjdk8-tests` and maybe just one Python version.

And have a new label `CI-Full` for PRs that adds `-Pnative` and `-Pjdk8-tests` and the jackson-integration-tests to enable the "whole CI machinery".

Note: not verified, but it seems that `${{ github.event.label.name == 'CI-Full' }}` could work.
See https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

snazy:
Proposal for three new labels:
* `pr-native` - triggers CI with `native` Maven profile (disabled by default)
* `pr-jdk8` - triggers CI with `jdk8-tests` Maven profile (disabled by default)
* `pr-jackson` - triggers CI with Jackson-client-test matrix (disabled by default)

Fixes #1575

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1660)

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -111,14 +111,25 @@ jobs:
         echo ::set-output name=CI_WITH_JACKSON::$(jq 'any(.pull_request.labels[]; .name == "pr-jackson")' < $GITHUB_EVENT_PATH)
     - name: Build with Maven
       env:
-        CI_WITH_NATIVE: ${{ steps.ci_options.outputs.CI_WITH_NATIVE }}
-        CI_WITH_JDK8: ${{ steps.ci_options.outputs.CI_WITH_JDK8 }}
         SPARK_LOCAL_IP: localhost
       run: |
-        PROFILES="-Pcode-coverage"
-        [[ $CI_WITH_JDK8 == "true" ]] && PROFILES="$PROFILES,jdk8-tests"
-        [[ $CI_WITH_NATIVE == "true" ]] && PROFILES="$PROFILES,native"
-        ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml $PROFILES -Dtest.log.level=WARN
+        ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml -Pcode-coverage -Dtest.log.level=WARN
+    - name: Build with Maven / JDK8 tests
+      if: ${{ steps.ci_options.outputs.CI_WITH_JDK8 == 'true' }}
+      env:
+        SPARK_LOCAL_IP: localhost
+      run: |
+        # No code-coverage here
+        ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml -Pjdk8-tests" -Dtest.log.level=WARN \
+          -pl :nessie-hms-hive2 \
+          -pl :nessie-hms-hive3
+    - name: Build with Maven / Native tests
+      if: ${{ steps.ci_options.outputs.CI_WITH_NATIVE == 'true' }}
+      run: |
+        # No code-coverage here
+        ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml -Pnative -Dtest.log.level=WARN \
+          -pl :nessie-quarkus \
+          -pl :nessie-lambda
     - name: Build with Gradle
       run: ./gradlew --rerun-tasks --no-daemon --info build
       working-directory: ./tools/apprunner-gradle-plugin
@@ -142,6 +153,7 @@ jobs:
         flags: java
         files: |
           code-coverage/target/site/jacoco-aggregate-all/jacoco.xml
+
   jackson-tests:
     name: Jackson Integration Tests
     needs: java
@@ -181,6 +193,7 @@ jobs:
             ${{ runner.os }}-maven-
       - name: Jackson Integration Tests ${{ matrix.target-library }}
         run: mvn verify -pl :nessie-client -am -Pjackson-tests -Djackson.test.version=${{ matrix.jackson-version }} -Dtest.log.level=WARN
+
   python:
     name: Python
     runs-on: ubuntu-latest
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -19,6 +19,8 @@ jobs:
   java:
     name: Java/Maven
     runs-on: ubuntu-latest
+    outputs:
+      CI_WITH_JACKSON: ${{ steps.ci_options.outputs.CI_WITH_JACKSON }}
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


rymurr (commented):
I just created the labels then added them to the build. They didn't trigger anythign new. Does this imply the labels have to be added before the build -> at the creation of the PR? Or are they added then we re-trigger the whole build?

nastra:
It probably needs to trigger CI on label events (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#label)

snazy:
Ah, oversight. Yes, the WF needs to be told to run when labels are added/removed (https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request)

rymurr:
Cool. In that case, to me this feels like they should be separate jobs (not sure of the terminology). New items in the list of checks basically. So adding a label 'x' produces another check which runs after the main PR build and uses the resources it created. Is that possible or are the resources never saved anywhere. Or maybe they re-build only the bits they need. 

The reason I say so is the addition of labels doesn't clearly indicate that the build changes configuration nor is there any visible output that it does. The PR workflow is also complicated a bit, we could create a new label workflow to run only the bits we need.

Thoughts?

snazy:
Sharing resources (like "things on disk") isn't possible between jobs.

Updated this change to not run CI when labels have changed, but when someone adds a comment containing the string `test this please`. There's no other (useful) way to trigger CI again.
(Rerunning a job in the workflow-UI isn't an option, because that uses the original event-json.)

The next step (not in this PR) would be to unify main.yml + pull-request.yml + site.yml so we don't have any more real duplicate code fragments in the workflows for CI.

snazy:
Ah no - strike that comment-thing... it becomes too [s|b]ad
Reverted this PR to re-run CI when the labels change - that way also only us can re-trigger CI

nastra (approved):
seems like it's working, so +1

.github/workflows/pull-request.yml
```diff
@@ -83,10 +101,22 @@ jobs:
...
+        echo ::set-output name=CI_WITH_JACKSON::$(jq 'any(.pull_request.labels[]; .name == "pr-jackson")' < $GITHUB_EVENT_PATH)
     - name: Build with Maven
-      run: ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml -Pcode-coverage,jdk8-tests -Dtest.log.level=WARN
       env:
+        CI_WITH_NATIVE: ${{ steps.ci_options.outputs.CI_WITH_NATIVE }}
+        CI_WITH_JDK8: ${{ steps.ci_options.outputs.CI_WITH_JDK8 }}
         SPARK_LOCAL_IP: localhost
+      run: |
+        PROFILES="-Pcode-coverage"
+        [[ $CI_WITH_JDK8 == "true" ]] && PROFILES="$PROFILES,jdk8-tests"
```
rymurr, line 117: can these be split up into different runs? They would make the build easier to read and we would see that jobs are actually run. It might prevent having to re-run the whole build for a single label change. I would rather be able to view when they have run and their results/logs. Especially for new users.

rymurr (changes requested):

.github/workflows/pull-request.yml
snazy: Moved those to separate `run`s

.github/workflows/pull-request.yml
```diff
@@ -84,9 +109,26 @@ jobs:
         key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
         restore-keys: ${{ runner.os }}-gradle
     - name: Build with Maven
-      run: ./mvnw -B install javadoc:javadoc-no-fork --file pom.xml -Pcode-coverage,jdk8-tests -Dtest.log.level=WARN
       env:
         SPARK_LOCAL_IP: localhost
+      run: |
```
rymurr, line 114: I just realised what jdk8 tests does. Isn't this the only way that hive tests are run? I think we have to keep this in CI. And anyways, I thikn we should remove `hmsbridge` modules

snazy:
WF-Run w/ all three labels: https://github.com/projectnessie/nessie/pull/1660/checks?check_run_id=3189803846

rymurr:
> WF-Run w/ all three labels: https://github.com/projectnessie/nessie/pull/1660/checks?check_run_id=3189803846

there are now 24 checks running?

rymurr:
There are also still references to jdk8 tests. Can that part of the PR be removed?

.github/workflows/pull-request.yml
laurentgo: I am of the same opinion as @rymurr : if someone modify the hive code (or any code the hive code depends on), we should not force people to figure out which tag to add to have an accurate CI picture

.github/workflows/pull-request.yml
```diff
@@ -2,6 +2,7 @@ name: PR Build Check
 
 on:
   pull_request:
+    types: [labeled, unlabeled, opened, synchronize, reopened]
```
laurentgo, line 5: wouldn't this cause also for the main job to run too simply by adding labels?

snazy: Nope, that's only for PRs

rymurr:
this broadly looks good to me. Just two questions:
1) why is the PR still reporting 16 checks? Looks like the whole build ran twice
2) how does the native build show up? I notice jackson shows up as a skipped job, I was kinda hoping the native test would show up the same

snazy:
> why is the PR still reporting 16 checks?

It says `9 successful and 1 skipped checks` here?

> how does the native build show up?

It doesn't. Better: it cannot w/o any more code, because only whole jobs appear as "checks" - not individual steps.

rymurr:
> It says `9 successful and 1 skipped checks` here?

Actions tab says 16 and it looks like the whole job ran twice

> It doesn't. Better: it cannot w/o any more code, because only whole jobs appear as "checks" - not individual steps.

so to make it show up as a check we would have to do most of the build again. Not great. would rather see that the job ran. But ok.

 can you add some documetation (possibly to CONTRIBUTING) talking about the labels and thier effects. Specifically mentioning that the native job is 'hidden'

snazy:
> Actions tab says 16 and it looks like the whole job ran twice

Probably true - twice against the same commit.

snazy:
> so to make it show up as a check we would have to do most of the build again

Not really. Will see if we can make it visible with some GH API calls like what's been there for the gatling-reports.

snazy:
> Will see if we can make it visible with some GH API calls

bummer... :(

rymurr:
> bummer... :(

What does this mean? 

> Probably true - twice against the same commit.

Sorry, why would it do that?

snazy:
> > bummer... :(
> 
> What does this mean?

It became too complicated - either the gh-wf for pull-request has no access to the secrets (to run the `gh` api calls) - or it becomes quite messy & complicated with 2nd-ary workflows.

> > Probably true - twice against the same commit.
> 
> Sorry, why would it do that?

CI runs again when you change the labels.

rymurr (approved):

PR #1671, opened 2021-07-29
Remove hmsbridge

snazy:
HMS-bridge is currently not well maintained and requires quite a bit more work.
We're focusing on Apache Iceberg and Delta, so the idea here is to just remove the whole HMS-bridge code - at least for now.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1671)


nastra (approved):
removing code is always fun, LGTM

rymurr (approved):

PR #1677, opened 2021-07-30
Create Helm chart for Nessie

nastra:
This is just a fairly simple MVP version so that Nessie can be installed via Helm.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1677)

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -1,0 +1,64 @@
+# Nessie Helm chart
+
+## Installation
+```bash
+$ helm install --namespace nessie-ns --name nessie helm/nessie
+```
+
+## Uninstalling the Chart
+
+```bash
+$ helm uninstall --namespace nessie-ns nessie
+```
+
+## Configuration
+### Nessie Configuration Parameters
+The following table lists the configurable parameters of the Nessie chart and their default values.
+
+| Parameter  | Description | Default |
+| -----------| ----------- | ------- |
+| replicaCount | The number of replicas to run | `1` |
+| image | Nessie Image settings | |
+| image.repository | The nessie image to use | `projectnessie/nessie` |
+| image.pullPolicy | Image pull policy, such as `IfNotPresent` / `Always` / `Never` | `IfNotPresent` |
+| image.tag | Overrides the image tag whose default is the chart appVersion | `appVersion` from `Chart.yaml` |
+| versionStoreType | Version store to use: `JGIT` / `INMEMORY` / `DYNAMO`. `JGIT` is best for local testing, `DYNAMO` preferred for production | `INMEMORY` |
+| dynamodb | Configuration specific to `versionStoreType: DYNAMO` | |
+| dynamodb.region | The region to configure for Dynamo | `us-west-2` |
+| dynamodb.secret.name | The name of the secret where credentials are stored | `awscreds` |
+| dynamodb.secret.awsAccessKeyId | The AWS access key id in the secret | `aws_access_key_id` |
+| dynamodb.secret.awsSecretAccessKey | The AWS secret access key | `aws_secret_access_key` |
+| jaegerTracing | Jaeger Tracing configuration | |
+| jaegerTracing.enabled | Determines whether Jaeger Tracing for Nessie is enabled or not via `true` / `false` | `false` |
+| jaegerTracing.endpoint | The traces endpoint, in case the client should connect directly to the Collector, such as `http://jaeger-collector:14268/api/traces` | None |
+| jaegerTracing.serviceName | The Jaeger service name | `nessie` |
+| jaegerTracing.publishMetrics | Determines whether metrics should be published or not via `true` / `false` | `true` |
+| jaegerTracing.samplerType | The sampler type (`const`, `probabilistic`, `ratelimiting` or `remote`) | `ratelimiting` |
+| jaegerTracing.samplerParam | 1=Sample all requests. Set `samplerParam` to somewhere between 0 and 1, e.g. 0.50, if you do not wish to sample all requests | `1` |
+| serviceMonitor | Configuration specific to a `ServiceMonitor` for the Prometheus Operator | |
+| serviceMonitor.enabled | Specifies whether a `ServiceMonitor` for Prometheus operator should be created | `true` |
+| serviceMonitor.interval | Interval at which metrics should be scraped | `30s` |
+| serviceMonitor.labels | Additional labels to add to the created `ServiceMonitor` so that Prometheus can properly pick it up | None |
+
+### Providing secrets for Dynamo Version Store
+
+* Make sure you have a Secret in the following form:
+```
+> cat $PWD/awscreds
+aws_access_key_id=YOURACCESSKEYDATA
+aws_secret_access_key=YOURSECRETKEYDATA
+```
+
+* Create the secret from the given file
+`kubectl create secret generic awscreds --from-env-file="$PWD/awscreds"`
+
+* Now you can use `DYNAMO` as the version store when installing Nessie via `helm install -n nessie-ns nessie helm/nessie --set versionStoreType=DYNAMO`.
+
+## Dev installation
+
+* Install Minikube as described in https://minikube.sigs.k8s.io/docs/start/
+* Install Helm as described in https://helm.sh/docs/intro/install/ 
+* Start Minikube cluster: `minikube start`
+* Create K8s Namespace: `kubectl create namespace nessie-dev`
+* Install Nessie Helm chart: `helm install nessie -n nessie-dev helm/nessie`
+
--- a/helm/README.md
+++ b/helm/README.md
@@ -25,9 +25,6 @@ The following table lists the configurable parameters of the Nessie chart and th
 | versionStoreType | Version store to use: `JGIT` / `INMEMORY` / `DYNAMO`. `JGIT` is best for local testing, `DYNAMO` preferred for production | `INMEMORY` |
 | dynamodb | Configuration specific to `versionStoreType: DYNAMO` | |
 | dynamodb.region | The region to configure for Dynamo | `us-west-2` |
-| dynamodb.secret.name | The name of the secret where credentials are stored | `awscreds` |
-| dynamodb.secret.awsAccessKeyId | The AWS access key id in the secret | `aws_access_key_id` |
-| dynamodb.secret.awsSecretAccessKey | The AWS secret access key | `aws_secret_access_key` |
 | jaegerTracing | Jaeger Tracing configuration | |
 | jaegerTracing.enabled | Determines whether Jaeger Tracing for Nessie is enabled or not via `true` / `false` | `false` |
 | jaegerTracing.endpoint | The traces endpoint, in case the client should connect directly to the Collector, such as `http://jaeger-collector:14268/api/traces` | None |
@@ -40,19 +37,6 @@ The following table lists the configurable parameters of the Nessie chart and th
 | serviceMonitor.interval | Interval at which metrics should be scraped | `30s` |
 | serviceMonitor.labels | Additional labels to add to the created `ServiceMonitor` so that Prometheus can properly pick it up | None |
 
-### Providing secrets for Dynamo Version Store
-
-* Make sure you have a Secret in the following form:
-```
-> cat $PWD/awscreds
-aws_access_key_id=YOURACCESSKEYDATA
-aws_secret_access_key=YOURSECRETKEYDATA
-```
-
-* Create the secret from the given file
-`kubectl create secret generic awscreds --from-env-file="$PWD/awscreds"`
-
-* Now you can use `DYNAMO` as the version store when installing Nessie via `helm install -n nessie-ns nessie helm/nessie --set versionStoreType=DYNAMO`.
 
 ## Dev installation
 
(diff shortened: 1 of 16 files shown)


servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -71,6 +71,12 @@ quarkus.resteasy.gzip.enabled=true
 quarkus.http.auth.basic=false
 quarkus.oidc.enabled=false
 
+## quarkus kubernetes settings (https://quarkus.io/guides/deploying-to-kubernetes)
+quarkus.kubernetes.name=nessie-server
+quarkus.kubernetes.version=0.8.3
```
snazy, line 76: Can you add some automation in `release-create.yml` GH workflow to update this?

nastra: I removed this as this isn't really necessary to have in the config

helm/nessie/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: 0.8.3
```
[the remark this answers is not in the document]
nastra: done

helm/README.md
```diff
@@ -0,0 +1,15 @@
+# Nessie Helm chart
+
+## Installation
```
rymurr, line 3: can this be filled out to describe the `Values.yaml` settings and what options a user has at install time?

helm/nessie/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
rymurr, line 1: What is this for? Can it just go in the `README.md`?

helm/nessie/templates/deployment.yaml
```diff
@@ -0,0 +1,88 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: NESSIE_VERSION_STORE_TYPE
+              value: {{ .Values.versionStoreType }}
+            {{- if .Values.dynamodb.region }}
+            - name: AWS_REGION
+              value: {{ .Values.dynamodb.region }}
+            {{- end }}
+            {{- if .Values.dynamodb.awsAccessKeyId }}
```
rymurr, line 43: typically this will be put in secrets rather directly in the values file. I don't think we want to encourage bad security practices here.

helm/nessie/values.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+service:
```
rymurr, line 44: An example of configuring this with a real ingest would be good for the docs

helm/nessie/values.yaml
```diff
@@ -0,0 +1,79 @@
...
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name: ""
+
+podAnnotations:
```
rymurr, line 27: rather than using the scrape annotation I think we should set up service monitors: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md

see: https://github.com/prometheus-operator/kube-prometheus/pull/16

helm/nessie/values.yaml
```diff
@@ -0,0 +1,79 @@
+replicaCount: 1
```
rymurr, line 1: there is no config to set up tracing here?

servers/quarkus-server/pom.xml
```diff
@@ -165,6 +165,14 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-container-image-jib</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-kubernetes</artifactId>
+    </dependency>
+    <dependency>
```
rymurr, line 172: why would minikube be a dependency?

servers/quarkus-server/pom.xml
```diff
@@ -165,6 +165,14 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-container-image-jib</artifactId>
     </dependency>
+    <dependency>
```
rymurr, line 168: what does this dependency provide? Do we really want to ship all servers w/ this?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -71,6 +71,11 @@ quarkus.resteasy.gzip.enabled=true
 quarkus.http.auth.basic=false
 quarkus.oidc.enabled=false
 
+## quarkus kubernetes settings (https://quarkus.io/guides/deploying-to-kubernetes)
```
rymurr, line 74: do these things ahve to live here? The server, up until now, is fairly agnostic to how we run right?

rymurr (changes requested):

helm/nessie/templates/NOTES.txt
nastra: The purpose of this is file is to show something after deploying the Helm chart (and in this case it shows how to do port forwarding). `NOTES.txt` is going through the templating engine and will show a copy/pastable string to do the port forwarding, thus it can't go into a `README.md`

servers/quarkus-server/pom.xml
nastra: it generate Kubernetes manifests (minikube.yaml and minikube.json) that are tailored to Minikube during the build

servers/quarkus-server/pom.xml
nastra: the purpose of this is to **only** generate kubernetes manifest files during the build

servers/quarkus-server/src/main/resources/application.properties
nastra: technically speaking, we don't need to have any of those k8s-specific settings here. I'll remove them

helm/nessie/values.yaml
nastra: the idea was to create an MVP version to be able to install Nessie via Helm and later add/parameterize what we need

helm/README.md
nastra: yep will add more docs around configurable things

helm/nessie/values.yaml
nastra: TIL about servicemonitor, will add that

helm/nessie/values.yaml
nastra: does that sound reasonable @rymurr?

helm/nessie/templates/deployment.yaml
nastra: yep agreed

helm/nessie/values.yaml
rymurr: tracing should be a 2 liner and is part of the MVP imho. it should be easy for a user to:
* start nessie 
* secure it
* attach it to a valid ingress
* observe it
We've done everything but the tracing part of observing

servers/quarkus-server/pom.xml
rymurr: do they get packaged into the jar file or docker image or just left in `target`?

servers/quarkus-server/pom.xml
rymurr: So w/o this we can't deploy to minikube?  do they get packaged into the jar file or docker image or just left in `target`?

servers/quarkus-server/pom.xml
nastra: they will just stay in `target` and are not making it into the jar

servers/quarkus-server/pom.xml
nastra: correct, we can't deploy the generated kubernetes manifest with minikube, since that plugin adds some stuff necessary for it to work with minikube (which lands under `target/kubernetes/minikube.yml`), but it doesn't add anything to the final `jar` and everything stays under `target`

helm/nessie/values.yaml
nastra: done

servers/quarkus-server/pom.xml
```diff
@@ -165,6 +165,14 @@
...
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-kubernetes</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-minikube</artifactId>
+    </dependency>
```
rymurr, line 175: can these go in a different module? similar to how we do the lambda? Also, why are they needed at all if the helm chart is already available?

nastra: strictly speaking, this is independent from the Helm chart, but this is the fastest/cheapest way to generate a functioning kubernetes manifest for nessie without having to look through K8s APIs. I'm basically using the generated manifest and copy/paste it into the Helm chart stuff where some values are then templated.

rymurr: so in theory it could be behind a profile or something? It really only needs to be run when re-generating the helm templates?

helm/README.md
nastra: done

helm/nessie/values.yaml
nastra: added tracing stuff

servers/quarkus-server/pom.xml
nastra: I put it behind a `k8s` profile

> It really only needs to be run when re-generating the helm templates?

it doesn't need to be run at all. everything can be done manually, but the generated k8s manifests are quite helpful as references

helm/nessie/values.yaml
nastra: I'll try and add a script that generates some data/commits and update the docs in a separate PR

helm/nessie/templates/deployment.yaml
nastra: fixed

servers/quarkus-server/pom.xml
```diff
@@ -327,5 +327,20 @@
         <quarkus.container-image.build>true</quarkus.container-image.build>
       </properties>
     </profile>
+    <profile>
+      <!-- Use the k8s profile to generate manifest files for K8s & Minikube, which will be located under target/kubernetes -->
```
snazy, line 331: Wonder whether we can add a jar-artifact to the deployment (e.g. `nessie-quarkus-1.2.3-k8s.jar`) that contains the generated manifest files. Would that be useful for user?

nastra: atm I don't think we need to ship these k8s artifacts at all. The way to install Nessie would be only via the Helm chart. If we decide that we want to ship those k8s artifacts at some point, then we would need to adjust a bunch of k8s settings in ´application.properties` but I don't see this in the scope of this PR.

helm/nessie/templates/deployment.yaml
```diff
@@ -0,0 +1,114 @@
...
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: NESSIE_VERSION_STORE_TYPE
+              value: {{ .Values.versionStoreType }}
+            {{- if eq .Values.versionStoreType "DYNAMO" }}
+            {{- if .Values.dynamodb.region }}
+            - name: AWS_REGION
+              value: {{ .Values.dynamodb.region }}
+            {{- end }}
+            - name: AWS_ACCESS_KEY_ID
```
rymurr, line 44: will these environment variables show up in logs anywhere? Or in a `docker ps` type command? I believe the image is set up to read from all the standard locations for AWS config so we could also write to `~/.aws/credentials` if these env variables would be visible anywhere.

helm/nessie/values.yaml
rymurr: by ingest I meant `nginx` and possibly ssl as opposed to ingesting data w/ nessie

helm/nessie/templates/deployment.yaml
nastra: I'm not aware that those env vars would be visible. They will only show up on the Deployment/Pod like this:
```
    Environment:
      NESSIE_VERSION_STORE_TYPE:  DYNAMO
      AWS_REGION:                 us-west-2
      AWS_ACCESS_KEY_ID:          <set to the key 'aws_access_key_id' in secret 'awscreds'>      Optional: false
      AWS_SECRET_ACCESS_KEY:      <set to the key 'aws_secret_access_key' in secret 'awscreds'>  Optional: false
```

nastra: actually inside the POD the secret will be exposed as a normal env var as described in https://kubernetes.io/docs/concepts/configuration/secret/#consuming-secret-values-from-environment-variables

rymurr (approved):

PR #1711, opened 2021-08-06
Fix scala and antlr shading

rymurr:
In this PR:

* downgrade antlr4 version to match spark
* rename classes to prevent clashing w/ Iceberg
* move grammar to its own module and relocate in a shaded jar

The relocation on the separate module is because maven shade does not support relocating scala classes. So we relocate the java stuff in a different module and import that into the scala module.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1711)


pom.xml
```diff
@@ -125,7 +125,7 @@
     <!--<dynamodb-local.repository.url>https://s3.us-west-2.amazonaws.com/dynamodb-local/release</dynamodb-local.repository.url>-->
 
     <!-- Version properties -->
-    <antlr.version>4.9.2</antlr.version>
+    <antlr.version>4.8</antlr.version>
```
nastra, line 128: nit: Iceberg uses 4.7.1

nastra (dismissed):

pom.xml
rymurr: yeah, i wanted to keep to spark3 rather than iceberg. I suspect iceberg probably wants to bump also.

nastra (dismissed):

snazy (approved):

nastra (approved):

nastra (approved):

PR #1712, opened 2021-08-06
TVS2 / DatabaseAdapter interface

snazy:
Adds the base `DatabaseAdapter` interface that contains all methods + functions to act as a backend for a `VersionStore` implementation.

`DatabaseAdapterFactory` contains functionality to find the requested database-adapter, configure and instantiate it.

The `org.projectnessie.versioned.persist.adapter` package contains the "version store facing" interfaces.

The `org.projectnessie.versioned.persist.adapter.spi` package contains "database-adapter implementation facing" code. Note: `AbstractDatabaseAdapter` has all the logic for commit/transplant/merge/etc.

Concrete database-adapter implementations fall into one category: non-transactional (think: NoSQL), transactional (think: JDBC).

Unit-test and benchmark as well as functionality for non-tx + tx + database-adapter implementations and a matching `VersionStore` implementations will come in follow-up PRs.

All code based on the `DatabaseAdapter` API/SPI is capable of tracking both global and per-reference state for contents objects. Contents objects are opaque to the database-adapter code.

Note for the curious people: [This TVS2 branch](https://github.com/snazy/nessie/tree/tvs2-with-legacy) contains everything, including the current version-store implementations. [This branch](https://github.com/snazy/nessie/tree/tvs2-with-legacy-with-deletes) builds on the former branch, but has the current version-store implementations removed.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1712)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
```diff
@@ -0,0 +1,89 @@
...
+  /** Creation timestamp in microseconds since epoch. */
+  long getCreatedTime();
+
+  Hash getHash();
+
+  /** Zero, one or more parent-entry hashes of this commit, nearest parent first. */
+  List<Hash> getParents();
+
+  /** Serialized commit-metadata. */
+  ByteString getMetadata();
```
nastra, line 41: are we using protobuf types here on purpose?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,975 @@
...
+                    done = true;
+                  }
+                });
+          }
+        };
+    return StreamSupport.stream(split, false).onClose(s::close);
+  }
+
+  /**
+   * Ensures that {@code ref} exists and that the hash in {@code hashOnRef} exists in that
```
nastra, line 404: maybe `hashExistsOnRef`?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+  Stream<WithHash<NamedRef>> namedRefs();
+
+  Hash create(NamedRef ref, Optional<NamedRef> target, Optional<Hash> targetHash)
+      throws ReferenceNotFoundException, ReferenceAlreadyExistsException;
+
+  void delete(NamedRef ref, Optional<Hash> hash)
+      throws ReferenceNotFoundException, ReferenceConflictException;
+
+  void assign(
+      NamedRef ref, Optional<Hash> expectedHash, NamedRef assignTo, Optional<Hash> assignToHash)
```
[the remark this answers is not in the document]
rymurr: is the reason to pass both down authz?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>See {@link org.projectnessie.versioned.persist.adapter.spi.DefaultDatabaseAdapterConfig} for
+ * the default, immutable-annotated type.
+ *
+ * <p>{@link DatabaseAdapter} implementations that need more configuration options must extend this
+ * interface, have the {@link Value.Immutable} annotation and declare the {@code with...} methods
+ * implemented by "immutables".
+ */
+public interface DatabaseAdapterConfig {
+
+  int DEFAULT_PARENTS_PER_COMMIT = 20;
```
rymurr, line 32: nit the `int`s are constants and the `String`s arent. Would be nice to do one or the other

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/SystemPropertiesConfigurer.java
```diff
@@ -0,0 +1,78 @@
...
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class to configure instances of {@link DatabaseAdapterConfig} via system properties for
+ * micro-benchmarks and other rather one-off things. Do <em>not</em> use this class in production
```
rymurr, line 27: can it live in the `test` module then? in my experience is if its in `main` you can assume someone will use it

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /** Builds a {@link ReferenceNotFoundException} exception with a human-readable message. */
+  protected static ReferenceNotFoundException hashNotFound(NamedRef ref, Hash hash) {
+    return new ReferenceNotFoundException(
+        String.format(
+            "Could not find commit '%s' in reference '%s'.", hash.asString(), ref.getName()));
+  }
+
+  protected static ReferenceNotFoundException refNotFound(NamedRef ref) {
```
rymurr, line 61: nit `referenceNotFound`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
```diff
@@ -0,0 +1,145 @@
...
+    return new ReferenceAlreadyExistsException(
+        String.format("Named reference '%s' already exists.", ref));
+  }
+
+  /** Returns the microseconds since epoch. */
+  protected static long currentTimeInMicros() {
+    // We only have System.currentTimeMillis() as the current wall-clock-value, but want
+    // microsecond "precision" here, which is implemented by taking the microsecond-part from
+    // System.nanoTime().
+    long microsFraction = TimeUnit.NANOSECONDS.toMicros(System.nanoTime());
```
rymurr, line 76: we can't rely on nanotime being fixed to a second so this is going to give us some microseconds since some completely arbitrary time. Why would we need such precision? If we do I would rather introduce a `Clock` and let more sophisticated clocks get injected. Finally can we get the time from `java.time` objects and make sure its in UTC

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
```diff
@@ -0,0 +1,145 @@
...
+import org.projectnessie.versioned.ReferenceConflictException;
+import org.projectnessie.versioned.ReferenceNotFoundException;
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapter;
+
+/**
+ * Utility methods for {@link DatabaseAdapter} implementations.
+ *
+ * <p>This class is just there to keep {@link AbstractDatabaseAdapter} cleaner and shorter.
+ */
+public abstract class DatabaseAdapterUtil implements DatabaseAdapter {
```
rymurr, line 38: I thikn this would be better as a `final class` rather than an `abstract class`. I would rather avoid interitance for such utils

versioned/pom.xml
```diff
@@ -37,5 +37,7 @@
     <module>tests</module>
     <module>tiered</module>
     <module>gc</module>
+
```
rymurr, line 40: nit extra line

rymurr (commented):
A few comments and nits.

My main concerns are:
1) are we pushing authz into the database layer
2) there are a lot of container objects. Do we end up doing a lot of marshalling between these objects and is there any way to simplify the number of container objects.
3) we are using `ByteString` i assume that means the layer above deals w/ serialization. I am not sure its clear that the database adapter cant modify the `ByteString`. Would we have to pass around as many objects and bytestrings if we pushed the serialization helper down to this layer?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
snazy: fixed

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/SystemPropertiesConfigurer.java
snazy: moved to the nessie-versioned-persist-tests artifact (not in this PR)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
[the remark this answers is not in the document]
snazy: Introduced a composite that holds the named-reference and the optional hash - makes a bunch of code much easier to read.

The `DatabaseAdapter`s do not do any authz. What those do however is to verify that the hash, if present, is reachable via the named-reference. With this, authz only needs to verify against the name of the reference. Without checking that a commit hash is reachable via a named reference, you could easily bypass authz. That "is commit on named-ref" check is a quite cheap operation for database-adapter-implementations, especially for commit/merge/transplant, which have to walk the commit-log anyway.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Tiered-Version-Store Database-adapter interface.
+ *
+ * <p>All returned {@link Stream}s must be closed.
+ */
+public interface DatabaseAdapter extends AutoCloseable {
+
+  /** Ensures that mandatory data is present in the repository, does not change an existing repo. */
+  void initializeRepo() throws ReferenceConflictException;
```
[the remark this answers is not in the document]
snazy: it wouldn't, removed

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+   * @param commitAttempt parameters for the commit
+   * @return optimistically written commit-log-entry
+   */
+  Hash commit(CommitAttempt commitAttempt)
+      throws ReferenceConflictException, ReferenceNotFoundException;
+
+  Hash transplant(
+      BranchName targetBranch,
+      Optional<Hash> expectedHash,
+      NamedRef source,
```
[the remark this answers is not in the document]
snazy: See https://github.com/projectnessie/nessie/pull/1712#discussion_r686720030

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+import org.projectnessie.versioned.ReferenceConflictException;
+import org.projectnessie.versioned.ReferenceNotFoundException;
+import org.projectnessie.versioned.WithHash;
+
+/**
+ * Tiered-Version-Store Database-adapter interface.
+ *
+ * <p>All returned {@link Stream}s must be closed.
+ */
+public interface DatabaseAdapter extends AutoCloseable {
```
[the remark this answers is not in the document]
snazy: added a bunch of javadoc

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -0,0 +1,133 @@
...
+
+  Hash merge(
+      NamedRef from, Optional<Hash> fromHash, BranchName toBranch, Optional<Hash> expectedHash)
+      throws ReferenceNotFoundException, ReferenceConflictException;
+
+  Hash toHash(NamedRef ref) throws ReferenceNotFoundException;
+
+  Stream<WithHash<NamedRef>> namedRefs();
+
+  Hash create(NamedRef ref, Optional<NamedRef> target, Optional<Hash> targetHash)
```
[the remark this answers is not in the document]
snazy: Introduced a composite that holds the named-reference and the optional hash - makes a bunch of code much easier to read.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
snazy: Had to move the "default branch name" out of `DatabaseAdapter`, so that particular option is gone here.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
snazy: done

snazy:
> 1. are we pushing authz into the database layer

Not really, just "natively" implement the check that a commit-hash is reachable from a named-reference. This is an operation that's performed anyways, so it's quite cheap to have it at that level.

> 2. there are a lot of container objects. Do we end up doing a lot of marshalling between these objects and is there any way to simplify the number of container objects.

There are:
* `CommitAttempt` encapsulating the parameters for the commit-operation. This is actually quite nice, because with that the `PersistVersionStore` can iteratively construct the parameters - in other words: the "ímmutables" library does all the temporary collections handling.
* `CommitLogEntry` contains all attributes that make up a commit. (see below about secondary indexes / predicate pushdown)
* `CommitOnReference` groups named-reference + optional-commit-hash - it is actually a simplification
* `ContentsAndState` could be refactored a bit, with the follow-up mentioned here https://github.com/projectnessie/nessie/pull/1712#discussion_r686715615

Leaves `KeyWithBytes` + `KeyWithType` to have the tuple key+contents-id+type - the difference is that `KeyWithBytes` also contains the value.

And `ContentsIdAndBytes` + `ContentsIdWithType` for the tuple contents-id+type. These are only used when exclusively dealing with global-state during Nessie-GC (global state doesn't know about the contents-key, which can change).

> 3. we are using `ByteString` i assume that means the layer above deals w/ serialization. I am not sure its clear that the database adapter cant modify the `ByteString`. Would we have to pass around as many objects and bytestrings if we pushed the serialization helper down to this layer?

Right, the `PersistVersionStore` (not in this PR) deals w/ (de)serialization via `StoreWorker` and `DatabaseAdapter` handles "opaque BLOBs". I tried to refactor it so that the database-adapters get the actual contents and commit-metadata objects, but that turned out to be quite complex w/o a refactoring of the store-worker mechanism: There are actual two reasons to tackle that refactoring as a follow-up. But for that, the database-adapters would need to have some "knowledge" of the currently opaque data.
1. Adding the "global-state" + "contents-id" into the mix makes the `StringSerializer` (for tests) and its `StoreWorker` implementation a bit messy.
2. Having to deal w/ (de)serialization at the version-store level and also for some database-adapters (e.g. Rocks) feels too much work. It also prevents adding secondary indexes (think: pushing down predicates to the database) later.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DatabaseAdapterUtil.java
snazy: Changed to `Clock`, but we do not need the microsecond part to be accurate.
Maybe we do not even need microsecond precision here (just took what's been used for that "dt" field in the current TVS).
All we basically need is some wall clock that tells us, whether a commit is "too recent" and "eligible" for Nessie-GC.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,936 @@
...
+          hasher.putString(e.getContentsId(), StandardCharsets.UTF_8);
+          hasher.putBytes(e.getValue().asReadOnlyByteBuffer());
+        });
+    unchanged.forEach(e -> hashKey(hasher, e));
+    deletes.forEach(e -> hashKey(hasher, e));
+    return Hash.of(UnsafeByteOperations.unsafeWrap(hasher.hash().asBytes()));
+  }
+
+  /** Adds a complete key-list to the given {@link CommitLogEntry}, will read from the database. */
+  protected CommitLogEntry buildKeyList(OP_CONTEXT ctx, CommitLogEntry unwrittenEntry) {
```
rymurr, line 557: How is the Key list stored/modelled? I am not 100% clear on that. It was effectively a linked-list in the original impl is it the same? The original impl also ensured that the list was semi-sorted

1. It can't only live on the commit right? 
2. Is it a separate object in the database?
3. Does every op have to read the entire list?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,936 @@
...
+          return entry.getParents();
+        });
+  }
+
+  /**
+   * Constructs a {@link Stream} of entries for either the global-state-log or a commit-log. Use
+   * {@link #commitLogFetcher(Object, Hash)} or the similar implementation for the global-log for
+   * non-transactional adapters.
+   */
+  protected <T> Stream<T> logFetcher(
```
rymurr, line 455: the javadoc here implies that the global-state has a history? Should it? Specifically to the iceberg case the global state of  a table becomes invalid once new global state is written. Just looking for clarification and to ensure that we can't time travel on teh global state

rymurr (commented):
Looks good, just asked for a few clarifications.

and:
1) regarding [this comment](https://github.com/projectnessie/nessie/pull/1712#discussion_r686720030) I think authz is leaking into the database impl. My understanding is that if a database adapter doesn't perform that check it implicitly could bypass authz. So authz concepts have leaked into the database impl. It puts a lot of responsibility on the database adapter implementer and is prone to being missed. I think the authz checks have to be done completely at a higher level and not rely on the database adapter to implicitly do it. This may slow down the commit rate but it keeps interfaces clean (ie its a premature optimisation imho)

2) I am not sure how to do this but I think the contents id is kinda a first class citizen now. Its the key in the global state object and has to be on all objects we persist. I don't think we have captured it well currently. Maybe it should be a `Key` also, or a subclass `GlobalKey` to help w/ type safety and avoid confusion. I am up for suggestions but I think an unattached String floating around the API feels squishy.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,936 @@
...
+  protected AbstractDatabaseAdapter(CONFIG config) {
+    Objects.requireNonNull(config, "config parameter must not be null");
+    this.config = config;
+  }
+
+  @Override
+  public Stream<KeyWithBytes> entries(
+      CommitOnReference commitOnReference, KeyFilterPredicate keyFilter)
+      throws ReferenceNotFoundException {
+    throw new UnsupportedOperationException("entries() is not used anywhere");
```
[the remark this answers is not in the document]
snazy: Argh - don't recall why I added that function. It's useless. Removed.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,936 @@
...
+   */
+  protected abstract Map<String, ByteString> fetchGlobalStates(
+      OP_CONTEXT ctx, Set<String> contentIds);
+
+  /**
+   * Write a new commit-entry with a best-effort approach to prevent hash-collisions but without any
+   * other consistency checks/guarantees. Some implementations however can enforce strict
+   * consistency checks/guarantees.
+   */
+  protected abstract void writeIndividualCommit(OP_CONTEXT ctx, CommitLogEntry entry)
```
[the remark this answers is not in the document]
snazy: Yes, implementations have to rather "blindly" write the given entry. No value or consistency checks by the implementations. Implementations however _can optionally_ ensure that the given hash is not already present - but since it might be tricky or quite expensive for some databases, it's not mandatory.

Clarified that in the javadoc.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,936 @@
...
+      throws ReferenceConflictException;
+
+  /**
+   * Write multiple new commit-entries with a best-effort approach to prevent hash-collisions but
+   * without any other consistency checks/guarantees. Some implementations however can enforce
+   * strict consistency checks/guarantees.
+   */
+  protected abstract void writeIndividualCommits(OP_CONTEXT ctx, List<CommitLogEntry> entries)
```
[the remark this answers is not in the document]
snazy: done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
rymurr: Ok, maybe I am confused. Is this the update of the ref or just the write of the commit w/ hash? The commit write cna be blind but the ref swap can't. Did I just misunderstand where the ref swap happens?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: The (current) concept to deal with the problem of getting a "full key list quickly" is this:
* The full key list (think: call to `DatabaseAdapter.keys()` / `VersionStore.getKeys`) is constructed from all puts+deletes from all commits
* Since the above is quite expensive, the implementation writes a "full key list" with every N-th commit (`DatabaseAdapterConfig.getKeyListDistance()`, defaults to `20`).

The above (full-key-list within a commit-entry) runs into the same problem as the current TVS implementation (if it becomes too big, the underlying database "runs into trouble"). The plan is to tackle this problem in a follow-up by limiting the number of keys per embedded-key-list and put "overflows" into separate database objects/rows (the commented `EmbeddedKeyList.getKeyListsIds()`). Probably limit the `CommitLogEntry.embeddedKeyList` to maybe a few dozen entries and then overflow to separate objects/rows. Another idea is to manage both the `CommitLogEntry.embeddedKeyList` and its "overflows" in an LRU-ish way - i.e. most recently touched keys end in `CommitLogEntry.embeddedKeyList`, older ones get "spilled out" into the "overflows".

rymurr: I am not sure I want to make the re-implementation less stable than the current one. The current one only runs into trouble if the embedded key list is rather large as it stores pointers to sub-lists of keys. I dont remember the number but its quite a bit larger than teh largest warehouses we would expect, This impl is limited to 10s or 100s of keys before it runs into trouble.

The algorithm you suggest is similar to how the TVS works today (at least in my understanding). The most often changed tables will be near the 'top' of the list of key-lists and the least often changed will sink to the bottom.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Right. There's only one global-state for any contents-id.
For no-sql databases it is implemented as a log-structure for technical reasons to have just one CAS operation.
For sql databases it just a table with contents-id + type as the primary key.

snazy: Note: global-state is not exposed as a log or history or the ability for time-travel.

snazy:
> Looks good, just asked for a few clarifications.

Thanks! Updated the PR according to the comments. Also removed the `CommitOnReference` composite, because it's no longer useful. Once this is in, I'll post a few follow-up PRs - one for Non-transactional DBs, one for SQL/JDBC DBs and maybe another one with common code for these two, if those become too big.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: Yea - this method just blindly writes the commit. The actual swap-operation depends on the database, so it's not in this PR.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: > This impl is limited to 10s or 100s of keys before it runs into trouble.

It's limited to the max size of a "db object" (-> e.g. the 400k limit in dynamo).

rymurr: agreed. which is orders of magnitude smaller than the current impl and for large table namespaces or delta tables that could be in the 100s of table names

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -0,0 +1,935 @@
...
+   * contain any of the given {@link Key}s.
+   *
+   * <p>Conflicts are reported via {@code mismatches}.
+   */
+  protected void checkConflictingKeysForCommit(
+      OP_CONTEXT ctx,
+      Hash upToCommitIncluding,
+      Hash sinceCommitExcluding,
+      Set<Key> keys,
+      boolean[] sinceSeen,
```
rymurr, line 710: why pass a mutable `boolean[1]` instead of returning a `boolean`

nastra: my guess when reviewing the code was so that the value can be modified

rymurr: it appears tp be just a return value in the current usage. its modified in a lambda scope in the method but i don't see why the boolean array couldn't be declared in the method and just returning the element at the end of the method couldn't work. Maybe i am jus tnot reading it correclty?

snazy: That `sinceSeen` value is computed within a Stream-lambda, so it can't be simply returned. Updated the code a little.

rymurr (approved):

PR #1720, opened 2021-08-09
Manually publish Nessie Helm chart to https://charts.projectnessie.org 

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1720)

I ran into some issues with the `chart-releaser-action` where it
would try and only release charts that changed (helm/chart-releaser-action#71).
I tried adding some changes locally to the chart and then running the
same commands as the action does, but I couldn't get it to work
properly, therefore I decided to do the publishing manually via commit

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -231,13 +231,18 @@ jobs:
         mv nessie-${RELEASE_VERSION}.tgz nessie-helm-${RELEASE_VERSION}.tgz
         echo ::set-output name=NESSIE_HELM_CHART::nessie-helm-${RELEASE_VERSION}.tgz
 
+    # Publish Nessie Helm chart to Helm Repo
     - name: Publish Nessie Helm chart to Helm Repo
-      uses: helm/chart-releaser-action@v1.2.1
-      if: ${{ !env.ACT }}
-      with:
-        charts_dir: helm
       env:
-        CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        RELEASE_VERSION: ${{ steps.get_version.outputs.VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-publish.yml
```diff
@@ -251,6 +251,7 @@ jobs:
       if: ${{ !env.ACT }}
       with:
         charts_dir: helm
+        charts_url: https://projectnessie.org/charts/
```
snazy, line 254: Update the URLs in the release-notes as well?

laurentgo: But I can have charts.projectnessie.org point to it, I guess?

nastra: > But I can have charts.projectnessie.org point to it, I guess?
yep, I changed the URL here to also point to charts.projectnessie.org

rymurr (dismissed):

rymurr (approved):

.github/workflows/release-publish.yml
```diff
@@ -231,13 +231,18 @@ jobs:
         mv nessie-${RELEASE_VERSION}.tgz nessie-helm-${RELEASE_VERSION}.tgz
         echo ::set-output name=NESSIE_HELM_CHART::nessie-helm-${RELEASE_VERSION}.tgz
 
+      # Publish Nessie Helm chart to Helm Repo
```
snazy, line 234:
```suggestion
    # Publish Nessie Helm chart to Helm Repo
```
Indent should be fixed (for IDEs)

nastra: fixed

snazy (approved):

PR #1745, opened 2021-08-12
Spark SQL Extensions - don't fail on create

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1745)


nastra (approved):
LGTM, altough I would probably update the commit msg / PR title to `...Introduce IF NOT EXISTS when creating branches/tags` as this makes it a bit clearer and everyone understands what `IF NOT EXISTS` means

snazy (approved):

PR #1746, opened 2021-08-12
Move Swagger examples to openapi.yaml

### LINKED ISSUE #1610: Fix openapi examples
the 'Try it Now!' button doesn't work for every endpoint b/c the example object doesn't render correctly and gets translated to the wrong url

`contents/{key}` was particularly bad in my testing.

dimas-b:
* This is mostly necessary for proper quoting of complex JSON
  payload examples (they got extra escaping when defined as java
  strings in annotations).

* Also move the top-level API metadata to openapi.yaml to keep
  all package-level configuration in one place.

* Minor fixes to Swagger example objects.

* Bump the API version to current project version: 0.9.1

Note: most Swagger examples still lead to "not found" errors when
executed, but at least they show syntactically valid argument
values now and Swagger UI is able to execute them.

Fixes #1610

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1746)


rymurr (commented):
LGTM, thanks @dimas-b !

Lets quickly discuss on standup what has to be done to make the examples work better. I suspect its:

1) add more example objects w/ different entries - eg empty ones
2) try and make the hash a valid hash, but that is store dependent...

Thoughts @snazy and @nastra

nastra (approved):

nastra:
+1 on adding more examples with different entries (as a follow-up)

dimas-b:
> try and make the hash a valid hash, but that is store dependent

The hash of the `main` is stable, but we still do not have any commits by default.

PR #1754, opened 2021-08-13
Keep generated API client code in a separate dir

dimas-b:
* Isolate code produced by openapi-generator into
  ui/src/generated

* Keeping the generated code under src/ to simplify
  code management during build (sources are required
  to be under the same root or symlinked)

* Add fix-generated-client.js to apply post-generation
  corrections (not sure why openapi-generator does not
  produce the expected code)

* Exclude ui/src/generated from version control

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1754)


ui/package-lock.json
```diff
@@ -17853,6 +17854,182 @@
         "node": ">=0.10"
       }
     },
+    "node_modules/replace-in-file": {
```
dimas-b, line 17857: Do we need this change? :thinking:

rymurr (approved):

snazy (approved):

PR #1763, opened 2021-08-16
TVS2 allow nested key-lists

snazy:
The current implementation aggregates all known content-`Key`s in `CommitLogEntry.keyList`, so a larger amount of content-keys would hit a database's hard (and also highly recommended) limit of a database-object/database-row.

This PR enhances the "embedded key list" `CommitLogEntry.keyList` with "nested" `KeyListEntity`s, so the total number of keys is rather "unlimited".

When a "complete-key-list" is constructed, the total size of both `CommitLogEntry` and `KeyListEntity` is limited to `DatabaseAdapterConfig.getMaxKeyListSize`. The "global default" for this value is unlimited (`Integer.MAX_VALUE`). Databases that do force a hard/highly-recommended limit must override `DatabaseAdapterConfig.getDefaultMaxKeyListSize`, to imply that as the default for the concrete database.

Current tests (not in this PR) exercise against a rather small value for max-key-list-size and pass for the current implementations (not in this PR) for in-memory, rocks-db, h2 + postgres.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1763)

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
@@ -53,15 +53,25 @@ public interface CommitLogEntry {
   List<Key> getDeletes();
 
   /**
-   * List of all "reachable" or "known" {@link Key}s up to this commit-log-entry's <em>parent</em>
-   * commit.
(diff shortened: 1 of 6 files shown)


rymurr (approved):
As discussed offline we will revisit the keylist algorithm when we can fully perf test it. We will look at:

* # of keys
* distance to full-key-list
* contention 
* retries
And try ot determine if its better to keep as many  keys as possible in the commit object or in the spill lists

PR #1765, opened 2021-08-16
TVS2: Serialization for database-adapters

snazy:
Some `DatabaseAdapter` implementations, for example RocksDB, store certain entities in their serialized form.

Introduces a new artifact `nessie-versioned-persist-serialize` that contains protobuf based serialization.

Objects that are exposed via the `DatabaseAdapter` interface are still based on "immutables" and not protobuf.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1765)

### DIFF
--- a/versioned/persist/serialize/src/main/proto/persist.proto
+++ b/versioned/persist/serialize/src/main/proto/persist.proto
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+syntax = "proto3";
+package nessie.persist;
+
+option java_package = "org.projectnessie.versioned.persist.serialize";
+option java_outer_classname = "AdapterTypes";
+option java_generate_equals_and_hash = true;
+
+message CommitLogEntry {
+  int64 createdTime = 1;
+  bytes hash = 2;
+  repeated bytes parents = 3;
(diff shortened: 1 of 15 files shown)


nastra (approved):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DbObjectsSerializers.java
```diff
@@ -0,0 +1,273 @@
...
+    }
+    if (sb.length() > 0) {
+      b.addElements(sb.toString());
+      sb.setLength(0);
+    }
+    return b.build();
+  }
+
+  /**
+   * Flatten this {@link Key} into a single string.
```
rymurr, line 240: why deviate from the other parts of the code that serialize . as \00?

rymurr (changes requested):
why straight bytes over protobuf when protobuf is used as our serialization format everywhere else? How do we handle versioning, upgrades, changing of these objects. It seems like while protobuf will create more code for us here it will be more stable long term as this on disk format changes

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/DbObjectsSerializers.java
snazy: The thought here was that `\000` is maybe not a "nice" character in a `VARCHAR` column

snazy:
PR completely redone w/ protobuf.

rymurr (dismissed):

versioned/persist/serialize/src/main/proto/persist.proto
```diff
@@ -0,0 +1,94 @@
...
+message KeyList {
+  repeated KeyWithType keys = 1;
+}
+
+message ContentsId {
+  string id = 1;
+}
+
+message ContentsIdWithType {
+  ContentsId contents_id = 2;
```
nastra, line 62: any particular reason why this doesn't start at `1`? Same for `ContentsIdWithBytes`

nastra (dismissed):

versioned/persist/serialize/src/main/proto/persist.proto
snazy: ehm - no - let me fix that

nastra (approved):

PR #1767, opened 2021-08-16
WIP: Initial gRPC implementation - Dont review yet

nastra:


clients/client/src/main/java/org/projectnessie/client/HttpClientBuilder.java
```diff
@@ -0,0 +1,238 @@
...
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_URL;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_USERNAME;
+
+import java.net.URI;
+import java.util.function.Function;
+import org.projectnessie.client.http.HttpClientException;
+import org.projectnessie.client.http.HttpClientReadTimeoutException;
+import org.projectnessie.model.Validation;
+
+public class HttpClientBuilder {
```
nastra, line 33: nothing new in this class. It was just moved out of `NessieClient`

clients/spark-extensions/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
```diff
@@ -101,7 +101,7 @@ object NessieUtils {
     val catalogConf = SparkSession.active.sparkContext.conf
       .getAllWithPrefix(s"spark.sql.catalog.$catalogName.")
       .toMap
-    HttpClientBuilder
+    NessieClient
```
nastra: this is coming from the iceberg-spark3-runtime dependency and not from the `NessieClient` version that we currently have

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcExceptionMapper.java
```diff
@@ -0,0 +1,181 @@
...
+   *     matches a {@link NessieNotFoundException}
+   * @throws NessieConflictException If the callable threw a gRPC exception, where the status
+   *     matches a {@link NessieConflictException}
+   * @throws NessieBadRequestException If the callable threw a gRPC exception, where the status
+   *     matches a {@link NessieBadRequestException}
+   * @throws StatusRuntimeException If the underlying exception couldn't be converted to the
+   *     mentioned Nessie-specific exceptions, then a {@link StatusRuntimeException} with {@link
+   *     Status#UNKNOWN} is thrown.
+   */
+  public static <T> T handle(java.util.concurrent.Callable<T> callable)
```
snazy, line 100: Import for `Callabcle`

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcClientBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Builds a new {@link NessieGrpcClient}.
+   *
+   * @return A new {@link NessieGrpcClient}.
+   */
+  public NessieClient build() {
+    ManagedChannel channel = ManagedChannelBuilder.forTarget(endpoint).usePlaintext().build();
+    GrpcConfigApi configApi = new GrpcConfigApi(channel);
```
snazy, line 56: Nit: inline

model/src/main/java/org/projectnessie/api/ConfigApi.java
```diff
@@ -26,9 +25,7 @@
 import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
 import org.projectnessie.model.NessieConfiguration;
 
-@Path("config")
```
snazy, line 29: Why only move the REST and OpenAPI annotations from the interfaces `ConfigApi`, `ContentsApi` and `TreeApi` but not the REST and OpenAPI annotations from the methods/function + parameters?

grpc/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-grpc</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>io.quarkus</groupId>
```
snazy, line 51: I think this artifact should not use Quarkus but "plain protobuf + grpc".

grpc/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-model</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
```
snazy, line 43: I think this artifact should not use Quarkus but "plain protobuf + grpc".

clients/grpc-client/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie-clients</artifactId>
+    <version>0.9.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>grpc-client</artifactId>
```
snazy: Why is this a separate artifact and not part of `:nessie-client` with an optional dependency to `:nessie-grpc`?

grpc/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie</artifactId>
+    <version>0.9.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>grpc</artifactId>
```
snazy: Should be `nessie-grpc`

nastra:
@laurentgo / @snazy / @rymurr this should be good for a first pass. In the meantime I'm working on adding some tests for the `ProtoUtil` conversion stuff and looking into gRPC authentication.

grpc/pom.xml
snazy: Should be `:nessie-grpc`

grpc/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+
+  <name>Nessie - GRPC</name>
```
snazy, line 30: However, I suspect this artifact can go into `:nessie-model` with _optional_ dependencies to protobuf + grpc

model/src/main/java/org/projectnessie/api/HttpConfigApi.java
```diff
@@ -0,0 +1,25 @@
...
+package org.projectnessie.api;
+
+import javax.ws.rs.Path;
+
+/**
+ * The purpose of this class is to only put the {@link Path} annotation onto it as otherwise other
+ * (non-http) implementors of {@link ConfigApi} could be used by CDI when a request comes in.
+ */
+@Path("config")
+public interface HttpConfigApi extends ConfigApi {}
```
snazy, line 25: Feels like the `Http*Api` interfaces should live in `org.projectnessie.api.http`

servers/quarkus-server/src/main/java/org/projectnessie/server/grpc/TreeService.java
```diff
@@ -0,0 +1,227 @@
...
+  public void getAllReferences(Empty request, StreamObserver<GetAllReferencesResponse> observer) {
+    try {
+      observer.onNext(
+          GetAllReferencesResponse.newBuilder()
+              .addAllReference(
+                  treeResource.getAllReferences().stream()
+                      .map(ProtoUtil::refToProto)
+                      .collect(Collectors.toList()))
+              .build());
+      observer.onCompleted();
```
snazy, line 70: The `onNext` + `onCompleted`/`onError` can probably be shortened / wrapped as you did on the client side

grpc/src/main/java/org/projectnessie/grpc/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+    if (null != commitMeta.getCommitter()) builder.setCommitter(commitMeta.getCommitter());
+    if (null != commitMeta.getCommitTime())
+      builder.setCommitTime(
+          Timestamp.newBuilder()
+              .setSeconds(commitMeta.getCommitTime().getEpochSecond())
+              .setNanos(commitMeta.getCommitTime().getNano())
+              .build());
+    if (null != commitMeta.getAuthorTime())
+      builder.setAuthorTime(
+          Timestamp.newBuilder()
```
snazy, line 314: duplicated - should be a util method

grpc/src/main/java/org/projectnessie/grpc/ProtoUtil.java
```diff
@@ -0,0 +1,474 @@
...
+
+  public static CommitMeta fromProto(org.projectnessie.api.grpc.CommitMeta commitMeta) {
+    return CommitMeta.builder()
+        .author(commitMeta.getAuthor())
+        // .committer(commitMeta.getCommitter())
+        .hash(commitMeta.getHash())
+        .message(commitMeta.getMessage())
+        .properties(commitMeta.getPropertiesMap())
+        .commitTime(
+            Instant.ofEpochSecond(
```
snazy, line 331: duplicated - should be a util method

clients/grpc-client/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+
+  <name>Nessie - Client - GRPC</name>
+
+  <dependencies>
```
snazy, line 32: Having the code in this artifact in `:nessie-client` should also help to revert the `nessie-client` artifact-dance in the (spark-)tests

snazy (changes requested):

clients/client/src/main/java/org/projectnessie/client/HttpClientBuilder.java
```diff
@@ -0,0 +1,241 @@
...
+import java.net.URI;
+import java.util.function.Function;
+import org.projectnessie.client.http.HttpClientException;
+import org.projectnessie.client.http.HttpClientReadTimeoutException;
+import org.projectnessie.model.Validation;
+
+/**
+ * A builder class that creates a {@link NessieHttpClient} via {@link HttpClientBuilder#builder()}.
+ */
+public class HttpClientBuilder {
```
snazy, line 36: move to `org.projectnessie.client.http` or so

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcClientBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+package org.projectnessie.client.grpc;
+
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import org.projectnessie.client.NessieClient;
+
+/**
+ * A builder class that creates a {@link NessieGrpcClient} via {@link GrpcClientBuilder#builder()}.
+ */
+public class GrpcClientBuilder {
```
snazy, line 25: Needs a test like `TestNessieHttpClient`

model/src/main/java/org/projectnessie/api/ConfigApi.java
nastra: see PR description for this

> all validation-specific (and HTTP-specific) annotations is still in the ConfigApi / ContentsApi / TreeApi interfaces because one can't add validation constraints to subclasses of ConfigApi / ContentsApi / TreeApi (the idea was that HttpConfigApi / HttpContentsApi / HttpTreeApi could define all those annotations but the Bean validation Spec doesn't allow that)

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcClientBuilder.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Sets the Nessie gRPC Server endpoint.
+   *
+   * @param endpoint The gRPC Server endpoint
+   * @return {@code this}
+   */
+  public GrpcClientBuilder withEndpoint(String endpoint) {
+    this.endpoint = endpoint;
```
snazy, line 45: Any reason why this cannot be a URI so HttpClientBuilder + GrpcClientBuilder can share that property (and have a commit AbstractClientBuilder class)?

model/src/main/java/org/projectnessie/api/HttpConfigApi.java
nastra: yep makes sense

clients/grpc-client/pom.xml
nastra: to me this clear distinctions seems better. Otherwise users that use the HTTP client (`nessie-client` module) all of a sudden get gRPC dependencies pulled in without expecting it

grpc/pom.xml
nastra: I was hesitant on doing that. Let's see what others think

clients/grpc-client/pom.xml
nastra: as mentioned in the other comment, I think http client + grpc client code should stay in their own modules

nastra: fwiw, moving the grpc client stuff into `nessie-client` doesn't fix the circular dependency issue that we have (where delta/iceberg can depend on potentially older versions of the nessie-client)

grpc/pom.xml
nastra: agreed. fixed

grpc/pom.xml
nastra: fixed

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcClientBuilder.java
nastra: made `endpoint` an `URI` but I don't see any good reason for introducing an `AbstractClientBuilder`

clients/grpc-client/pom.xml
snazy: Http isn't in its own module, no?

clients/grpc-client/pom.xml
snazy: Not if the protobuf + grpc dependencies are optional/provided

model/src/main/java/org/projectnessie/api/ConfigApi.java
snazy: Any other solution for this maybe?

snazy: Having REST + OpenAPI definitions "leak" into grpc doesn't feel good either.

clients/grpc-client/pom.xml
nastra: not sure I follow, can you elaborate what you mean exactly?

model/src/main/java/org/projectnessie/api/ConfigApi.java
nastra: oh I absolutely agree with you that this doesn't feel good, but I haven't found a good solution for this yet.

clients/grpc-client/pom.xml
snazy: You said: "_I think http client + grpc client code should stay in their own modules_" - but http is not in its own module - it's in `:nessie-client` - not in `:nessie-client-http` analogous to `:nessie-client-grpc`.

nastra: my assumption was that the current `nessie-client` == http client, therefore it's kind of in its own module, but intertwined with the general client APIs (since there was only one client implementation). I don't have a strong preference where the grpc client-code should live. Let's see what @rymurr thinks.

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcClientBuilder.java
nastra: done

grpc/src/main/java/org/projectnessie/grpc/ProtoUtil.java
nastra: done

servers/quarkus-server/src/main/java/org/projectnessie/server/grpc/TreeService.java
nastra: done

clients/grpc-client/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <!-- javax.ws.rs-api / resteasy-core-spi /com.fasterxml.jackson.core are only needed for
+    handling WebApplicationException/ResteasyViolationException/JsonParseException/JsonMappingException -->
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+      <version>2.1.1</version>
```
rymurr, line 38: this version should prob be in parent pom

clients/grpc-client/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty-shaded</artifactId>
+      <version>${grpc.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
```
rymurr, line 79: this and hte one below should be tests scope?

grpc/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+      <artifactId>grpc-stub</artifactId>
+      <version>${grpc.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
```
rymurr, line 58: test scope

clients/grpc-client/pom.xml
nastra: fixed

clients/grpc-client/pom.xml
nastra: fixed

grpc/pom.xml
nastra: fixed

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcExceptionMapper.java
```diff
@@ -0,0 +1,199 @@
...
+    if (ex instanceof JsonParseException
+        || ex instanceof JsonMappingException
+        || ex instanceof IllegalArgumentException
+        || ex instanceof WebApplicationException) {
+      return Status.INVALID_ARGUMENT
+          .withDescription(ex.getMessage())
+          .withCause(ex)
+          .asRuntimeException();
+    }
+    if (ex instanceof ResteasyViolationException) {
```
dimas-b, line 67: nit: style: we have new lines before `if` is some cases here, and no new line in others

grpc/src/main/proto/contents_service.proto
```diff
@@ -0,0 +1,85 @@
...
+message ContentsRequest {
+  ContentsKey contentsKey = 1;
+  string ref = 2;
+  string hashOnRef = 3;
+}
+
+message MultipleContentsRequest {
+  string ref = 1;
+  string hashOnRef = 2;
+  repeated ContentsKey requestedKeys = 3;
```
dimas-b, line 45: nit: inconsistent order of fields... or was it intentional?

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+  string hash = 2;
+}
+
+message Hash {
+  string hash = 1;
+}
+
+message CreateReferenceRequest {
+  string sourceRefName = 1;
+  Reference reference = 2;
```
dimas-b, line 68: I know it matches current java API, but does it make sense to have `Reference` as a parameter here? We cannot create a `Hash` this way and for `Branch` and `Tag` we cannot accept their pre-set `hash` attributes... WDYT?

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+message GetReferenceByNameRequest {
+  string namedRef = 1;
+}
+
+message AssignReferenceRequest {
+  string namedRef = 1;
+  string oldHash = 2;
+  oneof assignTo {
+    Branch branch = 3;
+    Tag tag = 4;
```
dimas-b, line 84: Similar concern here: We cannot accept pre-set `hash` attributes of `Branch` or `Tag` in this context, can we? Would it make sense to redefine the request fields?

dimas-b: From my POV, `namedRef` + `oldHash` could be wrapped into a `Reference` instead of the `assignTo` value :thinking:

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+  }
+}
+
+message DeleteReferenceRequest {
+  string namedRef = 1;
+  string hash = 2;
```
dimas-b, line 90: Could this be a simple `Reference` objects (assuming we deny deleting plain `Hash` references)?

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+}
+
+message CommitLogRequest {
+  string namedRef = 1;
+  string startHash = 2;
+  string endHash = 3;
+  int32 maxRecords = 4;
+  string pageToken = 5;
```
dimas-b, line 98: nit: `pageToken` -> `pagingState`? The work "token" looks a bit odd to me in this context :shrug:

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+}
+
+message Entry {
+  ContentsType type = 1;
+  ContentsKey contentsKey = 2;
+}
+
+message TransplantRequest {
+  string branchName = 1;
+  string hash = 2;
```
dimas-b, line 149: nit: `hash` -> `expectedHash`?

grpc/src/main/proto/tree_service.proto
```diff
@@ -0,0 +1,192 @@
...
+  string message = 3;
+  string fromRefName = 4;
+  repeated string hashesToTransplant = 5;
+}
+
+message MergeRequest {
+  string fromHash = 1;
+  string toBranch = 2;
+  string expectedHash = 3;
+  string fromRefName = 4;
```
dimas-b, line 159: nit: inconsistent parameter order... keep related params together (e.g. `fromRefName` and `fromHash`)?

dimas-b: I wonder it we could systematically use `Reference` (and sub-types) for arguments that have a "name" and "expected hash"... for example `MergeRequest` could be:
```
message MergeRequest {
  Reference from = 1;
  Branch to = 2;
}
```
WDYT?

dimas-b: ^ that is not really specific to the gRPC stuff, but more of a general API review for 1.0, I guess :)

clients/grpc-client/src/main/java/org/projectnessie/client/grpc/GrpcExceptionMapper.java
nastra: fixed

grpc/src/main/proto/contents_service.proto
nastra: you mean in comparison with `ContentsRequest`? yes that's on purpose. The order of fields is mapped in the same way as they are in the `ContentsResource` class

grpc/src/main/proto/tree_service.proto
nastra: yeah right now it's really mapped 1:1 with the Java API until we have a good reason to deviate from that

grpc/src/main/proto/tree_service.proto
nastra: same here, it's mapped 1:1 with the Java API, where the validation logic on the server accepts a Reference object

grpc/src/main/proto/tree_service.proto
nastra: I think we should then rename it in the REST API first as naming is a 1:1 mapping.

grpc/src/main/proto/tree_service.proto
nastra: yeah overall I think we should fix stuff like this first in the REST API

dimas-b: Yeah, I mentioned that in #1804 . Let's review APIs under that issue :+1:

dimas-b (approved):

harshm-dremio:
Is this PR missing anything? curious as there's not much update in recent times.

PR #1776, opened 2021-08-17
TVS throw ref not found + expose hash for "beginning of time"

snazy:
Current `DatabaseAdapter` API returns an empty stream if a ref could not be found, which is a left-over from the removal of all the added `NamedRef`s. `DatabaseAdapter` should throw `ReferenceNotFoundException`s instread.

Also exposes a function to return the hash for "beginning of time", which will become useful in following PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1776)

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -217,11 +217,13 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
 
     // 3. Collect commit-log-entries
     List<CommitLogEntry> toEntriesReverseChronological =
-        takeUntilExcludeLast(readCommitLogStream(ctx, toHead), e -> e.getHash().equals(commonAncestor))
+        takeUntilExcludeLast(
+                readCommitLogStream(ctx, toHead), e -> e.getHash().equals(commonAncestor))
             .collect(Collectors.toList());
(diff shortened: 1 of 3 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -431,8 +446,16 @@ protected Hash hashOnRef(
   protected abstract List<CommitLogEntry> fetchPageFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes);
 
   /** Reads from the commit-log starting at the given commit-log-hash. */
-  protected Stream<CommitLogEntry> commitLogFetcher(OP_CONTEXT ctx, Hash initialHash) {
-    return logFetcher(ctx, initialHash, this::fetchPageFromCommitLog, CommitLogEntry::getParents);
+  protected Stream<CommitLogEntry> commitLogFetcher(OP_CONTEXT ctx, Hash initialHash)
```
rymurr, line 449: nit: this name is slightly confusing. Its name is basically the same as the ones above and its arguments are the same. Could you rename it?

rymurr (commented):
one nit looks gpod otherwise.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: done

rymurr (approved):

PR #1777, opened 2021-08-17
Support bearer auth tokens

dimas-b:
* Using Quarkus OpenID Client


clients/client/src/test/java/org/projectnessie/client/auth/TestBearerAuthFilter.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client.auth;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
snazy, line 18: Nit: use assertj (we use it everywhere else)

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -110,7 +110,8 @@ quarkus.resteasy.gzip.enabled=true
 #quarkus.oidc.auth-server-url=
 # fixed at buildtime
 quarkus.http.auth.basic=false
-quarkus.oidc.enabled=false
+quarkus.oidc.enabled=true
```
snazy, line 113: Not actually a fan of enabling this by default. I suspect, that all requests even for a "try it out w/ docker" scenario would require OIDC, then?

snazy: For test classes, you can use a custom quarkus test profile:
* a class that extends QuarkusTestProfile - one for native tests and one for non-native tests
* annotate the tests using `@io.quarkus.test.junit.TestProfile`

dimas-b: Do we not want to have OIDC enabled by default? I though that was the intent. It has to be enabled at build time if it is to be usable. 

I believe we can have it enabled but not used in OSS docker images by default. As far as I can tell it should not break anything. Users can specify a keycloak URL at run time as a parameter if they want to integrate with it (this part I have not done yet).

 @rymurr WDYT?

rymurr: The ideal scenario is that OIDC is off by default but is easily activated. We want to cater for:

1. easy to try out and play w/ Nessie docker image w/o building an auth mechanism
2. being able to integrate the docker image into a prod setup w/ OIDC enabled

Is that possible? By the sounds of it you are saying auth is off unless config parameters are set at runtime to enable oidc explicitly? to me 1 is more important than 2 for the base image 

An aside, did you test anything other than keycloak? I believe there were some issues w/ other providers...

clients/client/src/test/java/org/projectnessie/client/auth/TestBearerAuthFilter.java
dimas-b: Migrated to AssertJ in 4a3d18717ad8bc8d65bfc422bdb34a3c7abe5084

servers/quarkus-server/src/main/resources/application.properties
dimas-b: > you are saying auth is off unless config parameters are set at runtime to enable oidc explicitly?

That's my understanding of Quarkus docs. `quarkus.oidc.enabled` is one of a few settings that must be set at build time. I'll try and confirm manually.

https://quarkus.io/guides/security-openid-connect#configuring-using-the-application-properties-file (`quarkus.oidc.enabled` is marked with a lock sign).

dimas-b: > did you test anything other than keycloak? I believe there were some issues w/ other providers...

So far I testing only with keycloak. Will test others too.

dimas-b: Confirmed: `quarkus.oidc.enabled` _cannot_ be overwritten to `true` at runtime.

```
2021-08-18 11:04:30,022 WARN  [io.qua.run.con.ConfigChangeRecorder] (main) Build time property cannot be changed at runtime:
 - quarkus.oidc.enabled was 'false' at build time and is now 'true'
```

I'll see if it is possible to have reasonable defaults with OIDC enabled to avoid 500 errors in case a user tries `Bearer` auth without properly configuring the server.

README.md
```diff
@@ -35,6 +35,42 @@ From there, you can use one of our technology integrations such those for
...
+user identity.
+
+In production, Nessie supports OpenID, which can be enabled by setting the following Quarkus properties:
+* `nessie.auth.enabled=true`
+* `quarkus.oidc.auth-server-url=<OpenID Server URL>`
+
+For example, to run Nessie with Keycloak in docker first start Keycloak:
+
+```
+docker run -p 8080:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin --name keycloak quay.io/keycloak/keycloak:15.0.1
```
snazy, line 50: It's a bit confusing here. The paragraph above say "in production", but this docker-run is pretty insecure. Not sure whether we should guide people about how to run Keycloak. Is there a docs page for Keycloak about how to run it in Docker for a test drive?

snazy: would `quay.io/keycloak/keycloak:latest` be better here?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
```diff
@@ -0,0 +1,105 @@
...
+        return anonymous();
+      }
+    }
+
+    return super.attemptAuthentication(context)
+        .onItem()
+        .transformToUni(
+            securityIdentity -> {
+              if (securityIdentity == null) {
+                if (authEnabled) {
```
snazy, line 81: How can `authEnabled` be `false` here?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
```diff
@@ -0,0 +1,105 @@
...
+ * <ul>
+ *   <li>Prevents the Quarkus OIDC authentication mechanism from attempting authentication when it
+ *       is not configured.
+ *   <li>Completely disallows unauthenticated requests when authentication is enabled.
+ * </ul>
+ */
+@Alternative // @Alternative + @Priority ensure the original HttpAuthenticator bean is not used
+@Priority(1)
+@ApplicationScoped
+public class NessieHttpAuthenticator extends HttpAuthenticator {
```
snazy, line 47: I think, the mechanism should really live inside Quarkus. Wanna raise a PR against Quarkus to add a runtime-config-flag there? WDYT?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -110,8 +110,11 @@ quarkus.resteasy.gzip.enabled=true
 #quarkus.oidc.auth-server-url=
 # fixed at buildtime
 quarkus.http.auth.basic=false
-quarkus.oidc.enabled=false
-
+quarkus.oidc.enabled=true
+quarkus.oidc.client-id=projectnessie
+nessie.auth.enabled=false
+# to be overwritten by end user when enabling OpenID validation
+quarkus.oidc.auth-server-url=http://localhost/auth/realms/unset/
```
snazy, line 117: Is this parameter mandatory if `quarkus.oidc.enabled=true` is set?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -110,8 +110,11 @@ quarkus.resteasy.gzip.enabled=true
 #quarkus.oidc.auth-server-url=
 # fixed at buildtime
 quarkus.http.auth.basic=false
-quarkus.oidc.enabled=false
-
+quarkus.oidc.enabled=true
+quarkus.oidc.client-id=projectnessie
```
snazy, line 114: Is this parameter mandatory if `quarkus.oidc.enabled=true` is set?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -172,6 +175,8 @@ quarkus.container-image.name=nessie
 %test.quarkus.security.users.embedded.roles.admin_user=admin,user
 %test.quarkus.security.users.embedded.roles.test_user=test123
 %test.nessie.version.store.jgit.type=INMEMORY
+%test.quarkus.keycloak.devservices.enabled=false
```
snazy, line 178: Do we need these two properties here or could those go into the test-profile?

README.md
dimas-b: I did not want to use `latest` to avoid surprises when things change, but it's probably stable enough for this example... will change to `latest`

dimas-b: Re: "in production". I guess this really means we do not support `Basic` auth in production. I can rephrase to that effect.

dimas-b: Re: Keycloak, they certainly have "how to" docs, but I wanted to give a specific example that people can try and run. Would it be preferable to have a generic "auth URL" in the example and let them figure out how to configure the auth server? 

Another option, I suppose, is to put a specific example in docs rather than README. WDYT?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: it can happen in test. The top-level `if` for the "anonymous" case catches only header-based auth, but some tests use annotation-based auth that does not use the header.

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: Good point. Let me think about how we can structure that in Quarkus code :thinking:

servers/quarkus-server/src/main/resources/application.properties
dimas-b: It has to be syntactically valid, otherwise the server will not start. It will not be used unless `quarkus.oidc.enabled` is `true`.

servers/quarkus-server/src/main/resources/application.properties
dimas-b: `devservices` configs must be set at build time.... but let me try if setting them in test profiles work.

servers/quarkus-server/src/main/resources/application.properties
dimas-b: No. Removed in 089e6c36d10a943566eff5e094a161b4653da443

servers/quarkus-server/src/main/resources/application.properties
dimas-b: Even though Quarkus docs say it defaults to `true`, in practice it turned out setting this parameter was not necessary.... removed in 089e6c36d10a943566eff5e094a161b4653da443

servers/quarkus-server/src/main/resources/application.properties
dimas-b: I changed the default URL to use a documentation IP address in 089e6c36d10a943566eff5e094a161b4653da443

README.md
dimas-b: I've pushed a new version of README in 50fd26f200704543af4418ed53349858d0fbd5b8. How does it sound?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -110,8 +110,10 @@ quarkus.resteasy.gzip.enabled=true
 #quarkus.oidc.auth-server-url=
 # fixed at buildtime
 quarkus.http.auth.basic=false
-quarkus.oidc.enabled=false
-
+quarkus.oidc.enabled=true
+nessie.auth.enabled=false
```
nastra, line 114: maybe rename this to `nessie.server.authentication.enabled` to be consistent with other settings that we have

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
snazy: AH, okay. Can you add a comment explaining that?

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.server.authn;
+
+import com.google.common.collect.ImmutableMap;
+import io.quarkus.test.junit.QuarkusTestProfile;
+import java.util.Map;
+
+/** A simple {@link QuarkusTestProfile} that disabled Nessie authentication. */
+public class AuthenticationDisabledProfile implements QuarkusTestProfile {
```
snazy, line 23: One thing I noticed w/ Quarkus test profiles:
You need actually two profiles: one for "normal" unit tests and one when running the tests against a native image. The latter needs
```
  @Override
  public String getConfigProfile() {
    return "prod";
  }
```

servers/quarkus-server/src/main/resources/application.properties
dimas-b: will do

clients/client/src/main/java/org/projectnessie/client/NessieClient.java
```diff
@@ -216,6 +217,16 @@ public Builder withPassword(String password) {
       return this;
     }
 
+    /**
+     * Set the token for {@link AuthType#BEARER} authentication.
+     *
+     * @param token bearer token
+     * @return {@code this}
+     */
+    public Builder withBearerToken(String token) {
```
laurentgo, line 226: Is it the right approach to have specific methods for each auth type? It doesn't seem very extensible, and would force to handle more cases where this kind of method would be invoked with an invalid auth type. Maybe create a separate auth interface with various implementations to capture supported auth methods?

dimas-b: Yes, I'm +1 to extracting individual interfaces for each auth type. This conflicts with #1816 though... which one do we want to merge first?

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
dimas-b: Do you think we should have separate tests for auth in native mode? I assumed it was not necessary since the logic is contained in a basic java class.

dimas-b: I've added `ITNativeOpenIdAuthentication` in 739575aeb2cbc3975b57cb11b1eda4127434ed5a. I guess it is not necessary to run native tests for basic auth since it is only meant to be supported for unit tests. WDYT?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: I hope the latest version of `NessieHttpAuthenticator` is cleaner. We do not have to work around any Quarkus-specific behaviour now. I guess we do not really need any quarkus-side fixes now. WDYT?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: I've refactored the configs and `NessieHttpAuthenticator` `if` statements in 9afbec6a279d242e5565ed38bd53523f6ca9ab67. I hope it is clearer now.

clients/client/src/main/java/org/projectnessie/client/NessieClient.java
snazy: #1819 is there as a draft (builds upon #1816). I don't mind merging this one first.

README.md
```diff
@@ -35,6 +35,26 @@ From there, you can use one of our technology integrations such those for
 Have fun! We have a Google Group and a Slack channel we use for both developers and 
 users. Check them out [here](https://projectnessie.org/develop/).
 
+### Authentication
+
+By default, Nessie servers run with authentication disabled and all requests are processed under the "anonymous"
+user identity.
+
+Nessie supports bearer tokens and uses [OpenID Connect](https://openid.net/connect/) for validating them.
```
snazy, line 43: nit: just mention something like "and HTTP BASIC for test purposes" or so.

servers/quarkus-server/src/main/resources/application.properties
dimas-b: Renamed to `nessie.server.auth.enabled` (for brevity) in 9afbec6a279d242e5565ed38bd53523f6ca9ab67

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
snazy: I'm okay with that.

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
snazy: Yea - that's okay I think.

snazy (commented):
LGTM!
Just a few very minor nits and then I think this one is ready for merge.

README.md
dimas-b: I'm not sure. Prebuilt Nessie jars and docker images actually do not have Basic Auth enabled. Basic auth works only in unit tests (when the `test` quarkus` profile is active)... but it is worth putting in README?

snazy: Ah, true, you‘re right

nastra (dismissed):

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
```diff
@@ -23,7 +23,7 @@
 public class AuthenticationDisabledProfile implements QuarkusTestProfile {
 
   public static final Map<String, String> CONFIG_OVERRIDES =
-      ImmutableMap.of("nessie.auth.enabled", "false");
+      ImmutableMap.of("nessie.authentication.enabled", "false");
```
snazy, line 26: Should this be `nessie.server.authentication.enabled`?

servers/quarkus-server/src/test/java/org/projectnessie/server/TestDisabledAuthentication.java
```diff
@@ -26,8 +26,8 @@
 import org.projectnessie.server.authn.AuthenticationDisabledProfile;
 
 /**
- * This test validates that setting `nessie.auth.enabled=false` allows all requests regardless of
- * their authentication type.
+ * This test validates that setting `nessie.authentication.enabled=false` allows all requests
```
snazy, line 29: Should this be `nessie.server.authentication.enabled`?

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
dimas-b: ah... should not have renamed late in the evening :facepalm:

servers/quarkus-server/src/test/java/org/projectnessie/server/TestDisabledAuthentication.java
dimas-b: good catch, thx!

servers/quarkus-server/src/test/java/org/projectnessie/server/authn/AuthenticationDisabledProfile.java
dimas-b: fixed in a68aa0d0341b44dc7e39c4cf2cfa28f30dcb062e

servers/quarkus-server/src/test/java/org/projectnessie/server/TestDisabledAuthentication.java
dimas-b: fixed in a68aa0d0341b44dc7e39c4cf2cfa28f30dcb062e

snazy (approved):

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -107,11 +107,12 @@ quarkus.resteasy.gzip.enabled=true
 ## Quarkus auth settings
 #quarkus.oidc.credentials.secret=
 #quarkus.oidc.client-id=
-#quarkus.oidc.auth-server-url=
+nessie.server.authentication.enabled=false
+# to be overwritten by end user when enabling OpenID validation
+quarkus.oidc.auth-server-url=http://127.255.0.0:0/auth/realms/unset/
```
dimas-b, line 112: @snazy : I had to go back to a loopback address here, because using example addresses like `192.0.2.1` apparently adds about 10 sec to server startup time... I guess Quarkus tries to access it on startup, however, using a fake address like this does not break the startup.

README.md
```diff
@@ -35,6 +35,26 @@ From there, you can use one of our technology integrations such those for
...
+
+Authentication can be enabled by setting the following Quarkus properties:
+* `nessie.server.authentication.enabled=true`
+* `quarkus.oidc.auth-server-url=<OpenID Server URL>`
```
rymurr, line 47: as a follow up PR can we have a little blog showing a full keycloak or google example?

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
```diff
@@ -0,0 +1,89 @@
...
+
+    return super.attemptAuthentication(context)
+        .onItem()
+        .transform(
+            securityIdentity -> {
+              if (securityIdentity == null) {
+                // Disallow unauthenticated requests when requested by configuration.
+                // Note: Quarkus by default permits unauthenticated requests unless there are
+                // specific authorization rules that validate the security identity.
+                throw new AuthenticationFailedException("Not authenticated");
```
rymurr, line 79: Is this failed because no auth was provided or because auth failed?

If its because no auth provided maybe we can state that in the exception

snazy (approved):

nastra (dismissed):

README.md
dimas-b: I was thinking about a section in the docs, but a blog post is probably even better.

servers/quarkus-server/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: In practice this usually happens when the use did not provide any credentials, but it might happen when the credentials provided were not recognized by the server (e.g. sending `Basic` auth when basic is disabled).

rymurr: It would be good if the error message were a bit cleaner. Assuming that error message would eventually get to the user?

dimas-b: I've rebased and updated the message to `Missing or unrecognized credentials`

dimas-b: I do not think this message gets to the user now, but it might in the future :thinking:

PR #1778, opened 2021-08-18
Make setup.py executable

dimas-b:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1778)


rymurr:
Why do you want to do this @dimas-b ? I don't normally see `setup.py` be executable. One usually runs `python setup.py bdist_wheel` for example.

dimas-b:
I just noticed the shebang line at the start of `setup.py`. It does not make sense without the executable flag, so I thought it was just missed :)

dimas-b:
I'm fine with keeping it as a plain text file.

PR #1781, opened 2021-08-18
Fix "named-ref-or-throw" for getCommitLog() + fix stream-handling

snazy:
1. Stream handling: ensure that Java streams, which might hold open resources, are properly closed once
those have been used.

2. Always check the "namedRef" in TreeResource.getCommitLog()

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1781)

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
@@ -20,6 +20,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -60,10 +61,6 @@ protected BaseResource(
     this.accessChecker = accessChecker;
   }
 
-  Optional<Hash> getHash(String ref) {
```
rymurr, line 63: were these methods not used? Seems like we just added them no?

snazy: Nope, no longer used.

rymurr (approved):

PR #1783, opened 2021-08-18
Improve circular dependency handling to NessieClient

nastra:
The circular dependecy is fixed by introducing the same NessieClient version that Iceberg uses.

This also moves NessieClient instantiation into subclasses of AbstractSparkTest so that each subclass can control how to instantiate the NessieClient (depending on the Nessie version they pull in)

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1783)

### DIFF
--- a/clients/spark-extensions/pom.xml
+++ b/clients/spark-extensions/pom.xml
@@ -33,7 +33,19 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${project.version}</version>
+      <version>${iceberg.nessie.version}</version>
+      <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
+      <exclusions>
+        <exclusion>
+          <groupId>org.projectnessie</groupId>
+          <artifactId>nessie-model</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-model</artifactId>
+      <version>${iceberg.nessie.version}</version>
     </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
@@ -136,10 +148,10 @@
       <plugin>
         <groupId>org.projectnessie</groupId>
         <artifactId>nessie-apprunner-maven-plugin</artifactId>
-        <version>${project.version}</version>
+        <version>${iceberg.nessie.version}</version>
         <configuration>
           <skip>${skipTests}</skip>
-          <appArtifactId>org.projectnessie:nessie-quarkus:${project.version}</appArtifactId>
+          <appArtifactId>org.projectnessie:nessie-quarkus:${iceberg.nessie.version}</appArtifactId>
           <applicationProperties>
             <quarkus.http.test-port>0</quarkus.http.test-port>
           </applicationProperties>
@@ -151,7 +163,7 @@
           <dependency>
             <groupId>org.projectnessie</groupId>
             <artifactId>nessie-quarkus</artifactId>
-            <version>${project.version}</version>
+            <version>${iceberg.nessie.version}</version>
           </dependency>
         </dependencies>
         <executions>
--- a/clients/spark-extensions/pom.xml
+++ b/clients/spark-extensions/pom.xml
@@ -34,18 +34,7 @@
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
       <version>${client.nessie.version}</version>
-      <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
-      <exclusions>
-        <exclusion>
-          <groupId>org.projectnessie</groupId>
-          <artifactId>nessie-model</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>org.projectnessie</groupId>
-      <artifactId>nessie-model</artifactId>
-      <version>${client.nessie.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
--- a/clients/spark-extensions/pom.xml
+++ b/clients/spark-extensions/pom.xml
@@ -33,7 +33,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${iceberg.nessie.version}</version>
+      <version>${client.nessie.version}</version>
       <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
       <exclusions>
         <exclusion>
@@ -45,7 +45,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-model</artifactId>
-      <version>${iceberg.nessie.version}</version>
+      <version>${client.nessie.version}</version>
     </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
@@ -148,10 +148,10 @@
       <plugin>
         <groupId>org.projectnessie</groupId>
         <artifactId>nessie-apprunner-maven-plugin</artifactId>
-        <version>${iceberg.nessie.version}</version>
+        <version>${client.nessie.version}</version>
         <configuration>
           <skip>${skipTests}</skip>
-          <appArtifactId>org.projectnessie:nessie-quarkus:${iceberg.nessie.version}</appArtifactId>
+          <appArtifactId>org.projectnessie:nessie-quarkus:${client.nessie.version}</appArtifactId>
           <applicationProperties>
             <quarkus.http.test-port>0</quarkus.http.test-port>
           </applicationProperties>
@@ -163,7 +163,7 @@
           <dependency>
             <groupId>org.projectnessie</groupId>
             <artifactId>nessie-quarkus</artifactId>
-            <version>${iceberg.nessie.version}</version>
+            <version>${client.nessie.version}</version>
           </dependency>
         </dependencies>
         <executions>
--- a/clients/tests/pom.xml
+++ b/clients/tests/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${project.version}</version>
+      <version>${iceberg.nessie.version}</version>
       <scope>provided</scope>
     </dependency>
 
--- a/clients/tests/pom.xml
+++ b/clients/tests/pom.xml
@@ -51,7 +51,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${iceberg.nessie.version}</version>
+      <version>${client.nessie.version}</version>
       <scope>provided</scope>
     </dependency>
 
--- a/versioned/gc/iceberg-actions/pom.xml
+++ b/versioned/gc/iceberg-actions/pom.xml
@@ -34,7 +34,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${iceberg.nessie.version}</version>
+      <version>${client.nessie.version}</version>
       <scope>test</scope>
       <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
       <exclusions>
@@ -258,10 +258,10 @@
       <plugin>
         <groupId>org.projectnessie</groupId>
         <artifactId>nessie-apprunner-maven-plugin</artifactId>
-        <version>${iceberg.nessie.version}</version>
+        <version>${client.nessie.version}</version>
         <configuration>
           <skip>${skipTests}</skip>
-          <appArtifactId>org.projectnessie:nessie-quarkus:${iceberg.nessie.version}</appArtifactId>
+          <appArtifactId>org.projectnessie:nessie-quarkus:${client.nessie.version}</appArtifactId>
           <applicationProperties>
             <quarkus.http.test-port>0</quarkus.http.test-port>
             <nessie.version.store.dynamo.initialize>true</nessie.version.store.dynamo.initialize>
@@ -281,7 +281,7 @@
           <dependency>
             <groupId>org.projectnessie</groupId>
             <artifactId>nessie-quarkus</artifactId>
-            <version>${iceberg.nessie.version}</version>
+            <version>${client.nessie.version}</version>
           </dependency>
         </dependencies>
         <executions>
--- a/versioned/gc/iceberg/pom.xml
+++ b/versioned/gc/iceberg/pom.xml
@@ -34,7 +34,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${iceberg.nessie.version}</version>
+      <version>${client.nessie.version}</version>
       <scope>test</scope>
       <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
       <exclusions>
@@ -250,10 +250,10 @@
       <plugin>
         <groupId>org.projectnessie</groupId>
         <artifactId>nessie-apprunner-maven-plugin</artifactId>
-        <version>${iceberg.nessie.version}</version>
+        <version>${client.nessie.version}</version>
         <configuration>
           <skip>${skipTests}</skip>
-          <appArtifactId>org.projectnessie:nessie-quarkus:${iceberg.nessie.version}</appArtifactId>
+          <appArtifactId>org.projectnessie:nessie-quarkus:${client.nessie.version}</appArtifactId>
           <applicationProperties>
             <quarkus.http.test-port>0</quarkus.http.test-port>
             <nessie.version.store.dynamo.initialize>true</nessie.version.store.dynamo.initialize>
@@ -273,7 +273,7 @@
           <dependency>
             <groupId>org.projectnessie</groupId>
             <artifactId>nessie-quarkus</artifactId>
-            <version>${iceberg.nessie.version}</version>
+            <version>${client.nessie.version}</version>
           </dependency>
         </dependencies>
         <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -130,6 +130,8 @@
     <awssdk.version>2.17.25</awssdk.version>
     <bouncycastle.version>1.66</bouncycastle.version>
     <cel.version>0.2.0</cel.version>
+    <!-- to fix circular dependencies with NessieClient, certain projects need to use the same Nessie version as Iceberg/Delta has -->
+    <client.nessie.version>0.9.0</client.nessie.version>
     <deltalake2.version>0.6.2-nessie</deltalake2.version>
     <deltalake3.version>0.8.0-nessie</deltalake3.version>
     <dynamodb.version>1.12.0</dynamodb.version>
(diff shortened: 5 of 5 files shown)


clients/spark-extensions/pom.xml
```diff
@@ -63,7 +58,6 @@
     <dependency>
       <groupId>org.apache.iceberg</groupId>
       <artifactId>iceberg-spark3-runtime</artifactId>
-      <scope>test</scope>
```
rymurr, line 66: this has to be a test scope. The code has no dependency on iceberg runtime. This would ship the iceberg runtime w/ delta artifacts, which we don't want

rymurr (changes requested):
I think it would be good to follow a similar pattern to the gc poms. Where the server is started up on a certain version of nessie so we are explicit about what is supported here. We have to provide some nessie client w/o explicitly adding iceberg or delta as runtime dependencies though,

nastra:
I just opened https://github.com/projectnessie/nessie/pull/1785 to fix a similar issue with GC stuff

clients/spark-extensions/pom.xml
nastra: `NessieUtils` is using the `NessieClient`, therefore I removed the `test` scope so that `NessieUtils` uses whatever Nessie version that comes with `iceberg-spark3-runtime`

rymurr: yes, but if we do that then the iceberg runtime (all 100MB of it) becomes a hard dependency. We don't want that. This module isn't iceberg specific and we don't want to add iceberg as a dependency

rymurr:
> I just opened #1785 to fix a similar issue with GC stuff

I think the gc stuff is correct, it doesnt have a circular dependency currently

rymurr:
or more correctly, it manages the circular dependency by being speciifc about which versions are used in tests

nastra:
@rymurr I pushed a new version that forces a specific version of the NessieClient on the Spark Sql extensions and the GC modules

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
```diff
@@ -49,6 +49,8 @@
 
   protected static NessieClient nessieClient;
 
+  protected abstract void initNessieClient();
```
snazy, line 52: this should return the `NessieClient` and not let the implementations update the field in this class IMO

snazy: the new mechanism also creates a new `NessieClient` instance for each test, but never closes them. either create only one instance or close the instances after each test (I'd prefer one instance per class)

snazy:
Can you elaborate more on the "circular" dependeny in the PR description (and later commit summary)?

rymurr:
would it be easier to make the abstract class not depend on a nessie client or make the dependency <optional>? Seems like an awkward way to ensure the tests run?

nastra:
> Can you elaborate more on the "circular" dependeny in the PR description (and later commit summary)?

The problem is that the Spark SQL extentions for example rely on `iceberg-spark3-runtime`, which in turn is compiled against a Nessie version that isn't necessarily the same that we use here. Now when the tests run, we run against a `nessie-client` version that is different from Iceberg and stuff breaks. 
This happens when moving things around in the `nessie-client` project (as I'm doing in https://github.com/projectnessie/nessie/pull/1767).

nastra:
> would it be easier to make the abstract class not depend on a nessie client or make the dependency ? Seems like an awkward way to ensure the tests run?

yes that would probably better. Let me remove the client entirely from that project

versioned/gc/iceberg/pom.xml
```diff
@@ -30,11 +30,13 @@
...
   <name>Nessie - Versioned - GC for Spark - Iceberg</name>
 
-  <properties>
-    <iceberg.nessie.version>${project.version}</iceberg.nessie.version>
-  </properties>
-
   <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-client</artifactId>
```
rymurr, line 36: for this and above I would rather not add a nessie version explicitly. There is one in the iceberg runtime and we want to make sure the iceberg runtime one is used. As taht is how it will be used in production. This could mask issues w/ the client versions

clients/tests/pom.xml
```diff
@@ -48,13 +48,6 @@
       <version>${spark3.version}</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>org.projectnessie</groupId>
-      <artifactId>nessie-client</artifactId>
-      <version>${project.version}</version>
-      <scope>provided</scope>
```
rymurr, line 55: Why doesn't provided or optional work? Are the upcoming changes to the client so drastic that we have to add this code duplication? why can't the clients/tests module import the `iceberg.nessie.version` of the client instead of this refactor?

nastra: the idea is that delta for example might end up having a different version of the `nessie-client` than iceberg or not? That's why it seems safer to move the `nessie-client` to where it's actually required (into the subprojects)

versioned/gc/iceberg/pom.xml
nastra: let me double-check whether this works without adding an explicit dependency to `nessie-client`

nastra: ah now I remember why it was failing. I missed the fact that `iceberg.nessie.version` was defined in each of the GC poms and I only updated one of them and wondered why it failed, thus leading me to adding an explicit `nessie-client` dependency

clients/tests/pom.xml
nastra: so technically speaking we could then also have a `deltalake.nessie.version` which is being used in the delta projects, thus making our lives easier in case Delta&Iceberg stuff uses different nessie versions

versioned/gc/iceberg/pom.xml
nastra: fixed it

nastra: ok i double checked this and it doesn't work without forcing the `nessie-client` version here to be `${iceberg.nessie.version}`. It's either that, or we run into the jackson issue that you already ran into (where you moved the `iceberg-spark3-runtime` dependency to the very end of the pom:
```
Caused by: org.apache.iceberg.shaded.com.fasterxml.jackson.databind.exc.InvalidDefinitionException: 
Cannot construct instance of `org.projectnessie.model.Reference` (no Creators, like default constructor, exist): abstract types either need to be mapped to concrete types, have custom deserializer, or contain additional type information
```

snazy (commented):
The logic to create the nessie-client, delete branches after tests and close the client is duplicated a lot. Can you move that into `AbstractSparkTest` - at best deleting all existing branches and then re-creating the default branch?
A `compileOnly` scope should work.

nastra:
> The logic to create the nessie-client, delete branches after tests and close the client is duplicated a lot. Can you move that into `AbstractSparkTest` - at best deleting all existing branches and then re-creating the default branch?
> A `compileOnly` scope should work.

The whole idea is to leave the `nessie-client` out of `clients/tests/pom.xml` because it's possible that subprojects (delta, iceberg) at different versions might depend on different versions of `nessie-client`. Right now this is only reflected via `iceberg.nessie.version` but I imagine us also having a `delta.nessie.version` field

versioned/gc/iceberg-actions/pom.xml
```diff
@@ -30,11 +30,12 @@
...
-  <properties>
-    <iceberg.nessie.version>${project.version}</iceberg.nessie.version>
-  </properties>
-
   <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-client</artifactId>
+      <version>${iceberg.nessie.version}</version>
+    </dependency>
```
snazy, line 38: Could be 
```

    <dependency>
      <groupId>org.projectnessie</groupId>
      <artifactId>nessie-client</artifactId>
      <version>${iceberg.nessie.version}</version>
      <scope>test</scope>
      <exclusions>
        <exclusion>
          <groupId>org.projectnessie</groupId>
          <artifactId>nessie-model</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
 ```

Comment (nastra) [reply versioned/gc/iceberg-actions/pom.xml:38]: context: we discussed this during standup where @snazy had a similar problem.

@snazy do you want to fix this as part of a follow-up PR or should I fix it here?

Comment (nastra) [reply versioned/gc/iceberg-actions/pom.xml:38]: I'll make a follow up PR shortly with a concrete problem description for this part

**Review by snazy — COMMENTED** [review]
I think, all dependencies to `<artifactId>nessie-client</artifactId>` in the changed poms should have the `<scope>test</scope>`.

File: clients/spark-extensions/pom.xml [hunk 6]
```diff
@@ -33,7 +33,7 @@
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${project.version}</version>
+      <version>${iceberg.nessie.version}</version>
```
snazy, line 36: Does this particular one really need to be bumped?
I'm a bit worried about this particular one, because it's a production-dependency and `nessie-client` depends on `nessie-model`.

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
```diff
@@ -67,8 +64,6 @@ protected void create() throws IOException {
         .set("spark.sql.catalog.nessie", "org.apache.iceberg.spark.SparkCatalog");
     spark = SparkSession.builder().master("local[2]").config(conf).getOrCreate();
     spark.sparkContext().setLogLevel("WARN");
-
-    nessieClient = NessieClient.builder().withUri(url).build();
   }
 
   @AfterAll
```
snazy, line 69: Can you change this to `@AfterEach` as well? It feels weird to initialize a SparkSession for every test but only clean up the last one. Can be a follow-up though.

clients/spark-extensions/pom.xml
nastra: yes, this definitely needs to be bumped

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
nastra: yeah should probably be a follow-up

clients/spark-extensions/pom.xml
nastra: you mean we should also exclude the model classes here?

nastra: I had another look at the pom and turns out this was running with the nessie-quarkus `project.version` and not the `iceberg.nessie.version`. Pushed a fix for this and also forced the `nessie-model` version to be `iceberg.nessie.version`

clients/spark-extensions/pom.xml
```diff
@@ -136,10 +148,10 @@
       <plugin>
         <groupId>org.projectnessie</groupId>
         <artifactId>nessie-apprunner-maven-plugin</artifactId>
-        <version>${project.version}</version>
+        <version>${iceberg.nessie.version}</version>
```
snazy, line 151: Unnecessary change - the plugin can still be the project's version

clients/spark-extensions/pom.xml
snazy: Really not happy w/ having these versions as _production_ dependencies here. I think, that makes it really complicated to reason about and prone to more difficult dependency issues.

It's already unclear, which nessie and 3rd party dependencies will land here in which versions. `nessie-client-tests` brings in more dependencies.

The changes to the gc-modules are fine IMO, because those only affect tests, but this change affects production dependencies.

clients/spark-extensions/pom.xml
nastra: it's necessary with the `nessie-model` changes and startup breaks

clients/spark-extensions/pom.xml
nastra: @rymurr thoughts on @snazy's concerns? I currently don't see any easier way around this problem than running the spark sql extensions code with the nessie version that comes with iceberg.
imo we need to rethink the dependencies between nessie and iceberg entirely as the way it is currently structured is just error prone

rymurr: why can't this just be `provided` or `optional` scope? We just don't ship it with the shaded jar or as a dependency and then the delta or iceberg runtimes use their nessie client that is already available to them.

I thikn we need to make sure it generates a dependency reduced pom and that shading is correct though

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
```diff
@@ -67,12 +64,10 @@ protected void create() throws IOException {
...
     spark = SparkSession.builder().master("local[2]").config(conf).getOrCreate();
     spark.sparkContext().setLogLevel("WARN");
-
-    nessieClient = NessieClient.builder().withUri(url).build();
   }
 
-  @AfterAll
-  static void tearDown() throws Exception {
+  @AfterEach
+  protected void tearDown() {
```
rymurr, line 70: I don't think we want to stop spark after each run. Why change this?

versioned/gc/iceberg-actions/pom.xml
```diff
@@ -30,11 +30,20 @@
 
   <name>Nessie - Versioned - GC for Spark - Iceberg - Actions</name>
 
-  <properties>
-    <iceberg.nessie.version>${project.version}</iceberg.nessie.version>
-  </properties>
-
   <dependencies>
+    <dependency>
```
rymurr, line 34: why does this have to be added? When it exists in the iceberg runtime?

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
```diff
@@ -44,6 +47,21 @@
 
   @TempDir File tempPath;
 
+  private NessieClient nessieClient;
```
rymurr, line 50: Why does this have to live here, why not in abstract test? (same for other tests)

rymurr (commented):
I am not 100% sure why we cant have a single `iceberg.nessie.version`, which is renamed `client.nessie.version` that we support and assume all the things we ship to clients only work on those. Until we sort out backwards compatibility permanently (eg until 1.0). I think this change can be reduced in scope then. We are saying we only support a single version of pre-1.0 nessie at any time.

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
nastra: see my comment in https://github.com/projectnessie/nessie/pull/1783/files#r691250922

versioned/gc/iceberg-actions/pom.xml
nastra: outdated code

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
nastra: requested in https://github.com/projectnessie/nessie/pull/1783#discussion_r693755112

rymurr: Spark initialization is a bit tricky and stopping it like this could lead to funky issues. It maintains a lot of static state. It also takes a long time to start up so this adds a ton of time to tests.

@snazy we aren't initializing the Spark server in every run, we are initializing it in the first then getting it in the rest (hence `getOrCreate`). If this is particularly bothersome i would `create` in `BeforeAll` and `get` in `BeforeEach`

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
rymurr: yes, I don't think that answered the question. As per my [comment](https://github.com/projectnessie/nessie/pull/1783#pullrequestreview-736888401) I don't see what is wrong w forcing 1 (pre1.0) client version for all these tests and using it from the abstract test. The solution here creates a lot of duplicate code and confusing pom changes imho

nastra: but won't we eventually end up having different nessie-client versions for Delta and Iceberg? How do we guarantee that all the things use the same `client.nessie.version` when Delta/Iceberg release at a different pace?

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
snazy: Bummer - missed the `getOrCreate`

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
rymurr: my thinking is the following:

1. we only need this pre-1.0 as we will be introducing backwards compatibility soon
2. the delta code will soon either be deleted or moved to the delta repo depending on discussions w/ Databricks.

snazy:
I doubt we need this change with #1816, because that abstracts changes to Nessie-Client and its implementation(s) from the call sites.

rymurr:
> I doubt we need this change with #1816, because that abstracts changes to Nessie-Client and its implementation(s) from the call sites.

can you expand on that...i am not sure what you mean.

clients/deltalake/core/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
nastra: ok seems reasonable. Pushed a new simplified version and renamed `iceberg.nessie.version` to `client.nessie.version`

snazy:
> > I doubt we need this change with #1816, because that abstracts changes to Nessie-Client and its implementation(s) from the call sites.
> 
> can you expand on that...i am not sure what you mean.

AFAIK this PR and #1818 are "caused by" #1767, which changes `NessieClient` and its builders (breaks backwards compatibility). #1816 introduces client-builder interfaces but keeps the existing `NessieClient.Builder` for backwards-compatibility (can be removed once all projects that depend on it - aka Iceberg) have been migrated.

nastra:
> > > I doubt we need this change with #1816, because that abstracts changes to Nessie-Client and its implementation(s) from the call sites.
> > 
> > 
> > can you expand on that...i am not sure what you mean.
> 
> AFAIK this PR and #1818 are "caused by" #1767, which changes `NessieClient` and its builders (breaks backwards compatibility). #1816 introduces client-builder interfaces but keeps the existing `NessieClient.Builder` for backwards-compatibility (can be removed once all projects that depend on it - aka Iceberg) have been migrated.

Right, but #1816 only overcomes the problem for the HttpClient stuff. Next time when somebody changes something in `nessie-client` before we have backwards compatibility in place, that person runs into the same problem as we have right now

snazy:
> Right, but #1816 only overcomes the problem for the HttpClient stuff. Next time when somebody changes something in `nessie-client` before we have backwards compatibility in place, that person runs into the same problem as we have right now

Maybe, maybe not. Any change post GA must be implemented in a backwards compatible way w/ a migration strategy.

clients/tests/src/main/java/org/projectnessie/client/tests/AbstractSparkTest.java
nastra: reverted those changes

clients/spark-extensions/pom.xml
nastra: fixed

rymurr (approved):

PR #1785, opened 2021-08-18
Fix circular dependency to NessieClient in gc modules

nastra:
We remove the circular dependency by directly depending on the
NessieClient version that comes with the `iceberg-spark3-runtime`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1785)

### DIFF
--- a/versioned/gc/iceberg-actions/pom.xml
+++ b/versioned/gc/iceberg-actions/pom.xml
@@ -36,6 +36,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.iceberg</groupId>
+      <artifactId>iceberg-spark3-runtime</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-versioned-gc-iceberg</artifactId>
       <version>${project.version}</version>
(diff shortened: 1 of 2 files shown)


versioned/gc/iceberg-actions/pom.xml
```diff
@@ -142,11 +146,6 @@
       <artifactId>DynamoDBLocal</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
```
rymurr, line 145: we want it to be provided by the spark environment and it was at the bottom because of a dependency issue in jackson.

versioned/gc/iceberg/pom.xml
```diff
@@ -46,11 +50,6 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.apache.iceberg</groupId>
-      <artifactId>iceberg-spark3-runtime</artifactId>
-      <scope>provided</scope>
```
rymurr, line 52: same, we want it to be provided by the spark runtime

rymurr (changes requested):

nastra:
closing this. I'll fix stuff in #1783

PR #1796, opened 2021-08-19
Remove remainders of Hive

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1796)


nastra (dismissed):

model/src/main/java/org/projectnessie/model/SqlView.java
```diff
@@ -29,7 +29,6 @@
 public abstract class SqlView extends Contents {
 
   public static enum Dialect {
-    HIVE,
```
rymurr, line 32: Hive is still a valid dialect, this shouldn't have been removed

servers/store/src/main/proto/table.proto
```diff
@@ -23,10 +23,8 @@ option java_generate_equals_and_hash = true;
 message Contents {
   oneof object_type {
     IcebergTable iceberg_table = 1;
-    HiveTable hive_table = 2;
```
rymurr, line 26: imho we shouldn't remove/reorder protobufs. Fine given its pre-1.0 but we shouldn't do so again

PR #1797, opened 2021-08-19
API-versioning w/ HTTP headers - changes to nessie-http

snazy:
Another apporach to add "versioning" to the (REST) API is to leverage the `Content-Type` + `Accept` headers,
which can be "selected" in JAX-RS via the `@Consumes` + `@Produces` annotations. I.e. a changed API endpoint
keeps its old implementation and gets another one using the same `@Path` and HTTP-method but with different
media-types.

This change just prepares the `HttpClient` + `HttpRequest` classes for this scenario.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1797)

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
@@ -32,7 +32,6 @@ import javax.net.ssl.SSLContext;
 public class HttpClient {
 
   private final URI baseUri;
-  private final String accept;
   private final ObjectMapper mapper;
   private final SSLContext sslContext;
   private final int readTimeoutMillis;
@@ -51,19 +50,16 @@ public class HttpClient {
    * Construct an HTTP client with a universal Accept header.
    *
    * @param baseUri uri base eg https://example.com
-   * @param accept Accept header eg "application/json"
    * @param readTimeoutMillis timeout to wait for response from server, in milliseconds
    * @param connectionTimeoutMillis timeout to wait to connecto to server, in milliseconds
    */
   private HttpClient(
       URI baseUri,
-      String accept,
       ObjectMapper mapper,
       SSLContext sslContext,
       int readTimeoutMillis,
       int connectionTimeoutMillis) {
     this.baseUri = Objects.requireNonNull(baseUri);
-    this.accept = HttpUtils.checkNonNullTrim(accept);
     this.mapper = mapper;
     this.sslContext = sslContext;
     this.readTimeoutMillis = readTimeoutMillis;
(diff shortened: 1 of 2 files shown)


dimas-b:
@snazy : Just curious, what would versioned `Content-Type` and `Accept` header values look like?

snazy:
> @snazy : Just curious, what would versioned `Content-Type` and `Accept` header values look like?

It can be really any content-type. The only thing that's important is that the subtype ends with `+json`  (e.g. `application/nessie+globalstate+json`) so that the built-in `MediaTypeReader`s + `MediaTypeWriter`s work for those types.

clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
```diff
@@ -132,15 +126,6 @@ public HttpClientBuilder setBaseUri(URI baseUri) {
       return this;
     }
 
-    public String getAccept() {
-      return accept;
-    }
-
-    public HttpClientBuilder setAccept(String accept) {
```
dimas-b, line 139: Do we need to worry about breaking existing clients by removing this method? I guess not since we're still at version `0.*`, but just wanted to double check.

snazy: No, it didn't make sense to set it to a different value - otherwise all API requests would have been broken.

dimas-b (dismissed):

clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
```diff
@@ -102,14 +112,15 @@ private HttpResponse executeRequest(Method method, Object body) throws HttpClien
...
+        putHeader("Accept", accept, headers);
         headers.entrySet().stream()
             .flatMap(e -> e.getValue().stream().map(x -> new SimpleImmutableEntry<>(e.getKey(), x)))
             .forEach(x -> con.setRequestProperty(x.getKey(), x.getValue()));
         con.setRequestMethod(method.name());
         if (method.equals(Method.PUT) || method.equals(Method.POST)) {
           // Need to set the Content-Type even if body==null, otherwise the server responds with
           // RESTEASY003065: Cannot consume content type
-          con.setRequestProperty("Content-Type", "application/json; charset=utf-8");
+          con.setRequestProperty("Content-Type", contentsType + "; charset=utf-8");
```
rymurr, line 123: i think `; charset=utf-8` should be part of the static above.

snazy: updated

rymurr (approved):

nastra (approved):

PR #1814, opened 2021-08-23
Replace usage of deprecated ConfigProperties with ConfigMapping

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1814)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -155,7 +155,7 @@
     <openapi.version>2.0</openapi.version>
     <prometheus.version>0.9.0</prometheus.version>
     <protobuf.version>3.17.3</protobuf.version>
-    <quarkus.version>2.1.3.Final</quarkus.version>
+    <quarkus.version>2.2.0.CR1</quarkus.version>
     <reactor.version>2020.0.10</reactor.version>
     <scala2.12.version>2.12.13</scala2.12.version>
     <scala2.13.version>2.13.6</scala2.13.version>
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusServerConfig.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusServerConfig.java
@@ -15,19 +15,22 @@
  */
 package org.projectnessie.server.config;
 
-import io.quarkus.arc.config.ConfigProperties;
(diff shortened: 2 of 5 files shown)


servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusServerConfig.java
```diff
@@ -16,18 +16,21 @@
 package org.projectnessie.server.config;
 
 import io.quarkus.arc.config.ConfigProperties;
-import org.eclipse.microprofile.config.inject.ConfigProperty;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
 import org.projectnessie.services.config.ServerConfig;
 
 /** Nessie server config for Quarkus. */
 @ConfigProperties(prefix = "nessie.server")
```
nastra, line 24: switching this one currently still fails and I need to investigate why

dimas-b (approved):

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusServerConfig.java
dimas-b: Did we have to upgrade Quarkus to fix this?

dimas-b (approved):

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusServerConfig.java
nastra: yep, it was wrongly marked in their release notes that it's fixed in 2.1.3.Final, but it was actually fixed in 2.2.0.CR1

pom.xml
```diff
@@ -155,7 +155,7 @@
     <openapi.version>2.0</openapi.version>
     <prometheus.version>0.9.0</prometheus.version>
     <protobuf.version>3.17.3</protobuf.version>
-    <quarkus.version>2.1.3.Final</quarkus.version>
+    <quarkus.version>2.2.0.CR1</quarkus.version>
```
nastra, line 158: `2.2.0.Final` should be released today, so we/dependabot can switch to that new version then

PR #1816, opened 2021-08-23
Abstract NessieClientBuilder + Move HTTP client to separate package

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1816)

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
@@ -71,18 +71,6 @@ public interface NessieClientBuilder<IMPL extends NessieClientBuilder<IMPL>> {
   IMPL withUri(String uri);
 
   /**
-   * Set the repository owner + ID within the Nessie instance.
-   *
-   * <p>{@link NessieClient#getContentsApi()} and {@link NessieClient#getTreeApi()} are scoped to
-   * the repository specified by this option.
-   *
-   * @param owner repository owner
-   * @param repo repository ID
-   * @return {@code this}
-   */
-  IMPL withRepoOwner(String owner, String repo);
-
-  /**
    * Set the username for {@link AuthType#BASIC} authentication.
    *
    * @param username username
--- a/clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client;
+
+import java.net.URI;
+import java.util.function.Function;
+import org.projectnessie.client.NessieClient.AuthType;
+
+/**
+ * {@link NessieClient} builder interface.
+ *
+ * @param <IMPL> concrete nessie-client builder type
+ */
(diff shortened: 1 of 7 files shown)


clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+    return this;
+  }
+
+  /**
+   * Set the password for {@link AuthType#BASIC} authentication.
+   *
+   * @param password password
+   * @return {@code this}
+   */
+  public HttpClientBuilder withPassword(String password) {
```
dimas-b, line 186: Would it make sense to move auth parameters to the base `NessieClientBuilder` interface? I imagine we ought to support the same set of credential types across all clients (even though the credential transport may have to be different).

snazy: Right, those should not really differ semantically. Added that.

dimas-b (approved):

nastra (approved):
LGTM, the only thing that I noticed is that there's already a `HttpClient` in the `http` package, so that might cause confusion as having a `HttpClientBuilder` could imply that one creates a `HttpClient`. Not sure if we should name the builder `NessieHttpClientBuilder` or so

clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+import java.net.URI;
+import java.util.function.Function;
+import org.projectnessie.client.NessieClient.AuthType;
+
+/**
+ * {@link NessieClient} builder interface.
+ *
+ * @param <IMPL> concrete nessie-client builder type
+ */
+public interface NessieClientBuilder<IMPL extends NessieClientBuilder<IMPL>> {
```
rymurr, line 27: we discussed a protocol version and another parameter here also right?

Is this the only (future) way to get a Nessie client?

clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -0,0 +1,247 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client.http;
+
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_AUTH_TYPE;
```
rymurr, line 18: some of these constants are beyond the scope of just http client right? eg auth stuff, tracing

clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -0,0 +1,247 @@
...
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_OWNER;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_PASSWORD;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_REPOSITORY;
```
rymurr, line 21: can we just do a straight remove of all the repo/owner stuff. Nothing downstream is using it. I think it just adds confusion here.

rymurr (commented):
agreed w/ @nastra re renaming and a few other comments

clients/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
snazy: Yes, this should become the only way to build+get a Nessie client. The current `NessieClient.Builder` shall go away.

Once this one and a follow-up to add the "API version" + "API invocation builder pattern" are in, we can nuke `NessieClient` + the current `TreeApi` + `ConfigApi` + `ContentsApi` interfaces. But it's a long(er) way to get there.

clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
snazy: Can do it in an immediate follow-up.

clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
snazy: yes

PR #1818, opened 2021-08-23
Remove circular dependency to NessieClient in GC modules

nastra:
The circular dependecy is fixed by introducing the same NessieClient
version that Iceberg uses.

Note that we need to exclude nessie-model from the GC modules so that we
run the code with the same model classes that ship with iceberg's nessie
version.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1818)

### DIFF
--- a/versioned/gc/iceberg-actions/pom.xml
+++ b/versioned/gc/iceberg-actions/pom.xml
@@ -30,11 +30,20 @@
 
   <name>Nessie - Versioned - GC for Spark - Iceberg - Actions</name>
 
-  <properties>
-    <iceberg.nessie.version>${project.version}</iceberg.nessie.version>
-  </properties>
-
   <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-client</artifactId>
+      <version>${iceberg.nessie.version}</version>
+      <scope>test</scope>
+      <!-- we need to exclude this here so that we run the code with the same model classes that ship with iceberg's nessie version -->
+      <exclusions>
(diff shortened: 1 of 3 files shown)


snazy (approved):

versioned/gc/iceberg-actions/pom.xml
```diff
@@ -30,11 +30,20 @@
...
-  <properties>
-    <iceberg.nessie.version>${project.version}</iceberg.nessie.version>
-  </properties>
-
   <dependencies>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-client</artifactId>
+      <version>${iceberg.nessie.version}</version>
+      <scope>test</scope>
```
rymurr, line 38: why did we have to add this

nastra: are you referring to the exclusions? @snazy ran into a very similar problem as I did. He had breaking changes in the model classes and so that means we need to use the correct model classes here

rymurr: I was referring to the entire dependency.

nastra: see my comment on https://github.com/projectnessie/nessie/pull/1783#discussion_r691368847. It doesn't work without forcing the `nessie-client` version

rymurr: all we ddi was move a property to the high level pom. Why did it require the introduction of new dependencies ot the module?

PR #1819, opened 2021-08-23
Auth providers

snazy:
Replaces string-based authentication provider mechanism with a service-loader based mechanism using Java interfaces.

Authentication providers for `NessieClient` can be either instantiated direct or indirect via `NessieClientBuilder.fromConfig()` or `NessieClientBuilder.withAuthenticationFromConfig()`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1819)


clients/client/src/main/java/org/projectnessie/client/auth/NessieAuthenticationProvider.java
```diff
@@ -0,0 +1,80 @@
...
+          String.format("No authentication provider for '%s' found.", authType));
+    }
+    return null;
+  }
+
+  /**
+   * The authentication type discriminator. If authentication is configured via properties, like
+   * {@link org.projectnessie.client.NessieClientBuilder#fromConfig(Function)}/{@link
+   * org.projectnessie.client.NessieClientBuilder#withAuthenticationFromConfig(Function)}, set the a
+   * [@link NessieAuthenticationProvider} instance is used, if {@link
```
dimas-b, line 66: Invalid link opening char: `[`

clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -0,0 +1,211 @@
...
+   * @return {@code this}
+   */
+  @Override
+  public HttpClientBuilder withUri(String uri) {
+    return withUri(URI.create(uri));
+  }
+
+  @Override
+  public HttpClientBuilder withAuthentication(NessieAuthentication authentication) {
+    this.authentication = authentication;
```
dimas-b, line 144: I guess we can check the specific type of `NessieAuthentication` rather than in the http client when it has to build headers. WDYT?

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -0,0 +1,90 @@
...
+      authHeaderValue = "Basic " + encodedString;
+    }
+
+    @Override
+    public RequestFilter buildRequestFilter() {
+      return new BasicHttpAuthentication(authHeaderValue);
+    }
+  }
+
+  private static class BasicHttpAuthentication implements RequestFilter {
```
dimas-b, line 77: nit: Rename to `BasicHttpAuthenticationFilter` to avoid confusion with `BasicAuthentication`?

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -0,0 +1,90 @@
...
+
+    if (username == null || password == null) {
+      throw new NullPointerException(
+          "username and password parameters must be present for auth type " + AUTH_TYPE_VALUE);
+    }
+
+    return new BasicAuthentication(username, password);
+  }
+
+  private static class BasicAuthentication implements NessieAuthentication, HttpAuthentication {
```
dimas-b, line 60: nit: `HttpAuthentication` extends `NessieAuthentication`... why do we declare both?

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -0,0 +1,90 @@
...
+ * HTTP BASIC authentication provider.
+ *
+ * <p>Takes parameters {@link org.projectnessie.client.NessieConfigConstants#CONF_NESSIE_USERNAME}
+ * and {@link org.projectnessie.client.NessieConfigConstants#CONF_NESSIE_PASSWORD}.
+ */
+public class BasicAuthenticationProvider implements NessieAuthenticationProvider {
+
+  public static final String AUTH_TYPE_VALUE = "BASIC";
+
+  public static NessieAuthentication of(String username, String password) {
```
dimas-b, line 38: Could we return `HttpAuthentication` to be more specific?

dimas-b: nit: `BasicAuthenticationProvider.of(name,pwd)` looks odd to me because it return an "authentication" not a provider...

How about:
```
public class BasicHttpAuthentication implements NessieAuthenticationProvider {
  public static HttpAuthentication of(String username, String password) {... }
}
}

File: clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java (hunk 2 above)
Comment (snazy) [reply clients/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java]: right, sounds good!

File: clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java (hunk 4 above)
Comment (snazy) [reply clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java]: no specific reason actually, reduced to one

File: clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java (hunk 5 above)
Comment (snazy) [reply clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java]: Hm - I wanted to avoid exposing "internal" classes. Maybe rename `of` to `create`?

Comment (dimas-b) [reply clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java]: `create` SGTM :+1:

**Review by rymurr — APPROVED** [review]

File: clients/client/src/main/java/org/projectnessie/client/NessieClient.java [hunk 6]
```diff
@@ -127,6 +161,28 @@ public Builder withConnectionTimeout(int connectionTimeoutMillis) {
...
+            s -> {
+              switch (s) {
+                case NessieConfigConstants.CONF_NESSIE_AUTH_TYPE:
+                  return auth.name();
+                case NessieConfigConstants.CONF_NESSIE_USERNAME:
+                  return username;
+                case NessieConfigConstants.CONF_NESSIE_PASSWORD:
+                  return password;
+                case NessieConfigConstants.CONF_NESSIE_AWS_REGION:
+                  return null;
```
dimas-b, line 179: Would it make sense to throw an `IllegalArgumentException` in this case? It does not look like AWS if fully supported in by this builder.

snazy: It's a new config option, there was no way to set it in the old builder that's eventually going away.

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -0,0 +1,80 @@
...
+      this.username = username;
+      this.password = password;
+    }
+
+    @Override
+    public void applyToHttpClient(HttpClient client) {
+      String userPass = username + ':' + password;
+      byte[] encoded = Base64.getEncoder().encode(userPass.getBytes(StandardCharsets.UTF_8));
+      String encodedString = new String(encoded, StandardCharsets.UTF_8);
+      String authHeaderValue = "Basic " + encodedString;
```
dimas-b, line 75: nit: It might be preferable to do this in the constructor. No functional difference, but we avoid holding on to plain text passwords, which is a good practice, IMHO.

clients/client/src/main/java/org/projectnessie/client/auth/NessieAuthenticationProvider.java
```diff
@@ -0,0 +1,105 @@
...
+              + "AuthenticationProvider";
+
+      NessieAuthenticationProvider ap = null;
+      try {
+        @SuppressWarnings("unchecked")
+        Class<NessieAuthenticationProvider> provider =
+            (Class<NessieAuthenticationProvider>) Class.forName(reflectionClassName);
+        ap = provider.getDeclaredConstructor().newInstance();
+      } catch (Exception ignore) {
+        // ignore
```
dimas-b, line 75: Should we at least debug log it? Class loading problems can be very hard to debug without evidence :thinking:

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
rymurr: yikes! good point! +1 from me on that one

clients/client/src/main/java/org/projectnessie/client/auth/NessieAuthenticationProvider.java
dimas-b: Here it could be loading problems, cast problem, clinit problems... WDYT?

clients/client/src/main/java/org/projectnessie/client/NessieClient.java
dimas-b: noted

clients/client/src/main/java/org/projectnessie/client/auth/NessieAuthenticationProvider.java
rymurr: +1 its an awful broad exception

PR #1909, opened 2021-09-07
start adding unit tests to UI

rymurr:
We do several things here:

* remove dead svg
* add jest unit tests
* add jest snapshots
* start refactoring to make UI easier to test and use fewer REST calls

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1909)


ui/src/Explore/CommitHeader.tsx
```diff
@@ -16,43 +16,24 @@
...
-      if(data.operations && data.operations.length > 0) {
-        setLog(data.operations[0]);
-      }
-    }).catch(t => console.log(t));
-}
-
-function CommitHeader(props: {currentRef: string}) {
-
-  const [currentLog, setLog] = useState<CommitMeta>();
+function CommitHeader({committer, author, hash, message, properties, commitTime}: CommitMeta) {
```
dimas-b, line 22: WDYT about using `CommitMeta` as the only arg to simplify parameter passing to this method?

dimas-b (approved):

ui/src/Explore/CommitHeader.tsx
rymurr: It has no effect on parameter passing. You can only pass basic types through these react components. This is syntactic sugar to avoid having to reference each element of the commit log

.github/actions/ci-results/action.yml
```diff
@@ -21,3 +21,9 @@ runs:
         flags: java
         files: |
           code-coverage/target/site/jacoco-aggregate-all/jacoco.xml
+    - uses: codecov/codecov-action@v1
+      with:
+        flags: javascript
```
snazy, line 26: why not put `java,javascript` into the action above?

.github/actions/ci-results/action.yml
```diff
@@ -21,3 +21,9 @@ runs:
         flags: java
         files: |
           code-coverage/target/site/jacoco-aggregate-all/jacoco.xml
+    - uses: codecov/codecov-action@v1
```
snazy, line 24: could bump to `v2` btw

.github/actions/ci-results/action.yml
rymurr: the tag is per upload. So all java changes would get both tags and all javascript changes would get both tags

.github/actions/ci-results/action.yml
rymurr: yeah, have a separate PR for that

.github/actions/ci-results/action.yml
snazy: urgh

snazy (approved):

ui/src/Explore/CommitHeader.tsx
dimas-b: TIL :+1:

PR #1922, opened 2021-09-08
Global-states: API changes (non-intrusive)

snazy:
This PR only adds API+model changes, but does **not** change any behavior. Actual behavior changes and
implementations for global-state are subject to a follow-up PR.

Introduces the model changes for global-state.

* `IcebergTable` gets a new attribute `snapshotId`. The already existing `metadataLocation` will represent
  the "global state" and `snapshotId` the "per-reference state`

Operations:

* `Put` updates the `IcebergTable` for a key and gets a new attribute `expectedValue`. When updating an
  Iceberg table, clients pass the "last known" version of `IcebergTable` as the `expectedValue`. The server
  side automatically extracts the global-state from that `expectedValue`.
* No put-global operation.

No other API changes.

The initial version for global-states looked similar to this one. An intermediate version of this change
did model the global-state and on-reference-state as separate object types - on-reference-states were
updated using `Put` and global-states updated using an added `PutGlobal` operation. The latter approach
distinguished the meaning of the individual objects more clearly, but was much more difficult to implement
and use.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1922)

### DIFF
--- a/model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
@@ -47,7 +47,21 @@ import org.projectnessie.model.Validation;
 @Path("contents")
 public interface HttpContentsApi extends ContentsApi {
 
-  /** Get the properties of an object. */
+  /**
+   * This operation returns the {@link Contents} for a {@link ContentsKey} in a named-reference (a
+   * {@link org.projectnessie.model.Branch} or {@link org.projectnessie.model.Tag}).
+   *
+   * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Contents}
+   * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID).
+   *
+   * @param key the {@link ContentsKey}s to retrieve
+   * @param ref named-reference to retrieve the contents for
+   * @param hashOnRef hash on {@code ref} to retrieve the contents for, translates to {@code HEAD},
+   *     if missing/{@code null}
+   * @return list of {@link org.projectnessie.model.MultiGetContentsResponse.ContentsWithKey}s
+   * @throws NessieNotFoundException if {@code ref} or {@code hashOnRef} does not exist
+   */
   @GET
   @Produces(MediaType.APPLICATION_JSON)
   @Path("{key}")
@@ -99,6 +113,23 @@ public interface HttpContentsApi extends ContentsApi {
           String hashOnRef)
       throws NessieNotFoundException;
 
+  /**
+   * Similar to {@link #getContents(ContentsKey, String, String)}, but takes multiple {@link
+   * ContentsKey}s and returns the {@link Contents} for the one or more {@link ContentsKey}s in a
+   * named-reference (a {@link org.projectnessie.model.Branch} or {@link
+   * org.projectnessie.model.Tag}).
+   *
+   * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Contents}
+   * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID).
+   *
+   * @param ref named-reference to retrieve the contents for
+   * @param hashOnRef hash on {@code ref} to retrieve the contents for, translates to {@code HEAD},
+   *     if missing/{@code null}
+   * @param request the {@link ContentsKey}s to retrieve
+   * @return list of {@link org.projectnessie.model.MultiGetContentsResponse.ContentsWithKey}s
+   * @throws NessieNotFoundException if {@code ref} or {@code hashOnRef} does not exist
+   */
   @POST
   @Consumes(MediaType.APPLICATION_JSON)
   @Operation(
--- a/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
@@ -571,6 +571,19 @@ public interface HttpTreeApi extends TreeApi {
           Merge merge)
       throws NessieNotFoundException, NessieConflictException;
 
+  /**
+   * Commit multiple operations against the given branch expecting that branch to have the given
+   * hash as its latest commit. The hash in the successful response contains the hash of the commit
+   * that contains the operations of the invocation.
+   *
+   * @param branchName Branch to change, defaults to default branch.
+   * @param hash Expected hash of branch.
+   * @param operations {@link Operations} to apply
+   * @return updated {@link Branch} objects with the hash of the new HEAD
+   * @throws NessieNotFoundException if {@code branchName} could not be found
+   * @throws NessieConflictException if the operations could not be applied to some conflict, which
+   *     is either caused by a conflicting commit or concurrent commits.
+   */
   @POST
   @Path("branch/{branchName}/commit")
   @Consumes(MediaType.APPLICATION_JSON)
--- a/model/src/main/java/org/projectnessie/model/IcebergTable.java
+++ b/model/src/main/java/org/projectnessie/model/IcebergTable.java
@@ -18,21 +18,80 @@ package org.projectnessie.model;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.immutables.value.Value;
 
+/**
+ * Represents the state of an Iceberg table in Nessie. An Iceberg table is globally identified via
+ * its {@link Contents#getId() unique ID}.
+ *
+ * <p>The Iceberg-table-state consists of the location to the table-metadata and the snapshot-ID.
+ *
+ * <p>The table-metadata-location is managed globally within Nessie, which means that all versions
+ * of the same table share the table-metadata across all named-references (branches and tags). There
+ * is only one version, the current version, of Iceberg's table-metadata.
+ *
+ * <p>Within each named-reference (branch or tag), the (current) Iceberg-snapshot-ID will be
+ * different. In other words: changes to an Iceberg table update the snapshot-ID.
+ *
+ * <p>When adding a new table (aka contents-object identified by a contents-id), use a {@link
+ * org.projectnessie.model.Operation.Put} without an expected-value. In all other cases (updating an
+ * existing table). always pass the last known version of {@link IcebergTable} as the expected-value
+ * within the put-operation.
+ */
+@Schema(
+    type = SchemaType.OBJECT,
+    title = "Iceberg table global state",
+    description =
+        "Represents the global state of an Iceberg table in Nessie. An Iceberg table is globally "
+            + "identified via its unique 'Contents.id'.\n"
+            + "\n"
+            + "A Nessie commit-operation, performed via 'TreeApi.commitMultipleOperations', for Iceberg "
+            + "for Iceberg consists of a 'Operation.Put' with an 'IcebergTable' and an "
+            + "'IcebergTableGlobal' as the expected-global-state via 'Operation.PutGlobal' using the "
+            + "same 'ContentsKey' and 'Contents.id'.\n"
+            + "\n"
+            + "During a commit-operation, Nessie checks whether the known global state of the "
+            + "Iceberg table is compatible (think: equal) to 'Operation.PutGlobal.expectedContents'.")
 @Value.Immutable(prehash = true)
 @JsonSerialize(as = ImmutableIcebergTable.class)
 @JsonDeserialize(as = ImmutableIcebergTable.class)
 @JsonTypeName("ICEBERG_TABLE")
 public abstract class IcebergTable extends Contents {
 
+  /**
+   * Location where Iceberg stored its {@code TableMetadata} file. The location depends on the
+   * (implementation of) Iceberg's {@code FileIO} configured for the particular Iceberg table.
+   */
   @NotNull
   @NotBlank
   public abstract String getMetadataLocation();
 
+  /** ID of the current Iceberg snapshot. This value is not present, if there is no snapshot. */
+  @Nullable // TODO to be removed with PR#1923
+  public abstract Long getSnapshotId();
+
+  // TODO to be removed with PR#1923
   public static IcebergTable of(String metadataLocation) {
     return ImmutableIcebergTable.builder().metadataLocation(metadataLocation).build();
+  }
+
+  public static IcebergTable of(String metadataLocation, long snapshotId) {
+    return ImmutableIcebergTable.builder()
+        .metadataLocation(metadataLocation)
+        .snapshotId(snapshotId)
+        .build();
+  }
+
+  public static IcebergTable of(String metadataLocation, long snapshotId, String contentsId) {
+    return ImmutableIcebergTable.builder()
+        .metadataLocation(metadataLocation)
+        .snapshotId(snapshotId)
+        .id(contentsId)
+        .build();
   }
 }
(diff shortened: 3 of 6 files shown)


model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -571,6 +572,85 @@ void mergeRefIntoBranch(
...
+        }),
+    @APIResponse(responseCode = "400", description = "Invalid input, ref/hash name not valid"),
+    @APIResponse(responseCode = "401", description = "Invalid credentials provided"),
+    @APIResponse(
+        responseCode = "403",
+        description = "Not allowed to view the given reference or perform commits"),
+    @APIResponse(responseCode = "404", description = "Provided ref doesn't exists"),
+    @APIResponse(responseCode = "409", description = "Update conflict")
+  })
+  Branch commitMultipleOperationsWithGlobalState(
```
dimas-b, line 615: Could we handle this as a `default` method and redirect to `commitMultipleOperations`? It looks like `Operations` is able to contain both per-ref and global operations at the same time... Are we going to have different processing logic in this API method?

model/src/main/java/org/projectnessie/model/CompleteState.java
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model;
+
+/**
+ * Marker interface for {@link Contents} objects that represent both the global and per-reference
```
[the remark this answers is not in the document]
snazy: true, updated the comment

model/src/main/java/org/projectnessie/model/CompleteState.java
```diff
@@ -0,0 +1,31 @@
...
+ * org.projectnessie.api.ContentsApi#getContents(ContentsKey, String, String)} and {@link
+ * org.projectnessie.api.ContentsApi#getMultipleContents(String, String, MultiGetContentsRequest)}.
+ *
+ * @param <G> {@link GlobalState} object type
+ * @param <R> {@link OnReferenceState} object type
+ */
+public interface CompleteState<G extends Contents, R extends Contents> {
+  R extractRefState();
+
+  G extractGlobalState();
```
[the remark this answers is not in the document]
snazy: renamed

model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -56,18 +72,51 @@
...
+
+    /**
+     * When present, this {@code Put} operation will be a CAS (compare-and-swap) operation based on
+     * the expected global state. Only useful if {@link #getContents()} represents a {@link
+     * OnReferenceState}.
+     */
+    @Nullable
+    Contents getExpectedState();
+
+    static PutGlobal of(ContentsKey key, Contents contents, Contents expectedState) {
```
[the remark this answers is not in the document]
snazy: yea - better safe than sorry.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
snazy: Yes. those will be a little bit different. `commitMultipleOperations` only takes contents using the current (aka Nessie API 0.9) semantics w/o global state. `commitMultipleOperationsWithGlobalState` expects Nessie API v1 semantics w/ per-reference state in `Put` and global-state via `PutGlobal`.

It's a bit tricky to untangle the object hierarchy here for several reasons:
* type-parameters in {@code StoreWorker} and everything that depends on it. adding another type-parameter adds a ton of noise.
* backwards-compatibility of the API makes it necessary to rely on the most common-type (`Contents`)
* content-types for delta-lake + sql-view do not have a global state, therefore do not extend `OnReferenceState`, which expects the global state type

dimas-b: Right, but if we simply validate that in v0.9 API we do not have `PutGlobal` operations (and error out if we do), will the subsequent processing not be the same?

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
```diff
@@ -47,7 +48,87 @@
...
+                examples = {@ExampleObject(ref = "iceberg")},
+                schema = @Schema(implementation = Contents.class))),
+    @APIResponse(responseCode = "400", description = "Invalid input, ref name not valid"),
+    @APIResponse(responseCode = "401", description = "Invalid credentials provided"),
+    @APIResponse(
+        responseCode = "403",
+        description = "Not allowed to view the given reference or read object content for a key"),
+    @APIResponse(responseCode = "404", description = "Table not found on ref")
+  })
+  Contents getContentsWithGlobalState(
```
rymurr, line 95: why have two versions of get Contents? They seem the same to me? Why does the client care if its global or not

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: same comment as before, why does the end user care about the global part of some contents.

model/src/main/java/org/projectnessie/model/Contents.java
```diff
@@ -30,22 +30,52 @@
 @Schema(
     type = SchemaType.OBJECT,
     title = "Contents",
-    oneOf = {IcebergTable.class, DeltaLakeTable.class, SqlView.class},
+    oneOf = {
+      IcebergTable.class,
+      IcebergSnapshot.class,
```
rymurr, line 35: why have separate API objects to track the global vs local state? I see how we would want that distinction inside the version store but to me this is leaking version store implemention details to the user.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: additionally, I don't think we should make any allowances for old pre-1.0 apis. We should remove them from 1.0 if at all possible. Clean break from pre-1.0

rymurr (changes requested):
To me we are leaking a lot of implementation details out to the REST API consumer w/ this change. Can we remove global state from the API and only have it as an internal VersionStore concern?

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
snazy: Backwards compatibilty

model/src/main/java/org/projectnessie/model/Contents.java
snazy: `Put` operations contain the per-reference state - the iceberg-snapshot-id.
`PutGlobal` operations contain the global-state (iceberg-table-metadata) to set and the expected global-state

A single `Put` would otherwise leak global-state "logic" into per-reference-state-only types (deltalake + sql-views)

There are two options:
* `Put` w/ per-reference state + `PutGlobal` w/ global-state + expected-state or
* `Put` w/ a composite object and expected global-state. IIRC we discarded this option of a `Put` "with everything" a couple weeks ago.

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
rymurr: with what? I don't think we should try to maintain backwards compatibility now that our last pre-1.0 release is out

model/src/main/java/org/projectnessie/model/Contents.java
rymurr: I had hoped that `IcebergTable` would be `{Long snapshotId, String metadataLocation}` and the REST API or the version store would split that into a global change and a local change. When performing a get we then jsut return the iceberg table. To me the user shouldn't see global state at all and there should be one iceberg related Contents. IMHO global state is purely an artifact of the kernel and not somethign a user should care about.

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
snazy: Okay, that's new information.

model/src/main/java/org/projectnessie/model/Contents.java
snazy: That was actually the initial approach (adding `snapshotId` to `IcebergTable`).
Anyway, I've updated this PR accordingly. It's rather simple now (see PR description).

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -18,21 +18,80 @@
...
 
+/**
+ * Represents the state of an Iceberg table in Nessie. An Iceberg table is globally identified via
+ * its {@link Contents#getId() unique ID}.
+ *
+ * <p>The Iceberg-table-state consists of the location to the table-metadata and the snapshot-ID.
+ *
+ * <p>The table-metadata-location is managed globally within Nessie, which means that all versions
+ * of the same table share the table-metadata across all named-references (branches and tags). There
+ * is only one version, the current version, of Iceberg's table-metadata.
```
dimas-b, line 36: nit: Would it make sense to emphasize that the global metadata is conceptually cumulative (contains information that applies equally to all tables, current and historic)?

model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -48,6 +49,14 @@
...
   ContentsKey getKey();
 
+  @Schema(
+      type = SchemaType.OBJECT,
+      title = "Put-'Contents'-operation for a 'ContentsKey'.",
+      description =
+          "Add or replace (put) a 'Contents' object for a 'ContentsKey'. "
+              + "If the actual table type tracks the 'global state' of individual tables (Iceberg "
+              + "as of today), every 'Put' must be accompanied by a 'PutGlobal' to update the global "
+              + "state.")
```
dimas-b, line 59: If I understand correctly, for an Iceberg `Put` operation `Contents` would be an instance of `IcebergTable`, right? Since, `IcebergTable` already tracks both global and per-branch state, why do we require a separate `PutGlobal` operation? Could we not split the per-branch data and global data on the server side when we're about to store it? :thinking:

model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -56,9 +65,20 @@
     @NotNull
     Contents getContents();
 
+    @Nullable
+    Contents getExpectedContents();
```
rymurr, line 69: what does this do?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -100,7 +100,6 @@ public Reference getReferenceByName(String refName) throws NessieNotFoundExcepti
   @Override
   public Reference createReference(@Nullable String sourceRefName, Reference reference)
       throws NessieNotFoundException, NessieConflictException {
-    // TODO remove after Iceberg > 0.12
```
rymurr, line 103: what changed that this comment can be removed?

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
rymurr: do you disagree? Looking at these changes it seems like it would be hard to support both pre-1.0 and 1.0 apis well.

It means Iceberg wont be able to use Nessie > 1,0 until the next release of Iceberg, which is a bummer but do you think it can be avoided?

rymurr: Lets discuss this today...we should try and keep backwards compatibility if its not too onerous

model/src/main/java/org/projectnessie/model/Operation.java
snazy: > hy do we require a separate PutGlobal operation? 

There is no PutGlobal operation (see updated PR description).

> Could we not split the per-branch data and global data on the server side when we're about to store it?

That's what will happen - split on the server side.

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: Not sure though. It may change in the future.

model/src/main/java/org/projectnessie/model/Operation.java
snazy: That's the `if` for the global-state-CAS-operation (think: "UPDATE global_state SET value = contents.table_metadata IF value = expected_contents.table_metadata")

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
snazy: We always default to the default branch now. Nothing specific to this PR though - slipped though.

model/src/main/java/org/projectnessie/api/http/HttpContentsApi.java
snazy: It's confusing: _"I don't think we should try to maintain backwards compatibility"_ vs _"we should try and keep backwards compatibility"_

rymurr: yes it is confusing. How do we make it less confusing? What do you think in terms of backwards compatibility? 

For me there are competing concerns:
* Part of me doesn't think we should expend the effort as we are explicitly going to 1.0 and have no compatibiliy guarantees. 
* Im also concerned that Iceberg and other tools can't use Nessie 1.0 until their next releases (possibly months?)
* Iceberg in particular probably shoudln't use pre-1.0 Nessie since we can't do migrations of that data structure

snazy: I think, a "clear cut" for 1.0 is the better way forwards, because probably nobody is using Nessie in production yet. The argument that current Iceberg (0.12.0) wouldn't be able to run against Nessie 1.0 is legit though - couple options there:
- either we can ask to release Iceberg 0.12.1 or (0.13.0?) w/ Nessie 1.0
- release a custom artifact `org.projectnessie:iceberg-nessie:0.12.1` for the time until Iceberg 0.13.0

Ensuring backwards compatibility forces us to add a bunch of complexity to the code base (think: `commitMultipleOperationsNessie09`, `getContentsNessie09`, a Nessie 0.9 version of `IcebergTable`, a Nessie 1.0 version of `IcebergTable`). It may also mean that people continue to use Iceberg 0.12 w/ Nessie 1.0 but actually risking data corruption, because that doesn't know about global-states.

model/src/main/java/org/projectnessie/model/Operation.java
rymurr: the doc is out of date

model/src/main/java/org/projectnessie/model/Operation.java
rymurr: so if null it doesn't check? or if null and set it will fail?

Mildly annoying that we have to send this whole object back but prob no easy way around it. Thoughts?

snazy: Yea, it's the easiest way for tools: just send back what you got. Otherwise we'd have to introduce a new type just for the global-state + conversion/comparison functionality.

If it's null, it doesn't check, which can be used like a "git push --force". It must be null for non-global-state-content-types (Deltalake) and also when you've got no previous global-state (think: a new table). Sounds more complicated than it actually is (it's really just passing the existing field `NessieTableOperations.table` as the expectedValue).

model/src/main/java/org/projectnessie/model/Operation.java
snazy: argh - fixed

model/src/main/java/org/projectnessie/model/Operation.java
rymurr: Can we add some javadoc somewhere to that effect? I think it would be weird to reject a DeltaLake op if its non-null and have no clear reason why.

snazy: I've added a TODO to re-think that behavior in the `PersistVersionStore` (no PR yet)

model/src/main/java/org/projectnessie/model/IcebergTable.java
dimas-b: However, even in general from Nessies' perspective it probably only makes sense to track something as "global" state only if it is cumulative in nature, right? Otherwise, it'd be part of the change/commit log, would it not?

dimas-b (approved):

rymurr (approved):

PR #1930, opened 2021-09-09
Make from-ref-name optional, because old Nessie-API clients do not support it

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1930)


model/src/main/java/org/projectnessie/model/Transplant.java
```diff
@@ -46,7 +46,8 @@
   @Size(min = 1)
   List<String> getHashesToTransplant();
 
-  @NotBlank
+  @Nullable
+  @Size(min = 1) // TODO migrate to @NotBlank once all Nessie-0.9-API clients are "gone"
```
rymurr, line 50: can you make an issue also

snazy: opened #1931

rymurr (approved):

PR #1932, opened 2021-09-09
Fix dependency issues for `:nessie-deltalake` via `:nessie-client-tests`

snazy:
Tests _partly_ use `client.nessie.version` (the one that uses `:nessie-client-tests`, some use `project.version`.
Changing the pom to use `client.nessie.version` for the whole module.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1932)

### DIFF
--- a/clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
+++ b/clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
@@ -29,18 +29,24 @@ import org.apache.spark.sql.Encoders;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.functions;
 import org.apache.spark.util.Utils;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
 import org.junit.jupiter.api.io.TempDir;
+import org.projectnessie.client.api.NessieApiV1;
+import org.projectnessie.client.api.NessieApiVersion;
+import org.projectnessie.client.http.HttpClientBuilder;
 import org.projectnessie.client.tests.AbstractSparkTest;
 import org.projectnessie.model.Branch;
-import org.projectnessie.model.ImmutableMerge;
 import org.projectnessie.model.Reference;
 import scala.Tuple2;
 
 class ITDeltaLog extends AbstractSparkTest {
+
+  NessieApiV1 api;
 
   @TempDir File tempPath;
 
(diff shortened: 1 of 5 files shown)


clients/deltalake/pom.xml
```diff
@@ -30,15 +30,10 @@
...
   <dependencies>
-    <dependency>
-      <groupId>org.projectnessie</groupId>
-      <artifactId>nessie-model</artifactId>
-      <version>${project.version}</version>
-    </dependency>
     <dependency>
       <groupId>org.projectnessie</groupId>
       <artifactId>nessie-client</artifactId>
-      <version>${project.version}</version>
```
rymurr, line 41: we want to use the up to date version though don't we?

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLog.java
```diff
@@ -52,6 +58,17 @@ protected static void createDelta() {
...
   }
 
+  @BeforeEach
+  void createClient() {
+    api = HttpClientBuilder.builder().withUri(url).build(NessieApiVersion.V_1, NessieApiV1.class);
+  }
+
+  @AfterEach
+  void closeClient() {
+    api.close();
```
nastra, line 68: maybe add a null check in case api init failed

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
```diff
@@ -45,7 +47,7 @@
 
 class ITDeltaLogBranches extends AbstractSparkTest {
 
-  private NessieClient client;
+  static NessieApiV1 api;
```
nastra, line 50: nit: does this need to be static?

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
```diff
@@ -58,23 +60,23 @@ protected static void createDelta() {
...
+    api = HttpClientBuilder.builder().withUri(url).build(NessieApiVersion.V_1, NessieApiV1.class);
   }
 
   @AfterEach
-  public void closeClient() throws NessieNotFoundException, NessieConflictException {
+  void closeClient() throws NessieNotFoundException, NessieConflictException {
     Reference ref = null;
     try {
-      ref = client.getTreeApi().getReferenceByName("test");
+      ref = api.getReference().refName("test").submit();
```
nastra, line 71: do we need to do the same cleanup stuff in the other classes?

clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
```diff
@@ -60,6 +62,7 @@ protected static void createDelta() {
 
   @BeforeEach
   void getHash() throws NessieNotFoundException {
+    nessieClient = NessieClient.builder().withUri(url).build();
```
nastra, line 65: why not also use the same Nessie API as in the other classes here?

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
snazy: Don't think so

clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
snazy: It's using "old Nessie" here

nastra (approved):
LGTM but PR description needs to be deleted/updated to reflect the new changes :)

PR #1934, opened 2021-09-09
New rest endpoint for partial namespaces in the UI

rymurr:
This new endpoint is primarily used by the UI to only return the required namespace for a view. This simplifies the UI and reduces the number and weight of API calls by the UI.

For example keys:
`a.b.c.d`
`a.b.e`
`a.f`

we can use the combination of prefix and depth to return:
`("",1) -> a`
`("a",2) -> a.b,a.f`
etc

Notes:
* I implemented this simplistically w/o using EntriesParams or similar. We could make this more complicated and generic but I don't ahve immediate use cases for that in the UI
* We perhaps want to use the CEL expression directly rather than generating it in code if we want more complex filters than just the namespace depth
* I haven't exposed this to Java or Python clients as its really a UI concern.
* this doesn't work with paging so I have disabled paging. We are assuming the single depth of namespace should be small-ish

Please weigh in on the above if you disagree!

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1934)

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
@@ -49,7 +49,7 @@ final class HttpGetEntries extends BaseHttpOnReferenceRequest<GetEntriesBuilder>
   }
 
   @Override
-  public GetEntriesBuilder queryExpression(Integer namespaceDepth) {
+  public GetEntriesBuilder namespaceDepth(Integer namespaceDepth) {
     params.namespaceDepth(namespaceDepth);
     return this;
   }
--- a/clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
@@ -49,6 +49,12 @@ final class HttpGetEntries extends BaseHttpOnReferenceRequest<GetEntriesBuilder>
   }
 
   @Override
+  public GetEntriesBuilder queryExpression(Integer namespaceDepth) {
+    params.namespaceDepth(namespaceDepth);
+    return this;
+  }
+
+  @Override
   public EntriesResponse submit() throws NessieNotFoundException {
     params.hashOnRef(hashOnRef);
     return client.getTreeApi().getEntries(refName, params.build());
--- a/model/src/main/java/org/projectnessie/api/TreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/TreeApi.java
@@ -95,6 +95,15 @@ public interface TreeApi {
       @NotNull @Valid EntriesParams params)
       throws NessieNotFoundException;
 
+  EntriesResponse getNamespaceEntries(
+      @NotNull @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
+          String refName,
+      @Nullable @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
+          String hashOnRef,
+      @Nullable String namespacePrefix,
+      @Nullable Integer depth)
+      throws NessieNotFoundException;
+
   /**
    * Retrieve the commit log for a ref, potentially truncated by the backend.
    *
--- a/model/src/main/java/org/projectnessie/api/TreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/TreeApi.java
@@ -25,6 +25,7 @@ import org.projectnessie.api.params.EntriesParams;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
+import org.projectnessie.model.ContentsKey;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.LogResponse;
 import org.projectnessie.model.Merge;
@@ -100,7 +101,7 @@ public interface TreeApi {
           String refName,
       @Nullable @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
           String hashOnRef,
-      @Nullable String namespacePrefix,
+      @Nullable ContentsKey namespacePrefix,
       @Nullable Integer depth)
       throws NessieNotFoundException;
 
--- a/model/src/main/java/org/projectnessie/api/TreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/TreeApi.java
@@ -25,7 +25,6 @@ import org.projectnessie.api.params.EntriesParams;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
-import org.projectnessie.model.ContentsKey;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.LogResponse;
 import org.projectnessie.model.Merge;
@@ -94,15 +93,6 @@ public interface TreeApi {
       @NotNull @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
           String refName,
       @NotNull @Valid EntriesParams params)
-      throws NessieNotFoundException;
-
-  EntriesResponse getNamespaceEntries(
-      @NotNull @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
-          String refName,
-      @Nullable @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
-          String hashOnRef,
-      @Nullable ContentsKey namespacePrefix,
-      @Nullable Integer depth)
       throws NessieNotFoundException;
 
   /**
--- a/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
@@ -46,7 +46,6 @@ import org.projectnessie.api.params.EntriesParams;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
-import org.projectnessie.model.ContentsKey;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.LogResponse;
 import org.projectnessie.model.Merge;
@@ -229,7 +228,11 @@ public interface HttpTreeApi extends TreeApi {
               + "will return all commit log entries.\n"
               + "\n"
               + "The 'query_expression' parameter allows for advanced filtering capabilities using the Common Expression Language (CEL).\n"
-              + "An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n")
+              + "An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n"
+              + "\n"
+              + "The 'namespaceDepth' parameter returns only the ContentsKey components up to the depth of 'namespaceDepth'.\n"
+              + "For example they key 'a.b.c.d' with a depth of 3 will return 'a.b.c'. The operation is guaranteed to not return \n"
+              + "duplicates and therefore will never page.")
   @APIResponses({
     @APIResponse(
         description = "all objects for a reference",
@@ -256,73 +259,6 @@ public interface HttpTreeApi extends TreeApi {
           @PathParam("ref")
           String refName,
       @NotNull @Valid @BeanParam EntriesParams params)
-      throws NessieNotFoundException;
-
-  /**
-   * Retrieve namespace and objects for a ref.
-   *
-   * <p>Retrieves objects and namespaces for the given named reference (tag or branch) or the given
-   * {@link org.projectnessie.model.Hash}.
-   *
-   * <p>This is used to primarily to display the elements at a given level of a namespace.
-   *
-   * <p>This method does not page as it must guarantee uniqueness.
-   *
-   * <p>Example with objects: a.b.c.table1 a.b.table2 a.table3
-   *
-   * <p>With depth 1 we return a only with depth 2 and namespacePrefix a we return a.b and a.table3
-   * with depth 3 and namespacePrefix a.b we return a.b.c and a.b.table2
-   */
-  @GET
-  @Produces(MediaType.APPLICATION_JSON)
-  @Path("tree/{ref}/namespaces")
-  @Operation(
-      summary = "Fetch all namespaces and objects for a given depth and prefix",
-      description =
-          "Retrieves objects for a ref, this won't be paged due to the requirement for uniqueness.\n")
-  @APIResponses({
-    @APIResponse(
-        description =
-            "all unique objects and namespaces for a reference at a certain depth and prefix",
-        content = {
-          @Content(
-              mediaType = MediaType.APPLICATION_JSON,
-              examples = {@ExampleObject(ref = "entriesResponse")},
-              schema = @Schema(implementation = EntriesResponse.class))
-        }),
-    @APIResponse(responseCode = "200", description = "Returned successfully."),
-    @APIResponse(responseCode = "400", description = "Invalid input, ref name not valid"),
-    @APIResponse(responseCode = "401", description = "Invalid credentials provided"),
-    @APIResponse(
-        responseCode = "403",
-        description = "Not allowed to view the given reference or fetch entries for it"),
-    @APIResponse(responseCode = "404", description = "Ref not found")
-  })
-  EntriesResponse getNamespaceEntries(
-      @NotNull
-          @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
-          @Parameter(
-              description = "name of ref to fetch from",
-              examples = {@ExampleObject(ref = "ref")})
-          @PathParam("ref")
-          String refName,
-      @Nullable
-          @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
-          @Parameter(
-              description = "a particular hash on the given ref",
-              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
-          @QueryParam("hashOnRef")
-          String hashOnRef,
-      @Nullable
-          @Parameter(
-              description = "prefix namespace to start at",
-              examples = {@ExampleObject(ref = "namespace")})
-          @QueryParam("namespace")
-          ContentsKey namespacePrefix,
-      @Nullable
-          @Parameter(description = "total depth of namesapce to fetch")
-          @QueryParam("namespaceDepth")
-          Integer depth)
       throws NessieNotFoundException;
 
   /**
--- a/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
@@ -258,6 +258,79 @@ public interface HttpTreeApi extends TreeApi {
       throws NessieNotFoundException;
 
   /**
+   * Retrieve objects for a ref, potentially truncated by the backend.
+   *
+   * <p>Retrieves up to {@code maxRecords} objects for the given named reference (tag or branch) or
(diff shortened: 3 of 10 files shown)


servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +336,50 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+  }
+
+  EntriesResponse.Entry truncate(EntriesResponse.Entry entry, Integer depth) {
+    if (depth == null || depth < 1) {
+      return entry;
+    }
+    Type type = entry.getName().getElements().size() > depth ? Type.UNKNOWN : entry.getType();
+    ContentsKey key = ContentsKey.of(entry.getName().getElements().subList(0, depth));
+    return EntriesResponse.Entry.builder().type(type).name(key).build();
+  }
```
snazy, line 382: Nit: newline after closing bracket

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -257,6 +257,79 @@ EntriesResponse getEntries(
...
+              description = "a particular hash on the given ref",
+              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
+          @QueryParam("hashOnRef")
+          String hashOnRef,
+      @Nullable
+          @Parameter(
+              description = "prefix namespace to start at",
+              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
+          @QueryParam("namespace")
+          String namespacePrefix,
```
snazy, line 324: I think, this should this be a `ContentsKey` instead of `String.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -257,6 +257,79 @@ EntriesResponse getEntries(
...
+          @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
+          @Parameter(
+              description = "a particular hash on the given ref",
+              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
+          @QueryParam("hashOnRef")
+          String hashOnRef,
+      @Nullable
+          @Parameter(
+              description = "prefix namespace to start at",
+              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
```
snazy, line 322: hash? copy-paste?

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -257,6 +257,79 @@ EntriesResponse getEntries(
...
+          @QueryParam("namespace")
+          String namespacePrefix,
+      @Nullable
+          @Parameter(
+              description = "total depth of namesapce to fetch",
+              examples = {@ExampleObject(ref = "nullHash"), @ExampleObject(ref = "hash")})
```
snazy, line 328: more copy-pasted-hashes

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -257,6 +257,79 @@ EntriesResponse getEntries(
       @NotNull @Valid @BeanParam EntriesParams params)
       throws NessieNotFoundException;
 
+  /**
+   * Retrieve objects for a ref, potentially truncated by the backend.
```
snazy, line 261: The doc doesn't seem to match the functionality

model/src/main/java/org/projectnessie/api/TreeApi.java
```diff
@@ -95,6 +95,15 @@ EntriesResponse getEntries(
       @NotNull @Valid EntriesParams params)
       throws NessieNotFoundException;
 
+  EntriesResponse getNamespaceEntries(
```
snazy, line 98: No need to add it to `TreeApi`

snazy (changes requested):

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +336,50 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+  public EntriesResponse getNamespaceEntries(
+      String refName,
+      @Nullable String hashOnRef,
+      @Nullable String namespacePrefix,
+      @Nullable Integer depth)
+      throws NessieNotFoundException {
+    WithHash<NamedRef> refWithHash = namedRefWithHashOrThrow(refName, hashOnRef);
+    try {
+      List<EntriesResponse.Entry> entries;
+      try (Stream<EntriesResponse.Entry> s =
```
omarsmak, line 349: Would it make sense to change `s` to a more meaningful name, let's say `stream`? Not a big deal though

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +336,50 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+              .map(
+                  key ->
+                      EntriesResponse.Entry.builder()
+                          .name(fromKey(key.getValue()))
+                          .type((Type) key.getType())
+                          .build())) {
+        entries =
+            filterEntries(
+                    s,
+                    (namespacePrefix == null || namespacePrefix.isEmpty())
```
omarsmak, line 361: I am thinking, would it be more efficient to create a utility function that can be used across other classes to check if the value is null first and then check emptiness for every type instead of doing the double check everytime? e.g `isEmpty(Object value)`. If you guys think it make sense, I would happy to add it as my main contribution to Nessie :D

model/src/main/java/org/projectnessie/api/TreeApi.java
dimas-b: +1 to not changing legacy API in this case. The UI should be able to use v1 API, I assume.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: done

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: done

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: i didn't do a lot of docs, wanted an opinion on the concept first. Fixed up now.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: i didn't do a lot of docs, wanted an opinion on the concept first. Fixed up now.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
rymurr: i didn't do a lot of docs, wanted an opinion on the concept first. Fixed up now.

model/src/main/java/org/projectnessie/api/TreeApi.java
rymurr: why? Then it isn't in the client and I can't use it in tests.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: done

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: I changed the condition a bit so probably not as relevant, but happy  to have some `Preconditions` like utilities in the codebase

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -940,6 +940,71 @@ public void filterEntriesByNamespace() throws NessieConflictException, NessieNot
...
+    entries = tree.getNamespaceEntries(branch, null, "a", 0).getEntries();
+    assertThat(entries).isNotNull().hasSize(4);
+
+    entries = tree.getNamespaceEntries(branch, null, "a", 1).getEntries();
+    assertThat(entries).hasSize(1);
+    entries.forEach(e -> assertThat(e.getName().toPathString()).startsWith("a"));
+
+    entries = tree.getNamespaceEntries(branch, null, "a", 2).getEntries();
+    assertThat(entries).hasSize(2);
+    entries.forEach(e -> assertThat(e.getName().toPathString()).startsWith("a"));
```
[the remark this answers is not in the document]
rymurr: changed the assertions aroudn a bit. Better?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +336,50 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+                        ? null
+                        : String.format("entry.namespace.startsWith('%s')", namespacePrefix))
```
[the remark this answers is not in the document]
rymurr: kept CEL but am using a regex now.

omarsmak (dismissed):

model/src/main/java/org/projectnessie/api/TreeApi.java
dimas-b: Can't test use v1 API?

dimas-b: I do not mind adding a method to the old API as much as I'm wondering why it is necessary to add it to the old API rather than to v1 API :thinking:

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -257,6 +258,73 @@ EntriesResponse getEntries(
...
 
+  /**
+   * Retrieve namespace and objects for a ref.
+   *
+   * <p>Retrieves objects and namespaces for the given named reference (tag or branch) or the given
+   * {@link org.projectnessie.model.Hash}.
+   *
+   * <p>This is used to primarily to display the elements at a given level of a namespace.
+   *
+   * <p>This method does not page as it must guarantee uniqueness.
```
dimas-b, line 269: Is it a concern with potentially large sets of object names?

dimas-b: Would it be worth adding a limit parameter (and a max limit guardrail config)?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +337,56 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+    } catch (ReferenceNotFoundException e) {
+      throw new NessieNotFoundException(
+          String.format("Unable to find the reference [%s].", refName), e);
+    }
+  }
+
+  private String buildRegex(ContentsKey namespacePrefix) {
+    return String.format(
+        "entry.namespace.matches('%s(\\\\.|$)')",
+        Joiner.on("\\\\.").join(namespacePrefix.getElements()));
```
dimas-b, line 379: Should we also regex-quote each element. Are names allowed to contain special chars?

dimas-b: Would it be more robust to have a custom prefix-matching filter in java?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +337,56 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+                      EntriesResponse.Entry.builder()
+                          .name(fromKey(key.getValue()))
+                          .type((Type) key.getType())
+                          .build())) {
+        entries =
+            filterEntries(
+                    keyStream,
+                    (namespacePrefix == null || namespacePrefix.getElements().isEmpty())
+                        ? null
+                        : buildRegex(namespacePrefix))
```
harshm-dremio, line 364: A. Do we have to restrict to custom prefix use-case only? Why not take regex as an input from the ReST API instead.
B. Is it possible to pushdown the filter to the underlying store/DB adapter (alternatively)?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -335,6 +337,56 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+                .map(e -> truncate(e, depth))
```
harshm-dremio, line 365: Can't we construct the regex in a way, that it takes care of depth (matches only if the number of separators= depth)? If yes, a secondary truncate isn't necessary.

rymurr: no, unfortunately. There are no namespace objects so there is no `a.b` only `a.b.table` so we always have to truncate the `table` part when searching for `a.b`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: A. Ive been thinking that we should probably integrate this w/ the `getEntries` api endpoint. We would only have to add a depth parameter to the API then. Filtering would only happen if depth > 0.

Thoughts? cc @snazy @dimas-b

rymurr: B. that would be a goal for a later PR

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Sorry for confusing, I meant use of Matcher.group(..) ; but this version is also fine.

harshm-dremio:
LGTM

omarsmak (dismissed):

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
dimas-b: A: from my POV it makes sense to add an optional `depth` parameter to `getEntries`.

I'm not sure about exposing a RegEx filter to the end user is worth the effort. I suppose the main use case is hierarchical traversal, so depth alone should be sufficient. Depth should also be easier to push down to storage level  than a generic RegEx, I guess.

model/src/main/java/org/projectnessie/api/params/EntriesParams.java
```diff
@@ -50,6 +50,12 @@
   @QueryParam("pageToken")
   private String pageToken;
 
+  @Parameter(
+      description =
+          "If set > 0 will filter the results to only return namespaces/tables to the depth of namespaceDepth. If not set or <=0 has no effect")
```
dimas-b, line 55: nit: Should we also mention that paging is not supported if depth is set?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -325,8 +325,17 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
...
+        Stream<EntriesResponse.Entry> entriesStream =
+            filterEntries(entryStream, params.queryExpression());
+        if (params.namespaceDepth() != null && params.namespaceDepth() > 0) {
+          entries =
+              entriesStream
+                  .map(e -> truncate(e, params.namespaceDepth()))
+                  .distinct()
+                  .collect(ImmutableList.toImmutableList());
+        } else {
+          entries = entriesStream.collect(ImmutableList.toImmutableList());
```
dimas-b, line 337: Do we respect `EntriesParams.maxRecords`? Should we document that it does not apply in this case?

Same concern about `pageToken`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -940,6 +940,122 @@ public void filterEntriesByNamespace() throws NessieConflictException, NessieNot
...
+  public void filterEntriesByNamespaceAndPrefixDepth()
+      throws NessieConflictException, NessieNotFoundException {
+    final String branch = "filterEntriesByNamespaceAndPrefixDepth";
+    Reference r = tree.createReference("main", Branch.of(branch, null));
+    ContentsKey first = ContentsKey.of("a", "b", "c", "firstTable");
+    ContentsKey second = ContentsKey.of("a", "b", "c", "secondTable");
+    ContentsKey third = ContentsKey.of("a", "thirdTable");
+    ContentsKey fourth = ContentsKey.of("a", "b", "fourthTable");
+    ContentsKey fifth = ContentsKey.of("a", "boo", "fifthTable");
+    tree.commitMultipleOperations(
```
snazy, line 953: Better put those into a `for (int i = 1; i <= 5; i++)` loop (less code)

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -325,8 +325,17 @@ public EntriesResponse getEntries(String namedRef, EntriesParams params)
                           .name(fromKey(key.getValue()))
                           .type((Type) key.getType())
                           .build())) {
-        entries =
-            filterEntries(s, params.queryExpression()).collect(ImmutableList.toImmutableList());
+        Stream<EntriesResponse.Entry> entriesStream =
+            filterEntries(entryStream, params.queryExpression());
+        if (params.namespaceDepth() != null && params.namespaceDepth() > 0) {
+          entries =
```
snazy, line 331: nit: this can be:
```
entriesStream = entriesStream
                  .map(e -> truncate(e, params.namespaceDepth()))
                  .distinct();
```
and remove the `else`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
snazy: no newline there?

model/src/main/java/org/projectnessie/api/params/EntriesParams.java
rymurr: yes, added

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: There is a block of todo above regarding paging. I would rather not modify any of the paging specific stuff in `getEntries` and keep it to the depth token for this PR.

clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
```diff
@@ -48,6 +48,12 @@ public GetEntriesBuilder queryExpression(String queryExpression) {
     return this;
   }
 
+  @Override
+  public GetEntriesBuilder queryExpression(Integer namespaceDepth) {
```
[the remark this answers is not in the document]
rymurr: can you give more context to your comment?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
rymurr: done

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: done

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
rymurr: spotless keeps eating it

clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpGetEntries.java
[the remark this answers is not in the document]
snazy: method name's wrong ;)

snazy (approved):

dimas-b (approved):

PR #1943, opened 2021-09-10
Add MongoDatabaseAdapter

dimas-b:
Using naive keyPrefix handling for now.


versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    return "MongoDB";
+  }
+
+  @Override
+  public Builder<MongoDatabaseAdapterConfig> newBuilder() {
+    return new Builder<MongoDatabaseAdapterConfig>() {
+      @Override
+      protected MongoDatabaseAdapterConfig getDefaultConfig() {
+        return ImmutableMongoDatabaseAdapterConfig.builder()
+            .connectionString("unknown")
```
snazy, line 34: why do we need those `unknown`s?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,416 @@
...
+  protected MongoDatabaseAdapter(MongoDatabaseAdapterConfig config) {
+    super(config);
+
+    keyPrefix = ByteString.copyFromUtf8(config.getKeyPrefix());
+    globalPointerKey = keyPrefix.toByteArray();
+
+    ConnectionString cs = new ConnectionString(config.getConnectionString());
+    MongoClientSettings settings = MongoClientSettings.builder().applyConnectionString(cs).build();
+
+    mongoClient = MongoClients.create(settings);
```
snazy, line 87: `MongoDatabaseAdapterConfig` needs a functionality to provide a `MongoClient` instance to `MongoDatabaseAdapterConfig` (like `RocksDatabaseAdapter` does). I.e. if the `MongoDatabaseAdapterConfig` provides a non-null one, use that, otherwise create one. Probably better to wrap that `MongoClient` in a class similar to `RocksDbInstance`, which would also care about inquiring the collections (the `database.getCollection()` calls / keep the `MongoCollection` instances.

In other words: this constructor should be fast and cheap, if the MongoClient is provided via the config.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,416 @@
...
+public class MongoDatabaseAdapter
+    extends NonTransactionalDatabaseAdapter<MongoDatabaseAdapterConfig> {
+
+  private static final String GLOBAL_POINTER = "global_pointer";
+  private static final String GLOBAL_LOG = "global_log";
+  private static final String COMMIT_LOG = "commit_log";
+  private static final String KEY_LIST = "key_list";
+
+  private final ByteString keyPrefix;
+  private final byte[] globalPointerKey;
```
snazy, line 70: Can we make this a `String`?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,416 @@
...
+
+    super.initializeRepo(defaultBranchName);
+  }
+
+  @Override
+  public void close() {
+    mongoClient.close();
+  }
+
+  private byte[] toId(Hash id) {
```
snazy, line 110: Represent hashes as Strings in Mongo?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
harshm-dremio: possibly it's because of a security concern, since strings stay in memory, whereas arrays are cleared once code is done with it.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterFactory.java
dimas-b: I did not want to make the optional in the "immutable" config since they are actually required for any connection.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: I see. Did not realize this was the intention :)

However, why would we pass proper "configuration" settings to this class in this case? Why not always pass the "client" / "instance" object and let the factory configure it?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: We could use `String` of course, but it looks like all the `persist.proto` objects convert to byte arrays naturally (including keys). So I figured we could use a byte array for the global pointer too.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Sure, but I think it's going to be more processing work with the current `persist.proto` objects :thinking: I'll give it a try.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: It's not about the `byte[]` per se, it's to represent the `globalPointerKey` in a say "mongo db type friedly" way.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: It's really just about representing it using the types available in mongo. (note: the jdbc-like adapters use `VARCHAR` as well).

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterFactory.java
snazy: not if the connection-instance is passed ;)

dimas-b: Yeah, but then we have a logical conflict between different ways of defining the connection in the same config object, don't we?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Mongo uses binary serialization internally, AFAIK, and supports byte arrays, but I see the rationale for using a human-readable type for keys :+1:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Reworked to `String` in 1f3f1e4c3afa0f1266f7906e09a58fb549318f66

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Reworked to `String` in 1f3f1e4c3afa0f1266f7906e09a58fb549318f66

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: I'm thinking about splitting "config" and "database instance" method at the factoring level :hourglass:

dimas-b: I took an approach similar to `RocksDbInstance` in cc90e03fe60a625171481980ca86e887f73a7abe for now.

Still thinking about refactoring the config/factoring classes :thinking:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,32 @@
...
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapterConfig;
+
+@Value.Immutable(lazyhash = true)
+public interface MongoDatabaseAdapterConfig extends DatabaseAdapterConfig {
+
+  String getConnectionString();
+
+  String getDatabaseName();
+
+  @Value.Derived
```
snazy, line 28: I suspect this should be `@Value.Default` ("derived" means it's always built, "default" means its built when not explicitly set)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,401 @@
...
+    return loadById(client.getGlobalLog(), id, GlobalStateLogEntry::parseFrom);
+  }
+
+  @Override
+  protected List<GlobalStateLogEntry> fetchPageFromGlobalLog(
+      NonTransactionalOperationContext ctx, List<Hash> hashes) {
+    return fetchPage(client.getGlobalLog(), hashes, GlobalStateLogEntry::parseFrom);
+  }
+
+  @FunctionalInterface
```
snazy, line 397: Can you move this interface to `ProtoSerialization` so we can reuse it?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,401 @@
...
+    byte[] expectedGlobalId = expected.getGlobalId().toByteArray();
+
+    UpdateResult result =
+        client
+            .getGlobalPointers()
+            .updateOne(
+                Filters.and(Filters.eq(globalPointerKey), Filters.eq("globalId", expectedGlobalId)),
+                new Document("$set", doc));
+
+    return result.wasAcknowledged()
```
snazy, line 364: This CAS-operation LGTM from my understanding of MongoDB. But would be good if someone else could confirm.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,401 @@
...
+  }
+
+  @Override
+  protected CommitLogEntry fetchFromCommitLog(NonTransactionalOperationContext ctx, Hash hash) {
+    return loadById(client.getCommitLog(), hash, ProtoSerialization::protoToCommitLogEntry);
+  }
+
+  private Hash idAsHash(Document doc) {
+    String id = doc.get("_id", String.class);
+    String hash = id.substring(keyPrefix.length() + 1); // + 1 for the separator char
```
snazy, line 256: Better use String.indexOf() here instead of assuming the same key-prefix (unlikely that this might happen)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,401 @@
...
+   * Fetches a collection of documents by hash and returns them in the order of hashes requested.
+   *
+   * <p>Note: unknown hashes will correspond to {@code null} elements in the result list.
+   */
+  private <T> List<T> fetchMappedPage(
+      MongoCollection<Document> collection, List<Hash> hashes, Function<Document, T> mapper) {
+    List<String> ids = hashes.stream().map(this::toId).collect(Collectors.toList());
+    FindIterable<Document> docs = collection.find(Filters.in("_id", ids)).limit(hashes.size());
+
+    HashMap<Hash, Document> loaded = new HashMap<>(hashes.size());
```
snazy, line 215: The initial-capacity is wrong here. You need `hashes.size() * 4 / 3 + 1` here.

snazy (commented):
Mostly LGTM, few minor things

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
dimas-b: That's what I meant actually. `Derived` -> built on first get and cached. Since the constructor is trivial, I thought it'd make sense instead of forcing it to be created and set from the outside. WDYT?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Good point. Fixed in e6b4f6eebea447420690ab610184702adf28f57c

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: I've added some validation in e6b4f6eebea447420690ab610184702adf28f57c

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Moved: dbf2425824c655c50cc7d19bc8df6fe742afff59

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
snazy: Immutables takes care of that piece in the generated code.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: That's a lot ;)
I'm okay with that.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
dimas-b: Yes, so "derived" is created only on first access. Why is "default" preferable?

dimas-b: Made `Client` settable in d7a1e320814af38cdfd0b5d4c75487e841fd7de9

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,34 @@
...
+import org.immutables.value.Value;
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapterConfig;
+
+@Value.Immutable(lazyhash = true)
+public interface MongoDatabaseAdapterConfig extends DatabaseAdapterConfig {
+
+  String getConnectionString();
+
+  String getDatabaseName();
+
```
snazy, line 28: Just realized: need `withDatabaseName` + `withConnectionString` here as well.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterFactory.java
snazy: It having these defaults would prevent checks in `MongoDatabaseClient` whether the properties are set though, right?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -0,0 +1,415 @@
...
+      NonTransactionalOperationContext ctx,
+      GlobalStatePointer expected,
+      GlobalStatePointer newPointer) {
+    Document doc = toDoc(newPointer);
+    byte[] expectedGlobalId = expected.getGlobalId().toByteArray();
+
+    UpdateResult result =
+        client
+            .getGlobalPointers()
+            .updateOne(
```
rymurr, line 379: How has this been tested? From my reading of the [docs](https://docs.mongodb.com/manual/reference/method/db.collection.updateOne/) the `updateOne` method in this way is the best/correct operator. However most google searches about conditional updates talk about pipeline operators which makes me wonder if its possible that the filter in the first argument returns an entry then the 2nd argument modifies it but these two operations aren't done atomically and there could be a race under high load.

dimas-b: Mongo docs mention that changes are atomic at a specific document level... but I'll double check.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterFactory.java
dimas-b: I've made Mongo config properties nullable to match the pattern on other adapters: 4ba74c3889b28dcb70c6b3fb4d9a3048c92f2095

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: https://docs.mongodb.com/manual/core/write-operations-atomicity/

All write operations are supposed to be atomic at the level of individual documents. I have to assume that `updateOne` is considered one update operation (both the filter part and the change part), so if concurrent operations target the same doc, I suppose one of them will not "find" the changed data, which we catch in the result check (a few lines below).

Having reviewed the docs once more, I believe it might be preferable to use `replaceOne` instead, actually.

dimas-b: Switched to `replaceOne` in 681efd77e4b9657f67632fc35c82476ff7283ab1.

I do not think it makes a material difference in term of atomicity / consistency, but it looks cleaner code-wise.

dimas-b: Re: "conditional update". From my POV those pipelines are mostly used for deriving the updated value on the server side. We do not need to do that in our case as we have the complete new doc available at the start of `globalPointerCas()`

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapterConfig;
+
+@Value.Immutable(lazyhash = true)
+public interface MongoDatabaseAdapterConfig extends DatabaseAdapterConfig {
+
+  @Nullable
+  String getConnectionString();
+
```
snazy, line 27: Missing `MongoDatabaseClient withConnectionString(String connectionString)` +  `MongoDatabaseClient withDatabaseName(String databaseName)`

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/TestMongoDatabaseAdapter.java
```diff
@@ -0,0 +1,114 @@
...
+
+    mongo = starter.prepare(mongodConfig);
+    mongo.start();
+
+    assertThat(port).hasPositiveValue();
+
+    createAdapter(
+        "MongoDB",
+        config -> {
+          MongoDatabaseAdapterConfig testConfig =
```
snazy, line 97: Better move this into `MongoDatabaseAdapter.<init>()` if `config.getClient()` returns `null`.
It's not the greatest approach, but let's keep the functionality the same across all adapter implementations.

snazy: That way it's also guaranteed that the `MongoDatabaseClient` is closed when the `MongoDatabaseAdpater` is closed and not left behind when a test finishes.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
dimas-b: Why are they required?

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/TestMongoDatabaseAdapter.java
dimas-b: The closing of `MongoDatabaseClient` is handled by reference counting already. Will move code for consistency anyway.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
snazy: Just completeness - every *DatabaseAdapterConfig sub-type has those.

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/TestMongoDatabaseAdapter.java
dimas-b: Added default client creation logic to the adapter in a1c226a4c4965e21b3c581d638a8fc302d8a4061

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapterConfig.java
dimas-b: Added "with" methods in a1c226a4c4965e21b3c581d638a8fc302d8a4061

snazy (approved):
Ship it :)

PR #1966, opened 2021-09-14
Use `@ConfigProperties` instead of `@ConfigMappings`

snazy:
Reason: https://github.com/quarkusio/quarkus/issues/19990

Reverts: #1814

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1966)

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
@@ -57,8 +57,6 @@ public class CompiledAuthorizationRules {
    * @return A map of compiled authorization rules
    */
   private Map<String, Script> compileAuthorizationRules() {
-
-    // Map<String, String> rules = new HashMap<>(config.rules());
     Map<String, String> rules = new HashMap<>(rulesFromConfig());
     // by default we allow viewing all references until there's a user-defined VIEW_REFERENCE rule
     if (rules.entrySet().stream().noneMatch(r -> r.getValue().contains(VIEW_REFERENCE.name()))) {
@@ -87,8 +85,8 @@ public class CompiledAuthorizationRules {
     return ImmutableMap.copyOf(scripts);
   }
 
-  // TODO Quarkus/microprofile config API do not allow maps. Using this workaround here until
-  //  https://github.com/quarkusio/quarkus/issues/19990 is fixed.
+  // Quarkus/microprofile config API do not allow maps.
+  // Cannot use @ConfigMapping until https://github.com/quarkusio/quarkus/issues/19990 is fixed.
   private static Map<String, String> rulesFromConfig() {
     String prefix = "nessie.server.authorization.rules";
     Config config = ConfigProvider.getConfig();
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
@@ -21,8 +21,12 @@ import com.google.common.collect.ImmutableMap;
 import io.quarkus.runtime.Startup;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import org.eclipse.microprofile.config.Config;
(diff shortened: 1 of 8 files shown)


nastra (approved):

omarsmak (approved):

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
```diff
@@ -53,7 +57,9 @@ public CompiledAuthorizationRules(QuarkusNessieAuthorizationConfig config) {
    * @return A map of compiled authorization rules
    */
   private Map<String, Script> compileAuthorizationRules() {
-    Map<String, String> rules = new HashMap<>(config.rules());
+
+    // Map<String, String> rules = new HashMap<>(config.rules());
```
dimas-b, line 61: nit: remove this comment?

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
```diff
@@ -81,6 +87,26 @@ public CompiledAuthorizationRules(QuarkusNessieAuthorizationConfig config) {
...
+    return StreamSupport.stream(config.getPropertyNames().spliterator(), false)
+        .filter(name -> name.startsWith(prefix) && !name.equalsIgnoreCase(prefix))
+        .collect(
+            Collectors.toMap(
+                prop -> cleanPropertyName(prop.substring(prefix.length() + 1)),
+                prop -> config.getOptionalValue(prop, String.class).orElse("")));
+  }
+
+  private static String cleanPropertyName(String name) {
+    if (name.startsWith("\"") && name.endsWith("\"")) {
```
dimas-b, line 104: How can quotes occur in this context, given that `name` is a substring of a quarkus property?

dimas-b (dismissed):

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
snazy: It's the old code. Just left it so it's maybe easier to revert once the Quarkus issue is fixed.

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusNessieAuthorizationConfig.java
```diff
@@ -15,29 +15,28 @@
...
-import io.smallrye.config.ConfigMapping;
-import io.smallrye.config.WithDefault;
-import io.smallrye.config.WithName;
-import java.util.Map;
+import io.quarkus.arc.config.ConfigProperties;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 
 /** Configuration for Nessie authorization settings. */
-@ConfigMapping(prefix = "nessie.server.authorization")
+@ConfigProperties(prefix = "nessie.server.authorization")
```
nastra, line 22: TBH we shouldn't roll back these changes. Then we also wouldn't need the workaround that was added in `CompiledAuthorizationRules`.

Also in fact I believe we only need to roll back the changes in `QuarkusServerConfig`. Everything else can use the new `ConfigMapping`

omarsmak (approved):

nastra (changes requested):
I believe we only need to roll back the changes in `QuarkusServerConfig`. Everything else can use the new `ConfigMapping`

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CompiledAuthorizationRules.java
dimas-b: TIL :)

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusNessieAuthorizationConfig.java
snazy: > only need to roll back the changes in QuarkusServerConfig

Everything that's `@Inject`ed via a constructor is affected by the Quarkus isssue.

nastra: as far as I remember, the problem was always the injection of `ServerConfig` into `NessieExceptionMapper`. The changes from https://github.com/projectnessie/nessie/pull/1967 work for me locally. Let's see if native tests pass with that minimal set of changes

PR #1968, opened 2021-09-14
Unify authentication config property names

### LINKED ISSUE #1850: Use "nessie.authentication" prefix for all auth-related NessieClient config properties
This is to unify the client config naming scheme.

Impacts: Iceberg, DeltaLake, Spark integrations

dimas-b:
Fixes #1850

----

This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/1968)

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
@@ -29,14 +29,16 @@ public final class NessieConfigConstants {
    * Config property name ({@value #CONF_NESSIE_USERNAME}) for the user name used for (basic)
    * authentication.
    *
-   * @deprecated "basic" HTTP authentication is not considered secure.
+   * @deprecated "basic" HTTP authentication is not considered secure. Use {@link
+   *     #CONF_NESSIE_AUTH_TOKEN} instead.
    */
   @Deprecated public static final String CONF_NESSIE_USERNAME = "nessie.authentication.username";
   /**
    * Config property name ({@value #CONF_NESSIE_PASSWORD}) for the password used for (basic)
    * authentication.
    *
-   * @deprecated "basic" HTTP authentication is not considered secure.
+   * @deprecated "basic" HTTP authentication is not considered secure. Use {@link
+   *     #CONF_NESSIE_AUTH_TOKEN} instead.
    */
   @Deprecated public static final String CONF_NESSIE_PASSWORD = "nessie.authentication.password";
   /**
@@ -53,7 +55,7 @@ public final class NessieConfigConstants {
    * Config property name ({@value #CONF_NESSIE_AUTH_TYPE}) for the authentication provider ID.
    * Valid values are {@code BASIC}, {@code BEARER} and {@code AWS}.
    *
-   * <p>Note that "basic" HTTP authentication is not considered secure.
+   * <p>Note that "basic" HTTP authentication is not considered secure, use {@code BEARER} instead.
    */
   public static final String CONF_NESSIE_AUTH_TYPE = "nessie.authentication.type";
   /** Config property name ({@value #CONF_NESSIE_REF}) for the nessie reference used by clients. */
--- a/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
@@ -28,13 +28,17 @@ public final class NessieConfigConstants {
   /**
    * Config property name ({@value #CONF_NESSIE_USERNAME}) for the user name used for (basic)
    * authentication.
+   *
+   * @deprecated "basic" HTTP authentication is not considered secure.
    */
-  public static final String CONF_NESSIE_USERNAME = "nessie.username";
+  @Deprecated public static final String CONF_NESSIE_USERNAME = "nessie.authentication.username";
   /**
    * Config property name ({@value #CONF_NESSIE_PASSWORD}) for the password used for (basic)
    * authentication.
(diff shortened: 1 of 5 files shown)


clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -28,13 +28,17 @@
   /**
    * Config property name ({@value #CONF_NESSIE_USERNAME}) for the user name used for (basic)
    * authentication.
+   *
+   * @deprecated "basic" HTTP authentication is not considered secure.
    */
-  public static final String CONF_NESSIE_USERNAME = "nessie.username";
+  @Deprecated public static final String CONF_NESSIE_USERNAME = "nessie.authentication.username";
```
omarsmak, line 34: So to understand this correctly, the legacy is `nessie.username` and the new one `nessie.authentication.username`? 
If yes,  would it make sense to just keep the legacy config variables but only with `@Deprecated` annotation? E.g:
```
@Deprecated public static final String CONF_NESSIE_USERNAME = "nessie.username";
public static final String CONF_NESSIE_AUTHENTICATION_USERNAME = "nessie.authentication.username"
```
And then we can use the legacy config variables to fallback as you already doing [there](https://github.com/projectnessie/nessie/pull/1968/files#diff-16b70889c228cd93cc9caa0811e507ffe3f350b3b420790f2cc02be2ae2859b1R51) but using the deprecated variable.

site/docs/tools/spark.md
```diff
@@ -374,11 +374,15 @@ SparkSession.builder
 
 The Nessie LogStore needs the following parameters set in the Spark/Hadoop config.
 
+Note: "BASIC" authentication is not supported by Nessie servers in "production" mode.
+
 ```
 nessie.url = full url to nessie
-nessie.username = username if using basic auth, omitted otherwise
```
omarsmak, line 379: Maybe I wouldn't remove that section completely but I will just add a note that these configs are being deprecated.

clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
nastra, line 34: renaming the underlying property name and having a fallback to the old one makes sense to me, but what's the reason for deprecating the `CONF_NESSIE_USERNAME` field itself?

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -47,7 +47,14 @@ public String getAuthTypeValue() {
   @Override
   public HttpAuthentication build(Function<String, String> configSupplier) {
     String username = configSupplier.apply(CONF_NESSIE_USERNAME);
+    if (username == null) {
```
omarsmak, line 50: I am wondering, does it make sense to check as well for the String emptiness as well as Null or only Null check suffice here?

clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
dimas-b: `@Deprecated` is also about the security implications, plus production Nessie server do not support username/password (aka "basic") authentication.

dimas-b: production Nessie servers do not support username/password (aka "basic") authentication.

dimas-b: "Basic" auth works only in "dev" and "test" environments

clients/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
dimas-b: I'm not sure it's worth complicating the code for that. This `if` is just for backward compatibility with existing clients. I imagine we're going to remove it once all clients migrate to Nessie 1.0+.

site/docs/tools/spark.md
dimas-b: I'm not sure I understand, sorry. This is not a removal, but a rename of properties... the new names are mentioned in the "added" lines below.

clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
nastra: ok makes sense

nastra: ok makes sense, thanks

clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -28,13 +28,17 @@
   /**
    * Config property name ({@value #CONF_NESSIE_USERNAME}) for the user name used for (basic)
    * authentication.
+   *
+   * @deprecated "basic" HTTP authentication is not considered secure.
```
nastra, line 32: nit: maybe add `Please use other authentication mechanisms, such as XYZ`

nastra (dismissed):

omarsmak (approved):
Got it, thanks

clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
dimas-b: Good point. Redirected to `BEARER` in f85e83986ea536bb39fe2c16eee11c6e02d1cc8f

nastra (approved):

PR #2007, opened 2021-09-16
Update Jersey/Weld to fix CVE violations from OWASP report

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2007)


snazy:
Why doesn't dependabot catch those?

nastra:
I think because dependabot once updated those dependencies to the latest version, but those implied changing lots of package names (javax.ws.rs.core changed to jakarta.ws.rs.core), so we didn't update and probably that's why dependabot doesn't bug us anymore

snazy (approved):

snazy:
Diappointing :(

PR #2010, opened 2021-09-16
TVS: use junit extensions

snazy:
TVS: use junit extensions

Testing the varions database-adapter implementations is currently based on logic in the subclasses
of `AbstractTieredCommitsTest`. This is okay, as there is currently only one kind of test class
(the ones extending `AbstractTieredCommitsTest`).

However, there will be kind of tests exercising various database adapter implementations:
* tests per database adapter extending `AbstractVersionStoreTest`
* tests for Nessie-GC exercising against a couple database adapters

Copying the per-database-adapter logic to multiple test classes is error-prone and also unnecessary.

The proposed approach introduces a JUnit extension `DatabaseAdapterExtension` with a couple of
annotations that define/declare _which_ and _how_ a `DatabaseAdapter`, and later a derived
`VersionStore`, is configured.

A per-database-adapter test class looks like this:
```
@NessieExternalDatabase(RocksTestConnectionProviderSource.class)
class TestTieredCommitsRocks extends AbstractTieredCommitsTest {}
```
and the abstract test class like this:
```
@NessieAdapterConfigItem(name = "max.key.list.size", value = "2048")
public abstract class AbstractTieredCommitsTest {
  @NessieAdapter protected static DatabaseAdapter databaseAdapter;
```

Once the new version store has landed in `main`, the extension can also inject a matching
version-store implementation:
```
  @NessieAdapter VersionStore<String, String, StringStoreWorker.TestEnum> store;
```

The `@NessieAdapter` and `@NessieAdapterConfigItem` annotations can be used on static fields,
instance fields and test method parameters.

Test method parameter example, which provides two `DatabaseAdapter` instances with different
configurations but both share the same database:
```
  @Test
  void keyPrefixBasic(
      @NessieAdapter @NessieAdapterConfigItem(name = "key.prefix", value = "foo")
          DatabaseAdapter foo,
      @NessieAdapter @NessieAdapterConfigItem(name = "key.prefix", value = "bar")
          DatabaseAdapter bar)
      throws Exception { ... }
```

This change also separates the configuration for the (local) database-connection-provider
from the configuration of the database-adapter. `DatabaseAdapter` implementations do now
expect a non-null database-connection-provider via `DatabaseAdapterConfig`.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2010)


versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/DatabaseAdapterExtension.java
```diff
@@ -0,0 +1,194 @@
...
+
+                TestConfigurer configurer =
+                    makeAccessible(annotation.configurer().getDeclaredConstructor()).newInstance();
+                DatabaseAdapter adapter =
+                    context
+                        .getStore(NAMESPACE)
+                        .getOrComputeIfAbsent(
+                            KEY,
+                            key -> new AdapterResource(createAdapter(context, configurer)),
+                            AdapterResource.class)
```
dimas-b, line 155: This looks like we will be creating and closing adapters and related storage databases per test class, if I'm not mistaken. Is that intended? Currently we create one storage for all adapter tests.

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/TestMongoDatabaseAdapter.java
```diff
@@ -17,100 +17,17 @@
...
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.projectnessie.versioned.persist.tests.AbstractTieredCommitsTest;
 
+@ExtendWith(MongoDatabaseAdapterExtension.class)
 public class TestMongoDatabaseAdapter extends AbstractTieredCommitsTest {
```
dimas-b, line 25: Just as an idea for consideration: We could use two extensions: one for managing the storage (MongoDB, RockDB, containers, etc) and another for managing adapters. That would allow tests for pick and choose adapter-storage permutations.

Also, since a lot of the tests are written in an adapter-neutral way, we could reuse the test classes for all adapter without requiring test sub-classes. That would allow us to bring nested tests to the top level to make them executable directly from the IDE.

Unfortunately, it does not seem possible to produce test contexts in JUnit5 from an extension without interfering with `@ParameterizedTest`. That is an extension can produce new test context for different adapter names and storage settings, but that will only work with simple `@Test` methods. `@ParameterizedTest` will not be able to resolve their parameters.

However, if we introduced a custom JUnit5 test _engine_ we would be able to create new branches in the test context hierarchy that would be compatible with `@ParameterizedTest`. Our own extensions would use the `UniqueId` segment information for figuring out storage settings and adapter names.

For example: currently a test with `UniqueId` of `[engine:junit-jupiter]/[class:org.projectnessie.versioned.persist.rocks.TestTieredCommitsRocks]/[nested-class:Concurrency]/[test-template:concurrency(org.projectnessie.versioned.persist.tests.AbstractConcurrency$Variation)]/[test-template-invocation:#1]`

Would become something like `[engine:junit-jupiter]/[storage:rocks-local][adapter:RocksDB][class:org.projectnessie.versioned.persist.rocks.TestTieredCommitsRocks]/[nested-class:Concurrency]/[test-template:concurrency(org.projectnessie.versioned.persist.tests.AbstractConcurrency$Variation)]/[test-template-invocation:#1]`

I have a sketch of this approach at: https://github.com/dimas-b/stargate/commit/40a88a4f9e9950338c38794915b2d68cbf3ee639 (cf. `StorageAwareEngine`)

Some highlights:
* one set of tests for all adapters would live in a separate maven module
* each adapter module will depend on common tests and define resources (or annotations) for discovering their specific database config settings
* The custom test engine will use annotations to find related tests and customize the text context tree for them during the JUnit 5 discovery phase
* IDEs will be able to run individual tests from source and re-run / debug specific test configurations from the test run pane by virtue of passing a specific test `UniqueId` to JUnit 5
* Maven will run all tests with all database configrations by introspecting test dependencies
* Our custom test _extensions_ will be able to injects relevant database / adapter objects based on `UniqueId` segments and test annotations.

dimas-b (commented):
I only had a very brief looks at this PR, but realized I was trying to solve a similar problem about a year ago. Please see my comments below. Just wondering what you think. If you'd like to stay on the pure extension approach I'll continue reviewing :)

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/DatabaseAdapterExtension.java
snazy: No, it's created for the whole "adapter test class" including all subclasses

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/TestMongoDatabaseAdapter.java
snazy: Not sure whether I'd like to go down the route of implementing a JUnit5 test engine, even if delegating to the "standard" one. I'd rather pay the price of adding "mostly empty" test classes at the moment.

It's not too hard to inject adapters with custom configurations into specific test cases from an extension, if the `DatabaseAdapter` (or `VersionStore`) is injected as a test method parameter. Not sure whether `@ParameterizedTest` really doesn't work with `org.junit.jupiter.api.extension.ParameterResolver` if the parameterized-test is annotated with the "right" `@ExtendWith`.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/DatabaseAdapterExtension.java
```diff
@@ -0,0 +1,190 @@
...
+    @Override
+    public void close() throws Throwable {
+      databaseAdapter.close();
+    }
+  }
+
+  @Override
+  public void beforeAll(ExtensionContext context) throws Exception {
+    Class<?> testClass = context.getRequiredTestClass();
+    findAnnotatedFields(testClass, Adapter.class, ReflectionUtils::isStatic)
```
dimas-b, line 130: I'd do it differently. Since we assume the adapter is created in the `static` context we could declare the configurer in a class-level annotation.

This extension would them get the adapter name + configuration method from the annotation and always create the adapter through the factory. There's no need to subclass the extension itself. I think we can also get away without a `TestConfigurer`

For example:
```
@NessieAdapterConfig(createMethod="createAdapter")
public class TestMongoDatabaseAdapter extends AbstractTieredCommitsTest {
  static DatabaseAdapter createAdapter(...) {
    ...
  }
}
```


Then, we'd also implement `ParameterResolver` and inject the adapter into any/all lifecycle method parameters, be them static or per-instance.

`AbstractTieredCommitsTest` would have this method to receive the pre-configured adapter.
```
`@BeforeAll 
static void setup(DataAdapter adapter) {
  databaseAdapter = adapter;
}
```

WDYT?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/DatabaseAdapterExtension.java
```diff
@@ -0,0 +1,190 @@
...
+  @Override
+  public void beforeEach(ExtensionContext context) {
+    AdapterResource adapter = context.getStore(NAMESPACE).get(KEY, AdapterResource.class);
+
+    if (adapter == null) {
+      throw new JUnitException(
+          "Test class does not declare mandatory '@DatabaseAdapterExtension.Adapter static DatabaseAdapter databaseAdapter;' field");
+    }
+
+    adapter.databaseAdapter.reinitializeRepo("main");
```
dimas-b, line 170: I'd keep this part in a `@BeforeEach` method on `AbstractTieredCommitsTest`. it kind of belongs more with the actual testing logic, while "extensions" can concentrate of the storage and adapter setup part.

snazy:
I went with a slightly different approach as described in the PR description. It decouples the "connection-provider" from the "database-adapter". The "connection-provider" is rather "global/static" for each test class, but `DatabaseAdapter`/`VersionStore` can be either per-test-class (static field), per-test-instance (instance field) or per-test-case (test method parameter).

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/postgres/PostgresTestConnectionProviderSource.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.testcontainers.containers.CockroachContainer;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+
+public class PostgresTestConnectionProviderSource extends ContainerTestConnectionProviderSource {
+
+  @Override
+  protected JdbcDatabaseContainer<?> createContainer() {
+    String version = System.getProperty("it.nessie.container.cockroach.tag", "v21.1.6");
+    return new CockroachContainer("cockroachdb/cockroach:" + version);
```
nastra, line 26: copy/paste error or why is this using Cockroach?

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/postgres/ITTieredCommitsCockroach.java
```diff
@@ -15,32 +15,12 @@
...
-            .withConfig(ImmutableDefaultLocalDatabaseAdapterConfig.builder().build()),
-        config -> container.configureDatabaseAdapter(config));
-  }
-
-  @AfterAll
-  static void stopContainer() {
-    container.stop();
-  }
-}
+@NessieAdapterName("PostgreSQL")
```
nastra, line 24: copy/paste error? Should this be `Cockroach`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+package org.projectnessie.versioned.persist.tests.extension;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface NessieAdapter {}
```
nastra, line 25: maybe worth renaming to `NessieDbAdapter`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieAdapterName.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * The name of the database-adapter to use in the annotated test class.
+ *
+ * <p>Can be omitted, if there is only one {@link
+ * org.projectnessie.versioned.persist.adapter.DatabaseAdapterFactory} implementation on the
+ * classpath.
+ */
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface NessieAdapterName {
```
nastra, line 32: maybe rename to `NessieDbAdapterName`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieAdapterConfigItem.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+@Repeatable(NessieAdapterConfigItems.class)
+public @interface NessieAdapterConfigItem {
```
nastra, line 27: maybe rename to `NessieDbAdapterConfigItem`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieAdapterConfigItems.java
```diff
@@ -0,0 +1,27 @@
...
+package org.projectnessie.versioned.persist.tests.extension;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.TYPE, ElementType.FIELD, ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface NessieAdapterConfigItems {
```
nastra, line 25: maybe rename to `NessieDbAdapterConfigItems`

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractTieredCommitsTest.java
```diff
@@ -408,4 +375,65 @@ void recreateDefaultBranch() throws Exception {
...
+            .commitMetaSerialized(ByteString.copyFromUtf8("meta-bar"))
+            .addPuts(
+                KeyWithBytes.of(
+                    Key.of("bar"), ContentsId.of("bar"), (byte) 0, ByteString.copyFromUtf8("bar")))
+            .build());
+
+    Hash fooMain = foo.toHash(main);
+    Hash barMain = bar.toHash(main);
+
+    Hash fooBranch = foo.create(BranchName.of("foo-branch"), fooMain);
```
nastra, line 409: nit: probably worth extracting `BranchName.of("foo-branch")` into a local variable for reuse. Same for the `barBranch`

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractTieredCommitsTest.java
```diff
@@ -408,4 +375,65 @@ void recreateDefaultBranch() throws Exception {
...
+
+    assertThatThrownBy(() -> foo.commitLog(barBranch))
+        .isInstanceOf(ReferenceNotFoundException.class);
+    assertThatThrownBy(() -> bar.commitLog(fooBranch))
+        .isInstanceOf(ReferenceNotFoundException.class);
+
+    try (Stream<CommitLogEntry> log = foo.commitLog(fooBranch)) {
+      assertThat(log.collect(Collectors.toList()))
+          .extracting(CommitLogEntry::getMetadata)
+          .containsExactlyInAnyOrder(ByteString.copyFromUtf8("meta-foo"));
```
nastra, line 431: maybe extract `ByteString.copyFromUtf8("meta-foo")` into a variable. Same for `meta-bar`

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * <p>The test class must be annotated with {@link
+ * org.junit.jupiter.api.extension.ExtendWith @ExtendWith}.
+ */
+public class DatabaseAdapterExtension
+    implements BeforeAllCallback, BeforeEachCallback, ParameterResolver {
+  private static final Namespace NAMESPACE = Namespace.create(DatabaseAdapterExtension.class);
+  private static final String KEY_STATICS = "static-adapters";
+
+  private static class StaticAdapters implements CloseableResource {
```
nastra, line 60: I find `StaticAdapters` a bit confusing in terms of its name. Maybe just `DbAdapters`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
```diff
@@ -0,0 +1,262 @@
...
+                          .getTestClass()
+                          .ifPresent(
+                              cls -> {
+                                for (; cls != Object.class; cls = cls.getSuperclass()) {
+                                  collector.accept(cls);
+                                }
+                              });
+
+                      return configs.stream()
+                          .filter(n -> ("nessie.store." + n.name()).equals(property))
```
nastra, line 228: not sure if it's worth putting this prefix into a static final somewhere so that all the other places that use this also use the same prefix

nastra (changes requested):
I like the direction where this is going and overall the changes LGTM and make sense. Just a few things need to be updated/fixed.

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/postgres/PostgresTestConnectionProviderSource.java
snazy: copy-paste bug, fixed

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/postgres/ITTieredCommitsCockroach.java
snazy: that's correct - they're using the same code

nastra: isn't PostgreSQL using `PostgreSQLContainer` and Cockroach is using `CockroachContainer`?

nastra (approved):

PR #2083, opened 2021-09-20
TVS: actual `VersionStore` implementation incl tests

snazy:
Depends on #2010

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2083)

PR #2097, opened 2021-09-21
Fix NPE when hash is null

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2097)

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
@@ -142,6 +142,16 @@ public abstract class AbstractTestRest {
     String branchName1 = "createReferences_branch1";
     String branchName2 = "createReferences_branch2";
     assertAll(
+        // invalid source ref & null hash
+        () ->
+            assertThatThrownBy(
+                    () ->
+                        api.createReference()
+                            .sourceRefName("unknownSource")
+                            .reference(Tag.of(tagName2, null))
+                            .create())
+                .isInstanceOf(NessieNotFoundException.class)
+                .hasMessage("Ref 'unknownSource' does not exist"),
         // Tag without hash
         () ->
             assertThatThrownBy(
--- a/servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
@@ -20,6 +20,7 @@ import static org.projectnessie.model.Operation.Put;
 
 import java.security.AccessControlException;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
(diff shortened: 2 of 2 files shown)


servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -135,8 +133,12 @@ protected void checkHashOnRef(String refName, String hash) throws NessieNotFound
...
-            || hash.equals(getStore().noAncestorHash().asString())) {
+        System.out.println("EX refName = " + refName);
+        System.out.println("EX hash = " + hash);
+        System.out.println("getStore().noAncestorHash() = " + getStore().noAncestorHash());
+
+        if (Objects.equals(getConfig().getDefaultBranch(), refName) && hash == null
+            || Objects.equals(getStore().noAncestorHash().asString(), hash)) {
           // prevent the "hash-on-ref" check, but still do the "can-view" check
           hash = null;
           return;
```
snazy, line 144: The `return` needs to be removed (sorry, oversight)

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -135,8 +133,12 @@ protected void checkHashOnRef(String refName, String hash) throws NessieNotFound
         // Special case, mostly for tests:
         // When the default branch is deleted, we cannot verify whether the branch already exists
         // (it does not, so the check would fail).
-        if (refName.equals(getConfig().getDefaultBranch()) && hash == null
-            || hash.equals(getStore().noAncestorHash().asString())) {
```
snazy, line 139: The `equals` should have probably be the other way around (as `hash` might be `null`??)

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
nastra: I'll have a follow-up PR where I'll merge this `hashOnRef` check with `namedHashOrThrow` as both are essentially doing the same

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
nastra: yep, `hash` ends up being null

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -18,6 +18,7 @@
 import static org.projectnessie.model.Operation.Delete;
 import static org.projectnessie.model.Operation.Put;
 
+import com.google.common.base.Objects;
```
snazy, line 21: Really prefer `java.util.Objects#equals`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -142,6 +142,16 @@ void createReferences() throws NessieNotFoundException {
     String branchName1 = "createReferences_branch1";
     String branchName2 = "createReferences_branch2";
     assertAll(
+        // invalid source ref & null hash
```
snazy, line 145: This will conflict with the addressed review-comment https://github.com/projectnessie/nessie/pull/2083#discussion_r713701618 which will eagerly throw an IAE when `hash` is null for a tag.

snazy: Better remove this test.

nastra: the null check introduced in #2083 happens after we're hitting the NPE so it's still valid

snazy: I suspect, the test will fail with that change ;)

nastra: rebased onto latest changes and re-running CI. Test is passing locally. As mentioned, the `checkHashOnRef` check (where the NPE is happening) is done before the null check you added in #2083

snazy (approved):

PR #2103, opened 2021-09-22
Cleanup/merge hashOnRef checks

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2103)

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
+++ b/servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
@@ -16,10 +16,8 @@
 package org.projectnessie.services.rest;
 
 import java.security.Principal;
-import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import javax.ws.rs.core.SecurityContext;
 import org.projectnessie.error.NessieNotFoundException;
@@ -30,6 +28,7 @@ import org.projectnessie.services.authz.ServerAccessContext;
 import org.projectnessie.services.config.ServerConfig;
 import org.projectnessie.versioned.Hash;
 import org.projectnessie.versioned.NamedRef;
+import org.projectnessie.versioned.Ref;
 import org.projectnessie.versioned.ReferenceNotFoundException;
 import org.projectnessie.versioned.VersionStore;
 import org.projectnessie.versioned.WithHash;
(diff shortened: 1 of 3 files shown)


servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
```diff
@@ -100,63 +96,24 @@ public Reference getReferenceByName(String refName) throws NessieNotFoundExcepti
...
-    if (sourceRefName == null) {
-      sourceRefName = getConfig().getDefaultBranch();
-    }
-
     if (reference instanceof Branch) {
       getAccessChecker()
           .canCreateReference(createAccessContext(), BranchName.of(reference.getName()));
     } else if (reference instanceof Tag) {
       getAccessChecker().canCreateReference(createAccessContext(), TagName.of(reference.getName()));
-    } else {
```
omarsmak, line 112: does this `else` block need to be removed?

snazy (commented):
Is this change compatible with #2083 ?

snazy:
Let's wait until the remaining TVS2-PRs are in.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResourceWithAuthorizationChecks.java
nastra: yes, the same check is happening in the super class. Also the purpose of `TreeResourceWithAuthorizationChecks` is to only perform authz checks and not validation logic (which was represented by this else part)

nastra: yes, the same check is happening in the super class. Also the purpose of `TreeResourceWithAuthorizationChecks` is to only perform authz checks and not validation logic (which was represented by this else part)

omarsmak: got it, thanks

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -55,30 +54,22 @@ protected BaseResource(
...
-                      new NessieNotFoundException(
-                          String.format("Named reference '%s' not found.", namedRef)));
+    try {
+      WithHash<Ref> refWithHash = getStore().toRef(namedRef);
+      Ref ref = refWithHash.getValue();
+      if (!(ref instanceof NamedRef)) {
+        throw new ReferenceNotFoundException(
+            String.format("Named reference '%s' not found", namedRef));
+      }
+      namedRefWithHash = WithHash.of(refWithHash.getHash(), (NamedRef) refWithHash.getValue());
```
snazy, line 70:
```suggestion
      namedRefWithHash = WithHash.of(refWithHash.getHash(), (NamedRef) ref);
```

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -87,17 +78,11 @@ protected BaseResource(
...
       // we need to make sure that the hash in fact exists on the named ref
-      Hash hash = Hash.of(hashOnRef);
-      try (Stream<WithHash<CommitMeta>> commits = store.getCommits(namedRefWithHash.getValue())) {
-        if (commits.noneMatch(c -> c.getHash().equals(hash))) {
-          throw new NessieNotFoundException(
-              String.format("Could not find commit '%s' in reference '%s'.", hashOnRef, namedRef));
-        }
-      }
-      return WithHash.of(hash, namedRefWithHash.getValue());
+      return WithHash.of(
```
snazy, line 81: I think, it makes sense to compare `hashOnRef` w/ `refWithHash.getHash()` and just return `namedRefWithHash` when equal to prevent the then useless call to `getStore().hashOnReference()` (could be placed into the above `if`

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
```diff
@@ -87,17 +78,11 @@ protected BaseResource(
...
+          getStore().hashOnReference(namedRefWithHash.getValue(), Optional.of(Hash.of(hashOnRef))),
+          namedRefWithHash.getValue());
     } catch (ReferenceNotFoundException e) {
-      throw new NessieNotFoundException(
-          String.format("Could not find commit '%s' in reference '%s'.", hashOnRef, namedRef));
+      throw new NessieNotFoundException(e.getMessage());
```
snazy, line 85: Unrelated to this PR, but WDYT about having a helper class to replace all these `new NesseXyzException(e.getMessage())` calls?

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
snazy: Would probably require some inline-doc to explain when the if-condition is met.

snazy (commented):
LGTM - few minor comments

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
nastra: yes that would definitely be useful. I saw that we already have similar things in `DatabaseAdapterUtil` but it would be better to have this at a more central place where `DatabaseAdapterUtil` and other components can use this.

servers/services/src/main/java/org/projectnessie/services/rest/BaseResource.java
nastra: yep that makes sense, implemented it and added a comment explaining why we can exit early

snazy (approved):

PR #2107, opened 2021-09-22
Speed up main ci

snazy:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2107)


.github/workflows/main.yml
```diff
@@ -60,12 +60,31 @@ jobs:
...
-      if: ${{ env.DOCKER_TOKEN }}
-      run: |
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup runner
+        uses: ./.github/actions/setup-runner
+      - name: Setup Java, Maven, Gradle
+        uses: ./.github/actions/dev-tool-java
+
+      - name: Build with Maven
```
nastra, line 76: is this something that can be extracted as we're almost doing the same thing in the pull-request-native.yml?

snazy: Rather not, because a composite-action is shown as a single step - i.e. the whole log for the build and the log for the relevant native step are unified. Having those as separate steps in the workflow-yaml is nicer.

dimas-b (approved):

PR #2112, opened 2021-09-22
TVS: cleanup

snazy:
Big-bang-cleanup PR

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2112)


snazy:
Note: there are two commits: one that "only" deletes a bunch of files and one commit with changes to files.

site/docs/try/configuration.md
```diff
@@ -15,20 +15,11 @@ docker run -p 8080:8080 projectnessie/nessie \
...
-## Dynamo version store specific configuration
-# should Nessie create its own dynamo tables
-nessie.version.store.dynamo.initialize=false
-
-## Dynamo Configuration
-quarkus.dynamodb.aws.region=us-west-2
-quarkus.dynamodb.aws.credentials.type=DEFAULT
-# quarkus.dynamodb.endpoint-override=http://localhost:8000
+# path if using RocksDB
+nessie.version.store.rocks.directory=/tmp/rocks-nessie
```
dimas-b, line 22: Isn't this effectively the same as "in memory"? Should we suggest a non-transient location perhaps?

versioned/tests/src/main/java/org/projectnessie/versioned/StringStoreWorker.java
```diff
@@ -89,24 +89,22 @@ public String getId(String contents) {
...
+    return 0;
   }
 
   @Override
   public TestEnum getType(Byte payload) {
-    return getValueSerializer().getType(payload);
+    if (payload == null) {
+      return StringStoreWorker.TestEnum.NULL;
+    }
+    return payload > 60 ? StringStoreWorker.TestEnum.YES : StringStoreWorker.TestEnum.NO;
```
dimas-b, line 100: why 60 and not 42 ;-)

dimas-b (commented):
Reviewed the second commit only :)

versioned/tests/src/main/java/org/projectnessie/versioned/StringStoreWorker.java
snazy: IDK - was like that before. It's a test-only thing though ;)

snazy: I'd have DEFINITELY chosen 42 :P

site/docs/try/configuration.md
snazy: `/tmp/` has the advantage of being accessible to all users (unlike `/var/lib` or so).

Note: `/tmp` is not always transient in Linux though. [FHS 3.0](https://refspecs.linuxbase.org/FHS_3.0/fhs/ch03s18.html) states that you should not _assume_ that files survive reboots, but at least Ubuntu does not clear `/tmp` nor does it periodically do that (anymore). I suspect, clearing `/tmp`just caused too many problems (a lot of things place important, long-living things there - e.g. Java's hsperfdata, systemd, X11, etc).

dimas-b: fair enough.

dimas-b (dismissed):

nastra (approved):

PR #2128, opened 2021-09-23
Don't run tests when httpClient is null / Fix eq+hashCode in EntriesParams

nastra:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2128)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -120,7 +121,7 @@ protected void init(URI uri) {
     init(api, httpClient);
   }
 
-  protected void init(NessieApiV1 api, HttpClient httpClient) {
+  protected void init(NessieApiV1 api, @Nullable HttpClient httpClient) {
```
nastra, line 124: The gRPC tests pass null for the `httpClient`, so we need to skip tests that rely on the `httpClient`

snazy (approved):

PR #2132, opened 2021-09-23
Separate configuration for # of parents in GlobalLogEntry

snazy:
The implementation is currently reusing the number of parents per CommitLogEntry for
the number of parents in a GlobalLogEntry. The payload sizes for both types are very
different and it makes sense to have different configuration options and different
defaults for both. CommitLogEntry can be assumed to be bigger than GlobalLogEntry.
It also makes sense to have more parents per GlobalLogEntry, because we do not (yet)
aggregate the global-contents-values within the global-log.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2132)

### DIFF
--- a/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
+++ b/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,37 @@
...
+public interface NonTransactionalDatabaseAdapterConfig<T extends DatabaseConnectionProvider<?>>
+    extends DatabaseAdapterConfig<T> {
+  int DEFAULT_PARENTS_PER_GLOBAL_COMMIT = 50;
+
+  /**
+   * The number of parent-global-commit-hashes stored in {@link
+   * GlobalStateLogEntry#getParentsList()}. Defaults to {@value #DEFAULT_PARENTS_PER_GLOBAL_COMMIT}.
+   */
+  @Value.Default
+  default int getParentsPerGlobalCommit() {
```
dimas-b, line 32: I'm still wondering whether it might be worth keeping a global commit log in _all_ adapter types :thinking: mainly for uniformity of the data model at a conceptual level. Also could be useful for recovery, I guess.

dimas-b (approved):

Issue #1981, opened 2021-09-15
Document auth


rymurr:

dimas-b:
"basic" auth is to be removed from docs as it is not supported in production. Tests can refer to source/javadoc.

snazy:
@dimas-b can this issue be closed now?

dimas-b:
Almost, I need some more time to review our auth scenarios... likely no more docs changes, but I just want to double check.

PR #2137, opened 2021-09-23
End-to-end auth example with Keycloak

dimas-b:
Contributes to #1981

----

Not sure if we should merge it before releasing 0.9.3 (or later) because currently published Nessie Docker images do not have full auth support.

### DIFF
--- a/site/docs/guides/keycloak.md
+++ b/site/docs/guides/keycloak.md
@@ -1,0 +1,72 @@
+# Authentication with Keycloak
+
+In this guide we walk through the process of configuring a Nessie Server to authenticate clients against 
+a local [Keycloak](https://www.keycloak.org/) server. Docker is use at the runtime environments for both servers.
+
+## Setting up Keycloak
+
+For the purposes of this guide we will only do use a simple Keycloak configuration, that is still sufficient to
+demonstrate how OpenID authentication works in Nessie servers.
+
+First, start a Keycloak container using its latest Docker image.
+
+```shell
+docker run -p 8080:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin \
+  --name keycloak quay.io/keycloak/keycloak:latest
+```
+
+Note the `admin` username and password. Those values will be required to log into the Keycloak Administration Console
+that should now be available at http://localhost:8080/auth/admin/
+
+The default realm is called `Master`. On the left-hand pane find the `Manage > User` page and click `Add User` on the
+right side of the (initially empty) users table.
+
+Enter the username "nessie" and click `Save`. Now, under the `Credentials` tab of the `nessie` user page set password
+to `nessie` and turn off the `Temporary` flag. Click `Set Password`.
+
+For the sake of convenience let's increase the default token expiration time.
+Goto `Clients` > `admin-cli` > `Advanced Settings`. Set `Access Token Lifespan` to 1 day and click `Save`.
+
+Now we are ready to generate an OpenID access token to the `nessie` user:
+
+```shell
+curl -X POST \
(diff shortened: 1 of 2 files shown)


site/docs/guides/keycloak.md
```diff
@@ -0,0 +1,72 @@
...
+
+```shell
+docker run -p 8080:8080 -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin \
+  --name keycloak quay.io/keycloak/keycloak:latest
+```
+
+Note the `admin` username and password. Those values will be required to log into the Keycloak Administration Console
+that should now be available at http://localhost:8080/auth/admin/
+
+The default realm is called `Master`. On the left-hand pane find the `Manage > User` page and click `Add User` on the
```
snazy, line 21:
```suggestion
The default realm is called `Master`. On the left-hand pane find the `Manage > Users` page and click `Add User` on the
```

site/docs/guides/keycloak.md
```diff
@@ -0,0 +1,72 @@
...
+Enter the username "nessie" and click `Save`. Now, under the `Credentials` tab of the `nessie` user page set password
+to `nessie` and turn off the `Temporary` flag. Click `Set Password`.
+
+For the sake of convenience let's increase the default token expiration time.
+Goto `Clients` > `admin-cli` > `Advanced Settings`. Set `Access Token Lifespan` to 1 day and click `Save`.
+
+Now we are ready to generate an OpenID access token to the `nessie` user:
+
+```shell
+curl -X POST \
```
snazy, line 33: Maybe 
```
export NESSIE_AUTH_TOKEN=$(
    <that curl command> 
)
```

site/docs/guides/keycloak.md
```diff
@@ -0,0 +1,72 @@
...
+docker run -p 19120:19120 \
+  -e QUARKUS_OIDC_AUTH_SERVER_URL=http://localhost:8080/auth/realms/master \
+  -e QUARKUS_OIDC_CLIENT_ID=projectnessie \
+  -e NESSIE_SERVER_AUTHENTICATION_ENABLED=true \
+  --network host projectnessie/nessie
+```
+
+## Using Nessie CLI
+
+Now that the Nessie server runs in authenticated mode with a Keycloak, [clients](../tools/cli.md) have to provide
```
snazy, line 58: Something's really odd here.
I ran through all the steps above and installed `pynessie`.
However, a `nessie remote show` works even without a `NESSIE_AUTH_TOKEN` or an invalid `NESSIE_AUTH_TOKEN`. Guess, it requires some authn rules and a negative example?

snazy (changes requested):

site/docs/guides/keycloak.md
dimas-b: I thought about that but I did not want to the example to be bash-specific ;)

site/docs/guides/keycloak.md
dimas-b: That's because the latest published docker image does not have auth in it (see my top-level comment).

Would it make sense to release 0.9.3 before 1.0?

dimas-b: If you build docker locally, the example should work

site/docs/guides/keycloak.md
dimas-b: WDYT about the latest variant? 906af61764b607fd71205d1708deec3f6f638766

snazy: :+1:

site/docs/guides/keycloak.md
snazy: Ah, right.

PR #2146, opened 2021-09-24
Re-set commit meta on transplant/merge

harshm-dev:
Fixes [831](https://github.com/projectnessie/nessie/issues/831)

The patch attempts following changes on the merge/transplant commits -
* Committer changed to user principal, empty string when user is anonymous
* Commit time set to current time

Setting an updated commit time during merge is crucial as it results in an time-ordered sequence of commits, which features such as GC will rely upon.

Approach -
The user principal is available in the TreeResource. This is required for setting the committer. However, copying/shifting commits are done in the layers below (DatabaseAdapter or VersionStore). Hence, approach prepares a lambda function, to translate commit metadata, and set properties.
The version store allows generic metadata object, and has access to serializer class (which is not available in the DBAdapter). Hence, version store bears responsibility of supplying the same on the function passed for metadata reset.

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2146)


harshm-dremio:
The tests will go through an iteration, but the code can be reviewed.

nastra:
@harshm-dremio can you rebase please? Jgit and the old inmemory implementation have been removed from `main`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +302,40 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
+        String mergeProp = String.format("%s_ref", mergeOp.name().toLowerCase());
+        CommitMeta mergeReadyMeta =
+            CommitMeta.builder()
+                .from(commitMeta)
+                .committer(currUser)
+                .commitTime(Instant.now())
+                .putProperties(mergeProp, fromRefName)
+                .build();
+        return ser.toBytes(mergeReadyMeta);
+      } catch (RuntimeException e) { // error in serialization
```
snazy, line 332: This should either fail completely or succeed, but only logging a warning isn't helpful IMO

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +302,40 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
     }
   }
 
+  private BiFunction<Serializer<CommitMeta>, ByteString, ByteString> getResetMergePropsFunc(
+      MergeOp mergeOp, String fromRefName) {
+    return (ser, in) -> {
+      try {
+        CommitMeta commitMeta = ser.fromBytes(in);
+        String currUser =
+            Optional.ofNullable(getSecurityContext().getUserPrincipal())
```
snazy, line 320: Why not `getPrincipal()` here?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +302,40 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
       throw new NessieConflictException(e.getMessage(), e);
     }
   }
 
+  private BiFunction<Serializer<CommitMeta>, ByteString, ByteString> getResetMergePropsFunc(
+      MergeOp mergeOp, String fromRefName) {
+    return (ser, in) -> {
+      try {
+        CommitMeta commitMeta = ser.fromBytes(in);
+        String currUser =
```
snazy, line 319: This whole assignment could be externalized to a separate method to match the behavior of the commit-code-path.
I.e. both should do the same thing for setting the committer + commitTime.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Is this mapping of properties considered as critical to fail the merge/translate operation?

Just want to confirm before I make that change.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Can you please elaborate, this is already an abstracted method, referred from both transplant and merge.

snazy: Unify with the code-path for the commit-operation + merge + transplant.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
snazy: Either it fails or succeeds. But failing with a logged warning that probably nobody looks at is not good IMO.
If serialization fails, then something is very broken, right?

harshm-dremio:
@snazy @nastra Changing the commit metadata is causing a deviation in the hash calculation here - [AbstractDatabaseAdapter.java#L581](https://github.com/projectnessie/nessie/blob/20d6ec68b392382ada33386b7661c0b52d1f71d3/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java#L581)
This deviation then in-turn leads to incorrect calculation of parents while making a merge.

There are two approaches I could think of -
a. Hash calculation to ignore updatable properties.
b. Track referred hash, and use that for matching purposes.

However, both of these solutions don't appear clean to me. Please suggest if you can think of something else.

snazy:
> Changing the commit metadata is causing a deviation in the hash calculation here

Sure, you change the contents of the commits.
You probably have to adopt the call-sites of `copyCommits` as well, so the updated commit-entries are persisted (your current PR does not).

python/requirements.txt
```diff
@@ -1,5 +1,5 @@
 attrs==21.2.0
-botocore==1.21.47
+botocore==1.21.48
```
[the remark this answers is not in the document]
harshm-dremio: These are unintentional, caused during rebase, I'll fix them.

harshm-dremio:
Rebased, and removed the unnecessary commits. 

Currently working on - 

- Incorporate feedback.
- Handle deviations in hash references when commit entry is modified.
- Tests: fix failing tests + write new client side tests

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractVersionStoreTest.java
```diff
@@ -527,7 +527,8 @@ void run(VersionStore<String, String, StringStoreWorker.TestEnum> store)
                     s.transplant(
                         BranchName.of("this-one-should-not-exist"),
                         Optional.empty(),
-                        singletonList(s.toHash(BranchName.of("main"))))),
+                        singletonList(s.toHash(BranchName.of("main"))),
+                        (a, b) -> b)),
```
snazy, line 531: `a` and `b` are rather confusing. Would be good to have better names.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractITVersionStore.java
```diff
@@ -904,6 +966,18 @@ protected void checkTransplantWithNoExpectedHash() throws VersionStoreException
               Optional.empty(),
               Optional.of("v4_1"),
               Optional.of("v5_1"));
+      List<WithHash<String>> commits = commitsList(newBranch);
+      store()
+          .getCommits(newBranch)
```
snazy, line 971: Must use `commitsList` or put the `Stream` into a try-with-resources.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractITVersionStore.java
```diff
@@ -904,6 +966,18 @@ protected void checkTransplantWithNoExpectedHash() throws VersionStoreException
...
+          .map(w -> w.getValue())
+          .filter(m -> !m.equals("Another commit"))
+          .forEach(m -> assertMetaIsReset(m));
```
snazy, line 974: I assume, this can be refactored into a better readable assertion w/o repetitive assertTrue/assertFalse

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractITVersionStore.java
```diff
@@ -953,10 +1029,23 @@ protected void transplantBasic() throws VersionStoreException {
...
-              newBranch, Optional.of(initialHash), Arrays.asList(firstCommit, secondCommit));
+              newBranch,
+              Optional.of(initialHash),
+              Arrays.asList(firstCommit, secondCommit),
+              transformMeta);
       assertThat(
               store().getValues(newBranch, Arrays.asList(Key.of("t1"), Key.of("t4"), Key.of("t5"))))
           .contains(Optional.of("v1_2"), Optional.of("v4_1"), Optional.of("v5_1"));
+      store()
+          .getCommits(newBranch)
```
snazy, line 1040: Same as above

harshm-dremio:
@snazy @nastra can you please give this PR another look.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +303,26 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
+              toHash(hash, true),
+              getResetMergePropsFunc(MergeOp.MERGE, merge.getFromRefName()));
     } catch (ReferenceNotFoundException e) {
       throw new NessieNotFoundException(e.getMessage(), e);
     } catch (ReferenceConflictException e) {
       throw new NessieConflictException(e.getMessage(), e);
     }
   }
 
+  private BiFunction<Serializer<CommitMeta>, ByteString, ByteString> getResetMergePropsFunc(
```
snazy, line 315: Better move the method into the enum

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -418,19 +450,15 @@ protected Hash doOps(
...
-      throws NessieConflictException {
-    if (commitMeta.getCommitter() != null) {
-      throw new NessieConflictException(
-          "Cannot set the committer on the client side. It is set by the server.");
-    }
-    String committer = principal == null ? "" : principal.getName();
-    Instant now = Instant.now();
-    return commitMeta.toBuilder()
+  private CommitMeta meta(final CommitMeta referredCommit, Map<String, String> additionalProps) {
+    final String committer = Optional.ofNullable(getPrincipal()).map(p -> p.getName()).orElse("");
```
snazy, line 454: Let's not add `final` here. It's superfluous.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -405,11 +433,15 @@ protected Hash doOps(
...
+      if (commitMeta.getCommitter() != null) {
+        throw new NessieConflictException(
+            "Cannot set the committer on the client side. It is set by the server.");
+      }
       return getStore()
           .commit(
               BranchName.of(Optional.ofNullable(branch).orElse(getConfig().getDefaultBranch())),
               Optional.ofNullable(hash).map(Hash::of),
-              meta(getPrincipal(), commitMeta),
+              meta(commitMeta, Collections.EMPTY_MAP),
```
snazy, line 444: nit: `Collections.emptyMap()`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -418,19 +450,15 @@ protected Hash doOps(
...
-  private static CommitMeta meta(Principal principal, CommitMeta commitMeta)
-      throws NessieConflictException {
-    if (commitMeta.getCommitter() != null) {
-      throw new NessieConflictException(
-          "Cannot set the committer on the client side. It is set by the server.");
-    }
-    String committer = principal == null ? "" : principal.getName();
-    Instant now = Instant.now();
-    return commitMeta.toBuilder()
+  private CommitMeta meta(final CommitMeta referredCommit, Map<String, String> additionalProps) {
```
snazy, line 453: `final` is superfluous

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -418,19 +450,15 @@ protected Hash doOps(
...
+    final String committer = Optional.ofNullable(getPrincipal()).map(p -> p.getName()).orElse("");
+    final Instant now = Instant.now();
+    return referredCommit.toBuilder()
         .committer(committer)
         .commitTime(now)
-        .author(commitMeta.getAuthor() == null ? committer : commitMeta.getAuthor())
-        .authorTime(commitMeta.getAuthorTime() == null ? now : commitMeta.getAuthorTime())
+        .author(referredCommit.getAuthor() == null ? committer : referredCommit.getAuthor())
```
snazy, line 459: Why is this needed with `referredCommit.toBuilder()`?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
snazy: The name `referredCommit` is a bit misleading, because we're not _referring_ to it.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
```diff
@@ -89,4 +89,8 @@ static CommitLogEntry of(
         .addAllKeyListsIds(keyListIds)
         .build();
   }
+
+  static CommitLogEntry withNewMeta(CommitLogEntry sourceCommit, ByteString metadata) {
```
snazy, line 93: Can simplify by just adding a `CommitLogEntry withMetadata(ByteString metadata)` to the `CommitLogEntry` interface. (see e.g. `DatabaseAdapterConfig` interface)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -150,7 +152,10 @@ Hash commit(CommitAttempt commitAttempt)
    *     expected hEAD
    */
   Hash transplant(
-      BranchName targetBranch, Optional<Hash> expectedHead, List<Hash> sequenceToTransplant)
+      BranchName targetBranch,
+      Optional<Hash> expectedHead,
+      List<Hash> sequenceToTransplant,
+      Function<ByteString, ByteString> resetWithMergeProps)
```
snazy, line 158: Can you replace `resetWithMergeProps` here and elsewhere with `updateCommitMetadata`?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -861,6 +876,18 @@ protected void checkForModifiedKeysBetweenExpectedAndCurrentCommit(
...
+  /**
+   * Updates the commit log entry, against the commit log hash. All values of the given {@link
+   * CommitLogEntry} can be considered valid and consistent. Since, it is an override operation, the
+   * last update will be final version of the entry.
+   *
+   * @param ctx
+   * @param entry
+   * @throws ReferenceNotFoundException
+   */
+  protected abstract void overrideCommitEntry(OP_CONTEXT ctx, CommitLogEntry entry)
```
snazy, line 888: Definitely do not overwrite/update/modify any existing commit log entry. This would change the *original* commit -log-entry, right?
Merge + transplant write _new_ commit-log-entries.

snazy (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1047,6 +1075,15 @@ protected Hash copyCommits(
       } else {
         parents.add(0, targetHead);
       }
+      ByteString commitMetadata = resetWithMergeProps.apply(sourceCommit.getMetadata());
+      sourceCommit = CommitLogEntry.withNewMeta(sourceCommit, commitMetadata);
+      overrideCommitEntry(ctx, sourceCommit);
```
snazy, line 1080: This call doesn't make sense, no?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1047,6 +1075,15 @@ protected Hash copyCommits(
       } else {
         parents.add(0, targetHead);
       }
+      ByteString commitMetadata = resetWithMergeProps.apply(sourceCommit.getMetadata());
+      sourceCommit = CommitLogEntry.withNewMeta(sourceCommit, commitMetadata);
```
snazy, line 1079: Why this?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1060,15 +1097,15 @@ protected Hash copyCommits(
               newKeyLists);
       keyListDistance = newEntry.getKeyListDistance();
 
-      if (!newEntry.getHash().equals(sourceCommit.getHash())) {
+      if (!newEntry.getHash().equals(sourceCommitValueHash)) {
```
snazy, line 1100: Why this change and the ones below?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: This is around the concern I raised earlier - 

> @snazy @nastra Changing the commit metadata is causing a deviation in the hash calculation here - AbstractDatabaseAdapter.java#L581. This deviation then in-turn leads to incorrect calculation of parents while making a merge.
> There are two approaches I could think of -
> a. Hash calculation to ignore updatable properties.
> b. Track referred hash, and use that for matching purposes.
> However, both of these solutions don't appear clean to me. Please suggest if you can think of something else.

I believe I misunderstood your [comment](https://github.com/projectnessie/nessie/pull/2146#issuecomment-927941173) properly, assuming suggestion was to change and persist - 
> Sure, you change the contents of the commits.
> You probably have to adopt the call-sites of copyCommits as well, so the updated commit-entries are persisted (your current PR does not). 


Let me take another attempt to explain the problem, if we take following example - 
image
Before this change, if there's a second merge from a different branch, only commits that have to be merged are the ones after first merge and the merge hash. The hash is used to do this match. Now that metadata is different, the hash at merge commit (in the main branch) is also different and the matching algorithm doesn't know the point of last merge and as a result, now we try to pull in all the commits from the incoming branch (which is incorrect, guardrail catches the conflict).

Out of the approaches, that I initially thought of - 

- a. Hash calculation to ignore updatable properties -> since metadata is generic in the adapter classes, we should rule out selective serialisation.
- b. Track referred hash, and use that for matching purposes -> May be add a "mergeHash" in CommitEntry, and enhance the matching logic to check if the incoming hash is equal to either mergeHash or commit hash.

Please let me know your thoughts. If you are available for a short 10 mins call today discussing this problem, it would be helpful.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Taking some input parameters, hence the function

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: I guess it was a matter of habit that I picked to make use of final keyword liberally, thinking of the advantages and the intent (that a variable not supposed to be reinitialised should be declared final). 

I'll remove it, and try to be consistent with the rest of the code.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Isn't it same?

```
 public static final <K, V> Map<K, V> emptyMap() {
        return EMPTY_MAP;
    }
```

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: I'll refactor it to `from`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: It's existing code, let me know if it is unintended - https://github.com/projectnessie/nessie/blob/c0f2530649c240e89b68a45157bd068c38401178/servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java#L431

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: Same reason as described here - https://github.com/projectnessie/nessie/pull/2146#discussion_r719170422

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: Because of  - https://github.com/projectnessie/nessie/pull/2146#discussion_r719170422

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: Because of https://github.com/projectnessie/nessie/pull/2146#discussion_r719170422

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: copyCommits does exactly what the method name suggests: it **copies** commits - means: it creates new commit-log-entries. Why update if those are new??

Again: we will **never** update commit-log-entries - those are immutable.

harshm-dremio: I ran into an issue, which happens when there are multiple merges from a branch. Let me come up with a solution.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: The newEntry is created with reset version of commit metadata. The hash won't match in intended cases with the original source metadata hash. Hence, recomputing the source metadata hash and use the same.

harshm-dremio:
@snazy can you please review

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
snazy: It is not. Generics.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -418,19 +445,15 @@ protected Hash doOps(
...
 
-  private static CommitMeta meta(Principal principal, CommitMeta commitMeta)
-      throws NessieConflictException {
-    if (commitMeta.getCommitter() != null) {
-      throw new NessieConflictException(
-          "Cannot set the committer on the client side. It is set by the server.");
-    }
-    String committer = principal == null ? "" : principal.getName();
+  private CommitMeta meta(CommitMeta commitMeta, Map<String, String> additionalProps) {
+    String committer = Optional.ofNullable(getPrincipal()).map(p -> p.getName()).orElse("");
```
snazy, line 449: Rather leave the ternary as it was before.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -123,13 +125,17 @@ Hash commit(
    * @param referenceHash The hash to use as a reference for conflict detection. If not present, do
    *     not perform conflict detection
    * @param sequenceToTransplant The sequence of hashes to transplant.
+   * @param updateCommitMetadata Function to allow resetting properties application due to merge.
```
snazy, line 128: Please reword. "resetting properties application" is confusing. `merge` is wrong.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -150,12 +156,17 @@ void transplant(
    * @param fromHash The hash we are using to get additional commits
    * @param toBranch The branch that we are merging into
    * @param expectedHash The current head of the branch to validate before updating (optional).
+   * @param updateCommitMetadata Function to allow resetting properties application due to merge.
```
snazy, line 159: Please reword. "resetting properties application" is confusing.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -141,6 +142,7 @@ Hash commit(CommitAttempt commitAttempt)
    *     defaults to the named reference's HEAD.
    * @param expectedHead if present, {@code target}'s current HEAD must be equal to this value
    * @param sequenceToTransplant commits in {@code source} to cherry-pick onto {@code targetBranch}
+   * @param updateCommitMetadata Function to allow resetting properties application due to merge.
```
snazy, line 145: Please reword. "resetting properties application" is confusing. `merge` is wrong.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -163,6 +168,7 @@ Hash transplant(
    * @param from commit-hash to start reading commits from.
    * @param toBranch target branch to commit to
    * @param expectedHead if present, {@code toBranch}'s current HEAD must be equal to this value
+   * @param updateCommitMetadata Function to allow resetting properties application due to merge.
```
snazy, line 171: Please reword. "resetting properties application" is confusing.

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/SqlStatements.java
```diff
@@ -92,6 +92,8 @@
       String.format("DELETE FROM %s WHERE key_prefix = ?", TABLE_COMMIT_LOG);
   public static final String INSERT_COMMIT_LOG =
       String.format("INSERT INTO %s (key_prefix, hash, value) VALUES (?, ?, ?)", TABLE_COMMIT_LOG);
+  public static final String UPDATE_COMMIT_LOG =
```
snazy, line 95: Why is this still here?

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +299,25 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
+              toHash(hash, true),
+              updateMetaWithMergePropsFunc(MergeOp.MERGE, merge.getFromRefName()));
     } catch (ReferenceNotFoundException e) {
       throw new NessieNotFoundException(e.getMessage(), e);
     } catch (ReferenceConflictException e) {
       throw new NessieConflictException(e.getMessage(), e);
     }
   }
 
+  private BiFunction<Serializer<CommitMeta>, ByteString, ByteString> updateMetaWithMergePropsFunc(
```
snazy, line 311: This should be moved into the enum. The enum's is actually not needed at all wrt to the comment below.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
```diff
@@ -64,24 +78,119 @@ void transplant() throws Exception {
...
     assertThatThrownBy(
-            () -> databaseAdapter.transplant(conflict, Optional.empty(), Collections.emptyList()))
+            () ->
+                databaseAdapter.transplant(
+                    conflict, Optional.empty(), Collections.emptyList(), Function.identity()))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessage("No hashes to transplant given.");
   }
 
+  @Test
```
snazy, line 106: Both new tests duplicate a lot of code from `merg()` and `transplant()` and should be refactored to parameterized tests.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractITVersionStore.java
```diff
@@ -57,9 +62,32 @@
...
 public abstract class AbstractITVersionStore {
+  protected static BiFunction<Serializer<String>, ByteString, ByteString> NOOP = (s, b) -> b;
 
   protected abstract VersionStore<String, String, StringStoreWorker.TestEnum> store();
 
+  // Puts the metadata through a transformation, that appends a suffix
+  protected BiFunction<Serializer<String>, ByteString, ByteString> transformMeta =
+      (ser, inBytes) -> ser.toBytes(ser.fromBytes(inBytes) + "_suffix");
+
+  protected void assertMetaIsReset(
```
snazy, line 73: The term `Reset` is rather confusing and differs from `unchanged` elsewhere.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
```diff
@@ -287,14 +299,25 @@ public void mergeRefIntoBranch(String branchName, String hash, Merge merge)
...
+      MergeOp mergeOp, String fromRefName) {
+    return (ser, in) -> {
+      CommitMeta commitMeta = ser.fromBytes(in);
+      String mergeProp = String.format("%s_ref", mergeOp.name().toLowerCase());
```
snazy, line 315: Let's not add that property using this PR. It's a different scope and we are not yet sure whether it should be added at all.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
```diff
@@ -36,6 +36,8 @@
   /** Serialized commit-metadata. */
   ByteString getMetadata();
 
+  CommitLogEntry withMetadata(ByteString metadata);
```
snazy, line 39: Is this being used?

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -123,13 +125,17 @@ Hash commit(
...
    *     values for {@code branch}
    * @throws ReferenceNotFoundException if {@code branch} or if any of the hashes from {@code
    *     sequenceToTransplant} is not present in the store.
    */
   void transplant(
-      BranchName targetBranch, Optional<Hash> referenceHash, List<Hash> sequenceToTransplant)
+      BranchName targetBranch,
+      Optional<Hash> referenceHash,
+      List<Hash> sequenceToTransplant,
+      BiFunction<Serializer<METADATA>, ByteString, ByteString> updateCommitMetadata)
```
snazy, line 138: This should be a `Function<METADATA, METADATA>`

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -150,12 +156,17 @@ void transplant(
...
    *     for {@code toBranch}
    * @throws ReferenceNotFoundException if {@code toBranch} or {@code fromHash} is not present in
    *     the store.
    */
-  void merge(Hash fromHash, BranchName toBranch, Optional<Hash> expectedHash)
+  void merge(
+      Hash fromHash,
+      BranchName toBranch,
+      Optional<Hash> expectedHash,
+      BiFunction<Serializer<METADATA>, ByteString, ByteString> updateCommitMetadata)
```
snazy, line 169: This should be a `Function<METADATA, METADATA>`

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
snazy, line 311: This function should not do any serialization.

snazy (changes requested):

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Not sure if I got the concern.
The serialisation doesn't happen on this invocation, it just returns the lambda.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
harshm-dremio: The serialiser is available with the StoreWorker, and appears to be available at an implementation layer of the version store `storeWorker.getMetadataSerializer()`. However, the logic to update CommitMeta has to be at TreeResource. 
Hence, expecting a serialiser as input, and supplied by the VersionStore implementation.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
harshm-dremio: The serialiser is available with the StoreWorker, and appears to be available at an implementation layer of the version store storeWorker.getMetadataSerializer(). However, the logic to update CommitMeta has to be at TreeResource.
Hence, expecting a serialiser as input, and supplied by the VersionStore implementation.

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: ok

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/SqlStatements.java
harshm-dremio: oops, will revert

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
harshm-dremio: ok

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
harshm-dremio: ok

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
harshm-dremio: ok

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Should Principal be supplied as a parameter? If yes, why since it is available as a class level function?

I can move the implementation to following variant if you feel that's easier to read - 
```
    Principal principal = getPrincipal();
    String committer = principal == null ? "" : principal.getName();
```

servers/services/src/main/java/org/projectnessie/services/rest/TreeResource.java
harshm-dremio: Removing the code related to properties, this param will not be needed

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
snazy: Serialization is a concern of the implementation, not the `VersionStore` interface.

harshm-dremio:
@snazy pls review

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1060,15 +1084,15 @@ protected Hash copyCommits(
...
         commitsChronological.set(i, newEntry);
+        targetHead = newEntry.getHash();
       } else {
         // Newly built CommitLogEntry is equal to the CommitLogEntry to transplant.
         // This can happen, if the commit to transplant has NO_ANCESTOR as its parent.
         commitsChronological.remove(i);
+        targetHead = sourceCommit.getHash();
```
snazy, line 1094: This assignment is very dangerous and can lead to data corruption.
The copied commit _will be deleted_ if the transplant/merge ran into an issue later. With this assignment, you tell it to delete the source-commit, which must not happen.

snazy (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
harshm-dremio: IMO this part of the code is for cases, when we choose "not to copy" the commits. As per my understanding, the else block applies to cases, when sourceCommit and newEntry arrive at exactly same definition. The if condition prior to this change can be a good ref - `newEntry.getHash().equals(sourceCommit.getHash())` 

With the commitUpdate function is triggered, the newEntry uses updated meta and the computed hash value can be different from source commit hash. To retain the behaviour, we are matching the commit definition by checking a recomputed source hash.
In this particular case, the updated commit metadata isn't saved anywhere because copy didn't happen. Hence, we point it back to sourceCommit's hash as it will be used as the targetHead.

> The copied commit will be deleted if the transplant/merge ran into an issue later. With this assignment, you tell it to delete the source-commit, which must not happen.
Looking at the code, I'm not able to visualise any scenario where a source commit would be deleted. Can you please elaborate.

harshm-dremio: Well, I mentioned this earlier - 

> @snazy @nastra Changing the commit metadata is causing a deviation in the hash calculation here - [AbstractDatabaseAdapter.java#L581](https://github.com/projectnessie/nessie/blob/20d6ec68b392382ada33386b7661c0b52d1f71d3/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java#L581)

newEntry is now constructed with an updated commitMeta, hence it'll have a different hash than source commit.
> With the commitUpdate function is triggered, the newEntry uses updated meta and the computed hash value can be different from source commit hash.

To retain the definition match, we're evaluating - ( `updateCommitMeta(commitsChronological.get(i))::hash` against `newEntry.getHash()`.  If copy isn't required, `updateCommitMeta` isn't required either, and targetHead is set to `commitsChronological.get(i).getHash()`

harshm-dremio: `sourceCommit.getHash()` is the original hash whereas `sourceCommitValueHash` is computed from the object values. I can create a new variable `updatedSourceCommit` if that would reduce the confusion.

The `targetHead` has to be set to the original commit right? `sourceCommitValueHash/newEntry.getHash()` aren't representing the same.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1047,6 +1063,14 @@ protected Hash copyCommits(
       } else {
         parents.add(0, targetHead);
       }
+      ByteString commitMetadata = updateCommitMetadata.apply(sourceCommit.getMetadata());
+      sourceCommit = sourceCommit.withMetadata(commitMetadata);
+      final Hash sourceCommitValueHash =
```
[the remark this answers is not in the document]
harshm-dremio: Recomputing the hash here after the commitMeta updated; used at #1087

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1047,6 +1063,14 @@ protected Hash copyCommits(
       } else {
         parents.add(0, targetHead);
       }
+      ByteString commitMetadata = updateCommitMetadata.apply(sourceCommit.getMetadata());
```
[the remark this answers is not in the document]
snazy: Why is that needed? Can you please try just the simple change?

harshm-dev:
ping

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -402,11 +407,8 @@ protected Hash doOps(
     }
   }
 
-  private static CommitMeta meta(Principal principal, CommitMeta commitMeta) {
-    if (commitMeta.getCommitter() != null) {
```
rymurr, line 406: why remove this check? Was it moved elsewhere?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -174,7 +182,11 @@ Hash transplant(
    *     branch due to a conflicting change or if the expected hash of {@code toBranch} is not its
    *     expected hEAD
    */
-  Hash merge(Hash from, BranchName toBranch, Optional<Hash> expectedHead)
+  Hash merge(
```
rymurr, line 185: I don't think its the responsibility of the caller to decide how commit metadata gets updated by a transplant or merge. I think we want to only update the commit time and committer when a commit has been transplanted so I think we really only need to pass down the principal

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
harshm-dev: Yes, at line 394

We're re-using this method for merge/transplant, but the check is specific to the new commits.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
harshm-dev: By the existing design of this code, the lower layers of code treat metadata as a blackbox - For VersionStore, METADATA is generic, and for DBAdapter, metadata is a ByteString.

The main reason for the approach, where caller dictates the transformation logic, is because only the caller is aware of the `CommitMeta` object structure. 
The attempt here is to align with the current state, and not deviate the dependency graph across modules. Setting the principal in the VersionStore/DBAdapter would otherwise, be required to understand the model class - CommitMeta.

rymurr: thats valid. I am torn here on one hand the reason for updating the metadata on transplant/merge is to ensure strong auditing (same as only setting principal on the server side) on the other hand you are correct we don't want to expose internal info about commit meta to the commit kernel.

Maybe I would feel more comfortable with a method to modify commit meta that was only called by transplant/merge. I think it would be easier to audit and could have a big warning comment about what is allowed to be changed in that method. Thoughts?

harshm-dev: The method essentially feeds commit time properties in the meta object, this is applicable for all types of commits (merge/transplant, as well as the regular ones). May be we should rename the method to exactly say what it does - `setCommitProperties` with a big warning comment preventing people from not doing anything else in there. WDYT?

snazy:
Superseded by #3122

PR #2149, opened 2021-09-25
Add MongoDB Version Store to Quarkus Server

### LINKED ISSUE #2134: Add MongoDB to Quarkus
MongoDB shoiuld be available as a backend database via Nessie-Quarkus.
If the creation of the MongoDB client is cheap (i.e. no expensive connection-pool initialization), consider adding it to the Nessie-Lambda module as well.

/cc @dimas-b

dimas-b:
Closes #2134

---
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/2149)

### DIFF
--- a/servers/jax-rs-tests/pom.xml
+++ b/servers/jax-rs-tests/pom.xml
@@ -89,5 +89,10 @@
       <version>${dockerjava.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>de.flapdoodle.embed</groupId>
+      <artifactId>de.flapdoodle.embed.mongo</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -232,6 +232,43 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-versioned-persist-mongodb</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mongodb</groupId>
+      <artifactId>mongodb-crypt</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-unixsocket</artifactId>
+      <version>0.38.11</version>
+    </dependency>
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <version>1.4.8-1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
+      <version>1.1.8.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-versioned-persist-mongodb</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>de.flapdoodle.embed</groupId>
+      <artifactId>de.flapdoodle.embed.mongo</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -239,23 +239,8 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.mongodb</groupId>
-      <artifactId>mongodb-crypt</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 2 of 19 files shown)


servers/quarkus-server/pom.xml
```diff
@@ -200,6 +200,43 @@
...
+    <dependency>
+      <groupId>com.github.luben</groupId>
+      <artifactId>zstd-jni</artifactId>
+      <version>1.4.8-1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
+      <version>1.1.8.4</version>
+    </dependency>
```
dimas-b, line 227: This is required for the native image build to be able to resolve java linkages. 

IDK why Quarkus cannot infer that from Maven manifests.

snazy: Have you tried to use the `quarkus-mongodb-client` module?

BTW: Versions should all be defined in the top-level pom

servers/quarkus-server/src/main/java/org/projectnessie/server/config/VersionStoreConfig.java
```diff
@@ -52,4 +53,13 @@
     @WithName("db-path")
     String getDbPath();
   }
+
+  @ConfigMapping(prefix = "nessie.version.store.mongo")
+  interface MongoVersionStoreConfig {
+    @WithName("connect-to")
+    String getConnectionString();
```
snazy, line 60: Maybe the `quarkus-mongodb-client` module already defines something that makes this superfluous.

servers/quarkus-server/src/test/java/org/projectnessie/server/ITNativeRestApiMongo.java
```diff
@@ -0,0 +1,35 @@
...
+import org.projectnessie.jaxrs.AbstractTestRest;
+import org.projectnessie.server.profiles.QuarkusNativeProfileMongo;
+
+@NativeImageTest
+@TestProfile(QuarkusNativeProfileMongo.class)
+public class ITNativeRestApiMongo extends AbstractTestRest {
+
+  @Override
+  @BeforeEach
+  public void setUp() throws Exception {
```
snazy, line 31: Feels like we should introduce some `AbstractQuarkusTestRest` class, because the `setUp` method's duplicated a lot.

servers/jax-rs-tests/pom.xml
```diff
@@ -77,5 +77,23 @@
...
+      <version>${project.version}</version>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-versioned-persist-mongodb</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
```
snazy, line 92: Doesn't this need dependencies to testcontainers as well?

dimas-b: Works "as is" :shrug:

servers/quarkus-server/pom.xml
dimas-b: Yes, that helped - thanks! 8a1ee77c980aad7108562e2ce8294f30b9dd7557

servers/quarkus-server/src/main/java/org/projectnessie/server/config/VersionStoreConfig.java
dimas-b: It does, but it will then configure its own instance of `MongoClient`. We currently have non-quarkus environments where we create our own `MongoClient`... Do you think it's worth refactoring the code to support both a locally created `MongoClient` and an injected one?

snazy: Yea, it's probably worth to refactor the Quarkus-code to use the `MongoClient` produced by Quarkus.
Ran into the same issue for the Dynamo-adapter It felt really strange to have (a ton of) configuration options via Quarkus but only use a subset of those (authentication, endpoint-detection, aws-http-client config, etc). So I changed the Dynamo-PR that it uses the DynamoDbClient produced by Quarkus, but use the one "produced by us" in other places.

servers/quarkus-server/src/test/java/org/projectnessie/server/ITNativeRestApiMongo.java
dimas-b: Sure. I'll do this is a smaller PR presently

snazy:
FYI #2162

servers/quarkus-server/src/test/java/org/projectnessie/server/ITNativeRestApiMongo.java
dimas-b: #2164

servers/quarkus-server/src/main/java/org/projectnessie/server/config/VersionStoreConfig.java
dimas-b: refactored in the latest commit (also rebased)

snazy (commented):
LGTM

Just need to get CI passing ;)

PR #2176, opened 2021-09-28
Configure database adapters via quarkus server config

### LINKED ISSUE #2148: Support configuring database adapters via Quarkus properties
Currently database adapters created automatically by setting `nessie.version.store.type=...` get default values for all of their "common" properties, e.g. "key prefix".

Nessie should have a generic mechanism for configuring database adapters via Quarkus properties.

dimas-b:
* Move the "connection provider" from adapter config to adapter factory builder

* Remove "with" methods from database adapter config classes because Quarkus does not allow config interfaces to have methods with parameters

* Add QuarkusDatabaseAdapterConfig to hold Quarkus config annotations

* Use the TrimmedStringConverter for the "key-prefix" property because it can be an empty string, but the default converter will turn empty strings into `null`.

Closes #2148


servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,70 @@
...
+import io.smallrye.config.WithName;
+import org.projectnessie.versioned.persist.nontx.NonTransactionalDatabaseAdapterConfig;
+
+@ConfigMapping(prefix = "nessie.adapter")
+@RegisterForReflection(targets = TrimmedStringConverter.class)
+public interface QuarkusDatabaseAdapterConfig extends NonTransactionalDatabaseAdapterConfig {
+
+  @WithName("key-prefix")
+  @WithDefault(DEFAULT_KEY_PREFIX)
+  @WithConverter(TrimmedStringConverter.class)
```
snazy, line 32: Mind adding an inline comment why `TrimmedStringConverter` is used here?

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/DynamoVersionStoreFactory.java
```diff
@@ -34,31 +35,31 @@
...
   }
 
   @Override
   public <VALUE, METADATA, VALUE_TYPE extends Enum<VALUE_TYPE>>
       VersionStore<VALUE, METADATA, VALUE_TYPE> newStore(
           StoreWorker<VALUE, METADATA, VALUE_TYPE> worker, ServerConfig serverConfig) {
 
+    DynamoDatabaseClient client = new DynamoDatabaseClient();
+    client.configure(
+        ImmutableProvidedDynamoClientConfig.builder().dynamoDbClient(dynamoDbClient).build());
```
snazy, line 55: Can you move the verbose `Immutable...` code into a static method `ProvidedDynamoClientConfig.of(DynamoDbClient dynamoDbClient)` (and similar for the other version store types)?

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/InMemVersionStoreFactory.java
```diff
@@ -30,14 +32,17 @@
...
   @Override
   public <VALUE, METADATA, VALUE_TYPE extends Enum<VALUE_TYPE>>
       VersionStore<VALUE, METADATA, VALUE_TYPE> newStore(
           StoreWorker<VALUE, METADATA, VALUE_TYPE> worker, ServerConfig serverConfig) {
     DatabaseAdapter databaseAdapter =
         new InmemoryDatabaseAdapterFactory()
             .newBuilder()
-            .configure(c -> c.withConnectionProvider(new InmemoryStore()))
+            .withConfig(config)
+            .withConnector(new InmemoryStore())
```
snazy, line 45: I think, this `InmemoryStore` instance needs to be injected (so Quarkus maintains a singleton instance).

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/DynamoVersionStoreFactory.java
```diff
@@ -34,31 +35,31 @@
...
     this.dynamoDbClient = dynamoDbClient;
+    this.config = config;
   }
 
   @Override
   public <VALUE, METADATA, VALUE_TYPE extends Enum<VALUE_TYPE>>
       VersionStore<VALUE, METADATA, VALUE_TYPE> newStore(
           StoreWorker<VALUE, METADATA, VALUE_TYPE> worker, ServerConfig serverConfig) {
 
+    DynamoDatabaseClient client = new DynamoDatabaseClient();
```
snazy, line 53: The `DynamoDatabaseClient` (and similar for the other store types) would be clearer managed via Quarkus by a `@Producer` bean and then inject `DynamoDatabaseClient` into this class. WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
```diff
@@ -33,44 +33,53 @@
  * @param <CONFIG> the configuration-options type used to configure database-adapters produced by
  *     this factory.
  */
-public interface DatabaseAdapterFactory<CONFIG extends DatabaseAdapterConfig<?>> {
-  Builder<CONFIG> newBuilder();
+public interface DatabaseAdapterFactory<CONFIG, CONNECTOR> {
```
snazy, line 36: Why remove the `extends DatabaseAdapterConfig<?>` / why no `extends` for `CONNECTOR`?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/TestConnectionProviderSource.java
```diff
@@ -91,15 +91,7 @@ boolean isCompatibleWith(
...
-  /**
-   * Updates the given {@code config} with the {@link DatabaseConnectionProvider} via {@link
-   * DatabaseAdapterConfig#withConnectionProvider(DatabaseConnectionProvider)}.
-   *
-   * @param config the current configuration
-   * @return {@code config} with a valid {@link DatabaseConnectionProvider} instance.
-   */
-  DatabaseAdapterConfig<DatabaseConnectionProvider<CONN_CONFIG>> updateConfig(
-      DatabaseAdapterConfig<DatabaseConnectionProvider<CONN_CONFIG>> config);
+  DatabaseConnectionProvider<CONN_CONFIG> getConnectionProvider();
```
snazy, line 94: A javadoc for this method would be nice - and outline the difference of `createConnectionProvider` vs `getConnectionProvider`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -43,11 +43,9 @@
    */
   @Value.Default
   default String getKeyPrefix() {
-    return "";
+    return DEFAULT_KEY_PREFIX;
   }
 
-  DatabaseAdapterConfig<T> withKeyPrefix(String keyPrefix);
```
snazy, line 49: Isn't there a way to tell Smallrye-Config/Quarkus to ignore these methods?

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,70 @@
...
+import io.quarkus.runtime.configuration.TrimmedStringConverter;
+import io.smallrye.config.ConfigMapping;
+import io.smallrye.config.WithConverter;
+import io.smallrye.config.WithDefault;
+import io.smallrye.config.WithName;
+import org.projectnessie.versioned.persist.nontx.NonTransactionalDatabaseAdapterConfig;
+
+@ConfigMapping(prefix = "nessie.adapter")
+@RegisterForReflection(targets = TrimmedStringConverter.class)
+public interface QuarkusDatabaseAdapterConfig extends NonTransactionalDatabaseAdapterConfig {
```
snazy, line 28: How do we get the config for `TxDatabaseAdapterConfig` here?
I also would not mind to have a producer-bean for `NonTransactionalDatabaseAdapterConfig` and `TxDatabaseAdapterConfig` that takes Quarks-specific config instances via `@Inject`, if that's helpful.

snazy (commented):
I like this approach!
Few comments though, but the direction's clear and looks good!

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: Hmm... I thought I did :) might have lost it along the way :)

Basically, this is to avoid empty strings being converted to `null`... will put in comments as well.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
dimas-b: This is an artifact of my local experiments... will restore

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/TestConnectionProviderSource.java
dimas-b: I'm not clear on this myself... perhaps we do not really need `createConnectionProvider` now :thinking:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
dimas-b: not sure... if we kept them how would quarkus-based sub-classes implement them? :thinking:

dimas-b: Custom code that wants to create configs from scratch can use those "with" and other builder methods from the `Immutable` implementations, I guess.

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: I was thinking we'd extend both Tx- and NonTransactional- config interfaces in one quarkus-based interfaces annotated with `@ConfigMapping`. Since all config variables are optionals/have defaults, it should not hurt to have all of them combined in one quarkus config group (i.e. `nessie.adapter`).

Another approach is to break down config groups into unrelated classes, I guess... something like
`@ConfigMapping(prefix = "nessie.adapter") QuarkusDatabaseAdapterConfig extends DatabaseAdapterConfig (extends Object)`:
* getKeyPrefix()
* etc....

`@ConfigMapping(prefix = "nessie.adapter.nontx") QuarkusNonTxDatabaseAdapterConfig extends NonTransactionalDatabaseAdapterConfig (extends Object)`:
* getParentsPerGlobalCommit()


`@ConfigMapping(prefix = "nessie.adapter.tx") QuarkusNonTxDatabaseAdapterConfig extends TxDatabaseAdapterConfig (extends Object)`:
* getBatchSize()

... but I personally like the approach where one quarkus class extends all adapter config classes. WDYT?

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/DynamoVersionStoreFactory.java
dimas-b: If we had that, quarkus would create all types of "client" objects all the time (for all backends), would it not? This way we only create the one selected by configuration.

dimas-b: Apparently I was wrong :facepalm: unused beans are not instantiated... however, `MongoClient` is always instantiated for some reason :thinking:

dimas-b: I'll refactor this tomorrow

snazy: > however, MongoClient is always instantiated for some reason

That sounds like a Quarkus bug. IIRC, Quarkus 2.2.3 has some fixes around Mongo, maybe that's been fixed as well.

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
snazy: I tend to agree (with "one thing extends all"). Maybe just have the `prefix`es so that we could split up later, if we need to?

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @WithName("key-prefix")
+  @WithDefault(DEFAULT_KEY_PREFIX)
+  @WithConverter(TrimmedStringConverter.class)
+  @Override
+  String getKeyPrefix();
+
+  @WithName("parents-per-global-commit")
```
omarsmak, line 36: Are we planning to add a comment for every config explaining the purpose of the particular config?

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: good point re: prefixes :+1:

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: We have javadoc on base methods (in superclasses)

omarsmak: I see, nevertheless. isn't easier to have some docs here that explains the purpose of the config class when you read the config interface? Also, IMHO it will make our life easier when we decide to automate generating the config docs through Quarkus `@ConfigMapping` annotation.

dimas-b: I'd expect Quarkus build to be able to extract javadoc from superclasses.

I'll add a top-level javadoc to this class too.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
dimas-b: type parameters narrowed in c96f9344d07e4ee78f3bd9596af9f5a366a98d47

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/DynamoVersionStoreFactory.java
dimas-b: Added client singletons in 90b8280422f3dab555be727df30151ec99be8856

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/InMemVersionStoreFactory.java
dimas-b: FIxed in 90b8280422f3dab555be727df30151ec99be8856

servers/quarkus-server/src/main/java/org/projectnessie/server/providers/DynamoVersionStoreFactory.java
dimas-b: Fixed in 90b8280422f3dab555be727df30151ec99be8856

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/TestConnectionProviderSource.java
dimas-b: I moved `createConnectionProvider` to the impl. class in 1b903edb794652ba6259d3878ad87d47eeaa3a37... hope it's clearer now.... also update javadoc.

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: updated in e094dc26344a0f68b3d7c3adba03d0d7f118280d

servers/quarkus-server/src/main/java/org/projectnessie/server/config/QuarkusDatabaseAdapterConfig.java
dimas-b: updated in e094dc26344a0f68b3d7c3adba03d0d7f118280d

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -163,6 +163,9 @@ quarkus.container-image.name=nessie
 %test.quarkus.security.users.embedded.users.test_user=test_user
 %test.quarkus.security.users.embedded.roles.admin_user=admin,user
 %test.quarkus.security.users.embedded.roles.test_user=test123
+%test.nessie.adapter.key-prefix=nessie-test
```
snazy, line 166: I wonder, whether we should move all `test` profile properties into appropriate classes (not in this PR) after 1.0.0. WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
snazy: I've added `with...` methods mostly to ease configuration using those `.configurate(c -> c)` methods. Seems, that this is mostly no longer necessary, so I'm fine with removing those.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
```diff
@@ -33,44 +33,55 @@
...
-public interface DatabaseAdapterFactory<CONFIG extends DatabaseAdapterConfig<?>> {
-  Builder<CONFIG> newBuilder();
+public interface DatabaseAdapterFactory<
+    CONFIG extends DatabaseAdapterConfig, CONNECTOR extends DatabaseConnectionProvider<?>> {
+  Builder<CONFIG, CONNECTOR> newBuilder();
 
   String getName();
 
-  abstract class Builder<CONFIG extends DatabaseAdapterConfig<?>> {
+  abstract class Builder<CONFIG, CONNECTOR> {
```
snazy, line 42: `extends` missing for the generic types?

versioned/persist/bench/src/main/java/org/projectnessie/versioned/persist/benchmarks/CommitBench.java
```diff
@@ -108,37 +108,34 @@ public void init() throws Exception {
...
-                  providerSource.start();
-                } catch (Exception e) {
-                  throw new RuntimeException(e);
-                }
-                c = providerSource.updateConfig((DatabaseAdapterConfig) c);
-                connectionProvider = c.getConnectionProvider();
-                return c;
-              })
-          .build();
+      DatabaseAdapterFactory.Builder<CONFIG, DatabaseConnectionProvider<DatabaseConnectionConfig>>
```
snazy, line 117: This part looks better now :)

servers/quarkus-server/src/main/resources/application.properties
dimas-b: +1 #2240

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
snazy: Ah, yea - it's a method not a class.

dimas-b:
rebased

snazy (approved):
+1

PR #2196, opened 2021-09-29
Declare classes for Quarkus reflection via annotation

dimas-b:
* Move the reflection declaration for NessieError from reflection-config.json
  to a java annotation.

* This is to simplify (hopefully) tracking quarkus reflection dependencies.


snazy (approved):
+1

Issue #756, opened 2021-02-02
Decide about Lambda module


rymurr:
The Nessie Lambda module is rather unmaintained. We need to decide what to do with it.

Currently:
1. the readme is out of date
2. the lambda hasn't been tested in some time
3. we don't run automated tests on the build lambda image
4. it is not published anywhere

Things to consider:
1. It is less useful for databases with "expensive" client/driver setups (long initialization time), which is actually the majority
2. It is not useful for "local" databases (RocksDB)

PR #2242, opened 2021-10-04
Support AWS auth profiles in java client

dimas-b:
Contributes to #756


clients/client/src/main/java/org/projectnessie/client/auth/AwsAuthenticationProvider.java
```diff
@@ -78,15 +84,18 @@ private static Region awsRegion(String regionName) {
...
   private static class AwsAuthentication implements HttpAuthentication {
     private final Region region;
+    private final AwsCredentialsProvider awsCredentialsProvider;
 
     @SuppressWarnings({"QsPrivateBeanMembersInspection", "CdiInjectionPointsInspection"})
-    private AwsAuthentication(Region region) {
+    private AwsAuthentication(Region region, String profile) {
       this.region = region;
+      this.awsCredentialsProvider =
+          DefaultCredentialsProvider.builder().profileName(profile).build();
```
snazy, line 93: Not sure whether it's worth to special-case `profile==null` to prevent that it may throw an NPE when they change the builder impl? WDYT?

PR #2257, opened 2021-10-05
Fix to allow re-creating the default branch

snazy:
Fixes Iceberg tests, otherwise [this line](https://github.com/apache/iceberg/blob/master/nessie/src/test/java/org/apache/iceberg/nessie/BaseTestIceberg.java#L83) never works.

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
@@ -125,13 +125,21 @@ public abstract class AbstractTestRest {
   }
 
   @BeforeEach
-  public void setUp() throws Exception {
+  public void setUp() {
     init(URI.create("http://localhost:19121/api/v1"));
   }
 
   @AfterEach
   public void tearDown() {
     api.close();
+  }
+
+  @Test
+  void createRecreateDefaultBranch() throws NessieConflictException, NessieNotFoundException {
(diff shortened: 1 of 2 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -134,6 +134,14 @@ public void tearDown() {
     api.close();
   }
 
+  @Test
+  void createRecreateDefaultBranch() throws NessieConflictException, NessieNotFoundException {
+    api.deleteBranch().branch(api.getDefaultBranch()).delete();
+
+    api.createReference().reference(Branch.of("main", null)).create();
+    api.getReference().refName("main").get();
```
omarsmak, line 142: We don't need any assertions for this test?

snazy: It shall just pass

nastra: maybe just add a `assertThat(api.getReference().refName("main").get()).isNotNull()` check

snazy: which will never return null but throw an exception, right?

nastra (approved):

PR #2258, opened 2021-10-05
Allow custom configuration via DatabaseAdapterExtension

snazy:
This is useful to inject a custom `java.time.Clock` (required for Nessie GC tests) or do other advanced things.


versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieDbAdapter.java
```diff
@@ -22,4 +22,6 @@
 
 @Target({ElementType.FIELD, ElementType.PARAMETER})
 @Retention(RetentionPolicy.RUNTIME)
-public @interface NessieDbAdapter {}
+public @interface NessieDbAdapter {
+  String configMethod() default "";
```
dimas-b, line 26: Could you add javadoc with expectations of the config (e.g. `static`, etc)?

dimas-b (approved):

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieDbAdapter.java
dimas-b: Suggestion: As an alternative we could use another annotation, e.g. `NessieDbAdapterConfigMethod`, to be used on the config method directly.... but current approach also LGTM... no preference, just an idea.

dimas-b: A minor benefit of a method-level annotation is that one can instruct the IDE not to worry about "method not used" checks with those methods.

snazy: Done. However, after the rebase I realized that removing all the `with*()` methods makes the approach impossible, because you can no longer "update" a configuration. Any idea?

snazy: > method-level annotation

Doesn't work, because you can have different DatabaseAdapter instances that require different DatabaseAdapterConfigs.

dimas-b: > However, after the rebase I realized that removing all the with*() methods makes the approach impossible, because you can no longer "update" a configuration.

The config method can still do something like `ImmutableTxDatabaseAdapterConfig.builder().from(configArg)...build()`. Will that work?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieDbAdapter.java
```diff
@@ -22,4 +22,33 @@
...
+   * </ul>
+   *
+   * <p>Example:
+   *
+   * <code><pre>
+   *   &#64;NessieDbAdapter(configMethod = "applyTestClock")
+   *   protected static DatabaseAdapter databaseAdapter;
+   *
+   *   static DatabaseAdapterConfig applyTestClock(DatabaseAdapterConfig config) {
+   *     return config.withClock(TEST_CLOCK);
```
dimas-b, line 49: How about:
```
 static TxDatabaseAdapterConfig applyTestClock(TxDatabaseAdapterConfig config) {
   return ImmutableTxDatabaseAdapterConfig.builder().from(config).clock(TEST_CLOCK).build();
}
```

snazy: But what if it’s a non-tx Adapter?
Would Quarkus complain about a `toBuilder()` function on the interfaces?

dimas-b: but, whoever uses the `NessieDbAdapter` annotation should know the expected adapter and config types, right?

dimas-b: ... not sure about `toBuilder()`... what is it's class? Quarkus-generated configs should be usage in `builder().from(config)` AFAIK

snazy: Reopened #2148 - it broke functionality.

PR #2259, opened 2021-10-05
Test fetching data via SQL extensions with enabled/disabled caching

nastra:


clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
```diff
@@ -99,8 +100,50 @@ void testCreateBranchInExists() throws NessieNotFoundException {
...
+        .hasSize(2)
+        .containsExactlyInAnyOrder(row(23, "test"), row(24, "test24"));
+  }
 
+  private static List<Object[]> sqlWithDisabledCaching(String query, Object... args) {
+    try (SparkSession sparkWithDisabledCaching = spark.newSession()) {
+      sparkWithDisabledCaching.conf().set("spark.sql.catalog.nessie.cache-enabled", "false");
+      List<Row> rows = sparkWithDisabledCaching.sql(String.format(query, args)).collectAsList();
+      return rows.stream().map(AbstractSparkTest::toJava).collect(Collectors.toList());
+    }
```
dimas-b, line 142: I'm not sure this actually validates the effect of disabled caching in a session, because this is a new session. Will it not have an empty cache to begin with?

clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
```diff
@@ -99,8 +100,50 @@ void testCreateBranchInExists() throws NessieNotFoundException {
...
+    assertThat(sql("SELECT * FROM nessie.`tbl@main`")).hasSize(1).containsExactly(row(23, "test"));
+    assertThat(sql("SELECT * FROM nessie.tbl"))
+        .hasSize(2)
+        .containsExactlyInAnyOrder(row(23, "test"), row(24, "test24"));
+
+    // this still sees old data because tbl@testBranch is being cached in Iceberg
+    // due to "spark.sql.catalog.nessie.cache-enabled=true" by default
+    assertThat(sql("SELECT * FROM nessie.`tbl@%s`", refName))
+        .hasSize(1)
+        .containsExactly(row(23, "test"));
```
dimas-b, line 129: Could this be flaky, e.g. if the cache expires for some behind-the-scenes reason?

clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
nastra: you're right. Turns out this is actually a bit more difficult to test than I thought. I changed the test to just use a fresh session with an empty cache

clients/spark-extensions/src/test/java/org/projectnessie/spark/extensions/ITNessieStatements.java
nastra: the cache is only invalidated when the table is being renamed or dropped, so shouldn't be flaky here

dimas-b (approved):

Issue #2195, opened 2021-09-29
Design squash


snazy:
Squashing multiple commits happens on two different levels:
1. In Nessie: All delete and put operations of the commits to be squashed would be collected into the squashed commit. Information in the commit-metadata would be aggregated, e.g. the commit messages concatenated. In case of Iceberg, all Iceberg snapshots would be retained as they were before, because Nessie alone does not know how to squash Iceberg snapshots.
2. Content type specific:
    1. If squashing/merging requires more logic than "just take the last on-reference-content-value", then the integration side (e.g. in Iceberg): The Squash operation happens on the integration side, probably individually for each table. The integration side e.g. Iceberg does know how "merge" (or "squash") multiple Iceberg snapshots.
    2. If squashing/merging for a content-type effectively results in "just take the last on-reference value", then this can happen in Nessie directly.

The complete process of squashing recent commits up to HEAD might result in a multi-step approach:
1. Squash requested Nessie commits into one commit containing all operations from the original commits.
2. (For case 2.i. from above) For each content-id (NOT content-key!), the corresponding integration side can then rewrite the squashed commit for each (physical) table.

Related:
* Iceberg - how to squash different snapshots, rather: how to squash different operations
* "automatic" conflict resolution for conflicting changes since the "last known hash"

snazy:
Updated the issue title from "Design squash / rebase", removed "rebase", because that's a completely different animal.

Squash means "use the latest state + (maybe) remove intermediate states"
Rebase means "replay the changes"

snazy:
Done with #3809 #3810

PR #2311, opened 2021-10-09
Explicitly define behavior of multiple operations for the same key in a commit

snazy:
While thinking about how Nessie could actually squash multiple commits, I realized
that this can result in multiple commit-operations for the same content-key.

While commit-operations are ordered on the (REST) API level (list of `Operation`s),
database adapters split those into separate collections of delete and put operations,
which makes it impossible to determine the original order of the squashed operations.

The Nessie spec does not define the behavior of a Nessie commit with multiple
operations for the same key in one commit.

The spec now clearly states that there must be no more than operation (put / delete)
for each content key.

This commit updates the spec and also adds tests on the `DatabaseAdapter` and
`VersionStore` levels and includes related fixes to `AbstractDatabaseAdapter`
and `PersistVersionStore`.

### DIFF
--- a/site/docs/develop/spec.md
+++ b/site/docs/develop/spec.md
@@ -139,8 +139,27 @@ recorded within the [_Put Operation_](#put-operation) of a Nessie commit.
 
 ## Operations in a Nessie commit
 
-Each Nessie commit carries one or more operations. Each operation contains the Content Key and
-comes in one of the following variations.
+Each Nessie commit carries one or more operations. Each operation contains the
+[Content Key](#content-key) and is either a [Put](#put-operation), [Delete](#delete-operation) or
+[Unmodified](#unmodified-operation) operation.
+
+One [Content Key](#content-key) must only occur once in a Nessie commit.
+
+Operations present in a commit are passed into Nessie as a list of operations.
+
+### Mapping SQL DDL to Nessie commit operations
+
+A `CREATE TABLE` is mapped to one _Put operation_.
+
+An `ALTER TABLE RENAME` is mapped to a _Delete operation_ using the Content Key for the table
+being renamed plus at least one _Put operation_ using the Content Key of the table's new name,
+using the Content Id of the table being renamed.
+
+A `DROP TABLE` is represented as a Nessie _Delete operation_ (without a _Put operation_ for the
+same Content Id).
+
+A `DROP TABLE` + `CREATE TABLE` using the same table name (Content Key) in a single commit are
+mapped to one _Put operation_ with a different Content Id.
 
 ### Put operation
 
@@ -158,15 +177,14 @@ definition (think: SQL DDL) or data (think: SQL DML).
 A _Delete operation_ does not carry any Content object and is used to indicate that a Content
 object is no longer referenced using the Content Key of the _Delete operation_.
 
-Example for a Nessie _Delete operation_ is an SQL `DROP TABLE`. An `ALTER TABLE RENAME` is mapped
-to a _Delete operation_ plus a _Put operation_.
-
 ### Unmodified operation
 
 An _Unmodified operation_ does not represent any change of the data, but can be included in a
 Nessie commit operation to enforce strict serializable transactions. The presence of an
 _Unmodified operation_ means that the Content object referred to via the operation's Content Key
 must not have been modified since the Nessie commit's `expectedHash`.
+
+The _Unmodified operation_ is not persisted.
 
 ## Version Store
 
--- a/site/docs/develop/spec.md
+++ b/site/docs/develop/spec.md
@@ -143,7 +143,7 @@ Each Nessie commit carries one or more operations. Each operation contains the
 [Content Key](#content-key) and is either a [Put](#put-operation), [Delete](#delete-operation) or
 [Unmodified](#unmodified-operation) operation.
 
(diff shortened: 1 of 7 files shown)


snazy:
Contributes to #2195

site/docs/develop/spec.md
```diff
@@ -139,8 +139,27 @@ recorded within the [_Put Operation_](#put-operation) of a Nessie commit.
 
 ## Operations in a Nessie commit
 
-Each Nessie commit carries one or more operations. Each operation contains the Content Key and
-comes in one of the following variations.
+Each Nessie commit carries one or more operations. Each operation contains the
+[Content Key](#content-key) and is either a [Put](#put-operation), [Delete](#delete-operation) or
+[Unmodified](#unmodified-operation) operation.
+
+One [Content Key](#content-key) must only occur once in a Nessie commit.
```
dimas-b, line 146: `A [Content Key](#content-key) must only occur at most once in a Nessie commit.`?

site/docs/develop/spec.md
```diff
@@ -158,16 +177,15 @@ definition (think: SQL DDL) or data (think: SQL DML).
...
-to a _Delete operation_ plus a _Put operation_.
-
 ### Unmodified operation
 
 An _Unmodified operation_ does not represent any change of the data, but can be included in a
 Nessie commit operation to enforce strict serializable transactions. The presence of an
 _Unmodified operation_ means that the Content object referred to via the operation's Content Key
 must not have been modified since the Nessie commit's `expectedHash`.
 
+The _Unmodified operation_ is not persisted.
```
dimas-b, line 187: Would we not need to validate `Unmodifed` on merges?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -135,8 +138,26 @@ public Hash commit(
...
+          if (!expectedStates.containsKey(contentId)) {
+            // First occurrence of contentsId in this commit.
+            commitAttempt.putExpectedStates(contentId, expectedValue);
+            expectedStates.put(contentId, expectedValue);
+          } else {
+            // Consecutive occurrence of contentsId in this commit.
+            if (expectedValue.isPresent()) {
+              // Operation expects a certain global, compare against the previous global-value
+              // from this commit.
+              if (!globals.get(contentId).equals(expectedValue.get())) {
```
dimas-b, line 152: According to `spec.md` I think we should deny this `Put` even if the global state matches, because the spec requires at most one `Put` per key... WDYT?

site/docs/develop/spec.md
snazy: Hm - that would allow a content-key to occur in multiple operations.

site/docs/develop/spec.md
snazy: It is validated, but not persisted.

site/docs/develop/spec.md
dimas-b: Right, e.g. for a table re-create: `Delete(key, oldId)` + `Put(key, newId)`

site/docs/develop/spec.md
dimas-b: How can we validate that `Unchanged` still holds when we merge (if it's not persisted)?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
dimas-b: sgtm

site/docs/develop/spec.md
snazy: Nope. That would be:
`Delete(oldKey)` +
`Put(newKey, newContentId)`

snazy: ^ that's for a "rename", sorry

A recreate with the same key would manifest as a single `Put` for the current `key`.

dimas-b: interesting...  understood :)

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
dimas-b: I confused keys with IDs, here, sorry... closing this conversation for now.

dimas-b (approved):

site/docs/develop/spec.md
```diff
@@ -139,8 +139,27 @@ recorded within the [_Put Operation_](#put-operation) of a Nessie commit.
...
+
+An `ALTER TABLE RENAME` is mapped to a _Delete operation_ using the Content Key for the table
+being renamed plus at least one _Put operation_ using the Content Key of the table's new name,
+using the Content Id of the table being renamed.
+
+A `DROP TABLE` is represented as a Nessie _Delete operation_ (without a _Put operation_ for the
+same Content Id).
+
+A `DROP TABLE` + `CREATE TABLE` using the same table name (Content Key) in a single commit are
+mapped to one _Put operation_ with a different Content Id.
```
nastra, line 162: maybe add a note why a put + different content id is sufficient in this case

PR #2329, opened 2021-10-11
Draft: Client compatibility updates

harshm-dev:
The commit updates the client to be compatible with the latest code. There's a cicrular dependency between iceberg and nessie primarily for integration testing and is yet to be addressed. While this change is still in a draft stage, it can be used for development needs as suited.

TODO: Resolve a way to handle circular dependency


nastra:
@harshm-dev we cannot upgrade the code now. We can only upgrade it to Nessie version **X** once that Nessie version actually made it into Iceberg/Delta and once there's a new release of Iceberg/Delta

nastra:
Closing based on the discussion above

PR #2330, opened 2021-10-11
Validation messages on missing context from the Nessie commands

harshm-dev:
Existing behaviour - 
```
spark-sql> list references;
21/10/05 09:19:22 ERROR SparkSQLDriver: Failed in [list references]
java.lang.IllegalArgumentException: Cannot construct Http client. Must have a non-null uri
```

A pre-check, ensuring catalog conf is present in the session, the command now outputs with the core problem, and the rectification mechanism -
```
spark-sql> list references;
21/10/11 19:51:22 ERROR SparkSQLDriver: Failed in [list references]
java.lang.IllegalArgumentException: requirement failed: Catalog is not defined. Either set the catalog in context before running the command [eg. `USE nessie`] or provide the catalog in the command [eg. `<COMMAND> IN nessie`]
```

Additionally, a check, that validates if a nessie server uri is present against a catalog configuration
```
spark-sql> list references in n1;
21/10/11 19:55:22 ERROR SparkSQLDriver: Failed in [list references in n1]
java.lang.IllegalArgumentException: requirement failed: Nessie catalog URI not defined. Please set a value for conf [spark.sql.catalog.n1.uri].
```


harshm-dev:
Try-catch will catch the exceptions at the time of client initialisation. It can be done, but less optimal as a trigger point if we want to inform the user about the missing catalog information. 
It'll be great I can get a better picture of when this check on catalog is irrelevant. Few questions -

1. Is it correct that the commands listed in [here](https://github.com/projectnessie/nessie/blob/main/clients/spark-antlr-grammar/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/NessieSqlExtensions.g4) work only if the catalog-impl is nessie type?
2. Is it possible to initialise spark-shell/spark-sql in a way, that the default catalog-impl is nessie? Can you please suggest a case when [currentCatalog](https://github.com/projectnessie/nessie/blob/bed7473646af72741e5d9964f893a4a1373533aa/clients/spark-extensions/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala#L170) is empty, but the scenario expects a nessie catalog interaction, and the commands should executed.

nastra:
> Try-catch will catch the exceptions at the time of client initialisation. It can be done, but less optimal as a trigger point if we want to inform the user about the missing catalog information. It'll be great I can get a better picture of when this check on catalog is irrelevant. Few questions -
> 
> 1. Is it correct that the commands listed in [here](https://github.com/projectnessie/nessie/blob/main/clients/spark-antlr-grammar/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/NessieSqlExtensions.g4) work only if the catalog-impl is nessie type?
> 2. Is it possible to initialise spark-shell/spark-sql in a way, that the default catalog-impl is nessie? Can you please suggest a case when [currentCatalog](https://github.com/projectnessie/nessie/blob/bed7473646af72741e5d9964f893a4a1373533aa/clients/spark-extensions/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala#L170) is empty, but the scenario expects a nessie catalog interaction, and the commands should executed.

re 1. I don't know. You need to probably check the code that hooks everything together
re 2. probably not

PR #2359, opened 2021-10-13
SQL extensions: SHOW LOG & SHOW REF need to point to the previously configured Hash

nastra:
ASSIGN and USE REFERENCE both provide a way to use a specific Hash. Once
a user then uses SHOW LOG / SHOW REF, then both those commands need to
use the specified Hash.


dimas-b:
Could we keep state in the NessieCatalog object itself? Is it fixed within a session?

PR #2369, opened 2021-10-13
Give rocks-db-config a default value so it can be omitted

snazy:


omarsmak:
Can you please update the documentation as well for the configs? https://github.com/projectnessie/nessie/blob/main/site/docs/try/configuration.md

PR #2414, opened 2021-10-18
docker-compose for store type and authn variants 

harshm-dev:
- Introduced docker-compose for in memory, mongo and oidc authentication
- Switched the default docker-compose to Mongo instead of DynamoDB
- Readme


omarsmak:
@harshm-dev please don't forget to add docs to the main website on how to use Nessie with docker and docker-compose

PR #2552, opened 2021-10-31
Fix Swagger API example schema for getContents()

harshm-dev:
The swagger api example is partially fixed with https://github.com/projectnessie/nessie/pull/2545. However, the example wasn't executing because the schema was set to that of ContentsKey; resulting in a - not a valid json error"

This change attempts to
a. Use String type as Key @schema to allow execution from swagger.
b. Reset example key as "example.key" as it aligns with other examples in there.


ajantha-bhat (approved):
LGTM.
If you can attach a screenshot of the latest UI run , it would be helpful. 
If we had done the same in previous PR, we would have caught it that time only as it is hard to catch from review.

harshm-dev:
contentskeyswagger

harshm-dev:
contentskeyswagger

PR #2570, opened 2021-11-02
Remove ContentsApi + move get-multiple-contents to TreeApi

snazy:
* Removes `ContentsApi` + related interfaces & classes
* Moves `getMultipleContents()` to `TreeApi`
* Includes Python changes

Related Nessie CLI update to list contents with the associated key for both textual and JSON
representation + indicate keys that were not found (instead of a bare Python exception).

Nit-fixes:
* use default-branch for Nessie CLI contents in all cases
* error out with appropriate message when the expected-hash is not present
* add missing `allow_none=True` in `model.py`


snazy:
Manually tested `nessie` CLI + web UI

python/pynessie/commands/contents.py
```diff
@@ -170,18 +182,13 @@ def _edit_contents(key: str, body: str) -> Optional[str]:
 
 
 def _get_contents(
-    nessie: NessieClient, ref: str, delete: bool = False, use_stdin: bool = False, expect_same_contents: bool = False, *keys: str
+    nessie: NessieClient, ref: str, keys: List[str], delete: bool = False, use_stdin: bool = False, expect_same_contents: bool = False
 ) -> List[Operation]:
     contents_altered: List[Operation] = list()
+    contents_dict = nessie.get_values(ref, keys)
```
omarsmak, line 188: We don't intend to check for `NessieNotFoundException` in case reference is not found here?

nastra (approved):
LGTM

python/pynessie/commands/contents.py
snazy: That's already handled. It prints something like
```
Requested URL: http://localhost:19120/api/v1/trees/contents?ref=foop
Server status: 404
Error code: REFERENCE_NOT_FOUND
Server message: Named reference 'foop' not found
Server traceback: None
```

omarsmak (approved):

PR #2578, opened 2021-11-03
Introduce pylint to pynessie in order to enforce more strict checking

omarsmak:
This PR intends to introduce `pylint` in order to enforce more strict static typing, this can work alongside `flake8` and `mypy`

### DIFF
--- a/python/.pylintrc
+++ b/python/.pylintrc
@@ -94,7 +94,8 @@ disable=raw-checker-failed,
         consider-using-f-string,
         no-value-for-parameter,
         consider-using-dict-items,
-        keyword-arg-before-vararg
+        keyword-arg-before-vararg,
+        unused-import
 
 # Enable the message, report, category or checker with the given id(s). You can
 # either give multiple identifier separated by comma (,) or put this option
--- a/python/.pylintrc
+++ b/python/.pylintrc
@@ -1,0 +1,574 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-allow-list=
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
+# for backward compatibility.)
+extension-pkg-whitelist=
+
+# Return non-zero exit code if any of these messages/categories are detected,
+# even if score is above --fail-under value. Syntax same as enable. Messages
+# specified are enabled, while categories only check already-enabled messages.
+fail-on=
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10.0
+
+# Files or directories to be skipped. They should be base names, not paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the ignore-list. The
+# regex matches against paths.
+ignore-paths=
+
+# Files or directories matching the regex patterns are skipped. The regex
+# matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Min Python version to use for version dependend checks. Will default to the
+# version used to run pylint.
+py-version=3.7
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED.
+confidence=
+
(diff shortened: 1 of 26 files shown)


python/.pylintrc
```diff
@@ -0,0 +1,574 @@
+[MASTER]
```
snazy, line 1: Question: where does this config file come from? (I'm in no way a Python guru - so bear with me ;) )

python/tox.ini
```diff
@@ -35,6 +35,15 @@ commands =
     safety check --ignore=41002
     mypy --ignore-missing-imports -p pynessie
 
+[testenv:pylint]
+basepython = python
+deps =
+    -r{toxinidir}/requirements_dev.txt
```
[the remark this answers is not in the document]
omarsmak: It should be enough, in fact only `requirements_lint.txt` is needed, I will change this

python/.pylintrc
omarsmak: no problems :). It is generated through `pylint` cli via this command `pylint --generate-rcfile > .pylintrc` and then from there, the config file can be modified as needed

snazy: So it's the default for now, right?

snazy:
I'm okay with this change, having stricter type checks and such things is a good thing IMO.
Maybe just mention it to @rymurr & @dimas-b whether they're okay with it as well.

python/.pylintrc
omarsmak: Nope, I have disabled some features like `consider-using-f-string`, `too-few-public-methods`  .. etc as I think they are not necessary needed for now and changed properties like `max-line-length`

python/tox.ini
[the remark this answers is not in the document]
omarsmak: fixed

snazy (approved):

python/pynessie/error.py
```diff
@@ -168,22 +168,22 @@ def _create_nessie_exception(error: dict, status: int, reason: str, url: str) ->
...
-    elif "REFERENCE_CONFLICT" == error_code:
+    if error_code == "REFERENCE_CONFLICT":
         return NessieReferenceConflictException(error, status, url, reason)
     return None
 
 
 def _create_exception(error: dict, status: int, reason: str, url: str) -> Exception:
     if 400 <= status < 500:
-        reason = u"Client Error (%s)" % reason
+        reason = f"Client Error {reason}"
```
dimas-b, line 184: TIL :) This looks a lot nicer :+1:

python/tox.ini
```diff
@@ -15,33 +15,37 @@
...
 [tox]
-envlist = py36, py37, py38, py39, flake8
+envlist = py36, py37, py38, py39, flake8, pylint
 
 [gh-actions]
 python =
     3.6: py36
     3.7: py37
-    3.8: py38, flake8
+    3.8: py38, flake8, pylint
```
dimas-b, line 24: just wondering: why do we run style tests on 3.8 and not 3.9? I'd guess going with the latest version for style-related stuff would be preferable from the general code evolution perspective?

python/tox.ini
```diff
@@ -15,33 +15,37 @@
...
     flake8 pynessie tests
     safety check --ignore=41002
     mypy --ignore-missing-imports -p pynessie
 
+[testenv:pylint]
+basepython = python
+deps =
+    -r{toxinidir}/requirements_lint.txt
+commands =
+    pylint pynessie
```
dimas-b, line 41: why not lint tests too (in the spirit of "test is code")?

dimas-b (commented):
LGTM overall :100: Just wondering if we should `pylint` the test code too. Plus a naming comment about `y_s` and related variables.

python/tox.ini
omarsmak: good question, from my side, I have no idea 😄.

python/tox.ini
omarsmak: good catch, will add that as well

python/pynessie/commands/branch.py
```diff
@@ -25,8 +25,8 @@
 
 
 @click.command(name="branch")
-@click.option("-l", "--list", cls=MutuallyExclusiveOption, is_flag=True, help="list branches", mutually_exclusive=["delete"])
-@click.option("-d", "--delete", cls=MutuallyExclusiveOption, is_flag=True, help="delete a branch", mutually_exclusive=["list"])
+@click.option("-l", "--list", "is_list", cls=MutuallyExclusiveOption, is_flag=True, help="list branches", mutually_exclusive=["delete"])
```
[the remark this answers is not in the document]
dimas-b: fair enough

python/pynessie/client/nessie_client.py
```diff
@@ -228,7 +228,7 @@ def get_log(
         """
         page_token = filtering_args.get("pageToken", None)
 
-        def fetch_logs(max: Optional[int], token: Optional[str] = page_token) -> LogResponse:
+        def fetch_logs(max_: Optional[int], token: Optional[str] = page_token) -> LogResponse:
```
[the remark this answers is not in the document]
omarsmak: Fixed

python/pynessie/conf/__init__.py
```diff
@@ -29,7 +29,7 @@
 
 def to_dict(config: confuse.Configuration) -> dict:
     """Convert confuse Configuration object to dict."""
-    ys = config.dump(redact=True)
-    fd = StringIO(ys)
-    dct = yaml.safe_load(fd)
+    y_s = config.dump(redact=True)
+    f_d = StringIO(y_s)
```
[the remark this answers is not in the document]
omarsmak: Changed it to something more meaningful

python/tox.ini
omarsmak: @dimas-b I have added pylint to the tests. However, I had to exchange the global variables with dataclass as pylint wasn't happy about using the global variables. Can you please check this out?

python/tests/conftest.py
```diff
@@ -98,15 +103,15 @@ def reset_nessie_server_state() -> None:
...
+    if nessie_test_config.cleanup:
         return None
     return request
 
 
 @pytest.fixture(scope="module")
 def vcr_config() -> dict:
     """VCR config that adds a custom before_record_request callback."""
-    pytest.nessie_cleanup = False
+    nessie_test_config.cleanup = False
```
dimas-b, line 114: nice catch :+1:

dimas-b (approved):

nastra (approved):
overall LGTM and I like the idea of doing more strict checking

PR #2603, opened 2021-11-05
Add pagination to `TreeApi.getAllReferences()`

### LINKED ISSUE #2595: Add pagination to all endpopints returning lists
Just to ensure that we're safe on the API/client side in case we have to implement pagination.

The goal of this issue is that all endpoints that return a list support pagingation.
It is not a goal to actually _implement_ pagination for endpoints - so existing endpoints will return a single page.

snazy:
* Includes changes to the web UI + pynessie.
* Splits the Java interfaces for the Nessie-API request builders
* Abstracts common code for `CommitLOgParams`/`EntriesParams`/`ReferencesParams`
* Adds support in `StreamingUtil`

Fixes #2595

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
+++ b/clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
@@ -26,12 +26,30 @@ import org.projectnessie.model.CommitMeta;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.EntriesResponse.Entry;
 import org.projectnessie.model.LogResponse;
+import org.projectnessie.model.Reference;
+import org.projectnessie.model.ReferencesResponse;
 
 /** Helper and utility methods around streaming of {@link NessieApiV1} et al. */
 public final class StreamingUtil {
 
   private StreamingUtil() {
     // intentionally blank
+  }
+
+  /**
+   * Default implementation to return a stream of references, functionally equivalent to calling
+   * {@link NessieApiV1#getAllReferences()} with manual paging.
+   *
+   * @return stream of {@link Reference} objects
+   */
+  public static Stream<Reference> getAllReferencesStream(
+      @NotNull NessieApiV1 api, OptionalInt maxRecords) throws NessieNotFoundException {
+
+    return new ResultStreamPaginator<>(
+            ReferencesResponse::getReferences,
+            (r, pageSize, token) ->
+                builderWithPaging(api.getAllReferences(), pageSize, token).get())
+        .generateStream(null, maxRecords);
   }
 
   /**
@@ -55,8 +73,10 @@ public final class StreamingUtil {
     return new ResultStreamPaginator<>(
             EntriesResponse::getEntries,
             (ref1, pageSize, token) ->
-                builderWithPagedQuery(
-                        api.getEntries(), ref1, hashOnRef, pageSize, token, queryExpression)
+                builderWithPaging(api.getEntries(), pageSize, token)
+                    .refName(ref1)
+                    .hashOnRef(hashOnRef)
+                    .queryExpression(queryExpression)
                     .get())
         .generateStream(ref, maxRecords);
   }
@@ -83,21 +103,17 @@ public final class StreamingUtil {
     return new ResultStreamPaginator<>(
             LogResponse::getOperations,
             (reference, pageSize, token) ->
-                builderWithPagedQuery(
-                        api.getCommitLog(), reference, hashOnRef, pageSize, token, queryExpression)
+                builderWithPaging(api.getCommitLog(), pageSize, token)
+                    .refName(reference)
+                    .hashOnRef(hashOnRef)
+                    .queryExpression(queryExpression)
                     .untilHash(untilHash)
                     .get())
         .generateStream(ref, maxRecords);
   }
 
-  private static <B extends PagingBuilder<B>> B builderWithPagedQuery(
-      B builder,
-      String reference,
-      String hashOnRef,
-      Integer pageSize,
-      String token,
-      String queryExpression) {
-    builder.refName(reference).hashOnRef(hashOnRef).queryExpression(queryExpression);
+  private static <B extends PagingBuilder<B>> B builderWithPaging(
+      B builder, Integer pageSize, String token) {
     if (pageSize != null) {
       builder.maxRecords(pageSize);
     }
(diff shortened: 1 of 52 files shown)


clients/client/src/main/java/org/projectnessie/client/api/QueryBuilder.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client.api;
+
+public interface QueryBuilder<R extends QueryBuilder<R>> {
+
+  /** Required: the <a href="https://github.com/projectnessie/cel-java">CEL</a> script. */
```
dimas-b, line 20: `Required`? Should it not be optional, meaning get "all" if no query expression is given?

model/src/main/java/org/projectnessie/api/params/AbstractParams.java
```diff
@@ -0,0 +1,67 @@
...
+package org.projectnessie.api.params;
+
+import javax.ws.rs.QueryParam;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+
+public abstract class AbstractParams {
+
+  @Parameter(description = "maximum number of entries to return, just a hint for the server")
+  @QueryParam("max")
+  private Integer maxRecords;
```
dimas-b, line 25: `maxRecordsPerPage` perhaps for the sake of clarity?

python/pynessie/client/nessie_client.py
```diff
@@ -75,13 +76,13 @@ def __init__(self: "NessieClient", config: confuse.Configuration) -> None:
         except confuse.exceptions.NotFoundError:
             self._base_branch = None
 
-    def list_references(self: "NessieClient") -> List[Reference]:
+    def list_references(self: "NessieClient") -> ReferencesResponse:
```
dimas-b, line 79: Since this method does not take pagination parameters, would it make sense to iterate all pages inside it and return the full list (as before)?

python/tests/cassettes/test_nessie_cli_auth/test_bearer_auth.yaml
```diff
@@ -1,4 +1,155 @@
 interactions:
```
dimas-b, line 1: Is re-recording "auth" cassettes necessary for this PR?

python/tests/cassettes/test_nessie_cli_auth/test_bearer_auth.yaml
```diff
@@ -1,4 +1,155 @@
...
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '0'
+      User-Agent:
+      - python-requests/2.26.0
+    method: DELETE
+    uri: http://localhost:19120/api/v1/trees/branch/main?expectedHash=2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d
```
dimas-b, line 66: Hmmm.... this looks like a clean-up request... I think it should normally be excluded from recordings :thinking:

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -55,8 +73,10 @@ private StreamingUtil() {
     return new ResultStreamPaginator<>(
             EntriesResponse::getEntries,
             (ref1, pageSize, token) ->
-                builderWithPagedQuery(
-                        api.getEntries(), ref1, hashOnRef, pageSize, token, queryExpression)
+                builderWithPaging(api.getEntries(), pageSize, token)
+                    .refName(ref1)
```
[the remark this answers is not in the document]
snazy: It's just using the existing code.

clients/client/src/main/java/org/projectnessie/client/api/QueryBuilder.java
snazy: This is just moving the existing code + comments.

python/tests/cassettes/test_nessie_cli_auth/test_bearer_auth.yaml
snazy: Yes

model/src/main/java/org/projectnessie/api/params/AbstractParams.java
snazy: Let's not change unrelated things in this PR.

python/pynessie/client/nessie_client.py
snazy: That's what it implicitly does, as the server doesn't support paging yet.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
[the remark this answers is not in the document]
dimas-b: Yes, but since we're touching pagination in this PR, would it be reasonable to make this part more direct?

clients/client/src/main/java/org/projectnessie/client/api/QueryBuilder.java
dimas-b: ok

model/src/main/java/org/projectnessie/api/params/AbstractParams.java
dimas-b: ok

python/pynessie/client/nessie_client.py
dimas-b: If we have iteration in the client lib, then client will have have to update when the server starts supporting pagination.

snazy: We do not have that for list-keys/tables either at the moment, probably because it's not supported on the server side. It's only implemented for the commit log.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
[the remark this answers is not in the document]
dimas-b: ok

python/pynessie/client/nessie_client.py
dimas-b: fair enough

dimas-b (approved):

python/tests/cassettes/test_nessie_cli_auth/test_bearer_auth.yaml
dimas-b: #2616

snazy: Nope, haven't changed the generated files.

PR #2626, opened 2021-11-09
Move table-metadata-pointer back to global state, use snapshot,schema,spec,sort-order ids as on-ref state

snazy:
This change basically reverts #2313. It moves the Iceberg table-metadata-pointer back to Nessie's global state and snapshot ID plus other relevant IDs (schema-ID, partition-spec-ID, sort-order-ID) to Nessie's on-reference state.

A working version of the Iceberg changes is in this branch: https://github.com/snazy/iceberg/tree/back-to-single-table-metadata

### DIFF
--- a/model/src/main/java/org/projectnessie/api/ContentApi.java
+++ b/model/src/main/java/org/projectnessie/api/ContentApi.java
@@ -39,7 +39,8 @@ public interface ContentApi {
    *
    * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Content}
    * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
-   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID).
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID,
+   * schema-ID, partition-spec-ID, default-sort-order-ID).
    *
    * @param key the {@link ContentKey}s to retrieve
    * @param ref named-reference to retrieve the content for
@@ -64,7 +65,8 @@ public interface ContentApi {
    *
    * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Content}
    * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
-   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID).
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID,
+   * schema-ID, partition-spec-ID, default-sort-order-ID).
    *
    * @param ref named-reference to retrieve the content for
    * @param hashOnRef hash on {@code ref} to retrieve the content for, translates to {@code HEAD},
--- a/model/src/main/java/org/projectnessie/api/ContentApi.java
+++ b/model/src/main/java/org/projectnessie/api/ContentApi.java
@@ -65,8 +65,8 @@ public interface ContentApi {
    *
    * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Content}
    * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
-   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID,
-   * schema-ID, partition-spec-ID, default-sort-order-ID).
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-id,
+   * schema-id, partition-spec-id, default-sort-order-id).
    *
    * @param ref named-reference to retrieve the content for
    * @param hashOnRef hash on {@code ref} to retrieve the content for, translates to {@code HEAD},
--- a/model/src/main/java/org/projectnessie/model/IcebergTable.java
+++ b/model/src/main/java/org/projectnessie/model/IcebergTable.java
@@ -28,14 +28,8 @@ import org.immutables.value.Value;
  * Represents the state of an Iceberg table in Nessie. An Iceberg table is globally identified via
  * its {@link Content#getId() unique ID}.
  *
- * <p>The Iceberg-table-state consists of the location to the table-metadata and the state of the ID
- * generators using the serialized version of Iceberg's {@code TableIdGenerators} object.
- *
- * <p>The table-metadata-location is managed on each Nessie reference, which means that all versions
- * of the same table have distinct table-metadata across all named-references (branches and tags).
- *
- * <p>The information needed to generate IDs for an Iceberg table that need to be globally unique,
- * for example the last-column-ID, is managed globally within Nessie.
+ * <p>The Iceberg-table-state consists of the location to the table-metadata and the state of
+ * relevant IDs using a serialized version of those.
  *
  * <p>When adding a new table (aka content-object identified by a content-id), use a {@link
  * org.projectnessie.model.Operation.Put} without an expected-value. In all other cases (updating an
@@ -69,20 +63,38 @@ public abstract class IcebergTable extends Content {
   @NotBlank
   public abstract String getMetadataLocation();
 
-  /** Opaque representation of Iceberg's {@code TableIdGenerators}. */
-  public abstract String getIdGenerators();
+  public abstract long getSnapshotId();
 
-  public static IcebergTable of(String metadataLocation, String idGenerators) {
+  public abstract int getSchemaId();
+
+  public abstract int getSpecId();
+
+  public abstract int getSortOrderId();
+
+  public static IcebergTable of(
+      String metadataLocation, long snapshotId, int schemaId, int specId, int sortOrderId) {
     return ImmutableIcebergTable.builder()
         .metadataLocation(metadataLocation)
-        .idGenerators(idGenerators)
+        .snapshotId(snapshotId)
+        .schemaId(schemaId)
+        .specId(specId)
+        .sortOrderId(sortOrderId)
         .build();
   }
 
-  public static IcebergTable of(String metadataLocation, String idGenerators, String contentId) {
+  public static IcebergTable of(
+      String metadataLocation,
+      long snapshotId,
+      int schemaId,
+      int specId,
+      int sortOrderId,
+      String contentId) {
     return ImmutableIcebergTable.builder()
         .metadataLocation(metadataLocation)
-        .idGenerators(idGenerators)
+        .snapshotId(snapshotId)
+        .schemaId(schemaId)
+        .specId(specId)
+        .sortOrderId(sortOrderId)
         .id(contentId)
         .build();
   }
--- a/python/pynessie/model.py
+++ b/python/pynessie/model.py
@@ -47,7 +47,10 @@ class IcebergTable(Content):
     """Dataclass for Nessie Content."""
 
     metadata_location: str = desert.ib(fields.Str(data_key="metadataLocation"))
-    id_generators: str = desert.ib(fields.Str(data_key="idGenerators"))
+    snapshot_id: int = desert.ib(fields.Int(data_key="snapshotId"))
+    schema_id: int = desert.ib(fields.Int(data_key="schemaId"))
+    spec_id: int = desert.ib(fields.Int(data_key="specId"))
+    sort_order_id: int = desert.ib(fields.Int(data_key="sortOrderId"))
 
     @staticmethod
     def requires_expected_state() -> bool:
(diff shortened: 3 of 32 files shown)


servers/store/src/test/java/org/projectnessie/server/store/TestStoreWorker.java
```diff
@@ -69,18 +69,18 @@ void testSerde(Map.Entry<ByteString, Content> entry) {
...
   void testSerdeIceberg() {
     String path = "foo/bar";
-    IcebergTable table = IcebergTable.of(path, "xyz", ID);
+    IcebergTable table = IcebergTable.of(path, "x", ID);
 
     ObjectTypes.Content protoTableGlobal =
         ObjectTypes.Content.newBuilder()
             .setId(ID)
-            .setIcebergGlobal(IcebergGlobal.newBuilder().setIdGenerators("xyz"))
+            .setIcebergMetadataPointer(
```
ajantha-bhat, line 77: I think if we keep this design. All the snapshot id will be present in current metadata pointer of table@ref.
So, this can cause the below problems.
a) Each branch, each table will lose the snapshot history. so time travel within the branch is not possible
b) Each branch, each table will not know the list of all the reachable files as snapshot history is lost. So, if the table is dropped, we cannot know which files to clean for that table@ref.
c) we lose table evolution history per branch, we cannot know how schema, sort order, partition spec is evolved for that table@ref.
d) table properties configured for table@ref will be used for the same table in other reference also. which is not good.

This is high level things I can recall. Yet to analyze how compaction per branch or other Iceberg operations are affected by this.

dimas-b (commented):
Code change LGTM, but I'm not sure I understand all the implications for Iceberg to approve :thinking: 

As a side note, if we keep Iceberg data in opaque form, would it not be easier to just have two properties in the `IcebergTable` class: `globalState` and `branchedState` and delegate interpretation to the Iceberg-side Nessie code?

servers/store/src/test/java/org/projectnessie/server/store/TestStoreWorker.java
snazy: > a) In each branch, each table will lose the snapshot history. So, time travel within the branch is not possible.

Time travel would still be possible via Nessie.

> b) In each branch, each table will not know the list of all the reachable files as snapshot history is lost.

Can you elaborate why `TableMetadata.snapshotLog()` is required for this?

> So, if the table is dropped in that branch, we cannot know which files to clean for that table@ref.

You can still figure that out by traversing the commit-log.

The only situation I can currently imagine is that you create a table on a new branch, add data and then just drop the branch. But every approach would run into this situation.

> c) we lose table evolution history per branch. We cannot know how schema, sort order, partition spec is evolved for that table@ref.

It's still there via the Nessie commits + the information in the referenced snapshots, right?

> d) Table properties configured for table@ref will be visible and used for the same table in other reference also. which can cause unexpected behaviour in other references for the table.

I don't think that there's an issue with exposing other references' information. Proper isolation would require way more than just thinking about it here.

What would those unexpected behaviors be?

> e) we lose metadata-log history per table in a branch. metadata-log is used for deleting old metadata files (and fixed number of maximum metadata files at given point of time). we have an auto action for cleaning the metadata files.

Why would we lose that metadata-log (`TableMetadata.previousFiles()`)?

There will be exactly one current and valid metadata (file).

ajantha-bhat: > b) In each branch, each table will not know the list of all the reachable files as snapshot history is lost.
> 
> Can you elaborate why TableMetadata.snapshotLog() is required for this?
I meant Table.snapshots(), which will have all branch's snapshot. So, when I do drop table on branch. At Iceberg side, I cannot know what all snapshots to delete for that branch.

ajantha-bhat: > d) Table properties configured for table@ref will be visible and used for the same table in other reference also. which can cause unexpected behaviour in other references for the table.

> I don't think that there's an issue with exposing other references' information. Proper isolation would require way more than just thinking about it here.
> 
> What would those unexpected behaviors be?

For example, If table in branch_1 is configured to use `write.target-file-size-bytes` = 10 MB, same table in all the branches will use that for writing file size. which may be unexpected for same table in other branch as they might be created for other experiments.

ajantha-bhat: > e) we lose metadata-log history per table in a branch. metadata-log is used for deleting old metadata files (and fixed number of maximum metadata files at given point of time). we have an auto action for cleaning the metadata files.
> 
> Why would we lose that metadata-log (TableMetadata.previousFiles())?

> There will be exactly one current and valid metadata (file).

I meant we lose per branch, not for overall table. Table in one ref now contains `TableMetadata.previousFiles()` which has metadata files from all the branches. So, when we add new commit in one branch, it might clean up other branch's metadata file when this configuration is enabled.

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -69,20 +63,19 @@
   @NotBlank
   public abstract String getMetadataLocation();
 
-  /** Opaque representation of Iceberg's {@code TableIdGenerators}. */
-  public abstract String getIdGenerators();
+  public abstract String getTableState();
```
rymurr, line 66: I thnk we should put actual properties for each thing here rather than an opaque `String` (same comment as @dimas-b  I think).

Off the top of my head I think we should save as actual properties:
* partition spec id
* schema id
* sequence number
but I may be missing some from V2 spec.

servers/store/src/test/java/org/projectnessie/server/store/TestStoreWorker.java
rymurr: > Now assume branch_1 has 1 commit and branch_2 has 99 commits. If metadata files to retain is configured as 100. For a new commit happens on branch_2, we delete branch_1's metadata file even though it is referenced in Nessie branch_1.

You are correct ajantha. We should be setting this property at table create time and running checks to ensure it isn't changed

> d) Table properties configured for table@ref will be visible and used for the same table in other reference also. which can cause unexpected behaviour in other references for the table

My thinking here is that for the time being table proeprties are global and when we come across a use case which requires table properties per branch we can adjust.

rymurr: on the topic of time travel and history generally: we are in a weird position currently as there are two timelines (iceberg via snapshots and Nessie via commits) and both are accessible via different APIs. We should strive to have 1 history API for a Nessie table ( which uses Nessie's history). So I am not overly worried about the iceberg timeline and iceberg apis to access it, the iceberg timeline may be goofy but the Nessie timeline is the source of truth. We do have to focus on extending Iceberg to ensure that certain API calls are routed to Nessie, these should be generic extension points in the iceberg metadata handling code and should make sense for all extenders of Iceberg (not just Nessie). Make sense?

model/src/main/java/org/projectnessie/model/IcebergTable.java
ajantha-bhat: @rymurr: sequence number is a global field. So no need to keep in the on_reference_state.
we just have to keep fields that are specific to current branch.

So, I think current chosen parameters are ok  (If table property need to maintain for per branch, we can add that as well)

And agree on the part that we can have a class called `TableCurrentState`, which keeps these parameters instead of opaque string.

rymurr: > we just have to keep fields that are specific to current branch.

ok cool. agreed

>  sequence number is a global field.

I would just like to amend this statement a bit for my own clarity. We have to save anything that can be changed w/o creating a snapshot. The real risk we face is there are iceberg transactions that don't create a snapshot. So global counters (column id, sequence number) are safe to ignore. Things like schema and partition id are not.

rymurr: Additionally I think we should create a test suite (probably along w/ the iceberg fix) to test all these assumtions

servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java
```diff
@@ -111,15 +115,18 @@ public Content valueFromStore(ByteString onReferenceValue, Optional<ByteString>
...
         }
         return IcebergTable.of(
-            content.getIcebergMetadataPointer().getMetadataLocation(),
-            global.getIcebergGlobal().getIdGenerators(),
+            global.getIcebergMetadataPointer().getMetadataLocation(),
+            content.getIcebergRefState().getSnapshotId(),
+            content.getIcebergRefState().getSchemaId(),
+            content.getIcebergRefState().getSpecId(),
+            content.getIcebergRefState().getSortOrderId(),
             content.getId());
```
dimas-b, line 130: nit: might using a builder be preferable here, to reduce the chance of mismatch in ID args, which have the same java type?

servers/store/src/main/proto/table.proto
```diff
@@ -34,8 +34,11 @@ message IcebergMetadataPointer {
   string metadata_location = 1;
 }
 
-message IcebergGlobal {
-  string id_generators = 1;
+message IcebergRefState {
```
ajantha-bhat, line 37: In the contentAPI descriptions and some other place of code, It still mentions that on-ref-state is just the snapshot id. can we change that as well?
please look up `Iceberg: snapshot-ID` in the files.

snazy: addressed

servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java
snazy: addressed

dimas-b (approved):
LGTM

model/src/main/java/org/projectnessie/api/ContentApi.java
```diff
@@ -39,7 +39,8 @@
    *
    * <p>If the table-metadata is tracked globally (Iceberg), Nessie returns a {@link Content}
    * object, that contains the most up-to-date part for the globally tracked part (Iceberg:
-   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID).
+   * table-metadata) plus the per-Nessie-reference/hash specific part (Iceberg: snapshot-ID,
+   * schema-ID, partition-spec-ID, default-sort-order-ID).
```
ajantha-bhat, line 43: nit: In the descriptions, If we are naming this variable(`default-sort-order-ID`) based on iceberg spec, then other two variable has to be `default-spec-id`, `current-schema-id`

or may be instead of mentioning each field, we just link it to `IcebergRefState`

ajantha-bhat (commented):
one small nit comment. 

overall LGTM from my side.

model/src/main/java/org/projectnessie/api/ContentApi.java
rymurr: nit: i thikn `ID` shjould be `id`

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -69,20 +63,38 @@
   @NotBlank
   public abstract String getMetadataLocation();
 
-  /** Opaque representation of Iceberg's {@code TableIdGenerators}. */
-  public abstract String getIdGenerators();
+  public abstract long getSnapshotId();
```
rymurr, line 66: do these need default values? particularly the shcema/specid etc? Or do they have to be specified every time

python/pynessie/model.py
```diff
@@ -55,7 +58,11 @@ def requires_expected_state() -> bool:
 
     def pretty_print(self: "IcebergTable") -> str:
         """Print out for cli."""
-        return "Iceberg table:\n\tmetadata-location:{}\n\tid-generators:{}".format(self.metadata_location, self.id_generators)
+        return (
+            f"Iceberg table:\n\tmetadata-location:{self.metadata_location}\n\tsnapshot-id:{self.snapshot_id}"
+            f"\n\tschema-id:{self.schema_id}"
+            f"\n\tpartition-spec-id:{self.spec_id}\n\tdefault-sort-order-id:{self.sort_order_id}"
```
rymurr, line 64: nit spaces after `:`

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: Don't think those should have default values. It uses what Iceberg's `TableMetadata` says and not rely on a default value.

ajantha-bhat (approved):

Issue #2614, opened 2021-11-08
[CLI] Generate CLI documentation for nested sub-commands


omarsmak:
Currently, `pynessie` only generate the documentation for top level command but it doesn't for sub-commands like `nessie content [commit|view|list]`. Ideally it should generate all documentation for all sub-commands

PR #2628, opened 2021-11-09
Generate python docs for nested commands

omarsmak:
Fixes: https://github.com/projectnessie/nessie/issues/2614

As well, refactoring and moving the docs generation script from the tests into its own script under `python/tools/generate_docs.py` that can be triggered through `tox` or `tox -e docs`

### DIFF
--- a/python/CONTRIBUTING.rst
+++ b/python/CONTRIBUTING.rst
@@ -77,10 +77,8 @@ Ready to contribute? Here's how to set up `nessie` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+   tests, including testing other Python versions with tox and all docs have been generated::
 
-    $ flake8 pynessie tests
-    $ python setup.py test or pytest
     $ tox
 
    To get flake8 and tox, just pip install them into your virtualenv.
--- a/python/docs/branch.rst
+++ b/python/docs/branch.rst
@@ -1,48 +1,48 @@
 .. code-block:: bash
 
-	Usage: cli branch [OPTIONS] [BRANCH] [BASE_REF]
-	
-	  Branch operations.
-	
-	  BRANCH name of branch to list or create/assign
-	
-	  BASE_REF name of branch or tag from which to create/assign the new BRANCH
-	
-	  Examples:
-	
-	      nessie branch -> list all branches
-	
-	      nessie branch -l -> list all branches
-	
-	      nessie branch -l main -> list only main
-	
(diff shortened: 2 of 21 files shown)


python/CONTRIBUTING.rst
```diff
@@ -77,10 +77,8 @@ Ready to contribute? Here's how to set up `nessie` for local development.
    Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass flake8 and the
-   tests, including testing other Python versions with tox::
+   tests, including testing other Python versions with tox and all docs have been generated::
```
dimas-b, line 80: nit: double `:` at the end of line?

python/CONTRIBUTING.rst
```diff
@@ -77,10 +77,8 @@ Ready to contribute? Here's how to set up `nessie` for local development.
...
 
-    $ flake8 pynessie tests
-    $ python setup.py test or pytest
     $ tox
 
    To get flake8 and tox, just pip install them into your virtualenv.
```
dimas-b, line 84: Do we need to mention `flake8` now? I guess `tox` will take care of it automatically?

dimas-b (approved):

python/CONTRIBUTING.rst
omarsmak: I think `rst` needs to have double `:` at the end of the line 🤔

dimas-b: TIL :facepalm:

python/CONTRIBUTING.rst
omarsmak: I think it could be the case you want to just run `flake8` alone though without tox 🤔

dimas-b: In my experience, the standalone `flake8` results do not always match what it detects under `tox` :shrug:

omarsmak: Weird, I have been always using the standalone `flake8`, maybe because in `tox` in needs to run both for `pynessie` and `tests` folders? 🤔

dimas-b: IDK, I hit a couple of cases when the `tox` run of `flake8` was stricter than a direct run :shrug:

omarsmak: > IDK, I hit a couple of cases when the `tox` run of `flake8` was stricter than a direct run 🤷

Ah maybe because on `tox -e flake8`, it runs as well `mypy` which Python`s type checker as you can see [here](https://github.com/projectnessie/nessie/blob/main/python/tox.ini#L34). I think should be separated into different tox step for clarity 🤔

Issue #2598, opened 2021-11-05
[CLI] Handle content key in form `a.b."c.d"` in `nessie content` commands group


omarsmak:
Related comment: https://github.com/projectnessie/nessie/pull/2593#discussion_r743716930

PR #2636, opened 2021-11-10
Handle ContentKey better in nessie CLI

omarsmak:
Fixes: https://github.com/projectnessie/nessie/issues/2598

As well, make `ContentKey` the source of truth instead of passing `str` in the CLI and make `ContentKey` _somewhat_ similar to the Java API


python/docs/content_commit.rst
```diff
@@ -5,7 +5,8 @@
      Commit content.
    
          KEY is the content key that is associated with a specific content to
-         commit. This accepts as well multiple keys with space in between.
+         commit. This accepts as well multiple keys with space in between. The key
+         can be in this format: 'table.key' or 'table."namespace.key"'.
```
ajantha-bhat, line 9: should it be `namespace.table.key` ?

rymurr (approved):
I am happy with this change once @ajantha-bhat comments are addressed. Nice one @omarsmak

python/pynessie/client/nessie_client.py
```diff
@@ -38,7 +38,7 @@
 from ._endpoints import list_tables
 from ._endpoints import merge
 from ..auth import setup_auth
-from ..model import Branch
+from ..model import Branch, ContentKey
```
rymurr, line 41: nit: inconsistent import style. why isnt black failing this?

omarsmak: True. It seems almost in many places, the import style is mixed up, I am wondering why non of the linter that we have didn't complain about that 🤔. I will look at the linter configs separately

omarsmak: Fixed that one

python/docs/content_commit.rst
omarsmak: Fixed

ajantha-bhat (approved):

snazy (approved):
LGTM!

I'm not sure what would really happen over the whole lifetime of an empty `ContentKey`.

Probably something to tackle, if needed, in a different issue/PR, as `org.projectnessie.model.ContentKey#fromPathString` doesn't assert on that as well using e.g. `@NotBlank` on the parameter.

python/pynessie/model.py
```diff
@@ -123,6 +124,38 @@ class ContentKey:
...
+    def to_path_string(self: "ContentKey") -> str:
+        """Convert this key to a url encoded path string."""
+        # false positives in pylint
+        # pylint: disable=E1133
+        return ".".join(i.replace(".", "\00") if i else "" for i in self.elements)
+
+    @staticmethod
+    def from_path_string(key: str) -> "ContentKey":
+        """Convert from path encoded string to normal string."""
+        return ContentKey([i for i in ContentKey._split_key_based_on_regex(key) if i])
```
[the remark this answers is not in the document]
omarsmak: In that Python, it will just create an `ContentKey` with empty elements which I think shouldn't happen as the Java API requires non-empty elements, I think the Python should behave similar to its Java counterpart. 
And true, this should be addressed in both Java API and Python API to prevent empty `key` from creating empty elements.
I have created this issue to address this in both APIs https://github.com/projectnessie/nessie/issues/2648

Issue #2616, opened 2021-11-08
test_nessie_cli_auth.py does not exclude clean-up requests from VCR recordings


dimas-b:
`test_nessie_cli_auth.py` has a custom `vcr_config`, which appears to interfere with `conftest.vcr_config` and causes auxiliary HTTP requests to be recorded by the VCR plugin (they should normally be excluded).

Related comment: https://github.com/projectnessie/nessie/pull/2603#discussion_r744941151

PR #2650, opened 2021-11-11
Prevent unwanted tests from VCR recording if rewrite mode is set in the command line

omarsmak:
Fixes https://github.com/projectnessie/nessie/issues/2616


python/tests/test_nessie_cli_auth.py
```diff
@@ -26,13 +26,13 @@ def vcr_config() -> dict:
     return {"match_on": ["method", "scheme", "host", "port", "path", "query", "headers"]}
 
 
-@pytest.mark.vcr
+@pytest.mark.vcr(record_mode="none")
```
omarsmak, line 29: @dimas-b does it make sense to disable the VCR recording for these tests since they need auth config in the Nessie server?

python/tests/conftest.py
```diff
@@ -148,8 +148,17 @@ def _clean_nessie_session_marker(request: "FixtureRequest", record_mode: str) ->
...
-    if request.node.get_closest_marker("vcr"):
+    vcr_marker = request.node.get_closest_marker("vcr")
+
+    # Clean the server state for tests that have the @pytest.mark.vcr annotation
+    if vcr_marker:
+        if "record_mode" in vcr_marker.kwargs:
+            # if we have record_mode set in @pytest.mark.vcr annotation, we test that first
+            test_record_mode = vcr_marker.kwargs["record_mode"]
+
+            if test_record_mode is None or test_record_mode == "none":
```
dimas-b, line 159: I'd treat `None` as an error in this case. Setting `record_mode` without a value does not look intuitive for _disabling_ the recording.

python/tests/test_nessie_cli_auth.py
dimas-b: This is a bit tricky. I think we do need to record HTTP traffic for these tests sometimes. Specifically when auth stuff changes.

However, excluding these test from recording by default probably makes sense.

Can we have a cmd. line option to force enable recording for these special tests too? Maybe rename the annotation property to `default_recording_mode`?

dimas-b (commented):
I suspect there may be some interference between our custom `vcr_config` methods, but I did not get to gebugging the root case - pls see #2616

python/tests/test_nessie_cli_auth.py
omarsmak: > However, excluding these test from recording by default probably makes sense.
> 
> Can we have a cmd. line option to force enable recording for these special tests too? Maybe rename the annotation property to `default_recording_mode`?

I think `record_mode` is used by `pytest-recording` library which handles these markers and handles how the VCR recordings should run 🤔 .

dimas-b: I believe we can adjust the default recording logic in our callbacks (like `before_record_cb`). Yet, my main concern is that we are going to need to enable auth test recording at some point.

omarsmak: We can try to explore how we can enable that through command line. However, I don't really think enabling this for two tests manually needed when is needed is a big deal, isn't? 🤔 
How about this, at least for now, we keep these tests manually disabled. Meanwhile, we research on how we can do that through command line, WDTY @dimas-b ?

python/tests/conftest.py
omarsmak: Fixed

python/tests/test_nessie_cli_auth.py
dimas-b: I'm not sure a manual action to enable recording for these tests is more useful than a manual action of not (re-)running these tests when recording is on :thinking: 

Basically, one can record VCR cassettes per test file, e.g. recording just `test_nessie_cli.py`. When auth tests are re-recorded I'd prefer the recording to fail out of the box as a reminding for the user to set up authentication in the Nessie server.

Looking deeper into the auth test, I'm not sure `test_bearer_auth` in `test_nessie_cli_auth.py` actually has that much value. Essentially, it validates that we add auth headers to HTTP requests, but that is already validated by `test_nessie_auth.py` without VCR. At the same time, proper bearer token handling in the server is validated by java tests over REST API.

Basically, `test_bearer_auth` is just a recorded fact that someone manually tested python CLI with auth enabled. However, the overhead of maintaining it seems to be too high ATM. Perhaps we can drop `test_bearer_auth`. Then, the `vcr_config` in its test file should no longer be necessary. I believe `test_bearer_auth_invalid` should work without any custom VCR config or Nessie Server setup. 

WDYT?
cc: @rymurr

omarsmak: It seems `test_bearer_auth_invalid` needs some special config in Nessie? 
```
$ python -m pynessie.cli --json --auth-token invalid remote show
Remote URL: http://localhost:19120/api/v1
Default branch: main
Remote branches:
	main
```

dimas-b: ah, yes, authentication is not enforced by default :thinking: ...

I believe I added this mostly to validate the conversion of server-side auth errors into proper client-side error messages. That should now be covered by `test_nessie_error.py`, I guess.

So I would not mind removing this test too so as to simplify the overall VCR recording workflow.

Do you see a use case that this test validates that is not covered by `test_nessie_error.py`? We have some general purpose error handling test cases in `test_nessie_cli_error.py`, which should cover the passing of error information from HTTP level to our custom python code, and that should cover auth errors too, I think.

omarsmak: Yeah I also think is not necessary, I will just remove it in order to keep our tests simpler.

omarsmak: It is removed: https://github.com/projectnessie/nessie/pull/2650/commits/9f601e8868f521ce75823a5540b8c198672307e9

dimas-b: Did you remove their cassettes too?

omarsmak: Good catch! Removed it as well

dimas-b (approved):
Thanks for bearing with my throughout this review. I think in the end it was good to remove the redundant auth tests from the VCR suite :)

omarsmak:
@dimas-b can you please approve it again, I had to rebase 😅

dimas-b (approved):

PR #2651, opened 2021-11-11
Mark Python VCR cassettes as binary

snazy:
Marking the Python VCR cassettes as binary shall improve readability
of pull requests, as those changes would only show up a a modified
file, but not the whole textual diff, which is massive and usually
useless.

### DIFF
--- a/python/tests/cassettes/.gitattributes
+++ b/python/tests/cassettes/.gitattributes
@@ -1,0 +1,1 @@
+*.yaml binary linguist-generated


dimas-b:
Will it affect how `git` cmd. line shows the diff?.. or IntelliJ? I often review VCR diffs as a sanity check for what is going on :)

dimas-b:
Another option, I guess, is imposing order to JSON serialization of Nessie data types. That should make VCR diffs much smaller and more reflective of actual code changes, I hope.

snazy:
> Will it affect how `git` cmd. line shows the diff?.. or IntelliJ? I often review VCR diffs as a sanity check for what is going on :)

By default, yes. It will show something like "binary file changed".
But you can still tell `git` to do a textual diff (with the `--text` option).

dimas-b:
I'm on the fence regarding this change. If other people want to go ahead, I'd be ok too. However, I think it might be worth exploring the option for stable JSON property order too - I think it will help CLI users too.

snazy:
Added the [`linguist-vendored` attribute]()https://github.com/github/linguist/blob/master/docs/overrides.md, which is recognized by github. With this, github does not show the diff by default, but you can still click on the "Show diff".

IntelliJ uses its own content detection (so it shows the textual diff)

`git` shows those files as binary, aka hidden by default, but overridable w/ `--text`.

dimas-b (approved):

omarsmak (approved):

PR #2689, opened 2021-11-18
Bump `immer` to `9.0.6`

dimas-b:
This is mostly to bump the `immer` transitive dependency
to v9.0.6 to resolve CVE-2021-23436 and CVE-2021-3757.

----

Related: https://github.com/projectnessie/nessie/security/dependabot/ui/package-lock.json/immer/open


nastra (approved):

ajantha-bhat:
@dimas-b : But dependabot says even 11.0.4 is mapped to V8.0.1 of immer,  but CVE fixed in 9.0.6 of immer.
Is dependabot wrong ? 🤔
Screenshot 2021-11-18 at 11 06 44 PM

ui/package.json
```diff
@@ -41,7 +41,7 @@
     "react": "^16.14.0",
     "react-app-polyfill": "^2.0.0",
     "react-bootstrap": "^2.0.2",
-    "react-dev-utils": "^11.0.3",
+    "react-dev-utils": "^11.0.4",
```
ajantha-bhat, line 44: https://github.com/facebook/create-react-app/blob/v4.0.3/packages/react-dev-utils/package.json

ajantha-bhat: Issue is that `react-dev-utils` master branch code is pointing to  immer:9.0.6

but latest release is still pointing to "immer": "8.0.1".

So, I don't think this fix will solve the problem. we need to wait for new release.

ajantha-bhat (changes requested):

ui/package.json
ajantha-bhat: dependabot also saying the same thing.

image

dimas-b: you're right... I check `react-dev-utils` sources, but missed that the `9.0.6` upgrade was not released yet :facepalm:

ajantha-bhat: cool. Let's wait for the new release. which dependabot will auto update when happens. we can sit back and relax !

dimas-b: however https://github.com/facebook/create-react-app/blob/main/packages/react-dev-utils/package.json shows `immer` at `9.0.6` in `react-dev-utils` `11.0.4` :thinking:

dimas-b: Should we try v. `12.0.0-next.47` for `react-dev-utils`?

ajantha-bhat: True. Confusing. I checked from last release tag which is mapping to 11.0.3. 
not sure how 11.0.4 is out if version is still in master branch !

ajantha-bhat: Are the independently releasing packages without sync ?

dimas-b: `11.0.4` shows available here: https://www.npmjs.com/package/react-dev-utils/v/11.0.4

dimas-b: but a local build with 11.0.4 still pulls `immer` 8.0.1 :(

dimas-b: however, `immer` is not pinned, it seems, let me try to bump it individually :hourglass:

ajantha-bhat: > Should we try v. 12.0.0-next.47 for react-dev-utils?
I think not good to use unreleased version (snapshots)

may be we can raise issue in their github and get it clarified or get info about next release.

dimas-b:
Note: `mvn quarkus:dev` does not start the UI after the most recent quarkus version bump. You can validate UI by building and running a Nessie docker image.

ui/package.json
dimas-b: Added a direct dependency on `immer` to force bump it to `9.0.6` (force pushed)

ajantha-bhat: but will it break `react-dev-utils` ? we can test once, if it is fine. we can merge it.

dimas-b: I smoke-tested the UI in a local build - it was fine, but I'm fairly new to the whole UI tools chain, so I'd appreciate if someone could double check.

ui/package-lock.json
```diff
@@ -38210,6 +38220,11 @@
             "slash": "^3.0.0"
           }
         },
+        "immer": {
+          "version": "8.0.1",
```
ajantha-bhat, line 38224: 8.0.1 has CVE, so as long as we are keeping it in the repo, we are vulnerable right ?

ui/package.json
dimas-b: > but will it break react-dev-utils ?

`create-react-app` apparently uses this version combination on `main`.

https://github.com/facebook/create-react-app/blob/main/packages/react-dev-utils/package.json

ui/package-lock.json
dimas-b: TBH, I do not know how it works. This is a nested dep. from `react-dev-utils`. The top-level entry for `immer` is at `9.0.6` now.

dimas-b: Is there a way to verify what is actually packaged?

ajantha-bhat: I too have no idea for this side of the code. Theoretically the current changes looks ok to me. But better to consult UI experts :)

ajantha-bhat (commented):
Theoretically the current changes looks ok to me compared to last one. But better to consult UI experts :)

ajantha-bhat (approved):

ui/package.json
dimas-b: It turned out to be complicated... I did another force push to use `npm-force-resolutions` for bumping nested `immer` dependencies to `9.0.6`.

ui/package-lock.json
dimas-b: It turned out to be complicated... I did another force push to use `npm-force-resolutions` for bumping nested `immer` dependencies to `9.0.6`.

ajantha-bhat: It shows 34KLOC diff, do we really need to handle this way ? I think we can consult `react-dev-utils` and ask them for a formal release ?

dimas-b:
Let's wait for https://github.com/facebook/create-react-app/issues/11660 ... closing

PR #2707, opened 2021-11-22
Fix openapi example for getAllReferences()

### LINKED ISSUE #2596: Fix Swagger examples for /trees
Currently the example does not show the type of tree (branch or tag)

nastra:
fixes #2596

### DIFF
--- a/model/src/main/resources/META-INF/openapi.yaml
+++ b/model/src/main/resources/META-INF/openapi.yaml
@@ -174,3 +174,18 @@ components:
               additionalProp2: "yyy"
               additionalProp3: "zzz"
             signedOffBy: "signedOffByName <signedOffBy@example.com>"
+
+    referencesResponse:
+      value:
+        token: null
+        hasMore: false
+        references:
+          - type: BRANCH
+            hash: "abcdef4242424242424242424242beef00dead42112233445566778899001122"
+            name: "exampleBranch"
+          - type: TAG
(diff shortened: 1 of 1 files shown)


snazy (approved):

model/src/main/resources/META-INF/openapi.yaml
```diff
@@ -174,3 +174,18 @@ components:
               additionalProp2: "yyy"
               additionalProp3: "zzz"
             signedOffBy: "signedOffByName <signedOffBy@example.com>"
+
+    referencesResponse:
+      value:
+        token: "xxx"
+        hasMore: false
```
dimas-b, line 181: nit: if `hasMore` is `false`, shouldn't the `token` be `null` or not present?

dimas-b: Can we have two examples: 1) for a paginated response 2) for a non-paginated (or last page) response?

dimas-b (approved):

model/src/main/resources/META-INF/openapi.yaml
nastra: I set the token to `false`, since pagination is currently not supported due to https://github.com/projectnessie/nessie/blob/0e0b8713efca8050b3eefc377c0ee910f770b08d/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java#L97

dimas-b (approved):

PR #2715, opened 2021-11-23
Clarify paging in in-line code comments

snazy:


nastra (dismissed):

omarsmak (approved):

clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
```diff
@@ -40,7 +40,7 @@
      *
      * @param ref reference
      * @param pageSize page size
-     * @param token paging token
+     * @param token paging continuation token
```
dimas-b, line 43: nit: in `AbstractParams` it is called `pageToken`... WDYT about renaming this param (and the method in `PaginatedResponse`) to use the same `pageToken` base name?.. just for the sake of clarity and consistency.

For me personally the bare word `token` sound more like something related authentication than to paging.

snazy: yup - opened #2719

PR #2726, opened 2021-11-24
Enhance `DatabaseAdapter.namedRefs()`

snazy:
Goal of this PR is to optionally expose more information about named references. The caller decides which kind of information shall be retrieved. The default behavior is unchanged and only returns a collection of reference names + type + HEAD hash.

`DatabaseAdapter.namedRefs()` gets a new argument `NamedReferencesParams`, that allows to request:
* exclusion of tags in the result
* exclusion of branches in the result
* retrieving the commit-meta of the named reference's HEAD
* retrieving the common-ancestor of the named reference and the default branch
* retrieving the number of commits behind + ahead relative to the common-ancestor

The type describing each named reference in the result of `DatabaseAdapter.namedRefs()` has been changed to `NamedReference<METADATA>` and allows returning the selected information.

The new types `NamedReference` and  `NamedReferencesParams` are defined in `:nessie-versioned-spi` and are not meant to be exposed via the REST API. Updating the REST API is work for a separate PR.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -40,6 +40,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.Spliterators.AbstractSpliterator;
 import java.util.concurrent.TimeUnit;
@@ -52,11 +53,15 @@ import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import org.projectnessie.versioned.BranchName;
 import org.projectnessie.versioned.Diff;
+import org.projectnessie.versioned.GetNamedRefsParams;
 import org.projectnessie.versioned.Hash;
 import org.projectnessie.versioned.Key;
 import org.projectnessie.versioned.NamedRef;
 import org.projectnessie.versioned.ReferenceConflictException;
+import org.projectnessie.versioned.ReferenceInfo;
+import org.projectnessie.versioned.ReferenceInfo.CommitsAheadBehind;
 import org.projectnessie.versioned.ReferenceNotFoundException;
+import org.projectnessie.versioned.TagName;
 import org.projectnessie.versioned.persist.adapter.CommitAttempt;
 import org.projectnessie.versioned.persist.adapter.CommitLogEntry;
 import org.projectnessie.versioned.persist.adapter.ContentAndState;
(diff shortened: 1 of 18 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/NamedReferencesParams.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Name of the default branch to use as a reference for the computation of {@link
+   * NamedReference#getAheadBehind()}. If this parameter is not {@code null}, the branch must exist.
+   *
+   * @return name of the default branch. Can be {@code null}, if not needed.
+   */
+  @Nullable
+  String getDefaultBranchName();
```
nastra, line 41: can we not just assume it's whatever the default branch is in the config instead of having to pass this explicitly?

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReferencesParams.java
```diff
@@ -0,0 +1,78 @@
...
+  default boolean isComputeCommonAncestor() {
+    return false;
+  }
+
+  /**
+   * Whether to retrieve the commit-meta information of each reference's HEAD commit in {@link
+   * NamedReference#getHeadCommitMeta()}, defaults to {@code false}.
+   */
+  @Value.Default
+  default boolean isRetrieveHead() {
```
nastra, line 63: maybe `isRetrieveCommitMetaForHead`? That's at least more understandable and wouldn't force people to use the javadoc to understand what this method is about ;)

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -114,17 +118,55 @@ public Hash hashOnReference(NamedRef namedReference, Optional<Hash> hashOnRefere
...
+                        || (params.isRetrieveTags()
+                            && r.getValue().getType() == RefPointer.Type.Tag))
+            .map(
+                r ->
+                    NamedReference.<ByteString>builder()
+                        .namedRef(toNamedRef(r.getValue().getType(), r.getKey()))
+                        .hash(Hash.of(r.getValue().getHash()))
+                        .build());
+
+    NonTransactionalOperationContext ctx = NON_TRANSACTIONAL_OPERATION_CONTEXT;
```
nastra, line 149: nit: can be inlined

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -114,17 +118,55 @@ public Hash hashOnReference(NamedRef namedReference, Optional<Hash> hashOnRefere
...
+    refs = namedRefsWithDefaultBranchRelatedInfo(ctx, refs, params, defaultBranchHead);
+
+    if (params.isRetrieveHead()) {
+      refs = refs.map(r -> namedReferenceWithCommitMeta(NON_TRANSACTIONAL_OPERATION_CONTEXT, r));
+    }
+
+    return refs;
+  }
+
+  private Hash defaultBranchHeadForNamedRefs(
```
nastra, line 159: maybe let's mark this as `@Nullable`

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReferencesParams.java
snazy: `DatabaseAdapter` doesn't know the `VersionStore` configuration

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
snazy: Can do in a separate PR (for all interesting places), but would like to keep it out of this one

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReference.java
```diff
@@ -0,0 +1,68 @@
...
+  @Nullable
+  METADATA getHeadCommitMeta();
+
+  NamedReference<METADATA> withHeadCommitMeta(@Nullable METADATA value);
+
+  @SuppressWarnings({"rawtypes", "unchecked"})
+  default <UPDATED_METADATA> NamedReference<UPDATED_METADATA> withUpdatedCommitMeta(
+      UPDATED_METADATA commitMeta) {
+    NamedReference updated = this;
+    return updated.withHeadCommitMeta(commitMeta);
```
dimas-b, line 46: Technically, we're breaking the API contract of `this` here. If its accessors were to be called via an old reference with the old type parameter, it would lead to a class cast exception.

Could we create a separate `NamedReference` object here? For example:
```
default <UPDATED_METADATA> NamedReference<UPDATED_METADATA> reinterpretCommitMeta(Function<METADATA, UPDATED_METADATA> parser) {
   return builder().from(this)...
}
```
WDYT?

nastra (approved):
LGTM with some nits. Addressing those in the PR that will add the same functionality when fetching a single reference seems ok

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReferencesParams.java
dimas-b: I suppose in UI / CLI applications this branch does not have to be the "default" branch. A user may want to compare two arbitrary branches. How about renaming to something like `getRelatedReference()`? I guess it could be a tag too.

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -174,8 +174,9 @@ public void delete(NamedRef ref, Optional<Hash> hash)
   }
 
   @Override
-  public Stream<WithHash<NamedRef>> getNamedRefs() {
-    return callStream("GetNamedRefs", b -> {}, delegate::getNamedRefs);
+  public Stream<NamedReference<METADATA>> getNamedRefs(NamedReferencesParams params) {
+    return callStream(
+        "GetNamedRefs", b -> {}, () -> delegate.getNamedRefs(NamedReferencesParams.DEFAULT));
```
dimas-b: Should we not pass `params` to the inner call?

versioned/spi/src/main/java/org/projectnessie/versioned/MetricsVersionStore.java
```diff
@@ -133,8 +133,9 @@ public void delete(NamedRef ref, Optional<Hash> hash)
   }
 
   @Override
-  public Stream<WithHash<NamedRef>> getNamedRefs() {
-    return delegateStream("getnamedrefs", delegate::getNamedRefs);
+  public Stream<NamedReference<METADATA>> getNamedRefs(NamedReferencesParams params) {
+    return delegateStream(
+        "getnamedrefs", () -> delegate.getNamedRefs(NamedReferencesParams.DEFAULT));
```
dimas-b: Pass `params` to the inner call?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -960,15 +1115,13 @@ protected Hash findCommonAncestor(OP_CONTEXT ctx, Hash from, NamedRef toBranch,
...
       }
 
-      for (Hash f : fromCommitHashes) {
-        if (toCommitHashes.contains(f)) {
-          return f;
+      for (int diffOnFrom = 0; diffOnFrom < fromCommitHashes.size(); diffOnFrom++) {
+        Hash f = fromCommitHashes.get(diffOnFrom);
+        if (state.contains(f)) {
+          return result.apply(diffOnFrom, f);
         }
```
dimas-b, line 1125: Could it be more efficient if we kept N + 1 hash sets (one for each ref, plus one for the "to" branch), fetches commits breadth-first and cross-check each commit against the other maps? I believe list searching would not be required and hopefully less transient data would be used. WDYT?

snazy: I think, it depends on the actual branch/tag model - whether all/most branches/tags are created off of the main branch or whether those are a rather "complex graph".
My assumption here were that:
* most branches are rather new(-ish) - aka: not that far away from the default branch's HEAD
* the UI does not request the extended information for tags

snazy: But yes, I also see that there's probably room for improvement here.

For the use case of showing the "additional information" for all branches: Git(Hub) seems to do it on the fly. At least the corresponding `git` command is pretty quick. But unlike Git we (Nessie) cannot access "everything quickly" (like just read the missing pieces off the disk).

For the use case of a merge/transplant: It's only one branch vs another branch - so probably not that much of an issue.

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReferencesParams.java
snazy: I'm fine with adding that to `DatabaseAdapter`. There's technically no difference for the implementation.

But I'd rather leave that functionality out of the REST API for now to not make it too complicated to use.

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReference.java
snazy: Why should it cause a CCE?
It's setting the commit-meta to an instance of `UPDATED_METADATA`, so the old one's is gone.

(I've only added this function so that call sites do not have to deal with raw types.)

dimas-b: I mean there's no problem in existing code that uses `NamedReference`. My comment was "in principle" :) ... something like this:

```
NamedReference<ByteString> ref0 = ...
NamedReference<CommitMeta> ref1 = ref0.withUpdatedCommitMeta(...);
ByteString rawBytes = ref0.getCommitMeta(); // compiler is happy here, but CCE in runtime, AFAIK
```

snazy: No problem. That `with*` creates a copy

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReference.java
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned;
+
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface NamedReference<METADATA> {
```
dimas-b, line 22: Just realized that we now have `NamedRef` and `NamedReference`... kind of confusing IMHO :)

How about renaming this to `ReferenceInfo` or `ReferenceDetails`?

snazy: Renamed `NamedReference` to `ReferenceInfo` and `NamedReferencesParams` to `GetNamedRefsParams`.

nastra: +1 to renaming

nastra (approved):

versioned/spi/src/main/java/org/projectnessie/versioned/NamedReference.java
dimas-b: Ah, yes, I missed the copy in generated `with` methods :facepalm:

PR #2734, opened 2021-11-25
Optionally fetch metadata properties and attach them to a Branch

nastra:
Additional metadata for Branches can now be fetched on the REST API via
the `fetchMetadata=true` flag.

The metadata is then returned within a `Branch` instance via the newly introduced
`metadataProperties` map.

The `metadataProperties` map currently has information under the
`numCommitsAhead` / `numCommitsBehind` / `headCommitMeta` /
`commonAncestorHash` keys.

Note that these changes currently do not affect the Python client in any way, since the Python client doesn't pass the `fetchAdditionalInfo=true` flag. This also means that we don't need to change the schema in the Python's client model classes due to https://github.com/projectnessie/nessie/pull/2734/files#diff-f943e12034e6350b04e3fba9735a5996106b2e88afe9302638244fe2ba77ea76R76.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
@@ -23,5 +23,20 @@ import org.projectnessie.model.ReferencesResponse;
  * @since {@link NessieApiV1}
  */
 public interface GetAllReferencesBuilder extends PagingBuilder<GetAllReferencesBuilder> {
+
+  /**
+   * Will fetch additional metadata about {@link org.projectnessie.model.Branch} instances, such as
+   * number of commits ahead/behind or the common ancestor in relation to the default branch, and
+   * the commit metadata for the HEAD commit.
+   *
+   * @return {@link GetAllReferencesBuilder}
+   */
+  GetAllReferencesBuilder fetchAdditionalInfo(boolean fetchAdditionalInfo);
+
+  /**
+   * Fetches all references and returns them in a {@link ReferencesResponse} instance.
+   *
+   * @return Fetches all references and returns them in a {@link ReferencesResponse} instance.
+   */
   ReferencesResponse get();
 }
--- a/model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
+++ b/model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
@@ -17,10 +17,13 @@ package org.projectnessie.api.params;
 
 import java.util.Objects;
 import java.util.StringJoiner;
+import javax.ws.rs.QueryParam;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.projectnessie.api.http.HttpTreeApi;
 
 /**
- * The purpose of this class is to include optional parameters that can be passed to {@code
- * HttpTreeApi#getEntries(String, EntriesParams)}.
+ * The purpose of this class is to include optional parameters that can be passed to {@link
+ * HttpTreeApi#getAllReferences(ReferencesParams)}
  *
  * <p>For easier usage of this class, there is {@link ReferencesParams#builder()}, which allows
  * configuring/setting the different parameters.
@@ -29,12 +32,29 @@ public class ReferencesParams extends AbstractParams {
 
   public ReferencesParams() {}
 
-  private ReferencesParams(Integer maxRecords, String pageToken) {
+  @Parameter(
+      description =
+          "If set to true, will fetch additional metadata for branches, such as number of commits ahead/behind or the common ancestor in relation to the default branch, and the commit metadata for the HEAD commit.\n\n"
+              + "A returned Branch instance will then have a map of 'metadataProperties' with the following keys:\n\n"
+              + "- numCommitsAhead (number of commits ahead of the default branch)\n\n"
+              + "- numCommitsBehind (number of commits behind the default branch)\n\n"
+              + "- headCommitMeta (the commit metadata of the HEAD commit)\n\n"
+              + "- commonAncestorHash (the hash of the common ancestor in relation to the default branch).\n\n"
+              + "Note that computing & fetching additional metadata might be computationally expensive on the server-side, so this flag should be used with care.")
+  @QueryParam("fetchAdditionalInfo")
+  private boolean fetchAdditionalInfo;
+
+  public boolean isFetchAdditionalInfo() {
+    return fetchAdditionalInfo;
+  }
+
+  private ReferencesParams(Integer maxRecords, String pageToken, boolean fetchAdditionalInfo) {
     super(maxRecords, pageToken);
+    this.fetchAdditionalInfo = fetchAdditionalInfo;
   }
 
   private ReferencesParams(Builder builder) {
-    this(builder.maxRecords, builder.pageToken);
+    this(builder.maxRecords, builder.pageToken, builder.fetchAdditionalInfo);
   }
 
   public static ReferencesParams.Builder builder() {
@@ -50,6 +70,7 @@ public class ReferencesParams extends AbstractParams {
     return new StringJoiner(", ", ReferencesParams.class.getSimpleName() + "[", "]")
         .add("maxRecords=" + maxRecords())
         .add("pageToken='" + pageToken() + "'")
+        .add("fetchAdditionalInfo=" + fetchAdditionalInfo)
         .toString();
   }
 
@@ -63,20 +84,30 @@ public class ReferencesParams extends AbstractParams {
     }
     ReferencesParams that = (ReferencesParams) o;
     return Objects.equals(maxRecords(), that.maxRecords())
-        && Objects.equals(pageToken(), that.pageToken());
+        && Objects.equals(pageToken(), that.pageToken())
+        && Objects.equals(fetchAdditionalInfo, that.fetchAdditionalInfo);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(maxRecords(), pageToken());
+    return Objects.hash(maxRecords(), pageToken(), fetchAdditionalInfo);
   }
 
   public static class Builder extends AbstractParams.Builder<Builder> {
 
     private Builder() {}
 
+    private boolean fetchAdditionalInfo = false;
+
     public ReferencesParams.Builder from(ReferencesParams params) {
-      return maxRecords(params.maxRecords()).pageToken(params.pageToken());
+      return maxRecords(params.maxRecords())
+          .pageToken(params.pageToken())
+          .fetchAdditionalInfo(params.fetchAdditionalInfo);
+    }
+
+    public Builder fetchAdditionalInfo(boolean fetchAdditionalInfo) {
+      this.fetchAdditionalInfo = fetchAdditionalInfo;
+      return this;
     }
 
     private void validate() {}
--- a/model/src/main/java/org/projectnessie/model/Branch.java
+++ b/model/src/main/java/org/projectnessie/model/Branch.java
@@ -17,9 +17,13 @@ package org.projectnessie.model;
 
 import static org.projectnessie.model.Validation.validateReferenceName;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.Collections;
+import java.util.Map;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.immutables.value.Value;
@@ -31,6 +35,10 @@ import org.immutables.value.Value;
 @JsonDeserialize(as = ImmutableBranch.class)
 @JsonTypeName("BRANCH")
 public interface Branch extends Reference {
+  String NUM_COMMITS_AHEAD = "numCommitsAhead";
+  String NUM_COMMITS_BEHIND = "numCommitsBehind";
+  String HEAD_COMMIT_META = "headCommitMeta";
+  String COMMON_ANCESTOR_HASH = "commonAncestorHash";
 
   /**
    * Validation rule using {@link org.projectnessie.model.Validation#validateReferenceName(String)}.
@@ -46,5 +54,27 @@ public interface Branch extends Reference {
 
   static Branch of(String name, String hash) {
     return builder().name(name).hash(hash).build();
+  }
+
+  /**
+   * Returns a map of metadata properties that describe this branch. Note that these properties
+   * <b>can be added</b> by the server when a {@link Branch} instance is returned. A returned {@link
+   * Branch} instance can then have the following keys:
+   *
+   * <ul>
+   *   <li>{@value Branch#NUM_COMMITS_AHEAD}: number of commits ahead of the default branch
+   *   <li>{@value Branch#NUM_COMMITS_BEHIND}: number of commits behind the default branch
+   *   <li>{@value Branch#HEAD_COMMIT_META}: the commit metadata of the HEAD commit
+   *   <li>{@value Branch#COMMON_ANCESTOR_HASH}: the hash of the common ancestor in relation to the
+   *       default branch.
+   * </ul>
+   *
+   * @return A map of metadata properties that describe this branch. Note that these properties
+   *     <b>can be added</b> by the server when a {@link Branch} instance is returned.
+   */
+  @Value.Default
+  @JsonInclude(Include.NON_EMPTY)
+  default Map<String, Object> metadataProperties() {
+    return Collections.emptyMap();
   }
 }
--- a/model/src/main/resources/META-INF/openapi.yaml
+++ b/model/src/main/resources/META-INF/openapi.yaml
@@ -189,3 +189,50 @@ components:
           - type: BRANCH
             hash: "2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d"
             name: "main"
+    referencesResponseWithMetadata:
+      value:
+        token: null
+        hasMore: false
+        references:
+          - type: BRANCH
+            hash: "84c57a20c5e956af4af40f3cc34ecc8a9028b9586da135c79011b1867aa3191d"
+            name: "main"
+            metadataProperties:
+              headCommitMeta:
+                hash: "84c57a20c5e956af4af40f3cc34ecc8a9028b9586da135c79011b1867aa3191d"
+                committer: ""
+                author: "nessie-author"
+                signedOffBy: null
+                message: "update table"
+                commitTime: "2021-11-26T08:01:13.855974Z"
+                authorTime: "2021-11-26T08:01:13.852826Z"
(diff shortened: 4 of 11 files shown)


model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -49,4 +51,15 @@ default void checkName() {
...
   }
+
+  /**
+   * Returns a map of metadata properties that describe this branch. Note that these properties
+   * <b>can be added</b> by the server when a {@link Branch} instance is returned.
+   *
+   * @return A map of metadata properties that describe this branch. Note that these properties
+   *     <b>can be added</b> by the server when a {@link Branch} instance is returned.
+   */
+  @Nullable
```
snazy, line 62: Should we maybe default to an empty map and mark this one as not-null?

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -29,12 +32,24 @@
 
   public ReferencesParams() {}
 
-  private ReferencesParams(Integer maxRecords, String pageToken) {
+  @Parameter(
+      description =
+          "If set to true, will fetch additional metadata for branches, such as number of commits ahead/behind or the common ancestor in relation to the default branch, and the commit metadata for the HEAD commit. "
+              + "Note that computing & fetching additional metadata might be computationally expensive on the server-side, so this flag should be used with care.")
+  @QueryParam("fetchMetadata")
+  private boolean fetchAdditionalMetadata;
```
snazy, line 40: We also need another parameter to specify the reference-type to fetch (only branches, only tags, branches + tags).
Reason is that computing the common-ancestor + commits-behind/ahead is expensive.
Similar to https://github.com/snazy/nessie/branches vs https://github.com/snazy/nessie/tags

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -496,15 +506,34 @@ private static Reference makeRef(NamedReference<CommitMeta> refWithHash) {
...
     } else {
       throw new UnsupportedOperationException("only converting tags or branches"); // todo
     }
   }
 
+  private static Map<String, Object> extractMetadataProperties(
+      NamedReference<CommitMeta> refWithHash) {
+    Map<String, Object> metadataProperties = new HashMap<>();
+    if (null != refWithHash.getAheadBehind()) {
+      metadataProperties.put("numCommitsAhead", refWithHash.getAheadBehind().getAhead());
```
snazy, line 525: Need contstants for these - and have the keys documented in the REST API, with an example for each.

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -175,8 +175,7 @@ public void delete(NamedRef ref, Optional<Hash> hash)
 
   @Override
   public Stream<NamedReference<METADATA>> getNamedRefs(NamedReferencesParams params) {
-    return callStream(
-        "GetNamedRefs", b -> {}, () -> delegate.getNamedRefs(NamedReferencesParams.DEFAULT));
+    return callStream("GetNamedRefs", b -> {}, () -> delegate.getNamedRefs(params));
```
snazy, line 178: This one (and the other one) are fixed in #2726 - probably also need to rebase

snazy (commented):
Just did a very brief first pass on this one, but left some comments.
Note: this needs a rebase.

model/src/main/java/org/projectnessie/model/Branch.java
dimas-b: +1 to a non-null empty map default

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -47,4 +51,15 @@ default void checkName() {
...
+  @JsonInclude(Include.NON_NULL)
+  Map<String, Object> metadataProperties();
```
dimas-b, line 64: Why not have a strongly typed holder class for the metadata? It could extend `Map` if we want to allow clients to write more generic code for processing those properties.

model/src/main/java/org/projectnessie/model/Branch.java
nastra: I was thinking about that and the reason why I thought that going with null & not serializing this is so that the python client isn't affected by this change (since the python client doesn't even ask for additional metadata).
We could also use an empty map by default and annotate it with `@JsonInclude(Include.NON_EMPTY)` but the first option seemed to make more sense. thoughts?

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
nastra: let's do this in a follow-up PR as otherwise that would pollute this one

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1783,6 +1784,62 @@ public void testAssignRefToFreshMain() throws BaseNessieClientServerException {
...
+                  .extracting("numCommitsAhead")
+                  .isEqualTo(commitsPerBranch);
+              assertThat(metadataProperties)
+                  .hasFieldOrProperty("numCommitsBehind")
+                  .extracting("numCommitsBehind")
+                  .isEqualTo(0);
+              assertThat(metadataProperties)
+                  .hasFieldOrProperty("headCommitMeta")
+                  .extracting("headCommitMeta")
+                  .isNotNull();
```
snazy, line 1835: Better assert the content of the `CommitMeta` here.

nastra: done

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1783,6 +1785,79 @@ public void testAssignRefToFreshMain() throws BaseNessieClientServerException {
...
+    // not fetching additional metadata
+    List<Reference> references = api.getAllReferences().get().getReferences();
+    Optional<Reference> main =
+        references.stream().filter(r -> r.getName().equals("main")).findFirst();
+    assertThat(main).isPresent();
+    List<Branch> branches =
+        references.stream()
+            .filter(r -> r.getName().startsWith(branchPrefix))
+            .map(r -> (Branch) r)
+            .collect(Collectors.toList());
```
snazy, line 1808: BTW: If you combine the two `assertThat(branches)` in both occurences, you can omit the `.collect(Collectors.toList())` and assert directly on the `Stream`

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
```diff
@@ -23,5 +23,20 @@
...
 public interface GetAllReferencesBuilder extends PagingBuilder<GetAllReferencesBuilder> {
+
+  /**
+   * Will fetch additional metadata about {@link org.projectnessie.model.Branch} instances, such as
+   * number of commits ahead/behind or the common ancestor in relation to the default branch, and
+   * the commit metadata for the HEAD commit.
+   *
+   * @return {@link GetAllReferencesBuilder}
+   */
+  GetAllReferencesBuilder fetchAdditionalMetadata();
```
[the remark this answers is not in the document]
nastra: renamed to `fetchAdditionalInfo`

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -50,6 +70,7 @@ public String toString() {
     return new StringJoiner(", ", ReferencesParams.class.getSimpleName() + "[", "]")
         .add("maxRecords=" + maxRecords())
         .add("pageToken='" + pageToken() + "'")
+        .add("fetchAdditionalMetadata=" + fetchAdditionalMetadata)
```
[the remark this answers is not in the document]
nastra: renamed to `fetchAdditionalInfo`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1783,6 +1785,79 @@ public void testAssignRefToFreshMain() throws BaseNessieClientServerException {
...
+                  .extracting(Branch.NUM_COMMITS_AHEAD)
+                  .isEqualTo(commitsPerBranch);
+              assertThat(metadataProperties)
+                  .hasFieldOrProperty(Branch.NUM_COMMITS_BEHIND)
+                  .extracting(Branch.NUM_COMMITS_BEHIND)
+                  .isEqualTo(0);
+              assertThat(metadataProperties)
+                  .hasFieldOrProperty(Branch.HEAD_COMMIT_META)
+                  .extracting(Branch.HEAD_COMMIT_META)
+                  .isNotNull()
```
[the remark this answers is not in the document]
nastra: yep good idea, will change accordingly

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1783,6 +1785,79 @@ public void testAssignRefToFreshMain() throws BaseNessieClientServerException {
...
+    Optional<Reference> main =
+        references.stream().filter(r -> r.getName().equals("main")).findFirst();
+    assertThat(main).isPresent();
+    List<Branch> branches =
+        references.stream()
+            .filter(r -> r.getName().startsWith(branchPrefix))
+            .map(r -> (Branch) r)
+            .collect(Collectors.toList());
+    assertThat(branches).hasSize(numBranches);
+    assertThat(branches).allSatisfy(branch -> assertThat(branch.metadataProperties()).isEmpty());
```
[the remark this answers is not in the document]
nastra: done

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
nastra: good point, updated

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
nastra: It's just a single parameter, but it requires additional testing and it's not exactly in the scope of "fetch me all metadata". Therefore I think it's cleaner to have this as a separate PR

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -63,20 +84,30 @@ public boolean equals(Object o) {
...
+    private boolean fetchAdditionalMetadata = false;
+
     public ReferencesParams.Builder from(ReferencesParams params) {
-      return maxRecords(params.maxRecords()).pageToken(params.pageToken());
+      return maxRecords(params.maxRecords())
+          .pageToken(params.pageToken())
+          .fetchAdditionalInfo(params.fetchAdditionalInfo);
+    }
+
+    public Builder fetchAdditionalInfo(boolean fetchAdditionalMetadata) {
```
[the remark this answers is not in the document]
nastra: looks like refactoring in IntelliJ didn't update everything :(

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -29,12 +32,29 @@
 
   public ReferencesParams() {}
 
-  private ReferencesParams(Integer maxRecords, String pageToken) {
+  @Parameter(
+      description =
+          "If set to true, will fetch additional metadata for branches, such as number of commits ahead/behind or the common ancestor in relation to the default branch, and the commit metadata for the HEAD commit.\n\n"
+              + "A returned Branch instance will then have a map of 'metadataProperties' with the following keys:\n\n"
+              + "- numCommitsAhead (number of commits ahead of the default branch)\n\n"
```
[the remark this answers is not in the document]
nastra: that's apparently the only way to actually have a newline in the Swagger UI when looking at the parameter docs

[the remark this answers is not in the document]
snazy: duh - annoying :(

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -63,20 +84,30 @@ public boolean equals(Object o) {
...
   public int hashCode() {
-    return Objects.hash(maxRecords(), pageToken());
+    return Objects.hash(maxRecords(), pageToken(), fetchAdditionalInfo);
   }
 
   public static class Builder extends AbstractParams.Builder<Builder> {
 
     private Builder() {}
 
+    private boolean fetchAdditionalMetadata = false;
```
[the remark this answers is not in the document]
nastra: updated

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -50,6 +70,7 @@ public String toString() {
     return new StringJoiner(", ", ReferencesParams.class.getSimpleName() + "[", "]")
         .add("maxRecords=" + maxRecords())
         .add("pageToken='" + pageToken() + "'")
+        .add("fetchAdditionalMetadata=" + fetchAdditionalInfo)
```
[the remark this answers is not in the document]
nastra: updated

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -47,4 +55,26 @@ default void checkName() {
...
     return builder().name(name).hash(hash).build();
   }
+
+  /**
+   * Returns a map of metadata properties that describe this branch. Note that these properties
+   * <b>can be added</b> by the server when a {@link Branch} instance is returned. A returned {@link
+   * Branch} instance can then have the following keys:
+   *
+   * <ul>
+   *   <li>{@link Branch#NUM_COMMITS_AHEAD}: number of commits ahead of the default branch
```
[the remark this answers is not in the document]
nastra: updated

snazy (approved):

model/src/main/java/org/projectnessie/model/Branch.java
nastra: what would be the benefit? To me it seemed having a simple `Map<String, Object>` would be flexible enough in case we want to put more info in the future

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -29,12 +32,29 @@
...
+              + "A returned Branch instance will then have a map of 'metadataProperties' with the following keys:\n\n"
+              + "- numCommitsAhead (number of commits ahead of the default branch)\n\n"
+              + "- numCommitsBehind (number of commits behind the default branch)\n\n"
+              + "- headCommitMeta (the commit metadata of the HEAD commit)\n\n"
+              + "- commonAncestorHash (the hash of the common ancestor in relation to the default branch).\n\n"
+              + "Note that computing & fetching additional metadata might be computationally expensive on the server-side, so this flag should be used with care.")
+  @QueryParam("fetchAdditionalInfo")
+  private boolean fetchAdditionalInfo;
+
+  public boolean isFetchAdditionalInfo() {
```
dimas-b, line 47: nit: odd method order: getter above constructor?

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -47,4 +55,26 @@ default void checkName() {
...
+
+  /**
+   * Returns a map of metadata properties that describe this branch. Note that these properties
+   * <b>can be added</b> by the server when a {@link Branch} instance is returned. A returned {@link
+   * Branch} instance can then have the following keys:
+   *
+   * <ul>
+   *   <li>{@value Branch#NUM_COMMITS_AHEAD}: number of commits ahead of the default branch
+   *   <li>{@value Branch#NUM_COMMITS_BEHIND}: number of commits behind the default branch
+   *   <li>{@value Branch#HEAD_COMMIT_META}: the commit metadata of the HEAD commit
```
dimas-b, line 67: Should we specify metadata value types, perhaps?

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -47,4 +55,26 @@ default void checkName() {
...
+   *   <li>{@value Branch#COMMON_ANCESTOR_HASH}: the hash of the common ancestor in relation to the
+   *       default branch.
+   * </ul>
+   *
+   * @return A map of metadata properties that describe this branch. Note that these properties
+   *     <b>can be added</b> by the server when a {@link Branch} instance is returned.
+   */
+  @Value.Default
+  @JsonInclude(Include.NON_EMPTY)
+  default Map<String, Object> metadataProperties() {
```
dimas-b, line 77: Would it work with a custom `Map` sub-class? This way we could have typed getters for well-known metadata keys. WDYT?

model/src/main/resources/META-INF/openapi.yaml
```diff
@@ -189,3 +189,50 @@ components:
...
+            hash: "da086850076827d2989c8ee1d7fd22f152f525d46a0441f2b22ad8119c0bbbe5"
+            name: "dev"
+            metadataProperties:
+              numCommitsAhead: 1
+              numCommitsBehind: 2
+              commonAncestorHash: "2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d"
+              headCommitMeta:
+                hash: "da086850076827d2989c8ee1d7fd22f152f525d46a0441f2b22ad8119c0bbbe5"
+                committer: ""
+                author: "nessie-author"
```
dimas-b, line 220: This will deserialize to a plain `Map` not to `CommitMeta`, right? Would it not be nice to expose the same value type in this case as in `LogResponse`?

snazy: That's true. Maybe just move the commit-meta to a separate field. The other attribtues are simple/primitive ones.

model/src/main/java/org/projectnessie/model/Branch.java
nastra: that's the part I'm not sure about. I'm a bit concerned that this might get complicated

model/src/main/java/org/projectnessie/model/Branch.java
nastra: that's the part I'm not sure about

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
nastra: I'll fix it in the next PR that is coming

dimas-b (approved):
we can certainly revisit strong typing points in a follow-up PR :)

model/src/main/java/org/projectnessie/model/Branch.java
dimas-b: it might, but at the same time, stronger compile time checks can help to ensure backward compatibility, I hope.

model/src/main/resources/META-INF/openapi.yaml
nastra: let me think about it a bit more how an implementation would actually look like. We agreed with @dimas-b to look at this in a follow-up

PR #2747, opened 2021-11-26
Fix build due to merge non-conflict

snazy:


servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -106,9 +107,8 @@ public ReferencesResponse getAllReferences(ReferencesParams params) {
         params.isFetchAdditionalInfo()
             ? GetNamedRefsParams.builder()
                 .baseReference(BranchName.of(this.getConfig().getDefaultBranch()))
-                .isComputeAheadBehind(true)
-                .isRetrieveCommitMetaForHead(true)
-                .isComputeCommonAncestor(true)
+                .branchRetrieveOptions(RetrieveOptions.BASE_REFERENCE_RELATED_AND_COMMIT_META)
+                .tagRetrieveOptions(RetrieveOptions.COMMIT_META)
```
nastra, line 111: could we also set this to `tagRetrieveOptions(RetrieveOptions.BARE)` since we currently don't show any metadata for Tags?

nastra (approved):

PR #2748, opened 2021-11-26
Stricter typing for Branch metadata / Move metadata properties into ReferenceMetadata class

nastra:


model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -57,24 +52,43 @@ static Branch of(String name, String hash) {
...
+          "Only returned by the server when explicitly requested by the client and contains the following information:\n\n"
+              + "- numCommitsAhead (number of commits ahead of the default branch)\n\n"
+              + "- numCommitsBehind (number of commits behind the default branch)\n\n"
+              + "- commitMetaOfHEAD (the commit metadata of the HEAD commit)\n\n"
+              + "- commonAncestorHash (the hash of the common ancestor in relation to the default branch).\n\n")
+  @Value.Immutable
+  @JsonSerialize(as = ImmutableBranchMetadata.class)
+  @JsonDeserialize(as = ImmutableBranchMetadata.class)
+  @JsonTypeName("BRANCH_METADATA")
+  interface BranchMetadata {
```
dimas-b, line 80: I wonder if it makes sense to make this a top-level interface :thinking: Nessie backend is capable of producing this kind metadata for tags as well.

dimas-b (commented):
LGTM overall. Thanks for looking into this.

model/src/main/java/org/projectnessie/model/Branch.java
nastra: that makes sense. I moved this into `ReferenceMetadata` and will add the metadata properties to Tags in the next PR

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -57,24 +52,15 @@ static Branch of(String name, String hash) {
...
+   *     by the client.
    */
-  @Value.Default
-  @JsonInclude(Include.NON_EMPTY)
-  default Map<String, Object> metadataProperties() {
-    return Collections.emptyMap();
-  }
+  @JsonInclude(Include.NON_NULL)
+  @Nullable
+  ReferenceMetadata metadata();
```
snazy, line 65: I think, this should live in `Reference`

nastra: fixed

snazy (approved):

PR #2749, opened 2021-11-27
Enhance commit log

snazy:
Allows to optionally return additional information from the commit log.

The `LogResponse` type has been changed to contain a list of `LogEntry` elements, which contain a `CommitMeta` and, if requested, also the operations (`Put` or `Delete`) of each commit.

The Python CLI behaves as it currently does. The `nessie log` command got a new flag to retrieve and show the additional information.

* Adds new parameter `fetchAdditionalInfo` to commit-log REST endpoint.
* Changes return type of commit-log REST endpoint, returns a list of `LogEntry`
  that contains the `CommitMeta` plus the optionally fetched list of operations.
* Adds a `--extended` flag to Nessie CLI's `log` command to include the additional
  information. Behaves exactly as before without the `--extended` flag.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/api/GetCommitLogBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/api/GetCommitLogBuilder.java
@@ -30,6 +30,14 @@ public interface GetCommitLogBuilder
     extends QueryBuilder<GetCommitLogBuilder>,
         PagingBuilder<GetCommitLogBuilder>,
         OnReferenceBuilder<GetCommitLogBuilder> {
+
+  /**
+   * Will fetch additional metadata about each commit like operations in a commit and parent hash.
+   *
+   * @return {@link GetAllReferencesBuilder}
+   */
+  GetCommitLogBuilder fetchAdditionalInfo(boolean fetchAdditionalInfo);
+
   GetCommitLogBuilder untilHash(
       @Nullable @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
           String untilHash);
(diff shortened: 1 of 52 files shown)


clients/client/src/main/java/org/projectnessie/client/api/GetCommitLogBuilder.java
```diff
@@ -30,6 +30,14 @@
     extends QueryBuilder<GetCommitLogBuilder>,
         PagingBuilder<GetCommitLogBuilder>,
         OnReferenceBuilder<GetCommitLogBuilder> {
+
+  /**
+   * Will fetch additional metadata about each commit: operations in a commit and parent hash.
```
nastra, line 35: maybe "about each commit, such as operations in a commit and parent hash"

model/src/test/java/org/projectnessie/api/params/CommitLogParamsTest.java
```diff
@@ -28,30 +29,53 @@ public void testBuilder() {
...
+  private void verify(
+      Integer maxRecords,
+      String startHash,
+      String endHash,
+      String pageToken,
+      String queryExpression,
+      boolean fetchAdditionalInfo,
+      Supplier<CommitLogParams> generator) {
+    assertThat(generator.get())
+        .isEqualTo(generator.get())
```
nastra, line 63: looks a bit weird to compare `generator.get()` with itself

nastra (commented):
LGTM, just a few nits

model/src/test/java/org/projectnessie/api/params/CommitLogParamsTest.java
snazy: It's not the _same_ object

nastra (approved):

PR #2767, opened 2021-11-29
Track total number of commits

snazy:
Allows retrieval of the total number of commits on a reference.

Via REST: in the `ReferenceMetadata`, if `fetchAdditionalInfo` is true.
Via VersionStore/DatabaseAdapter: if `retrieveCommitMetaForHead` is true.

The total number of commits is tracked as a field in every commit. The
first commit since "beginning of time" starts at `1`, following commits
"increment" this number. Since the commit history is immutable, the
number of commits since "beginning of time" is immutable as well.

### DIFF
--- a/model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
+++ b/model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
@@ -38,7 +38,8 @@ public class ReferencesParams extends AbstractParams {
               + "- numCommitsBehind (number of commits behind the default branch)\n\n"
               + "- commitMetaOfHEAD (the commit metadata of the HEAD commit)\n\n"
               + "- commonAncestorHash (the hash of the common ancestor in relation to the default branch).\n\n"
-              + "A returned Tag instance will only contain the 'commitMetaOfHEAD' field.\n\n"
+              + "- numTotalCommits (the total number of commits in this reference).\n\n"
+              + "A returned Tag instance will only contain the 'commitMetaOfHEAD' and 'numTotalCommits' fields.\n\n"
               + "Note that computing & fetching additional metadata might be computationally expensive on the server-side, so this flag should be used with care.")
   @QueryParam("fetchAdditionalInfo")
   private boolean fetchAdditionalInfo;
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
@@ -1906,6 +1906,7 @@ public abstract class AbstractTestRest {
     assertThat(referenceMetadata.numCommitsBehind()).isNull();
     assertThat(referenceMetadata.commitMetaOfHEAD()).isEqualTo(commitMeta);
     assertThat(referenceMetadata.commonAncestorHash()).isNull();
+    assertThat(referenceMetadata.numTotalCommits()).isEqualTo(10);
   }
 
   protected void unwrap(Executable exec) throws Throwable {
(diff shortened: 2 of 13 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1883,6 +1891,7 @@ private void verifyMetadataProperties(
     assertThat(referenceMetadata.numCommitsBehind()).isEqualTo(expectedCommitsBehind);
     assertThat(referenceMetadata.commitMetaOfHEAD()).isEqualTo(commitMeta);
     assertThat(referenceMetadata.commonAncestorHash()).isEqualTo(reference.getHash());
+    assertThat(referenceMetadata.numTotalCommits()).isEqualTo(expectedCommits);
```
nastra, line 1894: needs to have a `isNull` assertion in `verifyMetadataProperties(Tag tag)`

nastra (changes requested):
Requires doc updates in `GetReferenceParams` / `ReferenceMetadata` / `ReferencesParams` to mention the new field. The `openapi.yaml` needs to be updated as well to mention this new field in the example response

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
snazy: fixed

nastra (dismissed):

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -38,6 +38,7 @@
               + "- numCommitsBehind (number of commits behind the default branch)\n\n"
               + "- commitMetaOfHEAD (the commit metadata of the HEAD commit)\n\n"
               + "- commonAncestorHash (the hash of the common ancestor in relation to the default branch).\n\n"
+              + "- numTotalCommits (the total number of commits in this reference).\n\n"
               + "A returned Tag instance will only contain the 'commitMetaOfHEAD' field.\n\n"
```
nastra, line 42: This line (and the other ones in the other classes) needs adjustment given that we're now returning `numTotalCommits` for Tags as well

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -1897,6 +1906,7 @@ private void verifyMetadataProperties(Tag tag) throws NessieNotFoundException {
     assertThat(referenceMetadata.numCommitsBehind()).isNull();
     assertThat(referenceMetadata.commitMetaOfHEAD()).isEqualTo(commitMeta);
     assertThat(referenceMetadata.commonAncestorHash()).isNull();
+    assertThat(referenceMetadata.numTotalCommits()).isEqualTo(10);
```
nastra, line 1909: hm this should only be set if `fetchAdditionalInformation` was set to `true`, otherwise we'll be breaking the Python client (since it doesn't know anything about `ReferenceMetadata` yet.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -548,6 +548,10 @@ private static ReferenceMetadata extractReferenceMetadata(ReferenceInfo<CommitMe
       builder.commitMetaOfHEAD(
           addHashToCommitMeta(refWithHash.getHash(), refWithHash.getHeadCommitMeta()));
     }
+    if (0L != refWithHash.getCommitSeq()) {
+      found = true;
+      builder.numTotalCommits(refWithHash.getCommitSeq());
```
nastra, line 553: we should only add that if `fetchAdditionalInformation` was set to `true`, otherwise we'll be breaking the Python client (since it doesn't know anything about `ReferenceMetadata` yet.

nastra (changes requested):

nastra (approved):

PR #2768, opened 2021-11-29
Rename LogResponse#getOperations to LogResponse#getCommits

nastra:


snazy:
This is being done in #2749

PR #2782, opened 2021-11-30
Filter named references and keys

snazy:
Adds an optional `filter` parameter to the REST endpoints to get all
named references and get entries for a reference.

Filtering implementation is implemented in a very simple way, it just
uses `String.contains()` on the name of the reference / the string
representation of a `Key`.


rymurr:
what does this do that the query expression cant? sorry for the simpleton question

PR #2783, opened 2021-11-30
Introduce DiffApi to retrieve diffs between two references

nastra:


servers/services/src/main/java/org/projectnessie/services/impl/DiffApiImpl.java
```diff
@@ -0,0 +1,74 @@
...
+        builder.addAllDiffs(
+            diffs
+                .map(
+                    diff ->
+                        ImmutableDiff.builder()
+                            .key(ContentKey.of(diff.getKey().getElements()))
+                            .from(diff.getFromValue().orElse(null))
+                            .to(diff.getToValue().orElse(null))
+                            .build())
+                .collect(Collectors.toList()));
```
snazy, line 66: no need to `.collect()` here - just `.forEach(builder::addDiffs);`

model/src/main/java/org/projectnessie/api/http/HttpDiffApi.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
+import org.projectnessie.api.params.DiffParams;
+import org.projectnessie.error.NessieNotFoundException;
+import org.projectnessie.model.DiffResponse;
+
+@Path("diffs")
+public interface HttpDiffApi {
+
+  @GET
+  @Produces(MediaType.APPLICATION_JSON)
```
dimas-b, line 38: Please add `@Consumes` too... perhaps at the class level, for completeness.

model/src/main/resources/META-INF/openapi.yaml
```diff
@@ -298,3 +299,27 @@ components:
...
+              id: "f350b391-f492-41eb-9959-730a8c49f01e"
+              metadataLocation: "/path/to/metadata/"
+              snapshotId: 23
+              schemaId: 15
+              specId: 15
+              sortOrderId: 15
+            to:
+              type: ICEBERG_TABLE
+              id: "dec31d0a-7d4b-4534-8c24-24f08eda33b2"
+              metadataLocation: "/path/to/metadata/"
```
dimas-b, line 321: nit: since the `id` is different it might be best to use a different metadata path too, to minimize user confusion chances.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractResteasyTest.java
```diff
@@ -420,4 +438,31 @@ public void testGetContent() {
...
+            .statusCode(200)
+            .extract()
+            .as(DiffResponse.class);
+
+    assertThat(diffResponse).isNotNull();
+    assertThat(diffResponse.diffs()).hasSize(1);
+    Diff diff = diffResponse.diffs().get(0);
+    assertThat(diff.key()).isEqualTo(contentKey);
+    assertThat(diff.from()).isEqualTo(fromTable);
+    assertThat(diff.to()).isEqualTo(toTable);
```
dimas-b, line 466: add test cases to add/delete differences?

model/src/main/resources/META-INF/openapi.yaml
nastra: this is from an example where `metadataLocation` was actually the same, so it's still a valid example.

clients/client/src/main/java/org/projectnessie/client/api/GetDiffsBuilder.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Request builder for retrieving diffs between two references.
+ *
+ * @since {@link NessieApiV1}
+ */
+public interface GetDiffsBuilder {
+
+  GetDiffsBuilder fromRef(String fromRef);
+
+  GetDiffsBuilder toRef(String toRef);
```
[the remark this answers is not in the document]
nastra: while this would work on the builder, we can only use the string representation of a reference on the REST endpoint. Otherwise we would have to make it a POST request and I'd really like to avoid that.

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
```diff
@@ -99,4 +99,7 @@
   MergeReferenceBuilder mergeRefIntoBranch();
 
   CommitMultipleOperationsBuilder commitMultipleOperations();
+
+  /** Retrieve diffs between two references. */
+  GetDiffsBuilder getDiffs();
```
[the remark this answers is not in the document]
nastra: in the case of `Content` vs `Contents` it was more straight-forward, since `Content` would always only contain a single object. For the diffs however, we return a list of `Diff` objects, so it seemed more appropriate to me to use the plural form, but I don't have a strong opinion in this particular case, since `getDiff()` also makes sense to me.
@snazy thoughts?

model/src/main/java/org/projectnessie/api/http/HttpDiffApi.java
nastra: done

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractResteasyTest.java
nastra: added that to `AbstractTestRest`

clients/client/src/main/java/org/projectnessie/client/api/GetDiffsBuilder.java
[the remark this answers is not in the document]
snazy: Let's rename those then to `fromRefName` + `toRefName`, so it's possible to add `toRef` + `fromRef` taking a `Reference` later.

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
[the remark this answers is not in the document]
snazy: "diff" is probably the right term here. It's a single "diff" that's being asked for - the one diff contains a set/list of differences.

model/src/main/java/org/projectnessie/model/DiffResponse.java
```diff
@@ -0,0 +1,46 @@
...
+@JsonSerialize(as = ImmutableDiffResponse.class)
+@JsonDeserialize(as = ImmutableDiffResponse.class)
+public interface DiffResponse {
+
+  List<Diff> diffs();
+
+  @Value.Immutable
+  @JsonSerialize(as = ImmutableDiff.class)
+  @JsonDeserialize(as = ImmutableDiff.class)
+  interface Diff {
```
snazy, line 37: Maybe rename `Diff` here to `Difference` or maybe better `DiffEntry`.

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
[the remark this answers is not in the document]
nastra: updated to `GetDiffBuilder` and `getDiff`

clients/client/src/main/java/org/projectnessie/client/api/GetDiffsBuilder.java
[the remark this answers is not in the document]
nastra: updated

snazy (approved):

PR #2808, opened 2021-12-01
Fix NPE for unchanged operation for fetching commit log with additional info

ajantha-bhat:
Operation will be null for unchanged operation, but the code is expecting non-null object, Hence the NPE.

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
@@ -1955,6 +1955,7 @@ public abstract class AbstractTestRest {
                                 ContentKey.of("key" + i),
                                 IcebergTable.of("meta" + i, i, i, i, i, "cid" + i)))
                         .operation(Delete.of(ContentKey.of("delete" + i)))
+                        .operation(Unchanged.of(ContentKey.of("key" + i)))
                         .commitMeta(CommitMeta.fromMessage("Commit #" + i))
                         .branchName(branch)
                         .hash(head)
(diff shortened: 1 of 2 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -2089,6 +2089,26 @@ public void testDiff() throws BaseNessieClientServerException {
             });
   }
 
+  @Test
+  public void commitLogExtendedForUnchangedOperation() throws Exception {
+    String branch = "main";
```
dimas-b, line 2094: Could you use a test case-specific branch? I believe most test cases create a dedicated branch up front. That should minimize interference between tests.

ajantha-bhat: ok. done.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -2089,6 +2089,31 @@ public void testDiff() throws BaseNessieClientServerException {
...
+        .hash(head)
+        .commit();
+
+    List<LogEntry> logEntries =
+        api.getCommitLog().fetchAdditionalInfo(true).refName(branch).get().getLogEntries();
+
+    assertThat(
+            logEntries.size() == 1
+                && logEntries.get(0).getCommitMeta().getMessage().contains("Commit #1"))
+        .isTrue();
```
dimas-b, line 2114: Sorry about a late nitpick... I did not notice this before, but would it be clearer to use two `assertThat(...).isEqualTo(...)` statements instead of one `isTrue()` with an `AND` boolean construct? In the former case JUnit will produce more specific and easier to read failure messages, AFAIK.

dimas-b: Could you also add asserts about `CommitLogEntry.getPuts()` and `getDeletes()`?

ajantha-bhat: It is ImmutableLogEntry  object, not CommitLogEntry. So, added a check that operations should be null.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -236,18 +236,21 @@ public LogResponse getCommitLog(String namedRef, CommitLogParams params)
...
-                              Content content = ((Put<Content>) op).getValue();
-                              logEntry.addOperations(Operation.Put.of(key, content));
-                            }
-                            if (op instanceof Delete) {
-                              logEntry.addOperations(Operation.Delete.of(key));
-                            }
-                          });
+                  List<org.projectnessie.versioned.Operation<Content>> operations =
+                      commit.getOperations();
+                  if (operations != null) {
```
nastra, line 241: nit: no need to create a new variable, just use `commit.getOperations()`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -2089,6 +2089,29 @@ public void testDiff() throws BaseNessieClientServerException {
             });
   }
 
+  @Test
+  public void commitLogExtendedForUnchangedOperation() throws Exception {
```
nastra, line 2093: maybe just include this into the `commitLogExtended` test?

ajantha-bhat: I cannot add extra validation in existing `commitLogExtended` testcase as operations will have put and delete.
I added one unchanged operation now to that testcase.

I specifically wanted to check with single delete operation to be sure that we don't refer that operation somewhere else which can cause NPE. Hence a separate test case

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
ajantha-bhat: done

nastra (approved):

PR #2810, opened 2021-12-01
Allow CEL-filtering on optional operations in get-commit-log

### LINKED ISSUE #2801: Allow CEL-filtering on optional operations in get-commit-log
Probably a rather simple-ish change in `TreeApiImpl#filterCommitLog` to add the operations to the map passed to the CEL expression.

PR should include tests on different operation types (put + delete), operation type checks, key checks, key-prefix checks and such.

Might be nice to have some convenience methods for the CEL scripts on `Key` + `Operation` + `Put`.

snazy:
Adds `operations` attribute for the CEL to filter the commit log to
allow filtering on e.g. the operation type (PUT/DELETE), the simple
table name, the namespace. Filtering on the operations requires
`fetchAdditionalInfo`.

Fixes #2801

### DIFF
--- a/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
+++ b/model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
@@ -53,12 +53,18 @@ public class CommitLogParams extends AbstractParams {
 
   @Parameter(
       description =
-          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n"
-              + "Usable variables within the expression are 'commit.author' (string) / 'commit.committer' (string) / 'commit.commitTime' (timestamp) / 'commit.hash' (string) / 'commit.message' (string) / 'commit.properties' (map)",
+          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n\n"
+              + "Usable variables within the expression are:\n\n"
+              + "- 'commit' with fields 'author' (string), 'committer' (string), 'commitTime' (timestamp), 'hash' (string), ',message' (string), 'properties' (map)\n\n"
+              + "- 'operations' (list), each operation has the fields 'type' (string, either 'PUT' or 'DELETE'), 'key' (string, namespace + table name), 'keyElements' (list of strings), 'namespace' (string), 'namespaceElements' (list of strings) and 'name' (string, the \"simple\" table name)\n\n"
+              + "Note that the expression can only test against 'operations', if 'fetchAdditionalInfo' is true.\n\n"
+              + "Hint: when filtering commits, you can determine whether commits are \"missing\" (filtered) by checking whether 'LogEntry.parentCommitHash' is different from the hash of the previous commit in the log response.",
       examples = {
         @ExampleObject(ref = "expr_by_commit_author"),
         @ExampleObject(ref = "expr_by_commit_committer"),
-        @ExampleObject(ref = "expr_by_commitTime")
+        @ExampleObject(ref = "expr_by_commitTime"),
+        @ExampleObject(ref = "expr_by_commit_operations_table_name"),
+        @ExampleObject(ref = "expr_by_commit_operations_type")
       })
   @QueryParam("query_expression")
   private String queryExpression;
--- a/model/src/main/java/org/projectnessie/model/Content.java
+++ b/model/src/main/java/org/projectnessie/model/Content.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.model;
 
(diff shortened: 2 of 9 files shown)


model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -53,12 +53,18 @@
 
   @Parameter(
       description =
-          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n"
-              + "Usable variables within the expression are 'commit.author' (string) / 'commit.committer' (string) / 'commit.commitTime' (timestamp) / 'commit.hash' (string) / 'commit.message' (string) / 'commit.properties' (map)",
+          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n\n"
+              + "Usable variables within the expression are:\n\n"
+              + "- 'commit' with fields 'author' (string), 'committer' (string), 'commitTime' (timestamp), 'hash' (string), ',message' (string), 'properties' (map)\n\n"
```
dimas-b, line 58: Could you be more specific about what the `timestamp` value is? Java system time (millis) or formatted string?

model/src/main/java/org/projectnessie/model/Content.java
```diff
@@ -59,6 +60,10 @@ public String getId() {
     return UUID.randomUUID().toString();
   }
 
+  @Value.Redacted
+  @JsonIgnore
+  public abstract Type getType();
```
dimas-b, line 65: Could you add a comment that we're exposing this for filtering purposes, not for serialization but that the type value is the same as what is serialized in JSON?

model/src/main/java/org/projectnessie/model/DeltaLakeTable.java
```diff
@@ -37,4 +37,9 @@
 
   @Nullable
   public abstract String getLastCheckpoint();
+
+  @Override
+  public Type getType() {
+    return Type.DELTA_LAKE_TABLE;
```
dimas-b, line 43: Is it possible to reuse this constant in string form for the `@JsonTypeName` annotation above?

model/src/main/resources/META-INF/openapi.yaml
```diff
@@ -90,6 +90,12 @@ components:
       value: "commit.committer=='nessie_committer'"
     expr_by_commitTime:
       value: "timestamp(commit.commitTime) > timestamp('2021-05-31T08:23:15Z')"
+    expr_by_commit_operations_in_namespace:
+      value: "operations.exists(op, op.key.startsWith('some.name.space.'))"
+    expr_by_commit_operations_table_name:
+      value: "operations.exists(op, op.name == 'BaseTable')"
```
dimas-b, line 96: `op.name` looks a bit odd to me, it feels like the operation name... how about `op.entityName`?

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
```diff
@@ -69,12 +78,122 @@
...
   public static final CommitMeta EMPTY_COMMIT_META = CommitMeta.fromMessage("");
   public static final ReferenceMetadata EMPTY_REFERENCE_METADATA =
       ImmutableReferenceMetadata.builder().commitMetaOfHEAD(EMPTY_COMMIT_META).build();
+
+  /**
+   * 'Mirrored' interface wrapping a {@link Operation} for CEL to have convenience fields for CEL
+   * and to avoid missing fields due to {@code @JsonIgnore}.
+   */
+  @SuppressWarnings("unused")
+  public interface OperationForCel {
```
dimas-b, line 96: Could this be a class instead? It looks like we only have one impl. and it would be nice to keep the runtime logic together with the getter method definitions, IMHO.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -380,8 +382,17 @@ private static CommitMeta addHashToCommitMeta(Hash hash, CommitMeta commitMeta)
     return logEntries.filter(
         logEntry -> {
           try {
+            List<Operation> operations = logEntry.getOperations();
+            if (operations == null) {
+              operations = Collections.emptyList();
+            }
+            // ContentKey has some @JsonIgnore attributes, which would otherwise not be accessible.
```
dimas-b, line 389: Do we not control which attributes are visible to CEL expressions? Could we make `@JsonIgnore` attributes visible?

nastra (approved):

dimas-b (approved):
I'll try and see what can be done about my comments in a follow-up PR

PR #2822, opened 2021-12-02
Python: Add -x flag to fetch additional metadata for branches/tags

nastra:


python/pynessie/commands/_branch_tag_handlers.py
```diff
@@ -99,16 +101,33 @@ def _handle_json_output(input_data: list, ref_name: str) -> str:
...
+            num_total_commits = x.metadata.num_total_commits if x.metadata.num_total_commits else "?"
+            common_ancestor_hash = x.metadata.common_ancestor_hash if x.metadata.common_ancestor_hash else "?"
+            head = (
+                "[author: {} | commitTime: {} | msg: {}]".format(
+                    x.metadata.commit_meta_of_head.author, x.metadata.commit_meta_of_head.commitTime, x.metadata.commit_meta_of_head.message
+                )
+                if x.metadata.commit_meta_of_head
+                else "?"
+            )
+            additional_info = " [commits ahead: {} | commits behind: {} | total commits: {} | common ancestor: {} | HEAD: {}]".format(
```
dimas-b, line 121: nit: perhaps use simply `total commits: {}, behind: {}, ahead: {}...` to make it terser in console output? Also putting `behind` first aligns with how GH renders those counts in the UI, I guess, where the "behind" count is on the left.

dimas-b: rename `total commits` -> `commits from root` for clarity?

python/pynessie/commands/_branch_tag_handlers.py
```diff
@@ -99,16 +101,33 @@ def _handle_json_output(input_data: list, ref_name: str) -> str:
...
     max_width = max((len(i.name) for i in input_data), default=0)
     for x in input_data:
+        additional_info = ""
+        if show_additional_info and x.metadata:
+            num_commits_ahead = x.metadata.num_commits_ahead if x.metadata.num_commits_ahead else "?"
+            num_commits_behind = x.metadata.num_commits_behind if x.metadata.num_commits_behind else "?"
+            num_total_commits = x.metadata.num_total_commits if x.metadata.num_total_commits else "?"
+            common_ancestor_hash = x.metadata.common_ancestor_hash if x.metadata.common_ancestor_hash else "?"
+            head = (
+                "[author: {} | commitTime: {} | msg: {}]".format(
```
dimas-b, line 115: Is `msg` that useful? I imagine it can easily spam STDOUT... maybe abbreviate it to N chars or the first new line?

dimas-b (dismissed):

python/pynessie/commands/_branch_tag_handlers.py
nastra: updated to reflect your suggestion. regarding total commits -> commits from root rename, I wonder if that is really clearer for users. To me `total commits` is pretty clear, but `commits from root` is rather confusing

python/pynessie/commands/_branch_tag_handlers.py
nastra: I wasn't sure about which parts of a commit would be useful here either, but I figured if you're asking for the additional metadata, then the msg might be useful, although I don't have any strong opinions here, so any suggestion is welcome

nastra: I added some logic to abbreviate the msg if it gets longer than 35 chars

python/pynessie/commands/_branch_tag_handlers.py
```diff
@@ -99,16 +101,38 @@ def _handle_json_output(input_data: list, ref_name: str) -> str:
...
+                else x.metadata.commit_meta_of_head.message
+            )
+            head = (
+                "[author: {} | commitTime: {} | msg: {}]".format(
+                    x.metadata.commit_meta_of_head.author, x.metadata.commit_meta_of_head.commitTime, commit_msg
+                )
+                if x.metadata.commit_meta_of_head
+                else "?"
+            )
+            additional_info = " [total commits: {} | commits behind: {} | commits ahead: {} | ancestor: {} | HEAD: {}]".format(
```
omarsmak, line 126: Can we use here `f"{}"` instead of `.format`? I also find `f"{}"` syntax cleaner 😄

omarsmak (approved):

python/pynessie/commands/_branch_tag_handlers.py
nastra: yep good idea, updated.

omarsmak (approved):
LGTM!

python/pynessie/commands/_branch_tag_handlers.py
dimas-b: maybe `commits on branch`? `total` sounds to me like the number of commits for the whole repo, but if you think it's ok, I'd be fine with `total` too.

dimas-b: actually, for me personally the "number of commits from the common ancestor" would be interesting, but it's not for this PR for sure :)

nastra: given that a user is looking at the metadata for a single branch/tag I think it should be self-explanatory that `total` means the total commits within this given branch/tag. Therefore I would go with this for now. We can rename it if we see that it's unclear to users what that really means.

dimas-b (approved):

PR #2848, opened 2021-12-06
Rename 'fetchAdditionalInfo' query param to 'fetch' for better extensibility

nastra:
`fetchAdditionalInfo` was quite limited in its functionality, as a user could only provide `true`/`false` flags, but never actually control the level of information to retrieve.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.client.api;
 
+import org.projectnessie.api.params.FetchOption;
 import org.projectnessie.model.ReferencesResponse;
 
 /**
@@ -31,8 +32,9 @@ public interface GetAllReferencesBuilder
    * ancestor in relation to the default branch, and the commit metadata for the HEAD commit.
    *
    * @return {@link GetAllReferencesBuilder}
+   * @param fetchOption The option indicating how much info to fetch
    */
-  GetAllReferencesBuilder fetchAdditionalInfo(boolean fetchAdditionalInfo);
+  GetAllReferencesBuilder fetch(FetchOption fetchOption);
 
   /**
    * Fetches all references and returns them in a {@link ReferencesResponse} instance.
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpTreeClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpTreeClient.java
@@ -20,6 +20,7 @@ import javax.validation.constraints.NotNull;
 import org.projectnessie.api.http.HttpTreeApi;
 import org.projectnessie.api.params.CommitLogParams;
 import org.projectnessie.api.params.EntriesParams;
+import org.projectnessie.api.params.FetchOption;
 import org.projectnessie.api.params.GetReferenceParams;
 import org.projectnessie.api.params.ReferencesParams;
 import org.projectnessie.error.NessieConflictException;
@@ -49,7 +50,7 @@ class HttpTreeClient implements HttpTreeApi {
         .queryParam(
             "maxRecords", params.maxRecords() != null ? params.maxRecords().toString() : null)
         .queryParam("pageToken", params.pageToken())
-        .queryParam("fetchAdditionalInfo", Boolean.toString(params.isFetchAdditionalInfo()))
+        .queryParam("fetch", FetchOption.getFetchOptionName(params.fetchOption()))
         .queryParam("filter", params.filter())
         .get()
         .readEntity(ReferencesResponse.class);
@@ -72,7 +73,7 @@ class HttpTreeClient implements HttpTreeApi {
     return client
         .newRequest()
         .path("trees/tree/{ref}")
-        .queryParam("fetchAdditionalInfo", Boolean.toString(params.isFetchAdditionalInfo()))
+        .queryParam("fetch", FetchOption.getFetchOptionName(params.fetchOption()))
         .resolveTemplate("ref", params.getRefName())
         .get()
         .readEntity(Reference.class);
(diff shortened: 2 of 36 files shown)


dimas-b (commented):
Overall, I'm not sure using one enum "scale" for the level of details fetched by multiple API methods is clearer than having one "fetch additional" boolean parameter.... but I'd be ok with this approach. Added some minor naming comments.

clients/client/src/main/java/org/projectnessie/client/http/HttpTreeClient.java
```diff
@@ -72,7 +77,11 @@ public Reference getReferenceByName(@NotNull GetReferenceParams params)
     return client
         .newRequest()
         .path("trees/tree/{ref}")
-        .queryParam("fetchAdditionalInfo", Boolean.toString(params.isFetchAdditionalInfo()))
+        .queryParam(
+            "view",
+            null == params.getViewOptions()
```
omarsmak, line 82: Would it be possible if we move this block into a function since being used several times here?

```
            null == params.getViewOptions()
                ? ViewOption.BRIEF.toString()
                : params.getViewOptions().toString()
```

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
```diff
@@ -31,8 +32,9 @@
    * ancestor in relation to the default branch, and the commit metadata for the HEAD commit.
    *
    * @return {@link GetAllReferencesBuilder}
+   * @param viewOption
    */
-  GetAllReferencesBuilder fetchAdditionalInfo(boolean fetchAdditionalInfo);
+  GetAllReferencesBuilder view(ViewOption viewOption);
```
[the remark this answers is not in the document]
nastra: `fetch(FetchOptions)` sgtm. will adjust accordingly

model/src/main/java/org/projectnessie/api/params/ViewOption.java
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.api.params;
+
+public enum ViewOption {
+  BRIEF,
```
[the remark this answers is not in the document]
nastra: makes sense

clients/client/src/main/java/org/projectnessie/client/http/HttpTreeClient.java
nastra: yep, updated

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -119,6 +116,14 @@ public String filter() {
     return filter;
   }
 
+  public FetchOption fetchOption() {
+    return fetchOption;
+  }
+
+  public boolean isFetchAll() {
```
snazy, line 123: Not sure about adding `isFetchAll()` to the `*Params` classes. It's duplicated in `GetReferenceParams` + `ReferencesParams`.

WDYT about making `fetchOption` non-null and annotate with `@DefaultValue("MINIMAL")` so it cannot become `null` and adding a method `isAll()` to `FetchOption`?

nastra: we could probably do that, but we would then have to make sure that `fetchOption` is never null and validate that in the `validate()` methods of the *Param classes and also remove the `empty()` method from those.

python/pynessie/client/_endpoints.py
```diff
@@ -95,7 +95,7 @@ def all_references(base_url: str, auth: Optional[AuthBase], ssl_verify: bool = T
     """
     params = {}
     if fetch_additional_info:
```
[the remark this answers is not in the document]
nastra: I figured we can leave the python code as is, since we currently only have a distinction between "fetch nothing additional" and "fetch everything additional"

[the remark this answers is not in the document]
snazy: Feels a bit odd to rename `fetch_additional_info` in the Java code but leave it in the Python code. Maybe just replace `fetch_additional_info` to `fetch_all` for now.

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
snazy: My concern is that it's a helper method for a particular attribute's value.
If there are more values in `FetchOption`, it would look weird to me to add more `isFetchFoo` methods.

nastra: I moved `isFetchAll` into the enum itself. I think that makes things easier and we don't have repetitive code.

python/pynessie/client/_endpoints.py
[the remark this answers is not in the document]
nastra: makes sense, updated to `fetch_all` everywhere in the python code

model/src/main/java/org/projectnessie/api/params/ReferencesParams.java
```diff
@@ -67,18 +68,22 @@
...
+    this(builder.maxRecords, builder.pageToken, builder.fetchOption, builder.filter);
+  }
+
+  public FetchOption fetchOption() {
+    return fetchOption;
   }
 
-  public boolean isFetchAdditionalInfo() {
-    return fetchAdditionalInfo;
+  public boolean isFetchAll() {
```
snazy, line 85: Still there

nastra: don't know how I forgot to remove that...thanks for spotting. removed it

model/src/main/java/org/projectnessie/api/params/CommitLogParams.java
```diff
@@ -119,6 +116,10 @@ public String filter() {
     return filter;
   }
 
+  public FetchOption fetchOption() {
```
snazy, line 119: getFetchOption() (and in the other two places)

model/src/main/java/org/projectnessie/api/params/FetchOption.java
```diff
@@ -0,0 +1,47 @@
...
+package org.projectnessie.api.params;
+
+import javax.annotation.Nullable;
+
+public enum FetchOption {
+  MINIMAL,
+  ALL;
+
+  /**
+   * Will return <code>true</code> if the given <code>fetchOption</code> is {@link FetchOption#ALL},
```
snazy, line 25: Nit: for future: the `{@code ...}` javadoc syntax is a bit easier to read than HTML `<code>`

snazy (approved):

PR #2861, opened 2021-12-08
Use next free port in FlapdoodleMongoTestConnectionProviderSource

### LINKED ISSUE #2860: More robust port detection in FlapdoodleMongoTestConnectionProviderSource
`FlapdoodleMongoTestConnectionProviderSource` currently determines the port that Mongo is running on by looking at the log output of the mongo daemon and checking for a particular regex.
When upgrading flapdoodle from `3.1.4` to `3.2.2` in #2840 the logging pattern changed and so the regex had to be updated.
It would be good to make that port detection code more robust by e.g. avoiding log parsing

nastra:
fixes #2860

This uses `ServerSocket` to retrieve the next available port in random
fashion and pass it to the `MongodConfig` rather than having to parse
the log output to figure out on which random port Mongo itself is
running.

Flapdoodle itself is doing a similar approach in https://github.com/flapdoodle-oss/de.flapdoodle.embed.process/blob/master/src/main/java/de/flapdoodle/embed/process/runtime/Network.java#L134-L155

### DIFF
--- a/versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
+++ b/versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
@@ -22,27 +22,18 @@ import de.flapdoodle.embed.mongo.MongodExecutable;
 import de.flapdoodle.embed.mongo.MongodStarter;
 import de.flapdoodle.embed.mongo.config.Defaults;
 import de.flapdoodle.embed.mongo.config.MongodConfig;
-import de.flapdoodle.embed.mongo.config.MongodProcessOutputConfig;
 import de.flapdoodle.embed.mongo.config.Net;
 import de.flapdoodle.embed.mongo.distribution.Version;
-import de.flapdoodle.embed.process.config.io.ProcessOutput;
-import de.flapdoodle.embed.process.io.StreamProcessor;
 import de.flapdoodle.embed.process.runtime.Network;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.io.IOException;
+import java.net.ServerSocket;
 
 /**
  * MongoDB test connection-provider source using a locally spawned MongoDB instance via
  * "flapdoodle".
  */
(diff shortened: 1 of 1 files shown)


versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
```diff
@@ -120,4 +80,10 @@ public void stop() throws Exception {
       }
     }
   }
+
+  private Integer findRandomOpenPortOnAllLocalInterfaces() throws IOException {
+    try (ServerSocket socket = new ServerSocket(0)) {
+      return socket.getLocalPort();
```
dimas-b, line 86: Technically this does not guarantee that Mongo will be able to bind to this port later. Is monitoring Mongo logs that much of an inconvenience?

nastra: monitoring the mongo log depends on the logging format :shrug:

dimas-b: true, but it's a solid test failure when it changes vs. a random failure with port allocation in odd environments... :shrug: 

@snazy : What is your take of this? I believe we talked briefly about this before :)

nastra: alternatively I was thinking whether there's a way to ask flapdoodle for the port that mongo is running on

dimas-b: May a PR to Flapdoodle? ;-)

snazy: There's no way to let mongod tell you in a machine readable way which port it has chosen.

But Unix domain sockets help here #2863

nastra:
closing in favor of #2863

PR #2863, opened 2021-12-08
Let Mongo tests via flapdoodle use Unix sockets

snazy:
Using anything relying on TCP ports is actually bad:
* Using the local port from a previous `new ServerSocket(0).getLocalPort()` is racy
  (by definition)
* Using a TCP port from another PRNG is bad
* Using a fixed port is bad (you can't run tests in parallel)
* Telling mongod to use "port 0" and parse the log output is also error prone

But it's possible to let Mongod listen to a Unix domain socket, which is a file,
and we can pretty much guarantee (at least way better than the above TCP port
approaches) that the FS path for the Unix domain socket is "ours".

The only downside is that tests require `com.github.jnr:jnr-unixsocket`, but it's
for tests only, so feels acceptable.

### DIFF
--- a/versioned/persist/mongodb/pom.xml
+++ b/versioned/persist/mongodb/pom.xml
@@ -60,5 +60,10 @@
       <artifactId>de.flapdoodle.embed.mongo</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-unixsocket</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/versioned/persist/mongodb/pom.xml
+++ b/versioned/persist/mongodb/pom.xml
@@ -60,10 +60,5 @@
       <artifactId>de.flapdoodle.embed.mongo</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>com.github.jnr</groupId>
-      <artifactId>jnr-unixsocket</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
 </project>
--- a/versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
+++ b/versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
@@ -23,10 +23,12 @@ import de.flapdoodle.embed.mongo.MongodStarter;
 import de.flapdoodle.embed.mongo.config.Defaults;
 import de.flapdoodle.embed.mongo.config.MongodConfig;
 import de.flapdoodle.embed.mongo.config.Net;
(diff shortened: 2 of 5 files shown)


dimas-b (approved):

versioned/persist/mongodb/pom.xml
```diff
@@ -60,5 +60,11 @@
       <artifactId>de.flapdoodle.embed.mongo</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-unixsocket</artifactId>
+      <version>0.18</version>
```
nastra, line 66: should this maybe go into the `dependencyManagement` section of the main `pom.xml`?

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
```diff
@@ -22,82 +22,49 @@
...
             .version(Version.Main.PRODUCTION)
             .net(new Net(0, Network.localhostIsIPv6()))
+            .putArgs("--bind_ip", unixSocket)
             .build();
 
     mongo = starter.prepare(mongodConfig);
     mongo.start();
 
-    assertThat(port).hasPositiveValue();
+    assertThat(Files.exists(Paths.get(unixSocket))).isTrue();
```
nastra, line 65: nit: can be replaced with `assertThat(Paths.get(unixSocket)).exists()`

snazy:
Argh - so this becomes more complicated than expected.

dimas-b (approved):

snazy:
Sadly, Unix domain sockets don't work (without letting a ton of other stuff leak into the native image).

So I'm updating this one to use a different approach - better than `new ServerSocket(0).getLocalPort()` but still not perfect. Leveraging the fact that most Linux distros allocate those ports in the range 32768...60999 - so it's using a "plain random port" in the range 1024...32767.

versioned/persist/mongodb/src/test/java/org/projectnessie/versioned/persist/mongodb/FlapdoodleMongoTestConnectionProviderSource.java
```diff
@@ -22,82 +22,80 @@
...
-    mongo.start();
-
-    assertThat(port).hasPositiveValue();
-
-    String connectionString = String.format("mongodb://localhost:%d", port.get());
+        MongodStarter.getInstance(Defaults.runtimeConfigFor(Command.MongoD).build());
+
+    long timeoutAt = System.currentTimeMillis() + 600_000;
+    int port;
+    while (true) {
```
dimas-b, line 54: I personally prefer the (current) log scraping approach... it seems cleaner to me and the failure is easy to understand (I hope).

snazy: Yea, I agree

PR #2950, opened 2021-12-25
Reflog support in Nessie

ajantha-bhat:
Reflog backend entity support to get the complete details of what happened at the backend (similar to git reflog).

Reflog support for both non-transactional db and transactional db.
- write support (protobuf serialized entries) for all operations (commit, create/delete/assign ref, merge, transplant).

Reflog entry contents can be seen here:
https://github.com/ajantha-bhat/nessie/blob/reflog/versioned/persist/serialize/src/main/proto/persist.proto#L81-L106

Added two tables for transactional store, 
One for storing the data and one for storing the reflog head.
https://github.com/ajantha-bhat/nessie/blob/reflog/versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/SqlStatements.java#L128-L164

- read support with new "reflogs" REST resource and swagger API doc with 
  - pagination support
  - startHash , endHash support 
  - Authz support

phase2 TODO: (In the follow up PR)
a. python support

Please find the benchmarks attached. Run locally on my machine.
JMH version: 1.34
VM version: JDK 11.0.11, Java HotSpot(TM) 64-Bit Server VM, 11.0.11+9-LTS-194
VM invoker: /Library/Java/JavaVirtualMachines/jdk-11.0.11.jdk/Contents/Home/bin/java
VM options:
Blackhole mode: full + dont-inline hint (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
Warmup: 2 iterations, 2000 ms each
Measurement: 3 iterations, 5000 ms each
Timeout: 10 min per iteration
Threads: 4 threads, will synchronize iterations
Benchmark mode: Throughput, ops/time
Benchmark: org.projectnessie.versioned.persist.benchmarks.CommitBench.singleBranchUnsharedKeys
Parameters: (adapter = RocksDB, tablesPerCommit = 5)

Screenshot 2022-01-10 at 8 46 33 PM
Screenshot 2022-01-10 at 8 46 43 PM


versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -676,6 +811,14 @@ protected abstract void writeGlobalCommit(
       NonTransactionalOperationContext ctx, GlobalStateLogEntry entry)
       throws ReferenceConflictException;
 
+  /**
+   * Write a new refLog-entry with a best-effort approach to prevent hash-collisions but without any
+   * other consistency checks/guarantees. Some implementations however can enforce strict
+   * consistency checks/guarantees.
+   */
+  protected abstract void writeRefLog(NonTransactionalOperationContext ctx, RefLogEntry entry)
+      throws ReferenceConflictException;
```
ajantha-bhat, line 820: Thought of having `RefLogConflictException` instead of this. But then all the backend store doesn't throw exception. As reflog writing is common for all the operations, all signatures need to be changed from backend store to REST API If I add new exception.  Hence didn't modified. Any thoughts ?

snazy (changes requested):
Overall looks okay so far.

However, I don't like that database-adapter internales are exposed (list of ref-iog-id+commit-hash pairs).
The commit message should not be stored in the ref-log at all. Having the commit-hash is good enough. If the user wants information about the commit referenced by the ref-log (e.g. `CommitMeta` or `Operations`), it can be optionally retrieved and as a pro be retrieved for all non-commit operations.

CEL filtering should go out. Ref-log is "dumb" for now - no need fo add unneeded functionality for now.


> authz support

You need a new rule, no? Only special users (think: superuser, admin, gc-service-accounts) shall be able to retrieve the ref-log. It's not a "normal user operation".


> update openapi.yml

Means what?


> transactional db support (how to get head of reflog? keeping another table will create a bottleneck for concurrent write? Also at a time only one concurrent op can be success as parent id need rebase?)

It should be an `UPDATE ref_log_pointer SET ref_log_head_id = :new_head WHERE ... AND ref_log_head_id = :previous_head` - needs to be initialized with the repo.


> txn cleanup

???



> python support

Later.


> marking row to fetch intermediate rows and parallelize them from spark job ?

???

versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
```diff
@@ -322,4 +327,45 @@ protected int entitySize(CommitLogEntry entry) {
   protected int entitySize(KeyWithType entry) {
     return toProto(entry).getSerializedSize();
   }
+
+  @Override
+  protected void writeRefLog(NonTransactionalOperationContext ctx, AdapterTypes.RefLogEntry entry)
+      throws ReferenceConflictException {
+    Lock lock = dbInstance.getLock().writeLock();
```
harshm-dev, line 334: I'm a little worried that we're acquiring the lock separately for the ref log entry. If there's a commit done, the commit and reflog operations will take locks independent of each other, if commit operation passes, whereas reflog insert fails, it's gonna leave an inconsistent state.

versioned/spi/src/main/java/org/projectnessie/versioned/RefLogEntry.java
```diff
@@ -0,0 +1,55 @@
...
+  Hash getRefLogId();
+
+  String getRefName();
+
+  String getRefType();
+
+  Hash getCommitHash();
+
+  /** Zero, one or more ancestors of this entry, nearest parent first. */
+  List<Ancestor> getAncestors();
```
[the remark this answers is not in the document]
ajantha-bhat: finally I need list of commit hashes for gc to avoid fetching each row. But I want last reflog id also to continue calling back the api again. So, storing both parent reflog and commit hash

versioned/spi/src/main/java/org/projectnessie/versioned/RefLogNotFoundException.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package org.projectnessie.versioned;
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+import javax.annotation.Nonnull;
+
+/** Exception thrown when a reflog is not present in the store. */
+public class RefLogNotFoundException extends VersionStoreException {
```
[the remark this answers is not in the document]
ajantha-bhat: But reflog is not mapped to reference. So, added this

versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
ajantha-bhat: multi table transaction is handled by CAS operation loop present in the caller NonTransactionalDatabaseAdapter#casOpLoop.  

which will takes care of reverting all the entries when any one of the commit failed.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
```diff
@@ -0,0 +1,75 @@
...
+    String getCommitHash();
+
+    List<Ancestor> getAncestors();
+
+    /** time in microseconds. */
+    long getOperationTime();
+
+    String getOperation();
+
+    String getDetails();
```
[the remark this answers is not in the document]
ajantha-bhat: I decided to keep as "Commit on reference $refName with hash $comitHash" in the details and remove commit message. So, if user need more details, they can query commit log api with the information present in the details.

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+    Hash endRef = null == params.pageToken() ? endHash : Hash.of(params.pageToken());
+
+    try (Stream<RefLogEntry> entries = getStore().getRefLog(endRef)) {
+      Stream<RefLogResponse.RefLogResponseEntry> logEntries =
+          entries.map(
+              entry -> {
+                ImmutableRefLogResponseEntry.Builder logEntry =
+                    RefLogResponse.RefLogResponseEntry.builder();
+                if (entry.getAncestors() != null) {
```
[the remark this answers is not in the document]
snazy: Still, the *list* of parents/ancestors is an implementation detail that must not be exposed.

versioned/spi/src/main/java/org/projectnessie/versioned/RefLogEntry.java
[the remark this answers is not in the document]
snazy: First, it's still an implementation detail - second, you have the commit hash in RefLogEntry in the RefLogResponse, no?

model/src/main/java/org/projectnessie/model/RefLogResponse.java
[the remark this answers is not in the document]
snazy: That information is available in the respsctive fields - no need to keep another string around.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+      return Spliterators.emptySpliterator();
+    }
+    RefLogEntry initial = fetchFromRefLog(ctx, initialHash);
+    if (initial == null) {
+      throw RefLogNotFoundException.forRefLogId(initialHash.asString());
+    }
+    return refLogFetcher(ctx, initial, this::fetchPageFromRefLog, RefLogEntry::getAncestors);
+  }
+
+  protected <T> Spliterator<T> refLogFetcher(
```
snazy, line 1370: This one should be unified with the existing method - and not copied.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
```diff
@@ -0,0 +1,75 @@
...
+    String getRefType();
+
+    String getCommitHash();
+
+    List<Ancestor> getAncestors();
+
+    /** time in microseconds. */
+    long getOperationTime();
+
+    String getOperation();
```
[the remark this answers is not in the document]
ajantha-bhat: this is the end result. So string. 
Storing as enum to save space.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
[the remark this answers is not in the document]
ajantha-bhat: Isn't it better for UI if details has the string? instead of they forming string only for commit operation? Rest of operations also I am providing a nice string. It is duplicate storage for some operations.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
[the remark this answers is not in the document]
ajantha-bhat: similar to how we show hashes as string.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+      OP_CONTEXT ctx,
+      T initial,
```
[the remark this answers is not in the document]
ajantha-bhat: reusing the logfetcher now as I changed Ancestor to just Hash.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+  protected Spliterator<RefLogEntry> readRefLog(OP_CONTEXT ctx, Hash initialHash)
+      throws RefLogNotFoundException {
+    if (NO_ANCESTOR.equals(initialHash)) {
+      return Spliterators.emptySpliterator();
+    }
+    RefLogEntry initial = fetchFromRefLog(ctx, initialHash);
+    if (initial == null) {
+      throw RefLogNotFoundException.forRefLogId(initialHash.asString());
+    }
+    return refLogFetcher(ctx, initial, this::fetchPageFromRefLog, RefLogEntry::getAncestors);
```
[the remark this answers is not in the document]
ajantha-bhat: reusing the logfetcher now as I changed Ancestor to just Hash.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+  }
+
+  protected <T> Spliterator<T> refLogFetcher(
+      OP_CONTEXT ctx,
+      T initial,
+      BiFunction<OP_CONTEXT, List<RefLogEntry.Ancestor>, List<T>> fetcher,
+      Function<T, List<RefLogEntry.Ancestor>> nextPage) {
+    return new AbstractSpliterator<T>(Long.MAX_VALUE, 0) {
+      private Iterator<T> currentBatch;
+      private boolean eof;
```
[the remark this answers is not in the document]
ajantha-bhat: reusing the logfetcher now as I changed Ancestor to just Hash.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+      private T previous;
+
+      @Override
+      public boolean tryAdvance(Consumer<? super T> consumer) {
+        if (eof) {
+          return false;
+        } else if (currentBatch == null) {
+          currentBatch = Collections.singletonList(initial).iterator();
+        } else if (!currentBatch.hasNext()) {
+          if (previous == null) {
```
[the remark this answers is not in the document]
ajantha-bhat: reusing the logfetcher now as I changed Ancestor to just Hash.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,74 @@ protected void checkExpectedGlobalStates(
...
+            eof = true;
+            return false;
+          }
+          List<RefLogEntry.Ancestor> page = nextPage.apply(previous);
+          previous = null;
+          if (!page.isEmpty()) {
+            currentBatch = fetcher.apply(ctx, page).iterator();
```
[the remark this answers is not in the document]
ajantha-bhat: reusing the logfetcher now as I changed Ancestor to just Hash.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: yes. done.

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
```diff
@@ -0,0 +1,169 @@
...
+ * <p>For easier usage of this class, there is {@link RefLogParams#builder()}, which allows
+ * configuring/setting the different parameters.
+ */
+public class RefLogParams extends AbstractParams {
+
+  @Nullable
+  @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
+  @Parameter(
+      description =
+          "Hash on the given ref to start from (in chronological sense), the 'far' end of the reflog, returned "
```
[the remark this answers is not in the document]
ajantha-bhat: added

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
```diff
@@ -0,0 +1,160 @@
...
+              .build();
+    } catch (ScriptException e) {
+      throw new IllegalArgumentException(e);
+    }
+    return refLogEntries.filter(
+        refLogEntry -> {
+          try {
+            return script.execute(Boolean.class, ImmutableMap.of(VAR_REFLOG_ENTRY, refLogEntry));
+          } catch (ScriptException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
ajantha-bhat: removed CEL filter feature as it was concluded that no need.

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestRefLogResource.java
```diff
@@ -0,0 +1,73 @@
...
+      ServerConfig config,
+      VersionStore<Content, CommitMeta, Type> store,
+      AccessChecker accessChecker) {
+    this.config = config;
+    this.store = store;
+    this.accessChecker = accessChecker;
+  }
+
+  private RefLogApi resource() {
+    // TODO: need RefLogApiImplWithAuthorization.
```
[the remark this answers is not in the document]
ajantha-bhat: added RefLogApiImplWithAuthorization

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
```diff
@@ -0,0 +1,169 @@
...
+
+  @Parameter(
+      description =
+          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n\n"
+              + "Usable variables within the expression are:\n\n"
+              + "- 'reflog' with fields 'refName' (string), 'refType' (string), 'commitHash' (string), "
+              + "'operationTime' (long), "
+              + "'operation' (string), 'details' (string)\n\n")
+  @QueryParam("filter")
+  private String filter;
```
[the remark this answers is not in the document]
ajantha-bhat: removed.

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
```diff
@@ -0,0 +1,160 @@
...
+                  logEntries.spliterator(),
+                  x -> Objects.equals(x.getRefLogId(), params.startHash())),
+              false);
+
+      List<RefLogResponse.RefLogResponseEntry> items =
+          filterRefLog(logEntries, params.filter()).limit(max + 1).collect(Collectors.toList());
+
+      if (items.size() == max + 1) {
+        return ImmutableRefLogResponse.builder()
+            .addAllLogEntries(items.subList(0, max))
```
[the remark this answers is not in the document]
ajantha-bhat: sublist creates a view from original list, doesn't allocate new list.
And the allLogEntries is called only when the limit is reached. It is not called else.

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
```diff
@@ -0,0 +1,160 @@
...
+            .isHasMore(true)
+            .token(items.get(max).getRefLogId())
+            .build();
+      }
+      return ImmutableRefLogResponse.builder().addAllLogEntries(items).build();
+    } catch (RefLogNotFoundException e) {
```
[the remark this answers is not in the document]
ajantha-bhat: good point. need to discuss separately. I think reflog cannot support old store. So, I will create an issue to track it.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -402,9 +512,23 @@ public void initializeRepo(String defaultBranchName) {
...
+                ctx,
+                defaultBranchName,
+                RefLogEntry.RefType.Branch,
+                NO_ANCESTOR,
+                NO_ANCESTOR,
+                RefLogEntry.Operation.CREATE_REFERENCE,
+                commitTimeInMicros(),
+                ByteString.copyFromUtf8(
+                    String.format(
+                        "Create Default branch %s with %s", defaultBranchName, NO_ANCESTOR)));
```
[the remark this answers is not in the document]
ajantha-bhat: I want to make it related to CREATE_REFERENCE, hence this description.

ajantha-bhat:
@snazy , @harshm-dev : I have handled the comments and also finished TODO which was applicable for non-txn db.
please have a look at the PR again. Thanks.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
```diff
@@ -0,0 +1,75 @@
...
+import javax.validation.constraints.NotNull;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Schema(type = SchemaType.OBJECT, title = "RefLogResponse")
+@JsonSerialize(as = ImmutableRefLogResponse.class)
+@JsonDeserialize(as = ImmutableRefLogResponse.class)
+public interface RefLogResponse extends PaginatedResponse {
```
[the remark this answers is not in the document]
ajantha-bhat: added.

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
```diff
@@ -0,0 +1,106 @@
...
+      ServerConfig config,
+      VersionStore<Content, CommitMeta, Type> store,
+      AccessChecker accessChecker,
+      Principal principal) {
+    super(config, store, accessChecker, principal);
+  }
+
+  @Override
+  public RefLogResponse getRefLog(RefLogParams params) throws NessieNotFoundException {
+    int max =
```
snazy, line 55: The code here looks very much like the code to retrieve the commit log.
Can it be unified and the common parts implemented in BaseApiImpl?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -853,4 +997,72 @@ protected abstract GlobalStateLogEntry fetchFromGlobalLog(
...
+            details,
+            currentRefLogId,
+            parentEntry);
+
+    writeRefLog(ctx, refLogEntry);
+
+    return currentRefLogId;
+  }
+
+  private RefLogEntry buildRefLogEntry(
```
snazy, line 1031: This should live in `ABstractDatabaseAdapter` like `buildIndividualCommit` does

model/src/main/java/org/projectnessie/model/RefLogResponse.java
```diff
@@ -0,0 +1,76 @@
...
+    @NotNull
+    long getOperationTime();
+
+    /** Operation String mapped to ENUM in {@code RefLogEntry.Operation} of 'persist.proto' file. */
+    @NotNull
+    String getOperation();
+
+    /** Detailed message to explain the operation. */
+    @NotNull
+    String getDetails();
```
snazy, line 74: None of the implementations add any value to this property - i.e. all information in "details"  is available via the other fiels. Really no need to keep this property here or even persisted. It would be okay to add the merged/transplanted hashes as a separate, optional attribute. But I'm -1 on returning or persisting a human readable message that doesn't really add value.

versioned/persist/serialize/src/main/proto/persist.proto
```diff
@@ -77,10 +77,34 @@ message GlobalStateLogEntry {
...
+  enum Operation {
+    CREATE_REFERENCE = 0;
+    COMMIT = 1;
+    DELETE_REFERENCE = 2;
+    ASSIGN_REFERENCE = 3;
+    MERGE = 4;
+    TRANSPLANT = 5;
+  }
+  Operation operation = 7;
+  bytes details = 8;
```
snazy, line 100: None of the implementations add any value to this property - i.e. all information in "details"  is available via the other fiels. Really no need to keep this property. It would be okay to add the merged/transplanted hashes as a separate, optional attribute. But I'm -1 on persisting a human readable message that doesn't really add value.

versioned/spi/src/main/java/org/projectnessie/versioned/RefLogNotFoundException.java
snazy, line 24: Any test case that verifies that this excepiton is properly thrown?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -853,4 +997,72 @@ protected abstract GlobalStateLogEntry fetchFromGlobalLog(
 
   protected abstract List<GlobalStateLogEntry> fetchPageFromGlobalLog(
       NonTransactionalOperationContext ctx, List<Hash> hashes);
+
+  protected Hash writeRefLogEntry(
```
snazy, line 1001: This should live in `ABstractDatabaseAdapter`

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -208,8 +210,21 @@ public Hash merge(
                 writeGlobalCommit(
                     ctx, timeInMicros, Hash.of(pointer.getGlobalId()), Collections.emptyList());
 
+            Hash newRefLogId =
+                writeRefLogEntry(
```
snazy, line 214: This and all other call sites of `writeRefLogEntry` look tike they will be duplicated in `TxDatabaseAdapter` - better to not have duplicated code.

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -117,7 +120,7 @@ public void eraseRepo() {
     client.client.deleteItem(
         b -> b.tableName(TABLE_GLOBAL_POINTER).key(globalPointerKeyMap).build());
 
-    Stream.of(TABLE_GLOBAL_LOG, TABLE_GLOBAL_LOG, TABLE_KEY_LISTS)
+    Stream.of(TABLE_GLOBAL_LOG, TABLE_COMMIT_LOG, TABLE_KEY_LISTS, TABLE_REF_LOG)
```
snazy, line 123: Nice catch!

versioned/persist/serialize/src/main/proto/persist.proto
```diff
@@ -77,10 +77,34 @@ message GlobalStateLogEntry {
   repeated ContentIdWithBytes puts = 4;
 }
 
+message RefLogEntry {
```
snazy, line 80: Rename this one to avoid fully-qualified class names - clashes with `org.projectnessie.versioned.RefLogEntry` and  `org.projectnessie.model.RefLogResponse.RefLogEntry`

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -152,7 +154,6 @@ public Hash commit(
     }
 
     commitAttempt.commitMetaSerialized(serializeMetadata(metadata));
-
```
snazy, line 155: Let's not add unrelated changes - the PR is already quite big

versioned/spi/src/main/java/org/projectnessie/versioned/RefLogEntry.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.versioned;
+
+import java.util.List;
+import org.immutables.value.Value;
+
+/** Represents a reflog-entry stored in the database. */
+@Value.Immutable
+public interface RefLogEntry {
```
snazy, line 23: Rename this one to avoid fully-qualified class names - clashes with `RefLogEntry` in `persist.proto` and  `org.projectnessie.model.RefLogResponse.RefLogEntry`

snazy (changes requested):
Still -1 on having the "details" attribute. HTML/markdown/etc UI formatting details should definitely not leak into the REST API or even the storage.

The type name `RefLogEntry` is used for three different types - that's very confusing & requires fully-qualified type names in the code, so this should be changed. 

Rest looks mostly good. Couple of very likely common code parts should be moved to the actual base classes.

snazy (commented):
Another note: I'd really like to only merge this when the TxDatabaseAdapter PR is ready. Not great to have something in m̀ain` that doesn't work everywhere.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
ajantha-bhat: I intentionally didn't added there to avoid having `ReflogEntry` to `adaptertypes.ReflogEntry` conversion. If I keep it inside like `writeGlobalCommit`, I can avoid that extra conversion as I can build `adaptertypes.ReflogEntry `directly.

So, I am basically following like `writeGlobalCommit`

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
ajantha-bhat: As mentioned above,
I intentionally didn't added there to avoid having ReflogEntry to adaptertypes.ReflogEntry conversion. If I keep it inside like writeGlobalCommit, I can avoid that extra conversion as I can build adaptertypes.ReflogEntry directly.

So, I am basically following like writeGlobalCommit

versioned/persist/serialize/src/main/proto/persist.proto
ajantha-bhat: changed now.
proto is having RefLogEntry
RefLogResponse is having RefLogResponseEntry
spi is having RefLog

servers/quarkus-server/src/main/java/org/projectnessie/server/authz/CelAccessChecker.java
```diff
@@ -123,6 +124,22 @@ public void canDeleteEntity(AccessContext context, NamedRef ref, ContentKey key,
     canPerformOpOnPath(context, ref, key, AuthorizationRuleType.DELETE_ENTITY);
   }
 
+  @Override
+  public void canViewRefLog(AccessContext context) throws AccessControlException {
```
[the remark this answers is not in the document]
ajantha-bhat: canPerformOpOnReference works on ref, this is independent of ref. Hence added new. 
Refactoring current method again need check specific to reflog, hence added new.

servers/services/src/main/java/org/projectnessie/services/impl/RefLogApiImpl.java
ajantha-bhat: The building different objects here also we don't need filter and get commit log needs reference. hence cannot combine. Not much common code left. They look similar but for different purpose.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
ajantha-bhat: yeah. If I move this to abstract class, then I need to use `RefLog` instead of proto `ReflogEntry`.
So, it brings an extra conversion for each operation. 
Considering code duplication vs perf. I chose performance. 

Actually I tired adding proto dependency to abstract class. but its a circular dependency. I think some part is not designed common for txn and non-txn.

model/src/main/java/org/projectnessie/error/NessieRefLogNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.error;
+
+/** This exception is thrown when a requested reflog is not present in the store. */
+public class NessieRefLogNotFoundException extends NessieNotFoundException {
```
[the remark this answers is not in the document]
ajantha-bhat: added the testcase

ajantha-bhat:
@snazy 

> Still -1 on having the "details" attribute. HTML/markdown/etc UI formatting details should definitely not leak into the REST API or even the storage.

I have removed the details and added 'sourceHashes'

> The type name RefLogEntry is used for three different types - that's very confusing & requires fully-qualified type names in the code, so this should be changed.

done. 

> Rest looks mostly good. Couple of very likely common code parts should be moved to the actual base classes.

I have replied in line why I did like that. please have a look

model/src/main/java/org/projectnessie/model/RefLogResponse.java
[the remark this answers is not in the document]
snazy: No, that's actually not worth it. Every UI displays/formats it differently - also l10n i18n

servers/quarkus-server/src/test/java/org/projectnessie/server/authz/NessieAuthorizationTestProfile.java
```diff
@@ -71,6 +71,9 @@
           .put(
               "nessie.server.authorization.rules.allow_commits_without_entity_changes",
               "op=='COMMIT_CHANGE_AGAINST_REFERENCE' && role=='test_user2' && ref.startsWith('allowedBranch')")
+          .put(
```
snazy, line 74: Needs to be documented on application.properties, too

And probably the nessie-site as well

snazy (commented):
Looks much better now!

Can you run sime experiments using the benchmarking tool in versioned-persist-tests - comparing the differences for in-memory, rocks and mongo. Would be good to see the perf impact of this change. If necessary, it's fine to provision an EC2 instance for these tests, if your dev machine doesn't have enough resources.

On top: this change needs a follow-up to update the nessie web site.

model/src/main/java/org/projectnessie/model/RefLogResponse.java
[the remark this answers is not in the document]
ajantha-bhat: @snazy: I think you replied to old message. I don't have details field now.

servers/quarkus-server/src/test/java/org/projectnessie/server/authz/NessieAuthorizationTestProfile.java
ajantha-bhat: updated the site docs, application.properties and commit kernel doc also.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractResteasyTest.java
```diff
@@ -465,4 +466,32 @@ public void testGetDiff() {
     assertThat(diff.getFrom()).isEqualTo(fromTable);
     assertThat(diff.getTo()).isEqualTo(toTable);
   }
+
+  @Test
+  public void testGetRefLog() {
```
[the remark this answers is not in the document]
ajantha-bhat: Here this testcase mainly focus on testing rest request where as AbstractTestRest is API based. I didn't kept scenarios here as the main intention is to test rest request with query params.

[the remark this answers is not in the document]
ajantha-bhat: we have like this for comitt log, diff api etc. So I thought it is required to have rest easy testcase also.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -2318,4 +2321,148 @@ protected void unwrap(Executable exec) throws Throwable {
...
+        .branch(createdBranch3)
+        .transplant();
+    // reflog 11: delete branch
+    api.deleteBranch().branchName(branch1).hash(branch0.getHash()).delete();
+    // reflog 12: delete tag
+    api.deleteTag().tagName(tagName).hash(branch0.getHash()).delete();
+
+    RefLogResponse refLogResponse = api.getRefLog().get();
+    // verify reflog entries
+    assertThat(refLogResponse.getLogEntries().get(0).getOperation()).isEqualTo("DELETE_REFERENCE");
```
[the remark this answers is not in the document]
ajantha-bhat: you mean by building an `expectedResponse` and compare `assertThat(refLogResponse).isEqualTo(expectedResponse) `? 

It has eight fields, the commit time will be set inside of the code. So, I am not sure how to set that in my expectedResponse. Hence I followed this approach.

versioned/persist/serialize/src/main/proto/persist.proto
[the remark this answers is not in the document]
ajantha-bhat: done

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -243,7 +248,22 @@ public Hash merge(
                     h -> {},
                     h -> {},
                     updateCommitMetadata);
-            return tryMoveNamedReference(conn, toBranch, currentHead, toHead);
+            Hash resultHash = tryMoveNamedReference(conn, toBranch, currentHead, toHead);
+
+            Hash parentId = getRefLogHead(conn);
```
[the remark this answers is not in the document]
ajantha-bhat: done

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -853,4 +967,71 @@ protected abstract GlobalStateLogEntry fetchFromGlobalLog(
 
   protected abstract List<GlobalStateLogEntry> fetchPageFromGlobalLog(
       NonTransactionalOperationContext ctx, List<Hash> hashes);
+
+  protected Hash writeRefLogEntry(
+      NonTransactionalOperationContext ctx,
+      String refName,
```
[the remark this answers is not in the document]
ajantha-bhat: done

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1304,4 +1422,167 @@ protected boolean metadataUpperCase() {
...
+      Connection connection,
+      NamedRef ref,
+      Hash parentRefLogId,
+      Hash commitHash,
+      AdapterTypes.RefLogEntry.Operation operation,
+      long timeInMicros,
+      List<Hash> sourceHashes)
+      throws ReferenceConflictException {
+
+    Hash currentRefLogId = randomHash();
```
snazy, line 1436: Nit: Looks like we don't need to generate the `currentRefLogId` but can be moved into `buildRefLogEntry`

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -254,6 +269,20 @@ public Hash merge(
     }
   }
 
+  private void commitRefLog(
```
snazy, line 272: Nit: Move this method to the other new private methods for the ref-log below (try to keep pulic/protected implementation at the top - and private below)

versioned/spi/src/main/java/org/projectnessie/versioned/RefLog.java
```diff
@@ -0,0 +1,51 @@
...
+  String getRefName();
+
+  /** Reference type can be 'Branch' or 'Tag'. */
+  String getRefType();
+
+  /** Output commit hash of the operation. */
+  Hash getCommitHash();
+
+  /** Parent reflog id of the current entry. */
+  List<Hash> getParents();
```
snazy, line 38: Refactor to `Hash getParent()` (that list is internal to the database-adapter - nothing for the version-store)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1334,4 +1336,33 @@ protected void checkExpectedGlobalStates(
...
+   * exactly as many elements as in the parameter {@code hashes}. Non-existing hashes are returned
+   * as {@code null}.
+   */
+  protected abstract List<RefLog> fetchPageFromRefLog(OP_CONTEXT ctx, List<Hash> hashes);
+
+  /** Reads from the refLog starting at the given refLog-hash. */
+  protected Stream<RefLog> readRefLogStream(OP_CONTEXT ctx, Hash initialHash)
+      throws RefLogNotFoundException {
+    Spliterator<RefLog> split = readRefLog(ctx, initialHash);
+    return StreamSupport.stream(split, false);
```
snazy, line 1354: This requires an  onClose` on the stream to delegate the close to the underlying spliterator

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -591,6 +688,21 @@ protected int entitySize(KeyWithType entry) {
     return toProto(entry).getSerializedSize();
   }
 
+  @Override
+  public Stream<RefLog> refLog(Hash offset) throws RefLogNotFoundException {
```
snazy, line 692: This method should be in the "block" with the api implementations above

snazy (changes requested):
Pretty close.
Just need to get the list of parents out of the VS-SPI type, a missing onClose-delegation and a few nits.

ajantha-bhat:
@snazy : Please find the benchmarks attached. Run locally on my machine.
# JMH version: 1.34
# VM version: JDK 11.0.11, Java HotSpot(TM) 64-Bit Server VM, 11.0.11+9-LTS-194
# VM invoker: /Library/Java/JavaVirtualMachines/jdk-11.0.11.jdk/Contents/Home/bin/java
# VM options: <none>
# Blackhole mode: full + dont-inline hint (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 2 iterations, 2000 ms each
# Measurement: 3 iterations, 5000 ms each
# Timeout: 10 min per iteration
# Threads: 4 threads, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.projectnessie.versioned.persist.benchmarks.CommitBench.singleBranchUnsharedKeys
# Parameters: (adapter = RocksDB, tablesPerCommit = 5)

Screenshot 2022-01-10 at 8 46 33 PM
Screenshot 2022-01-10 at 8 46 43 PM

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
ajantha-bhat: I wanted to save one `Hash.of` conversion

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: I have already handled for Txn store, non-txn it is not present similar to `readCommitLogStream` because there is no releaseConnection impl.

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
ajantha-bhat: done

versioned/spi/src/main/java/org/projectnessie/versioned/RefLog.java
ajantha-bhat: done, Added one more interface. I was reusing it to avoid extra object.

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
ajantha-bhat: done

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
ajantha-bhat: moved.

ajantha-bhat:
> Pretty close.
Just need to get the list of parents out of the VS-SPI type, a missing onClose-delegation and a few nits.

Done. Please have a look. Thanks.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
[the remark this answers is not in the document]
snazy: `.extracting()` or `.matches*()`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Ah, right.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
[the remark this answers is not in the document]
snazy: But happy to get it fixed it in a follow-up

snazy (approved):
LGTM! Good job!

The perf regression is expected due to the nature of this change, but those don't look too bad to me.

ajantha-bhat:
@snazy : Thanks for the approval. I will refactor the test case in followup PR. I will merge this.

PR #2965, opened 2021-12-30
Add `/content` to single-page-app-routing in Quarkus

snazy:


servers/quarkus-server/src/main/java/org/projectnessie/server/filters/SinglePageAppRoutingFilter.java
```diff
@@ -23,7 +23,7 @@
 
   // maintain list of base paths that should serve root html.
   // must be updated as more UI paths are introduced
-  private static final String[] UI_ROUTES = {"/tree"};
+  private static final String[] UI_ROUTES = {"/tree", "/content", "/notfound"};
```
ajantha-bhat, line 26: we should add `Config` resource and `Diff` resource also?

snazy: Why? Where's "diff" in the UI?

ajantha-bhat: yes, we don't have `Diff` feature in UI. It is only supported from REST API.

ajantha-bhat (approved):

PR #2989, opened 2022-01-04
Simplify and fix GH workflow cache usage for pip + maven

snazy:
The change in `.github/actions/dev-tool-python/action.yml`, used by
`pull-request.yml`, now includes all requirements*.txt files, so a change
in a PR to for example `rquirements_dev.txt` no longer results in not
reusing an older cache version, that does not have an updated dependency.

See https://github.com/actions/setup-python#caching-packages-dependencies

### DIFF
--- a/.github/actions/dev-tool-python/action.yml
+++ b/.github/actions/dev-tool-python/action.yml
@@ -10,12 +10,8 @@ runs:
       uses: actions/setup-python@v2
       with:
         python-version: ${{ inputs.python-version }}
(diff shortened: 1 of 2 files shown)


snazy:
Note: this should help get #2974 in.

omarsmak (approved):

snazy:
> Note: this should help get #2974 in.

Meh - it doesn't. So this PR is rather a cosmetic change.

rymurr (approved):

PR #3000, opened 2022-01-05
WIP: early implementation of storing iceberg metadata on server side

rymurr:
This WIP adds storing metadata to the server. I am concerned at the moment with the direction of the patch rather than specifics around code style. Its a bit messy.

Please have a look especially at how things are being stored and let me kno what you think!

### DIFF
--- a/model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
+++ b/model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model;
+
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.Map;
+import javax.annotation.Nullable;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.immutables.value.Value;
+
+/**
+ * Represents the state of an Iceberg table in Nessie. An Iceberg table is globally identified via
+ * its {@link Content#getId() unique ID}.
+ *
+ * <p>The Iceberg-table-state consists of the location to the table-metadata and the state of
+ * relevant IDs using a serialized version of those.
+ *
+ * <p>When adding a new table (aka content-object identified by a content-id), use a {@link
+ * org.projectnessie.model.Operation.Put} without an expected-value. In all other cases (updating an
+ * existing table). always pass the last known version of {@link IcebergTable} as the expected-value
+ * within the put-operation.
+ */
+@Schema(
+    type = SchemaType.OBJECT,
+    title = "Iceberg table global state",
+    description =
+        "Represents the global state of an Iceberg table in Nessie. An Iceberg table is globally "
+            + "identified via its unique 'Content.id'.\n"
+            + "\n"
+            + "A Nessie commit-operation, performed via 'TreeApi.commitMultipleOperations', for Iceberg "
+            + "for Iceberg consists of a 'Operation.Put' with an 'IcebergTable' as in the 'content' "
+            + "field and the previous value of 'IcebergTable' in the 'expectedContent' field.\n"
+            + "\n"
+            + "During a commit-operation, Nessie checks whether the known global state of the "
+            + "Iceberg table is compatible (think: equal) to 'Operation.Put.expectedContent'.")
+@Value.Immutable
+@JsonSerialize(as = ImmutableIcebergTableMetadata.class)
+@JsonDeserialize(as = ImmutableIcebergTableMetadata.class)
+@JsonTypeName("ICEBERG_METADATA")
+public abstract class IcebergTableMetadata extends Content {
+
+  /**
+   * Location where Iceberg stored its {@code TableMetadata} file. The location depends on the
+   * (implementation of) Iceberg's {@code FileIO} configured for the particular Iceberg table.
+   */
+  @Nullable
+  public abstract String getMetadataLocation();
+
+  public abstract int getFormatVersion();
+
+  public abstract String getUuid();
+
+  public abstract String getLocation();
+
+  public abstract long getLastSequenceNumber();
+
+  public abstract long getLastUpdatedMillis();
+
+  public abstract int getLastColumnId();
+
+  public abstract String getSchema();
+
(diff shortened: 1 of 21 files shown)


servers/store/src/main/proto/table.proto
```diff
@@ -39,6 +44,70 @@ message IcebergRefState {
...
+  int32 last_column_id = 2;
+  int64 last_sequence_number = 3;
+  int64 last_updated_millis = 4;
+  string uuid = 5;
+  string location = 6;
+  int32 last_assigned_partition_id = 7;
+  repeated IcebergProperties properties = 8;
+}
+
+message IcebergProperties {
```
snazy, line 94: IcebertProperty - it's just one

model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public abstract SortOrder getSortOrder();
+
+  public abstract Snapshot getSnapshot();
+
+  public abstract Map<String, String> getProperties();
+
+  public abstract int getLastAssignedPartitionId();
+
+  public abstract long getSnapshotId();
```
snazy, line 89: Not sure, but I suspect we don't need the snapshot-id, schema-id, spec-id, sort-order-id here - should be in the respective objects.

model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public abstract int getLastColumnId();
+
+  public abstract String getSchema();
+
+  public abstract PartitionSpec getSpec();
+
+  public abstract SortOrder getSortOrder();
+
+  public abstract Snapshot getSnapshot();
```
snazy, line 83: Should be nullable - a new table hasn't necessarily a snapshot (same maybe for the partition spec + sort order ... schema?)

model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+
+  public abstract Map<String, String> getProperties();
+
+  public abstract int getLastAssignedPartitionId();
```
snazy, line 87: Shoud we make this ID (and last-seq-num, last-column-id) nullable instead of using Iceberg's "not present" sentinel? I'm fine with either though.

model/src/main/java/org/projectnessie/model/PartitionField.java
```diff
@@ -0,0 +1,36 @@
...
+@JsonTypeName("ICEBERG_PARTITION_FIELD")
+public interface PartitionField {
+
+  int getSourceId();
+
+  int getFieldId();
+
+  String getName();
+
+  String getTransform();
```
snazy, line 35: I think, this is an interface in Iceberg - not sure how that can be mapped to a string

model/src/main/java/org/projectnessie/model/SortField.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableSortField.class)
+@JsonDeserialize(as = ImmutableSortField.class)
+@JsonTypeName("ICEBERG_SORT_FIELD")
+public interface SortField {
+  String getTransform();
```
snazy, line 28: I think, this is an interface in Iceberg - not sure how that can be mapped to a string

model/src/main/java/org/projectnessie/model/SortOrder.java
```diff
@@ -0,0 +1,34 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableSortOrder.class)
+@JsonDeserialize(as = ImmutableSortOrder.class)
+@JsonTypeName("ICEBERG_SORT_ORDER")
+public interface SortOrder {
+  String getSchema();
```
snazy, line 29: I suspect, we don't need this one (same schema serialized twice)

model/src/main/java/org/projectnessie/model/PartitionSpec.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutablePartitionSpec.class)
+@JsonDeserialize(as = ImmutablePartitionSpec.class)
+@JsonTypeName("ICEBERG_PARTITION_SPEC")
+public interface PartitionSpec {
+
+  String getSchema();
```
snazy, line 30: I suspect, we don't need this one (same schema serialized twice)

model/src/main/java/org/projectnessie/model/IcebergTableMetadata.java
```diff
@@ -0,0 +1,101 @@
...
+   * (implementation of) Iceberg's {@code FileIO} configured for the particular Iceberg table.
+   */
+  @Nullable
+  public abstract String getMetadataLocation();
+
+  public abstract int getFormatVersion();
+
+  public abstract String getUuid();
+
+  public abstract String getLocation();
```
snazy, line 69: Do we need the location one in Nessie? Shouldn't _either_ Nessir _or_ Iceberg's FileIO have/hold the table metadata?

snazy (commented):
The approach to have the Iceberg table-metadata in Nessie is good IMO.

I suspect, figuring out which table-metadata attributes / child-attributes are mapped to "propert" types in Nessie and which are "just handled blobs" is WIP.

The global-state for an Iceberg table should IMO only hold the attributes that are globally relevant: last-column-id, last-sequence-number, last-assigned-partition-id and the set of snapshot-ids. Everything else can go into the on-reference state. However, even a protobuf representation of a snapshot and a schema can become somewhat big for the database, so I wonder, whether it makes sense to store the potentially big chunks in a separate table - but that can be figured out later. Should have a test that stores the "biggest snapshot with the biggest schema ever" would help to find the limits.

laurentgo:
Although the change is not that big, I'm still trying to digest the long term implications for Nessie and if we have the right interfaces/API in place to make it useful/efficient. Maybe a couple of examples to illustrate:

* Overall, the current proposal is to send/receive the whole metadata object, but an iceberg change is most likely to be represented by a couple of field changes. Same when reading a table possibly. Do we see a path to improve efficiency?
* The proposal also uses ids to keep track of sub-objects (like snapshot id, partition spec id, ...). @snazy proposed to replace those ids with the real object in the object model, which could make sense for the actual object model but is not very efficient as-is for transport/storage and/or might lose the identity link between the current snapshot and the latest snapshot in the list of snapshots. At the API level, does the use of ids become somehow a problem with other operations like merge/rebase for example (because this does require a semantic knowledge of what the ids represent)?

snazy:
Yea - the change isn't quite big, but pretty intrusive. I think of this one as a PoC/experiment to see how it could work in practice.

Iceberg's API doesn't provide us much meaningful information about what changed. We do only know whether the schema has changed (b/c the schema-id changed) and whether the snapshot has changed.

On the plus side: A get-content operation does only need to provide a "minimal" table-metadata object - i.e. only the current snapshot and the current schema. The big difference is that the table-metadata is sent to Nessie and not to S3.

Another topic that needs to be sorted out is when a snapshot and/or schema object becomes big - in other words: figuring out a viable way wrt to database limitations and performance of how to store and retrieve those objects.

It's practically not feasible to represent the table-metadata object model in its full extent (down to the actual field types for example) in Nessie. So this PoC/experiment is also about what information does Nessie currently and in the foreseeable future need to know and which pieces are better represented as "opaque blobs".

PR #3028, opened 2022-01-10
Track Nessie database repo state/version

### LINKED ISSUE #3009: Track Nessie database repo state/version
With the current work on implementing the ref-log it becomes kinda obvious that repository migration on the database level need to be handled post Nessie 1.0.0.

The idea is to have a single table with probably just a single row that has has the following (SQL-ish) layout:
```sql
CREATE TABLE repo_version (
  id INT PRIMARY KEY,  -- always 0 or 42 or whatever
  update_version LONG, -- some optimistic locking mechanism
  nessie_repo_version INT, -- internally tracked version number, first = 1, next = 2, etc
  properties BLOB -- somewhat generic properties, maybe used later - serialized protobuf
);
```

Rough idea:
* if the repo_version table does not exist, assume version `0` --> create the table, don't fail if it already exists (dumb concurrency failure "handling")
* if the repo_version table is empty, assume version `0` --> add the only row, don't fail if the row already exists (dumb concurrency failure "handling")

Version `0` would be what we have now (including the ref-log).

The scope of this issue is *not* to define how/when a schema-migration happens or who/what performs the schema migration or when/how a certain features added post 1.0.0 is enabled/disabled. This issue is just there to prepare Nessie to be theoretically able to handle older repo versions. A version may include a schema change or may add a feature that requires some data migration. It all depends on a lot of things.

snazy:
With the current work on implementing the ref-log it becomes kinda obvious that repository migration on the database level need to be handled.

The idea in this change is to have a description about the repository that provides these attributes:
* `repo_version` an integer version describing the repository, but not the database layout
* `properties` as a generic key-value bag describing the repository configuration

Behavior:
* if no repository-description exists, assume version `0` and an empty properties

Version `0` would be what we have now (including the ref-log).

Fixes #3009

### DIFF
--- a/versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
+++ b/versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
@@ -21,10 +21,12 @@ import static org.projectnessie.versioned.persist.dynamodb.Tables.TABLE_GLOBAL_L
 import static org.projectnessie.versioned.persist.dynamodb.Tables.TABLE_GLOBAL_POINTER;
 import static org.projectnessie.versioned.persist.dynamodb.Tables.TABLE_KEY_LISTS;
 import static org.projectnessie.versioned.persist.dynamodb.Tables.TABLE_REF_LOG;
+import static org.projectnessie.versioned.persist.dynamodb.Tables.TABLE_REPO_DESC;
 import static org.projectnessie.versioned.persist.dynamodb.Tables.VALUE_NAME;
 import static org.projectnessie.versioned.persist.serialize.ProtoSerialization.toProto;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableMap;
 import com.google.protobuf.InvalidProtocolBufferException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -45,6 +47,7 @@ import org.projectnessie.versioned.persist.adapter.KeyList;
 import org.projectnessie.versioned.persist.adapter.KeyListEntity;
 import org.projectnessie.versioned.persist.adapter.KeyWithType;
 import org.projectnessie.versioned.persist.adapter.RefLog;
+import org.projectnessie.versioned.persist.adapter.RepoDescription;
 import org.projectnessie.versioned.persist.nontx.NonTransactionalDatabaseAdapter;
 import org.projectnessie.versioned.persist.nontx.NonTransactionalDatabaseAdapterConfig;
 import org.projectnessie.versioned.persist.nontx.NonTransactionalOperationContext;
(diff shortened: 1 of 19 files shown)


versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -309,6 +312,52 @@ protected void cleanUpCommitCas(
...
+  protected boolean tryUpdateRepositoryDescription(
+      NonTransactionalOperationContext ctx, RepoDescription expected, RepoDescription updateTo) {
+    AttributeValue updateToBytes =
+        AttributeValue.builder().b(SdkBytes.fromByteArray(toProto(updateTo).toByteArray())).build();
+    try {
+      if (expected != null) {
+        AttributeValue expectedBytes =
+            AttributeValue.builder()
+                .b(SdkBytes.fromByteArray(toProto(expected).toByteArray()))
+                .build();
```
dimas-b, line 330: Do we guarantee a stable binary representation of `RepoDescription`?

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/serialize/ProtoSerialization.java
```diff
@@ -26,12 +26,53 @@
...
 import org.projectnessie.versioned.persist.adapter.KeyWithType;
 import org.projectnessie.versioned.persist.adapter.RefLog;
+import org.projectnessie.versioned.persist.adapter.RepoDescription;
 
 public class ProtoSerialization {
+
+  public static AdapterTypes.RepoProps toProto(RepoDescription repoDescription) {
+    AdapterTypes.RepoProps.Builder proto =
+        AdapterTypes.RepoProps.newBuilder().setRepoVersion(repoDescription.getRepoVersion());
+    repoDescription.getProperties().forEach(proto::putProperties);
```
dimas-b, line 41: The declaration is of `getProperties` implies a simple `Map`... Can we be sure that the order of entries is stable (which is assumed by database update logic)?

It is likely stable within one particular JVM instance, but might change across environments, I guess...

`TestSerialization.java` does not appear to validate this (unless I missed it).

versioned/persist/serialize/src/main/proto/persist.proto
```diff
@@ -116,3 +116,8 @@ message RefPointer {
   Type type = 1;
   bytes hash = 2;
 }
+
+message RepoProps {
+  int32 repo_version = 1;
+  map<string, string> properties = 2;
```
dimas-b, line 122: Does gRPC serialization guarantee a stable binary form of `map` entries?

snazy: (See comment above.)

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/serialize/ProtoSerialization.java
snazy: Great catch! Right, there's any guarantee that a (plain) `Map` has a stable order - it's not even guaranteed for e.g. a `HashMap` within the same JVM (unless a bunch of things are or behave exactly the same...).
protobuf 3 defines a map entry as a list of implicit elements - but that list is just the result of some `.entrySet()`.

Fix is easy: redefine the proto `map<string,string>` as a `repeated Entry`.

I'll post a follow-up PR to fix the same issue in `GlobalStatePointer`.

dimas-b (approved):
LGTM :+1: but CI complains about code formatting :)

nastra (approved):

PR #3042, opened 2022-01-11
Refactor test case validation for reflog

ajantha-bhat:
It is a followup PR as discussed in #2950


snazy (approved):
Looks much better now! Thanks!

PR #3057, opened 2022-01-12
WIP: Add expiration + txn flag to named references

snazy:
The most common use case of branches in Nessie is to "isolate" a
transaction. A "BEGIN TRANSACTION" translates to "create branch",
a "COMMIT TRANSACTION" to "merge" and a "ROLLBACK TRANSACTION" to
"drop branch". Transaction are rather short lived (seconds,
minutes, maybe a few hours - depending on the actual job/task).
Transactions should also have some sort of timeout.

OTOH there are developer branches or integration branches and such.
Those live for a long time, potentially forever.

End users are probably not really interested in seeing all currently
active "transactions", but probably do want to see experiments and
integrations and such - so all the long lived branches.

Therefore, this change adds three new attributes to named references:
* created timestamp (milliseconds since epoch), rather informational,
  can be handy.
* expire timestamp (milliseconds since epoch), optional, no automatic
  action in Nessie, but potentially a hint to some cleanup
  functionality.
* "is transaction" flag to mark branches as "created for a transaction".

The following endpoints have been updated in a backwards compatible way:
* create reference: add optional expire + "is transaction" parameters
* get all references: add optional flags to not return expired references
  and to optionally exclude branches with the transaction flag set.


clients/client/src/main/java/org/projectnessie/client/http/HttpTreeClient.java
```diff
@@ -52,17 +55,30 @@ public ReferencesResponse getAllReferences(ReferencesParams params) {
...
         .readEntity(ReferencesResponse.class);
   }
 
   @Override
-  public Reference createReference(String sourceRefName, @NotNull Reference reference)
+  public Reference createReference(
+      String sourceRefName,
+      @NotNull Reference reference,
+      @Nullable Instant expireAt,
+      @Nullable Boolean txn)
```
rymurr, line 73: transaction instead of txn? same in other places (particularly user and openapi facing methods

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -217,16 +218,18 @@ Hash merge(
...
    * Create a new named reference.
    *
-   * @param ref Named reference to create - either a {@link org.projectnessie.versioned.BranchName}
-   *     or {@link org.projectnessie.versioned.TagName}.
+   * @param ref Named reference to create - either a {@link BranchName} or {@link
+   *     org.projectnessie.versioned.TagName}.
    * @param target The already existing named reference with an optional hash on that branch. This
    *     parameter can be {@code null} for the edge case when the default branch is re-created after
    *     it has been dropped.
+   * @param expireAt
```
rymurr, line 226: nit: docs

snazy:
Let me see how this can be simplified w/o doing a full refactor of the public API.

PR #3061, opened 2022-01-12
Cleanup HttpClient

snazy:
* Introduce a `HttpRuntimeConfig` holding the runtime configuration in `HttpClient` passed into every `HttpRequest`
* Move request and response filters into `HttpRuntimeConfig` and configure those filters via the `HttpClient.Builder`
* Add missing configuration options for read + connect timeout
* Wire up missing SSLContext option
* Prepare for the HTTP-compression PR

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpRuntimeConfig.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpRuntimeConfig.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


clients/client/src/main/java/org/projectnessie/client/http/HttpRuntimeConfig.java
```diff
@@ -0,0 +1,83 @@
...
+package org.projectnessie.client.http;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.net.URI;
+import java.util.List;
+import java.util.Objects;
+import javax.net.ssl.SSLContext;
+
+/** Package-private HTTP configuration holder. */
+final class HttpRuntimeConfig {
```
nastra, line 25: nit, maybe rename to `HttpClientConfig`?

nastra (approved):
overall LGTM, the only minor confusion that we currently have in the code in terms of naming is `HttpClientBuilder` and `HttpClient.Builder`

clients/client/src/main/java/org/projectnessie/client/http/HttpRuntimeConfig.java
snazy: Explicitly named it like this as it's a mix of configuration from the builder(s) plus dynamic registrations (the request+response filters).

PR #3104, opened 2022-01-19
Check that the target ref can be read when authorization is enabled

### LINKED ISSUE #3058: Validate read access to all reference names passed to API methods
Conceptually the Nessie authorization model relies on reference (branch/tag) names. It stands to reason that the user should have at least `READ` access to all references used in an operation.

ATM, for example, the`assignReference` method does not validate access to the `assignTo` reference, it only validates access to the reference being changed. https://github.com/projectnessie/nessie/blob/dea9dadaabe6b821814350f02ae39d2f4e4024ea/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java#L129

This and possibly other use cases should be fixed to ensure access to all references used in API methods is validated.

nastra:
In this case the target ref is the one
* we assign a reference to
* merge a branch into
* transplant commits into

fixes #3058

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
+++ b/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
@@ -123,7 +123,7 @@ public class CELUtil {
   /**
    * 'Mirrors' Nessie model objects for CEL.
    *
-   * @param model Ńessie model object
+   * @param model Nessie model object
    * @return object suitable for CEL expressions
    */
   public static Object forCel(Object model) {
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
@@ -124,9 +124,12 @@ public class TreeApiImplWithAuthorization extends TreeApiImpl {
   @Override
   protected void assignReference(NamedRef ref, String oldHash, Reference assignTo)
       throws NessieNotFoundException, NessieConflictException {
-    namedRefWithHashOrThrow(assignTo.getName(), assignTo.getHash());
-
-    getAccessChecker().canAssignRefToHash(createAccessContext(), ref);
+    ServerAccessContext accessContext = createAccessContext();
+    getAccessChecker()
+        .canViewReference(
+            accessContext,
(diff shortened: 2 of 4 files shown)


snazy (dismissed):

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
```diff
@@ -124,9 +124,12 @@ public Reference createReference(@Nullable String sourceRefName, Reference refer
...
       throws NessieNotFoundException, NessieConflictException {
-    namedRefWithHashOrThrow(assignTo.getName(), assignTo.getHash());
-
-    getAccessChecker().canAssignRefToHash(createAccessContext(), ref);
+    ServerAccessContext accessContext = createAccessContext();
+    getAccessChecker()
+        .canViewReference(
+            accessContext,
+            namedRefWithHashOrThrow(assignTo.getName(), assignTo.getHash()).getValue());
+    getAccessChecker().canAssignRefToHash(accessContext, ref);
```
ajantha-bhat, line 132: I think we need to handle similarly for `transplant` also. Better to check all the API once and handle it as mentioned in #3058

nastra: indeed, we need to check merge+transplant

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
```diff
@@ -123,7 +123,7 @@
   /**
    * 'Mirrors' Nessie model objects for CEL.
    *
-   * @param model Ńessie model object
+   * @param model Nessie model object
```
dimas-b, line 126: :)

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
dimas-b: yes, please

nastra: fixed + added tests

ajantha-bhat (approved):

PR #3111, opened 2022-01-20
Ensure stable order of references in GlobalStatePointer

snazy:
As pointed out in the review of #3028, Java's `Map` does not ensure a stable order of elements.
This was definitely a problem in that PR.

Nessie's `GlobalStatePointer` also uses a `Map` for the named references. Although the
design of the code ensures that the order of elements in `GlobalStatePointer.namedReferences`
is stable, it feels safer to refactor that property to a list.

The "order" of the list is: most recently updated named references appear at the beginning of
that list. This change brings more advantages beside the "defined order or elements":
* Most recently updated references appear before "stale" references.
* Allows to "spill out" stale references to a separate table, if that becomes necessary, i.e.
  when there are too many references, the GSP becomes big and databases suffer from that size.
* (Probably neglectible:) a bit less (de)serialization effort as the map on top of the protobuf
  list is gone.

Note: in protobuf it is safe to "refactor" a `map<K, V>` to `repeated ELEM` where `ELEM ::= message { K key = 1; V value = 2 }` (which is how maps were "defined" before protobuf 3.

### DIFF
--- a/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
+++ b/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
@@ -48,6 +48,7 @@ import java.util.function.ToIntFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
+import javax.annotation.Nullable;
 import org.projectnessie.versioned.BranchName;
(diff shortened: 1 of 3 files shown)


nastra (approved):
LGTM, I'm assuming we can't use something like `LinkedHashMap`?

snazy:
> LGTM, I'm assuming we can't use something like `LinkedHashMap`?

In protobuf? Nope. TBH, it also always a single "Map.get(referenceName)" after deserialization - and it has to construct the map from a list anyways. So technically the `Map`'s mostly "byte code + heap overhead".

PR #3113, opened 2022-01-20
Use stricter regex in UI entries endpoint calls

### LINKED ISSUE #3106: Nessie UI uses incorrect filter expression that finds entries that don't exist at the specified path
**What happened**:
Nessie UI uses incorrect filter expression that finds entries that don't exist at the specified path

Step 3 below should show no entries

1. Add a table at `foo.bar.t1` to Nessie
2. Visit `http://localhost:19120/tree/main/foo`, shows folder `bar` correctly
3. Visit `http://localhost:19120/tree/main/bar`, shows folder `bar` incorrectly

**Details**:
- Nessie server type (docker/built from source) and version:
  - Docker image: `projectnessie/nessie:0.17.0` (present in latest as well)
  - UI & REST

The issue is that the UI makes a call to `http://localhost:19120/api/v1/trees/tree/main/entries?filter=entry.namespace.matches('bar(\\.|$)')&namespaceDepth=2`

The filter should actually be `entry.namespace.matches('^bar(\\.|$)')`, note the added `^` that enforces beginning of line

nastra:
fixes #3106

### DIFF
--- a/ui/src/Explore/Explore.test.tsx
+++ b/ui/src/Explore/Explore.test.tsx
@@ -42,9 +42,7 @@ it("Explore renders", async () => {
   const now = new Date();
   now.setDate(now.getDate() - 1);
   const scope1 = nock("http://localhost/api/v1")
-    .get(
-      "/trees/tree/main/entries?namespaceDepth=1&filter=entry.namespace.matches(%27(%5C%5C.%7C%24)%27)"
-    )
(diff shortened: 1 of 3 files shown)


dimas-b (approved):

nastra:
this is actually not working as expected, thus moving to WIP

snazy (approved):

PR #3122, opened 2022-01-21
Update committer to current user during merge+transplant

### LINKED ISSUE #631: Add tests to cover hash collisions in L1s/L2s
As per #607 we should add some tests to cover all TieredVersionStore operations when there are multiple commits that end up in the same L1 or L2 stripe. This will prevent errors like we saw in #607 and further test the version store.

Some possible test routes:
1) make the L1, L2, L3 size configurable at start up via a system property and run the integration test suite for with very small sizes
2) find a list of partial SHA256 hash collisions so that commits are forced to overlap

### LINKED ISSUE #831: re-set commit meta on transplant/merge

snazy:
This change updates the "committer" to the current user in the commits created by merge + transplant.

Tests to validate the behavior "with security" (i.e. when there's a `SecurityContext` available) are included in `AbstractTestJerseyRest`. Updated the `NessieJaxRsExtension` to provide a `@NessieSecurityContext` test parameter to provide the `SecurityContext` from tests.

Fixes #831

### DIFF
--- a/servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
+++ b/servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
@@ -17,16 +17,20 @@ package org.projectnessie.jaxrs.ext;
 
 import static org.projectnessie.services.config.ServerConfigExtension.SERVER_CONFIG;
 
+import java.util.function.Consumer;
 import java.util.function.Supplier;
+import javax.ws.rs.container.ContainerRequestFilter;
 import javax.ws.rs.core.Application;
+import javax.ws.rs.core.SecurityContext;
 import org.glassfish.jersey.message.DeflateEncoder;
 import org.glassfish.jersey.message.GZipEncoder;
 import org.glassfish.jersey.server.ResourceConfig;
 import org.glassfish.jersey.server.filter.EncodingFilter;
 import org.glassfish.jersey.test.JerseyTest;
 import org.jboss.weld.environment.se.Weld;
-import org.junit.jupiter.api.extension.AfterAllCallback;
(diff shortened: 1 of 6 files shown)


servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
```diff
@@ -86,15 +90,23 @@ public void beforeAll(ExtensionContext extensionContext) {
...
   @Override
-  public void afterAll(ExtensionContext extensionContext) {
-    // NOP - EnvHolder will be closed by JUnit5
+  public void afterEach(ExtensionContext extensionContext) throws Exception {
+    EnvHolder env = extensionContext.getStore(NAMESPACE).get(EnvHolder.class, EnvHolder.class);
+    env.resetSecurityContext();
+  }
+
+  @Override
+  public void beforeEach(ExtensionContext extensionContext) throws Exception {
```
nastra, line 99: nit: do we need to do the same thing in beforeeach + aftereach?

nastra (approved):

servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
snazy: Call me overly cautious ;)

PR #3124, opened 2022-01-22
Handle HTTP request headers case-insensitive

### LINKED ISSUE #2835: Case-insensitive HTTP header processing in HTTP client
HTTP headers are case-insensitive by the spec.

The Nessie HTTP Client should probably process them in a case-insentice way too (cf. `RequestContext.putHeader()`)

snazy:
Introduces new `HttpHeaders` and `HttpHeader` classes.

FIxes #2835

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpHeaders.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpHeaders.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


clients/client/src/main/java/org/projectnessie/client/http/HttpHeaders.java
```diff
@@ -0,0 +1,84 @@
...
+    return headers.containsKey(key);
+  }
+
+  public void remove(String name) {
+    String key = name.toLowerCase(Locale.ROOT);
+    headers.remove(key);
+  }
+
+  public Iterable<String> getValues(String name) {
+    String key = name.toLowerCase(Locale.ROOT);
```
dimas-b, line 46: nit: maybe extract this into a `private` method like `normalizeHeaderName`?

dimas-b (approved):
LGTM :+1: thx!

PR #3138, opened 2022-01-24
Ref convertion utility instead of repeated instanceof checks

snazy:


servers/services/src/test/java/org/projectnessie/services/impl/RefUtilTest.java
```diff
@@ -0,0 +1,137 @@
...
+                      @Override
+                      @Nonnull
+                      public String getName() {
+                        return REF_NAME;
+                      }
+                    },
+                    HASH_VALUE))
+        .isInstanceOf(IllegalArgumentException.class);
+    assertThatThrownBy(() -> toReference(() -> REF_NAME, HASH_VALUE))
+        .isInstanceOf(IllegalArgumentException.class);
```
nastra, line 109: nit: would be good to check that we get the correct error messages. Same for all the other places where things are verified via `assertThatThrownBy`

nastra (approved):

nastra (approved):

PR #3140, opened 2022-01-24
Fix get-commit-log access check

snazy:


dimas-b (approved):

dimas-b (approved):

nastra (commented):
Do we want to do checks when somebody does paginated requests? I would assume that you'd only want to do authz checks on the first request and not on subsequent requests

dimas-b:
I think so ^ otherwise we'd have a security hole for maliciously crafted "next page" requests, wouldn't we?

snazy:
Right - the page-token is currently just the hash. Pretty easy to figure that out - it's in the source code.

PR #3149, opened 2022-01-25
Unify TreeApi.deleteX() + TreeApi.assignX()

snazy:
Refactors the deleteBranch + deleteTag methods to deleteReference
and assignBranch + assignTag to assignReference.

The strings `tag` + `branch` in the REST paths become path
parameters based on an enum.

This is a non-breaking change for the REST API.

This eliminates duplicate code.

### DIFF
--- a/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
@@ -272,98 +272,40 @@ public interface HttpTreeApi extends TreeApi {
 
   @Override
   @PUT
-  @Path("tag/{tagName}")
+  @Path("{referenceType}/{referenceName}")
   @Operation(
-      summary = "Set a tag to a specific hash via a named-reference.",
+      summary = "Set a named reference to a specific hash via a named-reference.",
       description =
-          "This operation takes the name of the tag to reassign and the hash and the name of a "
+          "This operation takes the name of the named reference to reassign and the hash and the name of a "
               + "named-reference via which the caller has access to that hash.")
   @APIResponses({
     @APIResponse(responseCode = "204", description = "Assigned successfully"),
     @APIResponse(responseCode = "400", description = "Invalid input, ref/hash name not valid"),
     @APIResponse(responseCode = "401", description = "Invalid credentials provided"),
-    @APIResponse(responseCode = "403", description = "Not allowed to view or assign tag"),
-    @APIResponse(responseCode = "404", description = "One or more references don't exist"),
-    @APIResponse(responseCode = "412", description = "Update conflict")
-  })
-  void assignTag(
-      @Parameter(
-              description = "Tag name to reassign",
-              examples = {@ExampleObject(ref = "ref")})
-          @PathParam("tagName")
-          String tagName,
-      @Parameter(
-              description = "Expected previous hash of tag",
-              examples = {@ExampleObject(ref = "hash")})
-          @QueryParam("expectedHash")
-          String oldHash,
-      @RequestBody(
-              description =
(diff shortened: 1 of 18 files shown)


model/src/main/java/org/projectnessie/model/Reference.java
```diff
@@ -80,4 +81,19 @@ default void checkHash() {
...
+  ReferenceType getType();
+
+  /**
+   * The reference type, used inside REST API paths.
+   *
+   * <p>Enum values are intentionally lower-case, so this enum can be directly used as a REST API
+   * path parameter.
+   */
+  enum ReferenceType {
+    branch,
```
dimas-b, line 96: Upper case enum name?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -180,21 +181,13 @@ private GetNamedRefsParams getGetNamedRefsParams(boolean fetchMetadata) {
...
-              refType = "REFERENCE";
-            }
             return script.execute(
                 Boolean.class,
                 ImmutableMap.of(
                     VAR_REF,
                     reference,
                     VAR_REF_TYPE,
-                    refType,
+                    reference.getType().name().toUpperCase(Locale.ROOT),
```
dimas-b, line 190: This should not be necessary, I guess, if enum names were upper case.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -272,98 +272,40 @@ LogResponse getCommitLog(
...
-  void deleteTag(
+  void assignReference(
       @Parameter(
-              description = "Tag to delete",
+              description = "Reference type to reassign",
               examples = {@ExampleObject(ref = "ref")})
-          @PathParam("tagName")
-          String tagName,
+          @PathParam("referenceType")
+          Reference.ReferenceType referenceType,
```
dimas-b, line 294: Could you add examples for  `referenceType` params to `openapi.yaml`?

dimas-b (commented):
LGTM overall, minor comments.

model/src/main/java/org/projectnessie/model/Reference.java
snazy: See comment - intentionally lower case.

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
snazy: sure, added

model/src/main/java/org/projectnessie/model/Reference.java
dimas-b: Sorry missed that :facepalm:

dimas-b: On the other hand, could we manually convert the path parameter (`String`) to the enum and uphold  the upper case naming convention? Would it be "cleaner" code-wise?

I'm a bit hesitant because the `model` module is not really specific to HTTP, but in this case we introduce a hidden dependency on something that is specific to HTTP. WDYT?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
nastra: I agree, I think it would be nicer to rather do `reference.getType().name().toLowerCase()` in the places where the type is being used in the URI

model/src/main/java/org/projectnessie/model/Reference.java
snazy: Yup - updated that

nastra (approved):

PR #3150, opened 2022-01-25
Honour the client's commit meta commit time if set

ajantha-bhat:
Server is always setting the commit time to now. Which may be not same as client's timezone or client's scenario.

P.S: for my new nessie gc test case framework, I need commit timestamp to reflect the test case timestamp.


snazy (changes requested):
Commit timestamp is _always_ the server side timestamp and never modifiable by the user. User's time is recorded as authorTime.

ajantha-bhat:
@snazy : But it is exposed for configuration? I can understand that merge and transplant should use current time. But can commit use the time set by user (not mandatory, but if set by the user)? 

My GC algorithm need to be based on author time then ? else I cannot write a testcase that can verify GC on configured time.

snazy:
> My GC algorithm need to be based on author time then ? else I cannot write a testcase that can verify GC on configured time.

Commit timestamp in Nessie is always the server side current timestamp (analogous to Git).
You can refactor `TreeApiImpl` to get a `java.time.Clock` injected and take `now()` from there, right?

ajantha-bhat:
Closing this as commit time should always be set at server side.

PR #3163, opened 2022-01-26
Update hash-check rule to allow hex representation of 3..32 bytes

snazy:
Was 8..32 bytes before.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/Validation.java
+++ b/model/src/main/java/org/projectnessie/model/Validation.java
@@ -25,17 +25,17 @@ import java.util.regex.Pattern;
 
 /** Collection of validation rules. */
 public final class Validation {
-  public static final String HASH_REGEX = "^[0-9a-fA-F]{16,64}$";
+  public static final String HASH_REGEX = "^[0-9a-fA-F]{8,64}$";
   public static final String REF_NAME_REGEX =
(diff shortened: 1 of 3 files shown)


nastra (commented):
not sure I understand the reasoning behind this change, can you elaborate?

snazy:
Thinking of abbreviated commit hashes - repos likely start with a quite short abbreviated hash, so 6 hex digits feels a bit nicer than a minimum of 16 hex digits.
I don't insist on this changed - just stumbled upon this.

nastra:
> Thinking of abbreviated commit hashes - repos likely start with a quite short abbreviated hash, so 6 hex digits feels a bit nicer than a minimum of 16 hex digits. I don't insist on this changed - just stumbled upon this.

just wondering whether we should have at the very least a length of 7 or 8 instead of 6? Git uses 8 for abbreviated hashes

snazy:
I'm okay with 8

dimas-b:
git is dynamic AFAIK, it will use more digits if it expected higher clash probability

dimas-b (approved):

dimas-b (commented):
~~Actual change LGTM, but I'm not sure I understand the commit message :) the `4..32` and the `Was 8..32 bytes before` parts... the numbers do not seem to match code diffs~~

nvm, got it

PR #3169, opened 2022-01-26
Move tests in AbstractTestRest to multiple classes

snazy:
... so each class is not a "monster of couple thousand lines of code".

Splitting `AbstractTestRest` into separate classes and including those
via `@Nested` with plain JUnit 5 works fine. But it does not work as
a `@QuarkusTest`.

Using `@Nested` test classes like this:
```
public void MyTest {
  @Nested
  public class Group {
    @Test
    void foo() {}
  }
}
```
does not work, because Quarkus considers the non-static inner class
`Group` as a separate test class and fails horribly, mostly complaining
that `...MyTest$Group` is not in one of the expected class folders.

Even worse is that the whole Quarkus machinery is started for each
nested test class and not just for `MyTest`.

As an alternative it felt doable to refactor the former inner classes
to interfaces (so test methods become default methods), but that did
not work as well due to an NPE, when
`QuarkusSecurityTestExtension.beforeEach` tries to find the `@TestSecurity`
annotation - it just asserts whether the current `Class` is `!= Object.class`,
but `Class.getSuperclass()` returns `null` for interfaces - hence the
NPE there.

The "solution" here is to keep the separate classes but let each
extend another - so different groups of tests are kept in a single class.


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
```diff
@@ -15,99 +15,22 @@
...
-import org.projectnessie.model.SqlView.Dialect;
 import org.projectnessie.model.Tag;
-import org.projectnessie.model.Validation;
 
-public abstract class AbstractTestRest {
-  public static final String COMMA_VALID_HASH_1 =
-      ",1234567890123456789012345678901234567890123456789012345678901234";
-  public static final String COMMA_VALID_HASH_2 = ",1234567890123456789012345678901234567890";
-  public static final String COMMA_VALID_HASH_3 = ",1234567890123456";
+public abstract class AbstractTestRest extends AbstractRestRefLog {
```
dimas-b: nit: could you add javadoc to this class about the inheritance model and suggestions about when to add new test superclasses?

dimas-b (approved):
LGTM :+1: with flashbacks of multiple inheritance :wink: 

I wonder if it might work with `default` test methods in interfaces... have not tried... just thinking out loud.

snazy:
> I wonder if it might work with default test methods in interfaces... have not tried... just thinking out loud. If interface methods need "state" I suppose it could be injected via a Junit5 extension.

That's what I tried :(

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractTestRest.java
snazy: done

PR #3170, opened 2022-01-26
Skip integration tests by default for PRs

snazy:
Introduces a new Maven profile `withoutITs`.

PRs run without integration tests by default. The new label `pr-with-ITs` enables running ITs.

This improves CI runtime for PRs from ~30 minutes to <20 minutes.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -45,11 +45,20 @@ jobs:
     - name: Setup Java, Maven, Gradle
       uses: ./.github/actions/dev-tool-java
 
+    - name: Config
+      run: |
+        if [[ "${{ contains(github.event.pull_request.labels.*.name, 'pr-with-ITs') }}" == "false" ]]; then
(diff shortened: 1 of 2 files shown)


dimas-b (commented):
The time savings do not look that great considering review time. Is it really worth the added risk of breaking ITs inadvertently?

nastra:
honestly I don't know if it's a good idea to skip integration tests by default. I can see that flag being useful for local testing, but I believe integration tests should be run by default.
In order to improve CI time we could rather have a separate GH action that only runs those integration tests. That way we can parallelize things

omarsmak:
I also think that is not worth it to disable the IT by default, 10min for me doesn't seem much considering we have to wait for hours in other pipelines 😅.

snazy:
> 10min for me doesn't seem much considering we have to wait for hours in other pipelines

Well, it's a 50% improvement ;)

But what I read as a consensus is that this is potentially dangerous - so I'll close this one.

PR #3176, opened 2022-01-27
Filter support for reflog and streamUtil implementation

ajantha-bhat:
For new GC algorithm, need to filter just DELETE_REFERENCE and ASSIGN_REFERENCE reflog entries from the backend store. Having CEL filter can help in reducing the data in HTTP response by providing the filtered results.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
+++ b/clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
@@ -27,6 +27,7 @@ import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.EntriesResponse.Entry;
 import org.projectnessie.model.LogResponse;
 import org.projectnessie.model.LogResponse.LogEntry;
+import org.projectnessie.model.RefLogResponse;
 import org.projectnessie.model.Reference;
 import org.projectnessie.model.ReferencesResponse;
 
@@ -112,6 +113,34 @@ public final class StreamingUtil {
                     .untilHash(untilHash)
                     .get())
         .generateStream(ref, maxRecords);
+  }
+
+  /**
+   * Default implementation to return a stream of reflog entries, functionally equivalent to calling
+   * {@link NessieApiV1#getRefLog()} with manual paging.
+   *
+   * <p>The {@link Stream} returned by {@code getReflogStream(OptionalInt.empty())}, if not limited,
+   * returns all reflog entries.
+   *
+   * @param api The {@link NessieApiV1} to use
+   * @return stream of {@link RefLogResponse.RefLogResponseEntry} objects
+   */
+  public static Stream<RefLogResponse.RefLogResponseEntry> getReflogStream(
+      @NotNull NessieApiV1 api,
+      @Nullable String fromHash,
+      @Nullable String untilHash,
+      @Nullable String filter,
+      OptionalInt maxRecords)
+      throws NessieNotFoundException {
+    return new ResultStreamPaginator<>(
+            RefLogResponse::getLogEntries,
+            (reference, pageSize, token) ->
+                builderWithPaging(api.getRefLog(), pageSize, token)
+                    .fromHash(fromHash)
+                    .untilHash(untilHash)
+                    .filter(filter)
+                    .get())
+        .generateStream(null, maxRecords);
   }
 
   private static <B extends PagingBuilder<B>> B builderWithPaging(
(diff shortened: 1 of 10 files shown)


ajantha-bhat:
> For new GC algorithm, need to filter just DELETE_REFERENCE and ASSIGN_REFERENCE reflog entries from the backend store. Having CEL filter can help in reducing the data in HTTP response by providing the filtered results.

@snazy : I know I got a comment to remove filtering during reflog implementation. But now the GC algorithm demands it and it can be one of the usecase for filtering. Hence need this PR.

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
```diff
@@ -52,16 +52,30 @@
...
+              + "Hint: when filtering entries, you can determine whether entries are \"missing\" (filtered) by "
+              + "checking whether 'ReflogResponseEntry.parentRefLogId' is different from the hash of the previous "
+              + "reflog in the log response.")
+  @QueryParam("filter")
+  private String filter;
+
   public RefLogParams() {}
 
-  private RefLogParams(String startHash, String endHash, Integer maxRecords, String pageToken) {
+  private RefLogParams(
```
snazy, line 69: The whole class is missing a `RefLogParamsTest`, similar to the other params test classes.

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
```diff
@@ -52,16 +52,30 @@
...
+      description =
+          "A Common Expression Language (CEL) expression. An intro to CEL can be found at https://github.com/google/cel-spec/blob/master/doc/intro.md.\n\n"
+              + "Usable variables within the expression are:\n\n"
+              + "- 'reflog' with fields 'refLogId' (string), 'refName' (string), 'commitHash' (string), 'parentRefLogId' "
+              + "(string), ',operation' (string), 'operationTime' (long)\n\n"
+              + "Hint: when filtering entries, you can determine whether entries are \"missing\" (filtered) by "
+              + "checking whether 'ReflogResponseEntry.parentRefLogId' is different from the hash of the previous "
+              + "reflog in the log response.")
+  @QueryParam("filter")
+  private String filter;
```
snazy, line 65: `@Nullable` missing

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -114,6 +115,34 @@ private StreamingUtil() {
...
+   * Default implementation to return a stream of reflog entries, functionally equivalent to calling
+   * {@link NessieApiV1#getRefLog()} with manual paging.
+   *
+   * <p>The {@link Stream} returned by {@code getReflogStream(OptionalInt.empty())}, if not limited,
+   * returns all reflog entries.
+   *
+   * @param api The {@link NessieApiV1} to use
+   * @return stream of {@link RefLogResponse.RefLogResponseEntry} objects
+   */
+  public static Stream<RefLogResponse.RefLogResponseEntry> getReflogStream(
```
snazy, line 128: No tests for this?

snazy (commented):
Can you elaborate _why_ the filter is needed now?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
ajantha-bhat: the filter testcase is using this streamUtil itself. So, it is covered.

ajantha-bhat:
> Can you elaborate why the filter is needed now?
As mentioned in the GC algorithm document (shared internally long back). Originally I thought I will get all the commits from reflog and do anti join with LIVE commits to get expired commits. But finding live commits is no use as the commit can have both dead and live contents. So, the granularity should be contents not commits. So, dropped this algorithm. 

Instead changed algorithm to fetch dead contents using reflog by traversing only DELETE_REF and ASSIGN_REF (potentially unreachable) and union them with expired contents form live reference. Hence need filter to fetch reflog entries for only DELETE_REF and ASSIGN_REF.

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
ajantha-bhat: it is missed in commitLogParams also, I will handle it.

model/src/main/java/org/projectnessie/api/params/RefLogParams.java
ajantha-bhat: Added it similar to `CommitLogParamsTest`

nastra (approved):

Issue #3060, opened 2022-01-12
Improve test runtime


snazy:
There are several things that can be done to let Nessie build faster w/ tests:

* Let the tests in `:nessie-jaxrs-tests` run in parallel
* Let the tests in `:nessie-quarkus` run in parallel
* Improve the runtime of the tests in `:nessie-versioned-persist-dynamodb` (runs for about 9min locally)

PR #3177, opened 2022-01-27
Make tests that use Dynamo/Mongo/Rocks integration-tests

snazy:
Means: faster local builds with `-DskipITs`

Related: #3060

### DIFF
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -315,22 +315,25 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-failsafe-plugin</artifactId>
-        <configuration>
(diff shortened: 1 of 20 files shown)


servers/quarkus-server/pom.xml
```diff
@@ -279,6 +279,8 @@
               <filesets>
                 <fileset>
                   <directory>${project.build.directory}/rocksdb-test</directory>
```
dimas-b, line 281: Is the `rocksdb-test` directory still used now?

dimas-b (approved):

servers/quarkus-server/pom.xml
snazy: Not really, but I kept it just in case.

PR #3184, opened 2022-01-27
Add hashOnRef to DiffApi from + to

snazy:
Support "hashOnRef" on the `diffs/` endpoint for both "from" and "to".

The change is backwards compatible for older clients (older clients continue to work, newer clients may include the hashOnRef, which is not understood by older servers).

Example paths:
* `diffs/main...myBranch`
* `diffs/main...myBranch*1234567890123456`
* `diffs/main*1234567890123456...myBranch`
* `diffs/main*1234567890123456...myBranch*1234567890123456`

It's using the `*` character to separate the reference name from the hash, because that's one of the few characters that are not HTTP-escaped.

### DIFF
--- a/model/src/main/java/org/projectnessie/api/http/HttpDiffApi.java
+++ b/model/src/main/java/org/projectnessie/api/http/HttpDiffApi.java
@@ -38,8 +38,20 @@ public interface HttpDiffApi extends DiffApi {
 
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  @Path("{fromRef}...{toRef}")
-  @Operation(summary = "Get a diff for two given references")
+  @Path(
+      "{fromRef : [^*]+}{f : [*]?}{fromHashOnRef : ([^.]*)?}...{toRef : [^*]+}{t : [*]?}{toHashOnRef : ([^.]*)?}")
+  @Operation(
+      summary = "Get a diff for two given references",
+      description =
+          "The URL pattern is basically 'from' and 'to' separated by '...' (three dots). "
+              + "'from' and 'to' must start with a reference name, optionally followed by hash on "
+              + "that reference, the hash prefixed with the'*' character.\n"
+              + "\n"
+              + "Examples: \n"
+              + "  diffs/main...myBranch\n"
+              + "  diffs/main...myBranch*1234567890123456\n"
+              + "  diffs/main*1234567890123456...myBranch\n"
+              + "  diffs/main*1234567890123456...myBranch*1234567890123456\n")
   @APIResponses({
     @APIResponse(
         responseCode = "200",
--- a/model/src/main/java/org/projectnessie/model/Validation.java
+++ b/model/src/main/java/org/projectnessie/model/Validation.java
@@ -25,11 +25,14 @@ import java.util.regex.Pattern;
 
 /** Collection of validation rules. */
 public final class Validation {
-  public static final String HASH_REGEX = "^[0-9a-fA-F]{8,64}$";
-  public static final String REF_NAME_REGEX =
-      "^[A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])?$";
+
+  public static final String HASH_RAW_REGEX = "[0-9a-fA-F]{8,64}";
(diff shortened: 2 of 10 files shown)


nastra (dismissed):
LGTM, do we need to handle the python client as well for this functionality?

model/src/main/java/org/projectnessie/model/Validation.java
```diff
@@ -25,11 +25,14 @@
...
-      "^[A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])?$";
+
+  public static final String HASH_RAW_REGEX = "[0-9a-fA-F]{8,64}";
+  public static final String HASH_REGEX = "^" + HASH_RAW_REGEX + "$";
+  public static final String REF_NAME_RAW_REGEX =
+      "[A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])";
+  public static final String REF_NAME_REGEX = "^" + REF_NAME_RAW_REGEX + "?$";
   public static final String REF_NAME_OR_HASH_REGEX =
-      "^(([0-9a-fA-F]{8,64})|([A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])?))$";
+      "^((" + HASH_RAW_REGEX + ")|(" + REF_NAME_RAW_REGEX + "?))$";
```
dimas-b, line 35: Could the `?` be moved into `REF_NAME_RAW_REGEX` for the sake of clarity? It looks like it always follows it anyway.

model/src/main/java/org/projectnessie/model/Validation.java
```diff
@@ -25,11 +25,14 @@
...
 /** Collection of validation rules. */
 public final class Validation {
-  public static final String HASH_REGEX = "^[0-9a-fA-F]{8,64}$";
-  public static final String REF_NAME_REGEX =
-      "^[A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])?$";
+
+  public static final String HASH_RAW_REGEX = "[0-9a-fA-F]{8,64}";
+  public static final String HASH_REGEX = "^" + HASH_RAW_REGEX + "$";
+  public static final String REF_NAME_RAW_REGEX =
+      "[A-Za-z](((?![.][.])[A-Za-z0-9./_-])*[A-Za-z0-9._-])";
```
dimas-b, line 32: Single trailing dots (`.`) appear to be allowed in ref names... Will this work with the `getDiff` URL pattern? Will it not cause confusion with the `...` separator of the `from` and `to` sections?

If it works well, could you add a test case to make it explicit?

model/src/main/java/org/projectnessie/model/Validation.java
snazy: Nice catch! Yea - that's where it actually should be.

model/src/main/java/org/projectnessie/api/http/HttpDiffApi.java
```diff
@@ -38,8 +38,20 @@
 
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  @Path("{fromRef}...{toRef}")
-  @Operation(summary = "Get a diff for two given references")
+  @Path(
+      "{fromRef : [^*]+}{f : [*]?}{fromHashOnRef : ([^.]*)?}...{toRef : [^*]+}{t : [*]?}{toHashOnRef : ([^.]*)?}")
```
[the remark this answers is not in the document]
snazy: Makes sense, considering that it does conflict with the diff-URL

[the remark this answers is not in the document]
snazy: So, a `.` is valid _in_ the reference name, but the ref name must not _end_ with a `.`

model/src/main/java/org/projectnessie/model/Validation.java
snazy: See above, `.` no longer valid at the ref-name end.

dimas-b (approved):

PR #3185, opened 2022-01-27
Allow testing `AccessChecker` via `NessieJaxrsExtension`

snazy:
This allows upcoming PR(s) to run tests in nessie-jaxrs-tests via Jersey that supply mocked/custom `AccessContext` implementations. Also adds a convenience `DefaultAccessContext` implementation.

### DIFF
--- a/servers/jax-rs/src/main/java/org/projectnessie/services/authz/AccessCheckerExtension.java
+++ b/servers/jax-rs/src/main/java/org/projectnessie/services/authz/AccessCheckerExtension.java
@@ -16,6 +16,7 @@
 package org.projectnessie.services.authz;
 
 import java.security.AccessControlException;
+import java.util.function.Supplier;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 import javax.enterprise.inject.Default;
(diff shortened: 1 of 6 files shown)


servers/jax-rs/src/main/java/org/projectnessie/services/authz/AccessCheckerExtension.java
```diff
@@ -27,51 +28,99 @@
 
 /** This class needs to be in the same package as {@link AccessChecker}. */
 public class AccessCheckerExtension implements Extension {
-  public static final AccessChecker ACCESS_CHECKER =
+  private static final AccessChecker NOOP_ACCESS_CHECKER = new DefaultAccessChecker();
+
+  private volatile Supplier<AccessChecker> accessCheckerSupplier;
```
dimas-b, line 33: Could this be `final`? It looks like test methods do not have access to this field directly and have to go through `NessieJaxRsExtension` :thinking:

dimas-b (approved):

servers/jax-rs/src/main/java/org/projectnessie/services/authz/AccessCheckerExtension.java
snazy: Ah, no.
It doesn't allow `AccessCheckerExtension` (or anything else) to have constructor arguments.

PR #3186, opened 2022-01-27
Allow adjacent path templates in HttpClient

snazy:
This change is required by #3184

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/UriBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/UriBuilder.java
@@ -18,7 +18,6 @@ package org.projectnessie.client.http;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLEncoder;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


clients/client/src/main/java/org/projectnessie/client/http/UriBuilder.java
```diff
@@ -92,32 +89,35 @@ URI build() throws HttpClientException {
...
-                }
-                uriBuilder.append(x);
-              });
+
+      StringBuilder pathElement = new StringBuilder();
+      StringBuilder name = new StringBuilder();
+      int l = uri.length();
+      for (int i = 0; i < l; i++) {
+        char c = uri.charAt(i);
+        if (c == '/') {
```
dimas-b, line 98: Do we have to treat `/` chars in a special way? It looks like we can deal with `{` and `}` only. WDYT?

If we process `/` chars perhaps we can check that `{..}` groups do not intersect with path elements?.. However, I do not think it is necessary.

nastra (approved):

clients/client/src/main/java/org/projectnessie/client/http/UriBuilder.java
snazy: I thought the same, but sadly the `encode` function also escapes `/`.
I didn't want to implement a custom encode function, but maybe we have to at one point, IDK.

PR #3189, opened 2022-01-28
Allow fetching commit log from hash for dead reference for reflog user

ajantha-bhat:
For GC, need to fetch the expired contents (expired because reference is deleted or assigned) from the backend store. 
Currently we only support fetching commit log + contents for the live references. Hence, need a way to fetch from dead references. 

Changes:
When authz is enabled,
    allow fetching commit log from hash for dead reference for reflog user and continue the same restrictions for normal user.
When authz is disabled,
    anyone can fetch commit log from hash for dead reference.

Note: 
When Hash is not provided, cannot fetch commit log from the dead reference irrespective of the access control.


snazy (changes requested):
No need for this one.
I will have the bare-commit-id support ready pretty soon, which will allow reading the commit log from any commit

snazy:
#3196 is the change that will allow reading from bare commit-ids.

ajantha-bhat:
Closing the PR as there is an alternate implementation in #3196

PR #3193, opened 2022-01-28
CLI: Fix merge test

nastra:
Given that a merge now produces a different commit hash on the target
branch, it was required to adjust the test, since it was checking that
the commit hash is the same on the source & target branch.

### DIFF
--- a/python/tests/test_nessie_cli.py
+++ b/python/tests/test_nessie_cli.py
@@ -268,9 +268,12 @@ def test_merge() -> None:
     ref = ReferenceSchema().loads(execute_cli_command(["--json", "branch", "-l", "main"]), many=False)
     main_hash = ref.hash_
     execute_cli_command(["merge", "dev", "-c", main_hash])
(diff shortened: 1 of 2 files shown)


python/tests/test_nessie_cli.py
```diff
@@ -268,9 +268,12 @@ def test_merge() -> None:
     ref = ReferenceSchema().loads(execute_cli_command(["--json", "branch", "-l", "main"]), many=False)
     main_hash = ref.hash_
     execute_cli_command(["merge", "dev", "-c", main_hash])
-    branches = ReferenceSchema().loads(execute_cli_command(["--json", "branch"]), many=True)
-    refs = {i.name: i.hash_ for i in branches}
-    assert refs["main"] == refs["dev"]
+    logs = simplejson.loads(execute_cli_command(["--json", "log"]))
+    # we don't for equality of hashes here because a merge
```
snazy, line 272: typo

nastra: fixed

snazy (approved):

PR #3195, opened 2022-01-28
Remove `VersionStore.toRef()`

snazy:
`VersionStore.toRef()` was already only used from tests and was superseded with `getNamedRef()` a while ago.

### DIFF
--- a/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferences.java
+++ b/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferences.java
@@ -158,33 +158,48 @@ public abstract class AbstractReferences extends AbstractNestedVersionStore {
         ReferenceNotFoundException.class, () -> store().delete(tag, Optional.of(initialHash)));
   }
 
+  /**
+   * Rudimentary test for {@link VersionStore#getNamedRef(String, GetNamedRefsParams)}. Better tests
+   * in {@code AbstractGetNamedReferences} in {@code :nessie-versioned-persist-tests}.
+   */
   @Test
   void getNamedRef() throws VersionStoreException {
-    final BranchName branch = BranchName.of("toRef");
-    store().create(branch, Optional.empty());
-    store().hashOnReference(branch, Optional.empty());
+    final BranchName branch = BranchName.of("getNamedRef");
+    Hash hashFromCreate = store().create(branch, Optional.empty());
+    assertThat(store().hashOnReference(branch, Optional.empty())).isEqualTo(hashFromCreate);
 
-    final Hash firstCommit = commit("First Commit").toBranch(branch);
+    final Hash firstCommitHash = commit("First Commit").toBranch(branch);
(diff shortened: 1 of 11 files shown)


snazy:
Requires #3184

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferences.java
```diff
@@ -160,29 +159,37 @@ public void createAndDeleteTag() throws Exception {
...
     final Hash thirdCommit = commit("Third Commit").toBranch(branch);
 
     store().create(BranchName.of(thirdCommit.asString()), Optional.of(firstCommit));
     store().create(TagName.of(secondCommit.asString()), Optional.of(firstCommit));
 
-    assertThat(store().toRef(secondCommit.asString()))
-        .isEqualTo(WithHash.of(firstCommit, TagName.of(secondCommit.asString())));
-    assertThat(store().toRef(thirdCommit.asString()))
-        .isEqualTo(WithHash.of(firstCommit, BranchName.of(thirdCommit.asString())));
+    assertThat(store().getNamedRef(secondCommit.asString(), GetNamedRefsParams.DEFAULT))
```
dimas-b, line 179: Would it be too much trouble to refactor and use dedicated names/variables for branches/tags to avoid confusion with hashes... unless we're trying to assert that hashes can be used as branch names (which does not seems to be the case here)? The code in its current form was a bit hard to understand for me :)

snazy: Done

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferences.java
```diff
@@ -159,30 +158,58 @@ public void createAndDeleteTag() throws Exception {
...
-    // Is it correct to allow a reference with the sentinel reference?
-    // assertThat(store().toRef(initialCommit.asString()), is(WithHash.of(initialCommit,
-    // initialCommit)));
-    assertThrows(ReferenceNotFoundException.class, () -> store().toRef("unknown-ref"));
-    assertThrows(ReferenceNotFoundException.class, () -> store().toRef("1234567890abcdef"));
+    // Verifies that the result of "getNamedRef" for the branch created at "firstCommitHash" is
+    // correct
+    assertThat(store().getNamedRef(branchName.getName(), GetNamedRefsParams.DEFAULT))
+        .extracting(ReferenceInfo::getHash, ReferenceInfo::getNamedRef)
+        .containsExactly(secondCommitHash, branchName);
```
dimas-b, line 201: Is this not the same as lines 188-190? :thinking:

snazy: Oops, copy-paste-keyboard-bug

nastra (approved):

PR #3196, opened 2022-01-28
Reading with detached commit-IDs

snazy:
(Re)introduces reading from "detached" commit IDs, i.e. only specify the commit-id/hash instead of the reference name or reference name plus hash. This PR does not update or break the public API.

To allow reading with "detached" commit IDs via the existing REST APIs, this change uses a special reference name `DETACHED`.


nastra (approved):
looks great

PR #3199, opened 2022-01-29
Add ManyUniqueTablesSimulation

dimas-b:
This simulation attempts to represent a scenario where
multiple users create unique tables in the main branch.
    
The main purpose of this simulation is exploratory and
resilience testing of Nessie servers.

### DIFF
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/ManyUniqueTablesSimulation.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/ManyUniqueTablesSimulation.scala
@@ -1,0 +1,127 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.perftest.gatling
+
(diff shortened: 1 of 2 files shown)


perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/ManyUniqueTablesSimulation.scala
```diff
@@ -0,0 +1,111 @@
...
+  * This simulation is not very realistic from the end user perspective because
+  * using a new table UUID is each commit is not the intended way of using the
+  * Nessie Catalog. It is generally expected that table UUIDs are reused and
+  * each table is mentioned in more than one commit.
+  *
+  * The goal of this simulation is mostly to probe the robustness and
+  * scalability of the handling of global data in Nessie under contention when
+  * many table UUIDs are at play.
+  *
+  * This simulation is meant to run in two major modes: <ul> <li>Using a high
```
tomekl007, line 39: can we add two pre-defined configurations per those two scenarios?
I see that we have this:
```
object ConcurrentCommitsParams {
  def fromSystemProperties(): ConcurrentCommitsParams = {
    val base = BaseParams.fromSystemProperties()
    ConcurrentCommitsParams(
      Integer.getInteger("sim.warmup.seconds", 1).toInt,
      Integer.getInteger("sim.rampup.seconds", 1).toInt,
      Integer.getInteger("sim.plateau.seconds", 1).toInt,
      Integer.getInteger("sim.warmup.rate", 1).toInt,
      base.opRate,
      base.note
    )
  }
```
but it uses `1` as default. We could create two specific configs: `capacityScenario` and `responseTimes` with a defaults that make sense for testing those.

dimas-b: let me think about this :hourglass:

dimas-b: I've refactored param classes a bit and added scaladoc... hope it clarifies the use cases :)

dimas-b:
will revisit after the WIP model changes are merged... if this scenario is still relevant.

PR #3222, opened 2022-02-01
Revert "Bump error_prone_core from 2.10.0 to 2.11.0 (#3159)"

ajantha-bhat:
This reverts commit 1d789f198be3b0e46310eb18ac1de5ab06e92f84.

Due to the below problem.
Screenshot 2022-02-01 at 8 05 01 AM


snazy (approved):

nastra:
Alternatively we could set `-Xep:UnicodeInCode:OFF` in the pom

snazy:
> Alternatively we could set -Xep:UnicodeInCode:OFF in the pom

That would disable it for Maven builds as well.

PR #3293, opened 2022-02-09
Content attachments in Nessie

snazy:
Allow binary values related to content-ids via database-adapter.

This is a prerequisite to have Icebert table and view metadata in Nessie.

The content attachment keys is a composite of:
* content-id
* object-type (some string, e.g. `snapshot` or `schema`)
* object-id (some identifier, e.g. a hash over the content value or some
  other identifier)

Attachment values:
* persisted with "format" _and_ "compression" attributes.
* field "object ID", which represents e.g. the snapshot ID
* "version" field for optimistic locking + CAS operations

The new functionality via `DatabaseAdapter` allows:
* unconditional bulk writes and reads, used to allow bulk writes + reads of
  for example the Iceberg snapshot + partition spec + schema for an Iceberg
  table.
* conditional writes (think: "insert" + "update if") to store per-content-id
  information about a table - for Iceberg this would be for example the
  "last sequence number" or "last column id".
* listing all attachment keys for a content id, which may return keys for
  non-existent attachments (compromise for non-transactional databases).


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/BinaryContent.java
```diff
@@ -0,0 +1,58 @@
...
+  /** The object ID, part of the {@link BinaryKey}, for example the snapshot-ID or schema-ID. */
+  String getObjectId();
+
+  /** Compression algorithm needed to uncompress {@link #getData()}. */
+  Compression getCompression();
+
+  /** Potentially compressed binary data. */
+  ByteString getData();
+
+  /** Optional "hash" value, used for optimistic locking. */
```
dimas-b, line 44: Optional in what sense? Do we not control it completely on the server side?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/BinaryKey.java
```diff
@@ -0,0 +1,44 @@
...
+public interface BinaryKey {
+
+  /** Content ID to which the binary object identified by this key belongs to. */
+  ContentId getContentId();
+
+  /** The object type, for example {@code snapshot} or {@code schema}. */
+  String getObjectType();
+
+  /** The object ID, for example the snapshot-ID or schema-ID. */
+  String getObjectId();
```
dimas-b, line 31: Do we expect "object ID" to have weak uniqueness?.. and use "type" as part of the key to enhance its uniqueness?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -323,4 +323,35 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
    * @param offset initial reflog id to read from
    */
   Stream<RefLog> refLog(Hash offset) throws RefLogNotFoundException;
+
+  /**
+   * Retrieve the binary objects identified via {@code keys}. Binary objects that do not exist are
+   * not returned.
+   */
+  Stream<BinaryContent> getBinaries(Stream<BinaryKey> keys);
```
dimas-b, line 331: Well, `globalContent` effectively returns opaque binary data too... Could we use a different name for these objects? "Binaries" seems to be too generic to me :thinking:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -323,4 +323,35 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+
+  /**
+   * Consistent put-binary operation.
+   *
+   * <p>Either a "put-if-absent", if {@code expectedHash} is empty or a compare-and-swap based on
+   * the value of {@code expectedHash}. * *
+   *
+   * <p><b>IMPORTANT</b> keys of the {@link BinaryContent} binary objects must not be used via this
+   * method and {@link #putBinaries(Stream)}.
```
dimas-b, line 340: I'm not sure I understand this message, sorry :)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -96,18 +103,27 @@ public void eraseRepo() {
...
     // Note: the order of `put` calls matters
     idDoc.put(ID_REPO_NAME, repositoryId);
-    idDoc.put(ID_HASH_NAME, id.asString());
+    idDoc.put(ID_HASH_NAME, id);
     return idDoc;
   }
 
-  private List<Document> toIds(Collection<Hash> ids) {
-    return ids.stream().map(this::toId).collect(Collectors.toList());
+  private List<Document> toIdsStream(Stream<String> ids) {
```
dimas-b, line 121: nit: do we have to rename this method? It looks like there are no overloads :thinking:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -482,4 +518,84 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+              if (value.hasHash()) {
+                doc.put(VALUE_HASH_PROPERTY_NAME, value.getHash());
+              }
+              return doc;
+            })
+        .forEach(
+            doc ->
+                client
+                    .getBlobs()
+                    .updateOne(
```
dimas-b, line 538: Should we check the result of `updateOne` (e.g. `result.wasAcknowledged()`)? I'm not sure Mongo Client will throw exceptions in all cases the update did not happen :thinking:

dimas-b (commented):
I have a very brief look at this PR. Sorry about naming nitpicks... will try to have a deeper review tomorrow.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/BinaryContent.java
snazy: The issue here, actually: where I'm not yet sure about, is whether it can/should have a "proper hash" or whether some random number is sufficient. So the term "hash" is probably wrong.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/BinaryKey.java
snazy: We can't really enforce it, yes. It's what we get from the "outside system" i.e. Iceberg.
There will be some checks in higher level code that prevents reusing snapshot-IDs or schema-IDs though.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
snazy: Updated it.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: done.

snazy:
> Please correct me if I'm wrong: the overall idea is to have content objects versioned per branch with commit log and each of the versions can refer to zero or more "attachments" by their unique ID (`BinaryKey` or the renamed class). However, "attachments" themselves are immutable and not versioned... right?

Right. Those blobs are attached to a content on a commit. Multiple contents (for the same content id) can refer to the same attachment. Some are immutable (written via `DatabaseAdapter.putAttachments()`), some can be updated (via `DatabaseAdapter.consistentPutAttachment()`). The ones that can be updated represent some global state per content id - it just doesn't need the overall consistency guarantees that the current global-state w/ global-log + global-pointer provides.

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -468,4 +503,101 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+              VERSION_NAME, AttributeValue.builder().s(v.getVersion()).build());
+        });
+  }
+
+  @Override
+  protected boolean consistentWriteAttachment(
+      AttachmentKey key, AttachmentValue value, Optional<String> expectedVersion) {
+
+    AttributeValue newValueBytes =
+        AttributeValue.builder().b(SdkBytes.fromByteArray(value.toByteArray())).build();
```
tomekl007, line 529: why the `AttributeValue.builder` has those not descriptive methods like `s()` or `b()`?
Can we change those to some meaningful names?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
```diff
@@ -252,4 +257,69 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+            if (expectedVersion.isPresent()) {
+              if (current != null) {
+                try {
+                  AttachmentValue currentValue = AttachmentValue.parseFrom(current);
+                  if (currentValue.hasVersion()
+                      && currentValue.getVersion().equals(expectedVersion.get())) {
+                    return value.toByteString();
+                  }
+                } catch (InvalidProtocolBufferException e) {
+                  throw new RuntimeException(e);
```
tomekl007, line 283: shouldn't we create a custom exception that extends `RuntimeException` instead of using it directly?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
```diff
@@ -252,4 +257,69 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+  }
+
+  @Override
+  protected boolean consistentWriteAttachment(
+      AttachmentKey key, AttachmentValue value, Optional<String> expectedVersion) {
+    try {
+      store.attachments.compute(
+          dbKey(key.toByteString()),
+          (k, current) -> {
+            if (expectedVersion.isPresent()) {
```
tomekl007, line 274: Do we have guidelines around handling the `Optional` type?
can we use `expectedVersion.map().orElse()` instead of `if(isPresent())...`?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
```diff
@@ -252,4 +257,69 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+                  if (currentValue.hasVersion()
+                      && currentValue.getVersion().equals(expectedVersion.get())) {
+                    return value.toByteString();
+                  }
+                } catch (InvalidProtocolBufferException e) {
+                  throw new RuntimeException(e);
+                }
+              }
+              // consistent-update failed, key not present or hash!=expected, throw
+              throw new IllegalStateException();
```
tomekl007, line 287: please add some message to this exception. I think it can be even:
`throw new IllegalStateException(consistent-update failed, key not present or hash!=expected)`

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -229,26 +249,32 @@ private void delete(MongoCollection<Document> collection, Set<Hash> ids) {
...
   /**
    * Fetches a collection of documents by hash and returns them in the order of hashes requested.
    *
    * <p>Note: unknown hashes will correspond to {@code null} elements in the result list.
    */
-  private <T> List<T> fetchMappedPage(
-      MongoCollection<Document> collection, List<Hash> hashes, Function<Document, T> mapper) {
-    List<Document> ids = hashes.stream().map(this::toId).collect(Collectors.toList());
+  private <T> List<T> fetchMappedPageGeneric(
+      MongoCollection<Document> collection, List<String> idList, Function<Document, T> mapper) {
```
tomekl007, line 264: why do we get rid of type safety here (using `Hash`) and starting using `String` instead?
I see that the caller has a hash and transforms it to `String` (`Hash::asString`). 
Shouldn't we favor strong type instead of just passing `String`?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -485,4 +521,91 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
...
+                  Filters.and(
+                      Filters.eq(id),
+                      Filters.eq(VALUE_VERSION_PROPERTY_NAME, expectedVersion.get())),
+                  new Document("$set", doc));
+      if (!r.wasAcknowledged()) {
+        throw new IllegalStateException(
+            "Unacknowledged write to " + client.getAttachments().getNamespace());
+      }
+      return r.getMatchedCount() == 1L && r.getModifiedCount() == 1L;
+    } else {
```
tomekl007, line 574: do we want to execute the same else block for `getMatchedCount()>1` and `getMatchedCount()==0`?
Or should those two cases be handled separately?

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/serialize/ProtoSerialization.java
```diff
@@ -287,6 +293,126 @@ public static Key protoToKey(AdapterTypes.Key key) {
...
+    return String.format(
+        "%s::%s::%s", key.getContentId().getId(), key.getObjectType(), key.getObjectId());
+  }
+
+  public static String attachmentKeyAsString(AttachmentKey key) {
+    return String.format(
+        "%s::%s::%s", key.getContentId().getId(), key.getObjectType(), key.getObjectId());
+  }
+
+  public static AttachmentKey attachmentKeyFromString(String key) {
```
tomekl007, line 403: Should the methods that provide a construction of the `AttachmentKey` be in the `AttachmentKey.java` file?

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/serialize/TestSerialization.java
```diff
@@ -360,6 +388,41 @@ public void stableOrderOfRepoDescProps() {
     assertThat(repoProps.getPropertiesList()).containsExactlyElementsOf(expected);
   }
 
+  @Test
+  public void attachmentContentSerialization() throws Exception {
+    for (int i = 0; i < 500; i++) {
```
tomekl007, line 393: can you provide an explanation why `500` was chosen?

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1684,4 +1709,129 @@ private void writeRefLog(Connection connection, AdapterTypes.RefLogEntry entry)
...
+  public Stream<ContentAttachment> getAttachments(Stream<ContentAttachmentKey> keys) {
+    List<ContentAttachmentKey> keysList = keys.collect(Collectors.toList());
+    if (keysList.isEmpty()) {
+      return Stream.empty();
+    }
+    Connection conn = borrowConnection();
+    try {
+      String sql = sqlForManyPlaceholders(SqlStatements.SELECT_ATTACHMENTS, keysList.size());
+      try (PreparedStatement ps = conn.prepareStatement(sql)) {
+        ps.setString(1, config.getRepositoryId());
```
tomekl007, line 1723: why not start from `0`?

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1684,4 +1709,129 @@ private void writeRefLog(Connection connection, AdapterTypes.RefLogEntry entry)
...
+    Connection conn = borrowConnection();
+    try {
+      if (expectedVersion.isPresent()) {
+        try (PreparedStatement ps = conn.prepareStatement(SqlStatements.UPDATE_ATTACHMENT)) {
+          ps.setBytes(1, ProtoSerialization.toProtoValue(attachment).toByteArray());
+          ps.setString(2, attachment.getVersion());
+          ps.setString(3, config.getRepositoryId());
+          ps.setString(4, attachmentKeyAsString(ContentAttachmentKey.fromContent(attachment)));
+          ps.setString(5, expectedVersion.get());
+          if (ps.executeUpdate() != 1) {
```
tomekl007, line 1792: I think that this block:
```
          if (ps.executeUpdate() != 1) {
            return false;
          }
```
can be moved before `ps.set..` and provide fast-path returning false without the need for setting the values in prepared statements, right?

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1684,4 +1709,129 @@ private void writeRefLog(Connection connection, AdapterTypes.RefLogEntry entry)
...
+        }
+      } else {
+        try (PreparedStatement ps = conn.prepareStatement(SqlStatements.INSERT_ATTACHMENT)) {
+          attachmentInsert(ps, attachment);
+        }
+      }
+      txCommit(conn);
+    } catch (SQLException e) {
+      if (isIntegrityConstraintViolation(e)) {
+        return false;
```
tomekl007, line 1804: shouldn't we log that? Right now the information about the failure will be hidden (because we are just returning `false`)

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
snazy: Ehm - no? That's AWS SDK

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
snazy: Generally, yes.
But this is the inmemory implementation, that's only used in tests (hope nobody uses it in prod...)

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
snazy: We don't have guidelines - and I think we shouldn't have.
IMO it should be "readable". As long as the "conventional" approach is readable (or "equally readable" compared to stream-ish one, it's fine, because the implicit one's less work for the VM.

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
snazy: It's a bit of a hack here. Adding a different approach to make it a bit clearer.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: Not sure what you mean here wrt to `return` and the following `else`

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/serialize/ProtoSerialization.java
snazy: Tricky - `AttachmentKey` is a generated protobuf class

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/serialize/TestSerialization.java
snazy: I don't have a particular one - just to have some more randomness here.
500 is a bit too much - updated to 50

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
snazy: JDBC starts at 1

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
snazy: You're suggesting to execute the prepared-statements w/o all bind-variables configured?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
```diff
@@ -252,4 +257,69 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
         .map(ProtoSerialization::protoToRefLog)
         .collect(Collectors.toList());
   }
+
+  @Override
+  protected void writeAttachments(Stream<Entry<AttachmentKey, AttachmentValue>> attachments) {
```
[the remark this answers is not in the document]
snazy: Those are in `org.projectnessie.versioned.persist.tests.AbstractBinary`

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
tomekl007: ah, you are right. I thought that this value was pre-computed

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractBinary.java
```diff
@@ -0,0 +1,199 @@
...
+import org.projectnessie.versioned.persist.adapter.ContentId;
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapter;
+
+/**
+ * Verifies that a big-ish number of keys, split across multiple commits works and the correct
+ * results are returned for the commit-log, keys, global-states. This test is especially useful to
+ * verify that the embedded and nested key-lists (think: full-key-lists in a commit-log-entry) work
+ * correctly.
+ */
+public abstract class AbstractBinary {
```
tomekl007, line 36: How do we separate units from integration tests?

snazy: For the database adapters it's basically a "unit test", if the database adapter is in memory (so H2 + the inmem adapter) - if it's a separate process (e.g. Mongo, Dynamo, Postgres), then it's an integration test. Similar for the tests via Quarkus.

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
tomekl007: maybe add it in the Javadoc of this class (that it should not be used in prod)?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractBinary.java
snazy: IT = maven-failsafe-plugin
Test* or *Test = maven-surefire-plugin

You mean this?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
snazy: Different topic ;)
If, then it should be everywhere - especially application.properties, website, etc. Wanna take a stab on that?

tomekl007: Maybe we should create a ticket for that and decide if it is worth doing depending on our priorities?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
tomekl007: I am asking because many OS projects have guidelines around handling Optional. A lot of them prefer the non-functional approach (`isPresent`) and some of them functional (`map()..`) so it maybe beneficial to have unified rules regarding `Optional` handling

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/serialize/TestSerialization.java
tomekl007: If we need randomness, maybe we should consider property-based testing?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
snazy: Yup, please go ahead.

tomekl007: created: https://github.com/projectnessie/nessie/issues/3414

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -302,6 +304,44 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+  Stream<ContentAttachment> getAttachments(Stream<ContentAttachmentKey> keys);
+
+  /**
+   * Consistent put-attachment operation.
+   *
+   * <p>Either a "put-if-absent", if {@code expectedHash} is empty or a compare-and-swap based on
+   * the value of {@code expectedHash}. * *
+   *
+   * <p>Note: this method uses conditional put operations, unlike {@link #putAttachments(Stream)}.
+   * Should not use this method and {@link #putAttachments(Stream)} for the same keys.
```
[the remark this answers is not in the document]
dimas-b: Is this a hard requirement or more of a suggestion?

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/AbstractTxDatabaseAdapterTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.sql.SQLException;
+import java.sql.Statement;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.versioned.persist.tests.AbstractDatabaseAdapterTest;
+import org.projectnessie.versioned.persist.tx.TxDatabaseAdapter.NessieSqlDataType;
+
+public abstract class AbstractTxDatabaseAdapterTest extends AbstractDatabaseAdapterTest {
+  @Test
+  void isIntegrityConstraintViolation() throws Exception {
```
dimas-b, line 29: nit: rename to `insertOnConflict` perhaps (for clarity)?

versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryDatabaseAdapter.java
```diff
@@ -262,4 +268,149 @@ protected List<RefLog> doFetchPageFromRefLog(
...
+    attachments.forEach(
+        e -> {
+          AttachmentKey attachmentKey = e.getKey();
+          addAttachmentKey(attachmentKey);
+          store.attachments.put(dbKey(attachmentKey.toByteString()), e.getValue().toByteString());
+        });
+  }
+
+  private void addAttachmentKey(AttachmentKey attachmentKey) {
```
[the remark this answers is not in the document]
snazy: done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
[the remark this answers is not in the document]
snazy: The "layout" or "meaning" of the keys depends on the caller. The implementation would not complain, if the same key is used for both `consistentPutAttachment` and `putAttachments`, but the guarantees of `consistentPutAttachment` would be off.

versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
```diff
@@ -455,4 +463,177 @@ private void checkForHashCollision(ColumnFamilyHandle cf, byte[] key)
...
+      Stream<AttachmentKey> keys) {
+    try {
+      List<AttachmentKey> keyList = keys.collect(Collectors.toList());
+      if (keyList.isEmpty()) {
+        return Stream.empty();
+      }
+      List<ColumnFamilyHandle> handles =
+          IntStream.range(0, keyList.size())
+              .mapToObj(x -> dbInstance.getCfAttachments())
+              .collect(Collectors.toList());
```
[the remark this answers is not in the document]
snazy: :+1:

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/adapter/serialize/ProtoSerialization.java
```diff
@@ -286,6 +293,134 @@ public static Key protoToKey(AdapterTypes.Key key) {
...
+        !contentId.contains("::"), "Attributes of an attachment key must not contain '::'");
+    return String.format("%s::", contentId);
+  }
+
+  private static String attachmentKeyAsString(
+      String contentId, String objectType, String objectId) {
+    Preconditions.checkState(
+        !contentId.contains("::") && !objectType.contains("::") && !objectId.contains("::"),
+        "Attributes of an attachment key must not contain '::'");
+    return String.format("%s::%s::%s", contentId, objectType, objectId);
```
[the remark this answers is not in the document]
snazy: :+1:

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractBinary.java
```diff
@@ -0,0 +1,232 @@
...
+import org.junit.jupiter.api.Test;
+import org.projectnessie.versioned.ContentAttachment;
+import org.projectnessie.versioned.ContentAttachment.Compression;
+import org.projectnessie.versioned.ContentAttachmentKey;
+import org.projectnessie.versioned.persist.adapter.DatabaseAdapter;
+
+/**
+ * Verifies that a big-ish number of keys, split across multiple commits works and the correct
+ * results are returned for the commit-log, keys, global-states. This test is especially useful to
+ * verify that the embedded and nested key-lists (think: full-key-lists in a commit-log-entry) work
```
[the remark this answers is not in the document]
snazy: oops

versioned/persist/tx/src/test/java/org/projectnessie/versioned/persist/tx/AbstractTxDatabaseAdapterTest.java
snazy: :+1:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -492,6 +520,156 @@ protected List<RefLog> doFetchPageFromRefLog(
...
+                        .getAttachments()
+                        .updateOne(
+                            Filters.eq(doc.get(ID_PROPERTY_NAME)),
+                            new Document("$set", doc),
+                            new UpdateOptions().upsert(true)),
+                    client.getAttachments()));
+
+    attachmentsKeys.forEach(
+        (cid, attachmentKeys) ->
+            verifyAcknowledged(
```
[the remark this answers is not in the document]
snazy: Yea - I thought about this "nosql problem" quite a bit.
I suspect, there's no "always correct" solution here. OTOH it shouldn't matter, beside there might be rows in the content-attachments table that are never referenced.
If a write fails, the whole commit operation fails and the attachments, if not the same in another commit, will be just garbage in the database.
Trying each key individually might be too slow.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -100,18 +109,28 @@ public void eraseRepo() {
...
   private Document toId(Hash id) {
+    return toId(id.asString());
+  }
+
+  private Document toId(String id) {
     Document idDoc = new Document();
     // Note: the order of `put` calls matters
     idDoc.put(ID_REPO_NAME, repositoryId);
-    idDoc.put(ID_HASH_NAME, id.asString());
+    idDoc.put(ID_HASH_NAME, id);
```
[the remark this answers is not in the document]
snazy: Wanted to keep the behavior the same for all database adapters. It's tricky to handle decomposed keys in some - especially in SQL :(

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -302,6 +304,44 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
   Stream<RefLog> refLog(Hash offset) throws RefLogNotFoundException;
 
+  /** Retrieve the known attachment keys for a content ID. */
+  Stream<ContentAttachmentKey> getAttachmentKeys(String contentId);
+
+  /**
+   * Retrieve the content attachments identified via {@code keys}. Attachments that do not exist are
+   * not returned.
+   */
+  Stream<ContentAttachment> getAttachments(Stream<ContentAttachmentKey> keys);
```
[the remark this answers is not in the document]
snazy: :+1: 
Added a note that it's up to the implementation whether the implemetnation "terminates" the input stream or not.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -492,6 +520,156 @@ protected List<RefLog> doFetchPageFromRefLog(
...
+    List<String> keys = doc.getList(DATA_PROPERTY_NAME, String.class);
+    return keys.stream().distinct().map(ProtoSerialization::attachmentKeyFromString);
+  }
+
+  @Override
+  protected Stream<Entry<AttachmentKey, AttachmentValue>> fetchAttachments(
+      Stream<AttachmentKey> keys) {
+    return fetchMappedPageGeneric(
+            client.getAttachments(),
+            keys.map(ProtoSerialization::attachmentKeyAsString).collect(Collectors.toList()),
```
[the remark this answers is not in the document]
snazy: Hm - Mongo doesn't seem to complain with even 10000 keys here. But... Dynamo does: `'requestItems.attachments.member.keys' failed to satisfy constraint: Member must have length less than or equal to 100`
Added generic paging.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: Makes sense. That was my understanding.. just wanted to confirm :)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: Oh, I meant using decomposed keys only in Mongo, where it would be more meaningful, I hope.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: re: failures - my impression was that the attachment key list is by content ID, not by commit, so I though losing keys might have wider implications than just one commit. WDYT?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: Adapters receive the full key as a java object... would it be too much inconvenience to have different attachment key representation in some adapters (i.e. Mongo)?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: Duh - we don't do batch-writes anywhere for Mongo. Shall we tackle that as a follow-up and fix it everywhere?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: What do you have in mind for Mongo-only? Only thing that makes sense to split out would be the content-ID. But on the read-path, wouldn't it be better to "distribute" attachment-reads - i.e. not consider the content-ID as part of the "sharding-key" (or however it's called)?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: sure :+1:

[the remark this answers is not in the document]
snazy: > re: failures - my impression was that the attachment key list is by content ID, not by commit, so I thought losing keys might have wider implications than just one commit. WDYT?

Yes, attachments are per-content-ID. But it's the other way around: the "list of attachment keys" won't have missing entries, but potentially "too many". But never have a missing attachment (unless it's been explicitly deleted).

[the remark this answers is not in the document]
snazy: ^ that's to "work around" the issue w/ no-sql databases / no (cheap-ish) "multi-key" transactions (aka "all or nothing")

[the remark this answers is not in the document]
snazy: Follow-up for the batch-writes here: #4159

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: If we use opaque IDs (current state) then the whole key body will be used for sharding. If we use a composite key we can defer sharding decisions to deployment time, I think.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -224,26 +246,32 @@ private <T, ID> T loadById(MongoCollection<Document> collection, ID id, Parser<T
...
-    List<Document> ids = hashes.stream().map(this::toId).collect(Collectors.toList());
+  private <T> List<T> fetchMappedPageGeneric(
+      MongoCollection<Document> collection, List<String> idList, Function<Document, T> mapper) {
+    List<Document> ids = idList.stream().map(this::toId).collect(Collectors.toList());
     FindIterable<Document> docs =
-        collection.find(Filters.in(ID_PROPERTY_NAME, ids)).limit(hashes.size());
+        collection.find(Filters.in(ID_PROPERTY_NAME, ids)).limit(idList.size());
 
-    HashMap<Hash, Document> loaded = new HashMap<>(hashes.size() * 4 / 3 + 1, 0.75f);
+    HashMap<String, Document> loaded = new HashMap<>(idList.size() * 4 / 3 + 1, 0.75f);
```
dimas-b, line 266: nit: `com.google.common.collect.Maps.newHashMapWithExpectedSize(N)`?

site/docs/try/configuration.md
```diff
@@ -86,6 +86,7 @@ The following configurations are advanced configurations to configure how Nessie
...
 | `nessie.version.store.advanced.commit-retries`                  | `Integer.MAX_VALUE` | `int`    | Sets the maximum retries for CAS-like operations.                                                                                                                                                                                     |
 | `nessie.version.store.advanced.nontx.parents-per-global-commit` | `50`                | `int`    | Sets the number of parent-global-commit-hashes.                                                                                                                                                                                       |
+| `nessie.version.store.advanced.attachment-keys-batch-size`      | `100`               | `int`    | Sets the number of content attachments that are written or retrieved at once. Some implementations may silently adopt this value to database limits or implementation requirements.                                                   |
```
dimas-b, line 89: `adopt` -> `adapt`?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -492,6 +520,156 @@ protected List<RefLog> doFetchPageFromRefLog(
...
+
+  @Override
+  protected Stream<AttachmentKey> fetchAttachmentKeys(String contentId) {
+    Document doc =
+        client.getAttachmentKeys().find(Filters.eq(ID_PROPERTY_NAME, toId(contentId))).first();
+    if (doc == null) {
+      return Stream.empty();
+    }
+    List<String> keys = doc.getList(DATA_PROPERTY_NAME, String.class);
+    return keys.stream().distinct().map(ProtoSerialization::attachmentKeyFromString);
```
dimas-b, line 650: Do we really need `.distinct()` since we use Mongo set operations for writing this data?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -492,6 +520,156 @@ protected List<RefLog> doFetchPageFromRefLog(
...
+  }
+
+  @Override
+  protected Stream<AttachmentKey> fetchAttachmentKeys(String contentId) {
+    Document doc =
+        client.getAttachmentKeys().find(Filters.eq(ID_PROPERTY_NAME, toId(contentId))).first();
+    if (doc == null) {
+      return Stream.empty();
+    }
+    List<String> keys = doc.getList(DATA_PROPERTY_NAME, String.class);
```
dimas-b, line 649: Would it be worth trying to fetch the inner property (Mongo projection) and rely on Mongo's `FindIterable` to do the streaming?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: added that as a separate commit

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: I doubt

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -68,9 +76,13 @@
   private static final String ID_PROPERTY_NAME = "_id";
   private static final String ID_REPO_NAME = "repo";
   private static final String ID_HASH_NAME = "hash";
+  private static final String ID_ATT_CONTENT_ID = "cid";
+  private static final String ID_ATT_OBJECT_TYPE = "type";
+  private static final String ID_ATT_OBJECT_ID = "oid";
```
dimas-b, line 81: nit: I believe "attribute" is usually abbreviated as "attr" :)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -100,20 +112,52 @@ public void eraseRepo() {
...
     client.getRefLog().deleteMany(idPrefixFilter);
+    client.getAttachments().deleteMany(idPrefixFilter);
+    client.getAttachmentKeys().deleteMany(idPrefixFilter);
   }
 
   private Document toId(Hash id) {
+    return toId(id.asString());
+  }
+
+  private Document toId(String id) {
```
dimas-b, line 123: Do we need this change anymore? It looks like `toId(String)` is called only from `toId(Hash)` :thinking:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -100,20 +112,52 @@ public void eraseRepo() {
...
+
+  private Document toIdAttachmentKeyContentId(String contentId) {
+    Document idDoc = new Document();
+    // Note: the order of `put` calls matters
+    idDoc.put(ID_REPO_NAME, repositoryId);
+    idDoc.put(ID_ATT_CONTENT_ID, contentId);
+    return idDoc;
+  }
+
+  private List<Document> toIdsFromStrings(Collection<Hash> ids) {
```
dimas-b, line 149: hmmm, the name has "from Strings" but the arg is a collection of `Hash`s?

dimas-b (dismissed):

nastra:
will review the PR tomorrow, so please don't merge ;)

snazy:
Changes since the last review round:
* Attachment values persisted with "format" _and_ "compression" attributes
* Attachment values field "object ID", which represents e.g. the snapshot ID or schema ID to handle the case when the same content ID has different e.g. schemas or partition-specs or sort-orders on different branches but with the same schema-ID/partition-spec-ID/sort-order-ID

Attachment data serialization format can be JSON or CBOR - the latter wastes less space.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -291,6 +293,56 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+   *
+   * <p>Whether the input stream is "terminated" or "fluently" mapped, can vary between
+   * implementations.
+   */
+  Stream<ContentAttachment> mapToAttachment(Stream<ContentAttachmentKey> keys);
+
+  /**
+   * Consistent put-attachment operation.
+   *
+   * <p>Either a "put-if-absent", if {@code expectedHash} is empty or a compare-and-swap based on
```
dimas-b, line 317: `expectedHash` -> `expectedVersion`?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -291,6 +293,56 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+   * the value of {@code expectedVersion}.
+   *
+   * <p>Note: this method uses conditional put operations, unlike {@link #putAttachments(Stream)}.
+   * Users should not use this method and {@link #putAttachments(Stream)} for the same keys.
+   *
+   * @param attachment the attachment to write
+   * @param expectedVersion indicator for put-if-absent or the expected value on an existing item
+   */
+  boolean consistentPutAttachment(ContentAttachment attachment, Optional<String> expectedVersion);
```
dimas-b, line 326: Just wondering if we may need change log for attachments? Since the key is probably going to be frozen into a commit, could it be necessary for us to get previous versions of attachments when/after changes are made? Since we track top-level content changes in commits, should we also track attachment changes for the same of having full history?

dimas-b (approved):
Sorry, some last minute comments :) ... but minor / deferable.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
snazy: Not sure TBH. We do not need this consisten-put at the moment. It was intended to potentially allow sequence-numbers per content-id - like column-IDs and such, if we want to at some point.

PR #3294, opened 2022-02-09
Add nessie notification and webhook proposal.

hpolloni:


design/notifications.md
```diff
@@ -0,0 +1,102 @@
...
+Every time a user:
+- Creates a reference (branch or tag)
+- Deletes a reference (branch or tag)
+- Commits
+- Assigns reference to hash (force push)
+
+We will fire a notification. Each notification is associated with a particular event. We choose not to include content specific events (UpdateContent or DeleteContent) because these should happen at commit time. Instead a user will need to look into the list of operations performed during a commit event to filter for the event they are looking for. However, we support this use case through the use of subscriptions.
+
+## Notification SPI
+Nessie events are sent to a common "listener" implementation. Every time someone commits, creates a reference, deletes a reference or assigns a reference the `notifyEvent` method is called. Anyone that wishes to get notifications about a nessie event must implement the `NessieEventListener` interface.
```
dimas-b, line 14: How are listener implementations meant to be installed / discovered by the Nessie Server?

hpolloni: There's a single listener that delegates to other listeners. That event listener is injected (similar as the Authorizer is being injected). I guess in that way, listeners are not discovered, but they are registered on demand.

dimas-b: Do you mean the listener implementations will be packaged in non-Nessie jars and combined with the Nessie Server at deployment time (post-build)?

hpolloni:
Talked offline with Dmitri. I will be sending sending a separate proposal for webhooks and leave the notifications design as an implementation detail.

PR #3306, opened 2022-02-10
Use docker hostname for dynamodb integ test

hpolloni:
Now that Docker Desktop is no longer free. Some of us are migrating to Minikube. Minikube doesn't expose ports in localhost, so defaulting to "localhost" seems to assume that the only docker engine is Docker Desktop. I still need to test if this change works with other docker engines (like the linux one and docker desktop)


dimas-b (commented):
CI runs on Linux... let's see if it passes

dimas-b (approved):

hpolloni:
Confirmed that it works on my linux box.

snazy (approved):

PR #3316, opened 2022-02-11
parallel maven builds in workflows

XN137:
- we convert `-B` to `--batch-mode` in order to prefer meaningful argument names in scripts/code.
- we drop redundant `--file pom.xml`
- we use `--threads 1C` in CI mvn builds, which builds modules in parallel (should not impact test execution) which speeds up local builds for me by 33 % (I am aware we should use `mvnd` locally but this is not used in CI)

before:
```
~/code/nessie$ for i in {1..3}; do time bash -c './mvnw --batch-mode package -DskipTests >/dev/null 2>&1; echo $?'; done
0

real    1m29,632s
user    4m32,614s
sys     0m5,596s
0

real    1m34,567s
user    4m47,960s
sys     0m6,088s
0

real    1m36,726s
user    4m54,755s
sys     0m6,221s
```

after:
```
~/code/nessie$ for i in {1..3}; do time bash -c './mvnw --batch-mode --threads 1C package -DskipTests >/dev/null 2>&1; echo $?'; done
0

real    0m55,158s
user    4m49,586s
sys     0m6,825s
0

real    1m3,020s
user    5m33,859s
sys     0m8,042s
0

real    1m3,198s
user    5m35,195s
sys     0m8,650s
```

note that this warning will show up in the mvn output now:
```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in Nessie - Content Generator:
[WARNING] org.codehaus.mojo:tidy-maven-plugin:1.1.0
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
```
but in https://github.com/mojohaus/tidy-maven-plugin/commit/95044bf04604365ef663235fa170ef3543f50bf9 you can see that the plugin was already threadsafe (there just hasnt been a release since declaring it)

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -121,7 +121,7 @@ jobs:
           BUILD_ARGS="${BUILD_ARGS} -Dmaven.deploy.skip -Dgpg.skip"
         fi
 
-        ./mvnw -B ${GOALS} --file pom.xml ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
+        ./mvnw --batch-mode --threads 1C ${GOALS} ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
 
         # Build the native-image, it's not released to Maven Central anyway
         PROFILES="-Pnative -Pdocker"
@@ -129,7 +129,7 @@ jobs:
         if [[ -n ${ACT} ]] ; then
           BUILD_ARGS="${BUILD_ARGS} -Dquarkus.native.remote-container-build=true -Dquarkus.native.container-build=false -Dquarkus.native.native-image-xmx=6g"
         fi
-        ./mvnw -B install --file pom.xml ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
+        ./mvnw --batch-mode --threads 1C install ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
 
         # Add version to the openapi file name
         cp model/target/classes/META-INF/openapi/openapi.yaml model/target/nessie-openapi-${RELEASE_VERSION}.yaml
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -121,7 +121,7 @@ jobs:
           BUILD_ARGS="${BUILD_ARGS} -Dmaven.deploy.skip -Dgpg.skip"
         fi
 
-        ./mvnw --batch-mode --threads 1C ${GOALS} ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
+        ./mvnw --batch-mode ${GOALS} ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
 
         # Build the native-image, it's not released to Maven Central anyway
         PROFILES="-Pnative -Pdocker"
@@ -129,7 +129,7 @@ jobs:
         if [[ -n ${ACT} ]] ; then
           BUILD_ARGS="${BUILD_ARGS} -Dquarkus.native.remote-container-build=true -Dquarkus.native.container-build=false -Dquarkus.native.native-image-xmx=6g"
         fi
-        ./mvnw --batch-mode --threads 1C install ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
+        ./mvnw --batch-mode install ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
 
         # Add version to the openapi file name
         cp model/target/classes/META-INF/openapi/openapi.yaml model/target/nessie-openapi-${RELEASE_VERSION}.yaml
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,11 +34,11 @@ on the new feature or improvement.
 ### Maven tips
 
 A `./mvnw --threads 1C clean install` runs basically "everything" except release/deployment stuff. This is often
-not necessary. Some tips:
+not necessary. Use one of these parameters to speed things up:
 
-* `./mvnw --threads 1C package -Dquickly` Just compiles code, no tests, does not build code under `ui/` and `perftest/`.
-* `./mvnw --threads 1C package -DskipTests` Compiles everything, runs no tests.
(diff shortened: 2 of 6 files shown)


XN137:
marking as ready for review since CI + tests passed.

java/maven workflow before:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  31:59 min
[INFO] Finished at: 2022-02-11T14:54:23Z
[INFO] ------------------------------------------------------------------------
```
https://github.com/projectnessie/nessie/runs/5157348827?check_suite_focus=true

java/maven workflow after:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  28:52 min (Wall Clock)
[INFO] Finished at: 2022-02-11T15:38:18Z
[INFO] ------------------------------------------------------------------------
```
https://github.com/projectnessie/nessie/runs/5157978678?check_suite_focus=true

CONTRIBUTING.md
```diff
@@ -33,12 +33,12 @@ on the new feature or improvement.
...
 ### Maven tips
 
-A `./mvnw clean install` runs basically "everything" except release/deployment stuff. This is often
+A `./mvnw --threads 1C clean install` runs basically "everything" except release/deployment stuff. This is often
 not necessary. Some tips:
 
-* `./mvnw -Dquickly` Just compiles code, no tests, does not build code under `ui/` and `perftest/`.
-* `./mvnw -DskipTests` Compiles everything, runs no tests.
-* `./mvnw -DskipITs` Compiles everything, runs unit tests, but no integration tests.
+* `./mvnw --threads 1C package -Dquickly` Just compiles code, no tests, does not build code under `ui/` and `perftest/`.
```
XN137, line 39: using `package` here to just build the artifacts, not install them into the local m2 repo

CONTRIBUTING.md
```diff
@@ -33,12 +33,12 @@ on the new feature or improvement.
 
 ### Maven tips
 
-A `./mvnw clean install` runs basically "everything" except release/deployment stuff. This is often
+A `./mvnw --threads 1C clean install` runs basically "everything" except release/deployment stuff. This is often
```
XN137, line 36: i found no maven config setting to apply this universally, so it has to be added everywhere it seems

nastra (dismissed):
LGTM

CONTRIBUTING.md
dimas-b: I'm not sure about recommending `package` as the default goal. I usually do `install -Dquickly` and then run individual tests in IDE... but I'm not sure if that workflow will work with `package`. WDYT?

dimas-b: It's a matter of expectation I guess... if not running `install`, I think IDE will use over artifacts for dependencies from the local repo :thinking: I'm not 100% sure, though.

dimas-b: What I mean to say is that it might be preferable to leave the goal out of that example (as before) and let the use choose the goal to use with the "speed up" options.

XN137: i added a goal because as is the mvnw calls were invalid... so maybe we should just be explaining the speedup flags here, and leave out the `./mvnw <goal>` part completely?

dimas-b: +1 to that from my POV :)

XN137: ok i've done that in the new commit

dimas-b (dismissed):

.github/workflows/release-publish.yml
```diff
@@ -121,15 +121,15 @@ jobs:
...
+        ./mvnw --batch-mode --threads 1C ${GOALS} ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
 
         # Build the native-image, it's not released to Maven Central anyway
         PROFILES="-Pnative -Pdocker"
         BUILD_ARGS="-DskipTests"
         if [[ -n ${ACT} ]] ; then
           BUILD_ARGS="${BUILD_ARGS} -Dquarkus.native.remote-container-build=true -Dquarkus.native.container-build=false -Dquarkus.native.native-image-xmx=6g"
         fi
-        ./mvnw -B install --file pom.xml ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
+        ./mvnw --batch-mode --threads 1C install ${PROFILES} ${BUILD_ARGS} -pl servers/quarkus-server
```
snazy, line 132: `--threads 1C` is superfluous here, it's only one project

.github/workflows/release-publish.yml
```diff
@@ -121,15 +121,15 @@ jobs:
           BUILD_ARGS="${BUILD_ARGS} -Dmaven.deploy.skip -Dgpg.skip"
         fi
 
-        ./mvnw -B ${GOALS} --file pom.xml ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
+        ./mvnw --batch-mode --threads 1C ${GOALS} ${PROFILES} ${NEXUS_ARGS} ${BUILD_ARGS}
```
snazy, line 124: Is the `nexus-staging-maven-plugin` thread safe?
I know it's _marked_ as thread-safe, but I'm not sure TBH and would prefer not to risk anything for the Nessie release process.

snazy (commented):
Overall LGTM, just would like to not risk changing the actual Nessie release-deploy multi-threaded.
Note: that the snapshot deployment and uses a different Maven plugin.

.github/workflows/release-publish.yml
XN137: i think being marked as thread-safe is the strongest indicator of supporting it correctly there is... but we can remove the parameter here just for the release workflow

snazy (approved):

PR #3319, opened 2022-02-11
Make Spark SQL Extensions work with Spark 3.1 and 3.2.1

nastra:


clients/spark32-extensions/pom.xml
```diff
@@ -0,0 +1,346 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie-clients</artifactId>
+    <version>0.20.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nessie-spark32-extensions</artifactId>
```
snazy, line 28: Curious - can this be called `nessie-spark-3.2-extensions`?

nastra:
> Hm. That code duplication isn't particularly amazing - but I see why you did it.
> 
> The actual differences are rather tiny, but intrusive.
> 
> Maybe we can get around this with a different approach:
> 
> * Create folders `clients/spark-extensions-common/src/main/scala/org/apache/spark/sql/catalyst/plans/logical` and `clients/spark-extensions-common/src/main/scala/org/apache/spark/sql/execution/datasources/v2`
> * Move the existing `.scala` files for these packages from clients/spark-extension/src/main/scala... to the new folders
> * Remove the `.../plans/logical` and `.../datasources/v2`  directories in `spark-extensions` and `spark32-extensions`
> * Create symlinks from `spark-extensions` and `spark32-extensions` to the new folders
> * Let the "common" `*Command` and `*Exec` classes implement taits - maybe called `NessieCommandTrait` and `NessieExecTrait` and implement those specifically in spark-exentions + spark32-extensions
> 
> WDYT?

Let's do this as a follow-up

clients/spark32-extensions/pom.xml
nastra: updated

nastra:
opened https://github.com/projectnessie/nessie/issues/3383

clients/spark-extensions/pom.xml
```diff
@@ -41,10 +41,37 @@
...
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-spark-extensions-base</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-spark-extensions-base</artifactId>
+      <version>${project.version}</version>
+      <classifier>tests</classifier>
```
snazy, line 53: tests for non-test scope?

clients/spark-extensions-base/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>tidy-maven-plugin</artifactId>
```
snazy, line 95: Why tidy-maven-plugin here?

clients/spark-3.2-extensions/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-spark-extensions-grammar</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-spark-extensions-base</artifactId>
+      <version>${project.version}</version>
+      <exclusions>
```
snazy, line 48: Q: Is this really needed? I don't see any Spark dependencies that are not in provided scope.
(Same for the exclusion below)

clients/spark-3.2-extensions/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+        <exclusion>
+          <groupId>org.apache.spark</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-core_2.12</artifactId>
```
snazy, line 70: Wonder if all spark dependencies should actually be in provided scope, as a Spark extension runs in a Spark runtime, which should already have the right dependencies. (Thinking of patch-version mismatches of the version declared by Nessie and the patch-version running in the runtime.)

clients/spark-3.2-extensions/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>tidy-maven-plugin</artifactId>
```
snazy, line 157: THink I asked elsewhere about tidy-maven-plugin

clients/spark-3.2-extensions/pom.xml
```diff
@@ -0,0 +1,382 @@
...
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>spark-test-java</id>
+      <activation>
+        <jdk>[16,20)</jdk>
```
snazy, line 299: Curious: Have you re-tested this one?

clients/spark-extensions/pom.xml
nastra: fixed

clients/spark-3.2-extensions/pom.xml
nastra: it's not needed anymore. was a leftover from a previous approach

clients/spark-extensions-base/pom.xml
nastra: just a copy/paste from the spark-extensions pom.xml

clients/spark-3.2-extensions/pom.xml
nastra: just a copy/paste from the spark-extensions pom.xml

clients/spark-3.2-extensions/pom.xml
nastra: no, just a copy/paste from the spark-extensions pom.xml

clients/spark-3.2-extensions/pom.xml
nastra: done

snazy (approved):

tomekl007:
My findings. We have two sources of duplication. 

**The first one** is in the `plans.logical` package of both `spark-extensions` and `spark-3.2-extensions`.
All the `*Commands` are case classes that are built with the types imported directly from `spark_catalyst` engine, fore example:
`import org.apache.spark.sql.types.{DataTypes, Metadata, StructField, StructType}`
Here is an example `ShowReferenceCommand` from `spark-3.2`:
```

case class ShowReferenceCommand(val catalog: Option[String])
    extends LeafCommand {
  override lazy val output: Seq[Attribute] = new StructType(
    Array[StructField](
      StructField(
        "refType",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      ),
      StructField(
        "name",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      ),
      StructField(
        "hash",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      )
    )
  ).toAttributes

  override def simpleString(maxFields: Int): String = {
    s"ShowReference"
  }
}
```
and this is from `3.1`:
```
case class AssignReferenceCommand(
    reference: String,
    isBranch: Boolean,
    toRefName: Option[String],
    toHash: Option[String],
    catalog: Option[String]
) extends Command {
  override lazy val output: Seq[Attribute] = new StructType(
    Array[StructField](
      StructField(
        "refType",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      ),
      StructField(
        "name",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      ),
      StructField(
        "hash",
        DataTypes.StringType,
        nullable = false,
        Metadata.empty
      )
    )
  ).toAttributes

  override def simpleString(maxFields: Int): String = {
    s"AssignReference ${reference}"
  }
}
```
They look very similar but all the types used there are from a different library (spark 3.1 vs spark 3.2) so we cannot make it common. We could extract a trait for each command, for example:
```
trait AssignReferenceCommandTrait {
  def reference: String
  def isBranch: Boolean
  def toRefName: Option[String]
  def toHash: Option[String]
  def catalog: Option[String]
}
```
and make the commands extend it, but it will not remove any duplication. 

**The second source of duplication:**
The duplication in `*Exec` classes is already handled by the `spark-extensions-base` module and all `Base*` classes.

So to summarize, at this point I don't see a clear way to reduce duplication in those modules. Also, it is worth noting that the duplications come from the case classes that are anemic (without logic).
Any thoughts, comments? (@snazy @nastra)

snazy:
I assume, the comment should have gone to #3383 ? ;)

tomekl007:
yes, sorry. Re-posted there :)

PR #3320, opened 2022-02-11
use official maven wrapper

XN137:
see https://github.com/takari/maven-wrapper
there is now an official maven wrapper: https://maven.apache.org/wrapper/index.html

i only modified the pom then ran these commands to update the files:
```
./mvnw wrapper:wrapper -Dmaven=3.8.4 -Dtype=bin
./mvnw wrapper:wrapper -Dmaven=3.8.4 -Dtype=source
```
(the 2nd updates `MavenWrapperDownloader.java` which we may not actually need?).

i assume the license header in the affected files was accidentally from Dremio and keeping the original is better, wdyt?

### DIFF
--- a/.mvn/mvnd.properties
+++ b/.mvn/mvnd.properties
@@ -1,3 +1,19 @@
+#
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 mvnd.jvmArgs = \
   --add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED \
   --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED \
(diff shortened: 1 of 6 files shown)


mvnw.cmd
```diff
@@ -1,179 +1,188 @@
...
+@REM MAVEN_OPTS - parameters passed to the Java VM when running Maven
+@REM     e.g. to debug Maven itself, use
+@REM set MAVEN_OPTS=-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000
+@REM MAVEN_SKIP_RC - flag to disable loading of mavenrc files
+@REM ----------------------------------------------------------------------------
+
+@REM Begin all REM lines with '@' in case MAVEN_BATCH_ECHO is 'on'
+@echo off
+@REM set title of command window
+title %0
```
dimas-b, line 39: Do we need the Windows script at all? Will Nessie even work on WIndows?

XN137: i am fine with removing it if we dont need it (though it would be an additional manual step most likely when running the plugin goal in the future)

dimas-b: I did not realise the file was added automatically :facepalm: .... I'm fine to keep it :+1:

.mvn/mvnd.properties
```diff
@@ -1,19 +1,3 @@
-#
```
snazy, line 1: This is for mvnd, so a rather custom config file, not something that's been generated.

snazy (commented):
LGTM - just revert the change to mvnd.properties

.mvn/mvnd.properties
XN137: just to make sure: i should just revert the change in this file, but no longer making a license header check below `.mvn` folder you agree with?

snazy: yep

mvnw.cmd
snazy: We'll just point people to @XN137 when users complain about Windows builds ;) ;) ;)

snazy (approved):

PR #3322, opened 2022-02-13
Properly close RocksDB in Quarkus

snazy:


servers/quarkus-server/src/main/java/org/projectnessie/server/providers/RocksDbSource.java
```diff
@@ -0,0 +1,43 @@
...
+  RocksDbInstance createRocksDbInstance() {
+    RocksDbInstance rocksDbInstance = new RocksDbInstance();
+    rocksDbInstance.configure(
+        ImmutableRocksDbConfig.builder().dbPath(rocksConfig.getDbPath()).build());
+    rocksDbInstance.initialize();
+    return rocksDbInstance;
+  }
+
+  void close(@Disposes RocksDbInstance rocksDbInstance) {
+    rocksDbInstance.close();
```
nastra, line 41: maybe add a null check in case it never actually got initialized

snazy: How shall that happen? That instance is inhjected here by the container.

nastra (approved):

PR #3323, opened 2022-02-13
Framework for compatibility and upgrade checks

snazy:
Provides three compatibility-check approaches:
* Running older Nessie client versions against the current in-tree Nessie server
* Running the current in-tree Nessie client against older Nessie servers
* Upgrade scenarios:
  * Start Nessie using the "oldest" version, use the "oldest" Nessie client, run tests, stop server
  * Start Nessie using the "next" version, use the "next" Nessie client, run tests, stop server, and so on
  * Uses a database adapter that persists (RockdDB), so the "oldest" version starts with an emptry
    repository, each "next" version starts with the state of the "previous" version

Implementation:
* Uses JUnit 5 SPI functionality to "duplicate" test classes (one per Nessie version to be tested)
  using a JUnit engine implementation, tests are run using "standard" and well-known JUnit 5.
* Uses JUnit 5 extensions to properly populate and use current/old Nessie server/client instances and
  control their lifecycle - plus "helper" stuff like cached class loaders and such.

This change does not include required compatibility and upgrade checks, but a small set of tests to illustrate the behavior.


tools/pom.xml
```diff
@@ -34,5 +34,6 @@
     <module>apprunner-common</module>
     <module>apprunner-maven-plugin</module>
     <module>content-generator</module>
+    <module>compatibility</module>
```
nastra, line 37: putting this under `tools` seems rather weird. maybe just put this directly under root?

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
```diff
@@ -0,0 +1,145 @@
...
+
+    jerseyTest =
+        new JerseyTest() {
+          @Override
+          protected Application configure() {
+            ResourceConfig config = new ResourceConfig();
+            config.register(RestConfigResource.class);
+            config.register(RestTreeResource.class);
+            config.register(RestContentResource.class);
+            withClass("org.projectnessie.services.rest.RestDiffResource", config::register);
```
nastra, line 77: just thinking whether it would make sense to make it clearer that these classes are `new` classes that aren't available in the old version (whatever old means)

nastra: I'm thinking about separating those `new` classes out, something like:

```
private List<Class<?>> classesSinceX() {
    return Arrays.asList(withClass("org.projectnessie.services.rest.RestDiffResource"),withClass("org.projectnessie.services.rest.RestRefLogResource"), ...);
  }
```

and then doing something like `classesSinceX().forEach(config::register);`

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/DatabaseAdapters.java
```diff
@@ -0,0 +1,143 @@
...
+    return doBuild(builder, build);
+  }
+
+  private static DatabaseAdapter buildWithContentVariantSupplier(
+      Builder<DatabaseAdapterConfig, DatabaseAdapterConfig, DatabaseConnectionProvider<?>> builder)
+      throws NoSuchMethodException {
+    Method build =
+        DatabaseAdapterFactory.Builder.class.getMethod("build", ContentVariantSupplier.class);
+    return doBuild(
+        builder, build, new GenericContentVariantSupplier(new TableCommitMetaStoreWorker()));
```
nastra, line 130: should this be GenericContentVariantSupplier<>(...)?

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/ServerConfigExtension.java
```diff
@@ -0,0 +1,48 @@
...
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.Default;
+import javax.enterprise.inject.spi.AfterBeanDiscovery;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.Extension;
+import org.projectnessie.services.config.ServerConfig;
+
+public class ServerConfigExtension implements Extension {
```
nastra, line 26: seems like a1:1 copy/paste from the existing ServerConfigExtension :(

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/AuthorizerExtension.java
```diff
@@ -0,0 +1,58 @@
...
+import javax.enterprise.inject.Default;
+import javax.enterprise.inject.spi.AfterBeanDiscovery;
+import javax.enterprise.inject.spi.BeanManager;
+import javax.enterprise.inject.spi.Extension;
+import org.projectnessie.services.authz.AbstractBatchAccessChecker;
+import org.projectnessie.services.authz.AccessContext;
+import org.projectnessie.services.authz.Authorizer;
+import org.projectnessie.services.authz.BatchAccessChecker;
+
+public class AuthorizerExtension implements Extension {
```
nastra, line 30: 1:1 copy/paste from the existing AuthorizerExtension :(

tools/compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/TemporaryDirectory.java
```diff
@@ -0,0 +1,109 @@
...
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import org.junit.jupiter.api.extension.ExtensionContext.Store.CloseableResource;
+
+final class TemporaryDirectory implements CloseableResource {
```
nastra, line 32: can you elaborate why this is needed vs Junit5's TempDir? I think I know why but would be good to have some comments

tools/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
```diff
@@ -0,0 +1,87 @@
...
+        .isEqualTo(branch.getName());
+
+    LogResponse commitLog = api.getCommitLog().refName(branch.getName()).get();
+    assertThat(commitLog.getLogEntries())
+        .hasSize(1)
+        .map(LogEntry::getCommitMeta)
+        .map(CommitMeta::getMessage)
+        .containsExactly(commitMessage);
+  }
+}
```
nastra, line 87: I'm assuming we'll eventually add many more tests here?

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/ServerConfigExtension.java
snazy: yea - it just cannot use the other classes due to class loader isolation constraints :(

tools/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
snazy: yes (last sentence in the PR description)

tools/pom.xml
snazy: +1

tools/compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
snazy: Not a fan of `classesSinceX` (and X being a version number) - that would make the code here very very version specific (it's already a PITA that that module's needed due to class loader isolation requirements).

The alternative to this module would have been a lot of reflection based code - that's an even more a PITA.

snazy:
Pushed 3 more commits - one is the move to the top-level directory

nastra (approved):
I think it's ok to merge now and do follow-ups if we encounter issues. Having this merged and running CI is more valuable than holding off and making things perfect

PR #3343, opened 2022-02-14
use maven wrapper for jackson ITs

XN137:


.github/workflows/main.yml
```diff
@@ -119,7 +119,7 @@ jobs:
         with:
           java-version: ${{ matrix.java-version }}
       - name: Jackson Integration Tests ${{ matrix.target-library }}
-        run: mvn verify -pl :nessie-client -am -Pjackson-tests -Djackson.test.version=${{ matrix.jackson-version }} -Dtest.log.level=WARN
+        run: ./mvnw --batch-mode verify -pl :nessie-client -am -Pjackson-tests -Djackson.test.version=${{ matrix.jackson-version }} -Dtest.log.level=WARN
```
XN137, line 122: all maven usage should go through `mvnw` (which is already being used earlier in this workflow as well)

snazy (approved):

PR #3344, opened 2022-02-14
workaround intellj bug with shaded antlr

XN137:
this fixes the following compile problem in intellij:
intellj_shading_problem

the problem there is that intellj does not replace the original source with the shaded jar contents.... thus on compile it cant find some of the classes (that have been relocated in the shaded jar).
this works fine when using maven from the command line.

see also https://youtrack.jetbrains.com/issue/IDEA-126596 and https://youtrack.jetbrains.com/issue/IDEA-93855

Note that some of the workarounds mentioned in those tickets avoid the "class not found" problem, but not the "incompatible type" problem.

test to confirm the published pom is not negatively affected:
```
chris@xanadu:~/code/nessie$ rm ~/before.pom ~/after.pom; rm -rf ~/.m2/repository/org/projectnessie && git checkout main && ./mvnw --threads 0.5C clean install -DskipITs >/dev/null 2>&1 && cp -v ~/.m2/repository/org/projectnessie/nessie-spark-extensions/0.20.1-SNAPSHOT/nessie-spark-extensions-0.20.1-SNAPSHOT.pom ~/before.pom; rm -rf ~/.m2/repository/org/projectnessie && git checkout antlr-intellj-workaround && ./mvnw --threads 0.5C clean install -DskipITs >/dev/null 2>&1 && cp -v ~/.m2/repository/org/projectnessie/nessie-spark-extensions/0.20.1-SNAPSHOT/nessie-spark-extensions-0.20.1-SNAPSHOT.pom ~/after.pom; diff ~/before.pom ~/after.pom
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
'/home/chris/.m2/repository/org/projectnessie/nessie-spark-extensions/0.20.1-SNAPSHOT/nessie-spark-extensions-0.20.1-SNAPSHOT.pom' -> '/home/chris/before.pom'
Switched to branch 'antlr-intellj-workaround'
'/home/chris/.m2/repository/org/projectnessie/nessie-spark-extensions/0.20.1-SNAPSHOT/nessie-spark-extensions-0.20.1-SNAPSHOT.pom' -> '/home/chris/after.pom'
192a193,211
>     <profile>
>       <id>intellij-workaround</id>
>       <dependencies>
>         <dependency>
>           <groupId>org.projectnessie</groupId>
>           <artifactId>nessie-spark-extensions-grammar</artifactId>
>           <version>${project.version}</version>
>           <scope>runtime</scope>
>         </dependency>
>         <dependency>
>           <groupId>${project.groupId}</groupId>
>           <artifactId>nessie-spark-extensions-grammar</artifactId>
>           <version>FAKE-VERSION</version>
>           <classifier>FAKE-CLASSIFIER</classifier>
>           <scope>system</scope>
>           <systemPath>${project.basedir}/../spark-antlr-grammar/target/nessie-spark-extensions-grammar-${project.version}.jar</systemPath>
>         </dependency>
>       </dependencies>
>     </profile
```

### DIFF
--- a/clients/spark-extensions/pom.xml
+++ b/clients/spark-extensions/pom.xml
@@ -286,5 +286,41 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>intellij-workaround</id>
+      <activation>
+        <property>
+          <name>idea.version</name>
+        </property>
+      </activation>
+      <!--
+        IntelliJ IDEA gets confused when a module builds a shaded jar.
+        It does not replace the original source with the shaded jar contents.
+        To avoid compile errors due to relocations we have to use a workaround here.
+
+        Partly taken from https://stackoverflow.com/a/59654377
+      -->
+      <dependencies>
+        <dependency>
+          <groupId>org.projectnessie</groupId>
+          <artifactId>nessie-spark-extensions-grammar</artifactId>
+          <version>${project.version}</version>
+          <!--
+          We need this dependency to build the shaded jar beforehand, but the runtime scope
+          makes it so intellij compilation does not use the original source (which references
+          non-relocated antlr classes).
+          -->
+          <scope>runtime</scope>
+        </dependency>
+        <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>nessie-spark-extensions-grammar</artifactId>
+          <version>FAKE-VERSION</version>
+          <classifier>FAKE-CLASSIFIER</classifier>
+          <scope>system</scope>
+          <systemPath>${project.basedir}/../spark-antlr-grammar/target/nessie-spark-extensions-grammar-${project.version}.jar</systemPath>
+        </dependency>
+      </dependencies>
+    </profile>
   </profiles>
 </project>


nastra (approved):
I tried this out locally and this works, thanks for fixing!

rymurr:
can we determine if we are running under intellij and activate by default? eg w/ a vm arg or an env variaable? I think that can be done by eclipse...not sure if intellij can

snazy:
> can we determine if we are running under intellij and activate by default? eg w/ a vm arg or an env variaable? I think that can be done by eclipse...not sure if intellij can

Should be doable - @XN137 can you look into the files in `.idea`, whether there's something that can be checked into the codebase?

XN137:
> can we determine if we are running under intellij and activate by default? eg w/ a vm arg or an env variaable? I think that can be done by eclipse...not sure if intellij can

i found no easy way to do this.
as mentioned in the PR description universally setting a property in the intellj maven configuration does not work.
similarly setting an env variable there does not work.

my guess is that those are not injected when loading/refreshing a maven project (which is when the profile needs to be active).

> can you look into the files in .idea, whether there's something that can be checked into the codebase?

if we look for a file that is checked into the git repo, then the profile would activate in CI as well, which we dont want.
if we look for a file that exists locally but is not checked into git, this would allow to activate the profile locally but both inside and outside of intellj (i.e. any mvn command running inside the repo folder would also find the file and activate the profile).
This also does not seem to be what we want.

i explored the env var idea a bit more and if we start intellj itself with env var INTELLIJ_WORKAROUND=true set, then the profile actives correctly during the maven loading phase, so i've commited that.

XN137:
after some more searching i found an automatically available property we can rely on (somehow i did not find it in my earlier searches)

nastra (approved):

PR #3345, opened 2022-02-14
minor code cleanups

XN137:
WIP

a few LHF simplifications / unifications found via intellj static analysis


snazy:
Can you split those up into smaller PRs each addressing individual issues? That'll make it easier and less prone to merge conflicts.

XN137:
> Can you split those up into smaller PRs each addressing individual issues? That'll make it easier and less prone to merge conflicts.

yes i can do that but i am unsure what a good criteria to split by is, do you have a suggestion?

if i do it be type of change we would have:
- avoid uncessary toString calls
- avoid using deprecated junit MethodOrderer
- avoid jetbrains Nullable
- use more appropriate Assertion methods
- prefer ThreadLocal.withInitial

and 2-3 more like this, is that preferable?

PR #3347, opened 2022-02-14
Tighten python requirements

XN137:
- remove outdated `setup.py` test mechanism
- remove unused dependencies
- version pin all dependencies
- stricter mypy/pylint checks

### DIFF
--- a/python/pynessie/auth/aws.py
+++ b/python/pynessie/auth/aws.py
@@ -17,7 +17,7 @@
 """Use AWS4Auth and botocore to fetch credentials and sign requests."""
 from botocore.credentials import get_credentials
 from botocore.session import Session
-from requests_aws4auth import AWS4Auth  # type: ignore
+from requests_aws4auth import AWS4Auth
 
 
 def setup_aws_auth(region: str, profile: str = None) -> AWS4Auth:
--- a/python/pynessie/auth/aws.py
+++ b/python/pynessie/auth/aws.py
@@ -15,8 +15,8 @@
 # limitations under the License.
 #
 """Use AWS4Auth and botocore to fetch credentials and sign requests."""
-from botocore.credentials import get_credentials  # type: ignore
-from botocore.session import Session  # type: ignore
+from botocore.credentials import get_credentials
+from botocore.session import Session
 from requests_aws4auth import AWS4Auth  # type: ignore
 
 
--- a/python/pynessie/auth/aws.py
+++ b/python/pynessie/auth/aws.py
@@ -15,9 +15,9 @@
 # limitations under the License.
 #
 """Use AWS4Auth and botocore to fetch credentials and sign requests."""
-from botocore.credentials import get_credentials
-from botocore.session import Session
-from requests_aws4auth import AWS4Auth
+from botocore.credentials import get_credentials  # type: ignore
+from botocore.session import Session  # type: ignore
+from requests_aws4auth import AWS4Auth  # type: ignore
 
 
 def setup_aws_auth(region: str, profile: str = None) -> AWS4Auth:
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+attrs==21.4.0
 botocore==1.23.54
 Click==8.0.3
 confuse==1.7.0
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-attrs==21.4.0
 botocore==1.23.54
 Click==8.0.3
 confuse==1.7.0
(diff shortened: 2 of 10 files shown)


python/setup.py
```diff
@@ -72,9 +69,6 @@
     keywords="pynessie",
     name="pynessie",
     packages=find_packages(include=["pynessie", "pynessie.*"]),
-    setup_requires=setup_requirements,
-    test_suite="tests",
-    tests_require=[],
```
XN137, line 77: running tests through setup.py is an outdated mechanism, see deprecation notice here:
https://pypi.org/project/pytest-runner/

we are using tox for that as recommended

python/requirements_lint.txt
```diff
@@ -15,16 +15,12 @@
...
 flake8-black==0.2.4
 flake8-bugbear==22.1.11
 flake8-docstrings==1.6.0
 flake8-import-order==0.18.1
 pytest-mypy==0.9.1
 safety==1.10.3
-types-python-dateutil
-types-PyYAML
-types-requests
-types-simplejson
```
XN137, line 29: we can let mypy install these automatically via `--install-types`

python/tests/conftest.py
```diff
@@ -25,12 +25,9 @@
 import attr
 import pytest
 import simplejson
-from _pytest.config import Config
-from _pytest.fixtures import FixtureRequest
```
XN137, line 29: maybe this voodoo was required in earlier pytest versions but it is no longer the case

python/tox.ini
```diff
@@ -50,9 +55,8 @@ commands =
 [testenv]
 setenv =
     PYTHONPATH = {toxinidir}
-passenv = TOXENV CI TRAVIS TRAVIS_* CODECOV_*
+passenv = TOXENV CI CODECOV_*
 deps =
     -r{toxinidir}/requirements_dev.txt
 commands =
-    pip install -U pip
```
XN137, line 57: we pin pip in `requirements_dev.txt` so universally upgrading here made no sense imo

python/tox.ini
```diff
@@ -15,30 +15,35 @@
 #
 
 [tox]
-envlist = py37, py38, py39, py310, flake8, pylint, docs
+envlist = py37, py38, py39, py310, format, lint, docs
 
 [gh-actions]
 python =
     3.7: py37
-    3.8: py38, flake8, pylint
```
XN137, line 23: having `flake8` and `pylint` env split made no sense imo since flake8 also runs safety, mypy etc.
rather have grouping of `lint` and new `format`

dimas-b (dismissed):

rymurr (approved):

python/pynessie/auth/aws.py
```diff
@@ -15,9 +15,9 @@
 # limitations under the License.
 #
 """Use AWS4Auth and botocore to fetch credentials and sign requests."""
-from botocore.credentials import get_credentials
-from botocore.session import Session
-from requests_aws4auth import AWS4Auth
+from botocore.credentials import get_credentials  # type: ignore
```
omarsmak, line 18: Is this ignore for mypy?
```

**Review by omarsmak — APPROVED** [review]

File: python/pynessie/auth/aws.py (hunk 6 above)
Comment (XN137) [reply python/pynessie/auth/aws.py:18]: yes this is a more specific `--ignore-missing-imports`

File: python/requirements.txt [hunk 7]
```diff
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-attrs==21.4.0
```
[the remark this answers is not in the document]
XN137: update: i was confused and we still use `attrs` through the old `attr` api.
the removal just happened to not break things because `attrs` is a transitive dependency of `desert`, but we should not rely on that, since we use the library in our code directly.

python/pynessie/auth/aws.py
omarsmak: Can't we just add this in the `setup.cfg` instead, e.g:
```
[mypy-botocore.*]
ignore_missing_imports = True
```

XN137: ok i have pushed another commit to also ignore botocore imports via the config file (i used 3 usage spots as the cutoff point before).

python/pynessie/auth/aws.py
```diff
@@ -17,7 +17,7 @@
 """Use AWS4Auth and botocore to fetch credentials and sign requests."""
 from botocore.credentials import get_credentials
 from botocore.session import Session
-from requests_aws4auth import AWS4Auth
+from requests_aws4auth import AWS4Auth  # type: ignore
```
omarsmak, line 20: How about ignoring `requests_aws4auth` and `marshmallow_oneofschema` in the config instead?

XN137: as i mentioned before i based it on the amount of usage spots... since both are only imported in a single spot i ignored it locally.
if you insist we do all ignores via config file i can do that?

omarsmak: IMHO I would prefer via the config file to keep things in a central place to manage. WDYT?

XN137: ok done in latest commit

omarsmak (approved):

PR #3358, opened 2022-02-15
publish swaggerhub w/ release

rymurr:


.github/workflows/release-publish.yml
```diff
@@ -297,3 +297,11 @@ jobs:
...
           "${NESSIE_OPENAPI}" \
           "${NESSIE_HELM_CHART}"
+    - name: Update SwaggerHub
+      if: ${{ !env.ACT }}
+      uses: smartbear/swaggerhub-cli:latest
+      env:
+        SWAGGERHUB_API_KEY: ${{ secrets.SWAGGERHUB_API_KEY }}
+        SWAGGERHUB_URL: "https://api.swaggerhub.com"
+      with:
+        args: api:create projectnessie/nessie -f ${{ steps.build_maven.outputs.NESSIE_OPENAPI }} --published=publish --setdefault --visibility=public
```
snazy, line 307: Don't we need the version here / should we add it?
https://github.com/smartbear/swaggerhub-cli#swaggerhub-apicreate

rymurr: it gets read from the file. I thought that was safer/more reliable

snazy: Ah, fair enough

snazy (approved):

PR #3194, opened 2022-01-28
GC base module to identify expired contents

ajantha-bhat:
Introduced a `gc-base` module to identify the expired contents based on cutoff time.

The contents after or equal to cutoff time and the contents that are mapped to
commit head of live keys at the time of cutoff timestamp (for time travel) will be retained.

Two step traversal algorithm is implemented based on reflog, commit log, spark and bloom filter.
Detailed algorithm is added in javadoc of  `GCImpl#identifyExpiredContents()`

All the configuration for GC is provided from `GCParams` class.

Test cases with scenario diagrams can be found in `AbstractRestGCTest`

Note that only `IcebergTable` and `IcebergView` contents GC is supported as of now.

PR #3359, opened 2022-02-15
Refactor AbstractTestRest to support parallel hierarchy

ajantha-bhat:
For GC, need to run only in-memory tests and also should use fresh Nessie server to avoid huge data processing from dropped references. Hence we need a parallel hierarchy for current test.

These changes just extracted from #3194


servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/TestJerseyRestNaiveClientInMemory.java
```diff
@@ -48,7 +49,7 @@
 class TestJerseyRestNaiveClientInMemory extends AbstractTestJerseyRest {
 
   @Override
-  protected void init(NessieApiV1 api, @Nullable HttpClient.Builder httpClient) {
+  protected void init(NessieApiV1 api, @Nullable HttpClient.Builder httpClient, URI uri) {
```
ajantha-bhat, line 52: this URI configuration will be saved and used from testcase to build the client (for example, at spark executor )

snazy (approved):

PR #3387, opened 2022-02-18
Design proposal for webhooks/notifications

hpolloni:


design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+    branch:
+      $ref: Branch
+      description: The branch committed against.
+    newHash:
+      type : string
+      description: The hash of this commit.
+    metadata:
+      $ref: CommitMeta
+      description: This commit's metadata. Containing author, committer, etc.
+    operations:
```
rymurr, line 43: This potentially has repercussion on security. Do we have to check that every Operation is visible to the webhook register owner? I am initially inclined to not include this and require a call to the log endpoint to get this info

Or at the very least we have to define the behaviour for auth

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+    fromRefName:
+      type: string
+      description: The name of the reference being merged into.
+    fromHash:
+      type: string
+      description: The hash to merge from.
+    toBranchName:
+      type: string
+      description: The name of the branch to merge into.
+    expectedHash:
```
rymurr, line 96: why include this?

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+      type: string
+      description: The hash at which the branch was at the moment of merge.
+    newHash:
```
rymurr, line 99: there is no special merge commit like in git, what would this hash be in this context

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+      description: The name of the references to transplant commits from.
+    fromHashes:
+      type: array
+      description: The hashes of the commits to transplant
+      items:
+        type: string
+    toBranchName:
+      type: string
+      description: The name of the branch to transplant commits into.
+    expectedHash:
```
rymurr, line 144: same comments as merge

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+ReferenceCreated:
+  type: object
+  properties:
+    type:
+      type: string
+      description: Always REFERENCE_CREATED
+    eventTime:
+      type: string
+      format: date
+      description: The time the event is fired in ISO date format.
```
rymurr, line 181: can we specify that all timestamps are in UTC

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+      type: string
+      description: Always REFERENCE_DELETED
+    eventTime:
+      type: string
+      format: date
+      description: The time the event is fired in ISO date format.
+    referenceType:
+      type: string
+      description: The reference type. Either BRANCH or TAG.
+    referenceName:
```
rymurr, line 261: we include the hash in a lot of other places, should we include it here? eg the old hash for an assign feels like the same category

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+1. The client will register a webhook notification with Nessie by calling one of the APIs below.
+2. When an event happens in Nessie, it will call back using a `POST` request that:
+   1. Its content-type is `application/json`.
+   2. Its body contains the JSON payload of the event that happened. (See previous section for details on the expected schema).
+
+### Creating a new notification/webhook
+There are different event types a client might want to get notified about. In order to subscribe to a particular event 
+type, a client needs to make a `POST` request to one or more of the following endpoints.
+
+* `/notifications/commits`
```
rymurr, line 292: have you thought about the benefits/weaknesses of havnign a single endpoint and adding the type to the PUT request?

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+Content-type: application/json
+
+{
+  "id": "6bbdcc62-e5b6-4e1b-8f06-3b0bb3620d5e",
+  "type": "WEBHOOK",
+  "url": "http://www.example.org/webhook-callback"
+}
+```
+
+Important note is that nessie doesn't make any efforts to de-duplicate the URL and if a particular URL is subscribed
```
rymurr, line 337: should we de-dup? To me it sounds like we should try

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+twice to the same type of event. Nessie will notify twice, once per each notification.
+
+### Describing, updating and deleting notifications
+As previously described, when creating a notification, the response will contain the URI of the newly created
```
rymurr, line 341: how do we get a list of notifications that we own?

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+twice to the same type of event. Nessie will notify twice, once per each notification.
+
+### Describing, updating and deleting notifications
+As previously described, when creating a notification, the response will contain the URI of the newly created 
+notification. This URI is then used for describing, updating and deleting the specific notification. In general, 
+the following methods will be used.
+
+* `GET /notifications/{notificationId}` for describing a notification.
+* `PUT /notifications/{notificationId}` for updating a notification.
+* `DELETE /notifications/{notificationId}` for deleting a notification.
```
rymurr, line 347: After looking at this I think a more common arrangement would be:
```
GET /notifications  - list your notifications
POST /notifications - create a notification
PUT /notifications/{id} - modify 
GET /notifications/{id} - describe
DELETE /notifications/{id} - delete
```

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Implementation details
+### Observer extension point
+All the event types that we are interested are associated with the `/tree` API. We propose to extend the current 
+TreeApiImpl class with an observer pattern. Every time the specific method in the TreeApi is called, an event is sent
+to the observer. This extension point allows extending to other types of notifications in the future.
+
+### Notification persistence
```
rymurr, line 355: I don't like this part of the proposal. For other extensiosn to the core we have chosen an `SPI` and a simple storage mechanism. For example authz has an SPI and the reference impl uses configs provided at runtime. Implementers can then plug in more sophisticated backends with the SPI. Rather than modifying our database adapter and forcing the same storage mechanism i think the SPI gives more flexibility.

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+notification. This URI is then used for describing, updating and deleting the specific notification. In general, 
+the following methods will be used.
+
+* `GET /notifications/{notificationId}` for describing a notification.
+* `PUT /notifications/{notificationId}` for updating a notification.
+* `DELETE /notifications/{notificationId}` for deleting a notification.
+
+## Implementation details
+### Observer extension point
+All the event types that we are interested are associated with the `/tree` API. We propose to extend the current
```
rymurr, line 351: It seems that the simplest approach is the same as the auth and tracing idioms: create a wrapper class. This avoids having to make significant changes to the kernel

design/notifications.md
```diff
@@ -0,0 +1,360 @@
+
+# Notifications
+This document outlines the design for notifications in Nessie with an application for Webhooks. It outlines the different
+events that we support, the way the notifications are called, how they are stored and what is the API used for creating,
+listing, updating and deleting notifications for a nessie server.
+
+## Event types
+This section lists out all events we support. All events in Nessie contain two pieces of information: the event time and
```
nastra, line 8: nit: maybe make this a list so that it's immediately visible that there is event time +  event type

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+```yaml
+CommitEvent:
+  type: object
+  properties:
+    type:
+      type: string
+      description: Always COMMIT
+    eventTime:
+      type: string
+      format: date
```
nastra, line 32: I think this should be `date-time` (the same is used for `commitTime` in `CommitMeta`). Same comment applies to the other `eventTime` fields

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+  "fromRefName" : "dev",
+  "fromHash" : "e2dad0c9178ef5ff",
+  "toBranchName": "main",
+  "expectedHash" : "6a3efd4f40e88a36",
+  "newHash": "ffaabb78792ddd88"
+}
+```
+
+### Transplants
+Transplants events are triggered whenever transplants from one branch into another happens, i.e. whenever the
```
nastra, line 118: Transplants -> Transplant

design/notifications.md
nastra: agreed

design/notifications.md
nastra: agreed that having an SPI would be better and give us/others more flexibility

nastra (commented):
overall looks good, just a few comments

design/notifications.md
snazy: Probably mention above that all timestamps are in ISO format using UTC.

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+  type: object
+  properties:
+    type:
+      type: string
+      description: Always "MERGE"
+    eventTime:
+      type: string
+      format: date
+      description: The time the event was fired in ISO date format.
+    fromRefName:
```
snazy, line 87: `fromRefName` + `fromHash` can be replaced w/ `fromRef` using the `Reference` type, so the reference type is also clear here (branch/tag/detached).

design/notifications.md
snazy: The new HEAD

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+  type: object
+  properties:
+    type:
+      type: string
+      description: Always "TRANSPLANT"
+    eventTime:
+      type: string
+      format: date
+      description: The time the event was fired in ISO date format.
+    fromRefName:
```
snazy, line 133: Same comment re fromRef as for merge

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Implementation details
+### Observer extension point
+All the event types that we are interested are associated with the `/tree` API. We propose to extend the current 
+TreeApiImpl class with an observer pattern. Every time the specific method in the TreeApi is called, an event is sent
+to the observer. This extension point allows extending to other types of notifications in the future.
+
+### Notification persistence
+Because notifications need to survive between nessie service restarts. We propose to have a separate persistence layer
+for notifications. This persistence layer will use the same database connection as the VersionStore, but will not use
```
snazy, line 357: I'd prefer to leave out implementation specifics here. Traditional databases (RDBMS) and especially NoSQL databases are IMHO bad as a backend for events - it becomes really tricky to implement right.

snazy (commented):
The description of events LGTM - just a few comments about details.

But I'd prefer to move the management and especially the implementation details out of this document. Maybe we can start with a prototype that synchronously calls a statically configured endpoint. Maybe we tend to push events to a messaging system and have dedicated workers that are responsible for actually invoking the registered endpoint(s), deal with retries, error conditions, etc. And then think how we can actually implement a full-blown notification management system.

design/notifications.md
hpolloni: To be honest, I didn't put much thought on auth. I was thinking in making this an "admin" only feature.

design/notifications.md
hpolloni: do you mean like?

```
PUT /notifications

{
 eventType: "COMMIT"
 ...
}
```

OR merging the PUT and POST into a single operation and make the client choose the ID?

```
PUT /notifications/{client chosen ID}

{
 eventType: "COMMIT"
}
```

I like the idea for "futureproofing" for other types of events. I don't particularly like client generated IDs though.

design/notifications.md
hpolloni: I will add a `GET /notifications`

design/notifications.md
hpolloni: I didn't want to go through that rabbit hole. But the more I think about the more I like the idea of deduping both events and subscriptions. The problem is that we will need to document the dedup rules and I didn't want to get into doing that right now. A simple approach is to have deterministic IDs for subscriptions and/or notifications.

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+the following methods will be used.
+
+* `GET /notifications/{notificationId}` for describing a notification.
+* `PUT /notifications/{notificationId}` for updating a notification.
+* `DELETE /notifications/{notificationId}` for deleting a notification.
+
+## Implementation details
+### Observer extension point
+All the event types that we are interested are associated with the `/tree` API. We propose to extend the current 
+TreeApiImpl class with an observer pattern. Every time the specific method in the TreeApi is called, an event is sent
```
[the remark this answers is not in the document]
hpolloni: sounds good. Both Ryan and Eduard had concerns on this part of the proposal. Let's talk offline

design/notifications.md
```diff
@@ -0,0 +1,360 @@
...
+  type: object
+  properties:
+    type:
+      type: string
+      description: Always REFERENCE_DELETED
+    eventTime:
+      type: string
+      format: date
+      description: The time the event is fired in ISO date format.
+    referenceType:
```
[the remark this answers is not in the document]
hpolloni: Do we want the hash to be null or the last HEAD?

design/notifications.md
hpolloni: Will change this to be the reference being deleted.

dimas-b:
This PR has been dormant for about a year - closing. Expect another PR for notifications soon (with different design choices).

PR #3413, opened 2022-02-21
Proper hash collision check for RocksDB

snazy:
RocksDB's `keyMayExist` is a probablistic check, working against a bloom filter.
This means, that the function _may_ return false positives, which leads to
false positive hash-collision errors.

### DIFF
--- a/versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
+++ b/versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
@@ -148,10 +148,7 @@ public class RocksDatabaseAdapter
     lock.lock();
     try {
       byte[] key = dbKey(entry.getHash());
-      if (db.keyMayExist(dbInstance.getCfCommitLog(), key, new Holder<>())) {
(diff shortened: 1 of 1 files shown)


versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
```diff
@@ -430,4 +422,13 @@ protected RefLog fetchFromRefLog(NonTransactionalOperationContext ctx, Hash refL
       NonTransactionalOperationContext ctx, List<Hash> hashes) {
     return fetchPage(dbInstance.getCfRefLog(), hashes, ProtoSerialization::protoToRefLog);
   }
+
+  private void checkForHashCollision(ColumnFamilyHandle cf, byte[] key)
+      throws ReferenceConflictException, RocksDBException {
+    Holder<byte[]> value = new Holder<>();
+    // "may" exist is not "does really" exist, so check if a value was foun d
```
ajantha-bhat, line 429:
```suggestion
    // "may" exist is not "does really" exist, so check if a value was found
```

ajantha-bhat (approved):

PR #3415, opened 2022-02-21
Add DELETE_DEFAULT_BRANCH access check

### LINKED ISSUE #3401: Prevent accidental deletion of the default branch
In production environments it must not be possible to delete the default branch.

This is probably an easy fix in terms of code, but might involve touching a bunch of tests.
For the tests, it should be fine to re-assign the default branch to "beginning of time" instead of deleting + re-creating it.

XN137:
closes #3401 

first commit was my initial approach.
after some Discussion with Robert we thought handling everything in the access checker might better.

preventing accidental deletion now has to be done by setting up the right rules though (its not prevented out of the box, when there are 0 rules) ?

### DIFF
--- a/design/authorization.md
+++ b/design/authorization.md
@@ -69,6 +69,7 @@ which looks like this:
       BatchAccessChecker canViewReference(NamedRef ref);
       BatchAccessChecker canCreateReference(NamedRef ref);
       BatchAccessChecker canDeleteReference(NamedRef ref);
+      BatchAccessChecker canDeleteDefaultBranch();
       BatchAccessChecker canAssignRefToHash(NamedRef ref);
       BatchAccessChecker canReadEntries(NamedRef ref);
       BatchAccessChecker canListCommitLog(NamedRef ref);
@@ -113,6 +114,7 @@ An implementation of the `BatchAccessChecker` interface could be written with th
   * `VIEW_REFERENCE`
   * `CREATE_REFERENCE`
   * `DELETE_REFERENCE`
+  * `DELETE_DEFAULT_BRANCH`
   * `ASSIGN_REFERENCE_TO_HASH`
   * `READ_ENTRIES`,
   * `LIST_COMMIT_LOG`
@@ -132,6 +134,7 @@ nessie.server.authorization.rules.allow_branch_deletion="op=='DELETE_REFERENCE' 
 nessie.server.authorization.rules.allow_updating_entity="op=='UPDATE_ENTITY' && role=='test_user' && path.startsWith('allowed.')"
 nessie.server.authorization.rules.allow_deleting_entity="op=='DELETE_ENTITY' && role=='test_user' && path.startsWith('allowed.')"
 nessie.server.authorization.rules.allow_listing_reflog="op=='VIEW_REFLOG' && role=='admin_user'"
+nessie.server.authorization.rules.allow_deleting_default_branch="op=='DELETE_DEFAULT_BRANCH' && role=='admin_user'"
 ```
 
 > Written with [StackEdit](https://stackedit.io/).
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -60,6 +60,8 @@ nessie.server.send-stacktrace-to-client=false
 #   && role=='admin_user'
 # nessie.server.authorization.rules.allow_listing_reflog=\
 #   op=='VIEW_REFLOG' && role=='admin_user'
+# nessie.server.authorization.rules.allow_deleting_default_branch=\
+#   op=='DELETE_DEFAULT_BRANCH' && role=='admin_user'
 
 
 ### which type of version store to use: INMEMORY, ROCKS, DYNAMO, MONGO.
--- a/servers/services/src/main/java/org/projectnessie/services/authz/Check.java
+++ b/servers/services/src/main/java/org/projectnessie/services/authz/Check.java
@@ -42,6 +42,7 @@ public interface Check {
     CREATE_REFERENCE(true, false),
     ASSIGN_REFERENCE_TO_HASH(true, false),
     DELETE_REFERENCE(true, false),
+    DELETE_DEFAULT_BRANCH(false, false),
     READ_ENTRIES(true, false),
     LIST_COMMIT_LOG(true, false),
     COMMIT_CHANGE_AGAINST_REFERENCE(true, false),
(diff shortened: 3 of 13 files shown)


servers/services/src/main/java/org/projectnessie/services/authz/Check.java
```diff
@@ -42,6 +42,7 @@
     CREATE_REFERENCE(true, false),
     ASSIGN_REFERENCE_TO_HASH(true, false),
     DELETE_REFERENCE(true, false),
+    DELETE_DEFAULT_BRANCH(false, false),
```
XN137, line 45: due to its relation to `DELETE_REFERENCE` i put this right after (and did the same in all other files).
if we want (or need due to relying on enumeration number somewhere) new values to be added at the end, i can adjust them accordingly.

design/authorization.md
```diff
@@ -69,6 +69,7 @@ which looks like this:
       BatchAccessChecker canViewReference(NamedRef ref);
       BatchAccessChecker canCreateReference(NamedRef ref);
       BatchAccessChecker canDeleteReference(NamedRef ref);
+      BatchAccessChecker canDeleteDefaultBranch();
```
XN137, line 72: imo its a bit odd that we link to the code but also copy its contents here

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -60,6 +60,8 @@ nessie.server.send-stacktrace-to-client=false
 #   && role=='admin_user'
 # nessie.server.authorization.rules.allow_listing_reflog=\
 #   op=='VIEW_REFLOG' && role=='admin_user'
+# nessie.server.authorization.rules.allow_deleting_default_branch=\
+#   op=='DELETE_DEFAULT_BRANCH' && role=='admin_user'
```
XN137, line 64: imo its a bit odd that `VIEW_REFLOG` is not part of `allow_all` rule... but i followed this "new pattern" here.

snazy: Generally, the CEL expressions and rules are flexible and users can create/use what they need/want.

The examples are probably a bit misleading.

The VIEW_REFLOG example should probably use a different role name (superuser? gc-user?) and the example for DELETE_DEFAULT_BRANCH maybe use the "superuser" role. IDK.
But honestly, the "test_user" role names above are similar.

servers/services/src/main/java/org/projectnessie/services/authz/Check.java
snazy: That's fine here - we neither store nor use this enum's ordinal.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
```diff
@@ -124,7 +124,12 @@ protected void assignReference(NamedRef ref, String oldHash, Reference assignTo)
   @Override
   protected void deleteReference(NamedRef ref, String hash)
       throws NessieConflictException, NessieNotFoundException {
-    startAccessCheck().canDeleteReference(ref).checkAndThrow();
+    BatchAccessChecker check = startAccessCheck();
+    if (getConfig().getDefaultBranch().equals(ref.getName())) {
+      check.canDeleteDefaultBranch();
+    }
+    check.canDeleteReference(ref);
```
XN137, line 131: note: we keep the old check in place as well

snazy (approved):
+1

nastra (approved):
LGTM, but this is only preventing the delete when authorization is enabled, right? A user could still just delete the default branch when no authorization is enabled

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
ajantha-bhat: why do we need to check both here? This should be in else case ?

ajantha-bhat: Basically for default reference, if the user has access to delete it then it should not matter whether user has delete reference permission or not right?

@nastra, @snazy : WDYT ?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
```diff
@@ -124,7 +124,12 @@ protected void assignReference(NamedRef ref, String oldHash, Reference assignTo)
   @Override
   protected void deleteReference(NamedRef ref, String hash)
       throws NessieConflictException, NessieNotFoundException {
-    startAccessCheck().canDeleteReference(ref).checkAndThrow();
+    BatchAccessChecker check = startAccessCheck();
+    if (getConfig().getDefaultBranch().equals(ref.getName())) {
```
snazy, line 128: Actually, you need to check whether `ref` is a `BranchName`

XN137: addressed in latest commit

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
XN137: moved to else after confirming with Robert

ajantha-bhat (approved):

XN137:
> LGTM, but this is only preventing the delete when authorization is enabled, right? A user could still just delete the default branch when no authorization is enabled

yes indeed, which is why i mention this in the PR description:

> preventing accidental deletion now has to be done by setting up the right rules though (its not prevented out of the box, when there are 0 rules) ?

i am looking for ideas how we can now make it more difficult to delete the default branch out of the box :sweat_smile:

snazy (dismissed):

XN137:
until we have a good idea how to prevent default branch deletion out of the box i suggest to rename this PR to just `Add DELETE_DEFAULT_BRANCH access check`, wdyt?

ajantha-bhat:
> until we have a good idea how to prevent default branch deletion out of the box i suggest to rename this PR to just Add DELETE_DEFAULT_BRANCH access check, wdyt?

Just thinking out loud, 
when the access checker is enabled, we can follow the current code.
when the access checker is disabled, we need to know whether the access checker was there or directly executing. we can use a thread local to set some flag when access checker is present. If the thread local flag is present. That means already validated and allow deleting. Else always block deleting inside  `TreeApiImpl#deleteReference`

XN137:
PR is renamed and needs another review after recent test fixes

PR #3428, opened 2022-02-23
Create Exception Mapper that specifically handles ConstraintViolationExceptions

nastra:
The reason to do this is that it's possible that users register their
own exception mappers that handle `ConstraintViolationExceptions`. Given
that our current `ValidationExceptionMapper` is registered on the
superclass of `ConstraintViolationExceptions`, namely
`ValidationException`, it's possible that the exception handling code
for `ConstraintViolationException` in `ValidationExceptionMapper` isn't
called. This is because during exception validation, a distance for the
current exception is calculated to the exception that is being handled
by a particular exception mapper.


servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,66 @@
...
+import javax.validation.ElementKind;
+import javax.validation.Path;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+import org.projectnessie.error.ErrorCode;
+import org.projectnessie.services.config.ServerConfig;
+
+/**
+ * "Special" implementation for exceptions that extend {@link ConstraintViolationException}, as
+ * those do not "go through" {@link NessieExceptionMapper}.
```
XN137, line 30: should this mention that these exceptions should also not go through `ValidationExceptionMapper`  ?

XN137 (dismissed):
i dont fully understand the usecase yet... just at the java level this lgtm (but you should wait for another review i guess)

> it's possible that users register their
own exception mappers that handle ConstraintViolationExceptions

> it's possible that the exception handling code
for ConstraintViolationException in ValidationExceptionMapper isn't
called.

if they do that, doesnt it override the mapper we are adding in this PR? or how would it be decided which of the 2 gets applied?
if it overrides our mapper then as before, our logic would not get called.
so i am not sure how this helps with the described usecase.

but to me having a dedicated mapper for the specific exception type probably makes sense either way (avoids instanceof checks).

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Inject
+  public ConstraintViolationExceptionMapper(ServerConfig config) {
+    super(config);
+  }
+
+  @Override
+  public Response toResponse(ConstraintViolationException exception) {
+    ErrorCode errorCode = ErrorCode.BAD_REQUEST;
+    for (ConstraintViolation<?> violation : exception.getConstraintViolations()) {
```
snazy, line 50: Can't we do this in `ValidationExceptionMapper`?

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
```diff
@@ -0,0 +1,66 @@
...
+      for (final Path.Node node : violation.getPropertyPath()) {
+        final ElementKind kind = node.getKind();
+
+        if (ElementKind.RETURN_VALUE.equals(kind)) {
```
snazy, line 54: Nit: enum can use reference equality.

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
nastra: done

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
nastra: unfortunately no

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
nastra: fixed

servers/rest-services/src/main/java/org/projectnessie/services/rest/ConstraintViolationExceptionMapper.java
snazy: But `org.projectnessie.services.rest.ValidationExceptionMapper#toResponse` already has that `instanceof ConstraintViolationException` check. What's wrong there?

snazy: Oh, I re-read the "distance" calculation in the PR description (and finally understood what you're saying).

snazy:
Final Q - this is covered by our existing unit tests, right?

nastra:
> Final Q - this is covered by our existing unit tests, right?

Yes, all validation stuff is depending on this to properly work

snazy (approved):

PR #3429, opened 2022-02-23
Log fpp for merged ContentBloomFilters

XN137:
just an idea to at least be aware in case the merging of bloom filters results in deteriorated filter quality

see also:
https://github.com/google/guava/blob/a0e2577de61a0d7e8a3dd075be66a31c93ea0446/guava/src/com/google/common/hash/BloomFilter.java#L244-L252

maybe this scenario can never happen or is very unlikely and not worth the computation / space effort everytime, idk

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
@@ -63,8 +63,9 @@ public class ContentBloomFilter implements Serializable {
   }
 
   /**
-   * @return true if other filters were merged into this instance, which might indicate decreased
-   *     filter quality
+   * A merged bloomfilter might indicate decreased filter quality.
+   *
+   * @return true if other filters were merged into this instance
    */
   public boolean wasMerged() {
     return wasMerged;
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
@@ -62,6 +62,10 @@ public class ContentBloomFilter implements Serializable {
     return filter.expectedFpp();
   }
 
+  /**
+   * @return true if other filters were merged into this instance, which might indicate decreased
+   *     filter quality
+   */
   public boolean wasMerged() {
     return wasMerged;
   }
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
@@ -56,6 +56,10 @@ public class ContentBloomFilter implements Serializable {
     }
   }
 
+  public double getExpectedFpp() {
+    return filter.expectedFpp();
+  }
+
   private String getValue(Content content) {
     switch (content.getType()) {
       case ICEBERG_TABLE:
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
@@ -35,6 +35,7 @@ public class ContentBloomFilter implements Serializable {
   // as the consumer of GC results needs only this info for clean up.
   // String bloom filter with content type prefix + snapshot/version id.
   private final BloomFilter<String> filter;
+  private boolean wasMerged = false;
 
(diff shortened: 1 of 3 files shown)


gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
```diff
@@ -87,18 +92,32 @@ private static int getPartitionsCount(GCParams gcParams, List<Reference> referen
...
                     output.put(k, v);
                   }
                 }));
+    // Since we merged bloom filters log in case their quality deteriorated
+    mergedContentIds.stream()
+        .forEach(
+            contentId -> {
+              double fpp = output.get(contentId).getExpectedFpp();
+              if (fpp > bloomFilterFpp) {
+                LOGGER.info(
```
ajantha-bhat, line 115: a) good to have log level `WARN` for this.

b) Have we explored a testcase which merges equal size bloom filters and changes its fpp after merge? may be we can add one in `TestContentBloomFilter`.  It should validate whether fpp changed after merge.

XN137: i went with INFO because being slightly above the target could still be fine.

https://github.com/google/guava/blob/a0e2577de61a0d7e8a3dd075be66a31c93ea0446/guava/src/com/google/common/hash/BloomFilter.java#L177-L188

mentions only a significantly higher value would indicate a problem.
maybe we can put it on WARN or ERROR if it is above certain tresholds.

gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
```diff
@@ -87,18 +92,32 @@ private static int getPartitionsCount(GCParams gcParams, List<Reference> referen
...
+                    "Fpp of ContentBloomFilter for '{}': {}",
```
dimas-b, line 116: Could we expose this as a metric too?

XN137: maybe @ajantha-bhat has it planned already for the 2nd/3rd GC impl phase?

ajantha-bhat: no plans for metrics as of now. But always good to have. 
GC results we are storing in table, so we can run SQL get some metrics out of it. But nothing concrete as of now.

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
```diff
@@ -53,13 +54,18 @@ public boolean mightContain(Content content) {
...
       this.filter.putAll(filter.filter);
+      this.wasMerged = true;
     }
   }
 
   public double getExpectedFpp() {
     return filter.expectedFpp();
   }
 
+  public boolean wasMerged() {
```
snazy, line 65: Maybe add some javadoc explaining the reason for this flag.

gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
```diff
@@ -93,29 +91,31 @@ private static int getPartitionsCount(GCParams gcParams, List<Reference> referen
...
-            contentId -> {
-              double fpp = output.get(contentId).getExpectedFpp();
-              if (fpp > bloomFilterFpp) {
-                LOGGER.info(
-                    "Fpp of ContentBloomFilter for '{}': {}",
-                    contentId,
-                    String.format("%.3f", fpp));
+            e -> {
+              ContentBloomFilter bloomFilter = e.getValue();
+              if (bloomFilter.wasMerged()) {
```
snazy, line 110: Nit: can use `Stream.filter()` (also for the `if` below)

XN137: used for the first one but since we only want to call `expectedFpp()` once, mapping the entries to new `bstractMap.SimpleImmutableEntry` or `Pair` just to be able to use filter for the 2nd if, does not improve readability imo.

snazy (approved):

ajantha-bhat (approved):

XN137:
rebased so that codecov does not claim lost coverage in `[ValidationExceptionMapper.java]` anymore hopefully

snazy (approved):

PR #3431, opened 2022-02-23
Refactor the `StreamingUtil` class

### LINKED ISSUE #3402: Refactor `StreamingUtil`
`StreamingUtil` was intended to help users of the Java clients to use `Stream`s for commit-log, entries, etc.
Initially it was 2/3 methods with a few parameters, but evolved to have quite a few parameters per method now.

Instead of taking individual parameters, the methods in there should actually use builders - at best reusing the builders from `NessieApiV1` - either via a callback (think: `Consumer<GetCommitLogBuilder>`) passed into e.g. `getCommitLogStream` or by passing an already prepared builder into e.g. `getCommitLogStream`.

Goal: have less parameters for the methods in `StreamingUtil` and no need to change (think: break) the signature of these methods.

tomekl007:
Simplifies all API methods in the `StreamingUtil` class.
It allows the callers to pass a `Consumer<Builder>` that they may leverage to customize the builder.

Fixes #3402


clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -43,16 +48,30 @@ private StreamingUtil() {
...
+   *     Optional.empty()} the stream will be unbounded.
+   * @param filter - a filter expression to use when fetching a stream of references.
    * @return stream of {@link Reference} objects
    */
   public static Stream<Reference> getAllReferencesStream(
-      @NotNull NessieApiV1 api, OptionalInt maxRecords, @Nullable String filter)
+      @NotNull NessieApiV1 api,
+      @NotNull Consumer<GetAllReferencesBuilder> builderCustomizer,
+      @NotNull OptionalInt maxRecords,
+      @Nullable String filter)
```
snazy, line 64: It's fine to remove `filter`.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -63,25 +82,30 @@ private StreamingUtil() {
    * <p>The {@link Stream} returned by {@code getEntriesStream(ref, OptionalInt.empty())}, if not
    * limited, returns all commit-log entries.
    *
+   * @param api The {@link NessieApiV1} to use
    * @param ref a named reference (branch or tag name) or a commit-hash
```
snazy, line 86: This is wrong for a while now (cannot be a commit-hash anymore).
`ref` should actually be of type `Reference` nowadays

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -94,27 +118,32 @@ private StreamingUtil() {
    *
    * @param api The {@link NessieApiV1} to use
    * @param ref a named reference (branch or tag name) or a commit-hash
```
snazy, line 120: (same here)

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -94,27 +118,32 @@ private StreamingUtil() {
...
       @NotNull NessieApiV1 api,
-      @NotNull String ref,
-      @Nullable String hashOnRef,
-      @Nullable String untilHash,
-      @Nullable String filter,
-      OptionalInt maxRecords,
+      String ref,
+      @NotNull Consumer<GetCommitLogBuilder> builderCustomizer,
+      @NotNull OptionalInt maxRecords,
       boolean fetchAdditionalInfo)
```
snazy, line 135: It's fine to remove `fetchAdditionalInfo`

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -109,9 +109,7 @@ private IdentifiedResult computeExpiredContents(
         StreamingUtil.getCommitLogStream(
             gcStateParamsPerTask.getApi(),
             Detached.REF_NAME,
```
snazy, line 111: With `ref` param being of type `Reference`, this could be simplified to `Detached.of(<hash>)` and the consumer can be `null`

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -43,16 +48,30 @@ private StreamingUtil() {
...
+   *     with additional parameters (e.g.: filter).
+   * @param maxRecords - a maximum number of records in the stream. If it is {@code
+   *     Optional.empty()} the stream will be unbounded.
+   * @param filter - a filter expression to use when fetching a stream of references.
    * @return stream of {@link Reference} objects
    */
   public static Stream<Reference> getAllReferencesStream(
-      @NotNull NessieApiV1 api, OptionalInt maxRecords, @Nullable String filter)
+      @NotNull NessieApiV1 api,
+      @NotNull Consumer<GetAllReferencesBuilder> builderCustomizer,
```
snazy, line 62: Should we allow the consumer to be `null` in case you don't need any customization?

snazy (commented):
The `Consumer` based approach looks good.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
tomekl007: I am not sure if this is a good pattern to pass null as a `Consumer`.
The caller can pass a no_op consumer if he does not wish to alter the behavior.
 I was thinking about creating:
`public static final Consumer< GetAllReferencesBuilder> NO_OP_GET_ALL_REFERENCES = ...`
but we need to create it for all 4 types, so it is becoming verbose. 
The other solution would be to create an overloaded method without the:
`@NotNull Consumer<GetAllReferencesBuilder> builderCustomizer` parameter, and let the `StreamUtil` pass `(c) -{}` instead of the caller. I would favor the latter solution. Wdyt?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
tomekl007: done

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
snazy: Yea - let's just keep it `@NotNull`. No need for constants (`x -> {}` is short enough).

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -63,25 +76,28 @@ private StreamingUtil() {
...
     return new ResultStreamPaginator<>(
             EntriesResponse::getEntries,
-            (ref1, pageSize, token) ->
-                builderWithPaging(api.getEntries(), pageSize, token)
-                    .refName(ref1)
-                    .hashOnRef(hashOnRef)
-                    .filter(filter)
-                    .get())
+            (reference, pageSize, token) -> {
+              builderCustomizer.accept(api.getEntries());
```
snazy, line 98: Looks like this line has no effect?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -63,25 +76,28 @@ private StreamingUtil() {
...
+   *     Optional.empty()} the stream will be unbounded.
    * @return stream of {@link Entry} objects
    */
   public static Stream<Entry> getEntriesStream(
       @NotNull NessieApiV1 api,
       @NotNull String ref,
-      @Nullable String hashOnRef,
-      @Nullable String filter,
-      OptionalInt maxRecords)
+      @NotNull Consumer<GetEntriesBuilder> builderCustomizer,
```
snazy, line 91: Actually, `Consumer<X>` should be `Function<X, X>` because all builder methods return a builder instance (even it ifs `this`).

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
tomekl007: Do you think that we should change the API signature to accept not `Stirng ref`, but `Reference ref`?
In that case, the caller would need to do this:
`Detached.of(Detached.REF_NAME),`
It looks a bit verbose.
Also, in tests we are passing `branch.getName()` - I am not sure how to construct a `Reference` based only on the branch name. The `Branch.of()` needs two parameters: `name` and `hash`

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
tomekl007: Builders are mutable, so it will modify the entries builder.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
tomekl007: Yes, that would be a cleaner API even, the builders are mutable.

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -63,25 +76,28 @@ private StreamingUtil() {
...
+   * @param builderCustomizer a Function that takes and returns {@link GetEntriesBuilder} that is
+   *     passed to the caller of this method. It allows the caller to customize the
+   *     GetEntriesBuilder with additional parameters (e.g.: hashOnRef, filter).
+   * @param maxRecords - a maximum number of records in the stream. If it is {@code
+   *     Optional.empty()} the stream will be unbounded.
    * @return stream of {@link Entry} objects
    */
   public static Stream<Entry> getEntriesStream(
       @NotNull NessieApiV1 api,
       @NotNull String ref,
```
snazy, line 90: Just one thing left: `ref` should be of type `Reference`
(Noted it probably at the wrong place: https://github.com/projectnessie/nessie/pull/3431/files#r813583973)

tomekl007: yes, see my response here:
https://github.com/projectnessie/nessie/pull/3431#discussion_r813747650
are we ok with it?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -93,29 +109,28 @@ private StreamingUtil() {
...
-                    .hashOnRef(hashOnRef)
-                    .filter(filter)
-                    .untilHash(untilHash)
-                    .get())
-        .generateStream(ref, maxRecords);
+            (reference, pageSize, token) -> {
+              GetCommitLogBuilder builder = builderCustomizer.apply(api.getCommitLog());
+              return builderWithPaging(builder, pageSize, token).refName(reference).get();
+            })
+        .generateStream(ref.getHash(), maxRecords);
```
tomekl007, line 133: The last question: 
I see that `ref` that is passed here may be `Detached.of(hash)`, but also `Reference.of(name,null)`. For that reason, is it safe to pass hash here or should we have a conditional logic:
`ref.getHash != null ? ref.getHash() : ref.getName()`
?

snazy (approved):
LGTM!

Issue #3439, opened 2022-02-24
USE reference@hash is always setting reference head


ajantha-bhat:
In nessie spark sql extensions, 
`use reference main at $someHash`

is always setting the head reference instead of specified reference.

cause:
At nessie side we are only using the refName here and ignoring the ref hash
https://github.com/projectnessie/nessie/blob/main/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala#L193

At iceberg side, we need a to introduce a new spark configuration (to accept refHash along with refName)
https://github.com/apache/iceberg/blob/master/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java#L129
and the set this hash in NessieCatalog.initialize().

**Note**
So, it needs 2 PR one at iceberg and one at nessie.
I will only work on this as I have analyzed. PR will be up tomorrow as it is bit late in my timezone.

ajantha-bhat:
Raised #3440 from Nessie side.
Once this is merged and release is made, need to fix at Iceberg side using this configuration.
Later end to end test has to be updated in ITNessieStatements once Iceberg releases it.

ajantha-bhat:
Raised https://github.com/apache/iceberg/pull/4330 at Iceberg side.

ajantha-bhat:
Once, Nessie depends on 0.14.0 we need to add/update the test case at the Nessie side before closing this issue.

snazy:
What's the status of this issue?

ajantha-bhat:
> What's the status of this issue?

we need to update the testcase on Nessie side. I will track it to closure this week

ajantha-bhat:
Whenever the SQL "USE REFERENCE AT HASH" is used, defined hash will not be updated in the backend. Only client uses that information. Nessie integration testcases are independent of NessieClient.  

Also, we may need to update the "SHOW REFERENCE" command similar to this (#5081). But that is an independent issue. Hence, closing this ticket as completed.

Issue #3432, opened 2022-02-23
Extended logs showing always the same metadata location


hpolloni:
**What happened**:
All commits on the "extended" log format (i.e. `nessie log -x`) show the same metadata location. The same for `/api/v1/trees/tree/{branch}/log?fetch=ALL`

**What you expected to happen**:
For commits to show the iceberg metadata location at the time of the specific commit. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Setup a local instance of nessie. I tried with `mvn quarkus:dev`, docker and docker compose with MongoDB.
2. Setup pyspark following the same documentation. JARs are `org.apache.iceberg:iceberg-spark-runtime-3.1_2.12:0.13.1,org.projectnessie:nessie-spark-extensions:0.20.1`
3. Issue a CREATE, followed by a simple ALTER TABLE.

**Details**:
- Nessie server type (docker/built from source) and version: Repro in local quarkus, docker and docker-compose (mongo version)
- Client type (Ex: UI/Spark/pynessie ...) and version: pynessie and spark. 
- Spark version: 3.1.2
- Iceberg version: 0.13.1
- Nessie version: 0.20.1

ajantha-bhat:
@hpolloni : **This is proper behaviour. Not an issue.** 
Because,

If the commit log from specific hash shows table metadata at the time of commit, then that file may not exist at the present time due to Iceberg's cleanup from `write.metadata.previous-versions-max` ([check here](https://iceberg.apache.org/docs/latest/maintenance/#remove-old-metadata-files))

For Iceberg table, Nessie maintains global state and on-reference state. ([read here](https://projectnessie.org/develop/spec/#on-reference-state-vs-global-state))

So, global state always points to latest metadata file and on-reference state points to snapshot id at the time of commit.
we can use the latest metadata file and that snapshot id to get the iceberg table state at the time of commit or use it for time travel. 

Hope it clears.

hpolloni:
Hey @ajantha-bhat thanks for the explanation. I guess I wasn't being clear. I thought the problem that I was seeing is related to the way we are storing the metadata location, but according to your explanation this should work.

I have the following code. Basically this will do a `CREATE`, followed by an `ALTER` and a `USE REF`. I'd expect the table t1 to have a different schema after the `USE REF` but it doesn't. Both outputs are the same. Is this expected behavior?

This is the code that I'm testing with. 

```python
from pyspark.sql import *
from pyspark import SparkConf
import os

[...]

PR #3440, opened 2022-02-25
Nessie SQL extensions: Fix Use reference at hash

ajantha-bhat:
for `use reference @ hash` sql command, we are calculating/validating the ref+ hash. But in the end we are not setting hash to the catalog. only ref name is set to the catalog. Hence the issue.

These are changes from Nessie side for #3439 
Once this is merged and release is made, need to fix at Iceberg side using this configuration. 
Later end to end test has to be updated in `ITNessieStatements` once Iceberg releases it.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
+++ b/clients/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
@@ -57,8 +57,15 @@ public final class NessieConfigConstants {
    * <p>Note that "basic" HTTP authentication is not considered secure, use {@code BEARER} instead.
    */
   public static final String CONF_NESSIE_AUTH_TYPE = "nessie.authentication.type";
-  /** Config property name ({@value #CONF_NESSIE_REF}) for the nessie reference used by clients. */
+  /**
+   * Config property name ({@value #CONF_NESSIE_REF}) for the nessie reference name used by clients.
+   */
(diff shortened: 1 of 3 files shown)


nastra:
sounds similar to https://github.com/projectnessie/nessie/pull/2359

ajantha-bhat:
> sounds similar to https://github.com/projectnessie/nessie/pull/2359

yeah, it talks about show log and show reference. Assuming USE REFERENCE at HASH works. But USE REFERENCE at HASH itself doesn't work. So, this PR.

Actually derived from #3432

nastra:
Indeed that's an issue. We have a test that actually verifies the configured hash in https://github.com/projectnessie/nessie/blob/187859163a6000d32a79cc972f8d68cbccdc6fe7/clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java#L451-L458. We output the correctly calculated hash in https://github.com/projectnessie/nessie/blob/187859163a6000d32a79cc972f8d68cbccdc6fe7/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BaseUseReferenceExec.scala#L36-L37 but when we configure Spark with the ref, we currently only set the ref name, but not the hash so your fix definitely makes sense

nastra (approved):

PR #3445, opened 2022-02-25
Remove output_string from execute_cli_command

XN137:
this is less verbose and has improved type hints.
also make distinction between stdout and stderr of cli cmds.

### DIFF
--- a/python/tests/conftest.py
+++ b/python/tests/conftest.py
@@ -64,21 +64,20 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
     shutil.rmtree(nessie_test_config.config_dir)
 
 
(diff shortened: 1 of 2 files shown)


python/tests/conftest.py
```diff
@@ -64,23 +64,22 @@ def pytest_sessionfinish(session: pytest.Session, exitstatus: int) -> None:
...
-
-
-def _run(args: List[str], input_data: Optional[str] = None, ret_val: int = 0) -> Result:
     result = CliRunner().invoke(cli.cli, args, input=input_data)
     if result.exit_code != ret_val:
-        print(result.output)
-        print(result)
+        print(result.stdout)
+        print(result.stderr)
+        print(result)  # exception
```
XN137, line 73: compare with https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/testing.py#L133-L156

omarsmak (approved):

PR #3446, opened 2022-02-25
Regenerate Python CLI docs

XN137:
Running `tox -e docs` locally i noticed that the docs were out of date after recent CLI changes like https://github.com/projectnessie/nessie/pull/3214

This also fixes a trailing whitespace indentation issue i noticed.

Later I will do another PR that will fail if docs are not up-to-date as part of CI to prevent this from happening again.


snazy (approved):
Nice catch!

I think, the check you're thinking of can be integrated into the python-ci steps?

PR #3447, opened 2022-02-25
Add tracing to database-adapter internal operations

snazy:
Tracing spans around:
* high-level `DatabaseAdapter` methods, via `TracingDatabaseAdapter`
* for each try-loop attempt ("CAS")
* low-level around all operations around database accesses

Adds a (rudimentary) test to compare spans produced via
`VersionStore` -> `DatabaseAdapter`.

No new configuration option, uses the existing "tracing-enabled"
option in `VersionStoreConfig`.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -24,6 +24,7 @@ import static org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUti
 import static org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUtil.referenceNotFound;
 import static org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUtil.takeUntilExcludeLast;
 import static org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUtil.takeUntilIncludeLast;
+import static org.projectnessie.versioned.persist.adapter.spi.Traced.trace;
 
 import com.google.common.hash.Hasher;
 import com.google.protobuf.ByteString;
@@ -109,6 +110,8 @@ import org.projectnessie.versioned.persist.adapter.RefLog;
 public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends DatabaseAdapterConfig>
     implements DatabaseAdapter {
 
+  protected static final String TAG_HASH = "hash";
+  protected static final String TAG_COUNT = "count";
   protected final CONFIG config;
   protected final ContentVariantSupplier contentVariantSupplier;
 
@@ -626,14 +629,33 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
   }
 
   /** Load the commit-log entry for the given hash, return {@code null}, if not found. */
-  protected abstract CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash);
+  protected final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
+    try (Traced ignore = trace("fetchFromCommitLog").tag(TAG_HASH, hash.asString())) {
+      return doFetchFromCommitLog(ctx, hash);
+    }
+  }
+
+  protected abstract CommitLogEntry doFetchFromCommitLog(OP_CONTEXT ctx, Hash hash);
 
   /**
    * Fetch multiple {@link CommitLogEntry commit-log-entries} from the commit-log. The returned list
    * must have exactly as many elements as in the parameter {@code hashes}. Non-existing hashes are
    * returned as {@code null}.
    */
-  protected abstract List<CommitLogEntry> fetchPageFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes);
+  protected final List<CommitLogEntry> fetchPageFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes) {
+    if (hashes.isEmpty()) {
+      return Collections.emptyList();
+    }
+    try (Traced ignore =
+        trace("fetchPageFromCommitLog")
+            .tag(TAG_HASH, hashes.get(0).asString())
+            .tag(TAG_COUNT, hashes.size())) {
+      return doFetchPageFromCommitLog(ctx, hashes);
+    }
+  }
+
+  protected abstract List<CommitLogEntry> doFetchPageFromCommitLog(
+      OP_CONTEXT ctx, List<Hash> hashes);
 
   /** Reads from the commit-log starting at the given commit-log-hash. */
   protected Stream<CommitLogEntry> readCommitLogStream(OP_CONTEXT ctx, Hash initialHash)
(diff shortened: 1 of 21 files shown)


snazy:
This change became a bit bigger than I actually wanted it to be, but it does its job and should give us a lot of insights.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieMockedTracingExtension.java
```diff
@@ -0,0 +1,104 @@
...
+ * and parameters.
+ *
+ * <p>Note that the {@link MockTracer} instance is constant per class - like a {@code static final}
+ * field.
+ */
+public class NessieMockedTracingExtension
+    implements BeforeAllCallback, BeforeEachCallback, ParameterResolver {
+  private static final Namespace NAMESPACE = Namespace.create(NessieMockedTracingExtension.class);
+
+  private static final MockTracer TRACER = new MockTracer();
```
dimas-b, line 45: Could we use `MockTracer` to replace our `TestTracer`?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TracingDatabaseAdapter.java
```diff
@@ -0,0 +1,274 @@
...
+        trace("values").tag(TAG_HASH, commit.asString()).tag(TAG_COUNT, keys.size())) {
+      return delegate.values(commit, keys, keyFilter);
+    }
+  }
+
+  @Override
+  public Stream<CommitLogEntry> commitLog(Hash offset) throws ReferenceNotFoundException {
+    Traced trace = trace("commitLog").tag(TAG_HASH, offset.asString());
+    try {
+      return delegate.commitLog(offset).onClose(trace::close);
```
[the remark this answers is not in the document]
snazy: Makes sense.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TracingDatabaseAdapter.java
```diff
@@ -0,0 +1,274 @@
...
+    } catch (RuntimeException e) {
+      trace.close();
+      throw e;
+    }
+  }
+
+  @Override
+  public Optional<ContentIdAndBytes> globalContent(
+      ContentId contentId, ToIntFunction<ByteString> contentTypeExtractor) {
+    try (Traced ignore = trace("globalContent")) {
```
nastra, line 236: should we add the `contentId` as a tag here?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieMockedTracingExtension.java
snazy: Probably - but out of scope for this PR IMO.
`TestTracer` works fine - and since opentracing is rather deprecated, we'll have to touch it when moving to opentelemetry anyway.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
```diff
@@ -149,4 +169,12 @@ public void sleepMillis(long sleepMillis) {
       }
     }
   }
+
+  private void start() {
+    traced = trace(opName).tag(TAG_ATTEMPT, getRetries());
```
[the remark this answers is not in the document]
snazy: This one's fine - it's all properly nested.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -626,14 +629,33 @@ protected void validateHashExists(OP_CONTEXT ctx, Hash hash) throws ReferenceNot
...
    * Fetch multiple {@link CommitLogEntry commit-log-entries} from the commit-log. The returned list
    * must have exactly as many elements as in the parameter {@code hashes}. Non-existing hashes are
    * returned as {@code null}.
    */
-  protected abstract List<CommitLogEntry> fetchPageFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes);
+  protected final List<CommitLogEntry> fetchPageFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes) {
+    if (hashes.isEmpty()) {
+      return Collections.emptyList();
+    }
+    try (Traced ignore =
```
nastra, line 649: we're leaking tracing code into the existing implementation or am I missing something?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/Traced.java
```diff
@@ -0,0 +1,63 @@
...
+        t.buildSpan(spanName)
+            .asChildOf(t.activeSpan())
+            .withTag(tagName("operation"), opName)
+            .start();
+    scope = t.activateSpan(span);
+  }
+
+  @Override
+  public void close() {
+    scope.close();
```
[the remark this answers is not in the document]
snazy: Neither can actually be null here

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: It's tracing about the inner workings of the implementation, so yes.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TracingDatabaseAdapter.java
snazy: done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
nastra: what does this mean if tracing is not enabled? Is this a no-op or are we creating unnecessary objects?

snazy: noop

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/Traced.java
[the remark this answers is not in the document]
nastra: but what if setup code fails and then close() is called? isn't that a legitimate case where it could be null?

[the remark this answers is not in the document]
snazy: That would mean, that the tracing implementation is so borked, that it succeeded earlier and suddenly fails.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
nastra: I think it's still unclear to me from reading the code why this would be a noop when tracing is disabled. `Traced.trace()` creates & starts a span so it's still adding object creation overhead or not? Maybe this can be clarified with a comment or so in `Traced`?

snazy: It's still a Noop implementation in OpenTracing, right?

nastra: I haven't looked into OpenTracing, that's why I'm asking, because it's unclear to people reading the code when not knowing implementation details of OpenTracing itself

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/NessieMockedTracingExtension.java
dimas-b: fair enough.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: If disabled Open Tracing will return non-null noop implementations, but it looks like we will still wrap them in another layer of `Traced` objects.

From my POV it is fine, unless we want to optimize for the disabled tracing use case, which I guess we do not actually want since we expect tracing to be used with higher or lower degree of sampling depending on the deployment.

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -253,27 +253,20 @@ static String makeSpanName(String name) {
...
-      throw traceError(span.get(), e);
-    } finally {
-      // We cannot `catch (E1 e)`, so assume that the delegate threw an exception, when result==null
-      // and then close the span.
-      if (result == null) {
-        span.close();
+    try (SpanHolder span = createSpan(spanName + ".stream", spanBuilder);
+        Scope ignore = activeScope(span.get())) {
+      try {
+        return invoker.handle();
```
dimas-b, line 264: I'm ok with this change, but I think the comment above is no longer 100% accurate because now we close the span before returning from this method.

dimas-b: For background: my idea for the old impl. was to keep the span open until the stream is closed to allow tracing visualizations to display the whole duration of the stream processing as an "async" operation. I would not expect nested traced calls to be contained within the stream span, but I might be been mistaken :)

snazy: Well, there are usually nested calls that are also traced. It's usually not a problem when only the VersionStore's traced, but with this change, the probability is nearly 100% that consuming ("starting") the stream causes nested activity.

PR #3449, opened 2022-02-25
Add up-to-date check for Python CLI docs

XN137:
Follow-up to https://github.com/projectnessie/nessie/pull/3446

the 1st commit contains an intentional error and CI failed like this:
```
   docs_sync_check run-test: commands[0] | bash -c 'changed_docs=$(git status --porcelain /home/runner/work/nessie/nessie/python/docs); if [ -n "$changed_docs" ]; then echo "Docs are out of sync: $changed_docs"; exit 1; fi'
  Docs are out of sync:  M python/docs/content_list.rst
  Error: ERROR: InvocationError for command /usr/bin/bash -c 'changed_docs=$(git status --porcelain /home/runner/work/nessie/nessie/python/docs); if [ -n "$changed_docs" ]; then echo "Docs are out of sync: $changed_docs"; exit 1; fi' (exited with code 1)
___________________________________ summary ____________________________________
  py38: commands succeeded
  lint: commands succeeded
  docs: commands succeeded
Error: ERROR:   docs_sync_check: commands failed
Error: Process completed with exit code 1.
```
https://github.com/projectnessie/nessie/runs/5357440319

the later commit fixes the problem and passes CI successfully:
```
  docs_sync_check run-test: commands[0] | bash -c 'changed_docs=$(git status --porcelain /home/runner/work/nessie/nessie/python/docs); if [ -n "$changed_docs" ]; then echo "Docs are out of sync: $changed_docs"; exit 1; fi'
___________________________________ summary ____________________________________
  py38: commands succeeded
  lint: commands succeeded
  docs: commands succeeded
  docs_sync_check: commands succeeded
```
https://github.com/projectnessie/nessie/runs/5357493990

OUTDATED:
Note that i thought about moving this check into tox itself but i dont think it should reason about the local git repo state.
In the github action we have much more control over this, so imo thats the better place for this logic.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -79,7 +79,29 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Test with tox
-      run: tox
+      run: |
+        tox
+
+        if [[ "${{ matrix.python-version }}" == "3.8" ]]; then
+          echo "Starting Python docs up-to-date check ..."
+
+          git_changed_files=$(git status --porcelain)
+          if [ -n "${git_changed_files}" ]; then
+            echo "Python docs up-to-date check needs to start on a clean git repo!"
+            echo "Changed files: ${git_changed_files}"
+            exit 1
+          fi
+
+          tox -e docs
+
+          git_changed_files=$(git status --porcelain)
+          if [ -n "${git_changed_files}" ]; then
+            echo "Python docs up-to-date check has failed! Re-run 'tox -e docs' locally"
+            echo "Changed files: ${git_changed_files}"
+            exit 1
+          fi
+          echo "Python docs are fully up-to-date!"
+        fi
       working-directory: ${{env.working-directory}}
     - uses: codecov/codecov-action@v1
       with:
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -79,29 +79,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Test with tox
-      run: |
-        tox
-
-        if [[ "${{ matrix.python-version }}" == "3.8" ]]; then
-          echo "Starting Python docs up-to-date check ..."
-
-          git_changed_files=$(git status --porcelain)
-          if [ -n "${git_changed_files}" ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request.yml
```diff
@@ -79,7 +79,28 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - name: Test with tox
-      run: tox
+      run: |
+        tox
+
+        # make sure python docs are up-to-date
+        if [[ "${{ matrix.python-version }}" == "3.8" ]]; then
```
XN137, line 86: we only do this for 3.8 in the matrix because that version already receives special treatment in tox:
https://github.com/projectnessie/nessie/blob/42342a27ecf2edc65eee1c2d4536656de67e47b5/python/tox.ini#L20-L25

snazy: Hm. Is it possible to add this check to tox.ini instead of a special case here?

snazy: Or better: Why not just unconditionally add `docs` to the gh-actions for 3.8?

XN137: i already tried to explain in the PR description why i did not want to put this into tox, but if we are fine with reasoning about git status in there then i will move it

XN137:
ready for re-review

.github/workflows/pull-request.yml
snazy: Yea, I think, that's okay.
If the generated docs are different from the checked-in docs, that's IMO a "bug".
I'd be even fine to move it into the `docs` part.

nastra (approved):

.github/workflows/pull-request.yml
XN137: > I'd be even fine to move it into the docs part.

then you will get an error everytime you run `tox -e docs` to update to the latest docs locally, so not a good idea imo

the "code" is simpler now but its only a valid up-to-date check in the following context:
- clean git repo state
- you run `docs` env just before

these conditions are mostly only guaranteed in CI, which is why i put the more verbose logic there.
its just a matter of different trade-offs, so i hope the current solution is fine now?

snazy: Hm, yea - that's practically a bit odd.

snazy (approved):

PR #3463, opened 2022-02-28
Shave off two database reads reads per Nessie operation

### LINKED ISSUE #3448: Shave off two read-ops per Nessie write operation
While working on #3447 I realized that we can save two database read operations per Nessie CAS-loop attempt (or: one database read for transactional databases).

To write a new global log entry, Nessie reads the parent hashes from the current global log entry.
This can be avoided by storing the parent hashes in the `GlobalStatePointer` as well (nothing to do for transactional databases).

To write a new ref log entry, Nessie reads the parent hashes from the current ref log entry.
This can be avoided by storing the parent hashes in the `GlobalStatePointer` as well (transactional database: add the parent hashes to `TABLE_REF_LOG_HEAD`).

The change can easily be implemented to be backwards compatible:
* If the list or parent hashes is not present, read the parent hashes like the current implementation does.
* If the "head" of the parent hashes is not equal to the expected "head", a concurrently running Nessie server updated the `GlobalStatePointer` (or `TABLE_REF_LOG_HEAD`) -> fall back to the old behavior and read the parent hashes like the current implementation does.

Example with create-reference:
Currently traced operations:
```
      VersionStore.create
          DatabaseAdapter.create
              DatabaseAdapter.try-loop.createRef
                  DatabaseAdapter.fetchGlobalPointer
                  DatabaseAdapter.fetchFromGlobalLog
                  DatabaseAdapter.writeGlobalCommit
                  DatabaseAdapter.fetchFromRefLog
                  DatabaseAdapter.writeRefLog
                  DatabaseAdapter.globalPointerCas
```

Expected traced operations after this change:
```
      VersionStore.create
          DatabaseAdapter.create
              DatabaseAdapter.try-loop.createRef
                  DatabaseAdapter.fetchGlobalPointer
                  DatabaseAdapter.writeGlobalCommit
                  DatabaseAdapter.writeRefLog
                  DatabaseAdapter.globalPointerCas
```

The change requires two new fields in `GlobalStatePointer` in `persist.proto`:
```
message GlobalStatePointer {
  bytes global_id = 1;
  repeated NamedReference named_references = 2;
  bytes ref_log_id = 3;
  // NEW FIELDS:
  repeated bytes global_parents = 4;
  repeated bytes ref_log_parents = 5;
}
```

snazy:
To write a new global log entry, Nessie reads the parent hashes from the current global log entry.
This can be avoided by storing the parent hashes in the `GlobalStatePointer` as well (nothing to do for transactional databases).

To write a new ref log entry, Nessie reads the parent hashes from the current ref log entry.
This can be avoided by storing the parent hashes in the `GlobalStatePointer` as well (transactional database: add the parent hashes to `TABLE_REF_LOG_HEAD`).

The change can easily be implemented to be backwards compatible:
* If the list or parent hashes is not present, read the parent hashes like the current implementation does.
* If the "head" of the parent hashes is not equal to the expected "head", a concurrently running Nessie server updated the `GlobalStatePointer` (or `TABLE_REF_LOG_HEAD`) -> fall back to the old behavior and read the parent hashes like the current implementation does.

Fixes #3448

### DIFF
--- a/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
+++ b/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
@@ -933,6 +933,16 @@ public abstract class NonTransactionalDatabaseAdapter<
     Hash parentHash = Hash.of(pointer.getGlobalId());
     Hash hash = randomHash();
 
+    // Before Nessie 0.21.0: the global-state-pointer contains the "heads" for the ref-log and
+    // global-log, so it has to read the head ref-log & global-log to get the IDs of all the
+    // previous parents to fill the parents in the new global-log-entry & ref-log-entry.
+    //
+    // Since Nessie 0.21.0: the global-state-pointer contains the "heads" for the ref-log and
+    // global-log PLUS the parents of those, so Nessie no longer need to read the head entries
+    // from the ref-log + global-log.
+    //
+    // The check of the first entry is there to ensure backwards compatibility and also
+    // rolling-upgrades work.
     Stream<ByteString> newParents;
     if (pointer.getGlobalParentsInclHeadCount() == 0
         || !pointer.getGlobalId().equals(pointer.getGlobalParentsInclHead(0))) {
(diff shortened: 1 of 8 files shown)


dimas-b (commented):
LGTM :+1: Waiting for the base tracing PR to be merged before approving this specific set of changes :)

nastra (approved):

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -932,32 +923,45 @@ protected abstract void doWriteRefLog(NonTransactionalOperationContext ctx, RefL
...
       List<ContentIdAndBytes> globals)
       throws ReferenceConflictException {
-    GlobalStateLogEntry currentEntry = fetchFromGlobalLog(ctx, parentHash);
 
-    Stream<Hash> newParents = Stream.of(parentHash);
-    if (currentEntry != null) {
+    Hash parentHash = Hash.of(pointer.getGlobalId());
+    Hash hash = randomHash();
+
+    Stream<ByteString> newParents;
```
tomekl007, line 936: can you elaborate more on this logic?  (Maybe ad a Javadoc comment)

dimas-b (dismissed):

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
snazy: added some inline docs

dimas-b (approved):

PR #3467, opened 2022-03-01
Fix merge issue for commits more than keylist distance

### LINKED ISSUE #3466: Merge fails if more than 20 commits are present on the source branch
* Start Nessie server with `quarkus:dev`
* Create a new branch `test1`
* Start a local Spark (1 node is sufficient)
* Make more that 20 commits in `test1`
* Attempt to merge `test1` to the default branch
* Observe an error similar to the following:
```
[...]
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.175 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.163 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.165 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.184 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.143 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.138 seconds
spark-sql> insert into t2 (id, data) values (1, 'a');
Time taken: 0.191 seconds
```
```
$ nessie  merge test1 -c 2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d
Client Error Not Found: Commit '2b2a3501e5c50ce7b5ba562d4a932c0f9509d08ab019953c89ec9491e0b1dea6' not found (Status code: 404)
Requested URL: http://localhost:19120/api/v1/trees/branch/main/merge?expectedHash=2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d
Server status: 404
Error code: REFERENCE_NOT_FOUND
Server message: Commit '2b2a3501e5c50ce7b5ba562d4a932c0f9509d08ab019953c89ec9491e0b1dea6' not found
Server traceback: None
```
Note: the hash from the error message is not visible in Nessie commit logs in CLI.
Note: the hash from the merge error is different in each new merge attempt without making any other commits.

ajantha-bhat:
`AbstractDatabaseAdapter#copyCommits` is preparing new commit hashes for merged commits.
During this time, for new commits that are merging, if the keylist size is reached, we are writing the new keylist to backend store [buildIndividualCommit-> buildKeyList].
While writing keylist we are using common code that queries commit log from backend db. But these new unwritten commit (we write all commits at once for merge to provide multi table txn) hashes are not present in the backend db. Hence the error.

**solution**: Hybrid approach
On the branch that we are merging, new keylist to write can be from both old commits and new commits (commits that are coming from merge) for this branch.
So, for new commits, collect key list from the inmemory list of commits instead of querying backend. For old commits, query from backend and combine the result of both and write new key list to backend db.

Fixes #3466

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -36,6 +36,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -194,7 +195,8 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
             commitAttempt.getPuts(),
             commitAttempt.getDeletes(),
             currentBranchEntry != null ? currentBranchEntry.getKeyListDistance() : 0,
-            newKeyLists);
+            newKeyLists,
+            new LinkedList<>());
     writeIndividualCommit(ctx, newBranchCommit);
     return newBranchCommit;
   }
@@ -737,7 +739,8 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
       List<KeyWithBytes> puts,
       List<Key> deletes,
       int currentKeyListDistance,
-      Consumer<Hash> newKeyLists)
+      Consumer<Hash> newKeyLists,
+      List<CommitLogEntry> unwrittenEntriesForKeyList)
       throws ReferenceNotFoundException {
     Hash commitHash = individualCommitHash(parentHashes, commitMeta, puts, deletes);
 
@@ -755,9 +758,33 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
             keyListDistance,
             null,
             Collections.emptyList());
-
     if (keyListDistance >= config.getKeyListDistance()) {
-      entry = buildKeyList(ctx, entry, newKeyLists);
+      // 'startHash' is the hash to be used for querying commit log from backend.
+      Hash startHash;
+      if (unwrittenEntriesForKeyList.isEmpty()) {
+        // query all commits from backend till the last keyList distance 0 commit.
+        // Because there is no previous unwritten entries (created from merge).
+        // So, set the hash to first ancestor of this entry.
+        startHash = entry.getParents().get(0);
+      } else if (entry.getParents().size() > unwrittenEntriesForKeyList.size()) {
+        // parents can have both written and unwritten commits.
+        // So, pick the startHash as the first written commit for backend query.
+        startHash = entry.getParents().get(unwrittenEntriesForKeyList.size());
+      } else if (entry.getParents().size() == unwrittenEntriesForKeyList.size()) {
+        // while merging many commits to fresh branch which is pointing to NO_ANCESTOR,
+        // parents can be equal to unwritten entries. So, no need to query backend.
+        startHash = NO_ANCESTOR;
+      } else {
+        // this cannot happen as per the merge logic
+        // as each unwritten entry will have one parent at least.
+        throw new RuntimeException(
+            String.format(
+                "Unwritten entries count: %s is greater than parents count: %s",
+                unwrittenEntriesForKeyList.size(), entry.getParents().size()));
+      }
+      entry = buildKeyList(ctx, entry, newKeyLists, startHash, unwrittenEntriesForKeyList.stream());
+      // remove the entries as key list is added for these entries.
+      unwrittenEntriesForKeyList.clear();
     }
     return entry;
   }
@@ -780,7 +807,7 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
     return Hash.of(UnsafeByteOperations.unsafeWrap(hasher.hash().asBytes()));
   }
 
-  /** Helper object for {@link #buildKeyList(Object, CommitLogEntry, Consumer)}. */
+  /** Helper object for {@link #buildKeyList(Object, CommitLogEntry, Consumer, Hash, Stream)}. */
   private static class KeyListBuildState {
     final ImmutableCommitLogEntry.Builder newCommitEntry;
     /** Builder for {@link CommitLogEntry#getKeyList()}. */
@@ -843,12 +870,12 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
    * to both read and re-write those rows for {@link KeyListEntity}.
    */
   protected CommitLogEntry buildKeyList(
-      OP_CONTEXT ctx, CommitLogEntry unwrittenEntry, Consumer<Hash> newKeyLists)
+      OP_CONTEXT ctx,
+      CommitLogEntry unwrittenEntry,
+      Consumer<Hash> newKeyLists,
+      Hash startHash,
+      Stream<CommitLogEntry> unwrittenEntries)
       throws ReferenceNotFoundException {
-    // Read commit-log until the previous persisted key-list
-
-    Hash startHash = unwrittenEntry.getParents().get(0);
-
     // Return the new commit-log-entry with the complete-key-list
     ImmutableCommitLogEntry.Builder newCommitEntry =
         ImmutableCommitLogEntry.builder().from(unwrittenEntry).keyListDistance(0);
@@ -856,7 +883,8 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
     KeyListBuildState buildState =
         new KeyListBuildState(entitySize(unwrittenEntry), newCommitEntry);
 
-    keysForCommitEntry(ctx, startHash)
+    // Read commit-log until the previous persisted key-list
+    keysForCommitEntry(ctx, startHash, unwrittenEntries)
         .forEach(
             keyWithType -> {
               int keyTypeSize = entitySize(keyWithType);
@@ -943,17 +971,22 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
   /** Retrieve the content-keys and their types for the commit-log-entry with the given hash. */
   protected Stream<KeyWithType> keysForCommitEntry(
       OP_CONTEXT ctx, Hash hash, KeyFilterPredicate keyFilter) throws ReferenceNotFoundException {
-    return keysForCommitEntry(ctx, hash)
+    return keysForCommitEntry(ctx, hash, Stream.empty())
         .filter(kt -> keyFilter.check(kt.getKey(), kt.getContentId(), kt.getType()));
   }
 
-  /** Retrieve the content-keys and their types for the commit-log-entry with the given hash. */
-  protected Stream<KeyWithType> keysForCommitEntry(OP_CONTEXT ctx, Hash hash)
+  /**
+   * Retrieve the content-keys and their types for the commit-log-entry with the given hash. And
+   * consider the unwrittenEntries along with the backend entries.
+   */
+  protected Stream<KeyWithType> keysForCommitEntry(
+      OP_CONTEXT ctx, Hash hash, Stream<CommitLogEntry> unwrittenEntries)
       throws ReferenceNotFoundException {
     // walk the commit-logs in reverse order - starting with the last persisted key-list
 
     Set<Key> seen = new HashSet<>();
     Stream<CommitLogEntry> log = readCommitLogStream(ctx, hash);
+    log = Stream.concat(unwrittenEntries, log);
     log = takeUntilIncludeLast(log, e -> e.getKeyList() != null);
     return log.flatMap(
         e -> {
@@ -1280,6 +1313,7 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
 
     int keyListDistance = targetHeadCommit != null ? targetHeadCommit.getKeyListDistance() : 0;
 
+    List<CommitLogEntry> unwrittenEntriesForKeyList = new LinkedList<>();
     // Rewrite commits to transplant and store those in 'commitsToTransplantReverse'
     for (int i = commitsChronological.size() - 1; i >= 0; i--, commitSeq++) {
       CommitLogEntry sourceCommit = commitsChronological.get(i);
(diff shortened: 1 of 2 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -194,7 +196,8 @@ protected CommitLogEntry commitAttempt(
             commitAttempt.getPuts(),
             commitAttempt.getDeletes(),
             currentBranchEntry != null ? currentBranchEntry.getKeyListDistance() : 0,
-            newKeyLists);
+            newKeyLists,
+            new LinkedList<>());
```
snazy, line 200: Why LinkedList? It's a plain list - ArrayList is more than sufficient.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
```diff
@@ -148,14 +166,15 @@ void apply(
...
+  private Hash[] mergeTransplant(int numCommits, MergeOrTransplant mergeOrTransplant)
+      throws Exception {
     BranchName main = BranchName.of("main");
     BranchName branch = BranchName.of("branch");
     BranchName conflict = BranchName.of("conflict");
 
     databaseAdapter.create(branch, databaseAdapter.hashOnReference(main, Optional.empty()));
 
-    Hash[] commits = new Hash[3];
+    Hash[] commits = new Hash[numCommits];
```
snazy, line 177: Since the prod code change touches the handling of content-keys/values, it's appropriate to add assertions for the merged content-key/values on every merged/tranplanted commit. Should also have some delete-operations in the generated commits.

snazy (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -943,18 +962,23 @@ protected void checkForModifiedKeysBetweenExpectedAndCurrentCommit(
...
+   */
+  protected Stream<KeyWithType> keysForCommitEntry(
+      OP_CONTEXT ctx, Hash hash, Stream<CommitLogEntry> unwrittenEntries)
       throws ReferenceNotFoundException {
     // walk the commit-logs in reverse order - starting with the last persisted key-list
 
     Set<Key> seen = new HashSet<>();
     Stream<CommitLogEntry> log = readCommitLogStream(ctx, hash);
     log = takeUntilIncludeLast(log, e -> e.getKeyList() != null);
+    log = Stream.concat(unwrittenEntries, log);
```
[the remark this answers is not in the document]
ajantha-bhat: I was sure that `unwrittenEntries` will have emptyKeyList(). so, I have added after the `takeUntilIncludeLast`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1280,6 +1304,7 @@ protected Hash copyCommits(
 
     int keyListDistance = targetHeadCommit != null ? targetHeadCommit.getKeyListDistance() : 0;
 
+    Queue<CommitLogEntry> unwrittenEntriesForKeyList = new LinkedList<>();
```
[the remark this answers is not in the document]
ajantha-bhat: wanted to maintain the order, I can keep a list and insert at head aswell.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: wanted to maintain the order, I can keep a list and insert at head aswell.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -757,7 +761,21 @@ protected CommitLogEntry buildIndividualCommit(
             Collections.emptyList());
 
     if (keyListDistance >= config.getKeyListDistance()) {
-      entry = buildKeyList(ctx, entry, newKeyLists);
+      Hash startHash;
+      if (unwrittenEntriesForKeyList.isEmpty()) {
+        // query all commits from backend.
+        startHash = entry.getParents().get(0);
+      } else if (entry.getParents().size() > unwrittenEntriesForKeyList.size()) {
```
snazy, line 768: This comparison is wrong. the number of parent commits in a commit-entry should always be equal to  config.getParentsPerCommit(). I do not see the relationship to the size of the new collection.

snazy (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -757,7 +761,21 @@ protected CommitLogEntry buildIndividualCommit(
             Collections.emptyList());
 
     if (keyListDistance >= config.getKeyListDistance()) {
-      entry = buildKeyList(ctx, entry, newKeyLists);
+      Hash startHash;
+      if (unwrittenEntriesForKeyList.isEmpty()) {
+        // query all commits from backend.
+        startHash = entry.getParents().get(0);
```
tomekl007, line 767: the comment says query all commits but we do `get(0);`. Shouldn't it iterate over it as well?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: Are you sure?

consider a scenario where we configured `config.getParentsPerCommit() = 20.`

I have a branch1 (pointing to NO_ANCESTOR hash) and I am merging 3 commits to branch1. In that case `entry.getParents().size()` can only be 1 to 3, it can never be 20.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: if the startHash points to head commit (aka `get(0)`), we iterate all the commits from startHash inside `buildKeyList -> keysForCommitEntry`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Again: what is the relationship between the number of parents (in all cases) stored in a commit-entry and the number of unwritten entries? This definitely needs to be elaborated more in the comments. It's quite hard to guess what this code's really doing.

ajantha-bhat: Also, KeyList distance can be different than getParentsPerCommit.
So, I need to handle keylist size <, =, > parents size

and that key list size may not be pure unwritten, it can have both written and unwritten commits.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
ajantha-bhat: moved

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
ajantha-bhat: Added validation and delete commit

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: Added comments and refactored a bit for better understanding.

ajantha-bhat:
@snazy : Can you please take another look? I have added comments and refactored a bit for better understanding.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1280,6 +1313,7 @@ protected Hash copyCommits(
 
     int keyListDistance = targetHeadCommit != null ? targetHeadCommit.getKeyListDistance() : 0;
 
+    List<CommitLogEntry> unwrittenEntriesForKeyList = new LinkedList<>();
```
snazy, line 1316: Still - no need for an expensive LinkedList here. ArrayList is sufficient

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
```diff
@@ -180,7 +204,25 @@ void apply(
 
       try (Stream<CommitLogEntry> targetLog =
           databaseAdapter.commitLog(databaseAdapter.hashOnReference(target, Optional.empty()))) {
-        assertThat(targetLog).hasSize(i + 1);
+        List<CommitLogEntry> targetLogEntries = targetLog.collect(Collectors.toList());
```
snazy, line 207: The added tests do assert the operations in the commits, which is nice, but not what _should_ be tested.
The keys and contents for every merged/transplanted commit via DatabaseAdapter.keys() + DatabaseAdapter.values() should be tested - that's what your change's touching.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
```diff
@@ -148,26 +167,31 @@ void apply(
...
               .commitMetaSerialized(ByteString.copyFromUtf8("commit " + i));
-      for (int k = 0; k < 3; k++) {
-        commit.addPuts(
-            KeyWithBytes.of(
-                Key.of("key", Integer.toString(k)),
-                ContentId.of("C" + k),
-                (byte) 0,
-                ByteString.copyFromUtf8("value " + i + " for " + k)));
+      if (i == commits.length - 1) {
+        commit.addDeletes(Key.of("key", Integer.toString(1)));
```
snazy, line 185: Should also have deletes for previously added keys - also in commits "in the middle". We really want to be sure that the change works as expected.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
```diff
@@ -74,8 +85,15 @@ void merge() throws Exception {
         .hasMessageStartingWith("No hashes to merge from '");
   }
 
-  @Test
-  void transplant() throws Exception {
+  @ParameterizedTest
+  @CsvSource({
```
snazy, line 89: `@ValueSource` is sufficient here

snazy (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: The plain `List` interface definition implies ordered entries too.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
dimas-b: TIL :+1:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: I'm not sure adding this complex logic is worth the benefit. Could we solve the problem by introducing a two-stage look up for commits involved in the key list construction: 1) from memory 2) if not found, from the database.

Callers can provide a real in-memory lookup function or a default no-op one.

We will not have to deal with all the complexities involved in configuring parent lists as longer or shorter than the "key distance". WDYT?

I'll try an open a PR for this quickly to illustrate my thinking.

dimas-b (changes requested):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
ajantha-bhat: changed.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
ajantha-bhat: I added some validations for keys but I am finding hard to add it in the current testcase. I find easy to write my own testcase for this :(

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
ajantha-bhat: deleting in the middle now.

ajantha-bhat:
> I'll try an open a PR for this quickly to illustrate my thinking.

@dimas-b : Please go ahead and open a PR if you have better/clean approach. I can close this PR. Thanks.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractMergeTransplant.java
ajantha-bhat: changed.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: This is what I mean: #3486 if it looks ok in general, I will polish it of course and include feedback from this PR.

ajantha-bhat: sure. Go ahead. it looks good to me. I will close this PR.

ajantha-bhat:
#3486 is cleaner than this. Hence closing this.

PR #3481, opened 2022-03-01
Add command to read commits

nastra:


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -0,0 +1,75 @@
...
+      names = {"-r", "--ref"},
+      description = "Name of the branch/tag to read the commit log from, defaults to 'main'")
+  private String ref = "main";
+
+  @Spec private CommandSpec spec;
+
+  @Override
+  public void execute() throws NessieNotFoundException {
+    try (NessieApiV1 api = createNessieApiInstance()) {
+      spec.commandLine().getOut().printf("Reading all commits\n\n");
```
snazy, line 45: Maybe guard it with `if (verbose()` and add the source-ref to the logged message?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/AbstractCommand.java
```diff
@@ -36,6 +36,15 @@
       description = "Custom implementation of org.projectnessie.client.NessieClientBuilder.")
   private Class<?> customBuilder;
 
+  @Option(
+      names = {"-v", "--verbose"},
+      description = "Produce verbose output (if possible)")
+  private boolean verbose = false;
+
+  protected boolean verbose() {
```
snazy, line 44: Nit: `isVerbose`

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
snazy, line 45: "all" is probably wrong here, because it shows only the first returned page but not the whole commit log.

snazy (commented):
What's the benefit of adding "read commits" here vs having the functionality in pynessie?

nastra:
> What's the benefit of adding "read commits" here vs having the functionality in pynessie?

The benefit is that you don't have to setup a python env

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -0,0 +1,75 @@
...
+            .getOut()
+            .printf(
+                "%s\t%s\t%s [%s]\n",
+                Objects.requireNonNull(commitMeta.getHash()).substring(0, 8),
+                commitMeta.getAuthorTime(),
+                commitMeta.getMessage(),
+                commitMeta.getAuthor());
+
+        List<Operation> operations = logEntry.getOperations();
+        if (operations != null) {
```
tomekl007, line 60: is it normal that there can be no operations? (`operations==null`)?
Maybe we should log that if it is in verbose mode?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Override
+  public void execute() throws NessieNotFoundException {
+    try (NessieApiV1 api = createNessieApiInstance()) {
+      spec.commandLine().getOut().printf("Reading commits for ref '%s'\n\n", ref);
+      FetchOption fetchOption = isVerbose() ? FetchOption.ALL : FetchOption.MINIMAL;
+      LogResponse logResponse = api.getCommitLog().refName(ref).fetch(fetchOption).get();
+      for (LogResponse.LogEntry logEntry : logResponse.getLogEntries()) {
+        CommitMeta commitMeta = logEntry.getCommitMeta();
+        spec.commandLine()
```
tomekl007, line 50: Design question regarding direct usage of `System.out` and `printf`
Did we think about re-designing this code to construct and return objects and the formatting and printing to be handled in a separate component? 
So for example, this logic could construct a `List` of `ReadOutput` and a separate component could handle printing it out. Maybe we could make the printing logic more generic and work for other commands (for example https://github.com/projectnessie/nessie/pull/3482/files).
Also, it could simplify testing, so the println testing would be done only in one place.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -0,0 +1,75 @@
...
+  private String ref = "main";
+
+  @Spec private CommandSpec spec;
+
+  @Override
+  public void execute() throws NessieNotFoundException {
+    try (NessieApiV1 api = createNessieApiInstance()) {
+      spec.commandLine().getOut().printf("Reading commits for ref '%s'\n\n", ref);
+      FetchOption fetchOption = isVerbose() ? FetchOption.ALL : FetchOption.MINIMAL;
+      LogResponse logResponse = api.getCommitLog().refName(ref).fetch(fetchOption).get();
```
snazy, line 47: Should use `StreamingUtil.getCommitLogStream` to get all commits - otherwise you'll only get/print the first returned page.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -0,0 +1,75 @@
...
+      for (LogResponse.LogEntry logEntry : logResponse.getLogEntries()) {
```
snazy, line 48: The (total) number of commits can be huge. I think, this needs at least some rudimentary paging similar to the `log` command in pynessie.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
nastra: I'll add this as a follow-up

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
nastra: I'll add this as a follow-up

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
nastra: I improved this already in the next PR (just didn't push it yet). There's definitely things that can be improved but out of scope for this PR I would say.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
nastra: yes that's normal, `operations` will only be there when verbose mode is used

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
nastra: I'll add this as a follow-up

snazy (approved):

PR #3482, opened 2022-03-01
Add cmd to read all refs

nastra:


tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITReadReferences.java
```diff
@@ -0,0 +1,42 @@
...
+  void readReferences() throws UnsupportedEncodingException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    PrintStream out = new PrintStream(baos);
+    System.setOut(out);
+
+    assertThat(NessieContentGenerator.runMain(new String[] {"refs", "--uri", NESSIE_API_URI}))
+        .isEqualTo(0);
+
+    out.close();
+    String[] output = baos.toString(StandardCharsets.UTF_8.toString()).split("\n");
```
XN137, line 39: this pattern for capturing output looks bad...

the `CommandLine` class in `runMain` has methods `setOut` and `setErr` which seem like the right way to capture output in tests, wdyt?

if this does not work, then at least we should have a helper method (in base class) that does all this capturing ceremony.
as is we duplicate this boilerplate in 4 places now.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadReferences.java
```diff
@@ -0,0 +1,40 @@
...
+@Command(name = "refs", mixinStandardHelpOptions = true, description = "Read references")
+public class ReadReferences extends AbstractCommand {
+
+  @Spec private CommandSpec spec;
+
+  @Override
+  public void execute() {
+    try (NessieApiV1 api = createNessieApiInstance()) {
+      spec.commandLine().getOut().printf("Reading all references\n\n");
+      List<Reference> references = api.getAllReferences().get().getReferences();
```
snazy, line 35: Should use `StreamingUtil.getAllReferencesStream()` - probably not relevant now, but likely in the future.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
-                api.createReference()
-                    .reference(Branch.of(defaultBranchName, defaultBranch.getHash()))
-                    .sourceRefName(defaultBranch.getName())
-                    .create();
+      } else {
+        // Use the specified default branch.
+        try {
+          defaultBranch = (Branch) api.getReference().refName(defaultBranchName).get();
+        } catch (NessieReferenceNotFoundException e) {
+          // Create branch if it does not exist.
```
tomekl007, line 141: is it safe to create a branch automatically? If yes, then I think we should document that.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+                  api.createReference()
+                      .reference(Branch.of(defaultBranchName, defaultBranch.getHash()))
+                      .sourceRefName(defaultBranch.getName())
+                      .create();
+        }
+      }
+
+      List<String> branches = new ArrayList<>();
+      branches.add(defaultBranch.getName());
+      while (branches.size() < branchCount) {
```
tomekl007, line 154: in what use case may someone want to use `branchCount > 1`?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+        // Create a new branch
+        String newBranchName = "branch-" + runStartTime + "_" + (branches.size() - 1);
```
tomekl007, line 156: are we removing this branch after this logic?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
-    }
-
-    spec.commandLine().getOut().printf("Starting contents generation, %d commits...%n", numCommits);
+          .printf("Starting contents generation, %d commits...%n", numCommits);
 
-    for (int i = 0; i < numCommits; i++) {
-      // Choose a random branch to commit to
-      String branchName = branches.get(random.nextInt(branches.size()));
+      for (int i = 0; i < numCommits; i++) {
+        // Choose a random branch to commit to
```
tomekl007, line 172: why do we want to introduce randomness here?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadReferences.java
nastra: yep I'll add this as a follow-up once the commands are in place/merged

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+                    CommitMeta.builder()
+                        .message(
+                            String.format(
+                                "%s table %s on %s, commit #%d of %d",
+                                tableContents != null ? "Update" : "Create",
+                                tableName,
+                                branchName,
+                                i,
+                                numCommits))
+                        .author(System.getProperty("user.name"))
```
tomekl007, line 202: should we have null check for: `System.getProperty("user.name")`?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+                        .authorTime(Instant.now())
+                        .build());
+        if (newContents instanceof IcebergTable || newContents instanceof IcebergView) {
+          commit.operation(Put.of(tableName, newContents, tableContents));
+        } else {
+          commit.operation(Put.of(tableName, newContents));
+        }
+        Branch newHead = commit.commit();
+
+        if (random.nextDouble() < newTagProbability) {
```
tomekl007, line 212: again, why the randomness? Is it for performance reasons to not do it too often?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+          Tag tag = Tag.of("new-tag-" + random.nextLong(), newHead.getHash());
+          spec.commandLine()
+              .getOut()
+              .printf(
+                  "Creating tag '%s' from '%s' at %s%n", tag.getName(), branchName, tag.getHash());
+          api.createReference().reference(tag).sourceRefName(branchName).create();
+        }
+
+        try {
+          TimeUnit.NANOSECONDS.sleep(perCommitDuration.toNanos());
```
tomekl007, line 222: how the `perCommitDuration` is picked? And why do we need to sleep here?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -130,98 +128,102 @@ public void execute() throws BaseNessieClientServerException {
...
+        } catch (InterruptedException e) {
+          Thread.currentThread().interrupt();
```
tomekl007, line 224: again, a design question. Did we consider returning a `CompletableFuture` from the `execute` and let the caller decide whether this action should be blocking `CompletableFuture.get()` or not (chain subsequent actions)?

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITReadCommits.java
```diff
@@ -0,0 +1,98 @@
...
+    System.setOut(out);
+
+    assertThat(
+            NessieContentGenerator.runMain(
+                new String[] {"commits", "--uri", NESSIE_API_URI, "--ref", branch.getName()}))
+        .isEqualTo(0);
+
+    out.close();
+    String[] output = baos.toString(StandardCharsets.UTF_8.toString()).split("\n");
+    assertThat(output).anySatisfy(s -> assertThat(s).contains("testMessage"));
```
tomekl007, line 61: when looking at this test, it is not clear from where the `testMessage` and `first.second` come from.
Can we document that OR extract this special value and use it everywhere? (it would be easier to detect why and when it is used)

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
nastra: this code hasn't been touched by this PR

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITReadCommits.java
```diff
@@ -69,23 +70,22 @@ void readCommits() throws UnsupportedEncodingException, NessieNotFoundException
...
+          .isEqualTo(0);
 
-    out.close();
-    String[] output = baos.toString(StandardCharsets.UTF_8.toString()).split("\n");
-    assertThat(output).anySatisfy(s -> assertThat(s).contains("testMessage"));
-    assertThat(output).anySatisfy(s -> assertThat(s).contains("first.second"));
-    assertThat(output).anySatisfy(s -> assertThat(s).contains("key[0]: first"));
-    assertThat(output).anySatisfy(s -> assertThat(s).contains("key[1]: second"));
-    assertThat(output).anySatisfy(s -> assertThat(s).contains(contentId));
+      String[] output = baos.toString(StandardCharsets.UTF_8.toString()).split("\n");
```
XN137, line 82: the test output capturing is better than before (no more `System.setOut`), but imo still not minimal ceremony...
but we can improve that in a followup i guess.

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITReadCommits.java
```diff
@@ -69,23 +70,22 @@ void readCommits() throws UnsupportedEncodingException, NessieNotFoundException
...
+      assertThat(output).anySatisfy(s -> assertThat(s).contains("testMessage"));
+      assertThat(output).anySatisfy(s -> assertThat(s).contains("first.second"));
+      assertThat(output).anySatisfy(s -> assertThat(s).contains("key[0]: first"));
+      assertThat(output).anySatisfy(s -> assertThat(s).contains("key[1]: second"));
+      assertThat(output).anySatisfy(s -> assertThat(s).contains(contentId));
```
XN137, line 87: nit: slightly odd to split output by lines... only to the then confirm ANY line contains the given output

XN137 (approved):

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITReadCommits.java
nastra: yep, let's improve in a follow-up

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/AbstractContentGeneratorTest.java
```diff
@@ -38,11 +37,6 @@
 
   private static final PrintStream DEFAULT_STDOUT = System.out;
```
dimas-b, line 38: Do we need this variable anymore?

PR #3484, opened 2022-03-01
[RELEASE] Require released version being exercised in compatibility tests

snazy:
Ensure that the version being released is mentioned in `compatibility/compatibility-tests/src/test/resources/META-INF/nessie-compatibility.properties` and therefore exercised in compatibility tests.


snazy:
Meh - cannot work, as CI cannot vet against an unreleased version :disappointed:

PR #3488, opened 2022-03-02
Add utility methods to Version class for easier version comparisons

nastra:


compatibility/common/src/main/java/org/projectnessie/tools/compatibility/api/Version.java
```diff
@@ -80,6 +80,22 @@ private static int parsePart(String part) {
     return val;
   }
 
+  public boolean isGreaterThan(Version version) {
```
snazy, line 83: Mind adding a `isSame` using `compare`, too? It's not the same, because trailing `.0` are "ignored" - i.e. `"1.0".compareTo("1.0.0") == 0`.

nastra: done

snazy (approved):

PR #3493, opened 2022-03-02
Remove lazyhash=true from ContentKey

nastra:
Using `lazyhash=true` makes the `hashcode` field in the generated class
non-transient, which can lead to ser/de errors.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/ContentKey.java
+++ b/model/src/main/java/org/projectnessie/model/ContentKey.java
@@ -36,7 +36,7 @@ import org.projectnessie.model.ImmutableContentKey.Builder;
  * <p>For URL encoding, embedded periods within a segment are replaced with zero byte values before
  * passing in a url string.
  */
-@Value.Immutable(lazyhash = true)
+@Value.Immutable
 @JsonSerialize(as = ImmutableContentKey.class)
 @JsonDeserialize(as = ImmutableContentKey.class)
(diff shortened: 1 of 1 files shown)


ajantha-bhat:
I got review comment to remove contentKey from GC output and add it later with other values. 
But I forgot to revert serializable for the same. 
I may add back contentKey in the future. That time I have to make serializable.

snazy (approved):

model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -36,7 +36,7 @@
  * <p>For URL encoding, embedded periods within a segment are replaced with zero byte values before
  * passing in a url string.
  */
-@Value.Immutable(lazyhash = true)
+@Value.Immutable
 @JsonSerialize(as = ImmutableContentKey.class)
 @JsonDeserialize(as = ImmutableContentKey.class)
 public abstract class ContentKey implements Serializable {
```
ajantha-bhat, line 42: you can remove `Serializable`  also from `ContentKey` if you want. Later when I really need it for GC, I will add it back.

nastra: I think this is actually a bug in the Immutables lib. Also we want to eventually make all model classes Serializable, so IMO the more correct solution currently is to remove `lazyhash = true`

ajantha-bhat (approved):

PR #3510, opened 2022-03-03
Create Grafana Dashboard for Nessie service

nastra:

PR #3511, opened 2022-03-03
Reduce number of tags for micrometer

tomekl007:
Using the mechanism described in: https://access.redhat.com/documentation/en-us/red_hat_build_of_quarkus/1.11/html/collecting_metrics_in_your_quarkus_applications/con-http-metrics_quarkus-collecting-metrics

It reduces the number of tags reported by the metrics registry, by using `{path_parameters}`.

After the fix, tags look like this:
```
0 = "/api/v1/diffs/{diff_params}"
1 = "/api/v1/trees/tree/{ref}/entries"
2 = "/api/v1/trees"
3 = "/api/v1/trees/tree/commitLogExtended"
4 = "/api/v1/contents"
5 = "/api/v1/trees/tree"
6 = "/api/v1/trees/tag/testTag"
7 = "/api/v1/trees/tree/commitLogExtendedUnchanged"
8 = "/api/v1/trees/tree/main"
9 = "/api/v1/trees/branch/{ref}"
10 = "/api/v1/trees/tree/testTag"
11 = "/api/v1/trees/tree/testBranch"
12 = "/api/v1/trees/branch/{branchName}/commit"
13 = "/api/v1/trees/tree/{ref}/log"
```
so the number of tags is substantially reduced and the warning is no longer emitted.

### DIFF
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/ITMetrics.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/ITMetrics.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.restassured.RestAssured;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.jaxrs.AbstractTestRest;
+import org.projectnessie.server.profiles.QuarkusTestProfileInmemory;
+
+@QuarkusTest
+@TestProfile(value = QuarkusTestProfileInmemory.class)
+public class ITMetrics extends AbstractTestRest {
+  // We need to extend the AbstractTestRest because all Nessie metrics are created lazily.
+  // They will appear in the `/q/metrics` endpoint only when some REST actions are executed.
+
+  // this test is executed after all tests from the AbstractTestRest
+  @Test
+  void smokeTestMetrics() {
+    // when
+    String body =
+        RestAssured.given()
+            .when()
+            .basePath("/q/metrics")
+            .get()
+            .then()
+            .statusCode(200)
+            .extract()
+            .asString();
+
+    // then
+    assertThat(body).contains("jvm_threads_live_threads");
+    assertThat(body).contains("jvm_memory_committed_bytes");
+    assertThat(body).contains("nessie_versionstore_request_seconds_max");
+    assertThat(body).contains("nessie_versionstore_request_seconds_bucket");
+    assertThat(body).contains("nessie_versionstore_request_seconds_count");
+    assertThat(body).contains("nessie_versionstore_request_seconds_sum");
+    assertThat(body).contains("/api/v1/diffs/{diff_params}");
+    assertThat(body).contains("/api/v1/trees/{referenceType}/{ref}");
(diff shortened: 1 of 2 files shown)


snazy (commented):
Approach looks good. But I think there are a couple more places like
```
/content/{key}
/trees//tree/{refName}  (getReferenceByName)
/trees//tree/{refType}/{refName}    (assign/delete, can be a "tag" as well)
/trees//tree/branch/{refName}/transplant
/trees//tree/branch/{refName}/merge
```

nastra (commented):
Can we add some simple test to make sure we're getting all of the required metrics from https://github.com/projectnessie/nessie/pull/3510?


Something like:
```
@QuarkusTest
@TestProfile(value = QuarkusTestProfileInmemory.class)
public class ITTestMetrics {

  @Test
  void smokeTestMetrics() {
    String body = RestAssured.given()
      .when()
      .basePath("/q/metrics")
      .get()
      .then()
      .statusCode(200)
      .extract()
      .asString();

    assertThat(body).contains("jvm_threads_live_threads");
    // ... verify the other important metrics
  }
}
```

tomekl007:
> Approach looks good. But I think there are a couple more places like
> 
> ```
> `
> /trees//tree/{refName}  (getReferenceByName)
> /trees//tree/{refType}/{refName}    (assign/delete, can be a "tag" as well)
> /trees//tree/branch/{refName}/transplant
> /trees//tree/branch/{refName}/merge
> ```

I've added additional mappings, however, I am nos rue about the `/content/{key}`. Where it is used and created?

snazy:
> I've added additional mappings; however, I am not sure about the `/content/{key}`. Where is it used and created?

It's in org.projectnessie.api.http.HttpContentApi#getContent

snazy:
Out of scope and maybe something for later, but is there a way to automatically generate these configs from an openapi spec?

tomekl007:
> Out of scope and maybe something for later, but is there a way to automatically generate these configs from an openapi spec? Or maybe even just with some custom annotation processor (that we'd maybe have to build)?

We are hitting this: https://github.com/quarkusio/quarkus/issues/21034 so once it will be resolved, the mapping could be removed. So I am not sure if we want to create a mechanism for generating that.

snazy:
> We are hitting this: [quarkusio/quarkus#21034](https://github.com/quarkusio/quarkus/issues/21034) so once it will be resolved, the mapping could be removed. So I am not sure if we want to create a mechanism for generating that.

Ah! GTK!

snazy (dismissed):

nastra (dismissed):

tomekl007:
> LGTM!
> 
> Can you open a revert-PR that we can then just merge once a Quarkus version with the fix gets released? The PR as a "permanent reminder".

How do we prevent someone from accidentally merging such a PR? Maybe it's better to create a follow-up ticket?

snazy (approved):

servers/quarkus-server/src/test/java/org/projectnessie/server/ITMetrics.java
```diff
@@ -0,0 +1,62 @@
...
+            .when()
+            .basePath("/q/metrics")
+            .get()
+            .then()
+            .statusCode(200)
+            .extract()
+            .asString();
+
+    // then
+    assertThat(body).contains("jvm_threads_live_threads");
```
nastra, line 48: seems ok for now. We'll follow up with whatever other metrics we want to verify in a separate PR, so let's merge what we have

nastra (approved):

servers/quarkus-server/src/test/java/org/projectnessie/server/ITMetrics.java
snazy: Just open it as a draft, nobody will/should touch it then

ebullient:
I am having some trouble reproducing the issue with the provided reproducer. In your case, how are you making outbound Http requests? The current support looks for usage of the Rest Client (which declares templated URIs in a way we can look for at build time).

ebullient:
Thank you! This is server side, not client side, right? (received requests, rather than outbound)?

imjuoy:
I am also facing a similar issue of higher number of URI tags. Since I am planning to follow the same path that you guys have followed, I have a question regarding having multiple path parameters.
Everywhere I refer for examples of configurations, it's mostly just one path param. For example:

https://quarkus.io/guides/all-config#quarkus-micrometer_quarkus.micrometer.binder.http-server.match-patterns

https://github.com/tomekl007/nessie/blob/be4e348f4b1b4e12453912cb594a4781dc73a5d6/servers/quarkus-server/src/main/resources/application.properties#L173

In our case, the application has three path parameters. Let's say:
/v1/api/param1/param2/param3

Param2 has limited set of values and will never lead to high cardinality situations.
Is it possible to just record param2 and keep param1 and param3 templatized?

PR #3512, opened 2022-03-03
Allow IntelliJ to property generate proto sources

snazy:
Splits out the .proto files and related code generation to separate modules
without dependencies to not-yet-built artifacts.

IntelliJ's "Generate Sources and Update Folders" action won't work, if the
module for which sources should be generated (think: `process-sources` Maven
goal) has not-yet available dependencies.


CONTRIBUTING.md
```diff
@@ -31,6 +31,17 @@ on the new feature or improvement.
 
 ## Code changes
 
+### IntelliJ IDEA tips
+
+Nessie uses antlr and protoc to generate source files. Those files are needed to build the Nessie project.
+
+Tip: In the "Project" view (usually on the left) right-click on the `[nessie]` project, open the
+`Maven` sub-menu (at the bottom of the context menu), click on `Generate Sources and Update Folders`.
+This will generate the protobuf source files that necessary to build Nessie via IntelliJ IDEA.
```
XN137, line 40: grammar: "that necessary"

CONTRIBUTING.md
```diff
@@ -31,6 +31,17 @@ on the new feature or improvement.
...
+
+Note: Using the above method via the project's context menu is much quicker than running
+`Generate Sources and Update Folders for all projects` from the Maven view (usually on the right).
```
XN137, line 43: does the tip above mean to select a specific sub-project?
otherwise it is not immediately clear how "all projects" would be different from doing the same on the root project :sweat_smile: 
if its odd intellj behavior we should maybe mention that.

XN137 (approved):
typo in commit msg / PR title: `propertly`

apart from that LGTM

so basically put code generation logic into separate modules with minimal/no dependencies to other modules

CONTRIBUTING.md
snazy: `Generate Sources and Update Folders for all projects` is the name of the IntelliJ action

XN137: i know, my point was that its not immediately (at least to me) why running the action on the root project on the left has different behavior than running it on "all projects" on the right (since intellij only has a single project open and all project means all maven module afaict). no big deal.

XN137 (approved):
`propertly` typo still there

still lgtm

PR #3513, opened 2022-03-03
Use externally maintained nessie-antlr-runtime

snazy:
Uses the pre-generated and released `nessie-antlr-runtime` artifact instead of shading/relocating antlr runtime in the Nessie build, which makes it easier for IDEs to build w/o Maven.

Removes the "fake version workaround" to let IntelliJ import the project.

Updates the spark-grammar module to not use relocation but update the package name in the generated antlr java files.

With the above changes, it's possible to use `Maven --> Generate Sources and Update Folders` from the context menu of the top-level nessie module in IntelliJ IDEA project view.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,7 @@
     <maven.resolver.version>1.7.3</maven.resolver.version>
     <maven.version>3.8.4</maven.version>
     <mockito.version>4.3.1</mockito.version>
+    <nessie.antlr.version>4.9.2</nessie.antlr.version>
     <nessie.version>${project.version}</nessie.version>
     <openapi.version>3.0</openapi.version>
     <picocli.version>4.6.3</picocli.version>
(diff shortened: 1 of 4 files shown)


snazy:
Note: Intentionally used antlr 4.9.2 instead of 4.9.3 to also verify the upgrade-scenario from 4.9.2 to 4.9.3

pom.xml
```diff
@@ -196,6 +196,7 @@
     <maven.resolver.version>1.7.3</maven.resolver.version>
     <maven.version>3.8.4</maven.version>
     <mockito.version>4.3.1</mockito.version>
+    <nessie.antlr.version>4.9.2</nessie.antlr.version>
```
XN137, line 199: since we still use `antlr4-maven-plugin` we need to keep the `antlr.version` property, but does it mean we need to keep those versions in sync or whats the rule for updating one of the two in the future (i.e. `major.minor` version needs to stay the same)?

if there is any relationship we should imo move this line next to the antlr version and add a comment there.

XN137 (approved):
i assume you tested this also for maven builds in intellij

generally lgtm with minor comment/question

also maybe we want a developer readme somewhere describing these auxiliary projects (detailing where to find them and why they were introduced)

pom.xml
snazy: Yea - it's not strictly required and this property shall go away.
I had to add it to try the version-bump-scenario.
Will remove this property soon.

PR #3529, opened 2022-03-04
Add Nessie repos to README and tweak build intructions

snazy:


README.md
```diff
@@ -127,6 +130,13 @@ are specified as per https://github.com/eclipse/microprofile-config/blob/master/
 You can also deploy to AWS lambda function by following the steps in `servers/lambda/README.md`
  
 
+## Nessie related repositories
+
+* [Nessie Demos](https://github.com/projectnessie/nessie-demos): Demos for Nessie
+* [CEL Java](https://github.com/projectnessie/cel-java): Java port of the Common Expression Language
+* [Nessie antlr runtime](https://github.com/projectnessie/nessie-antlr-runtime): Shaded/relocated antlr runtime for Nessie Spark Extensions
```
XN137, line 137: wondering if its worth mentioning the intellj tickets we are working around (because without those we could be relocating on the fly probably)

XN137 (approved):
thanks for adding this!

lgtm

README.md
snazy: Not really. The reason (Maven not able to provide a meaningful build model for generated sources incl shading/relocation) for these external repos is "too far away" from a general README.

PR #3544, opened 2022-03-07
Use Immutable Builder functionality in HTTP Param classes to generate Builders

nastra:
The only reason we can't fully move those classes to true Immutables is
because of bean validation requiring those classes to have a default
constructor. Note that bean validation on those classes is done via
`@BeanParam`, such as in `getXyz(@BeanParam EntriesParams params)`.

At the very least, we can remove the hand-written builders and let the
Immutables lib do the work of generating those (since we already have
tests in place that make sure objects are properly constructed and
validated).


XN137 (approved):
changes lgtm so +1 assuming i can trust that we already have test coverage ;)

nastra:
> changes lgtm so +1 assuming i can trust that we already have test coverage ;)

For completeness, the existing test code lives in https://github.com/projectnessie/nessie/tree/f77aa64fd727e9768f338a93bf855170b84f6001/model/src/test/java/org/projectnessie/api/params

PR #3556, opened 2022-03-08
Remove unused 'type' + "type-extractor" functions

snazy:
The `type` field stored in the global-log was never really used and also always `0`.

This change removes the `type` field from the immutable-types, renames it in the serialized representation, and removes the "type extractor" arguments from the respective functions in `DatabaseAdapter`.

Note: the `type` field for "on reference" contents / content keys is correct and used.


versioned/persist/serialize-proto/src/main/proto/persist.proto
```diff
@@ -60,12 +60,12 @@ message ContentId {
 
 message ContentIdWithType {
   ContentId content_id = 1;
-  int32 type = 2;
+  int32 type_unused = 2; // ignore the value of this field
```
nastra, line 63: will this potentially cause issues after an upgrade?

snazy: Nope. protobuf uses the int values - not the names

nastra (approved):

PR #3557, opened 2022-03-08
Add delta test for nessie sql extensions

XN137:
- improves delta test cleanup
- improves error message in nessie catalog check

this is part 1 of https://github.com/projectnessie/nessie/issues/3552 were we add test coverage

PR #3558, opened 2022-03-08
Fix nessie catalog check for delta

### LINKED ISSUE #3552: Delta tests and docs dont use `NessieSparkSessionExtensions`
in the delta demo we are using the SQL nessie extensions:
https://github.com/projectnessie/nessie-demos/blame/099cb4a0b1e8a92e456770edeb909bcb4e1b1e7a/notebooks/nessie-delta-demo-nba.ipynb#L61

however in our delta tests and docs we do not.

Delta tests configuration:
https://github.com/projectnessie/nessie/blob/d36e031431f694ee83de6a7877c0417ad79ef94f/clients/deltalake/src/test/java/org/projectnessie/deltalake/AbstractDeltaTest.java#L99-L103

delta spark docs:
https://github.com/projectnessie/nessie/blame/dcd1406a4ebc95e46fd65c83b5bb6a4efe04a966/site/docs/tools/deltalake/spark.md#L43-L44

both should be changed to show and test all of the capabilities in a delta context.

XN137:
Follow-up to https://github.com/projectnessie/nessie/pull/3557

Fixes #3552

Original problem:
After https://github.com/projectnessie/nessie/pull/2377 the nessie spark sql extensions could no longer be used with delta.

Also updates the deltalake spark docs removing "spark 2" references and adding info about the nessie spark sql extenions.

### DIFF
--- a/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
+++ b/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
@@ -164,22 +164,30 @@ object NessieUtils {
       catalog: Option[String]
   ): NessieApiV1 = {
     val catalogName = catalog.getOrElse(currentCatalog.name)
-    val catalogConf = SparkSession.active.sparkContext.conf
+    val sparkConf = SparkSession.active.sparkContext.conf
+    val catalogConf = sparkConf
       .getAllWithPrefix(s"spark.sql.catalog.$catalogName.")
       .toMap
-    // Referring to https://github.com/apache/iceberg/blob/master/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
-    // Not using fully-qualified class name to provide protection from shading activities (if any)
-    val catalogImpl = catalogConf.get("catalog-impl")
-    val catalogInfo = catalogImpl match {
-      case Some(clazz) => s"but $catalogName is a $clazz"
-      case None =>
-        s"but spark.sql.catalog.$catalogName.catalog-impl is not set"
+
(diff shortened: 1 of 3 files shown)


clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
```diff
@@ -164,22 +164,30 @@ object NessieUtils {
...
+        case Some(clazz) => s"but $catalogName is a $clazz"
+        case None =>
+          s"but spark.sql.catalog.$catalogName.catalog-impl is not set"
+      }
+      // Referring to https://github.com/apache/iceberg/blob/master/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java
+      // Not using fully-qualified class name to provide protection from shading activities (if any)
+      require(
+        catalogImpl
+          .exists(impl => impl.endsWith(".NessieCatalog")),
+        s"The command works only when the catalog is a NessieCatalog ($catalogErrorDetail). Either set the catalog via USE <catalog_name> or provide the catalog during execution: <command> IN <catalog_name>."
```
XN137, line 187: Note that we have tests to make sure this still works in the iceberg case:
https://github.com/projectnessie/nessie/blob/8ba7079c54325c90e39ebf4e234c014711ca0e8e/clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java#L703-L721

I could not use the same style of testing for delta because i didnt know which non-delta catalog to use (no iceberg jar on classpath).

nastra (approved):

XN137:
tests are passing but codecov is reporting decreased coverage on untouched files... so I will just ignore that check

PR #3421, opened 2022-02-23
Store GC identify action's results into a table

ajantha-bhat:
Currently, output of `GCImpl#identifyExpiredContents` is stored in an in memory map by merging the results from each spark task.

The results can be stored in a table instead, to run clean up afterwards or to be used by consumer of a different process. 

With this PR, Spark is directly writing the task output into the iceberg table in a configured reference where each task's output will be mapped to multiple rows in a table (one row per content id).

`GCImpl#identifyExpiredContents` now returns the run id of the completed task. Using which user can query the output table. Also added an interface to fetch the latest completed run id.

All old gc test cases now query the output table to collect the results and validates it.
Added a new test case to test in-progress and marker row write to the table.

Also removed `Serializable` from `Content` along with removing intermediate classes (`contentValues`,`IdentifiedResult`)that was used for storing output in-memory map.

PR #3561, opened 2022-03-09
Remove Serializable from Reference

### LINKED ISSUE #3496: Remove `Serializable` from `Reference` + `Content`
We already have JSON serialization and should rely only on that method for serialization and not use/support Java serialization for the Nessie model.

ajantha-bhat:
We already have JSON serialization and should rely only on that method for serialization and not use/support Java serialization for the Nessie model.

Also for using these objects with spark dataframe (with java serialization), spark code-gen will fail because these objects don't have public constructor. Adding pubic constructor gives error in Immutables as these classes have validate annotation. 
So, better to remove java serialization for immutables objects and using Immutables built in JSON serialization. 

This partially fixes #3496

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
@@ -16,12 +16,12 @@
 package org.projectnessie.gc.base;
 
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.OptionalInt;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.spark.sql.SparkSession;
 import org.projectnessie.api.params.FetchOption;
@@ -100,8 +100,12 @@ public class GCImpl {
       DistributedIdentifyContents distributedIdentifyContents =
           new DistributedIdentifyContents(session, gcParams);
       List<Reference> liveReferences = api.getAllReferences().get().getReferences();
-      Map<Reference, Instant> droppedReferenceTimeMap = collectDeadReferences(api);
-      List<Reference> allRefs = new ArrayList<>(liveReferences);
+      Map<String, Instant> droppedReferenceTimeMap = collectDeadReferences(api);
+      // As this list of references is passed from Spark driver to executor,
+      // using available Immutables JSON serialization instead of adding java serialization to the
+      // classes.
+      List<String> allRefs =
+          liveReferences.stream().map(GCUtil::serializeReference).collect(Collectors.toList());
       if (droppedReferenceTimeMap.size() > 0) {
         allRefs.addAll(droppedReferenceTimeMap.keySet());
       }
@@ -134,8 +138,8 @@ public class GCImpl {
     return defaultRefMetadata.getNumTotalCommits();
   }
 
-  private static Map<Reference, Instant> collectDeadReferences(NessieApiV1 api) {
-    Map<Reference, Instant> droppedReferenceTimeMap = new HashMap<>();
+  private static Map<String, Instant> collectDeadReferences(NessieApiV1 api) {
+    Map<String, Instant> droppedReferenceTimeMap = new HashMap<>();
     Stream<RefLogResponse.RefLogResponseEntry> reflogStream;
     try {
       reflogStream =
@@ -168,12 +172,13 @@ public class GCImpl {
           switch (entry.getRefType()) {
             case RefLogResponse.RefLogResponseEntry.BRANCH:
               droppedReferenceTimeMap.put(
-                  Branch.of(entry.getRefName(), hash),
+                  GCUtil.serializeReference(Branch.of(entry.getRefName(), hash)),
                   getInstantFromMicros(entry.getOperationTime()));
               break;
             case RefLogResponse.RefLogResponseEntry.TAG:
               droppedReferenceTimeMap.put(
-                  Tag.of(entry.getRefName(), hash), getInstantFromMicros(entry.getOperationTime()));
+                  GCUtil.serializeReference(Tag.of(entry.getRefName(), hash)),
+                  getInstantFromMicros(entry.getOperationTime()));
               break;
             default:
               throw new RuntimeException(
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
@@ -15,6 +15,8 @@
  */
 package org.projectnessie.gc.base;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Map;
 import java.util.Spliterator;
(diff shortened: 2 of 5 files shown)


gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -263,6 +266,24 @@ private Instant getCutoffTimeForRef(
         ? gcParams.getDefaultCutOffTimestamp()
         : gcParams
             .getCutOffTimestampPerRef()
-            .getOrDefault(reference.getName(), gcParams.getDefaultCutOffTimestamp());
+            .getOrDefault(reference, gcParams.getDefaultCutOffTimestamp());
+  }
+
+  private static Reference deSerializeReference(String reference) {
```
nastra, line 272: deSerializeReference -> deserializeReference

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -263,6 +266,24 @@ private Instant getCutoffTimeForRef(
...
+    ObjectMapper objectMapper = new ObjectMapper();
+    try {
+      return objectMapper.readValue(reference, Reference.class);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static String serializeContent(Content content) {
+    ObjectMapper objectMapper = new ObjectMapper();
```
nastra, line 282: `objectMapper` should rather be kept in a static final. Also I think it would make sense to have this functionality in a utility class or something so that I can be reused (rather than copied)

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -62,13 +64,13 @@ public IdentifyContentsPerExecutor(GCParams gcParams) {
             bloomFilterSize);
   }
 
-  protected Function<Reference, IdentifiedResult> computeExpiredContentsFunc(
+  protected Function<String, IdentifiedResult> computeExpiredContentsFunc(
```
nastra, line 67: shouldn't the type still stay `Reference` rather than using the serialized JSON string?

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifiedResult.java
```diff
@@ -34,10 +33,10 @@
   // ContentValues per reference per content id.
   private final Map<String, Map<String, ContentValues>> contentValues = new ConcurrentHashMap<>();
 
-  public void addContent(String refName, Content content) {
+  public void addContent(String refName, String contentId, String content) {
```
nastra, line 36: shouldn't the type still stay `Content` rather than using the serialized JSON string?

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
```diff
@@ -189,4 +200,12 @@ private static Instant getInstantFromMicros(Long microsSinceEpoch) {
         TimeUnit.MICROSECONDS.toNanos(
             Math.floorMod(microsSinceEpoch, TimeUnit.SECONDS.toMicros(1))));
   }
+
+  private static String getSerializedReference(ObjectMapper objectMapper, Reference reference) {
```
nastra, line 204: this functionality should be defined exactly once and reused across classes/tests

nastra (changes requested):

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifiedResult.java
ajantha-bhat: I don't want to deserialize it here as these output will be written to spark rows (#3421) and it needs a serialized string. 
I will add a comment about it.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
ajantha-bhat: This Lambda is passed from driver to executor, hence spark needs JAVA serialization if we keep `Reference`.
As described in the PR description, wanted to avoid java serialization.

Once received in executor, deserializing it in the method that this Lamdba calls.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
ajantha-bhat: Moved to `GCUtil`

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
ajantha-bhat: Moved to `GCUtil`

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
ajantha-bhat: done.

ajantha-bhat:
@nastra , @snazy : Can you please take a look at this? once it is merged, I need to rebase #3421 and work on top of this.

snazy (changes requested):
This PR makes `Content` serializable, but after #3421 there's no need at all to serialize `Content` and `IdentifiedContents`, which makes me think that this change shoud go in after #3421, not before.

ajantha-bhat:
> This PR makes Content serializable, but after https://github.com/projectnessie/nessie/pull/3421 there's no need at all to serialize Content and IdentifiedContents, which makes me think that this change shoud go in after https://github.com/projectnessie/nessie/pull/3421, not before.

while working on #3421, I faced serialization problem with dataframe (for Reference). Hence, I need this change to make #3421 work.

I still think `Content` needs to be Json serilaized to store them in spark row. Later in #3421 we can see if `IdentifiedContents` is required or not. 

So, I feel this PR can go now. If you wish to wait till #3421 to finish, I will just cherry-pick this change and use it in that PR to solve that problem.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
```diff
@@ -26,12 +28,52 @@
...
 import org.projectnessie.client.http.HttpClientBuilder;
+import org.projectnessie.model.Content;
 import org.projectnessie.model.LogResponse;
+import org.projectnessie.model.Reference;
 
 public final class GCUtil {
 
   private GCUtil() {}
 
+  static final ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
ajantha-bhat: done

snazy:
I'd really prefer to limit this change to `Reference` instead of touching the same code multiple times.

rymurr:
> I still think Content needs to be Json serilaized to store them in spark row. Later in https://github.com/projectnessie/nessie/pull/3421 we can see if IdentifiedContents is required or not.

I don't think we should be json serializing for Spark rows. (for contents or reference) I think we should either have specialized serializers in the gc module and register them with spark or we should implement Kryo Serializable container objects (see old gc module for an example).

I do think we should simply drop the Serializable from model objects asap though.

ajantha-bhat:
@snazy , @rymurr : Thanks. I will explore about how to show `Content` in spark Row (by seeing older PR or reading more on the suggestion given)

So, `Content` still needs to serialized until I do that change whole storage logic.
As  @snazy suggested I will change this PR scope only for `Reference`

ajantha-bhat:
Changed and pushed the PR for only to be `Reference` specific.

nastra (approved):

snazy (approved):
Approved with a mandatory follow-up to change serialization from JSON to Kryo or a custom serializer for `Reference`.

PR #3569, opened 2022-03-09
Respect max-requests in batch-deletes for DynamoDB

snazy:


versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -288,33 +288,47 @@ protected void doCleanUpCommitCas(
...
+                  .deleteRequest(
+                      b ->
+                          b.key(
+                              Collections.singletonMap(
+                                  KEY_NAME,
+                                  AttributeValue.builder().s(keyPrefix + hash.asString()).build())))
+                  .build());
+      requests++;
+
+      if (requests == DYNAMO_BATCH_WRITE_MAX_REQUESTS) {
```
tomekl007, line 317: can we add a test for it?

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -288,33 +288,47 @@ protected void doCleanUpCommitCas(
...
+      }
+    }
+
+    // close() is actually a flush, implementing AutoCloseable for easier use of BatchDelete using
+    // try-with-resources.
+    @Override
+    public void close() {
+      if (requests > 0) {
+        client.client.batchWriteItem(b -> b.requestItems(requestItems));
+        requestItems.clear();
```
tomekl007, line 328: what happens to `requestItems` that are still in the buffer but not written?

snazy: It's an `AutoCloseable`, right?

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
snazy: will do

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
tomekl007: Yes, but it seems that we are ignoring additional records (`> DYNAMO_BATCH_WRITE_MAX_REQUESTS`).
Or is the `batchWriteItem` doing the write of the remaining items? If yes, we will still write `> DYNAMO_BATCH_WRITE_MAX_REQUESTS` records. Is it something that we expect?

snazy: `batchWriteItem` is the Dynamo-API call

tomekl007 (approved):

PR #3571, opened 2022-03-09
Remove duplicate, identical tests

snazy:


clients/client/src/test/java/org/projectnessie/client/TestNessieHttpClient.java
```diff
@@ -45,36 +41,6 @@ static void setupTracer() {
     JaegerTestTracer.register();
```
nastra, line 41: I think 
```
@BeforeAll
  static void setupTracer() {
    JaegerTestTracer.register();
  }
```
can be removed now right?

clients/client/src/test/java/org/projectnessie/client/http/TestHttpClientBuilder.java
```diff
@@ -145,49 +141,6 @@ void testAuthBasic(Function<HttpClientBuilder, HttpClientBuilder> config) throws
                     UTF_8));
   }
 
-  @Test
-  void testTracing() throws Exception {
```
nastra, line 149: where are we testing this now, given that we're removing it here?

snazy: In the class you mentioned in the above comment

clients/client/src/test/java/org/projectnessie/client/TestNessieHttpClient.java
snazy: Ehm - no - `testTracing` is in this class

nastra: then I guess we can remove ^ from `TestHttpClientBuilder` :)

clients/client/src/test/java/org/projectnessie/client/http/TestHttpClientBuilder.java
nastra: yeah I got confused by what is where, so def. makes sense to clean this up here

PR #3572, opened 2022-03-09
Support Namespaces

### LINKED ISSUE #3254: Support explicit Namespaces
Currently a Namespace in Nessie is implicit via the name prefix of a `ContentKey`. However, by making `Namespace` explicit, we can better match how existing databases/systems represent Namespaces.
In the mid-term we could also think about adding properties to Namespaces, such as security privileges.

nastra:
fixes #3254

This introduces a new `NamespaceApi` that allows creating/retrieving/deleting a single Namespace and also retrieving all Namespaces with a given prefix.

A Namespace is now a subtype of `Content` to keep changes to a minimum at the storage level.

Given that users currently don't explicitly create a Namespace when adding a new table 'a.b.c.table', the `NamespaceApi` is designed in a way where it first needs to evaluate:
* whether an explicitly created Namespace exists: An explicitly created Namespace is created via `NamespaceApi#createNamespace(NamespaceParams)`
* whether an implicitly created Namespace exists: An implicitly created namespace generally occurs when adding a table 'a.b.c.table' where the 'a.b.c' part represents the namespace without

This feature also adds 3 new Exceptions: `NessieNamespaceAlreadyExistsException` / `NessieNamespaceNotEmptyException` / `NessieNamespaceNotFoundException`

Note that this currently does not support Namespace properties, as those will be handled separately. Also authorization on Namespaces will be handled separately.


model/src/main/java/org/projectnessie/api/params/NamespaceParams.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class NamespaceParams {
+
+  @Parameter(
+      description = "name of ref to fetch",
+      examples = {@ExampleObject(ref = "ref")})
+  @PathParam("ref")
+  @NotNull
+  @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
+  private String refName;
```
snazy, line 37: Requires a `hashOnRef` as well - think: `DETATCHED`

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImplWithAuthorization.java
```diff
@@ -0,0 +1,68 @@
...
+import org.projectnessie.model.Content;
+import org.projectnessie.model.Content.Type;
+import org.projectnessie.model.GetNamespacesResponse;
+import org.projectnessie.model.Namespace;
+import org.projectnessie.services.authz.Authorizer;
+import org.projectnessie.services.config.ServerConfig;
+import org.projectnessie.versioned.VersionStore;
+
+/** Does authorization checks (if enabled) on the {@link NamespaceApiImpl}. */
+public class NamespaceApiImplWithAuthorization extends NamespaceApiImpl {
```
snazy, line 34: Isn't this class missing its actual implementation?

model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -33,24 +36,29 @@
 @Value.Immutable
 @JsonSerialize(as = ImmutableNamespace.class)
 @JsonDeserialize(as = ImmutableNamespace.class)
-public abstract class Namespace {
+@JsonTypeName("NAMESPACE")
+public abstract class Namespace extends Content {
```
snazy, line 40: I think it's wrong to have the namespace elements repeated in the `Namespace` `Content` object. The name of the `Namespace` is already determined by the `ContentKey`, so it's duplicated.
Maybe just add a new `NamespaceContent` object that extends `Content` and is currenty empty?

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+      BranchName branch = branchFromRefName(params);
+      if (getExplicitlyCreatedNamespace(params, branch).isPresent()) {
+        throw namespaceAlreadyExistsException(params);
+      }
+      if (getImplicitlyCreatedNamespace(params, branch).isPresent()) {
+        throw namespaceAlreadyExistsException(params);
+      }
+
+      Namespace namespace = params.getNamespace();
+      Put put = Put.of(ContentKey.of(namespace.getElements()), namespace);
```
snazy, line 75: This line makes it obvious, that the FQN of the namespace is encoded in a `ContentKey` but also duplicated in the value.
This can end in a situation where namespace `a.b.c` is committed as content-key `d.e.f`

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+  public NamespaceApiImpl(
+      ServerConfig config,
+      VersionStore<Content, CommitMeta, Content.Type> store,
+      Authorizer authorizer,
+      Principal principal) {
+    super(config, store, authorizer, principal);
+  }
+
+  @Override
+  public Namespace createNamespace(NamespaceParams params)
```
snazy, line 63: The implementation doesn't guarantee atomicity.
The checks at the top do not run within the atomic commit operation.

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+
+      return namespace;
+    } catch (ReferenceNotFoundException | ReferenceConflictException e) {
+      throw new NessieReferenceNotFoundException(e.getMessage(), e);
+    }
+  }
+
+  @Override
+  public void deleteNamespace(NamespaceParams params)
+      throws NessieReferenceNotFoundException, NessieNamespaceNotEmptyException,
```
snazy, line 86: Similar here - "not empty check" is running outside the atomic commit operation.

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+   * created namespace generally occurs when adding a table 'a.b.c.table' where the 'a.b.c' part
+   * represents the namespace.
+   *
+   * @param params The params to use for fetching the namespaces
+   * @param branch The ref
+   * @return A list of implicit namespaces that might contain duplicate entries.
+   */
+  private List<Namespace> getImplicitlyCreatedNamespaces(NamespaceParams params, BranchName branch)
+      throws ReferenceNotFoundException {
+    try (Stream<WithType<Key, Type>> stream = getImplicitNamespacesStream(params, branch)) {
```
snazy, line 202: Checks for explicit and implicit namespaces should be unified, because the stream returning the implicit ones includes the explicitly created one.

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+    return BranchName.of(
+        Optional.ofNullable(params.getRefName()).orElse(getConfig().getDefaultBranch()));
+  }
+
+  private NessieReferenceNotFoundException refNotFoundException(ReferenceNotFoundException e) {
+    return new NessieReferenceNotFoundException(e.getMessage(), e);
+  }
+
+  private void commit(BranchName branch, String commitMsg, Operation<Content> contentOperation)
+      throws ReferenceNotFoundException, ReferenceConflictException {
```
snazy, line 271: What prevents a user from messing with namespaces via TreeApi?

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,279 @@
...
+    getStore()
+        .commit(
```
snazy, line 273: I'll post a PR to add checks to ensure atomicity shortly.

compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
```diff
@@ -86,6 +86,9 @@ protected Application configure() {
             withClass(
                 "org.projectnessie.services.rest.ConstraintViolationExceptionMapper",
                 c -> config.register(c, 10));
+            withClass(
```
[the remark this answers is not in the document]
nastra: we do but I forgot to add it

model/src/main/java/org/projectnessie/model/Namespace.java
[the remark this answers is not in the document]
nastra: The only place where we exposed `Namespace` was https://github.com/projectnessie/nessie/blob/f9a71ce42abedaad707a6ee405e7801d3f74f26e/model/src/main/java/org/projectnessie/model/ContentKey.java#L56-L59, but we never actually serialized it anywhere, therefore we're not breaking anything (luckily)

model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -33,24 +36,29 @@
...
+  @Override
+  public Type getType() {
+    return Type.NAMESPACE;
+  }
 
-  @JsonValue
   @NotNull
   public abstract String name();
 
-  @Value.Redacted
```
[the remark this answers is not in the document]
nastra: removed it by accident

model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -55,7 +54,6 @@
    *     from {@link ContentKey#getElements()}.
    */
   @JsonIgnore
-  @Value.Redacted
```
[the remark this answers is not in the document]
nastra: accidental removal

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImplWithAuthorization.java
nastra: authorization will be handled in a separate PR

model/src/main/java/org/projectnessie/api/params/NamespaceParams.java
nastra: you're right, added

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
nastra: currently nothing

model/src/main/java/org/projectnessie/model/Namespace.java
nastra: While I agree that the namespace is already determined in `ContentKey`, I unfortunately don't see any better way atm as we need/want to expose the `Namespace` instance in the `NamespaceApi`. 
In terms of the relationship between `ContentKey` + `Namespace` I see it this way:

Creating an actual `Namespace` `a.b.c` via the `NamespaceApi` will do a `Put.of(ContentKey.of(namespace.getElements()), namespace)`, so the full `ContentKey` name will be the same as the actual `Namespace`, namely `a.b.c`. `ContentKey` then also has a `ContentKey.getNamespace()` method, which then will return `a.b.` (omitting the last element), which makes sense, since the `Namespace` of `a.b.c` is `a.b`.

clients/client/src/main/java/org/projectnessie/client/api/OnNamespaceBuilder.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.client.api;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import org.projectnessie.model.Namespace;
+import org.projectnessie.model.Validation;
+
+public interface OnNamespaceBuilder<R extends OnNamespaceBuilder<R>> {
```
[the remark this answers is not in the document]
snazy: ?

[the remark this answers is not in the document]
snazy: oh, not yet updated on your branch

snazy:
I wonder whether we should actually let `Namespace` extend `Content` in the public API.
`Namespace` shall, as it seems from this PR, exclusively be managed using the new REST API (the validations performed in the new endpoints), but users should actually not use `Namespace` via `TreeApi.commt()` (could put a `Namespace` using a different `ContentKey`, unconditionally delete namespaces, maybe more).

A merge operation, where namespace `a.b` was created for a table `a.b.C` on branch `foo` and namespace `a.b` was also created on the `main` branch for a table `a.b.Foo`, will fail with this change. Such a merge of `a.b.C` into `main` is currently perfectly valid. But with the new `Namespace` objects, the merge would fail, because `a.b` was modified on both the source and target branch since the common ancestor. It seems that there's a bunch more logic needed?

model/src/main/java/org/projectnessie/error/ErrorCode.java
```diff
@@ -34,7 +34,9 @@
   BAD_REQUEST(400, NessieBadRequestException::new),
   FORBIDDEN(403, NessieForbiddenException::new),
   TOO_MANY_REQUESTS(429, NessieBackendThrottledException::new),
-  ;
```
dimas-b, line 37: The idea of having the `;` on a new line was to avoid changing old lines when new enum values are added :)

model/src/main/java/org/projectnessie/api/NamespaceApi.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Creates a new namespace.
+   *
+   * @param params The {@link NamespaceParams} that includes the parameters for the API call.
+   * @return A {@link Namespace} instance if creating the namespace succeeded.
+   * @throws NessieReferenceNotFoundException If the reference could not be found.
+   * @throws NessieNamespaceAlreadyExistsException If the namespace already exists.
+   */
+  Namespace createNamespace(@NotNull NamespaceParams params)
+      throws NessieNamespaceAlreadyExistsException, NessieReferenceNotFoundException;
```
dimas-b, line 39: Since this operation makes a commit, I though we ought to return a `Branch` with the up-to-date reference info... same as what `TreeApi.commitMultipleOperations` returns.

Does returning `Namespace` to the caller actually communicates new information that the caller does not have already?

Also, since a commit is made I think we have to declare `NessieConflictException`. Even though it is not technically possible now, it may be possible later when namespace have properties.

model/src/main/java/org/projectnessie/api/params/NamespacesParams.java
```diff
@@ -0,0 +1,113 @@
...
+import javax.validation.constraints.Pattern;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
+import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.immutables.builder.Builder;
+import org.projectnessie.model.Namespace;
+import org.projectnessie.model.Validation;
+
+public class NamespacesParams {
```
dimas-b, line 31: Suggestion: rename to avoid confusion with `NamespaceParams`... it is very hard to tell them apart :) ... perhaps use `MultipleNamespacesParams` here?

model/src/main/java/org/projectnessie/api/http/HttpNamespaceApi.java
```diff
@@ -0,0 +1,131 @@
...
+    @APIResponse(responseCode = "401", description = "Invalid credentials provided"),
+    @APIResponse(responseCode = "403", description = "Not allowed to retrieve namespace"),
+    @APIResponse(responseCode = "404", description = "Reference or Namespace not found"),
+  })
+  Namespace getNamespace(@BeanParam @NotNull NamespaceParams params)
+      throws NessieNamespaceNotFoundException, NessieReferenceNotFoundException;
+
+  @Override
+  @GET
+  @Path("/{ref}")
```
dimas-b, line 112: What if we have a ref named `namespace`... will the path resolution work with respect to `getNamespace`?

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
```diff
@@ -123,4 +123,16 @@
    * <p>See {@code org.projectnessie.client.StreamingUtil} in {@code nessie-client}.
    */
   GetRefLogBuilder getRefLog();
+
+  /** Fetch a single {@link org.projectnessie.model.Namespace}. */
+  GetNamespaceBuilder getNamespace();
+
+  /** Fetch one or more {@link org.projectnessie.model.Namespace}s based on a given prefix. */
+  GetNamespacesBuilder getNamespaces();
```
dimas-b, line 131: Same naming concern: `GetNamespaceBuilder` is _very_ similar to `GetNamespacesBuilder`... could the latter be renamed?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/GenericContentVariantSupplier.java
```diff
@@ -37,4 +38,13 @@ public ContentVariant getContentVariant(byte type) {
     }
     return ContentVariant.ON_REF;
   }
+
+  @Override
+  public boolean isNamespace(ByteString type) {
+    try {
+      return storeWorker.getType(type).name().equals("NAMESPACE");
```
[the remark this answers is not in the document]
nastra: makes sense, done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1456,16 +1462,56 @@ protected void checkForKeyCollisions(
...
+    Predicate<Entry<Key, ContentAndState<ByteString>>> isNamespace =
+        e -> contentVariantSupplier.isNamespace(e.getValue().getRefState());
+    Set<Key> namespacesOnTarget =
+        fetchValues(ctx, hashFromTarget, keyCollisions, ALLOW_ALL).entrySet().stream()
+            .filter(isNamespace)
+            .map(Entry::getKey)
+            .collect(Collectors.toSet());
+
+    Set<Key> namespacesOnSource =
+        fetchValues(ctx, hashFromSource, keyCollisions, ALLOW_ALL).entrySet().stream()
```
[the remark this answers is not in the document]
nastra: makes sense, fixed

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1456,16 +1462,56 @@ protected void checkForKeyCollisions(
...
+      throws ReferenceNotFoundException {
+    Predicate<Entry<Key, ContentAndState<ByteString>>> isNamespace =
+        e -> contentVariantSupplier.isNamespace(e.getValue().getRefState());
+    Set<Key> namespacesOnTarget =
+        fetchValues(ctx, hashFromTarget, keyCollisions, ALLOW_ALL).entrySet().stream()
+            .filter(isNamespace)
+            .map(Entry::getKey)
+            .collect(Collectors.toSet());
+
+    Set<Key> namespacesOnSource =
```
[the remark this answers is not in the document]
nastra: let's leave this probably for a follow-up optimization

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -537,10 +540,9 @@ public Branch commitMultipleOperations(String branch, String hash, Operations op
     }
   }
 
-  private Function<CommitMeta, CommitMeta> commitMetaUpdate() {
+  static Function<CommitMeta, CommitMeta> commitMetaUpdate(Principal principal) {
```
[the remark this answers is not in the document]
nastra: done

model/src/main/java/org/projectnessie/api/NamespaceApi.java
nastra: returning a `Branch` seems a bit non-expected from a user's perspective. Returning a `Namespace` does not carry any new information currently, but it will potentially in the future

model/src/main/java/org/projectnessie/api/http/HttpNamespaceApi.java
nastra: yes this works. I updated one of the tests to use a Branch called `namespace`

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
nastra: renamed to `getMultipleNamespaces`

model/src/main/java/org/projectnessie/api/params/NamespacesParams.java
nastra: done

model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -68,13 +78,22 @@ public static Namespace of(String... elements) {
...
+        throw new IllegalArgumentException("A namespace must not contain a zero byte.");
+      }
+    }
+
     if (DOT.equals(elements[elements.length - 1])) {
       throw new IllegalArgumentException(
           String.format(ERROR_MSG_TEMPLATE, Arrays.toString(elements)));
     }
 
     String name = String.join(DOT, Arrays.asList(elements));
```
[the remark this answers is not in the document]
nastra: I believe we wanted to model the `Namespace` class back then as close as possible to the one in Iceberg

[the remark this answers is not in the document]
nastra: I'll follow up with a fix for this right after this PR as I don't want this PR to get bigger than it already is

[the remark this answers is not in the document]
nastra: fixed in https://github.com/projectnessie/nessie/pull/3632/commits/8cc3c6520018a79e1b26dee621ff6880bd4fa824

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: sgtm

model/src/main/java/org/projectnessie/model/GetNamespacesResponse.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
+import javax.validation.constraints.NotNull;
+import org.immutables.value.Value.Immutable;
+
+@Immutable
+@JsonSerialize(as = ImmutableGetNamespacesResponse.class)
+@JsonDeserialize(as = ImmutableGetNamespacesResponse.class)
+public interface GetNamespacesResponse {
```
dimas-b, line 27: `EntriesResponse` is paginated... Should we extend `PaginatedResponse` too, even if we do not support pagination ATM?

model/src/test/java/org/projectnessie/model/TestNamespace.java
```diff
@@ -98,6 +98,22 @@ void testInvalidParsing(String identifier) {
         .hasMessage(String.format(Namespace.ERROR_MSG_TEMPLATE, identifier));
   }
 
+  @ParameterizedTest
+  @ValueSource(strings = {"\u0000", "a.\u0000", "a.b.c.\u0000"})
+  void testZeroByteUsage() {
```
dimas-b, line 103: Parameters aren't used?

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -0,0 +1,299 @@
...
+  @Override
+  public Namespace createNamespace(NamespaceParams params)
+      throws NessieNamespaceAlreadyExistsException, NessieReferenceNotFoundException {
+    try {
+      BranchName branch = branchFromRefName(params.getRefName());
+
+      Callable<Void> validator =
+          () -> {
+            if (getExplicitlyCreatedNamespace(params, branch).isPresent()) {
+              throw namespaceAlreadyExistsException(params);
```
dimas-b, line 74: Technically, it could be a table with the same key, right? Should we adjust the message to say "Another content object already exists"... or something like that?

model/src/test/java/org/projectnessie/model/TestNamespace.java
nastra: fixed

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
dimas-b: ok, let's address in a follow-up PR

dimas-b (approved):

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
nastra: done in https://github.com/projectnessie/nessie/pull/3636

PR #3573, opened 2022-03-09
Add unique contentId & pass expectedContent when committing

tomekl007:
My current findings:
I've tried to split the state per `user_id` by making the `val contentId = tableName + "_" + userId.toString` containing the `userId`. It reduced number of errors for the `CommitToBranchSimulation` simulation:
Screenshot 2022-03-09 at 12 43 30
but there are still situations when the global state is not found/ or conflicts. Is this all we need to do to use the global state, or Am I missing something?

Regarding this comment in the codebase:
```         
// TODO the expectedContent is wrong!! commitNum is for the session, but we need the actual global state!!
```
However, Gatling does not give the possibility to share a state between runs for different sessions, and I think there is a good reason for that. If the `commitNumber` is global, but we have N threads, each inserting data for its own `user_id`, there will be race conditions and conflicts - for that scenario, there is no way that the expected state would match the inserted state.

Also, this test uses mode `BRANCH_PER_USER_SINGLE_TABLE`, so the same table is shared between all threads/users - is it expected? Because of that, the same table name is used for all sessions, and this is yet another place where conflicts occur. I've tried to fix it via making the table more granular: `path_on_disk_${tableName}_${userId}_${commitNum - 1}` but I am not sure if this is what we expect from this benchmark. Without those changes, the conflicts in the global state are also happening:
Screenshot 2022-03-09 at 13 31 53

Also, one general question to our Gatling tests:
Are they supposed to be run in isolation, or can they be run on the same instance of Nessie? I've noticed that when changing something in test A, test B may start having some errors. Is every test running in the fresh nessie instance, or are they reusing the same Nessie?
(@dimas-b @snazy thoughts?)

**Improvement**
With https://github.com/projectnessie/nessie/pull/3573/commits/433d5eb86598a09ce8bc985785bf852512e08102
I've added usage of a global state that fixes the conflicts. The run results look like this now:
Screenshot 2022-03-16 at 10 34 50
However, the change with the global state only does not solve this problem. We need global state + content-id per user_id that is implemented in this PR

### DIFF
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
@@ -64,15 +64,15 @@ class CommitToBranchSimulation extends Simulation {
               contentId
             )
 
-          val globalStates = client.getContent.reference(branch).key(key).get()
-          val globalIcebergTable = globalStates.get(key)
+          val existingTable =
+            client.getContent.reference(branch).key(key).get().get(key)
 
           val op =
-            if (commitNum > 0 && globalIcebergTable != null)
+            if (commitNum > 0 && existingTable != null)
               Put.of(
                 key,
                 tableMeta,
-                globalIcebergTable
+                existingTable
               )
             else Put.of(key, tableMeta);
 
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
@@ -64,20 +64,15 @@ class CommitToBranchSimulation extends Simulation {
               contentId
             )
 
-          // TODO the expectedContent is wrong!! commitNum is for the session, but we need the actual global state!!
+          val globalStates = client.getContent.reference(branch).key(key).get()
+          val globalIcebergTable = globalStates.get(key)
+
           val op =
-            if (commitNum > 0)
+            if (commitNum > 0 && globalIcebergTable != null)
               Put.of(
                 key,
                 tableMeta,
-                IcebergTable.of(
-                  s"path_on_disk_${tableName}_${userId}_${commitNum - 1}",
-                  42,
-                  43,
-                  44,
-                  45,
-                  contentId
-                )
+                globalIcebergTable
               )
             else Put.of(key, tableMeta);
 
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
@@ -51,12 +51,12 @@ class CommitToBranchSimulation extends Simulation {
           val tableName = params.makeTableName(session)
 
           // Call the Nessie client operation to perform a commit
-          val key = ContentKey.of("name", "space", tableName, userId.toString)
+          val key = ContentKey.of("name", "space", tableName)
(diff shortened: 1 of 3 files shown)


snazy:
Hm - sad that Gatling has no global state, but that makes sense, because you can run multiple Gatling load drivers.

The necessary change here should then be to read the current `IcebergMetadata` from Nessie. It it doesn't yet exist, it's a new table (and no expected state) - if it does exist, then use that as the expected state.

dimas-b (commented):
I'd say we should assume each simulation runs in isolation. Otherwise, the about of permutations we have to worry about is going to be unbearable.

I guess we could also run some single-threaded task during the "setup" phase if we need to pre-create some objects in Nessie.

tomekl007:
> Hm - sad that Gatling has no global state, but that makes sense, because you can run multiple Gatling load drivers.
> 
> The necessary change here should then be to read the current `IcebergMetadata` from Nessie. It it doesn't yet exist, it's a new table (and no expected state) - if it does exist, then use that as the expected state.

Done. However, we need global state + content-id per user_id.

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
```diff
@@ -51,33 +51,28 @@ class CommitToBranchSimulation extends Simulation {
...
+              s"path_on_disk_${tableName}_${userId}_$commitNum",
               42,
               43,
               44,
               45,
               contentId
             )
 
-          // TODO the expectedContent is wrong!! commitNum is for the session, but we need the actual global state!!
+          val globalStates = client.getContent.reference(branch).key(key).get()
```
nastra, line 67: it's not "global", it's rather just existing content

nastra (changes requested):

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
```diff
@@ -51,33 +51,28 @@ class CommitToBranchSimulation extends Simulation {
           val tableName = params.makeTableName(session)
 
           // Call the Nessie client operation to perform a commit
-          val key = ContentKey.of("name", "space", tableName)
-          val contentId = tableName
+          val key = ContentKey.of("name", "space", tableName, userId.toString)
+          val contentId = tableName + "_" + userId.toString
```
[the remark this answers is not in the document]
tomekl007: Without those changes, the same problems as before this PR occur:
```
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                            99 ( 10%)
> 800 ms < t < 1200 ms                                   1 (  0%)
> t > 1200 ms                                            0 (  0%)
> failed                                               900 ( 90%)
---- Errors --------------------------------------------------------------------
> org.projectnessie.error.NessieReferenceConflictException: Glob    900 (100.0%)
al-state for content-id 'someTable1647427912836' already exist...
```

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulation.scala
tomekl007: the existing `TODO` states that we should use the `global state`, that is why I named it this way. Isn't it global?

nastra: let's just change this to `val existingTable = client.getContent.reference(branch).key(key).get().get(key)`

nastra (commented):
let's update PR title to something like `Add unique contentId & pass expectedContent when committing` or something along those lines, as `try to fix...` is a bit misleading

nastra (approved):

nastra (approved):

PR #3582, opened 2022-03-10
Split nessie-quarkus into commons, tests, server

snazy:
We need a way to trigger global-log compaction (#3559). There were two options to implement it:
1. Add another REST API endpoint, which requires additional permission checks.
2. Introduce a "low level maintenance tool", aka a command-line tool that works against an instance of `DatabaseAdapter`

For option 2, the easiest way is to reuse the existing infrastructure from `nessie-quarkus` in a CLI application based on Quarkus using picocli, which Quarkus natively supports.

This change moves the "common" production-code parts out of `nessie-quarkus` into a new module `nessie-quarkus-common` and the common test-code parts out of `nessie-quarkus` into a new module `nessie-quarkus-tests`.

There are no functional changes in this PR.


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusNessieAuthenticationConfig.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.projectnessie.server.config;
+package org.projectnessie.quarkus.config;
```
dimas-b, line 16: Shouldn't this stay in `quarkus-server` because it is only used by the quarkus http authenticator?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusNessieAuthorizationConfig.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.projectnessie.server.config;
+package org.projectnessie.quarkus.config;
```
dimas-b, line 16: This config also appears to be used only inside `quarkus-server`... so why move it?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusNessieAuthenticationConfig.java
snazy: Yea - doesn't make sense. Moved back.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -148,6 +148,7 @@ quarkus.index-dependency.protobuf.group-id=com.google.protobuf
 quarkus.index-dependency.protobuf.artifact-id=protobuf-java
 quarkus.native.additional-build-args=\
   --rerun-class-initialization-at-runtime=org.rocksdb.RocksDB,\
+  --rerun-class-initialization-at-runtime=org.apache.http.impl.auth.NTLMEngineImpl,\
```
dimas-b, line 151: Why do we need to add NTLM?

snazy: I tried to investigate why, but couldn't figure out the actual cause.
But that class initialization triggers some `java.util.Random` initialized in a static final field.

dimas-b (approved):

PR #3586, opened 2022-03-10
Validation during atomic commit

snazy:
PR #3572 performs some validations, which need to happen inside the atomic-CAS (or transaction) of a Nessie commit.

This change allows a callback from within the CAS/Op-loop so that the caller can perform atomic validations.

Lots of changes in `TxDatabaseAdapter`, but those are "boring changes". The change to `TxDatabaseAdapter` is to let "nested" operations (aka commit -> validation -> validations performing reads via the version store) use the same JDBC connection/transaction, which required the `ConnectionWrapper` class.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitAttempt.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitAttempt.java
@@ -19,6 +19,7 @@ import com.google.protobuf.ByteString;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import javax.annotation.Nullable;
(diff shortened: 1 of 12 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitAttempt.java
```diff
@@ -68,4 +69,7 @@
 
   /** Serialized commit-metadata. */
   ByteString getCommitMetaSerialized();
+
+  @Nullable
+  Runnable getValidator();
```
dimas-b, line 74: nit: `@Nonnull` with a default no-op impl?

nastra (approved):

PR #3588, opened 2022-03-10
Add tx leak detection

snazy:


versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/ConnectionWrapper.java
```diff
@@ -39,6 +39,10 @@ public static ConnectionWrapper borrow(Supplier<Connection> newConnectionProduce
     return current;
   }
 
+  public static boolean threadHasOpenConnection() {
+    return INHERITED.get() != null;
```
dimas-b, line 43: This PR LGTM, but in general, I wonder if we really need to share connections on the same thread. I'm not sure it will always be correct. IIRC, Postgres does not reusing a connection to execute another statement if the result set of the first statement has not been exhausted / closed. WDYT?

dimas-b (approved):

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/ConnectionWrapper.java
snazy: Yea - it's a bit problematic, but I couldn't find a better way to satisfy the needs of #3586
It's intended to be used in a callback during a commit operation, so there shouldn't be any open result sets.

nastra (approved):

PR #3589, opened 2022-03-10
Remove utility method to create IcebergTable/IcebergView without ID

nastra:


snazy (changes requested):
Not sure whether we should merge this.
The variants without a content-id are there to create _new_ IcebergTable/View contents - the contract is to let Nessie generate the content-ID but not to let the caller define the new content-id.

dimas-b (commented):
Could we forbid different content IDs under the same key reachable from the same commit?

snazy:
> Could we forbid different content IDs under the same key reachable from the same commit?

Nope. It could actually become a boomerang - think that you have a squash-commit with a `put`  + `delete` for a table _and_ a `put` for the same table key. It's still theoretical, but is one thing that #2311 shall specify.

gc/gc-base/src/test/java/org/projectnessie/gc/base/TestContentBloomFilter.java
```diff
@@ -86,10 +87,12 @@ void testDeltaLakeContent() {
     for (int i = index; i < 10 + index; i++) {
       switch (contentType) {
         case ICEBERG_TABLE:
-          contents.add(IcebergTable.of("temp" + i, i, 42, 42, 42));
+          contents.add(IcebergTable.of(UUID.randomUUID().toString(), "temp" + i, i, 42, 42, 42));
```
tomekl007, line 90: I understand that the purpose of this change is to prevent misuse of our API, righ?

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -77,23 +77,12 @@ public Type getType() {
   }
 
   public static IcebergTable of(
```
tomekl007, line 79: Is this a public API that can be used outside of the project? 
If yes, shouldn't we deprecate that instead of removing it, for backward compatibility purpose?

PR #3590, opened 2022-03-10
`AbstractDatabaseAdapter.fetchValues()` looking for non-existing keys

snazy:
A `ContentApi.getContents()` with keys that do not exist leads into
traversing the whole commit log.

The current code in `AbstractDatabaseAdapter.fetchValues()` is optimized to stop
traversing the commit log while looking for content-values when the values for
all requested keys have been found.

In turn this means that the commit-log-traversal will look into _every_ commit
for keys that do not exist.

It's is rather easy to prevent that unnecessary whole-commit-log-traversal for
non-existing keys: Every 20th (by default) commit contains the whole key-lists,
so all visible (think: existing + undeleted) keys. The set of all keys can be
inspected to stop looking for keys that are known to not exist.

While it sounds like a minor optimization, it's in fact a rather big one.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -1116,7 +1116,31 @@ public abstract class AbstractDatabaseAdapter<OP_CONTEXT, CONFIG extends Databas
     Set<ContentId> contentIdsForGlobal = new HashSet<>();
     try (Stream<CommitLogEntry> log =
         takeUntilExcludeLast(readCommitLogStream(ctx, refHead), e -> remainingKeys.isEmpty())) {
-      log.peek(entry -> entry.getDeletes().forEach(remainingKeys::remove))
+      log.peek(
+              entry -> {
+                // remove deleted keys from keys to look for
+                entry.getDeletes().forEach(remainingKeys::remove);
+
+                if (entry.getKeyList() != null) {
+                  // CommitLogEntry has a KeyList.
+                  // All keys in 'remainingKeys', that are _not_ in the KeyList(s), can be removed,
+                  // because at this point we know that these do not exist.
+
+                  Set<Key> remainingInKeyList = new HashSet<>();
+                  try (Stream<KeyList> keyLists =
+                      Stream.concat(
+                          Stream.of(entry.getKeyList()),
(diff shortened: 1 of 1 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1116,7 +1116,31 @@ protected void checkForModifiedKeysBetweenExpectedAndCurrentCommit(
     Set<ContentId> contentIdsForGlobal = new HashSet<>();
     try (Stream<CommitLogEntry> log =
         takeUntilExcludeLast(readCommitLogStream(ctx, refHead), e -> remainingKeys.isEmpty())) {
-      log.peek(entry -> entry.getDeletes().forEach(remainingKeys::remove))
+      log.peek(
+              entry -> {
+                // remove deleted keys from keys to look for
+                entry.getDeletes().forEach(remainingKeys::remove);
+
+                if (entry.getKeyList() != null) {
```
dimas-b, line 1124: Possible extra optimization: I suppose once we process one full key list entry, we do not have to process full key lists again. For example when we are searching for an _exisiting_ key, whose entry is behind _several_ entries with full key lists.

dimas-b (approved):

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: Also, I wonder if it might be valuable to add some unreachable history to our correctness tests... i.e. first run tests with empty history, then run them again with N commits add and removing keys, but leaving no "active" keys in HEAD.... WDYT?

snazy: > Possible extra optimization: I suppose once we process one full key list entry, we do not have to process full key lists again. For example when we are searching for an _exisiting_ key, whose entry is behind _several_ entries with full key lists.

That's tackled by #3592

snazy: > Also, I wonder if it might be valuable to add some unreachable history to our correctness tests... i.e. first run tests with empty history, then run them again with N commits adding and removing keys, but leaving no "active" keys in HEAD.... WDYT? (a separate PR for sure)

Need to add tests for #3592 incl upgrade tests anyway, so I can put that into that PR.

PR #3592, opened 2022-03-10
Record commit-id in key-lists

### LINKED ISSUE #3591: Memoize last commit-ID for keys in `KeyList`
Fetching the values for content-keys currently requires traversing the commit-log up to the commit that last touched the content-key.
For keys that haven't been changed for a long time, this will lead into a lot of unnecessary commit-log fetches.

If the `KeyList` would also maintain the commit-ID for the last change of a key, that would reduce the amount commit-log fetches a lot.

This can be easily implemented as a backwards-compatible change and a change that's safe to use during a rolling upgrade.

/cc @dimas-b

snazy:
Fetching content values can become expensive as fetching the value from a
`Put` opartion may require reading a lot of commit-log-entries, the more
commit-log-entries the older the content.

This change adds the commit-id of the last `Put` operation to the already
existing entries in `KeyList`, which allows `DatabaseAdapter.fetchValues`
to quickly "navigate" to the commits that contain the most recent `Put`
operation that contains the content value for the requested content key.

Time/resource complexity for existing keys is reduced from "number of
commit log entries from HEAD to last Put operation" to at max two
bulk-fetch operations against the commit-log.

Updates & renames the type `KeyWithType` to `KeyListEntry`, adds the
commit-id to `KeyListEntry`.

This is not a breaking change and requires no data migration. Upgrade
tests are included. Although not tested, and therefore not guaranteed,
this change should also be safe to use during a rolling-upgrade.

Fixes #3591

### DIFF
--- a/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
+++ b/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
@@ -18,15 +18,21 @@ package org.projectnessie.tools.compatibility.tests;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.entry;
 import static org.assertj.core.api.Assertions.tuple;
+import static org.assertj.core.api.Assumptions.assumeThat;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.OptionalInt;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import org.assertj.core.api.InstanceOfAssertFactories;
 import org.assertj.core.groups.Tuple;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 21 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/KeyListEntry.java
```diff
@@ -15,19 +15,29 @@
...
 
+import javax.annotation.Nullable;
 import org.immutables.value.Value;
+import org.projectnessie.versioned.Hash;
 import org.projectnessie.versioned.Key;
 
 /** Composite of key, content-id, content-type. */
 @Value.Immutable(lazyhash = true) // this type is used as a map-key in an expensive test
-public interface KeyWithType {
+public interface KeyListEntry {
```
nastra, line 25: is the rename itself really necessary? also javadoc on the class needs to be updated

compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
```diff
@@ -188,18 +198,130 @@ private void expectedRefLogEntry(String op) {
...
+  // //////////////////////////////////////////////////////////////////////////////////////////
+  // Keys + values
+  // //////////////////////////////////////////////////////////////////////////////////////////
+
+  private static Branch keysUpgradeBranch;
+  private static final Map<String, Map<ContentKey, IcebergTable>> keysUpgradeAtHash =
+      new LinkedHashMap<>();
+  private static int keysUpgradeSequence;
+  // exceed both number-of-parents per commit-log-entry and per global-log-entry
+  private static final int keysUpgradeCommitsPerVersion = 65;
```
nastra, line 228: should this rather refer to whatever the default is and add 15 to it? I fear that if the default ever gets bumped, then this won't be testing what's expected anymore

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/KeyListEntry.java
snazy: Yea - make it clear that the type represents an entry of `KeyList` and isn't something "generic".

compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
snazy: done

PR #3602, opened 2022-03-11
Simplify python type hints

XN137:
we achieve the same type checks + hints but with less code.

see also inline comments

### DIFF
--- a/python/pynessie/client/nessie_client.py
+++ b/python/pynessie/client/nessie_client.py
@@ -193,6 +193,7 @@ class NessieClient:
 
         return ContentSchema().load(get_content(self._base_url, self._auth, ref_name, content_key, hash_on_ref, self._ssl_verify))
 
+    # pylint: disable=keyword-arg-before-vararg
     def commit(self, branch: str, old_hash: str, reason: Optional[str] = None, author: Optional[str] = None, *ops: Operation) -> Branch:
         """Modify a set of Nessie tables."""
         meta = CommitMeta(message=reason if reason else "")
@@ -259,6 +260,7 @@ class NessieClient:
         merge_json = MergeSchema().dump(Merge(from_ref, str(from_hash)))
         merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify)
 
+    # pylint: disable=keyword-arg-before-vararg
     def cherry_pick(self, branch: str, from_ref: str, old_hash: Optional[str] = None, *hashes: str) -> None:
         """Cherry pick a list of hashes to a branch."""
         if not old_hash:
--- a/python/pynessie/client/nessie_client.py
+++ b/python/pynessie/client/nessie_client.py
@@ -75,7 +75,7 @@ from ..model import TransplantSchema
 class NessieClient:
     """Base Nessie Client."""
 
-    def __init__(self: "NessieClient", config: confuse.Configuration) -> None:
+    def __init__(self, config: confuse.Configuration) -> None:
         """Create a Nessie Client from known config."""
         self._base_url = config["endpoint"].get()
         self._ssl_verify = config["verify"].get(bool)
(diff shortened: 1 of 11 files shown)


python/tox.ini
```diff
@@ -40,7 +40,8 @@ deps =
     -r{toxinidir}/requirements_lint.txt
 commands =
     # flake8 includes black check due to flake8-black
-    flake8 pynessie tests tools
+    # ANN101+102 should not be enabled by default, see https://github.com/sco1/flake8-annotations/issues/75
+    flake8 --extend-ignore=ANN101,ANN102 pynessie tests tools
```
omarsmak, line 44: Can we just add it to `setup.cfs` https://github.com/projectnessie/nessie/blob/main/python/setup.cfg#L20 ?

XN137: yes we can, i did not realize flake8 was configured here, since was only looking into `pyproject.toml` and `.pylintrc` :sweat_smile:

XN137:
hold off on reviewing this for now...
i must potentially drop the 2nd commit because i've realized they changed the type hints PEP:
image
https://peps.python.org/pep-0484/#union-types

so for some reason its no longer recommended type checker behavior to assume `x : int = None` to be of type `Optional[int]` :thinking:

XN137:
i've removed the 2nd commit for now... even though mypy and pycharm etc still support `x : int = None` as `Optional[int]`, we should no longer rely on it if its no longer recommended in the spec.
so ready for review again.

python/setup.cfg
```diff
@@ -17,14 +17,12 @@ universal = 1
...
 select = ANN,B,B9,B950,BLK,C,D,E,F,I,S,W
-ignore = E203,E510,W503,S101,D412,W503
-extend-ignore = E203
+ignore = ANN101,ANN102,S101,D412,W503
 max-line-length = 140
 max-complexity = 10
 application-import-names = pynessie,tests
 import-order-style = smarkets
 docstring-convention = google
-per-file-ignores = tests/*:S101
```
XN137, line 27: note that `ignore`, `extend-ignore` and `per-file-ignores` all built on top of eachother... so since `S101` and `E203` are in `ignore` already they were not needed in the later stages

also removed the `Exx` ones, since they are no longer present in the codebase.

python/pynessie/client/nessie_client.py
```diff
@@ -193,17 +193,16 @@ def get_content(self: "NessieClient", ref: str, content_key: ContentKey, hash_on
 
         return ContentSchema().load(get_content(self._base_url, self._auth, ref_name, content_key, hash_on_ref, self._ssl_verify))
 
-    def commit(
-        self: "NessieClient", branch: str, old_hash: str, reason: Optional[str] = None, author: Optional[str] = None, *ops: Operation
-    ) -> Branch:
+    # pylint: disable=keyword-arg-before-vararg
```
XN137, line 196: this is a bad pattern but we cant fix it here since this method is part of the public api, so we ignore the rule locally

python/pynessie/client/nessie_client.py
```diff
@@ -263,7 +260,8 @@ def merge(
         merge_json = MergeSchema().dump(Merge(from_ref, str(from_hash)))
         merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify)
 
-    def cherry_pick(self: "NessieClient", branch: str, from_ref: str, old_hash: Optional[str] = None, *hashes: str) -> None:
+    # pylint: disable=keyword-arg-before-vararg
```
omarsmak, line 263: Should we disable this as well from the config file?

XN137: it was disabled before in the config file but this PR removes it there. due to backwards compatibility we have to keep it here but ignore it.

XN137: same as for `commit` method above

omarsmak (approved):

PR #3617, opened 2022-03-14
Add ContentTypeSupplier

nastra:
The idea here is that we need a way to determine the actual content type
for a given `Key` within the DatabaseAdapters.

This is required for #3572 during key collision checks when merging/transplanting namespaces

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ContentType.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ContentType.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.persist.adapter;
+
+/** Determines the type of content that is stored in nessie. */
+public enum ContentType {
+  UNKNOWN,
+  ICEBERG_TABLE,
+  ICEBERG_VIEW,
+  DELTA_LAKE_TABLE
+}
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
@@ -71,7 +71,8 @@ public interface DatabaseAdapterFactory<
       return connector;
     }
 
-    public abstract DatabaseAdapter build(ContentVariantSupplier contentVariantSupplier);
+    public abstract DatabaseAdapter build(
+        ContentVariantSupplier contentVariantSupplier, ContentTypeSupplier contentTypeSupplier);
 
     public Builder<Config, AdjustableConfig, Connector> configure(
         Function<AdjustableConfig, Config> configurator) {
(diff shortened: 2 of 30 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
```diff
@@ -71,7 +71,8 @@ public Connector getConnector() {
       return connector;
     }
 
-    public abstract DatabaseAdapter build(ContentVariantSupplier contentVariantSupplier);
+    public abstract DatabaseAdapter build(
+        ContentVariantSupplier contentVariantSupplier, ContentTypeSupplier contentTypeSupplier);
```
dimas-b, line 75: The number of params here is growing, at the same time we already have a pattern of "with" methods.... Would it make sense to move these suppliers to `withContentVariantSupplier` and `withContentTypeSupplier`?

versioned/persist/bench/src/main/java/org/projectnessie/versioned/persist/benchmarks/CommitBench.java
```diff
@@ -153,7 +154,9 @@ private DatabaseAdapter adapterByName() {
 
       return builder
           .withConnector(providerSource.getConnectionProvider())
-          .build(new GenericContentVariantSupplier<>(SimpleStoreWorker.INSTANCE));
+          .build(
+              new GenericContentVariantSupplier<>(SimpleStoreWorker.INSTANCE),
```
dimas-b, line 158: nit: add `StoreWorker.getContentVariantSupplier()`?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/GenericContentTypeSupplier.java
```diff
@@ -0,0 +1,46 @@
...
+  private final StoreWorker<CONTENT, METADATA, CONTENT_TYPE> storeWorker;
+
+  public GenericContentTypeSupplier(StoreWorker<CONTENT, METADATA, CONTENT_TYPE> storeWorker) {
+    this.storeWorker = storeWorker;
+  }
+
+  @Override
+  public ContentType getContentType(ByteString type) {
+    CONTENT_TYPE typeEnum = storeWorker.getType(type);
+    if (typeEnum.name().equals("ICEBERG_TABLE")) {
```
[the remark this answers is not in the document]
dimas-b: Could we reduce the scope of the new information that we need to expose for namespace support so that the code could be generic across different content types? For example, namespaces are generic across different `StoreWorker` types... Would it make sense to only delineate contents vs. namespaces without exposing specific content types?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterFactory.java
dimas-b: Yeah, this is a minor point. We can refactor later if it becomes a burden.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ContentType.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.persist.adapter;
+
+/** Determines the type of content that is stored in nessie. */
+public enum ContentType {
```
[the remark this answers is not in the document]
nastra: I'm guessing because we don't want to have a dependency to the `model` module within this module itself

nastra:
we had a brainstorming session with @dimas-b and decided for now to go with a workaround that's within the existing `ContentVariantSupplier` (https://github.com/projectnessie/nessie/pull/3572/files#diff-8b7d7a1722e83d14e2a1eaa2b17d333510fae87765980ffc4b53f2c53fc84d14R29)

snazy:
> we had a brainstorming session with @dimas-b and decided for now to go with a workaround that's within the existing ContentVariantSupplier

Can this PR be closed then?

nastra:
yep, can be closed

PR #3645, opened 2022-03-17
Create new Gatling scenario: `CommitToBranchSimulationDifferentTables`

tomekl007:
The https://github.com/projectnessie/nessie/pull/3573 fixed the `CommitToBranchSimulationSameTable` to use a global state and share a table between users - so it validates commits to the same table/branch.
This PR is adding a new scenario `CommitToBranchSimulationDifferentTables` that is creating N tables - each user has its own dedicated table and commits the data. Because the users have dedicates tables, there should be no conflicts between commits.

### DIFF
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
@@ -28,8 +28,9 @@ import org.projectnessie.perftest.gatling.Predef.nessie
 import scala.concurrent.duration.{FiniteDuration, HOURS, NANOSECONDS, SECONDS}
 
 /** Gatling simulation to perform commits against Nessie. It commits the data to
-  * different table - each user has its own dedicated table. Has a bunch of
-  * configurables, see the `val`s defined at the top of this class.
+  * different table - each user has its own dedicated table. The ContentKey is
+  * created per userId. Has a bunch of configurables, see the `val`s defined at
+  * the top of this class.
   */
 class CommitToBranchSimulationDifferentTables extends Simulation {
 
--- a/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
+++ b/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
@@ -27,11 +27,11 @@ import org.projectnessie.perftest.gatling.Predef.nessie
 
 import scala.concurrent.duration.{FiniteDuration, HOURS, NANOSECONDS, SECONDS}
 
-/** Gatling simulation to perform commits against Nessie. Has a bunch of It
-  * commits the data the same table for all users. configurables, see the `val`s
-  * defined at the top of this class.
+/** Gatling simulation to perform commits against Nessie. It commits the data to
+  * different table - each user has its own dedicated table. Has a bunch of
+  * configurables, see the `val`s defined at the top of this class.
   */
-class CommitToBranchSimulationSameTable extends Simulation {
+class CommitToBranchSimulationDifferentTables extends Simulation {
 
   val params: CommitToBranchParams = CommitToBranchParams.fromSystemProperties()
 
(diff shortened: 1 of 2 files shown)


perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
```diff
@@ -0,0 +1,200 @@
...
+            .of(
+              s"path_on_disk_${tableName}_${userId}_$commitNum",
+              42,
+              43,
+              44,
+              45,
+              contentId
+            )
+
+          val existingTable =
```
tomekl007, line 68: @nastra I don't see how if should remove this state at all. Right now, the state is per table and table is per user so keeping this state is not problematic.

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationSameTable.scala
```diff
@@ -27,8 +27,9 @@ import org.projectnessie.perftest.gatling.Predef.nessie
 
 import scala.concurrent.duration.{FiniteDuration, HOURS, NANOSECONDS, SECONDS}
 
-/** Gatling simulation to perform commits against Nessie. Has a bunch of
-  * configurables, see the `val`s defined at the top of this class.
+/** Gatling simulation to perform commits against Nessie. Has a bunch of It
```
nastra, line 30: something is missing here in `Has a bunch of It`

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
```diff
@@ -0,0 +1,200 @@
...
+import org.projectnessie.client.http.HttpClientBuilder
+import org.projectnessie.error.NessieConflictException
+import org.projectnessie.model.Operation.Put
+import org.projectnessie.model._
+import org.projectnessie.perftest.gatling.Predef.nessie
+
+import scala.concurrent.duration.{FiniteDuration, HOURS, NANOSECONDS, SECONDS}
+
+/** Gatling simulation to perform commits against Nessie. It commits the data to
+  * different table - each user has its own dedicated table. Has a bunch of
```
nastra, line 31: I would probably mention that the `ContentKey` includes the userId, so that it's a bit more obvious what the difference between the 2 simulations is

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationSameTable.scala
tomekl007: please see next line

nastra: yes, the entire sentence is `Has a bunch of It commits the data...`, which seems that there's something missing

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
```diff
@@ -0,0 +1,201 @@
...
+  * the top of this class.
+  */
+class CommitToBranchSimulationDifferentTables extends Simulation {
+
+  val params: CommitToBranchParams = CommitToBranchParams.fromSystemProperties()
+
+  /** The actual benchmark code to measure Nessie-commit performance in various
+    * scenarios.
+    */
+  private def commitToBranch: ChainBuilder = {
```
nastra, line 42: given that this is almost a 1:1 copy of the other benchmark, I think it would be good to look into how this could be de-duplicated, but a follow-up is fine

nastra (approved):

perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/CommitToBranchSimulationDifferentTables.scala
tomekl007: yes, good point. I created a follow-up ticket for improving it: https://github.com/projectnessie/nessie/issues/3653

PR #3647, opened 2022-03-17
Use a random port injected to via `System.getProperty("quarkus.http.test-port")` in all gatling simulations

tomekl007:
Depends on:
https://github.com/projectnessie/nessie-apprunner/pull/10/


pom.xml
```diff
@@ -197,7 +197,7 @@
     <maven.version>3.8.5</maven.version>
     <mockito.version>4.4.0</mockito.version>
     <nessie.version>${project.version}</nessie.version>
-    <nessie-apprunner.version>0.21.4</nessie-apprunner.version>
+    <nessie-apprunner.version>0.21.5</nessie-apprunner.version>
```
tomekl007, line 200: it requires a release of the `nessie-apprunner`

dimas-b (dismissed):

perftest/simulations/README.md
```diff
@@ -27,7 +27,7 @@ class SimpleSimulation extends Simulation {
     .client(
       NessieClient
         .builder()
-        .withUri("http://127.0.0.1:19120/api/v1")
+        .withUri(s"http://127.0.0.1:${System.getProperties.getProperty("quarkus.http.test-port")}/api/v1")
```
XN137, line 30: could it make sense to put this into https://github.com/projectnessie/nessie/blob/a66390b1e4f2ba73e21c067b3576ec20dffb22e0/perftest/simulations/src/test/scala/org/projectnessie/perftest/gatling/BaseParams.scala#L57-L60 ?

dimas-b: I'd be fine without refactoring the client creation into a common place in this PR, but if it comes to that, please see my suggestion of `ProtocolParams` in #3199

dimas-b (dismissed):

perftest/simulations/README.md
tomekl007: yeah, I think we can make it common in a follow-up PR

nastra (dismissed):

perftest/simulations/README.md
dimas-b, line 30: Simulations do not have to target Quarkus servers. Since we control the name of the system property, would you mind changing it to something like `nessie.http.port`?

dimas-b (changes requested):

perftest/simulations/pom.xml
```diff
@@ -93,6 +93,7 @@
           <skip>${skipITs}</skip>
           <jvmArgs>
             <jvmArg>-Dsim.users=10</jvmArg>
+            <jvmArg>-Dquarkus.http.test-port=${quarkus.http.test-port}</jvmArg>
```
dimas-b, line 96: Could we simply use `-Dnessie.uri=http://127.0.0.1:${quarkus.http.test-port}` and avoid `.withUri(...)` calls in scala code?

dimas-b (changes requested):

perftest/simulations/README.md
tomekl007: Good point renamed param to `nessie.http.port`

perftest/simulations/pom.xml
tomekl007: I think that the logic for creating the URL should be in the code. Here we want to only inject port. even If I will do `-Dnessie.uri=http://127.0.0.1:${quarkus.http.test-port}`, the scala code needs to concatenate it with `/api/v1` so I would prefer to leave it as is (only changing the port name as in the https://github.com/projectnessie/nessie/pull/3647#discussion_r834565777 comment)

nastra (dismissed):

perftest/simulations/pom.xml
snazy: > I think that the logic for creating the URL should be in the code.

Actually not. `/api/v1` is, although coded in the Quarkus stuff, rather a configuration option. Similar for `http` vs `https` and the host. Not much work to pass in a full URL in case you wanna test a different endpoint using a manual test or in an external perf test environment.

tomekl007: so what's your suggestion?
is it: 
```
    <jvmArg>-Dnessie.uri=http://127.0.0.1:${quarkus.http.test-port}/api/v1</jvmArg>
```
?

dimas-b: I'm for `-Dnessie.uri=...`. Does it work?

tomekl007: done

perftest/simulations/README.md
```diff
@@ -27,7 +27,7 @@ class SimpleSimulation extends Simulation {
     .client(
       NessieClient
         .builder()
-        .withUri("http://127.0.0.1:19120/api/v1")
+        .withUri(s"${System.getProperties.getProperty("nessie.uri")}")
```
snazy, line 30: Not needed (same for the ones below)

dimas-b: Correct - not neeed - the client builder automatically uses the value of the `nessie.uri` system property.

tomekl007: nice! I didn't know that. Changed.

dimas-b (approved):

PR #3652, opened 2022-03-18
Make sure Namespace/ContentKey don't have empty strings in between

nastra:
This also required some changes to `NessieLogStore` because that part of
the code would simply do a `path.split("/")` where the first element
would then end up being an empty string if `path` started with a `/`.

The error messages have also been improved because if the check fails,
then it's difficult to tell exactly why it failed.

### DIFF
--- a/clients/deltalake/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
+++ b/clients/deltalake/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
@@ -34,6 +34,7 @@ import org.projectnessie.model.{
   TableReference
 }
 import org.apache.commons.io.IOUtils
+import org.apache.commons.lang3.StringUtils
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileStatus, Path}
 import org.apache.spark.SparkConf
@@ -364,7 +365,11 @@ class NessieLogStore(sparkConf: SparkConf, hadoopConf: Configuration)
   }
 
   def pathToKey(path: String): ContentKey = {
-    val parts = path.split("/").toList
+    // we can't simply do a path.toString.split("/") as that would result with
+    // "/tmp/a/b/c in a string array where the first element is an empty string,
+    // which isn't supported by ContentKey
+    val p = StringUtils.stripEnd(StringUtils.stripStart(path, "/"), "/")
+    val parts = p.split("/").toList
     ContentKey.of(parts.asJava)
   }
 
--- a/clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
+++ b/clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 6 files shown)


model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -86,16 +86,25 @@ public boolean isEmpty() {
    */
   public static Namespace of(String... elements) {
     Objects.requireNonNull(elements, "elements must be non-null");
-    if (elements.length == 0 || "".equals(elements[0])) {
+    if (elements.length == 0 || (elements.length == 1 && "".equals(elements[0]))) {
```
XN137, line 89: when does this usually get called with an array of length 1 containing just the empty string? it looks odd

clients/deltalake/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
```diff
@@ -364,7 +365,11 @@ class NessieLogStore(sparkConf: SparkConf, hadoopConf: Configuration)
   }
 
   def pathToKey(path: String): ContentKey = {
-    val parts = path.split("/").toList
+    // we can't simply do a path.toString.split("/") as that would result with
+    // "/tmp/a/b/c in a string array where the first element is an empty string,
+    // which isn't supported by ContentKey
+    val p = StringUtils.stripEnd(StringUtils.stripStart(path, "/"), "/")
+    val parts = p.split("/").toList
```
XN137, line 372: do we maybe have a utils class in deltalake module where we could put `buildContentKeyFromFilePath` or the like.
or make this method reusable from the test?

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
```diff
@@ -98,7 +99,12 @@ void testCheckpoint() throws NessieNotFoundException {
     Assertions.assertEquals(64, expectedSize);
 
     String tableName = tempPath.getAbsolutePath() + "/_delta_log";
-    ContentKey key = ContentKey.of(tableName.split("/"));
+    // we can't simply do a tableName.split("/") as that would result with
+    // "/tmp/a/b/c in a string array where the first element is an empty string,
+    // which isn't supported by ContentKey
+    String[] elements =
+        StringUtils.stripEnd(StringUtils.stripStart(tableName, "/"), "/").split("/");
+    ContentKey key = ContentKey.of(elements);
```
XN137, line 107: nit: its odd to have such a technical detail in the code of a single test... we should at least have a helper method with a proper name?

XN137 (approved):
not sure i understand all implications but code-wise it looks ok with some suggestions

model/src/main/java/org/projectnessie/model/Namespace.java
nastra: this is mainly for the `Namespace.of(""))` case and without this check `Namespace.of("", "something")` will produce `Namespace.EMPTY`

clients/deltalake/src/main/scala/org/projectnessie/deltalake/NessieLogStore.scala
nastra: we don't have one, but I'll add one in a follow-up with some tests

PR #3654, opened 2022-03-18
Extract Hadoop path conversion to ContentKey into util class

nastra:


clients/deltalake/src/main/java/org/projectnessie/deltalake/ContentKeyUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Arrays;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.fs.Path;
+import org.projectnessie.model.ContentKey;
+
+/**
+ * Utility class that produces a {@link ContentKey} instance from a Hadoop {@link Path} or a file
+ * path string.
+ */
+public class ContentKeyUtil {
```
nastra, line 28: given that this is really only required for Delta, I decided to put this into the deltalake package

clients/deltalake/src/main/java/org/projectnessie/deltalake/ContentKeyUtil.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  public static ContentKey fromFilePathString(String path) {
+    Preconditions.checkArgument(null != path && !path.isEmpty(), INVALID_PATH_MSG, path);
+    // we can't simply do a path.split("/") as that would result with
+    // "/tmp/a/b/c in a string array where the first element is an empty string,
+    // which isn't supported by ContentKey
+    String[] elements = StringUtils.stripEnd(StringUtils.stripStart(path, "/"), "/").split("/");
+    Preconditions.checkState(
+        elements.length > 0 && !Arrays.asList(elements).contains(""), INVALID_PATH_MSG, path);
```
XN137, line 46: nit: isnt the 2nd condition already checked inside `ContentKey` ?

clients/deltalake/src/main/java/org/projectnessie/deltalake/ContentKeyUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Arrays;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.fs.Path;
+import org.projectnessie.model.ContentKey;
+
+/**
+ * Utility class that produces a {@link ContentKey} instance from a Hadoop {@link Path} or a file
+ * path string.
+ */
+class ContentKeyUtil {
```
XN137, line 28: not sure if this is a good idea, but in order to not pollute the namespace, should this be `DeltaContentKeyUtil` maybe?
or `DeltaContentKey`.

nit: maybe add empty private constructor (for utility classes)

XN137 (dismissed):
thanks for doing this followup, lgtm

clients/deltalake/src/main/java/org/projectnessie/deltalake/ContentKeyUtil.java
nastra: it is, but the error msg can be a bit misleading after the path string was modified and passed to ContentKey

clients/deltalake/src/main/java/org/projectnessie/deltalake/ContentKeyUtil.java
nastra: done

XN137 (dismissed):

XN137 (approved):

PR #3656, opened 2022-03-18
Run scala:doc-jar on PRs along with javadoc task

nastra:
We had PRs fail after being merged to main because the `doc-jar` task
failed (since it is only part of the `release` profile).
However, given that we already run `javadoc` on PRs, we should do the
same for scala.

### DIFF
--- a/clients/deltalake/pom.xml
+++ b/clients/deltalake/pom.xml
@@ -308,6 +308,9 @@
                 <goals>
                   <goal>doc-jar</goal>
                 </goals>
+                <configuration>
+                  <scalaVersion>${scala2.12.version}</scalaVersion>
+                </configuration>
               </execution>
             </executions>
           </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -818,7 +818,6 @@
           <version>4.6.1</version>
           <configuration>
             <sendJavaToScalac>false</sendJavaToScalac>
-            <scalaVersion>${scala2.12.version}</scalaVersion>
           </configuration>
         </plugin>
         <plugin>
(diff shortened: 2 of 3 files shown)


XN137 (approved):
+1 since it makes sense to prevent breaking the main build
... hoping there are no negative sideeffects i.e. much longer CI run times :sweat_smile:

dimas-b (approved):

snazy:
@nastra wanna get this done?

nastra:
> @nastra wanna get this done?

no resources atm to take care of this :(

clients/deltalake/pom.xml
```diff
@@ -308,6 +308,9 @@
                 <goals>
                   <goal>doc-jar</goal>
                 </goals>
+                <configuration>
+                  <scalaVersion>${scala2.12.version}</scalaVersion>
```
XN137, line 312: does it not warrant a comment why only for this usage of the plugin we have to use a specific version and in the others we dont?

pom.xml
```diff
@@ -818,7 +818,6 @@
           <version>4.6.1</version>
           <configuration>
             <sendJavaToScalac>false</sendJavaToScalac>
-            <scalaVersion>${scala2.12.version}</scalaVersion>
```
XN137, line 821: what version will be used now that this is no longer specified?
was this always wrong or how is this related now to generating scala doc?

XN137 (commented):
theoretically +1 since CI is green i guess but i dont really understand it tbh (see questions)

pom.xml
nastra: setting this version globally resulted in a failure when the gatling module was executed. All of the other modules (except deltalake were already specifiying scala 2.12). Only the gatling module requires scala 2.13 but as I said, setting the version globally here resulted in issues

clients/deltalake/pom.xml
nastra: all of the other scala modules were already explicitly using scala 2.12, this is just aligning the deltalake module

PR #3660, opened 2022-03-21
SQL Extension: Fix Create reference from a hash on non-default reference

### LINKED ISSUE #3659: SQL Extension: Failure in Create reference from a hash on non-default reference
CREATE BRANCH dev IN nessie FROM MAIN;
USE REFERENCE dev IN nessie
CREATE TABLE nessie.db.tbl (id int, name string)
INSERT INTO nessie.db.tbl select 23, "test"
CREATE BRANCH test IN nessie FROM dev; --> fails as it looks for dev's latest hash in 'main' branch instead of 'dev' branch.

`Could not find commit '0eed82af48bbb879055b477862d9f9ec0ac1c975ca46e1ff373463bc700a2568' in reference 'main'.
`

Reason: Source ref name is not set in the code. Hence, it is looking for default reference.

ajantha-bhat:
_Steps to reproduce:_
CREATE BRANCH dev IN nessie FROM main;
USE REFERENCE dev IN nessie
CREATE TABLE nessie.db.tbl (id int, name string)
INSERT INTO nessie.db.tbl select 23, "test"
CREATE BRANCH test IN nessie FROM dev; --> fails as it looks for dev's latest hash in 'main' branch instead of 'dev' branch.

Could not find commit '0eed82af48bbb879055b477862d9f9ec0ac1c975ca46e1ff373463bc700a2568' in reference 'main'.

`Reason`: Source ref name is not set in the code. Hence, it is looking for default reference. There was no testcase to catch this scenario. 

Fixes #3659

### DIFF
--- a/clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java
+++ b/clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java
@@ -300,6 +300,48 @@ public abstract class AbstractSparkSqlTest {
   }
 
   @Test
+  void testCreateReferenceFromHashOnNonDefaultBranch() throws NessieNotFoundException {
+    assertThat(sql("CREATE BRANCH %s IN nessie FROM main", refName))
+        .containsExactly(row("Branch", refName, hash));
+    assertThat(api.getReference().refName(refName).get()).isEqualTo(Branch.of(refName, hash));
+    sql("USE REFERENCE %s IN nessie", refName);
+    sql("CREATE TABLE nessie.db.tbl (id int, name string)");
+    sql("INSERT INTO nessie.db.tbl select 23, \"test\"");
+    String newHash =
+        api.getCommitLog()
+            .refName(refName)
+            .maxRecords(1)
+            .get()
+            .getLogEntries()
+            .get(0)
+            .getCommitMeta()
+            .getHash();
+
+    String tempRef = refName + "_temp";
+    sql("CREATE BRANCH %s IN nessie FROM %s", tempRef, refName);
+    assertThat(api.getReference().refName(tempRef).get()).isEqualTo(Branch.of(tempRef, newHash));
+
+    String tag = refName + "_temp_tag";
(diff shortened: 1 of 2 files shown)


clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java
```diff
@@ -299,6 +299,48 @@ void testCreateTagInFrom() throws NessieNotFoundException {
...
+
+    String tempRef = refName + "_temp";
+    sql("CREATE BRANCH %s IN nessie FROM %s", tempRef, refName);
+    assertThat(api.getReference().refName(tempRef).get()).isEqualTo(Branch.of(tempRef, newHash));
+
+    String tag = refName + "_temp_tag";
+    sql("CREATE TAG %s IN nessie FROM %s", tag, refName);
+    assertThat(api.getReference().refName(tag).get()).isEqualTo(Tag.of(tag, newHash));
+
+    assertThat(sql("DROP BRANCH %s IN nessie", refName)).containsExactly(row("OK"));
```
nastra, line 328: those 3 DROP statements aren't really needed as they are being cleaned up automatically

ajantha-bhat: other testcases also has the same pattern. So, I followed the same.

nastra: yes I know, and this should rather be cleaned up. We should just have separate tests for dropping a branch/tag and not pollute existing tests with drop statements

nastra (approved):

clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java
ajantha-bhat: testCreateBranch, testCreateTag, testCreateBranchIn, testCreateTagIn, testCreateBranchInFrom, testCreateTagInFrom
all these has it, should I remove from them aswell or let's keep it?

PR #3676, opened 2022-03-21
Prevent maven-shade-plugin hang

snazy:
The maven-shade-plugin is "allergic" to some dependency configurations, which cause
it to loop infinitely. This can be reproduced with `mvnd clean install -DskipTests`,
which ends up with the two spark extensions modules and the content generator tool
to "build forever".

There are two workarounds for this:
* do not generate a dependency reduced pom (not applicable)
* exclude problematic dependencies - the option that this PR implements by excluding
  test and some provided dependencies.

Note: the end result's still the same.

See also https://issues.apache.org/jira/browse/MSHADE-148 (open for ~9 years now).


nastra (approved):
LGTM if it works

PR #3677, opened 2022-03-21
Revert "Prevent maven-shade-plugin hang (#3676)"

snazy:
This reverts commit 76123084783aa8bb68be3c2e6a0c4f6a90e58077.


snazy:
This is ... annoying.
It did work when I submitted #3676), but doesn't work anymore (infinite loop again).

PR #3691, opened 2022-03-22
Bump codecov/codecov-action from deprecated v1 to v2

snazy:


XN137 (approved):
some breaking changes but nothing that we use:
https://github.com/codecov/codecov-action#migration-from-v1-to-v2

so lgtm

PR #3698, opened 2022-03-23
Turn off the unicode check in ErrorProne

ajantha-bhat:
Turn off the unicode check in ErrorProne as it throws invalid error in IDE.

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1278,6 +1278,7 @@ limitations under the License.
             <configuration>
               <compilerArgs combine.children="append">
                 <arg>-XDcompilePolicy=simple</arg>
+                <!-- Turning off Unicode check as it gives invalid errors in ErrorProne 2.11.0 -->
                 <arg>-Xplugin:ErrorProne -Xep:UnicodeInCode:OFF -XepExcludedPaths:.*/target/generated-*sources/.*</arg>
               </compilerArgs>
               <annotationProcessorPaths combine.children="append">
--- a/pom.xml
+++ b/pom.xml
@@ -1278,7 +1278,7 @@ limitations under the License.
             <configuration>
               <compilerArgs combine.children="append">
                 <arg>-XDcompilePolicy=simple</arg>
-                <arg>-Xplugin:ErrorProne -XepExcludedPaths:.*/target/generated-*sources/.*</arg>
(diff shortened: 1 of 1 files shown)


XN137:
as an alternative we could try adding `-Xep:UnicodeInCode:OFF` here instead:
https://github.com/projectnessie/nessie/blob/eaca0145b4c4c76bda3dbbbbd30a07752f4d6552/pom.xml#L1281
(and dont lose anything because that check was only added in `2.11.0`)

wdyt?

ajantha-bhat:
> as an alternative we could try adding -Xep:UnicodeInCode:OFF here instead:

@XN137 : It will mask the actual errors also ?

XN137:
it will disable the check yes, but as i said, going back to the old version does not have this check at all.

snazy:
Don't see this error with IDEA Build #IU-213.7172.25, built on March 15, 2022

ajantha-bhat:
> Don't see this error with IDEA Build #IU-213.7172.25, built on March 15, 2022

@snazy : we merged this (https://github.com/projectnessie/nessie/pull/3640) two days ago

ajantha-bhat:
> it will disable the check yes, but as i said, going back to the old version does not have this check at all.

ok for me, Let's see what others think.

snazy:
> > Don't see this error with IDEA Build #IU-213.7172.25, built on March 15, 2022
> 
> @snazy : we merged this (#3640) two days ago

Yes??? Can you maybe check the build number of IDEA that you're using??

ajantha-bhat:
> Yes??? Can you maybe check the build number of IDEA that you're using??

where can I find it? I googled it, it pointed me to intellij version instead of build version.
Also build itself is failing for me, so there won't be a build number ?

image

XN137:
it still happens for me, even with the march 15 intellij build:
image

ajantha-bhat:
build number
image

XN137 (dismissed):
maybe add a comment why we disable it (so future us can now when we could try to re-enable it again)

but +1 as is

ajantha-bhat:
Even if it is solved by upgrading IDE, we can't force everyone to use the latest IDE. So, tuning it off as @XN137 suggested.
Reverting also not a good idea as someone might merge it again from dependabot.

ajantha-bhat:
> maybe add a comment why we disable it (so future us can now when we could try to re-enable it again)
Added the comment.

nastra (approved):

PR #3702, opened 2022-03-23
Filter out empty namespaces from non-namespace Content objects

nastra:


servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -221,7 +221,10 @@ public GetNamespacesResponse getNamespaces(MultipleNamespacesParams params)
       MultipleNamespacesParams params, BranchName branch) throws ReferenceNotFoundException {
     try (Stream<WithType<Key, Type>> stream =
         getImplicitNamespacesStream(params.getNamespace(), branch)) {
-      return stream.map(this::implicitNamespaceFrom).collect(Collectors.toList());
+      return stream
```
snazy, line 224: Feels like the two calls to getStore().getKeys() hwere and in getExplicitlyCreatedNamespaces() should be unified - in a follow-up.

nastra: yep will do a follow-up on this

snazy (approved):

PR #3732, opened 2022-03-27
Simplify ContentKey.of a little

snazy:


model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -75,7 +74,7 @@ public static ContentKey of(Namespace namespace, String name) {
 
   public static ContentKey of(String... elements) {
     Objects.requireNonNull(elements, "Elements array must not be null");
-    return ImmutableContentKey.builder().elements(Arrays.asList(elements)).build();
+    return ImmutableContentKey.builder().addElements(elements).build();
```
XN137, line 77: we have the same pattern in `Namespace.java`

XN137 (approved):

PR #3733, opened 2022-03-27
Let Quarkus Server tests run on Java 17+

snazy:
See https://github.com/quarkusio/quarkus/issues/24492

### DIFF
--- a/servers/quarkus-server/pom.xml
+++ b/servers/quarkus-server/pom.xml
@@ -168,6 +168,7 @@
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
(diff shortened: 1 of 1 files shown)


servers/quarkus-server/pom.xml
```diff
@@ -168,6 +168,7 @@
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <argLine>@{argLine} -Dfile.encoding=UTF-8 --add-opens=java.base/java.util=ALL-UNNAMED</argLine>
```
XN137, line 171: imo we should add a comment to explain why we added this (and to know when we can remove it again)

XN137 (approved):
+1 after added comment

PR #3792, opened 2022-03-31
Prevent explicit creation of empty namespaces

nastra:
This also includes minor simplifications in Namespace tests for different Web servers

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.services.impl;
 
+import com.google.common.base.Preconditions;
 import java.security.Principal;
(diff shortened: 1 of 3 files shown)


servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -88,6 +89,8 @@ public Namespace createNamespace(NamespaceParams params)
           };
 
       Namespace namespace = params.getNamespace();
+      Preconditions.checkArgument(
+          !Namespace.EMPTY.name().equals(namespace.name()), "Namespace name cannot be empty");
```
snazy, line 93: s/cannot/must not/

snazy: Why not just `namespace.name().isEmpty()` ?

snazy (approved):

PR #3793, opened 2022-03-31
Serialize NessieApi

ajantha-bhat:
to pass from spark driver to executor.

After changing bunch of classes to Serializable and making `SSLContext` and `ObjectMapper` transient because of below errors.

SSLContext:
`java.io.NotSerializableException: javax.net.ssl.SSLContext
`
ObjectMapper:
`Caused by: java.io.NotSerializableException: com.fasterxml.jackson.databind.jsontype.impl.TypeNameIdResolver
`
**Also modified GC spark code to send the nessie api from driver to executor and (close the API in driver)**

TODO: probably need to reconstruct SSLContext at executor by taking config, else HTTPS connections doesn't work.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
@@ -16,6 +16,7 @@
 package org.projectnessie.client.http;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.Serializable;
 import java.net.URI;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
@@ -28,7 +29,7 @@ import javax.net.ssl.SSLContext;
  * <p>Assumptions: - always send/receive JSON - set headers accordingly by default - very simple
  * interactions w/ API - no cookies - no caching of connections. Could be slow
  */
-public class HttpClient {
+public class HttpClient implements Serializable {
 
   private final HttpRuntimeConfig config;
 
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpConfigClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpConfigClient.java
@@ -15,10 +15,11 @@
  */
 package org.projectnessie.client.http;
 
+import java.io.Serializable;
 import org.projectnessie.api.http.HttpConfigApi;
 import org.projectnessie.model.NessieConfiguration;
 
-class HttpConfigClient implements HttpConfigApi {
+class HttpConfigClient implements HttpConfigApi, Serializable {
   private final HttpClient client;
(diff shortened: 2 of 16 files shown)


tomekl007:
Can you link corresponding spark changes?

nastra:
TBH I don't really understand why we need to make all of those APIs serializable. Can you please elaborate and describe the actual problem you're facing?

ajantha-bhat:
@nastra : please refer thread https://github.com/projectnessie/nessie/pull/3421#discussion_r836502224

I was just sending a api conf from driver to executor and building new instance. But because of spark lazy iterator, I couldn't free the api resources in negative scenarios. After lot of experimentation, decided that I don't create new instance in executor, I send from driver. But sending from driver to executor needs a serailizable instance.

ajantha-bhat:
> Can you link corresponding spark changes?

Update the description, it is in the same PR, GC module spark code.

nastra:
> @nastra : please refer thread [#3421 (comment)](https://github.com/projectnessie/nessie/pull/3421#discussion_r836502224)
> 
> I was just sending a api conf from driver to executor and building new instance of NessieAPI. But because of spark lazy iterator, I couldn't free the api resources in negative scenarios. After lot of experimentation, decided that I don't create new instance in executor, Instead I can send from driver and fee resource in driver after usage. But sending from driver to executor needs a serailizable instance.

I'm -1 on making the APIs serializable for solving the problem you're facing. We rather need to solve the actual root issue and not create workarounds

tomekl007:
> > @nastra : please refer thread [#3421 (comment)](https://github.com/projectnessie/nessie/pull/3421#discussion_r836502224)
> > I was just sending a api conf from driver to executor and building new instance of NessieAPI. But because of spark lazy iterator, I couldn't free the api resources in negative scenarios. After lot of experimentation, decided that I don't create new instance in executor, Instead I can send from driver and fee resource in driver after usage. But sending from driver to executor needs a serailizable instance.
> 
> I'm -1 on making the APIs serializable for solving the problem you're facing. We rather need to solve the actual root issue and not create workarounds

@ajantha-bhat can you check if other, "spark friendly" HTTP clients are serializable as well? If they are, it means that's the proper solution for spark. And if someone will want to use our `NessieAPI` form Spark, they will hit the same problem, therefore making it serializable would be a reasonable choice.

snazy:
I do not get what this change shall actually solves in which way.
We're talking about e.g. OS level resources here that need to be properly closed.

Definitely -1 on making the Nessie API serializable.

ajantha-bhat:
> I do not get what this change shall actually solves in which way.

this solves the problem of cleaning up resources of NessieAPI at executor by avoiding creating new instance in executor and pass the instance from driver. Because the original object reside in driver, driver is closing the resources.

PR #3813, opened 2022-04-02
Squash merge

### LINKED ISSUE #3809: Squash-Merge
Update the Nessie merge operation to squash the commits to be merged, this should become the default.

Currently, the commits from the source branch to be merged are copied onto the target branch.

Depends on point 1 in #2195 + #2311

snazy:
Updates the merge-operation to squash the commits to be merged, if there is more than 1
commit to be merged.

The most recent change to any content-key will be present in the squashed commit,
intermediate changes will not be present in the squashed commit.

Commit messages will be concatenated with a separating line consisting of a sequence of
`-` characters.

Commit properties are combined, last property value will be present in the squashed
commit.

Commit author will be reset to the user issuing the merge operation in a squashed
commit. Commit author-time will be reset to the current timestamp for a squashed commit.

This API-change is not changing the signature, but the behavior changes, as the default
is then to squash the commits to merge. I.e. not specifying the `keepIndividualCommits`
parameter for a merge operation means "squash" the commits. To use the previous behavior,
pass `keepIndividualCommits=true`.

Fixes #3809

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/api/MergeTransplantBuilder.java
+++ b/clients/client/src/main/java/org/projectnessie/client/api/MergeTransplantBuilder.java
@@ -13,21 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.projectnessie.versioned.persist.adapter;
+package org.projectnessie.client.api;
 
-import com.google.protobuf.ByteString;
-import java.util.List;
-import java.util.function.Function;
-import org.immutables.value.Value;
+import javax.validation.constraints.Pattern;
+import org.projectnessie.model.Validation;
 
-public interface MetadataRewriteParams extends ToBranchParams {
+public interface MergeTransplantBuilder<R extends MergeTransplantBuilder<R>>
+    extends OnBranchBuilder<R> {
+  R fromRefName(
+      @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
+          String fromRefName);
 
-  /** Whether to keep the individual commits and do not squash the commits to merge. */
-  @Value.Default
-  default boolean keepIndividualCommits() {
-    return false;
-  }
-
-  /** Function to rewrite the commit-metadata. */
-  Function<List<ByteString>, ByteString> getUpdateCommitMetadata();
+  R keepIndividualCommits(boolean keepIndividualCommits);
 }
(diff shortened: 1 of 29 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -387,24 +369,61 @@ protected Hash transplantAttempt(
...
+      // (no need to verify the global states during a merge)
+      // re-apply commits in 'sequenceToTransplant' onto 'targetBranch'
+      toHead =
+          copyCommits(
+              ctx,
+              timeInMicros,
+              toHead,
+              commitsToMergeChronological,
+              newKeyLists,
+              m -> params.getUpdateCommitMetadata().apply(Collections.singletonList(m)));
```
dimas-b, line 410: The relationship between the metadata rewrite function (with an implied `keepIndividualCommits` parameters ) and the explicit `keepIndividualCommits` parameters seems to be quite obscure to me.

Could we have a separate interface with _two_ methods for the metadata rewriter?

```
interface MetadataRewriter {
  CommitMeta rewriteSingle(CommitMeta metadata);
  CommitMeta squash(List<CommitMeta> metadata);
}
```

I suppose we'd only have to have one implementation this way - inside `BaseApiImpl`.

WDYT?

clients/client/src/main/java/org/projectnessie/client/api/MergeTransplantBuilder.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.validation.constraints.Pattern;
+import org.projectnessie.model.Validation;
+
+public interface MergeTransplantBuilder<R extends MergeTransplantBuilder<R>>
+    extends OnBranchBuilder<R> {
+  R fromRefName(
+      @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
+          String fromRefName);
+
+  R keepIndividualCommits(boolean keepIndividualCommits);
```
ajantha-bhat, line 27: By default squash is enabled with this during merge and transplant. If user wants to avoid squash they need to explicitly set this? 
I think other way around is better.

**But, If we squash commits we lose time travel support of those individual commits,** 
So, do we really need to support this feature ?

ajantha-bhat (commented):
If we squash commits we lose time travel support (using commit hashes) of those individual commits,
So, do we really need to support this feature ?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Yea, looks much better that way.

snazy:
The behavior re: squash by default follows the Git semantics.
Squashing makes a lot of sense because you see the atomic commit of a transaction and not the individual pieces of a transaction.

dimas-b (approved):

PR #3626, opened 2022-03-16
Support stored procedure for Identify expired contents

ajantha-bhat:
This PR depends on #3829. Need a rebase after that merge.

This PR focus on (Only top commit maps to this PR changes now)
- Implement `identify()`" stored procedure for `GCImpl#identifyExpiredContents()` functionality.
- Extend the same testcases and override performing gc to call procedure instead of API.

PR #3814, opened 2022-04-02
Refactor gc-base module to be reusable for gc-iceberg module

ajantha-bhat:
gc-iceberg module will depend on the gc-base module test jar and reuses its test cases by overriding the performGC to use call procedure instead of api as in https://github.com/projectnessie/nessie/pull/3626

- Make some private test methods to be protected for overriding.
- Move the timestamp to instant conversion to GCUtil to be reused in stored procedure.

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
@@ -18,9 +18,11 @@ package org.projectnessie.gc.base;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.lang.reflect.InvocationTargetException;
+import java.time.Instant;
 import java.util.Map;
 import java.util.Spliterator;
 import java.util.Spliterators;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.mutable.MutableBoolean;
@@ -53,6 +55,13 @@ public final class GCUtil {
     } catch (JsonProcessingException e) {
       throw new RuntimeException(e);
     }
+  }
+
+  public static Instant getInstantFromMicros(Long microsSinceEpoch) {
+    return Instant.ofEpochSecond(
+        TimeUnit.MICROSECONDS.toSeconds(microsSinceEpoch),
+        TimeUnit.MICROSECONDS.toNanos(
+            Math.floorMod(microsSinceEpoch, TimeUnit.SECONDS.toMicros(1))));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


gc/gc-base/src/test/java/org/projectnessie/gc/base/AbstractRestGC.java
```diff
@@ -90,7 +90,7 @@ void fillExpectedContents(Branch branch, int numCommits, List<Row> expected)
             });
   }
 
-  void performGc(
+  protected void performGc(
```
ajantha-bhat, line 93: overriding in another module (different package name), hence it cannot be package private.

ajantha-bhat: refer here: https://github.com/projectnessie/nessie/pull/3626

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCUtil.java
```diff
@@ -55,6 +57,13 @@ public static Reference deserializeReference(String reference) {
     }
   }
 
+  public static Instant getInstantFromMicros(Long microsSinceEpoch) {
```
tomekl007, line 60: since we are making it public, we need unit tests for it

ajantha-bhat: Added

gc/gc-base/src/test/java/org/projectnessie/gc/base/TestInstantConversion.java
```diff
@@ -0,0 +1,39 @@
...
+    Instant now = Instant.now();
+    long time = now.getEpochSecond();
+    long nano = now.getNano();
+    // current time in microseconds since epoch
+    long micro = TimeUnit.SECONDS.toMicros(time) + TimeUnit.NANOSECONDS.toMicros(nano);
+    Instant instantFromMicros = GCUtil.getInstantFromMicros(micro);
+    // Even though instant is capable of nanosecond precision,
+    // Instant.now() gives micro second precision by default.
+    // Because of that the below validation can pass.
+    assertThat(instantFromMicros).isEqualTo(now);
```
tomekl007, line 37: can we add a test that validates that Nanos is properly extracted as well? (`System.nanoTime()`)

ajantha-bhat: but the `getInstantFromMicros` expects a microseconds precision. So, instant will always be in micro seconds precision. 
Can you provide a full example  of the expected test?

tomekl007: ok, sounds good.

tomekl007 (approved):

Issue #3692, opened 2022-03-22
Nessie GC module should depend on client.nessie.version after next Iceberg release.


ajantha-bhat:
As nessie gc module [in #3421] depends on Iceberg spark run time jar, it should depend on `client.nessie.version` similar to other modules.

It cannot depend on it now itself as `client.nessie.version` is `0.18.0` which doesn't contain the gc related client changes (like detached reference, etc).
So, currently depending on the latest Nessie client. Once, Iceberg makes a new release, gc module has to be updated to  `client.nessie.version` 

note that if `gc-module` doesn't depends on `client.nessie.version`, if latest client breaks compatibility, gc module compile will fail.

ajantha-bhat:
Also fallback to uber jar (iceberg-spark3-runtime) dependency while depending on the `client.nessie.version`

ajantha-bhat:
Tried working on https://github.com/ajantha-bhat/nessie/commit/1bf78141b2e2f895d35bdcb417c3edc537916065

a) `nessie-jaxrs-testextension` and` nessie-jaxrs-tests` also probably has to depend on `client.nessie.version` along with client.
b) Cannot depend on a single iceberg uber jar (iceberg-spark-runtime jar) because finally, the Nessie module in iceberg has to depend on nessie-gc module. So that single runtime jar from iceberg can support nessie gc too. (if not we get that shaded jackson annotation errors)

PR #3829, opened 2022-04-04
Introduce gc-iceberg module to hold stored procedures based on gc-base

ajantha-bhat:
- Introduce `gc-Iceberg` module to handle Iceberg specific procedures.
- Supported `NessieIcebergGcSparkCatalog` and `NessieIcebergGcSparkSessionCatalog` to store the GC procedures under "nessie_gc" namespace.
- Added a dummy procedure to verify the framework.

### DIFF
--- a/gc/gc-iceberg/pom.xml
+++ b/gc/gc-iceberg/pom.xml
@@ -1,0 +1,121 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (C) 2020 Dremio
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.projectnessie</groupId>
+    <artifactId>nessie-gc</artifactId>
+    <version>0.24.1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>nessie-gc-iceberg</artifactId>
+
+  <name>Nessie - GC - Iceberg</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-jaxrs-testextension</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.projectnessie</groupId>
+      <artifactId>nessie-jaxrs-tests</artifactId>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <!--    spark dependencies -->
+    <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-sql_2.12</artifactId>
+      <version>${spark31.version}</version>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.apache.parquet</groupId>
+          <artifactId>parquet-hadoop</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>log4j-over-slf4j</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!--    Iceberg dependencies -->
+    <dependency>
+      <groupId>org.apache.iceberg</groupId>
+      <artifactId>iceberg-nessie</artifactId>
+      <version>${iceberg.version}</version>
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.projectnessie</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.iceberg</groupId>
+      <artifactId>iceberg-spark3</artifactId>
+      <version>${iceberg.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.iceberg</groupId>
+      <artifactId>iceberg-spark3-extensions</artifactId>
+      <version>${iceberg.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+      </plugin>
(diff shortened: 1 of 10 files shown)


gc/gc-iceberg/src/main/java/org/apache/iceberg/spark/procedures/BaseGcProcedure.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.iceberg.spark.procedures;
+
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+
+/**
+ * This class is a "bridge" between the Iceberg-Spark package and the Nessie package, because {@link
```
tomekl007, line 21: nit: too breat the doc line a bit earlier

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/DummyProcedure.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.spark.sql.catalyst.InternalRow;
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import org.apache.spark.sql.connector.iceberg.catalog.ProcedureParameter;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+
+/** Dummy procedure that accepts an input string and returns it. */
+public class DummyProcedure extends BaseGcProcedure {
```
tomekl007, line 28: why do we need a `DummyProcedure` in the production code? Shouldn't it be in the test code?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/GcProcedureUtil.java
```diff
@@ -0,0 +1,61 @@
...
+
+  static boolean isGcNamespace(Identifier identifier) {
+    return Arrays.equals(NAMESPACE_ARRAY, identifier.namespace());
+  }
+
+  static BaseGcProcedure loadGcProcedure(Identifier procedureIdentifier, TableCatalog catalog)
+      throws NoSuchProcedureException {
+    if (DummyProcedure.PROCEDURE_NAME.equals(procedureIdentifier.name())) {
+      return new DummyProcedure(catalog);
+    }
```
tomekl007, line 44: Will it be extended to a procedure that does something?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/GcProcedureUtil.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.spark.sql.connector.catalog.TableCatalog;
+import org.apache.spark.unsafe.types.UTF8String;
+import scala.collection.JavaConverters;
+import scala.collection.Seq;
+
+final class GcProcedureUtil {
+
+  private GcProcedureUtil() {}
+
+  static final String NAMESPACE = "nessie_gc";
```
tomekl007, line 33: We should document that `nessie_gc` namespace cannot be used by the Nessie clients now.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/NessieIcebergGcSparkSessionCatalog.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.spark.sql.connector.iceberg.catalog.Procedure;
+
+/**
+ * Extends {@link SparkSessionCatalog} to load the Nessie GC procedures in "nessie_gc" namespace. As
+ * there is no other way to "plug in" custom procedures in Iceberg yet.
+ *
+ * <p>This extension can be removed after Iceberg supports pluggable stored procedures.
+ */
+@SuppressWarnings("unused")
+public class NessieIcebergGcSparkSessionCatalog<T extends TableCatalog & SupportsNamespaces>
```
tomekl007, line 32: why do we need to provide this and `NessieIcebergGcSparkCatalog`?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/NessieIcebergGcSparkCatalog.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.spark.sql.connector.iceberg.catalog.Procedure;
+
+/**
+ * Extends {@link SparkCatalog} to load the Nessie GC procedures in "nessie_gc" namespace. As there
+ * is no other way to "plug in" custom procedures in Iceberg yet.
+ *
+ * <p>This extension can be removed after Iceberg supports pluggable stored procedures.
+ */
+@SuppressWarnings("unused")
+public class NessieIcebergGcSparkCatalog extends SparkCatalog {
```
tomekl007, line 30: nit: maybe change the name to `...ProcedureLoader` or similar?

gc/gc-iceberg/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+      <scope>provided</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.projectnessie</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.iceberg</groupId>
```
ajantha-bhat, line 100: Will use run time jar after next iceberg release.
Same as https://github.com/projectnessie/nessie/issues/3692

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/NessieIcebergGcSparkSessionCatalog.java
ajantha-bhat: These stored procedures are called from Iceberg-spark, Currently Iceberg doesn't support adding custom procedures. 
So, until they support this loading code has to stay in Nessie's gc-iceberg module.

I think java doc of this class explains this already.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/NessieIcebergGcSparkCatalog.java
ajantha-bhat: As it extends the `SparkCatalog`, current name is more suitable. Also this name the user will use for configuration from Iceberg.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/GcProcedureUtil.java
ajantha-bhat: I don't think we have that restriction. It can still be used as normal namespace.
Only restriction is that while calling the stored procedure, `nessie_gc` namespace is must. 
This restriction is already added in java doc of `NessieIcebergGcSparkCatalog` and `NessieIcebergGcSparkSessionCatalog`

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/DummyProcedure.java
ajantha-bhat: It will be removed after #3626 , because that PR was too big, I split it into this.

`DummyProcedure` is enough test the framework.

This class cannot be in a test code as the `GcProcedureUtil#loadGcProcedure` source code uses it.
I can still have some reflection way of loading it, but that looks ugly. Hence keeping it in the source.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/GcProcedureUtil.java
ajantha-bhat: Nope. This procedure is just to test framework. 
Instead of this two more gc procedures will be added `identify()` and `expire()`

gc/gc-iceberg/src/main/java/org/apache/iceberg/spark/procedures/BaseGcProcedure.java
ajantha-bhat: I think it is spotless doing ;)

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/DummyProcedure.java
tomekl007: I am not sure if having such a code in the sources of an open-source project won't be confusing for the users. I think we need a 2nd opinion about that, maybe @snazy or @nastra?

ajantha-bhat: Yeah, suggestions are welcome. This was the only way I found to split the PR #3626 to be a testable sub PR.
If we cannot have this, then I have to close this PR and review #3626 as a whole PR.

snazy:
The total number of PRs and changes and refactorings is getting quite high and becomes really hard to review.

I suggest another approach:
1. Close all existing PRs,
2. Finish GC including all tests and (Iceberg) integrations and performance reviews etc,
3. Get feedback about the implementation,
4. THEN split up the GC code into smaller PRs.

ajantha-bhat:
> I suggest another approach:

> Close all existing PRs,
Finish GC including all tests and (Iceberg) integrations and performance reviews etc,
Get feedback about the implementation,
THEN split up the GC code into smaller PRs.

OK. I will do the same.

PR #3845, opened 2022-04-05
Improve test runtime for REST tests a little

snazy:


nastra (approved):
does this really make any substantial difference?

PR #3848, opened 2022-04-05
Load correct view metadata for a given ref

### LINKED ISSUE #3734: The load and loadDefinition for NessieViewCatalog always return latest view info
**What happened**:
The loadDefinition and load in NessieViewCatalog return unexpected results in some cases.

When I call loadDefinition or load a view, I should not always get the info of the latest view.
The NessieViewCatalog should fetch the version and return the expected version.

Create branch A and create view V1 on it.
Create branch B from branch A and update view V1 on branch B  to V2.
When I call loadDefinition or load the view V2 on branch B, I will get the correct results.
When I call loadDefinition or load the view V1 on branch A, I will get the same result as V2.

**Details**:
- Nessie server type (docker/built from source) and version:
Build from source. 0.22
NessieViewCatalog

nastra:
This makes sure that we load the correct view metadata when specifying a
particular reference.

fixes #3734

### DIFF
--- a/clients/iceberg-views/src/main/java/org/apache/iceberg/nessie/NessieViewOperations.java
+++ b/clients/iceberg-views/src/main/java/org/apache/iceberg/nessie/NessieViewOperations.java
@@ -17,7 +17,10 @@ package org.apache.iceberg.nessie;
 
 import static org.apache.iceberg.view.ViewUtils.toCatalogTableIdentifier;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Predicate;
 import org.apache.iceberg.catalog.TableIdentifier;
 import org.apache.iceberg.exceptions.CommitFailedException;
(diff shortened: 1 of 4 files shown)


snazy (commented):
Is this basically changed to what we do for Tables in Iceberg for Nessie?

nastra:
@snazy yes

dimas-b (approved):

clients/iceberg-views/src/main/java/org/apache/iceberg/nessie/NessieViewOperations.java
```diff
@@ -26,7 +29,11 @@
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.util.Tasks;
 import org.apache.iceberg.view.BaseMetastoreViewOperations;
+import org.apache.iceberg.view.BaseVersion;
+import org.apache.iceberg.view.HistoryEntry;
+import org.apache.iceberg.view.Version;
```
dimas-b, line 34: Does this mean we have a circular dependency with Iceberg?

nastra: yes but the code here will move eventually into iceberg

PR #3850, opened 2022-04-05
Use group separator (u001D) character instead of zero byte for Namespace/ContentKey

nastra:
For backward-compatibility we still allow the zero byte, but convert it
to the group sep. character, since that's our new canonical representation
of a "." delimiter within a single element name.

Note that this also means that
`Namespace.parse("a.b\u001Dc.d")` is the same as `Namespace.parse("a.b\u0000c.d")`.

The reason for changing from the zero byte to the group sep. character is
because certain webservers (Jetty) don't allow the zero byte to be used
in the URL path
(https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/jetty-http/src/main/java/org/eclipse/jetty/http/HttpURI.java#L1203-L1206).
This means that the mentioned path check in Jetty fails even before for
example the `NamespaceParamConverter` gets called. Due to that we
decided to switch our canonical representation of a ContentKey/Namespace
to use a new character instead of the zero byte.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/ContentKey.java
+++ b/model/src/main/java/org/projectnessie/model/ContentKey.java
@@ -15,6 +15,10 @@
  */
 package org.projectnessie.model;
 
+import static org.projectnessie.model.UriUtil.DOT_STRING;
+import static org.projectnessie.model.UriUtil.GROUP_SEPARATOR_STRING;
+import static org.projectnessie.model.UriUtil.ZERO_BYTE_STRING;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -37,9 +41,6 @@ import org.projectnessie.model.ImmutableContentKey.Builder;
 @JsonSerialize(as = ImmutableContentKey.class)
 @JsonDeserialize(as = ImmutableContentKey.class)
 public abstract class ContentKey {
-
-  private static final char ZERO_BYTE = '\u0000';
-  private static final String ZERO_BYTE_STRING = Character.toString(ZERO_BYTE);
 
   @NotNull
   @Size(min = 1)
@@ -91,9 +92,11 @@ public abstract class ContentKey {
         throw new IllegalArgumentException(
             String.format("Content key '%s' must not contain a null element.", elements));
       }
-      if (e.contains(ZERO_BYTE_STRING)) {
+      if (e.contains(ZERO_BYTE_STRING) || e.contains(GROUP_SEPARATOR_STRING)) {
         throw new IllegalArgumentException(
-            String.format("Content key '%s' must not contain a zero byte.", elements));
+            String.format(
+                "Content key '%s' must not contain a zero byte (\\u0000) / group separator (\\u001D).",
+                elements));
       }
       if ("".equals(e)) {
         throw new IllegalArgumentException(
@@ -123,6 +126,6 @@ public abstract class ContentKey {
 
   @Override
   public String toString() {
-    return String.join(".", getElements());
+    return String.join(DOT_STRING, getElements());
   }
 }
--- a/model/src/main/java/org/projectnessie/model/Namespace.java
+++ b/model/src/main/java/org/projectnessie/model/Namespace.java
@@ -15,6 +15,8 @@
  */
 package org.projectnessie.model;
 
+import static org.projectnessie.model.UriUtil.DOT_STRING;
+import static org.projectnessie.model.UriUtil.GROUP_SEPARATOR_STRING;
 import static org.projectnessie.model.UriUtil.ZERO_BYTE_STRING;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -41,7 +43,6 @@ import org.immutables.value.Value.Derived;
 @JsonTypeName("NAMESPACE")
 public abstract class Namespace extends Content {
 
-  private static final String DOT = ".";
   static final String ERROR_MSG_TEMPLATE =
       "'%s' is not a valid namespace identifier (should not end with '.')";
 
@@ -96,10 +97,11 @@ public abstract class Namespace extends Content {
             String.format(
                 "Namespace '%s' must not contain a null element.", Arrays.toString(elements)));
       }
-      if (e.contains(ZERO_BYTE_STRING)) {
+      if (e.contains(ZERO_BYTE_STRING) || e.contains(GROUP_SEPARATOR_STRING)) {
         throw new IllegalArgumentException(
             String.format(
-                "Namespace '%s' must not contain a zero byte.", Arrays.toString(elements)));
+                "Namespace '%s' must not contain a zero byte (\\u0000) / group separator (\\u001D).",
+                Arrays.toString(elements)));
       }
       if ("".equals(e)) {
         throw new IllegalArgumentException(
@@ -108,7 +110,7 @@ public abstract class Namespace extends Content {
       }
     }
 
-    if (DOT.equals(elements[elements.length - 1])) {
+    if (DOT_STRING.equals(elements[elements.length - 1])) {
       throw new IllegalArgumentException(
           String.format(ERROR_MSG_TEMPLATE, Arrays.toString(elements)));
     }
@@ -142,7 +144,7 @@ public abstract class Namespace extends Content {
     if (identifier.isEmpty()) {
       return EMPTY;
     }
-    if (identifier.endsWith(DOT)) {
+    if (identifier.endsWith(DOT_STRING)) {
       throw new IllegalArgumentException(String.format(ERROR_MSG_TEMPLATE, identifier));
     }
     return Namespace.of(UriUtil.fromPathString(identifier));
--- a/model/src/main/java/org/projectnessie/model/UriUtil.java
+++ b/model/src/main/java/org/projectnessie/model/UriUtil.java
@@ -24,7 +24,11 @@ final class UriUtil {
   private UriUtil() {}
 
   public static final char ZERO_BYTE = '\u0000';
+  public static final char DOT = '.';
+  public static final char GROUP_SEPARATOR = '\u001D';
+  public static final String DOT_STRING = ".";
   public static final String ZERO_BYTE_STRING = Character.toString(ZERO_BYTE);
+  public static final String GROUP_SEPARATOR_STRING = Character.toString(GROUP_SEPARATOR);
 
   /**
    * Convert from path encoded string to normal string.
@@ -34,7 +38,7 @@ final class UriUtil {
    */
   public static List<String> fromPathString(String encoded) {
     return Arrays.stream(encoded.split("\\."))
-        .map(x -> x.replace('\u0000', '.'))
+        .map(x -> x.replace(GROUP_SEPARATOR, DOT).replace(ZERO_BYTE, DOT))
         .collect(Collectors.toList());
   }
 
@@ -44,6 +48,8 @@ final class UriUtil {
    * @return String encoded for path use.
    */
   public static String toPathString(List<String> elements) {
-    return elements.stream().map(x -> x.replace('.', '\u0000')).collect(Collectors.joining("."));
+    return elements.stream()
+        .map(x -> x.replace(DOT, GROUP_SEPARATOR).replace(ZERO_BYTE, GROUP_SEPARATOR))
+        .collect(Collectors.joining("."));
   }
 }
--- a/model/src/test/java/org/projectnessie/model/TestContentKey.java
+++ b/model/src/test/java/org/projectnessie/model/TestContentKey.java
@@ -31,6 +31,7 @@ import org.junit.jupiter.api.parallel.Execution;
 import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
 
 @Execution(ExecutionMode.CONCURRENT)
 class TestContentKey {
@@ -233,7 +234,13 @@ class TestContentKey {
   public void validation() {
     assertThatThrownBy(() -> ContentKey.of("a", "b", "\u0000", "c", "d"))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Content key '[a, b, \u0000, c, d]' must not contain a zero byte.");
+        .hasMessage(
+            "Content key '[a, b, \u0000, c, d]' must not contain a zero byte (\\u0000) / group separator (\\u001D).");
+
+    assertThatThrownBy(() -> ContentKey.of("a", "b", "\u001D", "c", "d"))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage(
+            "Content key '[a, b, \u001D, c, d]' must not contain a zero byte (\\u0000) / group separator (\\u001D).");
   }
 
   @Test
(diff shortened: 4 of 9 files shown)


model/src/main/java/org/projectnessie/model/UriUtil.java
```diff
@@ -44,6 +59,18 @@ private UriUtil() {}
    * @return String encoded for path use.
    */
   public static String toPathString(List<String> elements) {
-    return elements.stream().map(x -> x.replace('.', '\u0000')).collect(Collectors.joining("."));
+    return elements.stream()
+        .map(x -> x.replace(DOT, ZERO_BYTE))
+        .map(
+            x -> {
+              try {
+                return URLEncoder.encode(x, "UTF-8");
```
dimas-b, line 67: Use `StandardCharsets.UTF_8`, perhaps? Then we won't have to catch `UnsupportedEncodingException`

model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -91,7 +91,7 @@ protected void validate() {
         throw new IllegalArgumentException(
             String.format("Content key '%s' must not contain a null element.", elements));
       }
-      if (e.contains(ZERO_BYTE_STRING)) {
+      if (e.contains(ZERO_BYTE_STRING) || e.contains(ZERO_BYTE_STRING_URL_ENCODED)) {
```
snazy, line 94: Having two different representations for a zero-char is IMO not good.
The content-keys `a.f\u0000o.bar` and `a.f%00o.bar` are technically different.
However, for this error, not sure whether we need it, because `%` should already be forbidden IIRC.

model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -96,7 +97,7 @@ public static Namespace of(String... elements) {
             String.format(
                 "Namespace '%s' must not contain a null element.", Arrays.toString(elements)));
       }
-      if (e.contains(ZERO_BYTE_STRING)) {
+      if (e.contains(ZERO_BYTE_STRING) || e.contains(ZERO_BYTE_STRING_URL_ENCODED)) {
```
snazy, line 100: (same comment as for ContentKey)

model/src/test/java/org/projectnessie/model/TestContentKey.java
```diff
@@ -260,6 +260,7 @@ void doubleByte() {
   @Test
   void blockZeroByteUsage() {
     assertThrows(IllegalArgumentException.class, () -> ContentKey.of("\u0000"));
+    assertThrows(IllegalArgumentException.class, () -> ContentKey.of("%00"));
```
snazy, line 263: (same comment as in ContenbKey) IIRC the `%` char should already be forbidden. Maybe deserves a separate test?

model/src/test/java/org/projectnessie/model/TestNamespace.java
```diff
@@ -85,19 +85,19 @@ public void testOneElement() {
...
     assertThat(namespace.getElements()).isEqualTo(elements);
     assertThat(namespace.toString()).isEqualTo(pathString);
     assertThat(namespace.toPathString()).isEqualTo(pathString);
     assertThat(namespace.name()).startsWith("a.b");
-    assertThat(namespace.name()).startsWith("a.b\u0000");
-    assertThat(namespace.name()).startsWith("a.b\u0000c");
-    assertThat(namespace.name()).startsWith("a.b\u0000c.namespa");
+    assertThat(namespace.name()).startsWith("a.b%00");
+    assertThat(namespace.name()).startsWith("a.b%00c");
+    assertThat(namespace.name()).startsWith("a.b%00c.namespa");
```
snazy, line 98: Side note: the assertThat...startsWith repetition is weird - the last one should be sufficient

model/src/test/java/org/projectnessie/model/TestNamespace.java
```diff
@@ -85,19 +85,19 @@ public void testOneElement() {
...
     Namespace namespace = Namespace.parse(pathString);
     assertThat(namespace.name()).isEqualTo(pathString);
     assertThat(namespace.getElements()).isEqualTo(elements);
     assertThat(namespace.toString()).isEqualTo(pathString);
     assertThat(namespace.toPathString()).isEqualTo(pathString);
     assertThat(namespace.name()).startsWith("a.b");
-    assertThat(namespace.name()).startsWith("a.b\u0000");
-    assertThat(namespace.name()).startsWith("a.b\u0000c");
-    assertThat(namespace.name()).startsWith("a.b\u0000c.namespa");
+    assertThat(namespace.name()).startsWith("a.b%00");
```
snazy, line 96: Not happy with replacing `\u0000` with `%00` - those are different strings. Better keep `\u0000` for `ContentKey` and keep URL encoding exclusively to the REST layer

snazy (changes requested):

model/src/main/java/org/projectnessie/model/Namespace.java
dimas-b, line 100: The second check seems unnecessary to me. `URLEncoder` should be able to handle it in `toPathString`.

servers/rest-services/src/test/java/org/projectnessie/services/rest/ContentKeyParamConverterProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.projectnessie.services.rest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import javax.ws.rs.ext.ParamConverter;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.model.ContentKey;
+
+public class ContentKeyParamConverterProviderTest {
```
dimas-b, line 24: `TestContentKeyParamConverterProvider`? (Test in front)

servers/rest-services/src/test/java/org/projectnessie/services/rest/ContentKeyParamConverterProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+      new ContentKeyParamConverterProvider().getConverter(ContentKey.class, null, null);
+
+  @Test
+  public void testNulls() {
+    assertThat(converter.fromString(null)).isNull();
+    assertThat(converter.toString(null)).isNull();
+  }
+
+  @Test
+  public void testValid() {
```
dimas-b, line 36: Why "invalid"? The key value is supported, right? Maybre rename to `testDots`?

servers/rest-services/src/test/java/org/projectnessie/services/rest/NamespaceParamConverterProviderTest.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.projectnessie.services.rest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import javax.ws.rs.ext.ParamConverter;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.model.Namespace;
+
+public class NamespaceParamConverterProviderTest {
```
dimas-b, line 24: `TestNamespaceParamConverterProvider`?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -45,7 +45,7 @@
 public abstract class AbstractRestNamespace extends AbstractRestRefLog {
 
   @ParameterizedTest
-  @ValueSource(strings = {"a.b.c", "a.b\u0000c.d", "a.b.c.d"})
+  @ValueSource(strings = {"a.b.c", "a.b%00c.d", "a.b.c.d"})
```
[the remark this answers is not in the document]
dimas-b: but `%00` is the canonical representation of dots inside key components, though.

I think we should have both the new `a.b%00c.d` and the old `a.b\u0000c.d` (for compatibility) test case here.

model/src/main/java/org/projectnessie/model/ContentKey.java
dimas-b: My reading of this PR is that now, `a.f%00o.bar` is the canonical representation. `a.f\u0000o.bar` is supported for backward compatibility.

model/src/test/java/org/projectnessie/model/TestNamespace.java
dimas-b: My understanding is that Jetty does not permit null chars in paths. So we cannot use the null char is parameters that come from URL paths. @nastra probably has more knowledge on this, though.

snazy: Right - but `ContentKey` gets constructed from a URL-encoded string (with`%00`).
So `%00` is the representation in URLs and `\u0000` is the representation in `ContentKey`.
IMO it should be sufficient to handle `%00` in `UriUtil` via `ContentKeyParamConverterProvider` & `NamespaceParamConverterProvider`

servers/rest-services/src/test/java/org/projectnessie/services/rest/ContentKeyParamConverterProviderTest.java
nastra: it says `valid` :)

model/src/main/java/org/projectnessie/model/UriUtil.java
nastra: we're still using JDK 8 unfortunately due to https://github.com/projectnessie/nessie/blob/b750fad777b0676622185f8acc5e3f040d9aaae9/pom.xml#L158

model/src/main/java/org/projectnessie/model/ContentKey.java
nastra: `%` isn't forbidden btw

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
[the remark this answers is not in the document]
nastra: I agree with @dimas-b here. We need one canonical representation imo. We could easily end up in a situation where somebody creates a `Namespace` instance using the URL encoded form and ending up a in situation where `a.b%00c.d` != `a.b\u0000c.d` just because the request didn't go through the param converters

model/src/main/java/org/projectnessie/model/ContentKey.java
nastra: see also my above comment about having a single canonical representation

model/src/test/java/org/projectnessie/model/TestNamespace.java
nastra: this specifically tests that zero bytes are correctly taken into account, since we're using `startsWith` logic in the namespace API

model/src/test/java/org/projectnessie/model/TestNamespace.java
nastra: see my comment above about having one canonical representation.

model/src/main/java/org/projectnessie/model/Namespace.java
nastra: without this check something like `Namespace.of("%00")` will not fail, but `Namespace.of("\u0000") will. This is because validation happens before url encoding

model/src/main/java/org/projectnessie/model/UriUtil.java
snazy: `StandardCharsets` exists since Java 1.7

servers/rest-services/src/test/java/org/projectnessie/services/rest/ContentKeyParamConverterProviderTest.java
dimas-b: Yeah, I see it not :)

model/src/main/java/org/projectnessie/model/ContentKey.java
dimas-b: We reuse the URLEncoder for escaping null chars, but the `toPathString` form is not quite the same as the actual URL form. For example, `a` > `f.o` > `bar` will have the `toPathString` form of `a.f%00o.bar` but in a URL it will be `a.f%2500o.bar`.

So, essentially, we will have double URL encoding. If this is too confusing, what do you think about using some non-ascii unicode char instead of `\u0000`, e.g. `\u203B` (reference mark) https://www.unicode.org/charts/PDF/U2000.pdf

model/src/main/java/org/projectnessie/model/Namespace.java
dimas-b: true, but I think it would be correct.

model/src/main/java/org/projectnessie/model/UriUtil.java
```diff
@@ -34,7 +38,8 @@ private UriUtil() {}
    */
   public static List<String> fromPathString(String encoded) {
     return Arrays.stream(encoded.split("\\."))
-        .map(x -> x.replace('\u0000', '.'))
+        .map(x -> x.replace(GROUP_SEPARATOR, DOT))
+        .map(x -> x.replace(ZERO_BYTE, DOT))
```
snazy, line 42: Nit: chain the two `replace`s (below as well)

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -45,7 +45,7 @@
 public abstract class AbstractRestNamespace extends AbstractRestRefLog {
 
   @ParameterizedTest
-  @ValueSource(strings = {"a.b.c", "a.b\u0000c.d", "a.b.c.d"})
+  @ValueSource(strings = {"a.b.c", "a.b\u001Dc.d", "a.b.c.d"})
```
snazy, line 48: Let's keep the `\u0000` case here

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -363,7 +363,7 @@ public void testNamespacesWithAndWithoutZeroBytes() throws BaseNessieClientServe
     assertThat(
             getApi()
                 .getMultipleNamespaces()
-                .namespace("a.b\u0000c")
+                .namespace("a.b\u001Dc")
```
snazy, line 366: Let's keep the 0-test case as well

model/src/test/java/org/projectnessie/model/TestContentKey.java
```diff
@@ -233,7 +233,13 @@ void namespaceNotIncluded() throws IOException {
   public void validation() {
```
snazy, line 233: Nit: make it a `@ParameterizedTest` with `@ValueSource(strings = {...})`

model/src/test/java/org/projectnessie/model/TestContentKey.java
```diff
@@ -263,6 +269,13 @@ void doubleByte() {
   @Test
   void blockZeroByteUsage() {
     assertThrows(IllegalArgumentException.class, () -> ContentKey.of("\u0000"));
```
snazy, line 271: Nit: make it a `@ParameterizedTest` with `@ValueSource(strings = {...})`

model/src/main/java/org/projectnessie/model/UriUtil.java
```diff
@@ -44,6 +49,9 @@ private UriUtil() {}
    * @return String encoded for path use.
    */
   public static String toPathString(List<String> elements) {
-    return elements.stream().map(x -> x.replace('.', '\u0000')).collect(Collectors.joining("."));
+    return elements.stream()
+        .map(x -> x.replace(DOT, GROUP_SEPARATOR))
+        .map(x -> x.replace(ZERO_BYTE, GROUP_SEPARATOR))
```
dimas-b, line 54: Why do we need to replace `ZERO_BYTE`? It's not allowed inside element data, is it?

snazy: Old clients could still send it

nastra: I guess defensive programming :)

snazy (approved):

model/src/main/java/org/projectnessie/model/UriUtil.java
dimas-b: true, but this is the encoding path, how can zero bytes make their way here?

PR #3866, opened 2022-04-06
Remove global state for Iceberg tables + views

snazy:
Before this change, the location to the Iceberg table/view metadata was maintained
globally, so all commits/references/branches/tags shared the same metadata.
It was originally introduced to ensure that there can be no duplicate IDs for
column-ID, schema-ID, et al. On the flip side, the global metadata "leaked" all
table snapshots and schemas to all branches and even experimental branches could
(potentially) exhaust all possible column IDs for a table.

So this change moves the Iceberg metadata-pointer back to "per commit".

### DIFF
--- a/site/docs/develop/kernel.md
+++ b/site/docs/develop/kernel.md
@@ -75,7 +75,7 @@ The data model for non-transactional key-value databases relies on a single _glo
 which is technically a table with a single row pointing to the _current_ entry in the _global-log_,
 _current_ entry in the _ref-log_ and the "HEAD"s of all named references (branches and tags).
 
-The _global-log_ contains the changes of the _global-state_.
+The _global-log_ contains changes to _global-state_, which is needed for backwards compatibility.
 
 The _ref-log_ contains the history with details of operations 
 like COMMIT, MERGE, TRANSPLANT, CREATE_REFERENCE, DELETE_REFERENCE, ASSIGN_REFERENCE.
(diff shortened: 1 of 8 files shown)


site/docs/develop/kernel.md
```diff
@@ -75,8 +75,7 @@ The data model for non-transactional key-value databases relies on a single _glo
 which is technically a table with a single row pointing to the _current_ entry in the _global-log_,
 _current_ entry in the _ref-log_ and the "HEAD"s of all named references (branches and tags).
 
-The _global-log_ contains the changes of the _global-state_, like the location of Iceberg's
-table metadata.
+The _global-log_ contains the changes of the _global-state_.
```
nastra, line 78: should we even mention global-log?

site/docs/develop/kernel.md
```diff
@@ -191,11 +190,11 @@ support atomic CAS (compare-and-swap) operations against a single row/record, bu
 conditional updates to multiple rows/records is either not supported at all or extremely slow.
 
 Nessie differentiates between content types that do require so called _global-state_ and those
-that do not. Apache Iceberg is currently the only content type that supports global state:
-the pointer to the Iceberg "Table Metadata" is tracked as "global state" and 
-the Iceberg snapshot ID, schema ID, partition spec ID, sort order ID
-is tracker per _Nessie named reference_. For _Nessie commits_, which are atomic, this means that
-Nessie has to update both the global-state and the on-reference-state for the Iceberg table. While
+that do not. _Global-state_ is maintained globally and evaluated when a convent value object is
```
nastra, line 193: typo: convent -> content?

nastra: same here, should we even mentiond global-state here? maybe just remove the entire paragraph?

site/docs/develop/spec.md
```diff
@@ -134,7 +133,7 @@ recorded within the [_Put Operation_](#put-operation) of a Nessie commit.
 The state of an SQL view is represented using the attributes
 `sqlText` and `dialect` (currently one of `HIVE`, `SPARK`, `DREMIO`, `PRESTO`).
 
-SQL views are tracked without a _Global State_ in Nessie, i.e. those three attributes are
+Views are tracked without a _Global State_ in Nessie, i.e. those three attributes are
```
nastra, line 136: let's just remove global state here

nastra (approved):

PR #3877, opened 2022-04-07
CI: Capture test reports only on failure

snazy:
Saves ~ 40 seconds for successful CI runs.

See checks in #3876

### DIFF
--- a/.github/actions/ci-results/action.yml
+++ b/.github/actions/ci-results/action.yml
@@ -2,15 +2,6 @@ name: 'Capture test results'
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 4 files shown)


.github/actions/ci-results/action.yml
```diff
@@ -1,16 +1,11 @@
 name: 'Capture test results'
+inputs:
+  include-reports:
```
dimas-b, line 3: Is this input used here?

snazy: Urgh - nope. Forgot to remove the input-decl

dimas-b (approved):

PR #3878, opened 2022-04-07
GC/Test: use correct store worker

snazy:


gc/gc-base/src/test/java/org/projectnessie/gc/base/TestJerseyRestGcInMemory.java
```diff
@@ -35,7 +36,8 @@
 @ExtendWith(DatabaseAdapterExtension.class)
 class TestJerseyRestGcInMemory extends AbstractRestGCRepoTest {
 
-  @NessieDbAdapter static DatabaseAdapter databaseAdapter;
```
ajantha-bhat, line 38: I have previously referred from `TestNessieJaxRsExtension`. So, we should handle there as well?

snazy: No

nastra (approved):

PR #3879, opened 2022-04-07
Speed up `ui/` build (locally)

snazy:
Using the [speed-measure-webpack-plugin](https://github.com/stephencookdev/speed-measure-webpack-plugin)
to figure out the webpack plugins and loaders that contribute to the build time.

Most expensive plugins/loaders:
* ESLint (17+ seconds)
* Terser (5+ seconds)
* babel-loader (7+ seconds)

ESLint + Terser plugins are required for production builds and cannot be disabled by default.
However, this change adds Maven properties to disable those plugins locally (see `CONTRIBUTING.md`).

Moving the cache for the babel-loader to a different directory that's not cleaned by `mvn clean`
speeds up rebuilds.

### DIFF
--- a/.github/actions/dev-tool-java/action.yml
+++ b/.github/actions/dev-tool-java/action.yml
@@ -32,7 +32,9 @@ runs:
     - name: Cache local NPM repository
       uses: actions/cache@v3
       with:
-        path: ~/.npm
+        path: | 
+          ~/.npm
+          .mvn/.babel-cache
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
           ${{ runner.os }}-node-
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,6 +51,29 @@ not necessary. Use one of these parameters to speed things up:
 * `-DskipTests` Compiles everything, runs no tests.
 * `-DskipITs` Compiles everything, runs unit tests, but no integration tests.
 
+If you do not (regularly) update the `ui/` module, you may want to use the properties
+`-Dui.disable-terser=true` and `-Dui.disable-es-lint=true` to speed up the build of the ui module by
+about 20 seconds.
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -51,6 +51,29 @@ not necessary. Use one of these parameters to speed things up:
...
+Hint: you can define a default set of properties in your `~/.m2/settings.xml` like this:
+```xml
+<settings>
+  <profiles>
+    <profile>
+      <id>nessie-props</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <properties>
```
nastra, line 67: should we have a `skipUi` profile that sets those? I know we have `quickly`, but something to skip the UI only seems to be also helpful

nastra (approved):
haven't tried it out locally, but LGTM if it works

.github/actions/dev-tool-java/action.yml
```diff
@@ -32,7 +32,9 @@ runs:
     - name: Cache local NPM repository
       uses: actions/cache@v3
       with:
-        path: ~/.npm
+        path: | 
+          ~/.npm
+          .mvn/.babel-cache
```
dimas-b, line 37: nit: do we really need the inner directory to have the dot prefix?

CONTRIBUTING.md
snazy: That becomes super tricky w/ maven profiles :(

dimas-b: skipping UI may invalidate the `quarkus:run` task... At least IIRC that happened to be after a `clean install -Dquicky`. I believe the quarkus run task complained about missing UI modules... but I did not retest recently ;)

dimas-b: I suppose we could still add `ui.disable-terser` and `ui.disable-es-lint` into a profile.

PR #3880, opened 2022-04-07
CI: Automatically approve + merge dependabot PRs

snazy:
GitHub workflow to automatically approve and merge dependabot-PRs that
have passed CI.

Only workflow runs that were triggered by dependabot cause automatic
approval + merging. Workflow rungs that were triggered by other users
do not cause automatic approval + merging.

PRs to bump flapdoodle/mongo are intentionally ignored (no version
after 3.1.4 works fine on macOS)!

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,83 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GitHub workflow to automatically approve and merge dependabot-PRs that
+# have passed CI.
+
+name: Dependabot Auto Merge
+
+on:
+  workflow_run:
+    types:
+      - "completed"
+    workflows:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,83 @@
...
+      github.actor == 'dependabot[bot]' && (
+        startsWith(github.event.workflow_run.head_commit.message, 'Bump')
+      ) && !(
+        startsWith(github.event.workflow_run.head_commit.message, 'Bump de.flapdoodle.embed.mongo')
+      )
+
+    steps:
+      - name: Check commit status
+        run: |
+          echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
```
XN137, line 52: according to `gh help environment`:
```
GH_TOKEN, GITHUB_TOKEN (in order of precedence): an authentication token for github.com
API requests. Setting this avoids being prompted to authenticate and takes precedence over
previously stored credentials.

(...)
```
setting an env var would be enough, we dont need to pipe it in with echo, which seems slightly less secure?

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,83 @@
...
+          gh api repos/snazy/nessie/commits/${{ github.event.workflow_run.head_commit.id }}/check-runs --jq 'if ([.check_runs[] | select(.name | endswith(" release") or startswith("codecov/") or startswith("Report ") | not ) | select(.conclusion != "skipped") | .conclusion // "pending" ] | unique == ["success"]) then "OK" else error("Commit checks are not OK") end'
```
XN137, line 53: is "snazy" intentional here?

.github/workflows/dependabot-auto-merge.yml
snazy: That's the only way I could get it to work that time.

dimas-b:
`Workflow rungs`?

PR #3885, opened 2022-04-07
Nessie GC implementation  

ajantha-bhat:
1. Changes in `gc-base` module:
- Remove commitProtection time configuration and use cutoff time itself to protect the commits.
- Modify IdentifiedResultsRepo contents as per the requirement of GC consumer algorithm. [like add commit hash, metadata location column, isExpired]
- Replace `ContentBloomFilter` with `BloomFilter` of `ContentFunnel`
- Bloom filter is now per task instead of per table in a task.
- Make `deadReferenceCutOffTimeStamp` a mandatory config (to avoid fallback to live config)
- Split `IdentifyContentsPerExecutor` into separate classes for live and expired computations.
- Refactor some code for better maintainability. 

2. Changes in`gc-iceberg` module:
- Introduce `gc-iceberg` module
- support stored procedure framework by introducing `NessieIcebergGcSparkCatalog` and `NessieIcebergGcSparkSessionCatalog`
- Introduce `IdentifyExpiredContentsProcedure` to call `gc-base` module identify step and write output to table.
- Introduce `ExpireContentsProcedure` to for gc consumer algorithm that implements own expire snapshots.


gc/gc-base/src/main/java/org/projectnessie/gc/base/GCCheckPointRepo.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.iceberg.types.Types;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.RowFactory;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.analysis.NoSuchTableException;
+import org.apache.spark.sql.types.StructType;
+
+/** DDL + DML functionality for the "GCCheckPoint" table. */
+public final class GCCheckPointRepo extends BaseRepo {
```
snazy, line 38: There were serious concerns mentioned _against_ having multiple tables not so long ago.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -284,27 +331,42 @@ private Iterator<Row> walkAllCommitsInReference(
...
               .flatMap(operations -> operations.stream().filter(Operation.Put.class::isInstance))
               .map(Operation.Put.class::cast)
               .map(Operation.Put::getContent)
               .filter(
                   content ->
                       (content.getType() == Content.Type.ICEBERG_TABLE
                           || content.getType() == Content.Type.ICEBERG_VIEW))
-              .filter(expiredContentPredicate)
+              .filter(validSnapshotPredicate)
               .iterator();
```
snazy, line 353: An Iterator is most likely wrong here. Read up about terminal operations of Streams.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCCheckPointRepo.java
snazy, line 38: I doubt this "check pointing" works in all cases - what happens wghen the configuration changes between GC runs??

gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
```diff
@@ -52,17 +52,37 @@ public DistributedIdentifyContents(SparkSession session, GCParams gcParams) {
    * @param references list of all the references (JSON serialized)
    * @param bloomFilterSize size of bloom filter to be used
    * @param droppedRefTimeMap map of dropped time for reference@hash (JSON serialized)
+   * @param runId current run id of the gc
+   * @param startedAt gc start time
    * @return map of {@link ContentBloomFilter} per content-id.
```
snazy, line 57: This is _extremely_ resource heavy. One BF per CID is definitely not going to work.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+      // get Nessie catalog
+      Catalog nessieCatalog = GCUtil.loadNessieCatalog(spark(), gcCatalogName, expiryBranchName);
+      String finalRunId = runId;
+      expiredSnapshotsPerContentId
+          .entrySet()
+          .forEach(
+              entry -> {
+                // Expiry Algorithm:
+                // A. For each content id, fetch ANY one of the content.
+                // B. Use the newly created expiry branch. On this branch,
```
snazy, line 182: Why does this need a (temporary) branch, actually Nessie, at all?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+                        processedMetadataFiles.addAll(files);
+                      }
+                    });
+
+                // create a dummy table metadata which has all the snapshots for a content from all
+                // the references.
+                TableMetadata base = ((BaseTable) table).operations().current();
+                base = TableMetadata.buildFrom(base).build();
+                try {
+                  FieldUtils.writeField(base, "previousFiles", Collections.emptyList(), true);
```
snazy, line 230: Strong -1 on relying on Iceberg implementation details + using error-prone hacks!

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+                        contentId,
+                        entry.getValue(),
+                        result.deletedDataFilesCount(),
+                        result.deletedManifestListsCount(),
+                        result.deletedManifestsCount(),
+                        filesToBeDeleted);
+                outputRows.add(outputRow);
+              });
+
+      dropExpireProcedureBranch(expiryBranchName, api);
```
snazy, line 257: Why is this cleanup not in a `finally`?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+                // TODO: Remove after https://github.com/apache/iceberg/pull/4509
+                updateMetadataLocation((BaseTable) table);
+
+                Set<Snapshot> snapshotSet =
+                    new TreeSet<>((o1, o2) -> (int) (o1.timestampMillis() - o2.timestampMillis()));
+                Set<String> processedMetadataFiles = new HashSet<>();
+                List<String> metadataLocations =
+                    new ArrayList<>(metadataLocationsPerContentId.get(contentId));
+
+                metadataLocations.sort(Collections.reverseOrder());
```
snazy, line 209: Why does this need to be sorted?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+                    expiryBranchName, identifiedResultsRepo, api, finalRunId, contentId);
+                Table table =
+                    nessieCatalog.loadTable(
+                        TableIdentifier.parse(EXPIRE_NAMESPACE + "." + contentId));
+
+                // TODO: Remove after https://github.com/apache/iceberg/pull/4509
+                updateMetadataLocation((BaseTable) table);
+
+                Set<Snapshot> snapshotSet =
+                    new TreeSet<>((o1, o2) -> (int) (o1.timestampMillis() - o2.timestampMillis()));
```
snazy, line 204: This will be wrong when two snapshots have the same timestamp.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+                        snapshotSet.addAll(metadata.snapshots());
+                        List<String> files =
+                            metadata.previousFiles().stream()
+                                .map(TableMetadata.MetadataLogEntry::file)
+                                .collect(Collectors.toList());
+                        processedMetadataFiles.addAll(files);
+                      }
+                    });
+
+                // create a dummy table metadata which has all the snapshots for a content from all
```
snazy, line 225: If two snapshots from different references have the same snapshot-ID, your approach will not work.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+    Map<String, List<String>> metadataLocationsPerContentId =
+        getMetadataLocationsPerContentId(runId, identifiedResultsRepo);
+
+    List<InternalRow> outputRows = new ArrayList<>();
+    try (NessieApiV1 api = GCUtil.getApi(nessieClientConfig)) {
+      createGCBranch(expiryBranchName, api);
+      // get Nessie catalog
+      Catalog nessieCatalog = GCUtil.loadNessieCatalog(spark(), gcCatalogName, expiryBranchName);
+      String finalRunId = runId;
+      expiredSnapshotsPerContentId
```
snazy, line 176: This will be very slow, because it can only expire one table at a time.

snazy (changes requested):
I doubt the proposed approach will work, added a bunch of comments.

There are also serious concerns about time and space complexity, which still stand for quite a while (see original gc-base PR). GC as it is implemented consumes way too much heap, can only let Iceberg expire one table at a time (too slow) and I'm quite unsure about whether the proposed checkpoint-mechanism works.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCCheckPointRepo.java
ajantha-bhat: OK. I can add a row type and keep these in `IdentifiedResultsRepo` itself instead of new table.

gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
ajantha-bhat: Agree, but If I keep global bloom filter (one for all or group of content id), false positive will be more.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: there are lot of validations in table metadata, it expects new snapshot should be younger than previous. So, this validation

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: same timestamp will be grouped together. I don't see a problem here.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: > If two snapshots from different references have the same snapshot-ID, your approach will not work.

The probability of this is same as two commits in Iceberg having a same snapshot id allocated. 

snapshot id is UUID: 

https://github.com/apache/iceberg/blob/master/core/src/main/java/org/apache/iceberg/TableOperations.java#L116-L120

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: Agree, I will make it multi threaded.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: Dummy metadata has to have same UUID as the content's any of the metadata. 
There is no other interface from Iceberg to do that other than using the base object. 

Also due to lot of internal validations in table metadata, I had to do this.
I don't find other way.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: > Why does this need a (temporary) branch.
To create a dummy metadata table.  It may confuse existing users if created in other existing reference. Also, I need to drop that table as it will have reference to all the snapshots. Which will give incorrect results if not dropped.

yeah, other catalog works fine as well. Let me check it.

ajantha-bhat:
> I doubt the proposed approach will work, added a bunch of comments.

> There are also serious concerns about time and space complexity, which still stand for quite a while (see original gc-base PR). GC as it is implemented consumes way too much heap, can only let Iceberg expire one table at a time (too slow) and I'm quite unsure about whether the proposed checkpoint-mechanism works.

@snazy : Thanks for the review. I have replied to most of it. 

Few points I will rework based on your suggestion (like keeping one output table, no need of expiry branch and use other catalog, make calling expire snapshots distributed or multi threaded).

Check point mechanism works, I have shared the document and PR has test cases. 

Regarding overall time complexity, I tested 1000 commits in one reference on a spark cluster. One reference it took around 4 seconds to indentify the expired snapshots (this is with inmemory nessie backend store with local quarkus server).

I don't see other ways of doing GC. If you have a better solution, I am really happy to discuss. 

Also, I think it is hard to review overall PR. Hence, I am splitting up these into sub-PRs this week. 
Each topic we can discuss in depth on the individual PRs.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
snazy: Iceberg does not allow duplicate snapshot IDs - and GC should be resilient against duplicate snapshot IDs in different branches.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
snazy: Contract of `Comparator`?
BTW: Subtraction here esp. in combination with a cast to `int` can go sideways.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
snazy: You can create TableMetadata from scratch, no?

ajantha-bhat: Nope. That interface uses random UUID inside and it won't match with the table's UUID. So, some validations will fail.

ajantha-bhat: Also `new TableMeatdata()` is not public constructor. It is package private

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
snazy: I don't understand why the order of metadataLocations matters.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
```diff
@@ -0,0 +1,435 @@
...
+    } else {
+      expiredFilesConsumer = filesToBeDeleted::add;
+    }
+    return expiredFilesConsumer;
+  }
+
+  private static void updateMetadataLocation(BaseTable table) {
+    try {
+      BaseMetastoreTableOperations operations = (BaseMetastoreTableOperations) table.operations();
+      FieldUtils.writeField(
```
snazy, line 324: Same strong -1 regarding reflection here, too

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCCheckPointRepo.java
ajantha-bhat: checkpoints are written during traversal itself, only marker of checkpoints are written after all the tables are expired.

As content id is not understandable to the user. Selective table expiry is not exposed in the procedure. expire_snapshot procedure will expire all the tables.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: oh.. This one is an optimization from me.

for content1 from all the branches, expired contents are...
```
idx - localtionX.
idy - locationY.
idx - localtionX.
.
.
idz - locationZ.

```
So, I want to start from the latest version of the metadata files. (hence, the sort)
because it contains .previousFiles(). I can skip those contents metadata file reading if those files are already processed.  The recent file have the most previousFiles().


P.S: 
I will add more comments. I didn't expected this code to be reviewed today. this PR was only to get the overall idea and approach.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: this is temporary, as mentioned in the method caller, I fixed this bug at Iceberg side. I can remove it once Iceberg makes a release.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: Sorry, it is a same code for getting snapshot Id as mentioned above. 
As it is UUID, Iceberg believes it as random. 

The same code will be used for branches also. So, it will be unique. 

How can two branches can have same snapshot id for two different commits as it uses the same code?  (ignoring schema change as it doesn't affect gc)

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: Used a `GcMetadataUtil` with same package name as iceberg to prepare a metadata without history to avoid using reflections.

ajantha-bhat:
Tested the expire_snapshots of a dummy metadata that doesn't have some partition spec in its metadata for the spec id 
mentioned in some manifest list. It goes for NPE as partition summaries are collected for ALL_MANIFEST table, it is not required to collect but Iceberg uses some common code :(

Also, If two commits allocated same snapshot id very very rare scenario),
In Iceberg (without branching), I raised a PR to fix it. 
But with Nessie, we cannot know as the ALTER TABLE operation produces same snapshot id but different metadata. 
At Nessie side, I cannot know whether it is duplicate snapshot id or alter table operations. 

@snazy, @keithgchapman : I cannot go ahead without a global state or sudo global state.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireSnapshotsProcedure.java
ajantha-bhat: Do we have to restrict Nessie user to configure some other catalog other than Nessie catalog? I find it odd. So, I used Nessie catalog. Any thoughts ?

ajantha-bhat:
I have reworked on this PR.
a) Implemented own distributive (spark job) expire contents logic [`ExpireContentsProcedure.java`] that invalidates lot of previous comments (like temporary branch, duplicate snapshot id and slow sequential table deletes]
b) single bloom filter is per reference now instead of per content per reference. 
c) changed bloom filter key to use snapshot id + metadata file location to handle backward compatibility for global state contents.
d) Moved checkpoint info into `IdentifiedResultsRepo` table, instead of one more separate table.

I will polish up a PR bit. But overall logic is ready for review. 

cc: @snazy , @keithgchapman

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
ajantha-bhat: `Iterable` bears the expectation to be able to provide an `Iterator` more than once, which a `Stream` can’t fulfill.
But I don't need `Iterable`, one time `Iterator` is enough here.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,293 @@
...
+  private static Column computeExpiredFilesUDF(
+      Column metadataLocation, Set<String> expiredManifests, FileIO fileIO) {
+    return functions
+        .udf(
+            new ComputeExpiredFilesUDF(fileIO, expiredManifests),
+            DataTypes.createArrayType(DataTypes.StringType))
+        .apply(metadataLocation);
+  }
+
+  private static String updateRunId(
```
snazy, line 279: Naming: this does not update anything

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,293 @@
...
+    Dataset<Row> expiredContents =
+        getExpiredContents(runId, identifiedResultsRepo, expiredManifests, fileIO);
+
+    expiredContents.persist();
+
+    if (!dryRun) {
+      expiredContents.foreach(
+          row -> {
+            List<String> files = row.getList(2);
+            files.forEach(fileIO::deleteFile);
```
snazy, line 144: How does the whole thing behave, if one or more random files cannot be deleted?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,293 @@
...
+        dataset
+            .withColumn("manifestLocations", functions.explode(dataset.col("manifestLocations")))
+            .select("manifestLocations", "isExpired");
+    // compute the globally expired manifests by doing left anti join with locally expired dataset.
+    Dataset<Row> expired =
+        dataset.filter("isExpired=true").dropDuplicates().select("manifestLocations");
+    Dataset<Row> live =
+        dataset.filter("isExpired=false").dropDuplicates().select("manifestLocations");
+    Dataset<Row> unreachableManifests = expired.except(live);
+    return unreachableManifests.select("manifestLocations").collectAsList().stream()
```
snazy, line 195: This statement collects twice, which is unnecessary.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,293 @@
...
+    dataset =
+        dataset
+            .withColumn("manifestLocations", functions.explode(dataset.col("manifestLocations")))
+            .select("manifestLocations", "isExpired");
+    // compute the globally expired manifests by doing left anti join with locally expired dataset.
+    Dataset<Row> expired =
+        dataset.filter("isExpired=true").dropDuplicates().select("manifestLocations");
+    Dataset<Row> live =
+        dataset.filter("isExpired=false").dropDuplicates().select("manifestLocations");
+    Dataset<Row> unreachableManifests = expired.except(live);
```
snazy, line 194: "unreachable" is probably the wrong term here

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeExpiredFilesUDF.java
```diff
@@ -0,0 +1,64 @@
...
+  private final FileIO io;
+  private final Set<String> expiredManifests;
+
+  public ComputeExpiredFilesUDF(FileIO io, Set<String> expiredManifests) {
+    this.io = io;
+    this.expiredManifests = expiredManifests;
+  }
+
+  @Override
+  public List<String> call(String metadataLocation) {
```
snazy, line 39: This function collects all data files from all manifests to delete those. Can't a data file be in an expired **and** in a live snapshot's manifest?

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentBloomFilter.java
```diff
@@ -54,29 +53,24 @@ public boolean mightContain(Content content) {
...
-
   private String getValue(Content content) {
+    // For the contents before global state removal (Nessie version < 0.26.0)
+    // metadataLocation will not be unique.
+    // Hence, to handle both the version's compatibility, use snapshotId + metadataLocation as key.
     switch (content.getType()) {
       case ICEBERG_TABLE:
-        return ICEBERG_TABLE.name() + ((IcebergTable) content).getSnapshotId();
+        IcebergTable icebergTable = (IcebergTable) content;
+        return icebergTable.getSnapshotId() + icebergTable.getMetadataLocation();
```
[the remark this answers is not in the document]
ajantha-bhat: As there is no global state. Same table in other branch may get same snapshot id. So, keeping `content-id + snapshot id` as key will make the content in other branch as live even though it is expired.
In above case tableMetadata will be unique. So, Using that as a key.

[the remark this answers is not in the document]
snazy: The purpose of Nessie GC is to mark snapshots as live / identify snapshots that are not live and then expire particular snapshots. The key for that is content-ID + snapshot-ID. I don't understand your point about "there is no global state".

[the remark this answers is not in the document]
ajantha-bhat: same table in different branches can have the same snapshot id allocated for different operations right? In that case, maybe one operation is live and the other is expired. So, we cannot rely on just content-ID + snapshot-ID to identify live or not (because we are merging the bloom filter of all the references into one)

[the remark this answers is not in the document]
snazy: RIght. okay.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeManifestsUDF.java
```diff
@@ -0,0 +1,61 @@
...
+  public List<String> call(String metadataLocation) {
+    TableMetadata metadata = TableMetadataParser.read(io, metadataLocation);
+    List<String> manifestFiles = new ArrayList<>();
+    metadata
+        .snapshots()
+        .forEach(
+            snapshot -> {
+              try {
+                manifestFiles.addAll(
+                    snapshot.allManifests().stream()
```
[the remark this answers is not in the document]
ajantha-bhat: The intention is to log the warning and move on instead of failing the whole expiry contents. 
I am catching only the `NotFoundException`, are you expecting to catch IO exceptions aswell ?

I will catch `UncheckedIOException` (RuntimeIOException) which can be thrown from Iceberg.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
ajantha-bhat: I will catch IO or Runtime exceptions and log the warning. I don't want to halt the whole expiry process.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeExpiredFilesUDF.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.apache.iceberg.GcMetadataUtil;
+import org.apache.iceberg.TableMetadata;
+import org.apache.iceberg.TableMetadataParser;
+import org.apache.iceberg.io.FileIO;
+import org.apache.spark.sql.api.java.UDF1;
+
+class ComputeExpiredFilesUDF implements UDF1<String, List<String>> {
```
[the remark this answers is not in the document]
ajantha-bhat: I know. But there is manifest pruning optimization to avoid data files of the live snapshots. 
I will try to optimize it. If not, I will come back with a clear explanation of why it is present twice.

ajantha-bhat:
@snazy 

Assume this is the identify gc result

cid1, metadata1, expired, branch1
cid1, metadata1, live   , branch2
cid1, metadata2, live   , branch1
cid1, metadata3, expierd, branch3


now for each content, I will read the table metadata -> read manifest lists -> collect all the manifests path -> explode -> separate live and expired rows -> expired.except(live) -> get expired manifests.


cid1, metadata1, {m1,m2,m3},    expired, branch1
cid1, metadata1, {m1,m2,m3},    live,    branch2
cid1, metadata2, {m1,m2,m3,m4}, live,    branch3
cid1, metadata3, {m1,m2},       expired, branch3

now for each EXPIRED content, I will read the table metadata -> read manifest lists -> collect all the manifests path -> PRUNE with expired manifests filter -> get GenericManifestFile -> read data files -> return the result.


In Iceberg expire_snaphots one manifest file is read once as it works on only one table metadata.
expire contents has to work with N tableMetadata (total number of commits/contents).
Hence, I don't want to read the data files of the manifests before finding globally expired manifests.

Hence the two time read. 

The another problem with Iceberg interface is to read data files, manifestPath is not enough. It needs GenericManifestFile object.
On previous read, I can't store GenericManifestFile object. (bulky object and seriaization fails due to avro schema in that object with null type). So, for expired contents, I am reading table metadata and manifestList twice.

I know it is slow. But I don't have other solution.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeExpiredFilesUDF.java
ajantha-bhat: data files are not shared between manifests. Only manifests are reused between snapshots.
So, already this udf taken an arguments of `expiredManifests`. It just reads those manifests to get the data files. 

 The another problem with Iceberg interface is to read data files, manifestPath is not enough. It needs GenericManifestFile object.
On previous read, I can't store GenericManifestFile object. (bulky object and seriaization fails due to avro schema in that object with null type). So, for expired contents, I am reading table metadata and manifestList twice.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeManifestsUDF.java
```diff
@@ -0,0 +1,61 @@
...
+                        .map(ManifestFile::path)
```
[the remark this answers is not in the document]
snazy: > As hasExistingFiles check is costly, 

Where??

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ComputeExpiredFilesUDF.java
ajantha-bhat: hmm, I stand corrected. 
There is one exception to my above statement. **rewrite_manifests** :(

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,269 @@
...
+        new IdentifiedResultsRepo(
+            spark(), gcCatalogName, gcOutputBranchName, gcOutputTableIdentifier);
+
+    if (runId == null) {
+      runId = getLatestCompletedRunID(gcOutputTableIdentifier, identifiedResultsRepo);
+    }
+
+    FileIO fileIO = getFileIO(nessieClientConfig);
+    Dataset<Row> expiredContents = getExpiredContents(runId, identifiedResultsRepo, fileIO);
+    expiredContents.persist();
```
snazy, line 139: Why this `.persist()`? Deserves a comment.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
```diff
@@ -115,14 +120,15 @@ public String identifyExpiredContents(SparkSession session) {
...
-      getOrCreateEmptyBranch(api, gcParams.getOutputBranchName());
+      GCUtil.getOrCreateEmptyBranch(api, gcParams.getOutputBranchName());
+      // Identify the live contents and return the bloom filter
+      liveContentsBloomFilter =
+          distributedIdentifyContents.getLiveContentsBloomFilter(
+              allRefs, bloomFilterSize, droppedReferenceTimeMap, runId, startedAt);
     }
     // Identify the expired contents
-    return distributedIdentifyContents.identifyExpiredContents(liveContentsBloomFilterMap, allRefs);
+    return distributedIdentifyContents.identifyExpiredContents(
```
snazy, line 130: This call will use the **closed** `NessieApiV1`!

gc/gc-iceberg/src/test/java/org/projectnessie/gc/iceberg/ITJerseyRestExpireProcedureInMemory.java
```diff
@@ -0,0 +1,57 @@
...
+import org.projectnessie.versioned.persist.inmem.InmemoryTestConnectionProviderSource;
+import org.projectnessie.versioned.persist.tests.extension.DatabaseAdapterExtension;
+import org.projectnessie.versioned.persist.tests.extension.NessieDbAdapter;
+import org.projectnessie.versioned.persist.tests.extension.NessieDbAdapterName;
+import org.projectnessie.versioned.persist.tests.extension.NessieExternalDatabase;
+
+@NessieDbAdapterName(InmemoryDatabaseAdapterFactory.NAME)
+@NessieExternalDatabase(InmemoryTestConnectionProviderSource.class)
+@ExtendWith(DatabaseAdapterExtension.class)
+class ITJerseyRestExpireProcedureInMemory extends AbstractExpireProcedure {
```
snazy, line 37: This one's called `ITJerseryRest...` and the other IT just `ITRest...` - should use one naming pattern.

gc/gc-iceberg/src/test/java/org/projectnessie/gc/iceberg/AbstractExpireProcedure.java
```diff
@@ -0,0 +1,473 @@
...
+import org.apache.spark.sql.types.StructType;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.projectnessie.error.NessieNotFoundException;
+import org.projectnessie.gc.base.AbstractRestGC;
+import org.projectnessie.gc.base.IdentifiedResultsRepo;
+import org.projectnessie.model.Branch;
+import org.projectnessie.model.ContentKey;
+
+public abstract class AbstractExpireProcedure extends AbstractRestGC {
```
snazy, line 53: The abstraction is unnecessary for just one 20 line concrete test class.

gc/gc-iceberg/src/test/java/org/projectnessie/gc/iceberg/AbstractIdentifyProcedure.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+import org.apache.spark.sql.catalyst.analysis.NoSuchProcedureException;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.projectnessie.gc.base.AbstractRestGCTest;
+import org.projectnessie.gc.base.IdentifiedResultsRepo;
+
+/** Tests all the cases from {@link AbstractRestGCTest} using stored procedure. */
+public abstract class AbstractIdentifyProcedure extends AbstractRestGCTest {
```
snazy, line 37: The abstraction is unnecessary for just one 20 line concrete test class.

snazy (changes requested):
I think, the whole implementation could be much simpler / easier to understand. There are a lot of unnecessary `.collect()`/collection-instantiations and one dangerous misuse external state.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCImpl.java
ajantha-bhat: sorry, I don't see where I am using `NessieApiV1` in driver after line:128. 

The executor will build its own `NessieApiV1`.  So, I still don't see the problem here.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -193,15 +226,7 @@ private void handleLiveCommit(
...
-                          content.getId(),
-                          k ->
-                              new ContentBloomFilter(
-                                  gcStateParamsPerTask.getBloomFilterSize(),
-                                  gcParams.getBloomFilterFpp()))
-                      .put(content);
+                  bloomFilter.put(((Operation.Put) operation).getContent());
                 }
               });
     }
```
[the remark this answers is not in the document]
ajantha-bhat: oh.. Are you referring to the scenario of where `perReference` cutoff time is configured. And that reference is dropped after sometime.  Also, the `deadReference` cutoff time is not configured.


In that case, Currently GC is using `perReference` cutoff time instead of default cutoff time.
Ok, we can make deadReference cutofftime as mandatory config as you suggested.

snazy:
Can you please rebase this PR?

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifiedResultsRepo.java
```diff
@@ -103,72 +118,159 @@ public StructType getSchema() {
...
+            COL_GC_RUN_START);
+
+    if (latestRunIdWithMarker.isEmpty()) {
+      return Collections.emptyMap();
+    }
+
+    String runId = latestRunIdWithMarker.collectAsList().get(0).getString(0);
+
+    // collect all the checkpoint for the run id
+    List<Row> rows =
```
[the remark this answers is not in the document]
ajantha-bhat: From the dataFrame, I didn't find a way to collect as map. 
Hence, need one iteration to convert to map.

ajantha-bhat:
@snazy : I have addressed the comments. Please have a look at it again. Thanks.

gc/gc-iceberg/src/main/java/org/apache/iceberg/GcMetadataUtil.java
```diff
@@ -0,0 +1,81 @@
...
+                StreamSupport.stream(reader.entries().spliterator(), false)
+                    .map(
+                        entry ->
+                            ExpireContentsProcedure.FileType.DATA_FILE.name()
+                                + "#"
+                                + entry.file().path().toString())
+                    .collect(Collectors.toList());
+            dataFiles.addAll(files);
+          } catch (IOException | UncheckedIOException e) {
+            LOG.warn("Failed to read the Manifests: {}", e.getMessage());
```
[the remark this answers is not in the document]
ajantha-bhat: Not explicitly adding the file name as the callstack will have that info in caused by.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    List<Row> rows = expiredContents.collectAsList();
+
+    List<InternalRow> outputRows = new ArrayList<>();
+    rows.forEach(
+        row ->
+            outputRows.add(
+                GCProcedureUtil.internalRow(
+                    row.getString(0), row.getString(1), row.getInt(3), row.getList(2))));
```
snazy, line 154: Please add comments what each index means here

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentFunnel.java
```diff
@@ -0,0 +1,62 @@
...
+package org.projectnessie.gc.base;
+
+import static org.projectnessie.model.Content.Type.ICEBERG_TABLE;
+import static org.projectnessie.model.Content.Type.ICEBERG_VIEW;
+
+import com.google.common.hash.Funnel;
+import com.google.common.hash.PrimitiveSink;
+import java.nio.charset.StandardCharsets;
+import org.projectnessie.model.Content;
+import org.projectnessie.model.IcebergTable;
```
snazy, line 25: All Iceberg related code should live in gc-iceberg - not gc-base

snazy (changes requested):
Not a full review, but probably enough comments for now. Most require changes.

Also: the handling of dropped references is broken, because it relies on the _name_ of a dropped references, which will not work, when the same reference name is dropped multiple times. See `org.projectnessie.gc.base.IdentifyExpiredContentsPerExecutor#getExpiredContentRows`

Beside that: `GCImpl.collectDeadReferences()` should use the functionality added by #4303

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentFunnel.java
```diff
@@ -0,0 +1,62 @@
...
+        into.putLong(icebergView.getVersionId())
+            .putString(icebergView.getMetadataLocation(), StandardCharsets.UTF_8);
+        break;
+      default:
+        throw new RuntimeException("Unsupported type " + content.getType());
+    }
+  }
+
+  @Override
+  public int hashCode() {
```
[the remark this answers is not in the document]
ajantha-bhat: while merging the bloom filters, 
It expects them to be of same funnel type (https://github.com/google/guava/blob/master/guava/src/com/google/common/hash/BloomFilter.java#L274). Hence, `equals()` is based on funnel type.
As, `equals()` is based on type, `hashcode()` is always returns same value for this type of funnel.

p.s: same logic is used in older nessie code also [https://github.com/projectnessie/nessie/blob/nessie-0.9.2/versioned/gc/spark/src/main/java/org/projectnessie/versioned/gc/BinaryBloomFilter.java#L183-L191]

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentFunnel.java
```diff
@@ -0,0 +1,62 @@
...
+import java.nio.charset.StandardCharsets;
+import org.projectnessie.model.Content;
+import org.projectnessie.model.IcebergTable;
+import org.projectnessie.model.IcebergView;
+
+/** Bloom filter funnel implementation to hold {@link Content} info. */
+public class ContentFunnel implements Funnel<Content> {
+
+  @Override
+  public void funnel(Content content, PrimitiveSink into) {
```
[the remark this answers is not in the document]
ajantha-bhat: As explained above `hashCode + equals` is needed. 

> this should be refactored 
I am not sure which part needs to be refactored and refactored into what?

gc/gc-base/src/main/java/org/projectnessie/gc/base/DistributedIdentifyContents.java
```diff
@@ -46,50 +47,83 @@ public DistributedIdentifyContents(SparkSession session, GCParams gcParams) {
...
-    IdentifyContentsPerExecutor executor = new IdentifyContentsPerExecutor(gcParams);
-    List<Map<String, ContentBloomFilter>> bloomFilterMaps =
+  public BloomFilter<Content> getLiveContentsBloomFilter(
+      List<String> references,
+      long bloomFilterSize,
+      Map<String, Instant> droppedRefTimeMap,
+      String runId,
+      Timestamp startedAt) {
+    IdentifyLiveContentsPerExecutor executor = new IdentifyLiveContentsPerExecutor(gcParams);
+    List<LiveContentsResult> liveContentsResults =
```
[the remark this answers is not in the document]
ajantha-bhat: Can't pass directly `JavaRDD` because I need to collect bloom filter and merge it along with preparing checkpoint rows. If it was just checkpoint rows, it would have been possible to pass `JavaRDD` to the `createDataFrame()`

So, I will refactor only other parts (removing double traversal and extra lists)

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyLiveContentsPerExecutor.java
```diff
@@ -0,0 +1,258 @@
...
+            .api(api)
+            .reference(ref)
+            .liveCommitPredicate(liveCommitPredicate)
+            .bloomFilterSize(bloomFilterSize)
+            .build();
+
+    return walkLiveCommitsInReference(gcStateParamsPerTask);
+  }
+
+  private LiveContentsResult walkLiveCommitsInReference(GCStateParamsPerTask gcStateParamsPerTask) {
```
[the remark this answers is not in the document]
ajantha-bhat: `GCStateParamsPerTask` is still needed for `handleLiveCommit()` as it has too many arguments. 

So, I feel no need to inline this function  (`walkLiveCommitsInReference`) because the caller will become too many lines of code and we don't get any extra benefits from inlining.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifiedResultsRepo.java
```diff
@@ -103,72 +118,165 @@ public StructType getSchema() {
...
+  }
+
+  /**
+   * Finds the latest completed checkpoint. Checkpoint is considered as completed when it has a
+   * checkpoint marker row (written when the expiry operation is successful).
+   *
+   * @return a map of referenceName and checkpoint (its last traversed live commit hash from the
+   *     head).
+   */
+  Map<String, String> collectLatestCommitCheckPoint() {
```
snazy, line 232: What happens to this checkpoint information, if this happens:

BEFORE:
* branch A @ commit 123
* branch B @ commit 456

THEN: GC runs - checkpoints written

THEN: User assigns B to 123 and A to 456

The checkpoint information will be wrong, no?

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifiedResultsRepo.java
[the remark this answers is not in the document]
snazy: `.groupBy()` ??

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyLiveContentsPerExecutor.java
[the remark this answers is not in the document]
snazy: Keeping `GCStateParamsPerTask` for the sake of 4 or so arguments feels odd.

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentFunnel.java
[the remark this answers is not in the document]
ajantha-bhat: ok. I changed it to an enum (singleton) like how `LongFunnel` and other built-in implementations are there.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyLiveContentsPerExecutor.java
```diff
@@ -0,0 +1,258 @@
...
+   * reached expired commits.
+   *
+   * @param foundAllLiveCommitHeadsBeforeCutoffTime condition to stop traversing
+   * @param commits stream of {@link LogResponse.LogEntry}
+   * @param commitHandler consumer of {@link LogResponse.LogEntry}
+   * @return last visited commit hash. It is the commit hash when all the live contents heads at the
+   *     cutoff time are found.
+   */
+  static String traverseLiveCommits(
+      MutableBoolean foundAllLiveCommitHeadsBeforeCutoffTime,
```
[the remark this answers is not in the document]
ajantha-bhat: Refactored a bit in terms of naming. I agree that loose coupling is not good. But I don't see other alternatives.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Override
+    public Row call(Row value) {
+      List<String> files = value.getList(2);
+      files.forEach(
+          file -> {
+            try {
+              fileIO.deleteFile(file);
+            } catch (UncheckedIOException e) {
+              LOG.warn(
```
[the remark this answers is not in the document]
ajantha-bhat: I manually deleted the file before the actual delete and observed that an exception is **not** thrown when the file doesn't exist (the delete function just returns false). So, Just changed the error message.

snazy:
If #4626 causes too much trouble now, I'm okay to do it as an immediate follow-up.

ajantha-bhat:
3 comments are pending for this PR.

1. Understand and adopt to #4303 instead of using reflog
2. improve checkpoint logic (probably need to introduce reference ID and not sure about handling the compatibility yet)
3. #4626 (`gc-base` module should not depend on iceberg jars)

I will explore in the same order I mentioned above.

ajantha-bhat:
> PR title and probably description should be adopted.

Updated.

ajantha-bhat:
@snazy: 
Out of 3 open points. 
I have handled reverting the checkpoint logic. 

I believe the other two [#4626 ,  #4690] can be handled in the follow-up.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,280 @@
...
+  private static Dataset<Row> getExpiredContents(
+      String runId, IdentifiedResultsRepo identifiedResultsRepo, FileIO fileIO) {
+    // Read the expired content-rows from output table for this run id
+    Dataset<Row> expiredContents = identifiedResultsRepo.collectExpiredContentsAsDataSet(runId);
+    Dataset<Row> expiredContentsDF = computeAllFiles(fileIO, expiredContents);
+    // Read the live content-rows from output table for this run id
+    Dataset<Row> liveContents = identifiedResultsRepo.collectLiveContentsAsDataSet(runId);
+    Dataset<Row> liveContentsDF = computeAllFiles(fileIO, liveContents);
+    // remove the files which are used by live contents
+    Dataset<Row> expiredFilesDF = expiredContentsDF.except(liveContentsDF);
```
snazy, line 177: Does this really ensure that a data file that is referenced in an expired snapshot's manifest _and_ in a live snapshot's manifest is _not_ deleted?

rewrite_manifests?

gc/gc-iceberg/src/main/java/org/apache/iceberg/GCMetadataUtil.java
```diff
@@ -0,0 +1,99 @@
...
+  public static List<String> computeAllFiles(Snapshot snapshot, FileIO io) {
+    // New ManifestList file created after compaction will have new manifest files
+    // pointing to older data files with a marking (status) that it is deleted data files.
+    // So, skip those datafiles.
+    Predicate<ManifestEntry<?>> activeDataFilesPredicate =
+        entry -> entry.status() != ManifestEntry.Status.DELETED;
+
+    try {
+      Stream<String> dataFileStream =
+          snapshot.allManifests().stream()
```
snazy, line 50: Does all this work with table-metadata v1 as well?

snazy (changes requested):
The handling of references in the code is still concerning. The code especially does _not_ handle the following case properly, as it will handle the reference as a _dropped_ reference cut-off time:
```
CREATE BRANCH foo @ COMMIT 123
DROP BRANCH foo
CREATE BRANCH foo @ COMMIT 123
```
Considering the reference name for dropped references feels incorrect. It seems that it should only consider the commit-ID of dropped references, not the name, not the ref type. 

Additionally, I'm not sure whether serializing all `Reference`s as JSON is correct, especially since references could be re-serialized on different JVMs. The order of the fields in the JSON string representation isn’t guaranteed AFAIK, which can lead to wrong GC behavior.

gc/gc-iceberg/src/main/java/org/apache/iceberg/GCMetadataUtil.java
ajantha-bhat: Yes, as mentioned in the spec, this field is mandatory for both v1 and v2. So, it will be present always.

image

snazy: The question was: "Does **all** this work..." - aka scenarios with v1 table metadata instead of v2!

ajantha-bhat: By default (current test cases) Iceberg uses the V1 format. So, test cases cover V1 already. 
I will also V2 test cases (for this I need to upgrade spark to 3.2 as spark3.2 started supporting merge on read deletes). I will do it now.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyLiveContentsPerExecutor.java
```diff
@@ -0,0 +1,241 @@
...
+
+  public IdentifyLiveContentsPerExecutor(GCParams gcParams) {
+    super(gcParams);
+  }
+
+  protected Function<String, LiveContentsResult> computeLiveContentsFunc(
+      long bloomFilterSize, Map<String, Instant> droppedRefTimeMap) {
+    return reference ->
+        computeLiveContents(
+            getCutoffTimeForRef(reference, droppedRefTimeMap),
```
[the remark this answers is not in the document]
ajantha-bhat: good catch. changed to Detached reference now.

ajantha-bhat:
@snazy : Addressed all the comments except JSON one.

> The order of the fields in the JSON string representation isn’t guaranteed AFAIK, which can lead to wrong GC behavior.

I am not clear about the "wrong GC behaviour". Do you mean deserialization will fail?
I tested once in the cluster when I introduced this JSON serialization and it was working.

Also tried the below properties for ObjectMapper and it didn't make any difference in the JSON string output order.
ObjectMapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true)
ObjectMapper.configure(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS, true)

ajantha-bhat:
@snazy: While I am working on these two open points [https://github.com/projectnessie/nessie/issues/4626 , https://github.com/projectnessie/nessie/issues/4690] on a follow-up PR, 

can we please merge this PR if it is ok?

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    try (FileIO fileIO = getFileIO(nessieClientConfig)) {
+      Dataset<Row> expiredContentsDF = getExpiredContents(runId, identifiedResultsRepo, fileIO);
+      if (!dryRun) {
+        expiredContentsDF =
+            expiredContentsDF.map(
+                new DeleteFunction(fileIO), RowEncoder.apply(expiredContentsDF.schema()));
+      }
+      List<Row> rows = expiredContentsDF.collectAsList();
```
snazy, line 158: This still looks unnecessary. Why not `return expiredContentsDF.map().collect()` with a `MapFunction` ??

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,292 @@
...
+    // Read the expired content-rows from output table for this run id
+    Dataset<Row> expiredContents = identifiedResultsRepo.collectExpiredContentsAsDataSet(runId);
+    Dataset<Row> expiredContentsDF = computeAllFiles(fileIO, expiredContents);
+    // Read the live content-rows from output table for this run id
+    Dataset<Row> liveContents = identifiedResultsRepo.collectLiveContentsAsDataSet(runId);
+    Dataset<Row> liveContentsDF = computeAllFiles(fileIO, liveContents);
+    // remove the files which are used by live contents
+    Dataset<Row> expiredFilesDF = expiredContentsDF.except(liveContentsDF);
+    // final output
+    // Example output row:
```
snazy, line 189: If I understand correctly, `getExpiredContents` returns rows with _collections_ of files to delete.
Those collections can become very very big (expiring many 1000s of files at once) - and result in too much heap pressure - resulting in out-of-memory-errors.
IIRC Ryan's left a comment about not using collections in `Row`s a while ago as well.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
```diff
@@ -0,0 +1,292 @@
...
+      }
+      List<Row> rows = expiredContentsDF.collectAsList();
+      return rows.stream()
+          .map(
+              row ->
+                  GCProcedureUtil.internalRow(
+                      // "content_id", "deleted_files_type", "deleted_files_count",
+                      // "deleted_files_list"
+                      row.getString(0), row.getString(1), row.getInt(3), row.getList(2)))
+          .toArray(InternalRow[]::new);
```
snazy, line 166: This will result in _all_ rows with _all_ expired files manifesting on the Java heap at once. See comment about heap pressure below. This one, together with the `.collectAsList()` above, however is worse, because it manifests _all_ expired files.

I'm not sure whether it's worth to return every deleted file to the user. I'd expect that giving users some hint about _how_ to access the list of deleted files should be enough. I.e. a table that users can access separately.

snazy (changes requested):
Sorry, but the implementation is still very prone to easily run into out-of-memory situations.

gc/gc-iceberg/build.gradle.kts
```diff
@@ -0,0 +1,98 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+  `java-library`
+  jacoco
+  `maven-publish`
```
snazy, line 20: Also needs the `signing` plugin (mandatory)

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentWithCommitInfo.java
```diff
@@ -15,24 +15,18 @@
...
- */
+/** Intermediate object to hold content and some commit info of the content. */
 @Value.Immutable
-interface GCStateParamsPerTask {
-
-  NessieApiV1 getApi();
+public interface ContentWithCommitInfo extends Serializable {
 
-  Reference getReference();
+  Content getContent();
```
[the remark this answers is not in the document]
ajantha-bhat: `ContentId, SnapshotId, MetadataLocation, Type` -- these 4 fields are needed. I think it is better to provide a shallow copy that copies the content object without table-metadata object and use it here. Will check it in a follow up.

gc/gc-iceberg/src/main/java/org/projectnessie/gc/iceberg/ExpireContentsProcedure.java
ajantha-bhat: ok. I will write the output to table.

ajantha-bhat:
> Sorry, but the implementation is still very prone to easily run into out-of-memory situations.

@snazy : Fixed it by writing the expiry results to a table as suggested. Instead of writing to the same identify output table, I am using a new table under the same reference, as the identify result should not have expiry procedure output and the schema is totally different.

gc/gc-iceberg/build.gradle.kts
ajantha-bhat: done

gc/gc-base/src/main/java/org/projectnessie/gc/base/ContentWithCommitInfo.java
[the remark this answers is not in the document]
ajantha-bhat: BloomFilter needs `Content` object as it is ContentFunnel. So, the content Object is retained to check against bloom filter. 

So, I think the best option is to have a fetch option that will exclude content attachments while bringing contents from the backend.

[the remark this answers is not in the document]
ajantha-bhat: Also observed that even though we store attachments, currently it is not fetched at this place of code. So, content is still a lightweight object. Need to check why it is not fetching the content attachment even though the fetch option is ALL.

snazy:
Thanks for the effort, but closing in favor of #4991

PR #3894, opened 2022-04-08
Compare Namespaces based on their canonical elements

nastra:
Previously the `NamespaceApi` would only compare two namespaces using
their string representation. With the introduction of the group
separator character we need to actually check each canonical namespace
element individually.

For example, one would expect both of these checks
to produce the same result (because we treat \u001D and \u0000 equally):

```
assertThat(namespace.name()).startsWith("a.b\u001Dc");
assertThat(namespace.name()).startsWith("a.b\u0000c");
```

Thus we're introducing a `Namespace.isSameOrSubElementOf(Namespace)` method that
allows such a comparison.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/Namespace.java
+++ b/model/src/main/java/org/projectnessie/model/Namespace.java
@@ -151,6 +151,29 @@ public abstract class Namespace extends Content {
   }
 
   /**
+   * Checks whether the current {@link Namespace} is the same as or a sub-element of the given
+   * {@link Namespace} instance by comparing each canonical element.
+   *
+   * @param parent {@link Namespace} instance to compare with.
+   * @return <code>true</code> if the current {@link Namespace} is the same as or a sub-element of
+   *     the given {@link Namespace} instance by comparing each canonical element, <code>false
+   *     </code> otherwise.
+   */
+  public boolean isSameOrSubElementOf(Namespace parent) {
+    Objects.requireNonNull(parent, "namespace must be non-null");
+    if (getElements().size() < parent.getElements().size()) {
+      return false;
+    }
+    for (int i = 0; i < parent.getElements().size(); i++) {
+      // elements must match exactly
+      if (!getElements().get(i).equals(parent.getElements().get(i))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  /**
    * Convert from path encoded string to normal string.
    *
    * @param encoded Path encoded string
--- a/model/src/test/java/org/projectnessie/model/TestNamespace.java
+++ b/model/src/test/java/org/projectnessie/model/TestNamespace.java
@@ -104,8 +104,46 @@ public class TestNamespace {
   }
 
   @Test
+  public void testIsSameOrSubElementOf() {
+    Namespace namespace = Namespace.of(asList("a", "b.c", "namespace"));
+
+    assertThatThrownBy(() -> Namespace.EMPTY.isSameOrSubElementOf(null))
+        .hasMessage("namespace must be non-null");
+
+    assertThat(Namespace.EMPTY.isSameOrSubElementOf(Namespace.EMPTY)).isTrue();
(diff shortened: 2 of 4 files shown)


model/src/test/java/org/projectnessie/model/TestNamespace.java
```diff
@@ -88,19 +88,55 @@ public void testOneElement() {
...
+    assertThat(Namespace.EMPTY.startsWith(Namespace.of("a"))).isFalse();
+  }
+
+  @Test
+  public void testDifferentZeroByteRepresentations() {
+    assertThat(Namespace.parse("a.b\u001Dc.d")).isEqualTo(Namespace.parse("a.b\u0000c.d"));
+    assertThat(Namespace.parse("a.b\u001Dc.d").startsWith(Namespace.parse("a.b\u0000c.d")))
+        .isTrue();
+
+    assertThat(Namespace.parse("a.b\u001Dc.d").startsWith(Namespace.parse("a.b\u0000c"))).isTrue();
```
snazy, line 134: Nit: add another "pair" that tests for `startsWith(Namespace.parse("a")`

snazy: Hm - don't see it :(

model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -150,6 +150,32 @@ public static Namespace parse(String identifier) {
...
+   */
+  public boolean startsWith(Namespace other) {
+    Objects.requireNonNull(other, "namespace must be non-null");
+    if (getElements().size() < other.getElements().size()) {
+      return false;
+    }
+    for (int i = 0; i < other.getElements().size(); i++) {
+      // check last element using a string prefix
+      if (i + 1 == other.getElements().size()) {
+        return getElements().get(i).startsWith(other.getElements().get(i));
```
dimas-b, line 169: I find this confusing. Since the method's arg type is a `Namespace`, I think the "starts with logic" should alwasy apply to full namespace compomponents.

Does `Namespace.of("a", "bcd")` actually start with `Namespace.of("a", "b")`? In the first namespace, there's no "directory"  `b` at the second level :thinking:

snazy: Deserves a test-case ;)

dimas-b, line 169: I find this confusing. Since the method's arg type is a `Namespace`, I think the "starts with logic" should always apply to full namespace components.

Does `Namespace.of("a", "bcd", "e")` actually start with `Namespace.of("a", "b")`? In the first namespace, there's no "directory"  `b` at the second level :thinking:

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -114,9 +114,7 @@ public void deleteNamespace(NamespaceParams params)
             try (Stream<KeyEntry<Type>> keys = getStore().getKeys(branch)) {
               if (keys.anyMatch(
                   k ->
-                      Namespace.of(k.getKey().getElements())
-                              .name()
-                              .startsWith(params.getNamespace().name())
+                      Namespace.of(k.getKey().getElements()).startsWith(params.getNamespace())
```
dimas-b, line 117: Currently deleting a `Namespace.of("a", "b")` will delete `Key.of("a", "bc", "d")`... This does not look correct to me.

dimas-b (changes requested):

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
nastra: in hindsight I agree with your observation. I think we actually need to do equality comparisons for each individual element

snazy (approved):

PR #3947, opened 2022-04-13
Add `metadata` to `IcebergTable` + `IcebergView`

snazy:
This change only adds the fields to the model objects to include the Iceberg
metadata. No functionality added or changed.

The new field defaults to `null` and is not serialized if it is `null`.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/IcebergTable.java
+++ b/model/src/main/java/org/projectnessie/model/IcebergTable.java
@@ -15,9 +15,12 @@
  */
 package org.projectnessie.model;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
@@ -60,12 +63,16 @@ public abstract class IcebergTable extends Content {
   @NotBlank
   public abstract String getMetadataLocation();
 
+  /** Corresponds to Iceberg's {@code currentSnapshotId}. */
   public abstract long getSnapshotId();
 
+  /** Corresponds to Iceberg's {@code currentSchemaId}. */
   public abstract int getSchemaId();
 
+  /** Corresponds to Iceberg's {@code defaultSpecId}. */
   public abstract int getSpecId();
 
+  /** Corresponds to Iceberg's {@code defaultSortOrderId}. */
   public abstract int getSortOrderId();
 
   @Override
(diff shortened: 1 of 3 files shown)


snazy:
@nastra @dimas-b the goal here is to just have the `metadata` field. Iceberg should then:
* send the metadata as JSON for a put operation
* if a get-contents returns a non-null (and non-`NullNode`) `message` object, it should use that one instead of reading it via `FileIO` (which would never happen with only this PR though)

WDYT?

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -60,22 +71,39 @@ public abstract class IcebergTable extends Content {
...
 
   @Override
   public Type getType() {
     return Type.ICEBERG_TABLE;
   }
 
+  @Schema(type = SchemaType.OBJECT)
+  @Nullable
+  @JsonInclude(Include.NON_NULL)
+  public abstract JsonNode getMetadata();
```
dimas-b, line 98: nit: maybe add a separate Iceberg version to set a context for interpreting this free-form metadata?

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -52,6 +57,14 @@
 @JsonTypeName("ICEBERG_TABLE")
 public abstract class IcebergTable extends Content {
 
+  public static final int INITIAL_SPEC_ID = 0;
```
nastra, line 60: do we need those fields?

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -60,22 +73,38 @@ public abstract class IcebergTable extends Content {
...
+  /** Corresponds to Iceberg's {@code currentSnapshotId}. */
+  @Min(DEFAULT_SNAPSHOT_ID)
   public abstract long getSnapshotId();
 
+  /** Corresponds to Iceberg's {@code currentSchemaId}. */
+  @Min(INITIAL_SCHEMA_ID)
   public abstract int getSchemaId();
 
+  /** Corresponds to Iceberg's {@code defaultSpecId}. */
+  @Min(INITIAL_SPEC_ID)
```
nastra, line 85: `@PositiveOrZero`?

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -60,22 +73,38 @@ public abstract class IcebergTable extends Content {
   @NotBlank
   public abstract String getMetadataLocation();
 
+  /** Corresponds to Iceberg's {@code currentSnapshotId}. */
+  @Min(DEFAULT_SNAPSHOT_ID)
```
nastra, line 77: `@Min(-1)` maybe?

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: removed

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: removed

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: removed

snazy:
Removed the added constraints to prevent (unlikely) breaking changes.

model/src/main/java/org/projectnessie/model/IcebergTable.java
snazy: Moved the `JsonNode` into a new `GenericMetadata` composite, which has a "variant".

nastra (approved):

PR #3960, opened 2022-04-14
Don't throw exception in requiresGlobalState

nastra:
It's possible that an old storage format hits this code and fails to
upgrade if it uses and old storage format type, thus it's better to
return false rather than throwing an exception


snazy (approved):

dimas-b:
Could we add an upgrade test for this?

PR #4013, opened 2022-04-21
Fix typo for RocksDB version used in compatibility tests

snazy:


dimas-b (dismissed):

snazy:
Duh - the test failure is rather new - and I was already checking that one.
Weird that the test failure happened for the typo-fix as well. But the test failure's unrelated to this change.

PR #4027, opened 2022-04-22
Generate content ID on the server side & namespace content-IDs

### LINKED ISSUE #3731: Consider adding unique identifiers for namespaces
In order to be able to set privileges (authorization) on explicit namespaces differently in different branches/tags, namespaces would need a unique identifier the same way that tables/views have content IDs.

### LINKED ISSUE #4017: Add unique contentID to namespaces
**Currently a namespace's ID is derived from its full path from root, this means that two folders with the same path on 2 different branches would have the same contentID. This can cause issues when trying to identify a folder uniquely. It would be good for all first class content types in Nessie to have a contentID. Namespaces should be a first class content type in Nessie.**:

snazy:
This change is driven by the demand to have "unique" content-IDs for
namespaces. Technically there are two kinds of namespaces:
* explicitly created namespaces (via `NamespaceApi.createNamespace`)
  and
* implicitly "created" namespaces, which are returned by
  `NamespaceApi.getNamespaces`, if no explicitly created namespace
  exists.

The previous approach had a deterministic content ID, because
content ID == namespace-path. "Unique" (aka `UUID.randomUUID()`)
are not deterministic.

To be able to distinguish between implicitly and explicitly created
namespaces (think: retrieving an implicitly created one, adding
properties and persisting it vs. the same with an explicitly created
namespace), the content ID must be `null` for implicitly created
namespaces and non-`null` for explicitly created ones.

This mandates the change to make `Content.getId()` `@Nullable` and
remove the "implicit" creation of a content ID, if not present.

Also implicitly changes the technical behavior that content-IDs are
now _only_ created on the server side. Previously, the content-ID
was generated for clients using the Java `:nessie-model` classes.

A content object to be created must have no (== `null`) content ID
and the expected-content object must be `null` as well.
A content object to be updated must have a non-`null` content ID,
and the expected-content object must be non-`null` as well, with
the same content-ID.

Fixes #3731
Fixes #4017

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
@@ -186,27 +186,42 @@ public abstract class AbstractRestContents extends AbstractRestCommitLog {
   @Test
   public void multiget() throws BaseNessieClientServerException {
     Branch branch = createBranch("foo");
-    ContentKey a = ContentKey.of("a");
-    ContentKey b = ContentKey.of("b");
-    IcebergTable ta = IcebergTable.of("path1", 42, 42, 42, 42);
-    IcebergTable tb = IcebergTable.of("path2", 42, 42, 42, 42);
+    ContentKey keyA = ContentKey.of("a");
+    ContentKey keyB = ContentKey.of("b");
+    IcebergTable tableA = IcebergTable.of("path1", 42, 42, 42, 42);
+    IcebergTable tableB = IcebergTable.of("path2", 42, 42, 42, 42);
     getApi()
         .commitMultipleOperations()
         .branch(branch)
-        .operation(Put.of(a, ta))
+        .operation(Put.of(keyA, tableA))
         .commitMeta(CommitMeta.fromMessage("commit 1"))
         .commit();
     getApi()
         .commitMultipleOperations()
         .branch(branch)
-        .operation(Put.of(b, tb))
+        .operation(Put.of(keyB, tableB))
         .commitMeta(CommitMeta.fromMessage("commit 2"))
         .commit();
     Map<ContentKey, Content> response =
-        getApi().getContent().key(a).key(b).key(ContentKey.of("noexist")).refName("foo").get();
+        getApi()
+            .getContent()
+            .key(keyA)
+            .key(keyB)
+            .key(ContentKey.of("noexist"))
+            .refName("foo")
+            .get();
     assertThat(response)
-        .containsEntry(a, ta)
-        .containsEntry(b, tb)
+        .containsKeys(keyA, keyB)
+        .hasEntrySatisfying(
+            keyA,
+            content ->
+                assertThat(content)
+                    .isEqualTo(IcebergTable.builder().from(tableA).id(content.getId()).build()))
+        .hasEntrySatisfying(
+            keyB,
+            content ->
+                assertThat(content)
+                    .isEqualTo(IcebergTable.builder().from(tableB).id(content.getId()).build()))
         .doesNotContainKey(ContentKey.of("noexist"));
   }
 }
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestMisc.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestMisc.java
@@ -17,7 +17,6 @@ package org.projectnessie.jaxrs;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import org.junit.jupiter.api.Test;
 import org.projectnessie.error.BaseNessieClientServerException;
@@ -42,17 +41,22 @@ public abstract class AbstractRestMisc extends AbstractRestMergeTransplant {
   public void checkSpecialCharacterRoundTrip() throws BaseNessieClientServerException {
     Branch branch = createBranch("specialchar");
     // ContentKey k = ContentKey.of("/%国","国.国");
-    ContentKey k = ContentKey.of("a.b", "c.txt");
-    IcebergTable ta = IcebergTable.of("path1", 42, 42, 42, 42);
+    ContentKey key = ContentKey.of("a.b", "c.txt");
+    IcebergTable table = IcebergTable.of("path1", 42, 42, 42, 42);
     getApi()
         .commitMultipleOperations()
         .branch(branch)
-        .operation(Put.of(k, ta))
+        .operation(Put.of(key, table))
         .commitMeta(CommitMeta.fromMessage("commit 1"))
         .commit();
 
-    assertThat(getApi().getContent().key(k).refName(branch.getName()).get()).containsEntry(k, ta);
-    assertEquals(ta, getApi().getContent().key(k).refName(branch.getName()).get().get(k));
+    assertThat(getApi().getContent().key(key).refName(branch.getName()).get())
+        .containsKey(key)
+        .hasEntrySatisfying(
+            key,
+            content ->
+                assertThat(content)
+                    .isEqualTo(IcebergTable.builder().from(table).id(content.getId()).build()));
   }
 
   @Test
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import org.assertj.core.api.iterable.ThrowingExtractor;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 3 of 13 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -91,13 +100,21 @@ public void testNamespaces(String namespaceName) throws BaseNessieClientServerEx
   @Test
   public void testNamespacesRetrieval() throws BaseNessieClientServerException {
     Branch branch = createBranch("namespace");
-    Namespace one = Namespace.parse("a.b.c");
-    Namespace two = Namespace.parse("a.b.d");
-    Namespace three = Namespace.parse("x.y.z");
-    Namespace four = Namespace.parse("one.two");
+
+    ThrowingExtractor<String, Namespace, ?> createNamespace =
```
nastra, line 104: why are these changes necessary?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -321,28 +338,37 @@ public void testNamespacesWithAndWithoutZeroBytes() throws BaseNessieClientServe
...
+                          .refName(branch.getName())
+                          .namespace(namespace)
+                          .create())
+              .isInstanceOf(NessieNamespaceAlreadyExistsException.class)
+              .hasMessage(String.format("Namespace '%s' already exists", namespace.name()));
+
+          return created;
+        };
+
+    Namespace first = creator.apply(firstName);
```
nastra, line 369: also not sure why these changes are required. The only necessary change it seems is the comparison of `Namespace::getElements, Namespace::toPathString`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractResteasyTest.java
```diff
@@ -233,6 +256,13 @@ public void testBasic() {
         .statusCode(204);
   }
 
+  private static Content withoutId(Content content) {
+    if (content instanceof IcebergTable) {
+      return IcebergTable.builder().from(content).id(null).build();
+    }
+    throw new IllegalArgumentException("" + content);
```
nastra, line 263: maybe add a better error message here?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -111,44 +112,67 @@ public Hash commit(
...
-                storeWorker.getPayload(op.getValue()),
-                storeWorker.toStoreOnReferenceState(op.getValue())));
+                storeWorker.getPayload(content),
+                storeWorker.toStoreOnReferenceState(content)));
 
-        if (op.getExpectedValue() != null) {
-          ContentId expectedContentId = ContentId.of(storeWorker.getId(op.getExpectedValue()));
+        if (expected != null) {
+          String expectedId = storeWorker.getId(expected);
+          if (expectedId == null) {
```
nastra, line 144: maybe just use `Preconditions.checkArgument()`?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -111,44 +112,67 @@ public Hash commit(
...
       if (operation instanceof Put) {
         Put<CONTENT> op = (Put<CONTENT>) operation;
-        ContentId contentId = ContentId.of(storeWorker.getId(op.getValue()));
+        CONTENT content = op.getValue();
+        CONTENT expected = op.getExpectedValue();
+
+        if (storeWorker.getId(content) == null) {
+          // No content-ID --> New content
+
+          if (expected != null) {
```
nastra, line 121: maybe use `Preconditions.checkArgument()` here?

nastra (changes requested):

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
snazy: We don't have Guava in this module yet - and all other places in this module do it like this.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
snazy: to have the content-ID which is part of the .equals() comparison below

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
snazy: Also to not mess with all the assertions below - to have the content-ID

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestMisc.java
```diff
@@ -51,8 +50,10 @@ public void checkSpecialCharacterRoundTrip() throws BaseNessieClientServerExcept
         .commitMeta(CommitMeta.fromMessage("commit 1"))
         .commit();
 
-    assertThat(getApi().getContent().key(k).refName(branch.getName()).get()).containsEntry(k, ta);
-    assertEquals(ta, getApi().getContent().key(k).refName(branch.getName()).get().get(k));
+    assertThat(getApi().getContent().key(k).refName(branch.getName()).get())
+        .containsKey(k)
+        .hasEntrySatisfying(
+            k, c -> assertThat(c).isEqualTo(IcebergTable.builder().from(ta).id(c.getId()).build()));
```
dimas-b, line 56: nit: rename `k` to about confusion with `k` in the outer scope.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestNamespace.java
```diff
@@ -55,8 +56,16 @@ public void testNamespaces(String namespaceName) throws BaseNessieClientServerEx
...
-    assertThat(namespace).isNotNull().isEqualTo(ns);
-    assertThat(getApi().getNamespace().refName(branch.getName()).namespace(ns).get())
+    assertThat(namespace)
+        .isNotNull()
+        .extracting(Namespace::getElements, Namespace::toPathString)
+        .containsExactly(ns.getElements(), ns.toPathString());
+
+    Namespace got = getApi().getNamespace().refName(branch.getName()).namespace(ns).get();
+    assertThat(got)
+        .describedAs(
```
dimas-b, line 66: nit: do we need a custom description here? I suppose the default `Namespace.toString` would be clear in terms of mismatching fields... WDYT?

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -94,9 +95,13 @@ public Namespace createNamespace(NamespaceParams params, Namespace namespace)
       Preconditions.checkArgument(!namespace.isEmpty(), "Namespace name must not be empty");
 
       Put put = Put.of(ContentKey.of(namespace.getElements()), namespace);
-      commit(branch, "create namespace " + namespace.name(), TreeApiImpl.toOp(put), validator);
+      Hash hash =
+          commit(branch, "create namespace " + namespace.name(), TreeApiImpl.toOp(put), validator);
 
-      return namespace;
+      Namespace namespaceWithId =
+          (Namespace) getStore().getValue(hash, Key.of(namespace.getElements()));
```
dimas-b, line 102: Could we perhaps generify `getValue` to be `<T extends Content> T getValue(...)` to avoid trivial casts?

servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java
```diff
@@ -192,6 +194,23 @@ private static IllegalArgumentException noIcebergMetadataPointer() {
         "Iceberg content from reference must have global state, but has none");
   }
 
+  @Override
+  public Content applyId(Content content, String id) {
+    Objects.requireNonNull(content, "content must not be null");
+    Objects.requireNonNull(id, "id must not be null");
```
dimas-b, line 200: Should we also require that `content.getId` _is_ `null`?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
```diff
@@ -205,8 +205,11 @@ public void multiget() throws BaseNessieClientServerException {
     Map<ContentKey, Content> response =
         getApi().getContent().key(a).key(b).key(ContentKey.of("noexist")).refName("foo").get();
     assertThat(response)
-        .containsEntry(a, ta)
-        .containsEntry(b, tb)
+        .containsKeys(a, b)
+        .hasEntrySatisfying(
+            a, c -> assertThat(c).isEqualTo(IcebergTable.builder().from(ta).id(c.getId()).build()))
```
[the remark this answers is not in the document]
snazy: hm - still the same `a` though

renamed the vars anyway

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestMisc.java
snazy: (same)

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
snazy: It's only a special case for Namespace. Added a check.

dimas-b (dismissed):

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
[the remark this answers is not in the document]
dimas-b: Ah, sorry, I missed that `a` was an argument to `hasEntrySatisfying`... I thought it was a lambda parameter :facepalm:

nastra (changes requested):
mostly LGTM, just some small changes

nastra (approved):

PR #4031, opened 2022-04-22
Fix `BadImport` error-prone warnings

snazy:
```
[BadImport] Importing nested classes/static methods/static fields with commonly-used names can make code harder to read, because it may not be clear from the context exactly which type is being referred to. Qualifying the name with that of the containing class can make the code clearer. Here we recommend using qualified class: RefPointer.
```


XN137 (approved):
not sure how necessary this change is but as is the changes look "correct".

if we deeply care about this warning (or others as in the other PRs) we should consider turning them into errorprone errors instead (to enforce fixing them).

snazy:
Yep, opened #4053

PR #4034, opened 2022-04-22
errorprone: use of IdentityHashMap

snazy:


XN137 (approved):
nit: maybe add "explicit use" in commit message... i had to look up what the rule violation was (or mention the rule)

PR #4036, opened 2022-04-22
errorprone: java time / local zone

snazy:


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -115,6 +115,7 @@ public void execute() throws BaseNessieClientServerException {
 
     ThreadLocalRandom random = ThreadLocalRandom.current();
 
+    @SuppressWarnings("JavaTimeDefaultTimeZone")
     String runStartTime =
         DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss").format(LocalDateTime.now());
```
XN137, line 120: should we prefer `LocalDateTime.now(ZoneOffset.UTC);` over the suppression?

XN137 (approved):
since server time could be different than client time, i would favor using UTC but if we want to keep client time this looks correct i guess.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
snazy: Maybe not. It's a "user" visitble thing

PR #4039, opened 2022-04-22
errorprone: `protected` on final class

snazy:


tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/AbstractContentGeneratorTest.java
```diff
@@ -75,20 +75,20 @@ protected static final class ProcessResult {
...
       this.stdOut = stdOut;
     }
 
-    protected int getExitCode() {
+    int getExitCode() {
       return exitCode;
     }
 
-    protected String getStdOut() {
+    private String getStdOut() {
```
XN137, line 87: not sure if this should be private, since its part of the "interface"... it just happens to not be called atm

XN137 (approved):

PR #4047, opened 2022-04-22
errorprone: `Exception` constant

snazy:


XN137 (approved):
i feel dirty but given the use its correct i guess

PR #4048, opened 2022-04-22
errorprone: (false positive) `String.repeat` (Java 11)

snazy:


versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractTracing.java
```diff
@@ -141,7 +141,7 @@ public boolean equals(Object o) {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
```
nastra, line 144: kind of annoying that errorprone complains about generate equals() methods

nastra (approved):

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractTracing.java
snazy: yea :(

PR #4049, opened 2022-04-22
errorprone: (false positive) "class can be static" (not here)

snazy:


XN137 (approved):
is it because of the junit nested annotation that they cant be static?

snazy:
> is it because of the junit nested annotation that they cant be static?

yes

PR #4050, opened 2022-04-22
errorprone: generic for return type only

snazy:


versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
```diff
@@ -377,7 +377,7 @@ static DatabaseAdapter createAdapterResource(
     return applyCustomConfig;
   }
 
-  @SuppressWarnings("unchecked")
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
```
XN137, line 380: nit: maybe move the unchecked annotation to the cast statement, like in the other PR

XN137 (approved):

PR #4068, opened 2022-04-25
Explicitly define scala versions for scala-maven-plugin

nastra:


XN137 (commented):
whats our current understanding of why we cant set this default globally?

if only the gatling module is using `scala2.13.version` shouldnt it be the only module needing to override it?

nastra:
the problem is that the override is not working for the gatling module :(

PR #4073, opened 2022-04-25
errorprone: [DefaultCharset]

snazy:
Implicit use of the platform default charset, which can result in differing behaviour between JVM executions or incorrect behavior if the encoding of the data source doesn't match expectations.

### DIFF
--- a/clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
+++ b/clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
@@ -27,6 +27,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
```diff
@@ -77,9 +78,9 @@ void testGet() throws Exception {
         h -> {
           Assertions.assertEquals("GET", h.getRequestMethod());
           String response = MAPPER.writeValueAsString(inputBean);
-          h.sendResponseHeaders(200, response.getBytes().length);
+          h.sendResponseHeaders(200, response.getBytes(StandardCharsets.UTF_8).length);
           OutputStream os = h.getResponseBody();
```
XN137, line 82: nit: seems like all of these should use try-resources block

clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
```diff
@@ -417,7 +418,7 @@ public boolean equals(Object o) {
       if (this == o) {
         return true;
       }
-      if (o == null || getClass() != o.getClass()) {
+      if (!(o instanceof ExampleBean)) {
```
XN137, line 421: imo we should try to fix all violations of a single errorprone check in a single PR (and also potentially turn the check into an error)

XN137 (approved):

clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
snazy: Feel free to open a PR to fix it

PR #4094, opened 2022-04-26
Remove `fail_ci_if_error: true` for codecov

snazy:
... because that let's GH Workflows fail on Nessie forks.


XN137:
if code coverage upload fails, maybe the whole workflow step just should not run on forks?
that seems better than silently failing on forks (and possible failing again on the main repo) WDYT?

snazy:
See #4095

PR #4127, opened 2022-04-28
Test/contents: Add a couple more API operations

snazy:
In `AbstractRestContents`, add separte assertions against the committed contents/operations using:
* getEntries() (as present before)
* getDiff()
* getContent()
* getCommitLog()


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
```diff
@@ -129,27 +127,76 @@ public static Stream<ContentAndOperationType> contentAndOperationTypes() {
...
-    List<Entry> expect =
-        contentAndOperationTypes()
-            .filter(c -> c.operation instanceof Put)
-            .map(c -> Entry.builder().type(c.type).name(c.operation.getKey()).build())
-            .collect(Collectors.toList());
-    assertThat(entries).containsExactlyInAnyOrderElementsOf(expect);
+    contentAndOps.forEach(contentAndOp -> commit.operation(contentAndOp.operation));
+    Branch committed = commit.commit();
+
+    assertAll(
```
nastra, line 141: do we really need to put this into an `assertAll`? The problem with this is that it's more difficult to debug imo when code is inside an `assertAll`, so I would prefer to not use `assertAll` here

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
```diff
@@ -163,24 +210,129 @@ public void verifyContentAndOperationTypesIndividually(
...
-    List<Entry> expect =
-        contentAndOperationType.operation instanceof Put
-            ? singletonList(
-                Entry.builder().name(fixedContentKey).type(contentAndOperationType.type).build())
-            : emptyList();
-    assertThat(entries).containsExactlyInAnyOrderElementsOf(expect);
+
+    if (contentAndOperationType.operation instanceof Put) {
+      Put put = (Put) contentAndOperationType.operation;
+      assertAll(
```
nastra, line 223: same here, also at some point this gets hard to read, so I would prefer to not use `assertAll`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
```diff
@@ -163,24 +210,129 @@ public void verifyContentAndOperationTypesIndividually(
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  private Content clearIdOnContent(Content content) {
+    try {
+      return (Content)
+          content
+              .getClass()
+              .getDeclaredMethod("withId", String.class)
```
nastra, line 331: maybe move `withId` to a utility class so that we don't have to use reflection. Also should this rather be `withoutId`?

nastra (changes requested):

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
snazy: The opposite is actually true - it's easier to debug, because you see which operations actually fail.
Does it matter where you set the breakpoint?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
snazy: `withId` is generated code in all Immutable* classes that extend `Content`. Don't like to have a publicly accessible functionality to update or clear the content ID.

nastra: imo it would be good enough to move such functionality into a test utility class

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestContents.java
nastra: fair enough, however the argument about readability still holds. What's the big benefit using `assertAll` here?

nastra (approved):

PR #4149, opened 2022-05-02
Helm: Add ingress support for Kubernetes >=1.22

kerossin:
Update the ingress Helm template to support newer Kubernetes versions.

networking.k8s.io/v1beta1 was removed in Kubernetes 1.22 and networking.k8s.io/v1 is available since 1.19.

Source: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122


kostassabulis (approved):

snazy (approved):
LGTM!
Thanks for the fix!

PR #4163, opened 2022-05-04
Use hashOnRef when fetching Namespaces

nastra:
Currently the `hashOnRef` query flag was not used when fetching a
Namespace at a particular hash. This diff makes sure that the
`hashOnRef` flag is properly used when fetching namespaces.
Note that we don't use `hashOnRef` when creating/updating/deleting a
Namespace and thus pass a `null` for the hash param to `namedRefWithHashOrThrow(...)`


servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -86,20 +88,24 @@ public Namespace createNamespace(NamespaceParams params, Namespace namespace)
...
       Hash hash =
-          commit(branch, "create namespace " + namespace.name(), TreeApiImpl.toOp(put), validator);
+          commit(
+              BranchName.of(refWithHash.getValue().getName()),
+              "create namespace " + namespace.name(),
+              TreeApiImpl.toOp(put),
+              validator);
 
-      Content content = getExplicitlyCreatedNamespace(namespace, hash).orElse(null);
+      WithHash<NamedRef> updated = namedRefWithHashOrThrow(params.getRefName(), hash.asString());
```
snazy, line 106: `WithHash.of(hash, params.getRefName());` is sufficient

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -262,11 +276,11 @@ public void updateProperties(NamespaceParams params, NamespaceUpdate namespaceUp
 
   private Stream<KeyEntry<Content.Type>> getNamespacesKeyStream(
       @Nullable Namespace namespace,
-      BranchName branch,
+      WithHash<NamedRef> refWithHash,
```
snazy, line 279: The function only uses the `Hash` - maybe better to refactor to `Hash hash` (the name's irrelevant)

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -289,15 +303,16 @@ private Namespace namespaceFromType(KeyEntry<Content.Type> withType) {
     return Namespace.of(elements);
   }
 
-  private Optional<Content> getExplicitlyCreatedNamespace(Namespace namespace, Ref ref)
-      throws ReferenceNotFoundException {
-    return Optional.ofNullable(getStore().getValue(ref, Key.of(namespace.getElements())));
+  private Optional<Content> getExplicitlyCreatedNamespace(
+      Namespace namespace, WithHash<NamedRef> refWithHash) throws ReferenceNotFoundException {
```
snazy, line 307: Here and below: The function only uses the `Hash` - maybe better to refactor to `Hash hash` (the name's irrelevant)

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -159,31 +171,32 @@ public Namespace getNamespace(NamespaceParams params)
...
    *
    * @param namespace The namespace to fetch
-   * @param branch The ref to use
+   * @param refWithHash The ref to use
    * @return A {@link Namespace} instance
    * @throws ReferenceNotFoundException If the ref could not be found
    * @throws NessieNamespaceNotFoundException If the namespace could not be found
    */
-  private Namespace getNamespace(Namespace namespace, BranchName branch)
+  private Namespace getNamespace(Namespace namespace, WithHash<NamedRef> refWithHash)
```
snazy, line 179: The function only uses the `Hash` - maybe better to refactor to `Hash hash` (the name's irrelevant)

snazy (approved):

PR #4165, opened 2022-05-04
Spark SQL: Configure ref.hash for NessieCatalog only when explicitly requested

nastra:
This makes sure that we only configure the `ref.hash` parameter for the
`NessieCatalog` only when the Spark SQL cmd actually specifies a hash.
Specifying a hash means that generally a user wants to read data at a
particular hash. It is however not possible to update data at a
particular hash, thus the following sequence of operations will lead to
an unexpected exception:
* USE REFERENCE branchName IN nessie
* CREATE TABLE ....

This will lead to the error shown in
https://github.com/apache/iceberg/blob/b5f367bdedc26e0f676d88462912b57ff7bae167/nessie/src/main/java/org/apache/iceberg/nessie/UpdateableReference.java#L76
because once the `NessieCatalog` is set to a particular hash, it is
never updated (which makes sense).

Note that this issue only happens with the fix described in
https://github.com/apache/iceberg/pull/4330, hence the current tests
that we have are all passing.

### DIFF
--- a/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BaseUseReferenceExec.scala
+++ b/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BaseUseReferenceExec.scala
@@ -34,7 +34,12 @@ abstract class BaseUseReferenceExec(
   ): Seq[InternalRow] = {
 
     val ref = NessieUtils.calculateRef(branch, timestampOrHash, api)
-    NessieUtils.setCurrentRefForSpark(currentCatalog, catalog, ref)
+    NessieUtils.setCurrentRefForSpark(
+      currentCatalog,
+      catalog,
+      ref,
+      timestampOrHash.isDefined
+    )
 
     Seq(
       InternalRow(
--- a/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
+++ b/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
@@ -194,18 +194,37 @@ object NessieUtils {
       .build(classOf[NessieApiV1])
   }
 
+  /** @param currentCatalog
+    *   The current Spark catalog
+    * @param catalog
+    *   The catalog to configure for Spark
+    * @param ref
+    *   The reference to configure for Spark
+    * @param configureRefAtHash
+    *   Whether to configure the ref at its given hash. Note that this should
+    *   only be done when reading data from the given ref at its particular
+    *   hash.
+    */
   def setCurrentRefForSpark(
       currentCatalog: CatalogPlugin,
       catalog: Option[String],
-      ref: Reference
+      ref: Reference,
+      configureRefAtHash: Boolean
   ): Unit = {
     val catalogName = catalog.getOrElse(currentCatalog.name)
     val catalogImpl =
       SparkSession.active.sessionState.catalogManager.catalog(catalogName)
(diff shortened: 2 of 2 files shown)


clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/BaseUseReferenceExec.scala
```diff
@@ -34,7 +34,12 @@ abstract class BaseUseReferenceExec(
   ): Seq[InternalRow] = {
 
     val ref = NessieUtils.calculateRef(branch, timestampOrHash, api)
-    NessieUtils.setCurrentRefForSpark(currentCatalog, catalog, ref)
+    NessieUtils.setCurrentRefForSpark(
+      currentCatalog,
+      catalog,
+      ref,
+      null != NessieUtils.hashFrom(timestampOrHash)
```
ajantha-bhat, line 41: `NessieUtils.hashFrom` is called two times for `timestampOrHash`

one is inside `calculateRef` and one is outside.

, call we compute it once and reuse for both ?

ajantha-bhat, line 41: If it is a `timestamp`, `NessieUtils.hashFrom` will return null, but we what we want to set is proper hash mapping to that timestamp.

nastra: I actually simplified it to just checking `timestampOrHash.isDefined`

clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
```diff
@@ -194,18 +194,37 @@ object NessieUtils {
...
       .set(s"spark.sql.catalog.$catalogName.ref", ref.getName)
-    SparkSession.active.sparkContext.conf
-      .set(s"spark.sql.catalog.$catalogName.ref.hash", ref.getHash)
+    if (configureRefAtHash) {
+      // we only configure ref.hash if we're reading data
+      SparkSession.active.sparkContext.conf
+        .set(s"spark.sql.catalog.$catalogName.ref.hash", ref.getHash)
+    } else {
+      // we need to clear it in case it was previously set
+      SparkSession.active.sparkContext.conf
```
ajantha-bhat, line 225: I am not sure whether we need to clear this conf or set it to the current head of this reference.
I think it depends on catalog implementation.

Hope you have verified with end to end changes. 

These order of operations should work:
```
use reference@hash/timestamp ; --(points to old commit)
select * from table; -- data from old commit
insert to the table; --?? may be should not work.
use reference; -- (points to head)
select * from table; -- data from head
insert to the table; --should work.
select * from table@hash/timestamp ; -- data from old commit
```

ajantha-bhat (approved):
These changes looks good to me.

clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
ajantha-bhat: Resetting to head will become same as master branch code. So, I guess. This PR should help us to handle things at Iceberg side. 
LGTM.

nastra: yes resetting is actually required (I verified with end-to-end testing)

PR #4169, opened 2022-05-05
Use Junit5 assumptions in versioned tests + versioned/persist tests

snazy:


nastra (approved):
LGTM, but would be good to add some details to the commit msg why it's suddenly required now. Also should we add this in other modules?

PR #4176, opened 2022-05-06
Misc. fixes

snazy:
* make utility classes final w/ private c'tor
* enum using `==` instead of `.equals()`
* `String.isEmpty()` instead of `"".equals()`
* `HttpRequest` w/ `queryParam()` taking an `Integer`
* non-constant string in call to `printf` in `ReadReferences`
* collection field visibility in `RocksDbInstance`
* precompiled `Pattern`

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpRefLogClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpRefLogClient.java
@@ -33,8 +33,7 @@ class HttpRefLogClient implements HttpRefLogApi {
   public RefLogResponse getRefLog(@NotNull RefLogParams params) throws NessieNotFoundException {
     HttpRequest builder = client.newRequest().path("reflogs");
     return builder
-        .queryParam(
-            "maxRecords", params.maxRecords() != null ? params.maxRecords().toString() : null)
+        .queryParam("maxRecords", params.maxRecords())
         .queryParam("pageToken", params.pageToken())
         .queryParam("startHash", params.startHash())
         .queryParam("endHash", params.endHash())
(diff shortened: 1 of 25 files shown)


clients/iceberg-views/src/main/java/org/apache/iceberg/view/VersionParser.java
```diff
@@ -106,14 +106,19 @@ static Version fromJson(JsonNode node) {
...
+            break;
+          case CommonViewConstants.GENIE_ID:
+            builder.put(CommonViewConstants.GENIE_ID, JsonUtil.getString(field, summaryNode));
+            break;
+          case CommonViewConstants.ENGINE_VERSION:
+            builder.put(CommonViewConstants.ENGINE_VERSION, JsonUtil.getString(field, summaryNode));
+            break;
+          default:
+            builder.put(field, JsonUtil.getString(field, summaryNode));
+            break;
```
XN137, line 121: arent all switch cases doing the exact same thing? :thinking: 
```
builder.put(field, JsonUtil.getString(field, summaryNode));
```

XN137 (dismissed):
typo in commit message: quaryParam

lgtm with 2 comments worth looking at

clients/iceberg-views/src/main/java/org/apache/iceberg/view/VersionParser.java
snazy: Duh - did not want to touch iceberg-views at all.
Reverted this change.

@nastra there are a couple more issues, some are actual issues. But iceberg-views is WIP stuff - so IDK what to do with the module.

XN137 (approved):
commit msg still has typo ;)

PR #4177, opened 2022-05-06
Database-adapter: commit optimizations

snazy:
* Do only fetch the HEAD of the target branch once
* Never try to fetch NO_ANCESTOR (it can never exist)
* Allow diffent sizes for the "embedded key list" (in a commit) and
  key-list-entities (separate rows)
* Nit: add commit-hash to conflict message


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -111,6 +112,38 @@ default int getDefaultMaxKeyListSize() {
...
+   * Database adapter implementations that actually do have a hard technical or highly recommended
+   * limit on a maximum db-object / db-row size limitation should override this method and return a
+   * "good" value.
+   *
+   * <p>As for {@link #getMaxKeyListSize()}, this value must not be "on the edge" - means: it must
+   * leave enough room for database-serialization overhead * and similar.
+   *
+   * <p>Defaults to {@value #DEFAULT_MAX_KEY_LIST_ENTITY_SIZE}.
+   */
+  default int getDefaultMaxKeyListEntitySize() {
```
dimas-b, line 143: I'm not sure how it is going to work :thinking: In current config usage the `getMaxKeyListEntitySize` setting  apparently always overrides `getDefaultMaxKeyListEntitySize`... Is it worth having both of these settings? If we want the "default" size to be database-specific, should the code make sure `getDefaultMaxKeyListEntitySize` takes precedence?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -111,6 +112,38 @@ default int getDefaultMaxKeyListSize() {
     return DEFAULT_MAX_KEY_LIST_SIZE;
```
dimas-b, line 112: Do we need two database-level "hard" limit settings (wrt `getDefaultMaxKeyListEntitySize`)?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -725,6 +734,11 @@ protected final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
    * returned as {@code null}.
    */
   private List<CommitLogEntry> fetchMultipleFromCommitLog(OP_CONTEXT ctx, List<Hash> hashes) {
+    if (hashes.contains(NO_ANCESTOR)) {
+      // Do not try to fetch NO_ANCESTOR - it won't exist.
+      hashes = new ArrayList<>(hashes);
+      hashes.remove(NO_ANCESTOR);
```
dimas-b, line 740: javadoc: `The returned list must have exactly as many elements as in the parameter {@code hashes}.`

Aren't we violating the javadoc contract of this method here?

snazy: Duh! You're right!

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
snazy: But yea - having a configurable default for another configurable doesn't really make much sense. Removed the configurable defaults.

dimas-b: Yes, that makes sense, but I guess that is controlled by the "non-default" parameters, right? My reading of the "default" parameters was that they define a hard limit for database-level entity sizes, which would normally be the same for any entity, I guess :thinking:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
dimas-b: Should the method then be on the `DatabaseAdapter` interface? Or perhaps of the adapter factory?

dimas-b: right now this value comes from the same source (e.g. `QuarkusVersionStoreAdvancedConfig`) as the other "key list" size, which makes it a bit confusing from my POV :thinking:

site/docs/try/configuration.md
```diff
@@ -76,19 +76,21 @@ When setting `nessie.version.store.type=DYNAMO` which enables DynamoDB as the ve
...
+| `nessie.version.store.advanced.repository-id`                     |                     | `String` | Sets Nessie repository ID (optional). This ID can be used to distinguish multiple Nessie repositories that reside in the same storage instance.                                                                                       |
+| `nessie.version.store.advanced.parent-per-commit`                 | `20`                | `int`    | Sets the number of parent-commit-hashes stored in Nessie store.                                                                                                                                                                       |
+| `nessie.version.store.advanced.key-list-distance`                 | `20`                | `int`    | Sets the number "reachable" or "known" keys for each `CommitLogEntry`.                                                                                                                                                                |
```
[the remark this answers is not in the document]
snazy: fixed

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
snazy: (See below - will do in a separate PR)

snazy: But yes, it  (the hard limit) is an implementation concern and should be handled there.

dimas-b (dismissed):
LGTM :+1:

dimas-b (approved):

PR #4182, opened 2022-05-09
GC: Remove commit protection time configuration

ajantha-bhat:
In GC, cutoff time itself can be used to protect the new commits. No need of a separate configuration.
Hence removed it and used the cutoff time for commit protection


ajantha-bhat:
> I don't understand the purpose of the change.
cutoff-time is different from protection-time. The latter is to protect commits that have been recently added and not considered during identification.
Shouldn't "GC identification" collect the live contents and expire all unreferenced snapshots?

@snazy : "GC identification" step itself has two phases. It accepts a cutoff time (which cannot be from the future)
and uses the same cutoff time for both the phases.
If any new commits added during/after phase1 and before phase2, its commit time will be definitely newer than cutoff time. So, NO need of an extra configuration to protect those commits. It can be protected using the cutoff time itself.

snazy:
Have you thought about cutoff-time == 0?

ajantha-bhat:
>  Have you thought about cutoff-time == 0?

cutoff-time is not a duration, it is an Instant. So, I guess you are referring `cutofftime = now()` [**Also it cannot be from future**]
So, now() will be the beginning instant of phase1 of GC-identfy, so the new commits during/after phase1 can still be protected by cutoffTime as its commit time will be newer than cutoff time.

snazy:
And phase 2 expires all snapshots that are not identified as live in phase 1 - so there is definitely a gap.

ajantha-bhat:
> And phase 2 expires all snapshots that are not identified as live in phase 1 - so there is definitely a gap.

Nope. As I mentioned above, phase2 will check only the commits that are older cutoff time. So, these new commits will never be checked against bloom filter. So, it cannot be expired.  

Here is that logic of replacing commit protection time with cutoff time which handles the same.

https://github.com/projectnessie/nessie/pull/4182/files#diff-1bf51914a6143465d680b79d8ebcec9486ff4093c10dd712a6d443ff1fb48583R293

ajantha-bhat:
@snazy : Could you please take a fresh look at it? I still think there is no issues with this PR.

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCParams.java
```diff
@@ -89,14 +79,31 @@ default double getBloomFilterFpp() {
 
   @Value.Check
   default void validate() {
-    Integer taskCount = getSparkPartitionsCount();
-    if (taskCount != null && taskCount <= 0) {
-      throw new IllegalArgumentException("taskCount has invalid value: " + taskCount);
+    Instant now = Instant.now();
```
snazy, line 82: All these changes are unrelated to the scope of the PR.

gc/gc-base/src/main/java/org/projectnessie/gc/base/IdentifyContentsPerExecutor.java
```diff
@@ -241,30 +245,32 @@ private scala.collection.Iterator<Row> getExpiredContentRows(
...
-            JavaConverters.asScalaIterator(
-                    walkAllCommitsInReference(
-                        api,
-                        GCUtil.deserializeReference(reference),
-                        liveContentsBloomFilterMap,
-                        runId,
-                        startedAt))
-                .toTraversable());
+        reference -> {
+          Reference ref = GCUtil.deserializeReference(reference);
```
snazy, line 249: This is a noop change that unnecessarily increases the amount of changes and review effort

snazy (changes requested):

gc/gc-base/src/main/java/org/projectnessie/gc/base/GCParams.java
ajantha-bhat: I really wanted to have validations that ensure these timestamp objects are not from the future. 
It can be a separate PR, but don't want too many PR for nit stuffs.

snazy: Never ever make assumptions about the wall clock.
Yes, I do want to have this check moved out, because your assumption that wall clock is monotonically increasing is not correct and Instants in the future make sense.

ajantha-bhat: ok. moved out now.

ajantha-bhat:
Anything else needed for this PR to be approved?

ajantha-bhat:
#3885 will handle this.

PR #4204, opened 2022-05-10
Remove the configurable default for the configurable values

snazy:
The max entity size for the (embedded) key list is configured using two values from the
`DatabaseAdapterConfig` interface via `getMaxKeyListSize` and `getDefaultMaxKeyListSize`,
the latter serves as the default for the former. This is confusing and actually wrong.

This change removed the configurable default entirely and introduces an optional
"hard limit" enforced by database adapter implementations - DynamoDB configured to 375kB.


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1052,7 +1052,8 @@ protected CommitLogEntry buildKeyList(
               if (buildState.embedded) {
                 // filling the embedded key-list in CommitLogEntry
 
-                if (buildState.currentSize + keyTypeSize < config.getMaxKeyListSize()) {
+                if (buildState.currentSize + keyTypeSize
+                    < Math.min(maxEntitySize(), config.getMaxKeyListSize())) {
```
dimas-b, line 1056: nit: extract a method for the `min` logic since it's used in two places?

dimas-b: or perhaps fold it into the impl of `maxEntitySize`?

snazy: latter, also realized that there was another "configurable default" :(

PR #4223, opened 2022-05-12
Helm TRANSACTIONAL store type support

kerossin:
Add TRANSACTIONAL store type support to the Helm chart.

### DIFF
--- a/helm/nessie/templates/deployment.yaml
+++ b/helm/nessie/templates/deployment.yaml
@@ -81,6 +81,26 @@ spec:
             {{- end }}
             {{- end }}
 
+            {{- if eq .Values.versionStoreType "TRANSACTIONAL" }}
+            - name: QUARKUS_DATASOURCE_USERNAME
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.postgres.secret.name }}
+                  key: {{ .Values.postgres.secret.username }}
+            - name: QUARKUS_DATASOURCE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.postgres.secret.name }}
+                  key: {{ .Values.postgres.secret.password }}
+            - name: QUARKUS_DATASOURCE_JDBC_URL
+              value: {{ .Values.postgres.jdbcUrl }}
+            {{- end }}
+
+            {{- range .Values.versionStoreAdvancedConfig }}
+            - name: {{ .key }}
+              value: {{ .value | quote }}
+            {{- end }}
+
             {{- if .Values.authentication.enabled }}
             - name: NESSIE_SERVER_AUTHENTICATION_ENABLED
               value: {{ .Values.authentication.enabled | quote }}
--- a/helm/nessie/values.yaml
+++ b/helm/nessie/values.yaml
@@ -9,7 +9,7 @@ image:
 # The default logging level for the nessie server
 logLevel: INFO
 
-# which type of version store to use: INMEMORY, ROCKS, DYNAMO, MONGO
+# which type of version store to use: INMEMORY, ROCKS, DYNAMO, MONGO, TRANSACTIONAL
 versionStoreType: INMEMORY
 
 # this is needed when selecting ROCKS
@@ -28,6 +28,19 @@ dynamodb:
 mongodb:
   name: nessie
   connectionString: mongodb://localhost:27017
+
+# this is needed when selecting TRANSACTIONAL
+postgres:
+  jdbcUrl: jdbc:postgresql://localhost:5432/my_database
+  secret:
+    name: postgres-creds
+    username: postgres_username
+    password: postgres_password
+
+# environment variables for advanced version store configuration
+# versionStoreAdvancedConfig:
+#   - key: QUARKUS_DATASOURCE_JDBC_POOLING_ENABLED
+#     value: true
 
 # configures authentication for the nessie server
 authentication:
--- a/helm/nessie/README.md
+++ b/helm/nessie/README.md
@@ -30,7 +30,7 @@ The following table lists the configurable parameters of the Nessie chart and th
 | image.pullPolicy | Image pull policy, such as `IfNotPresent` / `Always` / `Never` | `IfNotPresent` |
 | image.tag | Overrides the image tag whose default is the chart version | `version` from `Chart.yaml` |
 | logLevel | The default logging level to be used | `INFO` |
-| versionStoreType | Version store to use: `INMEMORY` / `ROCKS` / `DYNAMO` / `MONGO` | `INMEMORY` |
+| versionStoreType | Version store to use: `INMEMORY` / `ROCKS` / `DYNAMO` / `MONGO` / `TRANSACTIONAL` | `INMEMORY` |
 | rocksdb | Configuration specific to `versionStoreType: ROCKS` | |
 | rocks.dbPath | Sets RocksDB storage path | `/tmp/rocks-nessie` |
 | dynamodb | Configuration specific to `versionStoreType: DYNAMO` | |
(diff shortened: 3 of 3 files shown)


helm/nessie/templates/deployment.yaml
```diff
@@ -81,6 +81,35 @@ spec:
             {{- end }}
             {{- end }}
 
+            {{- if eq .Values.versionStoreType "TRANSACTIONAL" }}
```
ajantha-bhat, line 84: As TRANSACTIONAL backend store, Nessie supports H2 also along with Postgres.

So, Instead of checking with TRANSACTIONAL, can we have separate one for each transactional adapters? (similar to how non-transactional backend type as one for each). 

It will be easy to add some more transactional type support in the future.

ajantha-bhat: oh, I see that enum itself has just trasnactional, 

https://github.com/projectnessie/nessie/blob/main/servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java#L35

@snazy, @nastra, @dimas-b:   Should we rename it as specific (like postgres) to support other transactional ones ?

kerossin: I'm not sure how this works but it seems this Nessie image is built only with Postgres: https://github.com/projectnessie/nessie/blob/main/servers/quarkus-cli/src/main/resources/application.properties#L26

ajantha-bhat: Yeah, I observed that and hence posted that question. 
@kerossin: Sorry for spamming your PR. I will keep it in a separate discussion.

dimas-b: We have tests for H2 as well. If there's interest it should be possible to build an image for it too.

dimas-b: but yes, at this point only Postgres can be selected at runtime. It looks like supporting other JDBC drivers will require some code changes... Please feel free to open an issue for this :) At this point H2 should be easy to enable, but other RDBMS' should be possible too.

dimas-b (commented):
Thanks for contributing. I added a couple of nit-picky comments for code changes.

These change look good to me overall, however, if `TRANSACTIONAL` becomes used in actual deployments it might be preferable to clean-up the database adapter setup code to make sure other database backends can be used in the future without breaking existing use cases. I suspect this might involve renaming the `TRANSACTIONAL` enum (among other changes).

What do you think about opening a PR or issue for enabling first class support for RDBMS storage (with Postres and H2 for now) and then rebasing this work when the dabasebase adapter changes are ironed out?

helm/nessie/values.yaml
```diff
@@ -37,6 +37,25 @@ authentication:
   # Set the OIDC client ID when authentication.enabled=true. Each application has a client ID that is used to identify the application
   oidcClientId: nessie
 
+# this is needed when selecting TRANSACTIONAL
+postgres:
+  secret:
+    name: postgres-creds
+    usernameKey: username
+    passwordKey: password
+  jdbcUrl: jdbc:postgresql://localhost:5432/my_database
```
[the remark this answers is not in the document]
kerossin: Yes, I meant this an example. I reused it from here https://github.com/projectnessie/nessie/blob/main/servers/quarkus-cli/src/main/resources/application.properties#L33

helm/nessie/templates/deployment.yaml
```diff
@@ -81,6 +81,35 @@ spec:
...
+            - name: QUARKUS_DATASOURCE_DB_KIND
+              value: postgresql
```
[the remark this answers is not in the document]
kerossin: You're right. Removed it.

helm/nessie/values.yaml
```diff
@@ -37,6 +37,25 @@ authentication:
   # Set the OIDC client ID when authentication.enabled=true. Each application has a client ID that is used to identify the application
   oidcClientId: nessie
 
+# this is needed when selecting TRANSACTIONAL
+postgres:
+  secret:
+    name: postgres-creds
+    usernameKey: username
```
nastra, line 44: nit: let's rename those to `username=postgres_username` and `password=postgres_password` as the `Key` suffix is rather confusing. This also requires some additional docs in the README, similar to how we have it for DYNAMO (see my example below). 


### Providing secrets for Postgres

* Make sure you have a Secret in the following form:
```
> cat $PWD/postgres-creds
postgres_username=YOUR_USERNAME
postgres_password=YOUR_PASSWORD
```

helm/nessie/values.yaml
```diff
@@ -37,6 +37,25 @@ authentication:
...
+    passwordKey: password
+  jdbcUrl: jdbc:postgresql://localhost:5432/my_database
+
+# JDBC configuration options for Quarkus https://quarkus.io/guides/datasource
+# quarkusJdbcOptions:
+#   - key: QUARKUS_DATASOURCE_JDBC_POOLING_ENABLED
+#     value: true
+
+# Sets the catalog name to use via JDBC
```
nastra, line 53: given that we might want to expose more settingsI think it would make sense to move these settings into its own `versionStoreAdvancedConfig` section

nastra (commented):
In general the changes LGTM, just a few small improvements that should be done.

It would be great to also update the documentation about those new settings in the `README`

helm/nessie/values.yaml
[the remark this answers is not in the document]
kerossin: Sorry, was on vacation last week. Will look into your comments this week.

Do you mean to move `jdbcUrl` like:
```
postgres:
  jdbcUrl:
  secret:
```
?

helm/nessie/values.yaml
kerossin: With the `Key` suffix I meant to clarify that you need to specify the name of the key in the secret rather than the actual username or password but I can change it.

helm/nessie/values.yaml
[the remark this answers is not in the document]
nastra: yes exactly

helm/nessie/values.yaml
kerossin: Sure.

Should I leave these ones or remove them and let someone configure them as part of `versionStoreAdvancedConfig` as needed?
```
# Sets the catalog name to use via JDBC
# versionStoreAdvancedTxJdbcCatalog: nessie

# Sets the schema name to use via JDBC
# versionStoreAdvancedTxJdbcSchema: nessie
```

helm/nessie/values.yaml
nastra: yeah but when referring to secrets you're generally always referring to the Key within that secret, thus no need to use the `Key` suffix. With AWS secrets it's obviously different because the `Key` suffix is already part of the name "awsSecretAccessKey"

helm/nessie/values.yaml
nastra: if they are not strictly necessary for now, then just leave them out

nastra:
@kerossin can you also please update the docs in the README.md file to reflect those new settings?

helm/README.md
```diff
@@ -41,6 +41,12 @@ The following table lists the configurable parameters of the Nessie chart and th
 | mongodb | Configuration specific to `versionStoreType: MONGO` | |
 | mongodb.name | The database name | `nessie` |
 | mongodb.connectionString | The database connection string | `mongodb://localhost:27017` |
+| postgres | Configuration specific to `versionStoreType: TRANSACTIONAL` | |
+| postgres.jdbcUrl | The database connection URL | `jdbc:postgresql://localhost:5432/my_database` |
+| postgres.secret.name | The name of the secret where database credentials are stored | `postgres-creds` |
+| postgres.secret.username | The name of the username key inside the secret | `postgres_username` |
+| postgres.secret.password | The name of the password key inside the secret | `postgres_password` |
```
nastra, line 48: see https://github.com/projectnessie/nessie/pull/4223#discussion_r873385806, let's please also add a section below with:
### Providing secrets for Postgres

* Make sure you have a Secret in the following form:
```
> cat $PWD/postgres-creds
postgres_username=YOUR_USERNAME
postgres_password=YOUR_PASSWORD
```

nastra:
@kerossin can you please rebase and fix the merge conflict so that I could do another round of review? Note that the README was moved to a different location

nastra (approved):

PR #4230, opened 2022-05-13
Do not process old key-lists when retrieving values.

snazy:
Only process the "newest" key-list - older key lists are irrelevant.
`KeyListEntry` written before Nessie 0.22.0 do not have the commit-ID field set,
which means the remaining commit-log needs to be searched for the commit that
added the key - but processing older key-lists "on the way" makes no sense,
because those will not have the key either.


dimas-b (approved):
nice catch :+1:

PR #4232, opened 2022-05-13
Add commit-ID to KeyListEntry when writing new key-lists

snazy:
`KeyListEntry` written before Nessie 0.22.0 do not have the commit-ID field set,
which means the remaining commit-log needs to be searched for the commit that
added the key.

This change adds the missing commit-ID when writing a new key-list.

### DIFF
--- a/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalOperationContext.java
+++ b/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalOperationContext.java
@@ -15,11 +15,14 @@
  */
 package org.projectnessie.versioned.persist.nontx;
 
-public class NonTransactionalOperationContext {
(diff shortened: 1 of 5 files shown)


versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalOperationContext.java
```diff
@@ -15,11 +15,14 @@
  */
 package org.projectnessie.versioned.persist.nontx;
 
-public class NonTransactionalOperationContext {
+public class NonTransactionalOperationContext implements AutoCloseable {
```
dimas-b, line 18: Is this needed for the key list fix? maybe a separate PR?

snazy: Just for this PR, yes.

dimas-b (approved):

PR #4234, opened 2022-05-13
Eliminiate global-pointer contention (non-tx database adapters)

snazy:
* Moves named-references from global-pointer to separate entities (one
  "row" per named reference), concurrent updates to the same named
  ref can content, but concurrent updates to the different named refs
  do not (minus ref-log, see below).
* Move ref-log HEAD from global-pointer to striped (mutliple) ref-log
  HEADs, using the name of the reference. This reduces the probability
  for contention to 1/16th (by default). Number of stripes can be
  increased (requires a restart of all nessie instances, no rolling
  upgrade).

Both changes are backwards compatible:
* Existing named references are migrated when Nessie finds existing
  named referenes in the global pointer.
* Existing ref-log does not need to be migrated, ref-log reads just
  "include" the ref-log HEAD in global-pointer as an additional
  stripe.

During a Nessie commit there is a (neglectible) chance that the
commit is written and the named-reference correctly updated, but the
corresponding ref-log-entry not written. This can happen when the
Nessie server crashes or is stopped right a successful CAS of the
named-reference HEAD but before the ref-log HEAD CAS. This is a
trade-off.

Upgrade procedure:
1. stop all existing Nessie instances
2. bump versions
3. restart.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
@@ -40,9 +40,12 @@ public class TryLoopState implements AutoCloseable {
   private final Function<TryLoopState, String> retryErrorMessage;
   private final BiConsumer<Boolean, TryLoopState> completionNotifier;
   private final long maxSleep;
+  private final long initialLowerBound;
+  private final long initialUpperBound;
   private long lowerBound;
   private long upperBound;
   private int retries;
+  private boolean unsuccessful;
 
   TryLoopState(
       String opName,
@@ -56,8 +59,8 @@ public class TryLoopState implements AutoCloseable {
     this.maxRetries = config.getCommitRetries();
     this.monotonicClock = monotonicClock;
     this.t0 = monotonicClock.currentNanos();
-    this.lowerBound = config.getRetryInitialSleepMillisLower();
-    this.upperBound = config.getRetryInitialSleepMillisUpper();
+    this.lowerBound = this.initialLowerBound = config.getRetryInitialSleepMillisLower();
+    this.upperBound = this.initialUpperBound = config.getRetryInitialSleepMillisUpper();
     this.maxSleep = config.getRetryMaxSleepMillis();
     this.completionNotifier = completionNotifier;
     start();
@@ -109,6 +112,10 @@ public class TryLoopState implements AutoCloseable {
    * exceeded the configured values.
    */
   public void retry() throws ReferenceRetryFailureException {
+    if (unsuccessful) {
+      throw unsuccessful();
+    }
+
     stop();
 
     retries++;
@@ -117,6 +124,7 @@ public class TryLoopState implements AutoCloseable {
     long elapsed = current - t0;
 
     if (maxTime < elapsed || maxRetries < retries) {
+      unsuccessful = true;
       throw unsuccessful();
     }
 
@@ -132,6 +140,27 @@ public class TryLoopState implements AutoCloseable {
       lowerBound *= 2;
       upperBound = upper;
     }
+
+    start();
+  }
+
+  /**
+   * Similar to {@link #retry()}, but never throws a {@link ReferenceRetryFailureException} (unless
+   * a preceding call to {@link #retry()} already raised it) and always uses the initial lower and
+   * upper bounds for the sleep duration (no backoff).
+   */
+  public void retryEndless() throws ReferenceRetryFailureException {
+    if (unsuccessful) {
+      throw unsuccessful();
+    }
+
+    stop();
+
+    retries++;
+
+    long sleepMillis = ThreadLocalRandom.current().nextLong(initialLowerBound, initialUpperBound);
+
+    monotonicClock.sleepMillis(sleepMillis);
 
     start();
   }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
@@ -30,6 +30,7 @@ public class TryLoopState implements AutoCloseable {
 
   private static final String TAG_ATTEMPT = "try-loop.attempt";
   private static final String TAG_RETRIES = "try-loop.retries";
+  private static final String TAG_ENDLESS_RETRIES = "try-loop.endless-retries";
 
   private Traced traced;
   private final String opName;
@@ -45,6 +46,7 @@ public class TryLoopState implements AutoCloseable {
   private long lowerBound;
   private long upperBound;
   private int retries;
+  private int endlessRetries;
   private boolean unsuccessful;
 
   TryLoopState(
@@ -83,6 +85,14 @@ public class TryLoopState implements AutoCloseable {
     return retries;
   }
 
+  public int getEndlessRetries() {
+    return endlessRetries;
+  }
+
+  public int getTotalRetries() {
+    return retries + endlessRetries;
+  }
+
   public long getDuration(TimeUnit timeUnit) {
     return timeUnit.convert(monotonicClock.currentNanos() - t0, TimeUnit.NANOSECONDS);
   }
@@ -99,7 +109,9 @@ public class TryLoopState implements AutoCloseable {
   private ReferenceRetryFailureException unsuccessful() {
     completionNotifier.accept(false, this);
     return new ReferenceRetryFailureException(
-        retryErrorMessage.apply(this), this.getRetries(), this.getDuration(TimeUnit.MILLISECONDS));
+        retryErrorMessage.apply(this),
+        this.getTotalRetries(),
+        this.getDuration(TimeUnit.MILLISECONDS));
   }
 
   /**
@@ -123,11 +135,40 @@ public class TryLoopState implements AutoCloseable {
     long current = monotonicClock.currentNanos();
     long elapsed = current - t0;
 
-    if (maxTime < elapsed || maxRetries < retries) {
+    if (maxTime < elapsed || maxRetries < retries + endlessRetries) {
       unsuccessful = true;
       throw unsuccessful();
     }
 
+    sleepAndBackoff(elapsed);
+
+    start();
+  }
+
+  public void resetBounds() {
+    lowerBound = initialLowerBound;
+    upperBound = initialUpperBound;
+  }
+
+  /**
+   * Similar to {@link #retry()}, but never throws a {@link ReferenceRetryFailureException} (unless
+   * a preceding call to {@link #retry()} already raised it).
+   */
+  public void retryEndless() throws ReferenceRetryFailureException {
+    if (unsuccessful) {
+      throw unsuccessful();
+    }
+
+    stop();
+
+    endlessRetries++;
+
+    sleepAndBackoff(0L);
+
+    start();
+  }
+
+  private void sleepAndBackoff(long elapsed) {
     long sleepMillis = ThreadLocalRandom.current().nextLong(lowerBound, upperBound);
 
     // Prevent that we "sleep" too long and exceed 'maxTime'
@@ -140,29 +181,6 @@ public class TryLoopState implements AutoCloseable {
       lowerBound *= 2;
       upperBound = upper;
     }
-
-    start();
-  }
-
-  /**
-   * Similar to {@link #retry()}, but never throws a {@link ReferenceRetryFailureException} (unless
-   * a preceding call to {@link #retry()} already raised it) and always uses the initial lower and
-   * upper bounds for the sleep duration (no backoff).
-   */
-  public void retryEndless() throws ReferenceRetryFailureException {
-    if (unsuccessful) {
-      throw unsuccessful();
-    }
-
-    stop();
-
-    retries++;
-
-    long sleepMillis = ThreadLocalRandom.current().nextLong(initialLowerBound, initialUpperBound);
-
-    monotonicClock.sleepMillis(sleepMillis);
-
-    start();
   }
 
   @Override
@@ -200,10 +218,10 @@ public class TryLoopState implements AutoCloseable {
   }
 
   private void start() {
-    traced = trace(opName).tag(TAG_ATTEMPT, getRetries());
+    traced = trace(opName).tag(TAG_ATTEMPT, getTotalRetries());
   }
 
   private void stop() {
-    traced.tag(TAG_RETRIES, getRetries()).close();
+    traced.tag(TAG_RETRIES, getRetries()).tag(TAG_ENDLESS_RETRIES, getEndlessRetries()).close();
   }
 }
--- a/versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
+++ b/versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
@@ -634,7 +634,9 @@ public class MongoDatabaseAdapter
   }
 
   @Override
-  protected void doCleanUpRefLogWrite(NonTransactionalOperationContext ctx, Hash refLogId) {}
+  protected void doCleanUpRefLogWrite(NonTransactionalOperationContext ctx, Hash refLogId) {
+    client.getRefLog().deleteOne(Filters.eq(toId(refLogId)));
+  }
 
   @Override
   protected List<ReferenceNames> doFetchReferenceNames(
--- a/versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
+++ b/versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
@@ -22,6 +22,7 @@ import com.google.common.collect.Maps;
 import com.google.protobuf.InvalidProtocolBufferException;
 import com.mongodb.DuplicateKeyException;
 import com.mongodb.ErrorCategory;
+import com.mongodb.MongoServerException;
 import com.mongodb.MongoWriteException;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCollection;
@@ -189,15 +190,13 @@ public class MongoDatabaseAdapter
     InsertOneResult result;
     try {
       result = collection.insertOne(doc);
-    } catch (MongoWriteException writeException) {
-      ErrorCategory category = writeException.getError().getCategory();
-      if (ErrorCategory.DUPLICATE_KEY == category) {
+    } catch (MongoServerException e) {
+      if (isDuplicateKeyError(e)) {
         ReferenceConflictException ex = DatabaseAdapterUtil.hashCollisionDetected();
-        ex.initCause(writeException);
+        ex.initCause(e);
         throw ex;
       }
-
-      throw writeException;
+      throw e;
     }
 
     verifyAcknowledged(result, collection);
@@ -212,15 +211,13 @@ public class MongoDatabaseAdapter
     InsertManyResult result;
     try {
       result = collection.insertMany(docs);
-    } catch (MongoWriteException writeException) {
-      ErrorCategory category = writeException.getError().getCategory();
-      if (ErrorCategory.DUPLICATE_KEY == category) {
+    } catch (MongoServerException e) {
+      if (isDuplicateKeyError(e)) {
         ReferenceConflictException ex = DatabaseAdapterUtil.hashCollisionDetected();
-        ex.initCause(writeException);
+        ex.initCause(e);
         throw ex;
       }
-
-      throw writeException;
+      throw e;
     }
 
     verifyAcknowledged(result, collection);
@@ -373,12 +370,11 @@ public class MongoDatabaseAdapter
     } else {
       try {
         return client.getRepoDesc().insertOne(doc).wasAcknowledged();
-      } catch (MongoWriteException writeException) {
-        ErrorCategory category = writeException.getError().getCategory();
-        if (ErrorCategory.DUPLICATE_KEY == category) {
+      } catch (MongoServerException e) {
+        if (isDuplicateKeyError(e)) {
           return false;
         }
-        throw writeException;
+        throw e;
       }
     }
   }
@@ -421,8 +417,11 @@ public class MongoDatabaseAdapter
       try {
         verifyAcknowledged(client.getRefLogHeads().insertOne(doc), client.getRefLogHeads());
         return true;
-      } catch (DuplicateKeyException e) {
-        return false;
+      } catch (MongoServerException e) {
+        if (isDuplicateKeyError(e)) {
+          return false;
+        }
+        throw e;
       }
     }
   }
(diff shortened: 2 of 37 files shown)


dimas-b:
Does `status WIP ...` in the description still apply?

snazy:
> Does `status WIP ...` in the description still apply?

Ooops. No.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -785,239 +722,550 @@ protected Hash casOpLoop(
...
+          if (!createNamedReference(
+              ctx,
+              toNamedRef(legacyNamedRef.getRef().getType(), legacyNamedRef.getName()),
+              Hash.of(legacyNamedRef.getRef().getHash()))) {
+            if (doFetchNamedReference(ctx, refName) == null) {
+              newPointer.addNamedReferences(legacyNamedRef);
+            }
+          }
+        }
+        // TODO better use CAS?
```
snazy, line 1040: WDYT? Shall we keep the "global pointer CAS" for this use case?

snazy: (Updated the migration-algo in a newer commit)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -434,27 +612,31 @@ protected boolean doGlobalPointerCas(
...
   @Override
-  protected void doCleanUpGlobalLog(
-      NonTransactionalOperationContext ctx, Collection<Hash> globalIds) {
-    delete(client.getGlobalLog(), globalIds);
+  protected void doCleanUpRefLogWrite(NonTransactionalOperationContext ctx, Hash refLogId) {}
+
+  @Override
+  protected ReferenceNames doFetchReferenceNames(
+      NonTransactionalOperationContext ctx, int segment) {
+    Document doc = client.getRefNames().find(Filters.eq(toId(segment))).first();
```
dimas-b, line 626: Could we do a full scan instead of looping over segments on our side?

dimas-b (commented):
Just some preliminary comments... will continue reviewing tomorrow.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -785,63 +727,538 @@ protected Hash casOpLoop(
...
+
+  protected final RefLogParents fetchRefLogParents(
+      NonTransactionalOperationContext ctx, int stripe) {
+    try (Traced ignore = trace("fetchRefLogParentsForReference")) {
+      return doFetchRefLogParents(ctx, stripe);
+    }
+  }
+
+  protected final int refLogStripeForName(String refName) {
+    return Math.abs(refName.hashCode() % config.getRefLogStripes());
```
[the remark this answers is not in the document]
snazy: shouldn't matter much - just needs to be deterministic. But changed.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -785,63 +727,538 @@ protected Hash casOpLoop(
...
+        int stripe = refLogStripeForName(ref.getName());
+        while (true) {
+          RefLogParents refLogParents = fetchRefLogParents(ctx, stripe);
+
+          RefLogEntry newRefLog = writeRefLogEntry(ctx, refLogParents, result.refLogEntry);
+
+          RefLogParents.Builder newRefLogParents =
+              RefLogParents.newBuilder()
+                  .addRefLogParentsInclHead(newRefLog.getRefLogId())
+                  .setLockId(randomHash().asBytes());
```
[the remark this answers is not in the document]
snazy: It doesn't identify the instance, but the "version" for the CAS operation.
`version` okay? Added a comment in `persist.proto`

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -323,16 +370,156 @@ protected boolean doTryUpdateRepositoryDescription(
...
+                .build());
+    return verifySuccessfulUpdate(
+        client
+            .getRefHeads()
+            .updateOne(
+                Filters.and(
+                    Filters.eq(toId(ref.getName())),
+                    Filters.eq(ID_HASH_NAME, refHead.getHash().toByteArray())),
+                new Document("$set", newDoc)),
+        client.getRefNames());
```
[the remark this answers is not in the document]
snazy: TBH, I don't see the the difference in the proposals (beside the collection mismatch).
Updated `verifySuccessfulUpdate` to take an "updater" `Function`

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: `version` sgtm :+1:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: Updated the code to fetch/resolve up to 25 named references at once.

snazy: And prefetch reference-names segments (1 by default).

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: ok :+1:

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Besides will Mongo not batch scan results on the server side (I hope, but do not know)?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -832,63 +761,525 @@ protected Hash casOpLoop(
...
+
+    private final Hash initialHash;
+    private boolean initialHashSeen;
+    private RefLog initialRefLog;
+
+    RegLogSpliterator(NonTransactionalOperationContext ctx, int stripes, Hash initialHash)
+        throws RefLogNotFoundException {
+      super(Long.MAX_VALUE, Spliterator.ORDERED);
+
+      this.initialHash = initialHash;
```
dimas-b, line 915: It looks like `initialHash` has dual purpose - proper initial hash and pagination token. I suppose in the latter case we could use a composite token to keep track of cursors in all stripes.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -832,63 +761,525 @@ protected Hash casOpLoop(
...
-        } else if (opVariant.commitOp) {
-          if (branchHead != null) {
-            individualCommits.add(branchHead);
+        // CAS against branch-heads succeeded, now try to write the ref-log-entry
+
+        int stripe = refLogStripeForName(ref.getName());
+        while (true) {
+          RefLogParents refLogParents = fetchRefLogParents(ctx, stripe);
+
+          RefLogEntry newRefLog = writeRefLogEntry(ctx, refLogParents, result.refLogEntry);
```
dimas-b, line 815: I guess ref log entries may be reordered with respect to commits, e.g. server S1 makes commit `1` for ref `A`, the server S2 makes commit `2` for ref `A`, then server S2 makes ref log `A2`, then server S1 makes ref log `A1`. Is that a problem?

If not, I suppose we might as well have a flexible number of stripes. Even different servers could be using different numbers of stripes (as long as they know an upper bound for iteration)... and we could select a stripe randomly for each write. WDYT?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/BatchSpliterator.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      if (nextMapped != null) {
+        continue;
+      }
+      if (sourceEof) {
+        return false;
+      }
+
+      while (true) {
+        if (!source.tryAdvance(this::setNextSource)) {
```
dimas-b, line 82: WDYT about making this `Spliterator` batch `SRC` into `List<SRC>` and then we'd do `Stream.flatMap` from `List<SRC>` to `DST` (maybe in a static utility method)? Would it simplify the code?

dimas-b (commented):
Next round of review... I'm progressing through this code, but slowly :) Will resume on Mon.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -832,63 +761,525 @@ protected Hash casOpLoop(
...
+            .map(ReferenceNames::getRefNamesList)
+            .flatMap(List::stream)
+            // The named references splits may contain duplicate names, which is probably very, very
+            // rare, but still possible.
+            .distinct()
+            .spliterator();
+
+    BatchSpliterator<String, NamedReference> batchSpliterator =
+        new BatchSpliterator<>(
+            25, allNames, batch -> fetchNamedReference(ctx, batch).spliterator());
```
dimas-b, line 1191: Make `25` configurable?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
snazy: Hm - it's only used to identify the initial hash, there's no pagination.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/BatchSpliterator.java
snazy: Would be fine for the particular case, but I wanted to keep this class rather generic.
Collecting the source into a list means consuming the whole source, which could be expensive.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
snazy: Hrm - you're right - this can violate assumptions.

With global state, we guaranteed that the order in the ref-log matches the order in the commit-log for each individual branch. With this change, there's a chance that the order ref-log entries for commits on a branch does not match the order of commit-log entries.

I'm okay with violating the previous guarantee, as I do not see a legit use case for it. And that guarantee complicates things a lot. Thankfully we never documented that guarantee.

We do have some legit requirements like knowing the "lost HEADs" for GC (deleted references, re-assigned references) - but  that's mostly it.

I'd like to keep the split-calculation based on the reference name though. It's a best-effort approach though - but probably good enough.

WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TryLoopState.java
```diff
@@ -136,6 +144,27 @@ public void retry() throws ReferenceRetryFailureException {
     start();
   }
 
+  /**
+   * Similar to {@link #retry()}, but never throws a {@link ReferenceRetryFailureException} (unless
+   * a preceding call to {@link #retry()} already raised it) and always uses the initial lower and
+   * upper bounds for the sleep duration (no backoff).
+   */
+  public void retryEndless() throws ReferenceRetryFailureException {
```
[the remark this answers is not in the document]
snazy: The rationale here was that the "original try-loop-state" could already be "near its end" (like nearly timed out or nearly at the max number of retries). While the "original operation" could take relatively long, the ref-log-entry operation is rather quick (fetch stripe parents, write ref-log entry, try CAS on stripe parents), so my initial thought was to just use no backoff for that. But now it doesn't feel right, so I've changed it to reset the bounds and use backoff.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: I suppose we may want to cross-check the ref list table with ref heads as part of the database "maintenance" operation (followup PR). WDYT?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
snazy: Oh *that* paging. Well, that's "external" paging, yea - that's a "bit" more inefficient at the moment.
Either leave it as it is in the current PR and "deprecate the reflog for removal" - or have a better "external" paging state, which isn't too hard either, but adds more code to this PR.

dimas-b: With the goal of deprecating the reflog API soon, current impl. looks good enough for me.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
```diff
@@ -15,6 +15,32 @@
  */
 package org.projectnessie.versioned.persist.nontx;
 
+import org.immutables.value.Value;
 import org.projectnessie.versioned.persist.adapter.DatabaseAdapterConfig;
 
-public interface NonTransactionalDatabaseAdapterConfig extends DatabaseAdapterConfig {}
+public interface NonTransactionalDatabaseAdapterConfig extends DatabaseAdapterConfig {
+  int DEFAULT_REFERENCES_SEGMENT_SIZE = 250_000;
+  int DEFAULT_REF_LOG_STRIPES = 8;
```
dimas-b, line 23: nit: Why decrease from the original 16? Does it hurt to have more stripes?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/RegLogSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * {@link Spliterator} of {@link RefLog} entries using a variable number of ref-log stripes,
+ * according to {@link NonTransactionalDatabaseAdapterConfig#getRefLogStripes()} to read {@link
+ * RefLog}s from in parallel.
+ *
+ * <p>The next {@link RefLog} returned by this implementation is the next available from all stripes
+ * with the lowest {@link RefLog#getOperationTime()}.
+ */
+final class RegLogSpliterator extends AbstractSpliterator<RefLog> {
```
dimas-b, line 39: typo: `Reg` -> `Ref`?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/RegLogSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+
+    splits = refLogStripeFetcher.map(RefLogSplit::new).collect(Collectors.toList());
+
+    // The API for DatabaseAdapter.refLog(Hash) requires to throw a RefLogNotFoundException,
+    // if no RefLog for the initial hash exists. This loop tries to find the initial RefLog entry
+    // and throw
+    AtomicReference<RefLog> initial = new AtomicReference<>();
+    while (!initialHashSeen) {
+      if (!tryAdvance(initial::set)) {
+        break;
```
dimas-b, line 62: Could we have `tryAdvance` in the `while` condition and break of receiving the initial entry? I think the code would be more readable that way. WDYT (see also the comment on line 100)?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/RegLogSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+
+    if (!initialHashSeen) {
+      initialHashSeen = refLog.getRefLogId().equals(initialHash);
+    }
+
+    if (initialHashSeen) {
+      action.accept(refLog);
+    }
+
+    return true;
```
dimas-b, line 104: Returning `true` without calling the `action` appears to violate the javadoc contract of the `tryAdvance` method. If we reversed `if/while` conditions on lines 60-61 we could call the action unconditionally, I guess.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -832,63 +759,416 @@ protected Hash casOpLoop(
...
+    //    allows retrieving the paging token. The latter is probably more efficient, but requires
+    //    a bit more "verbose" code.
+    // 2. Faster initial stripe fetching
+    //    Instead of retrieving all initial pages sequentially, perform a bulk-fetch for all initial
+    //    pages. Since ref-log queries are not performance critical and also rare operations, it is
+    //    not urgent to implement this optimization.
+
+    Stream<Spliterator<RefLog>> stripeFetchers =
+        IntStream.range(-1, config.getRefLogStripes())
+            .mapToObj(stripe -> refLogStripeFetcher(ctx, stripe));
```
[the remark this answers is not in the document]
snazy: :+1:

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/ReferenceNamesSpliterator.java
```diff
@@ -0,0 +1,65 @@
...
+  ReferenceNamesSpliterator(IntFunction<List<ReferenceNames>> fetchReferenceNames) {
+    super(Long.MAX_VALUE, Spliterator.NONNULL);
+    this.fetchReferenceNames = fetchReferenceNames;
+  }
+
+  @Override
+  public boolean tryAdvance(Consumer<? super ReferenceNames> action) {
+    if (segment >= offset + segments.size()) {
+      offset = segment;
+      segments = fetchReferenceNames.apply(offset);
```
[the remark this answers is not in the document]
snazy: `fetchReferenceNames` takes the segment number to fetch and returns (at least) that segment - `fetchReferenceNames` fetches the requested segment + one (depending on the configuration) additional segment.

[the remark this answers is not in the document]
snazy: added docs to this class

dimas-b (commented):
Another batch of comments... getting closer to the finish line :)

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/ReferenceNamesSpliterator.java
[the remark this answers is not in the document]
dimas-b: got it. thx!

dimas-b (commented):
LGTM overall. I'm done with reviewing. Just a couple of remaining comments.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: Becase ref heads and the ref name list are upgraded separately, so there's a chance the get out of sync?

snazy: Ah - nah.

When a named-ref's added, the name list is updated first and then the ref-head - so it might happen, that a name exists in the name list but not as a ref-head.
When a named-ref's deleted, the ref head is deleted first, then removed from the names list.

So there's a chance that a name exists in the names list, but not as a ref-head. This isn't a problem, because the named-refs listing only returns those that exist as a ref head as well. For "point queries" with the name, the names list isn't at play.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
snazy: Was a bit concerned about 16+1 stripes to read from, which "explodes" to (16+1)*20 = 340 initial reads at once.

versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
```diff
@@ -252,22 +241,7 @@ protected void doCleanUpCommitCas(
...
-        batch.delete(dbInstance.getCfGlobalLog(), dbKey(h));
-      }
-      db.write(new WriteOptions(), batch);
-    } catch (RocksDBException e) {
-      throw new RuntimeException(e);
-    } finally {
-      lock.unlock();
-    }
-  }
+  protected void doCleanUpRefLogWrite(NonTransactionalOperationContext ctx, Hash refLogId) {}
```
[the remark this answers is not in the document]
snazy: Ooops

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
dimas-b: fair enough :+1:

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/RefLogSpliterator.java
```diff
@@ -0,0 +1,174 @@
...
+      return true;
+    }
+
+    RefLog refLog = tryNext();
+    if (refLog != null) {
+      action.accept(refLog);
+      return true;
+    }
+
+    return true;
```
dimas-b, line 111: I believe this should be `return false` since we did not find a new element.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -498,52 +482,61 @@ public Stream<Difference> diff(Hash from, Hash to, KeyFilterPredicate keyFilter)
...
       try {
-        long timeInMicros = commitTimeInMicros();
-        GlobalStatePointer dummyPointer =
-            GlobalStatePointer.newBuilder()
-                .setGlobalId(NO_ANCESTOR.asBytes())
-                .setGlobalLogHead(NO_ANCESTOR.asBytes())
-                .addGlobalParentsInclHead(NO_ANCESTOR.asBytes())
-                .setRefLogId(NO_ANCESTOR.asBytes())
+        for (int stripe = 0; stripe < config.getRefLogStripes(); stripe++) {
+          unsafeWriteRefLogStripe(
```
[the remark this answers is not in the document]
dimas-b: So why not remove this from the init code and rely on the first write to create the entry?

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: Ah - you wanna show that via a maintenance job. Yea, legit.

dimas-b (approved):

snazy:
Thanks for the thorough review!

PR #4286, opened 2022-05-20
Add Ingress example with Minikube

nastra:


XN137 (approved):

helm/nessie/README.md
```diff
@@ -83,6 +83,35 @@ aws_secret_access_key=YOURSECRETKEYDATA
...
 * Install Nessie Helm chart: `helm install nessie -n nessie-ns helm/nessie`
 
+### Ingress with Minikube
+
+This is broadly following the example from https://kubernetes.io/docs/tasks/access-application-cluster/ingress-minikube/
+
+* Start Minikube cluster: `minikube start`
+* Enable NGINX Ingress controller: `minikube addons enable ingress`
+* Verify Ingress controller is running: `kubectl get pods -n ingress-nginx`
+* Configure Nessie with the following ingress settings
```
snazy, line 93: Where / how would a user do this?

helm/nessie/README.md
```diff
@@ -83,6 +83,35 @@ aws_secret_access_key=YOURSECRETKEYDATA
...
+  ```
+* Create K8s Namespace: `kubectl create namespace nessie-ns`
+* Install Nessie Helm chart: `helm install nessie -n nessie-ns helm/nessie`
+* Verify that the IP address is set:
+  ```
+  $ kubectl get ingress
+  NAME     CLASS   HOSTS   ADDRESS        PORTS   AGE
+  nessie   nginx   *       192.168.49.2   80      4m35s
+  ```
+* Add `192.168.49.2 chart-example.local` to `/etc/hosts`
```
snazy, line 112: Is it the recommended practice to update `/etc/hosts`?

nastra: this is just for a local example. In prod you'd also have routing properly configured

helm/nessie/README.md
nastra: in the `values.yaml` of the helm chart

nastra: I'll make this more clear in the next PR

helm/nessie/README.md
snazy, line 112: Also: I suspect the IP address isn't fixed. Maybe add a note to use the IP from the output above (for dummies).

snazy, line 112: Replace with 
```
echo "192.168.49.2 chart-example.local" | sudo tee /etc/hosts
```

PR #4289, opened 2022-05-20
Lint & Test Nessie Helm chart

nastra:
This uses https://github.com/helm/chart-testing-action (which in turn is
based on https://github.com/helm/chart-testing) to lint and test Helm
charts.

The chart testing itself actually looks into `helm/nessie/ci` and
installs the different `*-values.yaml` charts in alphabetical order
(note that only files ending with `values.yaml` are considered for
testing)

Additionally, this commit addresses some issues that were reporting when
running `ct lint`, such as the missing maintainers field.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -106,12 +106,40 @@ jobs:
         run: mkdocs build
         working-directory: ./site
 
-  lint-helm:
-    name: Lint Helm chart
+  helm-testing:
+    name: Lint & Test Helm chart
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - name: Lint Helm
-        uses: WyriHaximus/github-action-helm3@v2
+      - name: Checkout
+        uses: actions/checkout@v2
         with:
-          exec: helm lint ./helm/nessie/
+          fetch-depth: 0
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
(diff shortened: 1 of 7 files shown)


.github/workflows/pull-request.yml
```diff
@@ -106,12 +106,40 @@ jobs:
...
+          python-version: 3.8
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@main
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --target-branch ${{ github.event.repository.default_branch }})
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
```
nastra, line 132: I was initially thinking that we'd run the helm tests only if there are any changes, but for now maybe let's just run them every time for a while just to get a feel for the stability of those tests

helm/nessie/Chart.yaml
```diff
@@ -14,3 +14,7 @@ keywords:
   - data lake
   - transactional catalog
   - git-like semantics
+maintainers:
+  - name: nastra
+  - name: snazy
+  - name: dimas-b
```
nastra, line 20: shall we add all people here? I just added a few to make the linter happy

snazy: all good

snazy (approved):

PR #4303, opened 2022-05-23
Efficiently identify referenced and unreferenced heads

snazy:
Implement an efficient way to retrieve all unreferenced HEAD commits.

As a side effect (and up for discussion whether the information's useful), it also returns
* all fork-points (commits that are the parent of more than one other commits)
* map of commit-ID to named-refs

The implementation scans all commits and identifies all "heads" (aka commits that are not the parent of another commit) and all fork-points (commits that are the parent of more than one other commits).
In a next step all named-references and their commit-logs are scanned to identify the HEADs that are unreferenced and those that are the HEADs of named references.

"Efficient" means the only somewhat expensive part is the scan of all commits, but the computation/identification of all HEADs and fork-points (commit with more than one "child") plus the differentiation of unreferenced & referenced HEADs is not that expensive. Plus: it's a server-side implementation and interested clients (maintenance + GC) could just ask for this information.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
@@ -42,6 +42,7 @@ public interface DatabaseAdapterConfig {
   int DEFAULT_RETRY_INITIAL_SLEEP_MILLIS_LOWER = 5;
   int DEFAULT_RETRY_INITIAL_SLEEP_MILLIS_UPPER = 25;
   int DEFAULT_RETRY_MAX_SLEEP_MILLIS = 75;
+  long DEFAULT_ASSUMED_WALL_CLOCK_DRIFT_MICROS = 5_000_000L;
 
   /**
    * A free-form string that identifies a particular Nessie storage repository.
@@ -216,5 +217,11 @@ public interface DatabaseAdapterConfig {
   @Value.Default
   default int getParentsPerRefLogEntry() {
     return DEFAULT_PARENTS_PER_REFLOG_ENTRY;
+  }
+
+  /** The assumed wall-clock drift between multiple Nessie instances in microseconds. */
+  @Value.Default
+  default long getAssumedWallClockDriftMicros() {
+    return DEFAULT_ASSUMED_WALL_CLOCK_DRIFT_MICROS;
   }
 }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
@@ -1,0 +1,194 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.persist.adapter;
+
+import com.google.common.annotations.Beta;
+import com.google.protobuf.ByteString;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
+import org.agrona.collections.Hashing;
+import org.agrona.collections.Object2ObjectHashMap;
+import org.agrona.collections.ObjectHashSet;
+import org.projectnessie.versioned.GetNamedRefsParams;
+import org.projectnessie.versioned.Hash;
+import org.projectnessie.versioned.NamedRef;
+import org.projectnessie.versioned.ReferenceInfo;
+import org.projectnessie.versioned.ReferenceNotFoundException;
+
+@Beta
+public final class ReferencesUtil {
+  private final DatabaseAdapter databaseAdapter;
+  private final DatabaseAdapterConfig config;
+
+  private ReferencesUtil(DatabaseAdapter databaseAdapter, DatabaseAdapterConfig config) {
+    this.databaseAdapter = databaseAdapter;
+    this.config = config;
+  }
+
+  public static ReferencesUtil forDatabaseAdapter(
+      DatabaseAdapter databaseAdapter, DatabaseAdapterConfig config) {
+    return new ReferencesUtil(databaseAdapter, config);
+  }
+
+  private static <K, V> Map<K, V> newOpenAddressingHashMap() {
+    return new Object2ObjectHashMap<>(16, Hashing.DEFAULT_LOAD_FACTOR, false);
+  }
+
+  private static <T> Set<T> newOpenAddressingHashSet(int initialCapacity) {
+    return new ObjectHashSet<>(initialCapacity, Hashing.DEFAULT_LOAD_FACTOR, false);
+  }
+
+  private static <T> Set<T> newOpenAddressingHashSet(Set<T> source) {
+    ObjectHashSet<T> copy = new ObjectHashSet<>(source.size(), Hashing.DEFAULT_LOAD_FACTOR, false);
+    if (source instanceof ObjectHashSet) {
+      copy.addAll((ObjectHashSet<T>) source);
+    } else {
+      copy.addAll(source);
+    }
+    return copy;
+  }
+
+  private static <T> Set<T> newOpenAddressingHashSet() {
+    return newOpenAddressingHashSet(ObjectHashSet.DEFAULT_INITIAL_CAPACITY);
+  }
+
+  /**
+   * Identifies all heads and fork-points.
+   *
+   * <ul>
+   *   <li>"Heads" are commits that are not referenced by other commits.
+   *   <li>"Fork points" are commits that are the parent of more than one other commit. Knowing
+   *       these commits can help to optimize the traversal of commit logs of multiple heads.
+   * </ul>
+   *
+   * @param expectedCommitCount it is recommended to tell the implementation the total number of
+   *     commits in the Nessie repository
+   */
+  public HeadsAndForkPoints identifyAllHeadsAndForkPoints(int expectedCommitCount) {
+    // Using open-addressing implementation here, because it's much more space-efficient than
+    // java.util.HashSet.
+    Set<Hash> parents = newOpenAddressingHashSet(expectedCommitCount);
+    Set<Hash> heads = newOpenAddressingHashSet();
+    Set<Hash> forkPoints = newOpenAddressingHashSet();
+
+    // Need to remember the time when the identification started, so that a follow-up
+    // identifyReferencedAndUnreferencedHeads() knows when it can stop scanning a named-reference's
+    // commit-log. identifyReferencedAndUnreferencedHeads() has to read up to the first commit
+    // _before_ this timestamp to not commit-IDs as "unreferenced".
+    //
+    // Note: keep in mind, that scanAllCommitLogEntries() returns all commits in a
+    // non-deterministic order. Example: if (at least) two commits are added to a branch while this
+    // function is running, the original HEAD of that branch could otherwise be returned as
+    // "unreferenced".
+    long scanStartedAtInMicros = config.currentTimeInMicros();
+
+    // scanAllCommitLogEntries() returns all commits in no specific order, parents may be scanned
+    // before or after their children.
+    try (Stream<CommitLogEntry> scan = databaseAdapter.scanAllCommitLogEntries()) {
+      scan.forEach(
+          entry -> {
+            Hash parent = entry.getParents().get(0);
+            if (!parents.add(parent)) {
+              // If "parent" has already been added to the set of parents, then it must be a
+              // fork point.
+              forkPoints.add(parent);
+            } else {
+              // Commits in "parents" that are also contained in "heads" cannot be HEADs.
+              // This can happen because the commits are scanned in "random order".
+              // Should do this here to prevent the "heads" set from becoming unnecessarily big.
+              heads.remove(parent);
+            }
+
+            Hash commitId = entry.getHash();
+            if (!parents.contains(commitId)) {
+              // If the commit-ID is not present in "parents", it must be a HEAD
+              heads.add(commitId);
+            }
+          });
+    }
+
+    // Commits in "parents" that are also contained in "heads" cannot be HEADs.
+    // This can happen because the commits are scanned in "random order".
+    // TODO this 'removeIf' should not actually remove anything, because it should already be
+    //  properly handled above.
+    heads.removeIf(parents::contains);
+
+    return HeadsAndForkPoints.of(heads, forkPoints, scanStartedAtInMicros);
+  }
+
+  /**
+   * Identifies unreferenced heads and heads that are part of a named reference.
+   *
+   * <p>Requires the output of {@link #identifyAllHeadsAndForkPoints(int)}.
+   */
+  public ReferencedAndUnreferencedHeads identifyReferencedAndUnreferencedHeads(
+      HeadsAndForkPoints headsAndForkPoints) throws ReferenceNotFoundException {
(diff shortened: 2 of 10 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -825,6 +827,119 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
...
+
+                  Hash commitId = entry.getHash();
+
+                  if (referenced.containsKey(commitId)) {
+                    // Already saw this commit-ID, can break
+                    break;
+                  }
+
+                  if (heads.contains(commitId)) {
+                    unreferenced.remove(entry.getHash());
```
dimas-b, line 915: Could we keep track of unreferenced heads by having a log of "remove branch/tag" operations?... that is to avoid scanning?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -825,6 +827,119 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
...
+    long scanStartedAtInMicros = this.commitTimeInMicros();
+
+    try (Stream<CommitLogEntry> scan = scanAllCommitLogEntries()) {
+      scan.forEach(
+          entry -> {
+            Hash parent = entry.getParents().get(0);
+            if (!commits.add(parent)) {
+              // If "parent" has already been added to the set of commits, then it must be a
+              // common ancestor.
+              commonAncestors.add(parent);
```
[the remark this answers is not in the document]
snazy: > I'd rename it to "fork points" maybe?... just for the sake of clarity. In my mind a parent of a common ancestor is also a common ancestor thinking

We're using the term "common ancestor" elsewhere already (named-ref information, merges). Not sure if adding another term would be good.

> Might as well compute the number of branches stemming from this commit, I guess :)

That's what the other function `identifyReferencedAndUnreferencedHeads` does ;)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: That was what the ref-log was doing.
I thought about another "log" for those operations as well, but it's tricky and would eventually become a "single point of contention" as well. Or using anti-patterns for no-sql databases (think: time-series data)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -700,6 +702,18 @@ protected List<RefLog> doFetchPageFromRefLog(
     return fetchPage(client.getRefLog(), hashes, ProtoSerialization::protoToRefLog);
   }
 
+  @Override
+  protected Stream<CommitLogEntry> doScanAllCommitLogEntries(NonTransactionalOperationContext c) {
+    Bson idPrefixFilter = Filters.eq(ID_REPO_PATH, repositoryId);
+    FindIterable<Document> iter =
+        client.getCommitLog().find(idPrefixFilter, Document.class).batchSize(25);
```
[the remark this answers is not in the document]
snazy: done

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -825,6 +827,119 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
...
+    // non-deterministic order. Example: if (at least) two commits are added to a branch while this
+    // function is running, the original HEAD of that branch could otherwise be returned as
+    // "unreferenced".
+    long scanStartedAtInMicros = this.commitTimeInMicros();
+
+    try (Stream<CommitLogEntry> scan = scanAllCommitLogEntries()) {
+      scan.forEach(
+          entry -> {
+            Hash parent = entry.getParents().get(0);
+            if (!commits.add(parent)) {
```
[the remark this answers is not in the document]
snazy: Basically the "last" concern of this change.

The object structure for `commits` is:
```
HashSet
-> HashMap
   -> HashMap.Node[]
       -> Hash
           -> ByteString.LiteralByteString (usually that type)
               -> byte[]
```
It's a rather non-cheap structure when there are many commits.

There are two things that can be done here:
1. Optimize `Hash` heap pressure - inline the `byte[]` directly into `Hash`
2. Use a more efficient (computation + heap pressure) structure for the `HashSet commits`

These are separate exercises though ;)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: Sorry, I was not clear, I meant keeping a log of "obsolete" HEADs attached to the normal branch HEAD. For each branch we'd have one active HEAD and a chain on previous HEADs (for old deleted branched of the same name).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
dimas-b: In other code places the "common" ancestor is always in the context of two references, so the meaning is quite clear (at least for me :) ). Here, by contrast we're finding common ancestors even among unreferenced HEADs, so to me the meaning of "common" in this context is not as clear, because the entities that "share" a particular "common" commit themselves vary from case to case :thinking:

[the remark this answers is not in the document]
dimas-b: As a side benefit we'd get the previous ref name for unreferenced HEADs, which might be relevant for access control and historical investigation (by end users).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: As a side benefit we'd get the previous ref name for unreferenced HEADs, which might be relevant for access control and historical investigation (by end users).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -38,6 +38,7 @@ public interface DatabaseAdapterConfig {
   int DEFAULT_RETRY_INITIAL_SLEEP_MILLIS_LOWER = 5;
   int DEFAULT_RETRY_INITIAL_SLEEP_MILLIS_UPPER = 25;
   int DEFAULT_RETRY_MAX_SLEEP_MILLIS = 75;
+  long DEFAULT_ASSUMED_WALL_CLOCK_DRIFT_MICROS = 5 * 1000L * 1000L;
```
dimas-b, line 41: nit: `5_000_000L`?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -825,6 +827,120 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
...
+              try {
+                ctx.close();
+              } catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            });
+  }
+
+  @Override
+  public HeadsAndForkPoints identifyAllHeadsAndForkPoints(int expectedCommitCount) {
```
dimas-b, line 845: This functionality appears to be kind of external to `DatabaseAdapter`. Would it make sense to move this code to a utility class?.. or maybe a separate CDI bean? Do we expect to have adapter-specitic logic in this method the future? :thinking:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -825,6 +827,120 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
 
   protected abstract CommitLogEntry doFetchFromCommitLog(OP_CONTEXT ctx, Hash hash);
 
+  @Override
+  public Stream<CommitLogEntry> scanAllCommitLogEntries() {
```
dimas-b, line 831: WDYT about (optionally) providing a chronologically ordered stream of commits? That would reduce the amount of in-memory state, I guess (pruning non-heads sooner). Storage would have to maintain an index, which may not always be feasible, but perhaps in some cases?

snazy: Doubt it could work in all cases - like these:
* (Wall) clock drift could still let children appear before parents (parent create-timestamp < child create-timestamp)
* Indexes in non-tx databases are rather "hopefully consistent"

It's a tricky issue here and I'm not sure there's a way to get around it. Making that `Set<Hash> parents` a bloom filter is also not really going to work, because it will result in false-positives: non-head commits appearing as heads in the function's result. Which in turn would result in false-positives from `identifyReferencedAndUnreferencedHeads()`: returning those as unreferenced heads, which is risky, beacuse Nessie GC could use different expiry mechanisms, expire content quicker, leading to unavailable user data.

The "heap math" for the `Set<Hash> parents` is this:
* It's an open addressing hash set, with a load-factor of .65.
* Each `Hash` (of a 256 bit hash) has an on-heap size of 48 bytes: (up to) 16 bytes Java object header + 32 bytes for the 256 bits.
* The `parents` set size for e.g. 650000 objects would be an `Object[1048576]` (~ 8M heap) plus the `Hash`s (~32M) - so it's not too bad.
* The remaining objects are either short lived or small-ish collections.

Maybe we can go with this approach and see how big repositories will actually become?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: moved out

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: Renamed to "fork point"

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: Using an open addressing hash set SGTM :+1:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
```diff
@@ -0,0 +1,192 @@
...
+            if (!parents.contains(commitId)) {
+              // If the commit-ID is not present in "parents", it must be a HEAD
+              heads.add(commitId);
+            }
+          });
+    }
+
+    // Commits in "parents" that are also contained in "heads" cannot be HEADs.
+    // This can happen because the commits are scanned in "random order".
+    heads.removeIf(parents::contains);
```
dimas-b, line 127: Is it really possible to get false-positive heads at this point? We remove each new parent from `heads` and we only add to `heads` if the hash is not in `parents` (above)... it looks like false-positive are not possible.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: In theory, we could implement our custom open-addressing-set for 256-bit hashes, which would, based on the load-factor of .65, roughly reduce the heap pressure by the same amount that the `Object[]` occupies - and also have a better memory access pattern. It's an option - but likely none for yet.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -189,7 +189,8 @@ public Hash noAncestorHash() {
   // /////////////////////////////////////////////////////////////////////////////////////////////
 
   /** Returns the current time in microseconds since epoch. */
-  protected long commitTimeInMicros() {
+  @Override
+  public long commitTimeInMicros() {
```
dimas-b, line 193: maybe move the whole method to the config class?

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
```diff
@@ -23,6 +23,7 @@ public interface NonTransactionalDatabaseAdapterConfig extends DatabaseAdapterCo
   int DEFAULT_REF_LOG_STRIPES = 8;
   int DEFAULT_REFERENCES_SEGMENT_PREFETCH = 1;
   int DEFAULT_REFERENCE_NAMES_BATCH_SIZE = 25;
+  int DEFAULT_COMMIT_LOG_SCAN_PREFETCH = 25;
```
dimas-b, line 26: Why not use a large number? It looks like its used only in the context of the unreferenced heads search which is already a memory-heavy operation and will scan all entries, so using large batches may be preference, I guess.

dimas-b (commented):
LGTM overall :+1: minor comments

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
snazy: I'd rather be on the safe side here - we can experiment with it in "big scenarios" and bump the default based on the experiences. WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
snazy: Yea - I think the same (because it's handled in the loop above), because I stepped through here in the debugger and this removeIf never did anything.

Just not sure whether there's maybe a case where it could be. I can add a TODO to visit again. WDYT?

dimas-b: Yeah, a TODO is fine. The number of heads is not that great anyway.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapterConfig.java
dimas-b: sgtm

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
dimas-b: I suppose we can have a test where we purposefully execute all input permutations. That should validate the head identification logic even with a small dataset. I guess we really need a few liner segments and a few fork points (with nesting).

dimas-b (approved):

PR #4307, opened 2022-05-23
Expose merge-response via public API

snazy:
Exposes the merge & transplant responses and dry-run functionality
introduced by #4273 to the public API.

The changes are backwards compatible, old clients ignore the
response, new clients (against older servers) "just" return `null`
as the response. Compatibility tests added as well. However, if new
clients setting the new merge/transplant request properties, old
servers *will* complain, because the new request properties are
unknown.

Side note: `MergeResponse` cannot use a `Map<ContentKey, ...>` as
the internal `MergeResult` object does, because that requires the
implementation of custom "map key (de)serializers" for `ContentKey`.

Other than that, the change is rather simple, but sadly boilerplate.


compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
```diff
@@ -156,4 +160,73 @@ public void namespaceWithProperties() throws NessieNotFoundException, NessieConf
...
+
+    if (nessieWithMergeResponse()) {
+      assertThat(response).isNotNull();
+    } else {
+      assertThat(response).isNull();
+    }
+  }
+
+  boolean nessieWithMergeResponse() {
+    return version.compareTo(Version.parseVersion("0.30.0")) > 0;
```
nastra, line 230: version.isGreaterThan(...)

model/src/main/java/org/projectnessie/model/BaseMergeTransplant.java
```diff
@@ -44,11 +44,24 @@ public interface BaseMergeTransplant {
...
 
   @Nullable
+  @JsonInclude(Include.NON_NULL)
   Boolean isFetchAdditionalInfo();
 
+  /**
+   * When set to {@code true}, the {@link org.projectnessie.api.TreeApi#mergeRefIntoBranch(String,
+   * String, Merge)} and {@link org.projectnessie.api.TreeApi#transplantCommitsIntoBranch(String,
+   * String, String, Transplant)} operations will return {@link MergeResponse} object when a content
+   * based conflict cannot be reolved, instead of throwing a {@link
```
nastra, line 58: typo: reolved

nastra (commented):
overall LGTM, just a few nits

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
```diff
@@ -408,7 +421,7 @@ void transplantCommitsIntoBranch(
     @APIResponse(responseCode = "404", description = "Ref doesn't exists"),
     @APIResponse(responseCode = "409", description = "update conflict")
   })
-  void mergeRefIntoBranch(
+  MergeResponse mergeRefIntoBranch(
```
dimas-b, line 424: nit: Could you add OpenApi examples for the new output?

model/src/main/java/org/projectnessie/model/BaseMergeTransplant.java
```diff
@@ -44,11 +44,24 @@ public interface BaseMergeTransplant {
...
   @Nullable
+  @JsonInclude(Include.NON_NULL)
   Boolean isFetchAdditionalInfo();
 
+  /**
+   * When set to {@code true}, the {@link org.projectnessie.api.TreeApi#mergeRefIntoBranch(String,
+   * String, Merge)} and {@link org.projectnessie.api.TreeApi#transplantCommitsIntoBranch(String,
+   * String, String, Transplant)} operations will return {@link MergeResponse} object when a content
+   * based conflict cannot be resolved, instead of throwing a {@link
+   * org.projectnessie.error.NessieReferenceConflictException}.
```
dimas-b, line 59: tangential: it feels that we may want to have a dedicated exception for content conflicts in merges rather than `NessieReferenceConflictException`... but it's beyond the scope of this PR for sure.

dimas-b (approved):

model/src/main/java/org/projectnessie/api/http/HttpTreeApi.java
nastra: I wanted to comment on that as well and forgot, so +1 on adding an example

snazy: Just realized that there are a bunch of examples that aren't linked anywhere. WIll do it in a follow-up.

PR #4319, opened 2022-05-24
Testing: allow more concrete types than just `DatabaseAdapter.class`

snazy:
`DatabaseAdapterExtension` only allows the basic types `DatabaseAdapter`
and `VersionStore` as field or parameter types. This change allows
using more concrete types like `AbstractDatabaseAdpater` so the need
to manually cast the types becomes obsolete.

### DIFF
--- a/versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
+++ b/versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
@@ -230,9 +230,9 @@ public class DatabaseAdapterExtension
     }
 
     Object assign;
-    if (type.isAssignableFrom(DatabaseAdapter.class)) {
+    if (DatabaseAdapter.class.isAssignableFrom(type)) {
       assign = databaseAdapter;
-    } else if (type.isAssignableFrom(VersionStore.class)) {
+    } else if (VersionStore.class.isAssignableFrom(type)) {
       VersionStore<?, ?, ?> store = createStore(databaseAdapter, storeWorker);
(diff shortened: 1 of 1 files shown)


versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/extension/DatabaseAdapterExtension.java
```diff
@@ -230,9 +230,9 @@ private Object resolve(
     }
 
     Object assign;
-    if (type.isAssignableFrom(DatabaseAdapter.class)) {
+    if (DatabaseAdapter.class.isAssignableFrom(type)) {
       assign = databaseAdapter;
```
dimas-b, line 234: not sure how it works :thinking: We're returning `databaseAdapter`, which would be assigned to a variable of type `type`, right? In this case the old comparison was more correct, I think.

Did you mean `type.isAssignableFrom(databaseAdapter.getClass())`?

snazy: This one is to distinguish `DatabaseAdapter.class` from `VersionStore.class` - the other one is below.

dimas-b: Sorry, I was confused, please ignore my comment.

snazy: `isAssignableFrom` IS confusing

dimas-b (approved):

PR #4331, opened 2022-05-25
CI: Ignore OWASP errors

snazy:
OWASP started to fail on May 25, 2022.


nastra (approved):
so once it starts failing, how are we going to notice that it's actually broken?

PR #4333, opened 2022-05-25
Store and retrieve merge-parent

snazy:
Merge operations store the merged commit ID as an additional parent.

This change introduces a new field `additionalParents`, which is a
list of commit-IDs. This list is usually empty, except for
squashing merge commits, that are not single-commit fast-forward
merges.

Transplant operations do _not_ store an "additional parent".

Merge operations that are configured to keep separate commits do
_not_ store an "additional parent" as well, because this is effectively
the same as transplanting (cherry-picking).

This change _prepares_ the OpenAPI type `LogEntry` with a new attribute
`additionalParents`. The attribute will _not_ be populated though to
not break older clients. Otherwise, older clients will fail to
deserialize commit log entries.

This change however does return the first additional parent as a
property in `CommitMeta`, with the property key `_merge_parent`, so
clients that know about the meaning of that property, can handle it.

### DIFF
--- a/versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/adapter/serialize/TestSerialization.java
+++ b/versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/adapter/serialize/TestSerialization.java
@@ -434,7 +434,7 @@ class TestSerialization {
         System.nanoTime() / 1000L,
         randomHash(),
         randomLong(),
-        IntStream.range(0, 20).mapToObj(i -> randomHash()).collect(Collectors.toList()),
+        IntStream.range(5, 20).mapToObj(i -> randomHash()).collect(Collectors.toList()),
         randomBytes(256),
         IntStream.range(1, 1 + ThreadLocalRandom.current().nextInt(7))
             .mapToObj(i -> KeyWithBytes.of(randomKey(), randomId(), (byte) 1, randomBytes(32)))
--- a/model/src/main/java/org/projectnessie/model/CommitMeta.java
+++ b/model/src/main/java/org/projectnessie/model/CommitMeta.java
@@ -46,6 +46,14 @@ import org.immutables.value.Value;
 @JsonDeserialize(as = ImmutableCommitMeta.class)
 public abstract class CommitMeta {
 
+  /**
+   * Key of the {@link #getProperties() property} that indicates the merged commit ID, if the commit
+   * represents a merge commit.
+   *
+   * <p>If the commit is the result of a merge operation, the properties map <em>may</em> contain
+   * the key {@value #MERGE_PARENT_PROPERTY}, if the merge was performed with a Nessie server
+   * version, after 0.30.0, that persists the merged reference.
+   */
   public static final String MERGE_PARENT_PROPERTY = "_merge_parent";
 
   /**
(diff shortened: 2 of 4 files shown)


model/src/main/java/org/projectnessie/model/LogResponse.java
```diff
@@ -49,6 +51,9 @@ static ImmutableLogEntry.Builder builder() {
     @NotNull
     CommitMeta getCommitMeta();
 
+    @JsonInclude(Include.NON_EMPTY)
```
snazy, line 54: This one breaks compatibility w/ older clients, when the client retrieves a commit-log-entry that has an additional parent (a merge-commit).

dimas-b: Can we make JSON parsing more lenient on the client?

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/adapter/serialize/TestSerialization.java
```diff
@@ -434,43 +434,23 @@ static CommitLogEntry createEntry() {
...
+            .mapToObj(i -> KeyWithBytes.of(randomKey(), randomId(), (byte) 1, randomBytes(32)))
+            .collect(Collectors.toList()),
         Arrays.asList(randomKey(), randomKey(), randomKey(), randomKey(), randomKey()),
         42,
         KeyList.of(
             IntStream.range(0, 20)
                 .mapToObj(i -> createKeyListEntry())
                 .collect(Collectors.toList())),
-        IntStream.range(0, 20).mapToObj(i -> randomHash()).collect(Collectors.toList()));
+        IntStream.range(0, ThreadLocalRandom.current().nextInt(20))
```
dimas-b, line 448: Using random test input makes me uneasy :) ... could this be converted to a parameterized test to cover all meaningful ranges?

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/adapter/serialize/TestSerialization.java
```diff
@@ -434,43 +434,23 @@ static CommitLogEntry createEntry() {
...
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash(),
-            randomHash()),
+        IntStream.range(0, 20).mapToObj(i -> randomHash()).collect(Collectors.toList()),
```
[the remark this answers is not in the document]
snazy: yes

versioned/persist/serialize/src/test/java/org/projectnessie/versioned/persist/adapter/serialize/TestSerialization.java
snazy: Let's do that in another PR.

snazy:
> When transplanting, would it make sense to still store some information about the operation parameters as commit properties? The main use case would be historical analysis / trouble-shooting by a human being. That would be similar to how git generates the default commit message on `cherry-pick`, for example?... This would be separate from this PR, of course.

Yes, I think, the commit messages for those operations could be more informative. Not sure about Git's `cherry-pick` though, because that literally copies the original commit message.

dimas-b (approved):

PR #4350, opened 2022-05-27
Add Rolling-Upgrade test suite

### LINKED ISSUE #4348: Rolling upgrade tests
Nessie already has test suites to test against older client versions, older server versions and non-rolling-upgrades.

Having rolling upgrade tests is really helpful.

snazy:
Adds a new integration-tests for rolling upgrades including the
"machinery" to run rolling-upgrade tests.

That "machinery", implemented in `RollingUpgradesExtension` uses
MongoDB as the backend database. Using RocksDB doesn't work, because
rolling-upgrades requires that two server instances access the same
database - and RocksDB doesn't allow that.

Tests using `RollingUpgradesExtension` get two instances of
`NessieApiV1` - one configured against the "old" Nessie version and
the other against the "current" (in-tree) Nessie version.

`RollingUpgradesExtension` reinitializes the Nessie repository for
each combinations of Nessie versions.

The new `ITRollingUpgrades` tests a couple scenarios, against both
the old and new versions:
* get the default branch
* create new branch
* verify created branches
* commit a new table
* verify contents

`ITRollingUpgrades` is configured to require at least Nessie 0.31.0,
which is not yet released, because any rolling upgrade to current
in-tree will fail, because it is not safe for a rolling-upgrade
since #4234.

Fixes #4348

### DIFF
--- a/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgradeScenario.java
+++ b/compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgradeScenario.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.tools.compatibility.tests;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
+
+import com.google.common.collect.Maps;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.MethodOrderer;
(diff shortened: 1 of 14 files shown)


compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgradeScenario.java
```diff
@@ -0,0 +1,144 @@
...
+
+  static Stream<Arguments> verifyContent() {
+    return createdContent.entrySet().stream().map(e -> arguments(e.getKey(), e.getValue()));
+  }
+
+  @ParameterizedTest
+  @MethodSource("verifyContent")
+  @Order(109)
+  void verifyContentInOld(ContentKey key, IcebergTable table) throws Exception {
+    Map<ContentKey, Content> contents = apiOld.getContent().key(key).refName("main").get();
```
dimas-b, line 131: `apiOld` could probably be a parameter.

compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgrades.java
```diff
@@ -0,0 +1,209 @@
...
+    Branch defaultBranch = api.getDefaultBranch();
+    assertThat(defaultBranch).extracting(Branch::getName).isEqualTo("main");
+  }
+
+  @Test
+  @Order(101)
+  void defaultBranchInNew() throws Exception {
+    Branch defaultBranch = apiCurrent.getDefaultBranch();
+    assertThat(defaultBranch).extracting(Branch::getName).isEqualTo("main");
+  }
```
dimas-b, line 92: It feels like we could probably parameterize these test by `TargetVersion` and maybe use a `ParameterResolver` to inject the right API impl based on the enum value. WDYT?

dimas-b (approved):

compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgrades.java
snazy: Using `@NessieAPI` as a paramter already works. These tests (and the following) should work in a defined order though.

PR #4414, opened 2022-06-07
Add check-content Nessie CLI command

dimas-b:
* `check-content` validates whether Nessie content for active keys is readable.

* Minor refactoring of test code.

* Add testing infrastructure for Nessie CLI commands.

### DIFF
--- a/servers/quarkus-cli/pom.xml
+++ b/servers/quarkus-cli/pom.xml
@@ -66,6 +66,11 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-junit5</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
+++ b/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
@@ -1,0 +1,204 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.quarkus.cli;
+
+import com.google.protobuf.ByteString;
+import java.io.PrintWriter;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
+import org.projectnessie.versioned.GetNamedRefsParams;
+import org.projectnessie.versioned.Hash;
+import org.projectnessie.versioned.Key;
+import org.projectnessie.versioned.ReferenceInfo;
+import org.projectnessie.versioned.ReferenceNotFoundException;
+import org.projectnessie.versioned.StoreWorker;
+import org.projectnessie.versioned.persist.adapter.ContentAndState;
+import org.projectnessie.versioned.persist.adapter.KeyFilterPredicate;
+import org.projectnessie.versioned.persist.adapter.KeyListEntry;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "check-content",
+    mixinStandardHelpOptions = true,
+    description = "Check content readability of active keys.")
+public class CheckContent extends BaseCommand {
+
+  @CommandLine.Option(
+      names = {"-s", "--separator"},
+      defaultValue = ": ",
+      description = "Output field separator.")
+  private String separator;
+
+  @CommandLine.Option(
+      names = {"-k", "--key"},
+      description =
+          "Content key element (zero or more). If not set, all current keys will be checked.")
+  private List<String> contentKey;
+
+  @CommandLine.Option(
+      names = {"-c", "--show-content"},
+      description = "Print content details for each valid key.")
+  private boolean showContent;
+
+  @CommandLine.Option(
+      names = {"-B", "--batch"},
+      defaultValue = "25",
+      description = "The max number of keys to load at the same time.")
+  private int batchSize;
+
+  @CommandLine.Option(
+      names = {"-W", "--worker-class"},
+      defaultValue = "org.projectnessie.server.store.TableCommitMetaStoreWorker",
+      description = "The class name of the Nessie Store Worker (internal).")
+  private String workerClass;
+
+  @CommandLine.Option(
+      names = {"-r", "--ref"},
+      description = "Reference name to use (default branch, if not set).")
+  private String ref;
+
+  @CommandLine.Option(
+      names = {"-H", "--hash"},
+      description = "Commit hash to use (defaults to the HEAD of the specified reference).")
+  private String hash;
+
+  @CommandLine.Option(
+      names = {"-v", "--verbose"},
+      description = "Print extra error information to STDERR.")
+  private boolean verbose;
+
+  @CommandLine.Option(
+      names = {"-E", "--error-only"},
+      description = "Print status information only for keys with errors")
+  private boolean errorOnly;
+
+  private final AtomicInteger errorDetected = new AtomicInteger();
+
+  @Override
+  public Integer call() throws Exception {
+    warnOnInMemory();
+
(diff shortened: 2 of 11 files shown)


servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,204 @@
...
+      names = {"-s", "--separator"},
+      defaultValue = ": ",
+      description = "Output field separator.")
+  private String separator;
+
+  @CommandLine.Option(
+      names = {"-k", "--key"},
+      description =
+          "Content key element (zero or more). If not set, all current keys will be checked.")
+  private List<String> contentKey;
```
snazy, line 52: Nit: `contentKeyElements`?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,204 @@
...
+  @CommandLine.Option(
+      names = {"-s", "--separator"},
+      defaultValue = ": ",
+      description = "Output field separator.")
+  private String separator;
+
+  @CommandLine.Option(
+      names = {"-k", "--key"},
+      description =
+          "Content key element (zero or more). If not set, all current keys will be checked.")
```
snazy, line 51: Nit: `elements`?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,204 @@
...
+      names = {"-v", "--verbose"},
+      description = "Print extra error information to STDERR.")
+  private boolean verbose;
+
+  @CommandLine.Option(
+      names = {"-E", "--error-only"},
+      description = "Print status information only for keys with errors")
+  private boolean errorOnly;
+
+  private final AtomicInteger errorDetected = new AtomicInteger();
```
snazy, line 91: Looks like the number of errors is never used. Maybe `AtomicBoolean`?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    if (error == null && errorOnly) {
+      return;
+    }
+
+    PrintWriter out = spec.commandLine().getOut();
+    out.printf(error == null ? "OK" : "ERROR");
+    out.print(separator);
+    out.print(key);
```
snazy, line 199: Instead of just `key` maybe `UriUtil.toPathString(key.getElements())`

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,204 @@
...
+    if (error != null) {
+      errorDetected.incrementAndGet();
+    }
+
+    if (error == null && errorOnly) {
+      return;
+    }
+
+    PrintWriter out = spec.commandLine().getOut();
+    out.printf(error == null ? "OK" : "ERROR");
```
snazy, line 197: Wonder whether we could dump the whole information as a stream of JSON objects. I.e. create a `JsonGenerator` and open (and later close) a JSON array in `call()`  and dump JSON objects here. Could also properly dump the `Content` object as JSON as well. WDYT?

Not sure what to do with the stack trace dumped in the `report(key, Throwable)` - maybe dump that as JSON as well?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
snazy: Hm - can we use `UriUtil` here to make it a bit more user friedly?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: good point.

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: How about sending JSON output to a separate file, e.g. `check-content -o /tmp/results.json`?

snazy: WFM

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: I'll add a separate option for the encoded key form. In my experience using multiple options (one per key element) is sometimes very handy, e.g. when the user already has a key that is complex and not encoded.

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: Added `--summary` for this data.

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: On the second thought, I guess using encoded keys may not be very convenient in this situation. Most (if not all) other Nessie CLI tools produce unencoded keys (including this command that now produce JSON arrays for keys). So for an end user I think it would be more convenient to copy-paste that data into individual `-k` options. WDYT?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: I guess encoding is not required now that the output is JSON.

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,264 @@
...
+  @Override
+  public Integer call() throws Exception {
+    warnOnInMemory();
+
+    if (outputSpec != null) {
+      if ("-".equals(outputSpec)) {
+        check(spec.commandLine().getOut());
+        spec.commandLine().getOut().println();
+      } else {
+        try (PrintWriter out = new PrintWriter(outputSpec)) {
```
snazy, line 110: Can you add a charset to the PrintWriter constructor?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
```diff
@@ -0,0 +1,264 @@
...
+          check(out);
+        }
+      }
+    } else {
+      check(new PrintWriter(NullOutputStream.NULL_OUTPUT_STREAM));
```
snazy, line 115: Can you add a charset to the PrintWriter constructor?

servers/quarkus-cli/pom.xml
```diff
@@ -45,6 +45,15 @@
...
+    <dependency>
+      <groupId>org.immutables</groupId>
+      <artifactId>value</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+    </dependency>
     <dependency>
```
snazy, line 57: Adding the following dependency gets rid of the javac warnings
```xml
    <dependency>
      <groupId>org.eclipse.microprofile.openapi</groupId>
      <artifactId>microprofile-openapi-api</artifactId>
      <optional>true</optional>
    </dependency>
```

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
snazy: WFM

snazy (commented):
LGTM - would just be nice to get rid of the javac warnings

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: Good catch. Added.

How do you track compile warnings?

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/CheckContent.java
dimas-b: added

servers/quarkus-cli/pom.xml
dimas-b: Added with the `provided` scope

snazy (approved):

Issue #4428, opened 2022-06-07
Java Nessie http client does not follow HTTP/308 redirects


tmnd1991:
**What happened**: 
When invoking any `NessieApiV1` method which gets a response code 3xx (for example to redirect http to https) I get:

```
java.lang.UnsupportedOperationException: Unknown status code 308.
```

**What you expected to happen**:

I would expect that the client correctly performs the redirection.

**How to reproduce it (as minimally and precisely as possible)**:

I can reproduce it connecting to my Nessie instance, but I'm working on a reproducer into Nessie unit tests

**Details**:
- Nessie server type (docker/built from source) and version: 0.30.0
- Client type (Ex: UI/Spark/pynessie ...) and version: Spark/Java

snazy:
@tmnd1991 Thanks for the issue and PR.
Not against this change, but could you elaborate when the redirect actually happens?

snazy:
Hm - I wonder why the change's actually necessary, because the default for `java.net.HttpURLConnection#getInstanceFollowRedirects` is `true`. The only case when Java's implementation doesn't work is HTTP/308 (see `sun.net.www.protocol.http.HttpURLConnection#followRedirect`).

tmnd1991:
I see it happening when redirecting from http to https, might it be? without the patch applied I get the 308 error I mentioned

snazy:
Closing this one, because we have Java's new HTTP client in `:nessie-client` now.

PR #4429, opened 2022-06-07
Follow redirections

tmnd1991:
See #4428

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
@@ -37,12 +37,12 @@ import java.util.List;
 import java.util.function.BiConsumer;
 import java.util.zip.GZIPOutputStream;
 import javax.net.ssl.HttpsURLConnection;
-
 import org.projectnessie.client.http.HttpClient.Method;
 
 /** Class to hold an ongoing HTTP request and its parameters/filters. */
 public class HttpRequest {
 
+  private static final String LOCATION_HEADER = "Location";
   private final HttpRuntimeConfig config;
   private final UriBuilder uriBuilder;
   private final HttpHeaders headers = new HttpHeaders();
@@ -87,6 +87,11 @@ public class HttpRequest {
   }
 
   private HttpResponse executeRequest(Method method, Object body) throws HttpClientException {
+    return executeRequest(method, body, 0);
+  }
+
+  private HttpResponse executeRequest(Method method, Object body, int redirectCount)
+      throws HttpClientException {
     URI uri = uriBuilder.build();
     try {
       HttpURLConnection con = (HttpURLConnection) uri.toURL().openConnection();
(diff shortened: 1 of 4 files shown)


tmnd1991:
@snazy

clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
```diff
@@ -31,11 +31,13 @@
 import java.net.ProtocolException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.function.BiConsumer;
 import java.util.zip.GZIPOutputStream;
 import javax.net.ssl.HttpsURLConnection;
+
```
snazy, line 40: This empty line will cause checkstyle to complain

clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
```diff
@@ -140,8 +142,21 @@ private HttpResponse executeRequest(Method method, Object body) throws HttpClien
           }
         }
         con.connect();
-        con.getResponseCode(); // call to ensure http request is complete
-
+        con.getResponseCode();
```
snazy, line 145: Better get the response code into a variable (and keep the comment).

clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
```diff
@@ -140,8 +142,21 @@ private HttpResponse executeRequest(Method method, Object body) throws HttpClien
...
+        if (con.getResponseCode() == 308 || con.getResponseCode() == 307) {
```
snazy, line 146: Let's restrict to 308, because 307 is already handled by `HttpURLConnection`. Also use a new enum value in `Status` for 308.

clients/client/src/main/java/org/projectnessie/client/http/HttpRequest.java
```diff
@@ -140,8 +142,21 @@ private HttpResponse executeRequest(Method method, Object body) throws HttpClien
...
+          String newLocation = con.getHeaderField("Location");
+          return new HttpRequest(
```
snazy, line 148: Redirects should be guarded with a "max number of redirects" (which is what `HttpURLConnection` does).
In this case, for HTTP/308, it's probably sufficient to only handle the HTTP/308 case once, because a http->https redirect should only happen once.
Proposal:
A `do {} while (false)` loop around the outermost `try`-block.
In the `if` handling the 308 check if a HTTP/308 was already handled. If not, then update the uri to the `Location` header. If it was already handled, then error out.
If the `Location` header's missing, error out immediately (this is what all client implementation do).
This should guard the Nessie http client from an endless loop caused by a misconfigured server.

snazy:
I played a bit around with this "protocol upgrade" stuff and have some sad news.

It's not just that `sun.net.www.protocol.http.HttpURLConnection` does not implement HTTP/308, there's another issue in there, that would actually completely prevents a protocol change. And all that code has not been touched since the "initial load" of the Java source code into 14/15 years ago.

Java's new HTTP client (since Java 11) however does implement all legit redirect responses properly, in compliance with RFC 9110. If the new HTTP client is configured to handle redirects. Sadly we have to support Java 8 for some more time, so we cannot just use the new HTTP client.

I suspect the best solution for now is just to configure the right https base URI on the client side. This does also prevent an unnecessary redirect roundtrip and does not expose data + credentials in clear text over the network.

tmnd1991:
Thanks @snazy. The protocol update does not work on the same HttpUrlConnection instance, but if you create a new one, that's fine (this is what I tested locally). I think this PR (setting by default maxRetries to 0) is a step ahead, because at least, if someone wants to follow 308 redirects, can.

snazy:
Started working on #5280 , which would allow such redirects with Java 11+. Closing this one.

PR #4466, opened 2022-06-11
Allow dependabot to manage plugin dependencies

snazy:
Background: dependabot only manages dependencies (incl Gradle plugins) in build.gradle[.kts]
files. It scans the root build.gradle[.kts] fila and those in submodules referenced in
settings.gradle[.kts].
But dependabot does not manage managed plugin dependencies in settings.gradle[.kts].
However, since dependabot is a "dumb search and replace engine", we can use a trick:
1. Have this "dummy" build.gradle.kts file with all managed plugin dependencies.
2. Add an `include()` to this build file in settings.gradle.kts, surrounded with an
   `if (false)`, so Gradle does _not_ pick it up.
3. Parse this file in our settings.gradle.kts, provide a `ResolutionStrategy` to the
   plugin dependencies.


snazy:
Sample PRs:
* https://github.com/snazy/nessie/pull/1028
* https://github.com/snazy/nessie/pull/1030

nastra (approved):

PR #4472, opened 2022-06-13
Bump Iceberg to 0.13.2

nastra:
Iceberg 0.13.2 is about to be released and this PR makes sure that existing tests work with the new version


buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
```diff
@@ -160,6 +160,7 @@ fun Project.configureJava() {
     if (System.getProperty("withMavenLocal").toBoolean()) {
       mavenLocal()
     }
+    maven { setUrl("https://repository.apache.org/content/repositories/orgapacheiceberg-1088/") }
```
snazy, line 163: Move that up before `mavenCentral` and add a `content` + `include` rule to only include Iceberg.

snazy: And give it a `name`

nastra: this will be removed once 0.13.2 has been released

nastra:
superseded by https://github.com/projectnessie/nessie/pull/4477 thus not needed anymore

PR #4486, opened 2022-06-14
Revert "build(deps-dev): bump @testing-library/react from 12.1.4 to 13.3.0 in /ui"

snazy:
Reverts projectnessie/nessie#4362


XN137 (approved):
reason for revert?
i assume it broke tests silently? then +1

PR #4533, opened 2022-06-21
Fix reference deletion in RocksDB

snazy:


versioned/persist/rocks/src/main/java/org/projectnessie/versioned/persist/rocks/RocksDatabaseAdapter.java
```diff
@@ -598,7 +598,7 @@ protected void doRemoveFromNamedReferences(
         throw new RuntimeException(e);
       }
 
-      ReferenceNames.Builder newRefNames = referenceNames.toBuilder();
+      ReferenceNames.Builder newRefNames = referenceNames.toBuilder().clearRefNames();
```
XN137, line 601: whats the value of clearing the builder we get from `referenceNames` instead of just using a new one?

XN137 (commented):
doesnt this bug imply that we are missing test coverage for the `doRemoveFromNamedReferences` functionality?

XN137 (approved):
thanks for adding the test! lgtm

PR #4536, opened 2022-06-21
Hashed key-lists using open-addrressing

snazy:
Refactors key-lists handling in `CommitLogEntry`.

The `KeyListEntry`s in the key-list(s) in `CommitLogEntry` are currently organized by "least recently updated keys first", which degrades read performance with many keys per reference.

This PR re-organizes the keys using an open-addressing hash-map, with the content's `Key` as the open-addressing hash-map's key. The whole open-addressing hash-map is split into multiple segments: the first segment is the "embedded" key-list inside `CommitLogEntry`, with additional segments as `KeyListEntity`s. This allows effective point-queries, both to only fetch the required `KeyListEntry` and within the `KeyListEntry`.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
@@ -70,6 +70,15 @@ public interface CommitLogEntry {
    */
   List<Hash> getKeyListsIds();
 
+  @Nullable
+  List<Integer> getKeyListEntityOffsets();
+
+  @Nullable
+  Float getKeyListLoadFactor();
+
+  @Nullable
+  Integer getKeyListBucketCount();
+
   /** Number of commits since the last complete key-list. */
   int getKeyListDistance();
 
@@ -95,20 +104,25 @@ public interface CommitLogEntry {
       int keyListDistance,
       KeyList keyList,
       Iterable<Hash> keyListIds,
+      Iterable<Integer> keyListEntityOffsets,
       Iterable<Hash> additionalParents) {
-    return ImmutableCommitLogEntry.builder()
-        .createdTime(createdTime)
-        .hash(hash)
-        .commitSeq(commitSeq)
-        .parents(parents)
-        .metadata(metadata)
-        .puts(puts)
-        .deletes(deletes)
-        .keyListDistance(keyListDistance)
-        .keyList(keyList)
-        .addAllKeyListsIds(keyListIds)
-        .addAllAdditionalParents(additionalParents)
-        .build();
+    ImmutableCommitLogEntry.Builder c =
+        ImmutableCommitLogEntry.builder()
+            .createdTime(createdTime)
+            .hash(hash)
+            .commitSeq(commitSeq)
+            .parents(parents)
+            .metadata(metadata)
+            .puts(puts)
+            .deletes(deletes)
+            .keyListDistance(keyListDistance)
+            .keyList(keyList)
+            .addAllKeyListsIds(keyListIds)
+            .addAllAdditionalParents(additionalParents);
+    if (keyListEntityOffsets != null) {
+      c.addAllKeyListEntityOffsets(keyListEntityOffsets);
+    }
+    return c.build();
   }
 
   enum KeyListVariant {
@@ -125,20 +139,21 @@ public interface CommitLogEntry {
      */
     EMBEDDED_AND_EXTERNAL_MRU,
     /**
-     * The variant in which Nessie versions since 0.31.0 either write a <em>sorted</em> {@link
-     * CommitLogEntry#getKeyList() embedded key-list} <em>or</em> multiple {@link KeyListEntity
-     * key-list entities}, each representing a hash bucket.
+     * The variant in which Nessie versions since 0.31.0 maintains {@link
+     * CommitLogEntry#getKeyList() embedded key-list} and {@link KeyListEntity key-list entities}.
      *
-     * <p>If the total serialized size of all serialized {@link KeyListEntry} objects is up to
-     * {@link DatabaseAdapterConfig#getMaxKeyListSize()}, the {@link CommitLogEntry#getKeyList()
-     * embedded key-list} contains key-list-entries sorted by {@link KeyListEntry#getKey() content
-     * key}.
+     * <p>{@link KeyListEntry}s are maintained as an open-addressing hash map with {@link
+     * org.projectnessie.versioned.Key} as the map key.
      *
-     * <p>Otherwise the embedded key-list will be empty. All keys will be persisted as "external"
-     * {@link KeyListEntity key-list entities}. Each {@link KeyListEntity key-list entity}
-     * represents a hash bucket. All {@link KeyListEntry}s in each bucket are sorted by {@link
-     * KeyListEntry#getKey() content key}.
+     * <p>That open-addressing hash map is split into multiple segments, if necessary.
+     *
+     * <p>The first segment is represented by the {@link CommitLogEntry#getKeyList() embedded
+     * key-list} with a serialized size goal up to {@link
+     * DatabaseAdapterConfig#getMaxKeyListSize()}. All following segments have a serialized size up
+     * to {@link DatabaseAdapterConfig#getMaxKeyListEntitySize()} as the goal.
+     *
+     * <p>Maximum size constraints are fulfilled using a best-effort approach.
      */
-    HASHED_AND_SORTED
+    OPEN_ADDRESSING
   }
 }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
@@ -77,7 +77,7 @@ public interface CommitLogEntry {
   Float getKeyListLoadFactor();
 
   @Nullable
-  Integer getKeyListBucketCount();
+  Integer getKeyListSegmentCount();
 
   /** Number of commits since the last complete key-list. */
   int getKeyListDistance();
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
@@ -80,6 +80,10 @@ public interface CommitLogEntry {
     return KeyListVariant.EMBEDDED_AND_EXTERNAL_MRU;
   }
 
+  default boolean hasKeySummary() {
+    return getKeyListVariant() != KeyListVariant.EMBEDDED_AND_EXTERNAL_MRU || getKeyList() != null;
+  }
+
   static CommitLogEntry of(
       long createdTime,
       Hash hash,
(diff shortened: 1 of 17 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
```diff
@@ -67,62 +60,50 @@ class KeyListBuildState {
...
+    return Math.abs(entry.getKey().hashCode() % bucketCount);
   }
 
-  private void newKeyListEntity() {
-    currentSize = 0;
-    currentKeyList = ImmutableKeyList.builder();
-  }
+  List<KeyListEntity> finish() {
+    int totalSize = entries.stream().mapToInt(serializedEntrySize).sum();
+    int remainingEmbedded = maxEmbeddedKeyListSize - currentSize;
```
dimas-b, line 73: `currentSize` is the size of a `CommitLogEntry`, isn't it? Does it make sense to subtract it from the max "key list" size?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
```diff
@@ -67,62 +60,50 @@ class KeyListBuildState {
...
 
-  private void addToKeyListEntity(KeyListEntry keyListEntry, int keyTypeSize) {
-    currentSize += keyTypeSize;
-    currentKeyList.addKeys(keyListEntry);
-  }
+    int bucketCount = totalSize / maxBucketSize + 1;
+    List<List<KeyListEntry>> buckets = new ArrayList<>(bucketCount);
+    for (int i = 0; i < bucketCount; i++) {
+      buckets.add(new ArrayList<>());
+    }
```
dimas-b, line 81: I think buckets can still overflow in terms of "entity size" in this case if the key distribution is not strictly uniform. Would it make sense to start low and increase the bucket count iteratively until all buckets fit under the size limit?

Alternatively, this seems like a reasonable case for open hashing... That would be more flexible in terms of incremental inserts. WDYT?

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestKeyListBuildState.java
```diff
@@ -0,0 +1,191 @@
...
+              .sorted(Comparator.comparing(KeyListEntry::getKey))
+              .collect(Collectors.toList()));
+    }
+    List<KeyListEntry> expectedEmbedded =
+        entries.stream()
+            .filter(e -> keyListBuilder.bucket(e, bucketCount) == 0)
+            .sorted(Comparator.comparing(KeyListEntry::getKey))
+            .collect(Collectors.toList());
+
+    List<KeyListEntity> entities = keyListBuilder.finish();
```
dimas-b, line 154: Assert serialized size is under the configured limit?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
```diff
@@ -101,4 +106,35 @@ static CommitLogEntry of(
         .addAllAdditionalParents(additionalParents)
         .build();
   }
+
+  enum KeyListVariant {
```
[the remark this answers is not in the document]
snazy: That's actually the enum value 0 - backwards compatibility :(

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/CommitLogEntry.java
```diff
@@ -101,4 +106,35 @@ static CommitLogEntry of(
...
+     *
+     * <p>If the total serialized size of all serialized {@link KeyListEntry} objects is up to
+     * {@link DatabaseAdapterConfig#getMaxKeyListSize()}, the {@link CommitLogEntry#getKeyList()
+     * embedded key-list} contains key-list-entries sorted by {@link KeyListEntry#getKey() content
+     * key}.
+     *
+     * <p>Otherwise the embedded key-list will be empty. All keys will be persisted as "external"
+     * {@link KeyListEntity key-list entities}. Each {@link KeyListEntity key-list entity}
+     * represents a hash bucket. All {@link KeyListEntry}s in each bucket are sorted by {@link
+     * KeyListEntry#getKey() content key}.
```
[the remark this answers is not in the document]
snazy: done in the other PR

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
snazy: > buckets can still overflow in terms of "entity size"

they actually do - both due to length of the keys _and_ unlucky hash distribution (buckets w/ 0 entries .. buckets with 3x more entries).

snazy:
Pushed another update that writes the key-lists using an open-addressing map, split into segments (described in the javadoc of `CommitLogEntry.KeyListVariant#OPEN_ADDRESSING`).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
```diff
@@ -15,114 +15,155 @@
...
+
+    List<KeyListEntity> builtEntities =
+        keyLists.stream()
+            .skip(1)
+            .map(keyList -> KeyListEntity.of(randomHash(), keyList))
+            .collect(Collectors.toList());
+
+    if (!builtEntities.isEmpty()) {
+      builtEntities.stream().map(KeyListEntity::getId).forEach(newCommitEntry::addKeyListsIds);
+      newCommitEntry.addAllKeyListEntityOffsets(offsets);
```
dimas-b, line 159: Another idea, I was considering was to hash over large buckets, but have a sorted list inside each bucket.. I guess that would be a hybrid approach between a hash table of lists and a total open addressing hash table. I suppose the benefit would be that we do not have to keep a separate list on indexes and handling collisions within a bucket is easier (by sorted list). Also buckets can fill up independently without cause a re-hash or extra I/O hops (I suppose it is faster to search within a bucket than load the next hop).

We could even do perfect hashing within each bucket... but a sorted list would be fine, I guess.

So the lookup / insertion would go like this:
* bucket_index = hash % num_buckets
* on add, try fill first bucket up to size limit, rehash if overflow. Only one bucket needs to be re-written unless there's an overflow.
* on read, one bucket load is sufficient (by hash index) if not found, the entry does not exist anywhere.
* on rehash, we undersize buckets to accommodate future inserts.

WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapterConfig.java
```diff
@@ -118,6 +119,23 @@ default int getMaxKeyListEntitySize() {
     return DEFAULT_MAX_KEY_LIST_ENTITY_SIZE;
   }
 
+  /**
+   * Configures key-list hash-table load-factor for the open-addressing hash map used for key-lists.
+   *
+   * <p>Small segments and load factors above 0.65 are * bad, as those would lead to many database
```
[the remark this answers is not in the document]
snazy: IntelliJ format artifact....

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
snazy: Actually on the fence here between a "pure open-addressing" approach and a "mixed" approach (open-addressing for segmenting + sorted segments).
I've added a new test `AbstractManyKeys.pointContentKeyLookups()` to #4551 - the setting for the test (with ~160 bytes for each key-entry & 16kB entity size) leads to very few "additional fetches". Bumping that 16kB entity size to 64kB yielded no observed. (Hint: can be observed by adding a breakpoint in the line calling `fetchKeyLists()` in `AbstractDatabaseAdapter.fetchValuesHandleOpenAddressingKeyList()` with the condition `round > 0`.)

As I said, not sure yet. Thinking about other related stuff as well:
* Could we return a sorted content-key list to clients?
  Feels like this could be implemented by fetching all key-list-entities and merging the iterators. But that's a minor win.
* How could we eventually implement paging for "get content keys"?
  Feels like that could be easily implemented by with a paging token consisting of the commit-ID + bucket number. But also relatively easy w/ the old and mixed approach: commit-ID + segment number + offset in segment.

One argument against "fixed proportional segments" is the non-deterministic length of keys and potentially bad distribution of hash values, especially w/ non segment counts that are not a power of two. Even if you have e.g. 20000 keys, there can be (equally distributed) segments with no keys, key keys and "too many" keys (exceeding a database limit).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/KeyListBuildState.java
```diff
@@ -15,114 +15,156 @@
...
-  private void addToEmbedded(KeyListEntry keyListEntry, int keyTypeSize) {
-    currentSize += keyTypeSize;
-    embeddedBuilder.addKeys(keyListEntry);
-  }
+      if (entry != null) {
+        int entrySize = serializedEntrySize.applyAsInt(entry);
+
+        if (bucketSize + entrySize > maxBucketSize) {
+          maxBucketSize = maxKeyListEntitySize;
+          offsets.add(i);
```
dimas-b, line 128: Another idea: Use the full `long` value range for the "virtual" hash table, and have sorted listed in each bucket. We'd set index boundaries based on key size and load factor applied to each bucket (to reserve space for future inserts).

On get, we'd scan the list of offsets (same as now) to arrive at the bucket ID.
On update, we'd add to the bucket where the hash lands until it fills up, in which case we split this and following buckets (but not change to previous buckets).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,148 @@
...
+            ? keyListSegmentOffsetsList.stream().mapToInt(Integer::intValue).toArray()
+            : new int[0];
+    keyListCount = 1 + keyListSegmentOffsets.length;
+
+    keyListsArray = new KeyListEntry[keyListCount][];
+    keyListsArray[0] = entry.getKeyList().getKeys().toArray(new KeyListEntry[0]);
+
+    entityIdToSegment = new Object2IntHashMap<>(keyListCount, Hashing.DEFAULT_LOAD_FACTOR, -1);
+
+    hashMask = entry.getKeyListBucketCount() - 1;
```
dimas-b, line 53: nit: add a comment about `getKeyListBucketCount` being a power of 2? Should we perhaps also validate that on read?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,148 @@
...
+
+    keyListIds = entry.getKeyListsIds();
+  }
+
+  /** Calculates the open-addressing bucket for a key. */
+  int bucketForKey(Key key) {
+    return key.hashCode() & hashMask;
```
dimas-b, line 60: nit: Is `%` much slower? Using `%` would remove the dependency on the hash table size being a power of 2 on reads :thinking:

dimas-b (dismissed):

dimas-b (approved):

PR #4542, opened 2022-06-22
versioned/spi: remove `Collator` in `Key` 

snazy:
`java.text.Collator` is nice when working with real text, but content
keys are not "real text" (with for example german umlauts, french
accents, etc).

`Collator` is used in `org.projectnessie.versioned.Key` for all kinds
of preformance critical operations like `equals()`, `hashCode()` and
`compareTo()`.

Expecially #4536 suffered due to the use of `Collator`, which became
obvious in the test case `TestKeyListBuildState#manyBuckets`, because
it was super slow and a flame graph immediately pointed to the use
of `Collator`. A single iteration of that test takes ~2ms locally
in IntelliJ without `Collator`, but 70ms with `Collator` for 500 `Key`s.

(The change in AbstractDatabaseAdapter is rather a nice-to-have and
a unit test fix.)

### DIFF
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
@@ -15,22 +15,11 @@
  */
 package org.projectnessie.versioned;
 
-import java.text.Collator;
 import java.util.List;
-import java.util.Locale;
-import java.util.stream.Collectors;
 import org.immutables.value.Value.Immutable;
 
 @Immutable
 public abstract class Key implements Comparable<Key> {
-
-  private static final ThreadLocal<Collator> COLLATOR =
-      ThreadLocal.withInitial(
-          () -> {
-            Collator c = Collator.getInstance(Locale.US);
-            c.setStrength(Collator.PRIMARY);
-            return c;
-          });
 
   public abstract List<String> getElements();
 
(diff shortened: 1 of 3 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
```diff
@@ -42,12 +31,11 @@ static ImmutableKey.Builder builder() {
   /** Does a case insensitive comparison by key element. */
   @Override
   public final int compareTo(Key that) {
-    Collator collator = COLLATOR.get();
     List<String> a = this.getElements();
     List<String> b = that.getElements();
     int max = Math.min(a.size(), b.size());
     for (int i = 0; i < max; i++) {
-      int cmp = collator.compare(a.get(i), b.get(i));
+      int cmp = a.get(i).compareTo(b.get(0));
```
dimas-b, line 38: Is the comparison still case-insensitive as the javadoc implies?

versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
```diff
@@ -58,11 +46,11 @@ public final int compareTo(Key that) {
...
   @Override
   public int hashCode() {
-    final Collator collator = COLLATOR.get();
-    return getElements().stream()
-        .map(collator::getCollationKey)
-        .collect(Collectors.toList())
-        .hashCode();
+    int h = 1;
+    for (String element : getElements()) {
+      h = 31 + h + element.hashCode();
```
dimas-b, line 51: Does this match the old hash code or just an unrelated hash function? If the latter, wouldn't you normally want to _multiply_ by 31 here?

versioned/spi/src/test/java/org/projectnessie/versioned/TestKey.java
```diff
@@ -0,0 +1,64 @@
...
+            .describedAs("Compare of %s to %s expect %d", a, b, expectedCompare)
+            .extracting(k -> k.compareTo(b))
+            .asInstanceOf(INTEGER);
+    if (expectedCompare < 0) {
+      ass.isNegative();
+    } else if (expectedCompare > 0) {
+      ass.isPositive();
+    } else {
+      ass.isZero();
+    }
```
dimas-b, line 43: nit: `extracting(k -> Integer.signum(k.compareTo(b))).isEqualTo(expectedCompare)`?

dimas-b (approved):

dimas-b (approved):

PR #4547, opened 2022-06-23
Prevent accidental deletion of the default branch (#3401)

tomekl007:


servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -59,8 +59,6 @@ nessie.server.send-stacktrace-to-client=false
 #   'COMMIT_CHANGE_AGAINST_REFERENCE','ASSIGN_REFERENCE_TO_HASH','UPDATE_ENTITY','READ_ENTITY_VALUE','DELETE_ENTITY'] \
 #   && role=='admin_user'
 # nessie.server.authorization.rules.allow_listing_reflog=\
-#   op=='VIEW_REFLOG' && role=='admin_user'
-# nessie.server.authorization.rules.allow_deleting_default_branch=\
 #   op=='DELETE_DEFAULT_BRANCH' && role=='admin_user'
```
XN137, line 62: wrong lines got deleted here

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractReferences.java
```diff
@@ -246,6 +246,9 @@ void assign() throws Exception {
 
   @Test
   void recreateDefaultBranch() throws Exception {
+    // note: the default branch can not be deleted through the TreeApi,
+    // but the underlying DatabaseAdapter should still support this operati
```
XN137, line 250: typo: `operati` (maybe not everything was copied?)

also maybe we should think about this more and also add WHY the db adapter needs to be able to still do this when its not possible form the user facing apis :thinking:

XN137 (commented):
we might want to drop "accidental" from PR title, since we are generally preventing it now

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractReferences.java
tomekl007: Good point, I think that the main reasoning is that the underlying db adapter is a low-level API that may allow some internal code to remove the branch. On the other hand, we don't want to allow the same for the user-facing APIs.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
```diff
@@ -131,13 +131,11 @@ protected void assignReference(NamedRef ref, String expectedHash, Reference assi
   @Override
   protected void deleteReference(NamedRef ref, String expectedHash)
       throws NessieConflictException, NessieNotFoundException {
-    BatchAccessChecker check = startAccessCheck();
     if (ref instanceof BranchName && getConfig().getDefaultBranch().equals(ref.getName())) {
-      check.canDeleteDefaultBranch();
-    } else {
-      check.canDeleteReference(ref);
+      throw new IllegalArgumentException(
+          "Default branch '" + ref.getName() + "' can not be deleted.");
```
dimas-b, line 136: nit: `can not` -> `cannot`

tomekl007:
The Iceberg tests in the: `TestNessieIcebergViews` are failing because of the new behavior:
`org.projectnessie.error.NessieBadRequestException: Bad Request (HTTP/400): Default branch 'main' can not be deleted.`
Maybe we can migrate those tests to use a non-default branch, @snazy wdyt?

snazy:
> The Iceberg tests in the: `TestNessieIcebergViews` are failing because of the new behavior:
> `org.projectnessie.error.NessieBadRequestException: Bad Request (HTTP/400): Default branch 'main' can not be deleted.`
> Maybe we can migrate those tests to use a non-default branch, wdyt?

Sure, wfm

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
```diff
@@ -131,13 +131,11 @@ protected void assignReference(NamedRef ref, String expectedHash, Reference assi
   @Override
   protected void deleteReference(NamedRef ref, String expectedHash)
       throws NessieConflictException, NessieNotFoundException {
-    BatchAccessChecker check = startAccessCheck();
     if (ref instanceof BranchName && getConfig().getDefaultBranch().equals(ref.getName())) {
-      check.canDeleteDefaultBranch();
-    } else {
-      check.canDeleteReference(ref);
+      throw new IllegalArgumentException(
+          "Default branch '" + ref.getName() + "' cannot be deleted.");
```
dimas-b, line 136: I do not really have a strong opinion on forbidding deleting the default branch, but from a general usability perspective I guess it might be worth reversing the default on `canDeleteReference` to `false` so that we still allow full functionality at the API level, but the admin has to enabled deleting the default branch explicitly. WDYT?

dimas-b (commented):
Just a suggestion (below).

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImplWithAuthorization.java
tomekl007: That was the existing functionality, that we discussed in the ticket to change:
https://github.com/projectnessie/nessie/issues/3401

clients/deltalake/src/test/java/org/projectnessie/deltalake/AbstractDeltaTest.java
```diff
@@ -160,4 +165,13 @@ protected static Object[] toJava(Row row) {
   protected static Object[] row(Object... values) {
     return values;
   }
+
+  protected Branch getBranch() {
+    return api.getAllReferences().get().getReferences().stream()
```
snazy, line 170: `api.getReference()` instead of all-references

clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/BaseIcebergTest.java
```diff
@@ -80,13 +80,14 @@ public BaseIcebergTest(String branch) {
 
   private void resetData() throws NessieConflictException, NessieNotFoundException {
     for (Reference r : api.getAllReferences().get().getReferences()) {
-      if (r instanceof Branch) {
+      // remove only 'branch' that is used by the test.
+      if (r instanceof Branch && r.getName().equals(branch)) {
```
snazy, line 84: I'd prefer to delete all branches, that are not the default branch `main`. Otherwise tests that create more branches with the same name would start to fail.

clients/deltalake/src/test/java/org/projectnessie/deltalake/ITDeltaLogBranches.java
```diff
@@ -41,6 +41,12 @@ class ITDeltaLogBranches extends AbstractDeltaTest {
 
   @TempDir File tempPath;
 
+  private static final String BRANCH_NAME = "ITDeltaLog";
```
snazy, line 44: Nit: `ITDeltaLogBranches` ?

snazy (commented):
Feels like the special-case handling in `org.projectnessie.services.impl.TreeApiImplWithAuthorization#createReference` (the "conditional catch and re-throw") could go away as well now.

tomekl007:
> Feels like the special-case handling in `org.projectnessie.services.impl.TreeApiImplWithAuthorization#createReference` (the "conditional catch and re-throw") could go away as well now.

Are you sure? What if there will be a race when two independent users want to create a branch with the same name?

snazy (approved):
LGTM!

snazy:
Nit: when merging, remove the issue # from the commit/PR title and add `Fixes #3401` to the commit message

snazy:
> > Feels like the special-case handling in `org.projectnessie.services.impl.TreeApiImplWithAuthorization#createReference` (the "conditional catch and re-throw") could go away as well now.
> 
> Are you sure? What if there will be a race when two independent users want to create a branch with the same name?

Only one can win?

PR #4561, opened 2022-06-24
Limit key length and number of key elements

### LINKED ISSUE #4550: Enforce reasonable size limits
Objects like `ContentKey` or `Key` and are currently unrestricted in their serialized size, without throwing an error.

Especially for objects `ContentKey` or `Key` we should enforce a size limitation to prevent internal structures to become inefficient. I suspect that it is fine to restrict fully qualified table names to maybe 500 characters.

snazy:
Key length limit: max 500 characters
Key element limit: max 20 elements

Fixes #4550

### DIFF
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
@@ -15,11 +15,18 @@
  */
 package org.projectnessie.versioned;
 
+import com.google.common.base.Preconditions;
 import java.util.List;
-import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value;
 
-@Immutable
+@Value.Immutable
 public abstract class Key implements Comparable<Key> {
+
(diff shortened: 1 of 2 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/Key.java
```diff
@@ -15,12 +15,19 @@
...
 import java.util.List;
-import org.immutables.value.Value.Immutable;
+import org.immutables.value.Value;
 
-@Immutable
+@Value.Immutable
 public abstract class Key implements Comparable<Key> {
 
+  /** Maximum number of characters in a key. Note: characters can take up to 3 bytes via UTF-8. */
+  public static final int MAX_LENGTH = 500;
```
dimas-b, line 26: Would it be reasonable to allow configuring this via a java system property?

dimas-b (approved):

PR #4618, opened 2022-06-30
Remove Gradle type-safe project accessors

snazy:
Prerequisite for integrations and tools testing framework, which uses
Gradle's `includeBuild` functionality. Gradle type-safe project
accessors requires that project names must not contain dot characters,
but Iceberg project names _do_ use dot characters. Refactoring the
Iceberg build is not really an option.

It's a bit sad that it's necessary that we cannot use Gradle's
type-safe accessors.

For easier review, the whole change is split into 3 commits. The repetitive and "boring" changes in the `Remove Gradle type-safe project accessors` commit.


snazy:
> wondering: is the design for this framework outlined somewhere? do i understand correctly that we will include the iceberg build into our own? snapshot dependencies cant be used for this?

Yea - the overall idea is to use `includeBuild` to "wrap" at least Nessie + Iceberg in a single including build. Still fighting w/ some issues. The nice side of included builds is that "the whole thing", so tools/integrations-tests and Nessie and Iceberg appear in a single IDEA project - including all the code navigation, dependencies, etc. So I'd like to use that, if possible.

PR #4633, opened 2022-07-04
Use isort instead of flake8-import-order

omarsmak:
[Isort](https://github.com/PyCQA/isort) has a nice way to sort Python imports, thus introducing this instead of flake8-import-order. Plus one big bonus with isort, it will auto sort the imports as of `tox -e format` command without needing to manually fix the imports the orders from `flake8-import-order`.

### DIFF
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -26,3 +26,6 @@ docstring-convention = google
 
 [aliases]
 test = pytest
+
+[isort]
+profile=black
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -31,6 +31,7 @@ basepython = python
 deps =
     -r{toxinidir}/requirements_lint.txt
 commands =
+    isort pynessie tests tools
     black pynessie tests tools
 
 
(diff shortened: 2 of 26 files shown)


python/setup.cfg
```diff
@@ -26,3 +26,6 @@ docstring-convention = google
 
 [aliases]
 test = pytest
+
+[isort]
+profile=black
```
XN137, line 31: took me a while to find this... next time maybe do the setup in 1 commit and apply the format in another :sweat_smile:

python/tox.ini
```diff
@@ -31,6 +31,7 @@ basepython = python
 deps =
     -r{toxinidir}/requirements_lint.txt
 commands =
+    isort pynessie tests tools
```
XN137, line 34: just wondering:
did we confirm that lint cmd will find invalid import order as well?

is assume its happening because of the added `flake8-isort`, right?
maybe add a comment there similar as we have for `flake8-black`.

XN137 (approved):

python/setup.cfg
omarsmak: Forgot this is github not gerrit 🤣

XN137 (approved):
thx for the update

Issue #3766, opened 2022-03-29
StreamingUtil.getCommitLogStream is ignoring OptionalInt maxRecords


ajantha-bhat:
`AbstractRestCommitLog#commitLogPagingAndFilteringByAuthor
`
`completeLog` size should be same as `pageSizeHint`, but it is total log size.

https://github.com/projectnessie/nessie/blob/main/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestCommitLog.java#L395

Note:
Also need to verify other api in StreamingUtil for `pageSizeHint`

ajantha-bhat:
Testcase is already there as mentioned in the PR desvription

https://github.com/projectnessie/nessie/blob/main/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestCommitLog.java#L401

After line 401, just need to add an assert that completeLog.size() == pageSizeHint to see the issue.

dimas-b:
Re: `assert that completeLog.size() == pageSizeHint` strictly speaking the server does not _have_ to honour page size _hints_... it's just a suggestion coming from the client, but the server has authority over the page size.

PR #4638, opened 2022-07-05
NessieApi - automatic paging stream

### LINKED ISSUE #4291: Support pagination in the ReadCommits command
Currently it shows only the first page.

snazy:
Deprecate the `.get()` functions in NessieApi builder classes that
extend `PagingBuilder` and implement `.stream()` functions that
automatically perform paging. Using the `StreamingUtil` class
directly is also deprecated, because it is no longer needed.

This change "pushes" clients to use the "right" functionality.

Fixes #4291
Obsoletes #3766

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
@@ -54,6 +54,16 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
     this.fetcher = fetcher;
   }
 
+  Stream<E> limitedStream(OptionalInt maxRecords) throws NessieNotFoundException {
+    return limitedStream(maxRecords, OptionalInt.empty());
+  }
+
+  Stream<E> limitedStream(OptionalInt maxRecords, OptionalInt pageSizeHint)
+      throws NessieNotFoundException {
+    Stream<E> stream = generateStream(pageSizeHint);
+    return (maxRecords.isPresent()) ? stream.limit(maxRecords.getAsInt()) : stream;
+  }
+
   /**
    * Constructs the stream that uses paging under the covers.
    *
--- a/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
@@ -55,10 +55,10 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
   }
 
   Stream<E> limitedStream(OptionalInt maxRecords) throws NessieNotFoundException {
-    return limitedStream(maxRecords, OptionalInt.empty());
+    return limitedStream(OptionalInt.empty(), maxRecords);
   }
 
-  Stream<E> limitedStream(OptionalInt maxRecords, OptionalInt pageSizeHint)
+  Stream<E> limitedStream(OptionalInt pageSizeHint, OptionalInt maxRecords)
       throws NessieNotFoundException {
     Stream<E> stream = generateStream(pageSizeHint);
     return (maxRecords.isPresent()) ? stream.limit(maxRecords.getAsInt()) : stream;
--- a/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
@@ -54,16 +54,6 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
     this.fetcher = fetcher;
   }
 
-  Stream<E> limitedStream(OptionalInt maxRecords) throws NessieNotFoundException {
-    return limitedStream(maxRecords, OptionalInt.empty());
-  }
-
-  Stream<E> limitedStream(OptionalInt maxRecords, OptionalInt pageSizeHint)
-      throws NessieNotFoundException {
-    Stream<E> stream = generateStream(pageSizeHint);
-    return (maxRecords.isPresent()) ? stream.limit(maxRecords.getAsInt()) : stream;
-  }
-
   /**
    * Constructs the stream that uses paging under the covers.
    *
--- a/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
@@ -55,10 +55,10 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
   }
 
   Stream<E> limitedStream(OptionalInt maxRecords) throws NessieNotFoundException {
-    return limitedStream(OptionalInt.empty(), maxRecords);
+    return limitedStream(maxRecords, OptionalInt.empty());
   }
 
-  Stream<E> limitedStream(OptionalInt pageSizeHint, OptionalInt maxRecords)
+  Stream<E> limitedStream(OptionalInt maxRecords, OptionalInt pageSizeHint)
       throws NessieNotFoundException {
     Stream<E> stream = generateStream(pageSizeHint);
     return (maxRecords.isPresent()) ? stream.limit(maxRecords.getAsInt()) : stream;
--- a/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
+++ b/clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
@@ -16,7 +16,6 @@
 package org.projectnessie.client;
 
 import java.util.List;
-import java.util.OptionalInt;
 import java.util.Spliterator;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -38,12 +37,11 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
     /**
      * Fetch the first/next response.
      *
-     * @param pageSize page size
      * @param pagingToken paging continuation token
      * @return response object
      * @throws NessieNotFoundException if the ref does not exist
      */
-    R fetch(Integer pageSize, String pagingToken) throws NessieNotFoundException;
+    R fetch(String pagingToken) throws NessieNotFoundException;
   }
 
   private final Function<R, List<E>> entriesFromResponse;
@@ -60,12 +58,11 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
    * <p>This implementation fetches the first page eagerly to propagate {@link
    * NessieNotFoundException}.
    *
-   * @param pageSizeHint page size hint
    * @return stream of entries
-   * @throws NessieNotFoundException propagated from {@link Fetcher#fetch(Integer, String)}
+   * @throws NessieNotFoundException propagated from {@link Fetcher#fetch(String)}
    */
-  Stream<E> generateStream(OptionalInt pageSizeHint) throws NessieNotFoundException {
-    R firstPage = fetcher.fetch(pageSizeHint.isPresent() ? pageSizeHint.getAsInt() : null, null);
+  Stream<E> generateStream() throws NessieNotFoundException {
+    R firstPage = fetcher.fetch(null);
     Spliterator<E> spliterator =
         new Spliterator<E>() {
           private String pageToken;
@@ -93,9 +90,7 @@ final class ResultStreamPaginator<R extends PaginatedResponse, E> {
 
             if (currentPage == null) {
               try {
-                currentPage =
-                    fetcher.fetch(
-                        pageSizeHint.isPresent() ? pageSizeHint.getAsInt() : null, pageToken);
+                currentPage = fetcher.fetch(pageToken);
                 offsetInPage = 0;
                 // an empty returned page is probably an error, let's assume something went wrong
                 if (entriesFromResponse.apply(currentPage).isEmpty()) {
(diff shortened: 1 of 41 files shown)


snazy:
This PR is split into 4 commits - one with the implementation change and 3 more with different "kinds" of changes production, test and compatibility-test code.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -44,31 +44,32 @@ public void execute() throws NessieNotFoundException {
...
-        CommitMeta commitMeta = logEntry.getCommitMeta();
-        spec.commandLine()
-            .getOut()
-            .printf(
-                "%s\t%s\t%s [%s]\n",
-                Objects.requireNonNull(commitMeta.getHash()).substring(0, 8),
-                commitMeta.getAuthorTime(),
-                commitMeta.getMessage(),
-                commitMeta.getAuthor());
+      api.getCommitLog().refName(ref).fetch(fetchOption).stream(OptionalInt.empty())
```
dimas-b, line 47: Please mention that this PR fixes #4291

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
```diff
@@ -36,10 +42,15 @@
    */
   GetAllReferencesBuilder fetch(FetchOption fetchOption);
 
-  /**
-   * Fetches all references and returns them in a {@link ReferencesResponse} instance.
-   *
-   * @return Fetches all references and returns them in a {@link ReferencesResponse} instance.
-   */
+  @Override
   ReferencesResponse get();
```
dimas-b, line 46: Do we want to deprecate this method too? The `@Deprecated` annotation on the overridden method does not normally cause warnings on the more specific method.

By the way, it has usages not converted to `stream()` in `AbstractRestReferences`... Was that intentional?

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -15,16 +15,36 @@
...
+   */
+  @Deprecated
+  RESP get() throws NessieNotFoundException;
+
+  /**
+   * Retrieve entries/results as a Java {@link Stream}.
+   *
+   * @param maxTotalRecords {@link OptionalInt#empty()} for unlimited amount of results or a limit
+   */
+  Stream<ENTRY> stream(OptionalInt maxTotalRecords) throws NessieNotFoundException;
```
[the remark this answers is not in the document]
snazy: Added a `default` method

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -15,16 +15,36 @@
...
    */
   R pageToken(String pageToken);
+
+  /**
+   * Fetches responses.
+   *
+   * @deprecated Use {@link #stream(OptionalInt)} instead.
+   */
+  @Deprecated
+  RESP get() throws NessieNotFoundException;
```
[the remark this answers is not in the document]
snazy: Not a fan of implying such things, because `Stream`s can do "interesting" things on the way - it would effectively become a hard to track side effect and requires exposing state from the implementation.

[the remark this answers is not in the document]
snazy: If people want more insight, they can still implement manual paging.

clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/BaseIcebergTest.java
```diff
@@ -79,14 +79,20 @@ public BaseIcebergTest(String branch) {
...
+    Branch defaultBranch = api.getDefaultBranch();
+    api.getAllReferences().stream(OptionalInt.empty())
+        .forEach(
+            ref -> {
+              try {
+                if (ref instanceof Branch && !ref.getName().equals(defaultBranch.getName())) {
+                  api.deleteBranch().branch((Branch) ref).delete();
+                } else if (ref instanceof Tag) {
+                  api.deleteTag().tag((Tag) ref).delete();
+                }
```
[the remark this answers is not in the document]
snazy: That's rather a test-only thing. Not sure whether that should be added to user-facing prodcution code.

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
snazy: I've removed the deprecation, because as you mentioned, people may want to use explicit paging.

clients/client/src/main/java/org/projectnessie/client/ResultStreamPaginator.java
```diff
@@ -54,6 +54,16 @@
     this.fetcher = fetcher;
   }
 
+  Stream<E> limitedStream(OptionalInt maxRecords) throws NessieNotFoundException {
+    return limitedStream(maxRecords, OptionalInt.empty());
+  }
+
+  Stream<E> limitedStream(OptionalInt maxRecords, OptionalInt pageSizeHint)
+      throws NessieNotFoundException {
+    Stream<E> stream = generateStream(pageSizeHint);
```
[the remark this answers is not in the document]
snazy: Not a concern of this PR, because the added `.stream()` functions don't allow setting the page-size-hint.
Mind providing a test for this, since you raised the issue?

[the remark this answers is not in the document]
ajantha-bhat: I feel it is a concern as we are exposing a buggy interface to the user now.

Testcase is already there 
https://github.com/projectnessie/nessie/blob/main/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestCommitLog.java#L401

just need to add an assert that `completeLog.size() == pageSizeHint` to see the issue.

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -15,16 +15,53 @@
...
   /**
-   * For paged requests: pass the {@link EntriesResponse#getToken()} from the previous request,
-   * otherwise don't call this method.
+   * When using the manual paging, pass the {@code token} from the previous request, otherwise don't
+   * call this method.
    */
   R pageToken(String pageToken);
+
+  /**
+   * Fetches responses, but callers must implement paging on their own, if necessary. If in doubt,
```
dimas-b, line 37: suggestion: `Fetches a response chunk (may be one page or complete response depending on use case and parameters)...`

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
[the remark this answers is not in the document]
dimas-b: Undeprecating the `get()` method for manual paging SGTM (added a minor comment about its javadoc separately)

clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/BaseIcebergTest.java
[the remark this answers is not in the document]
dimas-b: fair enough

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
dimas-b: +1 to removing deprecation, but do we need this override now?

dimas-b (commented):
LGTM, just a few minor extra  comments :)

clients/client/src/main/java/org/projectnessie/client/api/GetAllReferencesBuilder.java
snazy: nope

dimas-b (dismissed):

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
```diff
@@ -55,16 +55,16 @@ private StreamingUtil() {
...
+   *     Optional.empty()} the stream will be unbounded.
    * @return stream of {@link Reference} objects
    * @deprecated Use {@link GetAllReferencesBuilder#stream(OptionalInt)}
    */
   @Deprecated
   public static Stream<Reference> getAllReferencesStream(
       @NotNull NessieApiV1 api,
       @NotNull Function<GetAllReferencesBuilder, GetAllReferencesBuilder> builderCustomizer,
-      @NotNull OptionalInt pageSizeHint)
+      @NotNull OptionalInt maxRecords)
```
dimas-b, line 67: This is potentially used by external code... Do we really have to change this parameter's semantics?

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -45,24 +45,24 @@
...
   /**
    * Retrieve entries/results as a Java {@link Stream}, uses automatic paging.
    *
+   * @param maxTotalRecords {@link OptionalInt#empty()} for unlimited amount of results or a limit
    * @param pageSizeHint {@link OptionalInt#empty()} to let the Nessie server choose the page size,
    *     or provide a <em>proposal</em> to the Nessie server
-   * @param maxTotalRecords {@link OptionalInt#empty()} for unlimited amount of results or a limit
    */
-  Stream<ENTRY> stream(OptionalInt pageSizeHint, OptionalInt maxTotalRecords)
+  Stream<ENTRY> stream(OptionalInt maxTotalRecords, OptionalInt pageSizeHint)
```
dimas-b, line 66: This could be confusing to end users too, I guess. WDYT about moving `maxTotalRecords` to a builder method?

clients/client/src/main/java/org/projectnessie/client/StreamingUtil.java
snazy: Where's that change??

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
snazy: We already have `maxRecords`, which is actually the page size. I tend to remove the `pageSizeHint` option - if people really need finer grained control, we can revisit this.

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -15,16 +15,49 @@
...
-public interface PagingBuilder<R extends PagingBuilder<R>> {
+public interface PagingBuilder<R extends PagingBuilder<R, RESP, ENTRY>, RESP, ENTRY> {
 
-  /** Recommended: specify the maximum number of records to return. */
+  /**
+   * When using the manual paging, it is recommended to set this parameter, this parameter is
+   * configured when necessary via {@link #stream(OptionalInt)}.
+   *
+   * <p>Only for manual paging via {@link #get()} - do <em>not</em> call when using any of the
+   * {@link #stream()} functions.
```
dimas-b, line 29: suggestion: `Sets the maximum number of records to be returned in a single response object from the get() method. When using stream() methods this parameter must not be set`.

Plus we'd error out if `stream()` is called _when_ this parameter is set. WDYT?

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
```diff
@@ -15,16 +15,49 @@
...
+  default Stream<ENTRY> stream() throws NessieNotFoundException {
+    return stream(OptionalInt.empty());
+  }
+
+  /**
+   * Retrieve entries/results as a Java {@link Stream}, uses automatic paging.
+   *
+   * @param maxTotalRecords {@link OptionalInt#empty()} for unlimited amount of results or a limit
+   */
+  Stream<ENTRY> stream(OptionalInt maxTotalRecords) throws NessieNotFoundException;
```
dimas-b, line 62: Maybe rename `maxTotalRecords` to `limit` so as to avoid confusion with `maxRecords()`?

On the other hard, we enforce stream limits only on the client side, so there's no real benefit in making it an API parameter ATM. Do you envision us passing it down to the server? Pehaps we could leave it out for now, keep only the parameterless `stream()` variant and add limits later, if/when they can be used to improve I/O?

For now a user can always do `builder.stream().limit(myLimit)`... WDYT?

clients/client/src/main/java/org/projectnessie/client/api/PagingBuilder.java
snazy: At that point, it's an unfinished builder. But I've applied the javadoc suggestion.

dimas-b: If mean in the `stream()` implementation we could check and throw an exception if there's a value for `maxRecords` in the builder. That would ensure the parameter is not misused with respect to `stream()`... WDYT?

dimas-b: Alternatively, we could treat `maxRecords()` as a `.limit()` on the `stream()` too. I think it would still match current API semantics. The caller would just lose the paging token when using `stream()` as opposed to `get()`, but that should be self-evident, I guess.

snazy: Oh my - good point. If a user fully consumes one stream, and then another one, it's fine. Concurrently... not so fine...

dimas-b (approved):

PR #4659, opened 2022-07-07
CI: Also test with Java 17 + 18

snazy:
Currently all CI jobs run with Java 11. To ensure that Nessie works
fine with Java versions 17 and newer as well (excluding Spark
related code + tests), this GH workflow is being introduced, which
runs daily against the main branch.

This new workflow can also be triggered manually, even in a fork,
choosing the branch to run against.

### DIFF
--- a/.github/workflows/newer-java.yml
+++ b/.github/workflows/newer-java.yml
@@ -1,0 +1,90 @@
+name: Newer Java versions
+
+on:
+  schedule:
+    # Run daily on week days
+    - cron:  '* 4 * * 1-5'
+  workflow_dispatch:
+
+jobs:
+  java:
+    name: Exercise Java version
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        include:
+          - java-version: 17
+          - java-version: 18
+    env:
+      SPARK_LOCAL_IP: localhost
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup runner
+      uses: ./.github/actions/setup-runner
+    - name: Setup Java, Gradle
+      uses: ./.github/actions/dev-tool-java
+      with:
+        java-version: ${{ matrix.java-version }}
+
+    - name: Gradle / spotless
+      uses: gradle/gradle-build-action@v2
+      with:
+        cache-read-only: true
+        # 'spotlessCheck' separate: workaround until https://github.com/diffplug/spotless/issues/1215 is fixed
+        arguments: spotlessCheck
+
+    - name: Gradle / compile
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: checkstyleMain checkstyleTest jar testClasses testJar
+
+    - name: Gradle / unit test
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: test
+
+    - name: Gradle / check incl. integ-test
+      uses: gradle/gradle-build-action@v2
+      with:
+        # '-x spotlessCheck': workaround until https://github.com/diffplug/spotless/issues/1215 is fixed
+        arguments: check -x spotlessCheck
+
+    - name: Gradle / Gatling simulations
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: gatlingRun
+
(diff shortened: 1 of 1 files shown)


snazy:
Sample run (still ruinning): https://github.com/snazy/nessie/actions/runs/2630538285

.github/workflows/newer-java.yml
```diff
@@ -0,0 +1,90 @@
...
+      uses: gradle/gradle-build-action@v2
+      with:
+        cache-read-only: true
+        # 'spotlessCheck' separate: workaround until https://github.com/diffplug/spotless/issues/1215 is fixed
+        arguments: spotlessCheck
+
+    - name: Gradle / compile
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: checkstyleMain checkstyleTest jar testClasses testJar
```
XN137, line 41: in the workflow logs i can see this warning:
```
> Task :buildSrc:compileKotlin
'compileJava' task (current target is 17) and 'compileKotlin' task (current target is 11) jvm target compatibility should be set to the same Java version.
```

also this:
```
GraalVM Native Image: Generating 'nessie-quarkus-0.30.1-SNAPSHOT-runner' (executable)...
========================================================================================================================
Warning: Method io.sentry.protocol.SdkVersion.<init>() not found.
Warning: Method io.sentry.protocol.SentryPackage.<init>() not found.
[1/7] Initializing...                                                                                   (16.0s @ 0.40GB)
 Version info: 'GraalVM 22.1.0 Java 11 CE'
 C compiler: gcc (redhat, x86_64, 8.5.0)
 Garbage collector: Serial GC
```
which makes me doubt i understand what combinations we are testing here.
what problems are we trying to catch by running these jobs regularly when quarkus tests still run on java11 ?

like does it even make sense for us to run java 18 tests before all of our libraries officially support java 18 ?
quarkus, delta, iceberg etc.
or are we just testing that java 18 can still run java11 bytecode? (sure there might be some stdlib changes like deprecations/removals etc... but most problems related to that might be outside our control and we would need to wait for the libraries to upgrade anyway).

imo we should comment explicitly in the workflow what types of bugs we are trying to catch here (i.e. what test coverage this is really adding... what java version are we compiling on, what version are we compiling for, what versions are our various tests types running on etc)

tbh atm i dont have a good understanding of what exact test coverage this is adding atm :sweat_smile:

XN137 (commented):
workflow itself seems ok but not sure i understand the test coverage scope exactly

.github/workflows/newer-java.yml
```diff
@@ -0,0 +1,90 @@
...
+jobs:
+  java:
+    name: Exercise Java version
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        include:
+          - java-version: 17
+          - java-version: 18
```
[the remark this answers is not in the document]
snazy: These arguments are the reason why this is a separate job. It's meant to find those things early.
BTW: we do already mention that Nessie builds with Java 17+18 - but have no CI for it

.github/workflows/newer-java.yml
snazy: You're mixing Java 11 for the native image with this change, which runs JVM based against 17+18.

Native images are currently built with Java 11, see #4660 to bump the Java version there.

snazy: That one `'compileJava' task (current target is 17) and 'compileKotlin' task (current target is 11) jvm target compatibility should be set to the same Java version.` isn't an actual issue - rather a hint.

dimas-b (approved):

.github/workflows/newer-java.yml
```diff
@@ -0,0 +1,90 @@
...
+          codeCoverageReport -x test -x intTest
+          publishToMavenLocal
+          -Puber-jar
+
+    - name: Gradle / build tools integration tests
+      uses: gradle/gradle-build-action@v2
+      with:
+        arguments: buildToolsIntegrationTest
+
+    - name: Gradle / integration test native
```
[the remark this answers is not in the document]
snazy: We do test the native image here.

PR #4664, opened 2022-07-08
Handle times in SQL extensions better

tomekl007:
We start accepting dateTime with time zone only. If the user passes the dateTime without zone:
"2022-06-29T09:26:07.974992" we should throw an exception denoting that the user needs to pass the zone explicitly. The new dataTime format should be: 2022-06-29T11:26:07.976777+02:00. It means that we are breaking the compatibility of the current API, but we concluded that having an exception with a meaningful message would be enough in that case. Also, Nessie is not 1.x version yet, so the API changes can be justified.

### DIFF
--- a/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
+++ b/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
@@ -32,7 +32,7 @@ import org.projectnessie.model.{
 }
 
 import java.time.format.DateTimeParseException
-import java.time.{Instant, LocalDateTime, ZoneOffset}
+import java.time.{Instant, LocalDateTime, ZoneOffset, ZonedDateTime}
 import java.util.OptionalInt
 import scala.collection.JavaConverters._
 
@@ -61,11 +61,14 @@ object NessieUtils {
       .map(x => x.replaceAll("`", ""))
       .map(x => {
         try {
-          LocalDateTime.parse(x).atZone(ZoneOffset.UTC).toInstant
+          ZonedDateTime.parse(x).toInstant
         } catch {
           case e: DateTimeParseException =>
             throw new NessieNotFoundException(
-              "Invalid timestamp provided: " + e.getMessage
+              String.format(
+                "Invalid timestamp provided: %s. You need to provide it with a zone info. For more info, see: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html",
+                e.getMessage
+              )
             )
         }
       })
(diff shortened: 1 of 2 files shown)


clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
```diff
@@ -61,11 +61,14 @@ object NessieUtils {
       .map(x => x.replaceAll("`", ""))
       .map(x => {
         try {
-          LocalDateTime.parse(x).atZone(ZoneOffset.UTC).toInstant
+          ZonedDateTime.parse(x).toInstant
```
ajantha-bhat, line 64: Apologies for the late response. 

Spark-3.3 introduced native syntax for time travel (similar to what delta has) and even they don't support timezones in the SQL time travel syntax.

So, just to be unified with spark. Can we use 
`SET TIME ZONE ` 'America/Los_Angeles';
or 
`spark.conf.set("spark.sql.session.timeZone", "UTC")`
syntax for setting the timezones?

for both of the above cases we can use the `spark.conf().get("spark.sql.session.timeZone")`
to get the currently set timezone and convert the timestamp here?

@snazy, @tomekl007 : WDYT?

tomekl007: We discussed that during our call and decided to not do it. We want to use a similar approach to Iceberg and Delta late and rely on explicit time zone passing.

clients/spark-extensions-base/src/test/java/org/projectnessie/spark/extensions/AbstractSparkSqlTest.java
```diff
@@ -439,11 +440,50 @@ void useShowReferencesIn() throws NessieNotFoundException {
   }
 
   @Test
-  void useShowReferencesAtTimestamp() throws NessieNotFoundException, NessieConflictException {
+  void throwWhenUseShowReferencesAtTimestampWithoutTimeZone()
```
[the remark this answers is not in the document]
tomekl007: > change the LocalDateTime.now(ZoneOffset.UTC) to something that does not use now(), but instead calculates it from the last commit to avoid the test to be prone to flakiness.

This particular test asserts that an exception is thrown when there is no time zone provided, so it does not matter if we pass `now()` or something else.


> have one commit before and one commit after the one the test asserts against. This would add a good amount of coverage to NessieUtils.findReferenceFromTimestamp

AND 

>add a test providing a timestamp that's before the first commit (throws an exception)

Done

snazy (approved):
LGTM!

clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
ajantha-bhat: ok.

ajantha-bhat (approved):

PR #4669, opened 2022-07-09
GH-WF: Add mandatory `description` fields

snazy:


dimas-b (approved):

.github/actions/dev-tool-java/action.yml
```diff
@@ -1,11 +1,14 @@
-name: 'Setup Java/Maven/Gradle'
+name: 'Setup Java/Gradle'
+description: 'Sets up Java + Gradle'
 inputs:
   java-version:
     required: false
     default: '11'
+    description: 'Java version, defaults to 11'
```
XN137, line 7: nit: why repeat the defaults? is it not visible otherwise?

XN137 (approved):
the added descriptions seem correct but TBH i dont understand why they are now "mandatory"?
if they dont add more value than the name, not sure why we need them.

PR #4676, opened 2022-07-11
UI/hygiene: Remove `pkg.json` package

snazy:
Fix for [security vulnerability](https://github.com/advisories/GHSA-7f3x-x4pr-wqhj),
transitively via `pkg.json@2.0.8 requires parse-url@^1.0.0 via a transitive dependency on git-up@1.2.1`


dimas-b (approved):
LGTM, but I did not actually check UI after this change :)

snazy:
> LGTM, but I did not actually check UI after this change :)

I checked it manually as well.

Issue #4436, opened 2022-06-08
Internal Server Error (500) on missing content type header - should be 400


dimas-b:
```
$ curl -X POST http://localhost:19120/api/v1/trees/branch/main/commit -d ''
{
  "status" : 500,
  "reason" : "Internal Server Error",
  "message" : "javax.ws.rs.NotSupportedException: RESTEASY003065: Cannot consume content type",
  "errorCode" : "UNKNOWN",
  "serverStackTrace" : null
}
```

Should be "Bad Request" `400`.

dalaro:
Hi.  I had a look at this, thinking it might make a good first contribution to the project.

For background, it seems `curl` defaults to POSTing `application/x-www-form-urlencoded` in this case:

```
$ curl --verbose -X POST http://localhost:19120/api/v1/trees/branch/main/commit -d ''
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:19120...
* Connected to localhost (127.0.0.1) port 19120 (#0)
> POST /api/v1/trees/branch/main/commit HTTP/1.1
> Host: localhost:19120
> User-Agent: curl/7.79.1
> Accept: */*
> Content-Length: 0
> Content-Type: application/x-www-form-urlencoded
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 415 Unsupported Media Type
< Content-Type: application/json
< content-length: 191
< 
{
  "status" : 415,
  "reason" : "Unsupported Media Type",
  "message" : "RESTEASY003065: Cannot consume content type",
  "errorCode" : "UNSUPPORTED_MEDIA_TYPE",
  "serverStackTrace" : null
* Connection #0 to host localhost left intact
```

If `curl` instead sends `Content-Type: application/json`, then Nessie responds with 400 Bad Request, as desired:

```
$ curl --verbose -H 'Content-Type: application/json' -X POST http://localhost:19120/api/v1/trees/branch/main/commit -d ''
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:19120...
* Connected to localhost (127.0.0.1) port 19120 (#0)
> POST /api/v1/trees/branch/main/commit HTTP/1.1
> Host: localhost:19120
> User-Agent: curl/7.79.1
> Accept: */*
> Content-Type: application/json
> Content-Length: 0
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 400 Bad Request
< validation-exception: true
[...]

PR #4711, opened 2022-07-14
Avoid converting 415 Unsupported Media Type into 500 ISE (#4436)

dalaro:
This is an attempt to address #4436, with context from my contemporaneous comment there.

Some notes on this PR:

- I've run `./gradlew test` and `gradlew intTest check` locally from the repo root, and I've run the newly introduced tests by themselves quite a bit during iteration; but maybe the github checks will reveal other problems, so it might be worth punting on review until the checks complete/fail.
- Is the exception message compatible with Nessie's style?  Resteasy generates the message "RESTEASY003065: Cannot consume content type" in the case touched by this commit, and NEM passes that message through to the client.  I noticed other response message strings prefixed by their status code for redundancy, and this one isn't prefixed that way.
- I think the newly-added `NessieUnsupportedMediaTypeException` fits the mold in `ErrorCode`, but maybe there's an argument for removing or renaming it.

_(the remainder of this descrption is just the message of the PR's sole commit)_

In cases where a client request specified a Content-Type that a Nessie-Quarkus method wouldn't accept, NessieExceptionMapper would convert the resulting 415 exception into a generic 500 exception.  The reason string still provided a useful hint ("Cannot consume content type"), but the code was more general than necessary.

This commit adds a case to JAX-WS NotSupportedException case to NessieExceptionMapper, preventing it from falling through to the catch-all 500 ISE case.  It also adds a couple of tests covering the change.

### DIFF
--- a/model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
+++ b/model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.error;
+
+/** This exception is thrown when the request's media type is not accepted at the specified path. */
+public class NessieUnsupportedMediaTypeException extends NessieConflictException {
+
+  public NessieUnsupportedMediaTypeException(NessieError error) {
+    super(error);
+  }
+
+  @Override
+  public ErrorCode getErrorCode() {
+    return ErrorCode.UNSUPPORTED_MEDIA_TYPE;
+  }
+}
--- a/model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
+++ b/model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
@@ -15,8 +15,8 @@
  */
 package org.projectnessie.error;
 
-/** This exception is thrown when the request's media type is not accepted at the specified path. */
-public class NessieUnsupportedMediaTypeException extends NessieConflictException {
+/** This exception is thrown when the request's media type is not accepted at its specified path. */
+public class NessieUnsupportedMediaTypeException extends NessieRuntimeException {
 
   public NessieUnsupportedMediaTypeException(NessieError error) {
     super(error);
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
@@ -144,7 +144,8 @@ class TestNessieError {
                         unwrap(
                             () -> client.newRequest().path("unsupportedMediaTypePut").put("foo")))
                 .isInstanceOf(NessieUnsupportedMediaTypeException.class)
-                .hasMessage("RESTEASY003065: Cannot consume content type"));
+                .hasMessage(
+                    "Unsupported Media Type (HTTP/415): RESTEASY003065: Cannot consume content type"));
   }
 
   @Test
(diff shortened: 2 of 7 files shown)


servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
```diff
@@ -134,6 +135,18 @@ void blankParameterQueryGet() {
                     "Bad Request (HTTP/400): blankParameterQueryGet.hash: must not be blank"));
   }
 
+  @Test
```
snazy, line 138: Do we need this test (and the change in ErrorTestService)? Not sure what's different to `AbstractResteasyTest.testGetUnknownContentType()`

model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.error;
+
+/** This exception is thrown when the request's media type is not accepted at the specified path. */
+public class NessieUnsupportedMediaTypeException extends NessieConflictException {
```
snazy, line 19: Wonder whether this should actually be a subclass of `NessieServiceException`, which is a `RuntimeExcepiton`.
`NessieConflictException` is meant for "application/data errors", but unsupported media types feel like network/configuration issues. WDYT?

dalaro: You're completely right.  That was careless on my part. I see now that ​`NessieConflictException`'s javadoc says prominently that it corresponds with HTTP status 409, and that every subclass of `NCE` corresponds to HTTP status 409 (except the one I introduced here).  That should have been my hint to look for something else, like `NessieServiceException`.  I'll change that.

servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
dalaro: I'd be fine with keeping either one of these and deleting the other.

I wrote this test first (with the associated change in `ErrorTestService`), then wrote the `AbstractRestEasyTest` case second.

I tentatively left them both in because they looked moderately cheap to execute -- locally ~10ms for the `TestNessieError` case and ~90ms for the other case on in-memory, though surely slower on other stores -- and I wasn't sure whether the difference in backends might eventually come in handy by revealing some interaction the `ErrorTestService` stub couldn't induce.  So, my response to that uncertainty was to err on the side of redundant coverage, but I really don't have a strong opinion here.

Incidentally, aside from deleting this, I'd be happy to add test coverage elsewhere, if you have suggestions.

model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
dalaro: Hmm, it seems `NessieServceException` is in clients/client, whereas `NessieUnsupportedMediaTypeException` is currently in model.

There are a bunch of ways to resolve this.  What do you think about making `NessieUnsupportedMediaTypeException` extend `NessieRuntimeException` instead, where the latter currently lives in model?

snazy: Argh - yea - I always run into that exception-hierarchy trap...

Yea, extending `NessieRuntimeException` seems fine.

servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
snazy: Hm - I don't have a strong opinion here. Fine to keep both to validate it in a slightly different way in Quarkus.

model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
dalaro: I've pushed this as a separate commit, eyeballing precedent in other closed PRs, but let me know if you'd rather see it squashed.

Aside from rationalizing the inheritance hierarchy, this also makes NessieUnsupportedMediaTypeException's message more consistent with other Nessie 4XX exception messages by indirectly involving NessieError#getFullMessage.  I've updated the one test that was sensitive to the change to match.

snazy: Yup - that looks good to me. But CI's failing now.

dalaro: I forgot to run spotless, my mistake.  It should be fully applied with the last force-push.

dimas-b (approved):

model/src/main/java/org/projectnessie/error/NessieUnsupportedMediaTypeException.java
dimas-b: +1 to extending `NessieRuntimeException`

PR #4712, opened 2022-07-14
Nessie REST (HTTP) client: verify JSON response

snazy:
When users configure the REST base URI incorrectly, e.g. by setting
it to `http://127.0.0.1:19120/` against a locally running
Nessie-Quarkus-Server, they just get some "cannot parse response
entity" error message, which isn't helpful.

This change verifies the the `Content-Type` returned by the server
is a valid JSON one, one that ends in either `/json` or `+json`.
If the response is not a JSON response, the thrown exceptions gives
a better hint to what _could_ be wrong.

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
@@ -19,7 +19,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import java.io.IOException;
 import java.io.InputStream;
-import org.projectnessie.client.rest.NessieServiceException;
+import org.projectnessie.client.rest.NessieBadResponseException;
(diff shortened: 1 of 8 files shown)


clients/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
```diff
@@ -44,13 +46,34 @@ private <V> V readEntity(ObjectReader reader) {
...
         return reader.readValue(is);
       }
     } catch (IOException e) {
       throw new HttpClientException("Cannot parse request.", e);
     }
   }
 
+  private void nonJsonResponse() throws IOException {
+    Status status = responseContext.getResponseCode();
+    throw new NessieServiceException(
```
dimas-b, line 62: I wonder if a `NessieBadResponseException` (extends `NessieServiceException`) may be more appropriate :thinking: It would allow interested clients to process this error case in a more robust way. WDYT?

PR #4713, opened 2022-07-14
Bump Gradle to 7.5

snazy:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5-all.zip
```
XN137, line 3: WDYT about adding `distributionSha256Sum=XYZ` from https://gradle.org/release-checksums here?

XN137 (commented):
imo we should run `./gradlew wrapper --gradle-version 7.5 --distribution-type all`
twice to also upgrade the gradle wrapper files to the new version

gradle/wrapper/gradle-wrapper.properties
snazy: makes sense!

XN137 (approved):

PR #4750, opened 2022-07-20
Fix POM transitive dependency issue

snazy:
Following _warning_ is emitted by the `buildToolIntegrationMaven` task
```
[WARNING] The POM for org.projectnessie:nessie-rest-services:jar:0.30.1-SNAPSHOT is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
```
which is caused by a missing dependency version for `jakarta.validation:jakarta.annotation-api` in the main nessie-pom

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -64,6 +64,7 @@ val versionImmutables = "2.9.0"
 val versionJackson = "2.13.3"
 val versionJacksonSpark3 = "2.13.3"
 val versionJacoco = "0.8.8"
+val versionJakartaAnnotationApi = "1.3.5"
 val versionJakartaEnterpriseCdiApi = "2.0.2"
 val versionJakartaValidationApi = "2.0.2"
 val versionJandex = "2.4.3.Final"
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -174,6 +175,7 @@ dependencies {
     api("io.quarkus:quarkus-bom:$versionQuarkus")
     api("io.quarkiverse.loggingsentry:quarkus-logging-sentry:$versionQuarkusLoggingSentry")
     api("io.rest-assured:rest-assured:$versionRestAssured")
+    api("jakarta.validation:jakarta.annotation-api:$versionJakartaAnnotationApi")
```
XN137, line 178: nit: put next to other `jakarta.validation` item

build.gradle.kts
```diff
@@ -64,6 +64,7 @@ val versionImmutables = "2.9.0"
 val versionJackson = "2.13.3"
 val versionJacksonSpark3 = "2.13.3"
 val versionJacoco = "0.8.8"
+val versionJakartaAnnotationApi = "1.3.5"
 val versionJakartaEnterpriseCdiApi = "2.0.2"
 val versionJakartaValidationApi = "2.0.2"
```
XN137, line 69: do these versions need to have any relation to each-other or can they be independently upgraded?

XN137 (approved):
lgtm with suggestion+question

i assume there is no way to let these kind of warnings fail the build in the future?

snazy:
> i assume there is no way to let these kind of warnings fail the build in the future?

There seems to be no way to tell maven to "fail on warning" :(

build.gradle.kts
snazy: :shrug:

PR #4751, opened 2022-07-20
Add Spark/Scala-version-matrix

### LINKED ISSUE #4668: Support Spark 3.3
Apache Iceberg already supports Spark 3.3 - so it's about time to support Spark 3.3 in Nessie as well.

### LINKED ISSUE #4739: Support Spark + Scala versions matrix
We currently build Nessie/Spark artifacts only for Scala 2.12 and for Spark 3.1 + 3.2.

This issue is about to support a version _matrix_ to support more Scala versions and more Spark versions.

Building for multiple Scala versions should be doable by reusing the same code base, but using different Scala compiler version.

This issue is also about renaming the artifacts for the Nessie Spark extensions by using the well adapted pattern:
```
artifact-id-${sparkMajorVersion}_${scalaMajorVersion}
```
so to have artifact IDs like `nessie-spark-extensions-3.2_2.13`.

Minimum set of version combinations:
| Spark | Scala | |
| --- | --- | --- |
| 3.1 | 2.12 | |
| 3.2 | 2.12 | |
| 3.2 | 2.13 | |
| 3.3 | 2.12 | (if this is a "plain copy" from 3.2) |
| 3.3 | 2.12 | (if this is a "plain copy" from 3.2) |

We should also publish artifacts with the old artifact IDs, that contain [relocations](https://maven.apache.org/guides/mini/guide-relocation.html) - so we'll publish the affected artifacts using the old _and_ new artifact IDs for a little while.

If easily doable, the PR for this issue can include #4668

snazy:
* Changes artifact notation for `nessie-spark-extensions-base` to `nessie-spark-extensions-base_<SCALA_VERSION>`
* Changes artifact notation for `nessie-spark-extensions` to `nessie-spark-extensions-<SPARK_VERSION>_<SCALA_VERSION`
* Artifact `nessie-spark-extensions` relocated to `nessie-spark-extensions-3.1_2.12`
* Artifact `nessie-spark-3.2-extensions` relocated to `nessie-spark-extensions-3.2_2.12`
* Artifact `nessie-spark-extensions-base` relocated to `nessie-spark-extensions-base_2.12`
* Adds support for Spark 3.3 (#4668). Spark 3.3 has been added by copying the sources for v3.2 to v3.3 (one tiny code change).
* Adds support for Scala 2.13. Required a minor code change in `BaseListReferenceExec#runInternal`.

With this change, the following Spark extensions artifacts will be produced:
* `nessie-spark-extensions-base_2.12`
* `nessie-spark-extensions-base_2.13`
* `nessie-spark-extensions-3.1_2.12`
* `nessie-spark-extensions-3.2_2.12`
* `nessie-spark-extensions-3.2_2.13`
* `nessie-spark-extensions-3.3_2.12`
* `nessie-spark-extensions-3.3_2.13`
* plus the the relocated artifact mentioned above

Includes the following Git renamses:
* `clients/nessie-spark-extensions/src` to `clients/nessie-spark-extensions/v3.1/src`
* `clients/nessie-spark-3.2-extensions/src` to `clients/nessie-spark-extensions/v3.2/src`

Fixes #4739
Fixes #4668


dimas-b (commented):
Can we avoid copying sources for Spark 3.3?... e.g. by building the same source with different Spark dependencies?

snazy:
> Can we avoid copying sources for Spark 3.3?... e.g. by building the same source with different Spark dependencies?

Nope - we already tried hard to do that. All we have is that `*-base` thing - but that's it.

snazy:
Hm - just realized that the CI duration for PRs has increased significantly. I'll do a follow-up to tweak the PR workflow.

dimas-b:
> Nope - we already tried hard to do that. All we have is that *-base thing - but that's it.

I was hoping with gradle it would be possible, no? Could we set the source locations in two modules to the same directory?

snazy:
> > Nope - we already tried hard to do that. All we have is that *-base thing - but that's it.
> 
> I was hoping with gradle it would be possible, no? Could we set the source locations in two modules to the same directory?
> 
> ... or at least do a source copy in gradle?

In theory yes, but I just realized that I had to do a tiny change in `NessieSparkSqlExtensionsParser.scala` (line 272 in 3.3)

dimas-b (approved):
The PR LGTM, but for some reason IntelliJ is not able to resolve the base class of `AssignReferenceExec` in all Spark version modules :shrug:

snazy:
> ... for some reason IntelliJ is not able to resolve ...

Damn it. But IntelliJ's behavior is correct. Since there are multiple Gradle projects referencing the same source directories, IntelliJ has to (randomly) chose one of the projects. It has to do that, to resolve cross-project dependencies from source files - including runtime and test dependencies.

Solved that by letting IntelliJ only import the project for first declared Scala version and also not import the relocating (legacy) projects. So you'll only see Scala 2.12 in IntelliJ.

Command-line Gradle runs however do see all projects.

dimas-b (commented):
Base classes are resolved now, but I'm not sure the overall effect is 100% positive, e.g. in v3.1 IDEA shows errors for `AssignReferenceExec` now... apparently due to incorrect class resolution:

image

snazy:
It's absolutely annoying how Spark/Scala require code duplication and special treatment...
Not sure whether there's anything that can be done here at the moment.
Compilation + tests via Gradle work - and 3.2 + 3.3 are fine.
I suspect, we can accept it as an IDE-only issue?

dimas-b (approved):
> I suspect, we can accept it as an IDE-only issue?

Fair enough.

PR #4759, opened 2022-07-21
Implement NessieApiV1 method for UnreachableReferenceHeads

### LINKED ISSUE #4690: Implement NessieApiV1#getUnreferencedHeads
Currently `DatabaseAdapter` level `ReferencesUtil` has the implementation to identify the unreferenced heads. 

To use this in GC (instead of reflog),  `NessieApiV1` should have the implementation which calls this via REST API endpoints.

ajantha-bhat:
- Support `NessieApiV1` API for Unreachable reference heads.
- Implement Authz with `VIEW_UNREACHABLE_REFERENCE_HEADS`.
- REST endpoint path for `unreachableHeads`.
- Move contents of `ReferenceUtils` to `AbstractDatabaseAdapter`.

TODO: `GCImpl#collectDeadReferences()` to use this new API. 
I tried replacing it, but GC algorithm needs to know when the reference is dropped (before or after cutoff time) to handle some time travel scenarios.  So, just the unreachable heads are not enough to replace reflog functionality. 

Fixes #4690


model/src/main/java/org/projectnessie/model/UnreachableHeadsResponse.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.List;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.immutables.value.Value;
+
+@Schema(type = SchemaType.OBJECT, title = "UnreachableHeadsResponse")
+@Value.Immutable
+@JsonSerialize(as = ImmutableUnreachableHeadsResponse.class)
+@JsonDeserialize(as = ImmutableUnreachableHeadsResponse.class)
+public interface UnreachableHeadsResponse {
```
snazy, line 29: This should implement paging

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
```diff
@@ -1,191 +0,0 @@
...
-import org.agrona.collections.Object2ObjectHashMap;
-import org.agrona.collections.ObjectHashSet;
-import org.projectnessie.versioned.GetNamedRefsParams;
-import org.projectnessie.versioned.Hash;
-import org.projectnessie.versioned.NamedRef;
-import org.projectnessie.versioned.ReferenceInfo;
-import org.projectnessie.versioned.ReferenceNotFoundException;
-
-@Beta
-public final class ReferencesUtil {
```
snazy, line 34: This refactoring isn't necessary and should be reverted

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -315,4 +316,7 @@ Stream<Commit<METADATA, VALUE>> getCommits(Ref ref, boolean fetchAdditionalInfo)
    * @return A stream of reflog entries.
    */
   Stream<RefLogDetails> getRefLog(Hash refLogId) throws RefLogNotFoundException;
+
+  /** Get all unreachable heads by reading all the commit logs. */
+  Set<Hash> getUnreachableHeads(long expectedCommitCount) throws ReferenceNotFoundException;
```
snazy, line 321: --> Stream

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
```diff
@@ -104,6 +105,12 @@ public interface NessieApiV1 extends NessieApi {
   /** Retrieve a diff between two references. */
   GetDiffBuilder getDiff();
 
+  /**
+   * Retrieve all the Unreachable reference heads (commit hash) created due to drop reference or
+   * assign reference operation.
+   */
+  List<String> getUnreachableReferenceHeads() throws NessieNotFoundException;
```
snazy, line 112: All other use cases implement a proper builder pattern - why not here? Lazy?

snazy (changes requested):

model/src/main/java/org/projectnessie/model/UnreachableHeadsResponse.java
ajantha-bhat: But the backend always computes a set, it cannot compute iteratively. 

Do you mean, build a stream on top of computed set<Hash> and do paging on that stream?

clients/client/src/main/java/org/projectnessie/client/api/NessieApiV1.java
ajantha-bhat: As there are no arguments to set for the builder, I felt like it will look odd to just have get() method.

Also in `getDefaultBranch()`, there was no builder pattern.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ReferencesUtil.java
ajantha-bhat: At `PersistVersionStore`, to call `ReferenceUtils`, I need `databaseAdapterConfig`
But it is hard to extract config from `TracingDatabaseAdapter`.
So, I moved those methods from 'ReferenceUtils' to `AbstractDatabaseAdapter`

Do you want me to add a method to `DatabaseAdapter` interface to get the configuration? I feel current refactoring makes these methods to be inlined with other methods.

model/src/main/java/org/projectnessie/model/UnreachableHeadsResponse.java
ajantha-bhat: Also, like commit log, backend doesn't persist the computed unreachable heads. So, pagination cannot work based on pageToken?

ajantha-bhat:
@snazy : I couldn't address any of the comments (as I have questions for them). I have replied to the comments. Please reply back.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
ajantha-bhat: As pagination is not supported, do we still need it as stream wrapping the HashSet?

ajantha-bhat:
https://github.com/projectnessie/nessie/issues/4690 is closed as this use case is not needed for GC. Hence, closing the PR aswell.

PR #4771, opened 2022-07-22
Shell script to validate publications

snazy:
Adds a shell script that can be run locally and in CI.

Simulates a release w/ signed artifacts and verifies that:
* `.asc` files exist for pom's module's, .jar's and -tests.jar's
* `-javadoc.jar` files exist for jars
* `-sources.jar` files exist for jars

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -80,14 +80,15 @@ jobs:
       with:
         arguments: gatlingRun
 
-    # Exercise these tasks to ensure those do not error out
-    - name: Gradle / assemble + publish local
+    - name: Gradle / code-coverage report
       uses: gradle/gradle-build-action@v2
       with:
-        arguments: |
-          assemble
-          publishToMavenLocal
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request.yml
```diff
@@ -80,14 +80,15 @@ jobs:
...
-        arguments: |
-          assemble
-          publishToMavenLocal
-          codeCoverageReport -x test -x intTest
+        arguments: codeCoverageReport -x test -x intTest
+
+    # Performs a `publishToMavenLocal` with `-Prelease` and a temporary GPG key - finds errors
+    # like missing .asc files and sources/javadoc jars
+    - name: Verify publication
+      run: build-tools-integration-tests/verify-publication.sh --noclean
```
snazy, line 91: Note: MUST NOT run this script in "main CI", because that'd break snapshot publishing.

XN137: nit: maybe add this as a comment to the script or workflow here (also why it would break it, to me its not immediately clear)

PR #4773, opened 2022-07-22
Add DeleteContent command

dimas-b:


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteContent.java
```diff
@@ -0,0 +1,78 @@
...
+      names = {"-m", "--message"},
+      description = "Commit message")
+  private String message;
+
+  @Spec private CommandSpec spec;
+
+  @Override
+  public void execute() throws NessieNotFoundException, NessieConflictException {
+    try (NessieApiV1 api = createNessieApiInstance()) {
+      ContentKey contentKey = ContentKey.of(key);
```
snazy, line 57: Maybe refactor to the following to allow deletion of multiple keys?
```java
List<Operation.Delete> deletes = keys.stream().map(ContentKey::fromPathString).map(Operation.Delete::of).collect(toList());
```

snazy: I know, it's like this in `ReadContent`, but I think having a fully qualified content-key is more intuitive. Wouldn't mind changing the `ReadContent` command to accept a content-key in the same way.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteContent.java
```diff
@@ -0,0 +1,78 @@
...
+
+      Reference refInfo = api.getReference().refName(ref).get();
+      if (refInfo.getType() != Reference.ReferenceType.BRANCH) {
```
snazy, line 60: Think you can skip this check - Nessie server will complain ;)

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteContent.java
dimas-b: In my experience using encoded key forms in CLI is very inconvenience... However, I happen to deal with a lot of cases where `.` is use a part of key element names. I guess it may not be a common use case in general, but for me it was easier to copy-paste a few key fragments into multiple `-k` args rather than manually encoding gots in element names :shrug:

dimas-b: I'll see about refactoring CLI tools to _print_ keys in encoded form, then I guess copy-paste will be easier.

dimas-b: Let's merge with current options to make the delete cmd available. I can open a follow-up issue for revisiting the generator CLI output and input key options/formatting (if you think it's still worth the effort)

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteContent.java
dimas-b: fixed

snazy (approved):

PR #4782, opened 2022-07-25
GH WF: Prevent duplicate WF runs

snazy:
The `Newer Java versions` shall run once every 24h on weekdays.
However, GH triggers the WF twice. This seems to be an issue in GH WFs,
but the simple workaround is to tweak the cron-expression.


XN137 (approved):
change lgtm but overall these run too frequently imo :sweat_smile:

PR #4795, opened 2022-07-26
Separate SNAPSHOT + Docker image publish

snazy:
Currently the "main CI" job automatically publishes SNAPSHOT versions
and a Docker image for every commit on Nessie's main branch. This can
especially lead to inconsistent SNAPSHOT publications, because unlike
release versions, _concurrent_ SNAPSHOT publications, which do happen
regularly, lead to inconsistent artifacts, which may or may not work
correctly.

This change removes the publication of SNAPSHOT artifacts and the
Docker image out of "main CI" into a scheduled GH workflow, that is
run twice every weekday. Only one instance of this new workflow is
allowed to run, so inconsistent SNAPSHOT releases are prevented.

### DIFF
--- a/.github/workflows/snapshot-publish.yml
+++ b/.github/workflows/snapshot-publish.yml
@@ -1,0 +1,67 @@
+name: Publish in-development builds from main
+
+on:
+  schedule:
+    # Run daily on week days
+    - cron:  '0 11,23 * * 1-5'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+jobs:
+  java:
+    name: Publish from main
+    runs-on: ubuntu-latest
+    env:
+      SPARK_LOCAL_IP: localhost
+    if: ${{ env.MAVEN_USERNAME }}
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup runner
+        uses: ./.github/actions/setup-runner
+      - name: Setup Java, Maven
+        uses: ./.github/actions/dev-tool-java
+
+      # Check that the commit that is becoming a release has passed CI.
+      # Note: intentionally hard-coded projectnessie/nessie, so this check even works when *testing*
+      # the workflow on a PR-branch.
+      - name: Check commit status
+        run: |
+          GIT_HEAD_SHA="$(git rev-parse HEAD)"
+          echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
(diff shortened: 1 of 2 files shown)


.github/workflows/snapshot-publish.yml
```diff
@@ -0,0 +1,67 @@
+name: Publish in-development builds from main
+
+on:
+  schedule:
+    # Run daily on week days
+    - cron:  '0 11,23 * * 1-5'
```
dimas-b, line 6: Will this run even when there are no _new_ commits?

.github/workflows/snapshot-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+            -Prelease -Puber-jar
+
+      - name: Gradle / integration test native
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: |
+            --no-daemon
+            :nessie-quarkus:quarkusBuild
+            -Pnative
+            -Pdocker
```
dimas-b, line 56: nit: I guess this does not actually run integration tests, so shall we drop `integration test native` from the name?

.github/workflows/snapshot-publish.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: ./.github/actions/dev-tool-java
+
+      # Check that the commit that is becoming a release has passed CI.
+      # Note: intentionally hard-coded projectnessie/nessie, so this check even works when *testing*
+      # the workflow on a PR-branch.
+      - name: Check commit status
+        run: |
+          GIT_HEAD_SHA="$(git rev-parse HEAD)"
+          echo ${{ secrets.GITHUB_TOKEN }} | gh auth login --with-token
+          gh api repos/projectnessie/nessie/commits/${GIT_HEAD_SHA}/check-runs --jq 'if ([.check_runs[] | select(.name | endswith(" release") or startswith("Dependabot ") or startswith("codecov/") or startswith("Report ") | not ) | select(.conclusion != "skipped") | .conclusion // "pending" ] | unique == ["success"]) then "OK" else error("Commit checks are not OK") end'
```
dimas-b, line 35: What if HEAD has CI in progress? Is this a concern? Could we use `gh api` to find the last commit that has complete CI results?

.github/workflows/snapshot-publish.yml
snazy: Yes - but given that we usually have multiple commits per day, I suspect that's okay.

snazy: (The hours (11+23 UTC) are outside of the "high commit traffic time")

.github/workflows/snapshot-publish.yml
snazy: Yea - CI in progress would let this check fail. Given that the wf is scheduled twice per day, that shouldn't be a big issue. And we can run it manually at any time.
Finding the last commit that's fine is probably over-engineering. It's doable, but it's probably not worth it.
Alternative could be to skip this check and assume that CI's fine.

dimas-b: Fair enough :+1:

dimas-b (approved):

PR #4802, opened 2022-07-28
Fix dependencies leaking into projects consuming Nessie

snazy:
While trying to bump Nessie in Iceberg to 0.40.1, the Iceberg build failed,
because it now receives "other" Iceberg artifacts via `nessie-client`.
This happens, because Nessie's dependency management (more precisely the
published Gradle module metadata) leaks into Iceberg's Gradle build.
Obviously, a workaround like adding an exclude-rule to the Iceberg build
would work, but it's not a solid fix.

This change tackles the problem by splitting the dependency-constraints
in `:nessie` into multiple dependency-constraints:
* `:nessie` stays, but is heavily reduced to (mostly) the bare minimum
  required by clients.
* `:nessie-deps-antlr` just for antlr
* `:nessie-deps-build-only` dependencies that are only required for the
  build (like annotation processors)
* `:nessie-deps-managed-only` only managed dependencies, that are used
  by plugins to e.g. google-java-formatter
* `:nessie-deps-persist` server-ish dependencies
* `:nessie-deps-quarkus` Quarkus related dependencies
* `:nessie-deps-testing` testing + microbenchmarks
* `:nessie-deps-iceberg` Iceberg, Spark + Deltalake

Sadly, this change also requires updating nearly all `build.gradle.kts`
files. (This "bulk work" is split out into a separate commit in the PR.)

Another sad thing is a "hack" to workaround an issue in Gradle regarding
multiple dependency-constraints from/via/using included-builds, in
`buildSrc/.../Utilities.kt`

`nessie-iceberg/settings.gradle.kts` has been updated to match the main
one for IntelliJ imports.

### DIFF
--- a/buildSrc/src/main/kotlin/Utilities.kt
+++ b/buildSrc/src/main/kotlin/Utilities.kt
@@ -160,7 +160,7 @@ fun DependencyHandlerScope.nessieProject(
  * Ideally, it should be sufficient to use [nessieProject], but that does not work properly and
  * results in this Gradle error: `Incompatible because this component declares a platform and the
  * consumer needed a library`. Although it is correct that the component declares a platform, it is
- * wrong that the consumer needs a librara...
+ * wrong that the consumer needs a library...
  */
 fun DependencyHandlerScope.nessieProjectPlatform(artifactId: String, gradle: Gradle): Dependency {
   if (!isIntegrationsTestingEnabled()) {
--- a/buildSrc/src/main/kotlin/Utilities.kt
+++ b/buildSrc/src/main/kotlin/Utilities.kt
@@ -25,6 +25,7 @@ import org.gradle.api.artifacts.ExternalModuleDependency
 import org.gradle.api.artifacts.ModuleDependency
 import org.gradle.api.artifacts.ProjectDependency
 import org.gradle.api.internal.artifacts.dependencies.DefaultProjectDependencyConstraint
+import org.gradle.api.invocation.Gradle
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.maven.MavenPublication
 import org.gradle.api.tasks.SourceSetContainer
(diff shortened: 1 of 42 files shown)


ajantha-bhat:
> While trying to bump Nessie in Iceberg to 0.40.1, the Iceberg build failed,

Thanks for fixing it. I too faced some error saying "classes are compiled with different version of JDK" while bumping to 0.40.0

settings.gradle.kts
```diff
@@ -112,7 +112,17 @@ gradle.beforeProject {
...
+    "code-coverage",
+    "nessie-deps-antlr",
+    "nessie-deps-build-only",
+    "nessie-deps-iceberg",
+    "nessie-deps-managed-only",
+    "nessie-deps-persist",
+    "nessie-deps-quarkus",
+    "nessie-deps-testing"
+  )
+  .forEach { include(it) }
```
dimas-b, line 125: Just curious: is this syntax preferable for any reason? I guess multiple `include(...)` calls would take less lines :)

buildSrc/src/main/kotlin/Utilities.kt
```diff
@@ -153,6 +154,30 @@ fun DependencyHandlerScope.nessieProject(
   }
 }
 
+/**
+ * Resolves a `platform()` dependency to a project in another Gradle build.
+ *
+ * Ideally, it should be sufficient to use [nessieProject], but that does not work properly and
+ * results in this Gradle error: `Incompatible because this component declares a platform and the
+ * consumer needed a library`. Although it is correct that the component declares a platform, it is
+ * wrong that the consumer needs a librara...
```
dimas-b, line 163: typo: `librara` -> `library`

dimas-b (approved):

settings.gradle.kts
snazy: Yea - it's just "less lines". spotless adds a blank line between each include :(

ajantha-bhat (approved):

settings.gradle.kts
dimas-b: TIL :)

dimas-b (approved):

PR #4833, opened 2022-08-01
Fix a few javac warnings

snazy:
* Unused variable/field
* Unnecessary lambda field


XN137 (approved):
changes lgtm

wondering:
do we have a convention on whether we use the `Test` as prefix or suffix?
because the affected files suggest it is not uniform it seems.

PR #4836, opened 2022-08-01
Remove Quarkus-BOM from non-Quarkus code

snazy:
The Quarkus-BOM references some dependencies that require at least
Java 11 (like `jboss-logging:3.5.0.Final`), which lets Iceberg's
Java 8 CI (and potentially other code running with Java 8) fail
with
```
java.lang.UnsupportedClassVersionError: org/jboss/logging/BasicLogger has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

This change removes all references to the `quarkus-bom` from non
Quarkus related Nessie projects.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,10 +80,12 @@ val versionJunit = "5.9.0"
 val versionLogback = "1.2.11"
 val versionMavenResolver = "1.7.3"
 val versionMaven = "3.8.6"
+val versionMicrometer = "1.9.2"
 val versionMockito = "4.6.1"
 val versionMongodbDriverSync = "4.7.1"
 val versionNessieApprunner = "0.21.4"
 val versionOpenapi = "3.0"
+val versionOpentracing = "0.33.0"
 val versionQuarkus = dependencyVersion("versionQuarkus") // ensure that plugin version is the same
 val versionQuarkusAmazon = "2.11.1.Final"
 val versionQuarkusLoggingSentry = "1.2.1"
@@ -105,6 +107,7 @@ versionIceberg = System.getProperty("nessie.versionIceberg", versionIceberg)
 versionClientNessie = System.getProperty("nessie.versionClientNessie", versionClientNessie)
 
 mapOf(
+    "versionAwssdk" to versionAwssdk,
     "versionCheckstyle" to versionCheckstyle,
     "versionClientNessie" to versionClientNessie,
     "versionErrorProneAnnotations" to versionErrorProneAnnotations,
@@ -114,6 +117,8 @@ mapOf(
     "versionGoogleJavaFormat" to versionGoogleJavaFormat,
     "versionJacoco" to versionJacoco,
     "versionJandex" to versionJandex,
+    "versionMicrometer" to versionMicrometer,
+    "versionOpentracing" to versionOpentracing,
     "versionProtobuf" to versionProtobuf,
     "versionRocksDb" to versionRocksDb,
     "quarkus.builder-image" to "quay.io/quarkus/ubi-quarkus-native-image:22.1-java17"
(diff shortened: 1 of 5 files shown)


build.gradle.kts
```diff
@@ -80,10 +80,12 @@ val versionJunit = "5.9.0"
 val versionLogback = "1.2.11"
 val versionMavenResolver = "1.7.3"
 val versionMaven = "3.8.6"
+val versionMicrometer = "1.9.2"
 val versionMockito = "4.6.1"
 val versionMongodbDriverSync = "4.7.1"
 val versionNessieApprunner = "0.21.4"
 val versionOpenapi = "3.0"
+val versionOpentracing = "0.33.0"
```
XN137, line 88: wondering:
are there any constraints around keeping these new version properties in sync with the versionQuarkus one?
like if we only upgrade quarkus, will the build now be downloading 2 different micrometer versions that could potentially be incompatible?

why do we need to explicitly declare versions for these dependencies when they are actually "pinned down" in the quarkus bom?

https://repo1.maven.org/maven2/io/quarkus/platform/quarkus-bom/2.11.1.Final/quarkus-bom-2.11.1.Final.pom
has:
```
<dependencyManagement>
  <dependencies>
  ...
    <dependency>
      <groupId>io.micrometer</groupId>
      <artifactId>micrometer-core</artifactId>
      <version>1.9.2</version>
    </dependency>
    ...
  </dependencies>
</dependencyManagement>
```

cant gradle manage these versions through the imported bom without actually pulling in these dependencies everywhere?

snazy: > are there any constraints around keeping these new version properties in sync with the versionQuarkus one?

it "just" has to "work" - no special requirements there - quarkus-bom's pulled in as an `enforcedPlatform` where really needed - to in doubt it "wins"

> why do we need to explicitly declare versions for these dependencies when they are actually "pinned down" in the quarkus bom?

because the quarkus-bom's not there in these places (which is the intent of this change)

> cant gradle manage these versions through the imported bom without actually ...

you can't pull in "just parts of it" - it's all or nothing

XN137: > you can't pull in "just parts of it" - it's all or nothing

thats the part i am not sure i understand...
normally if we import a BOM we want to import version constraints on artifacts... we dont import the artifacts themselves.
you can selectively use or not use artifacts from the bill but if you use one of them it will be used with the constrained version.

so i dont fully understand how importing these version constraints here can make the Iceberg Java 8 CI fail, unless something somewhere depends on `jboss-logging` directly and through the version management constraints from nessie, that version gets upgraded to a newer version unintentionally.

did we by any chance analyse the dependency tree on the iceberg build to pinpoint where the `jboss-logging` dependency gets pulled in?
i am guessing that something else was leaking here, possible these dependencies we are now removing:
```
versioned/spi/build.gradle.kts
33:  compileOnly(platform("io.quarkus:quarkus-bom"))
45:  testImplementation(platform("io.quarkus:quarkus-bom"))

versioned/persist/adapter/build.gradle.kts
33:  implementation(platform("io.quarkus:quarkus-bom"))
(...)
```
had the wrong scope or the like.

> quarkus-bom's pulled in as an enforcedPlatform where really needed - to in doubt it "wins"

ok, then the risk of incompatibilities is probably minimized even when those things to out of sync, so it is fine for now i guess

XN137 (approved):

build.gradle.kts
snazy: The "root cause" as likely me during the migration to the Gradle build - by adding the `quarkus-bom` in the projects changed in this PR.
The `quarkus-bom` ends in consuming projects as a Gradle dependency-constraint. Since `quakus-bom` also manages e.g. `jboss-logging` to 3.5.0.Final (which requires Java 11), `:iceberg-nessie:test` also gets `jboss-logging:3.5.0.Final`.

You can investigate it by updating the Nessie version in Iceberg's `versions.props` to `0.40.2` and running `./gradlew :iceberg-nessie:dependencies --configuration testRuntimeClasspath` - there you'll see that `jboss-logging` gets managed to `3.5.0.Final` and a bunch of lines above that that version management comes in via `quarkus-bom`.

PR #4846, opened 2022-08-02
Add `@MustBeClosed` on methods that return closeable `Stream`s

snazy:


versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -178,6 +179,7 @@ public Stream<CommitLogEntry> commitLog(Hash offset) throws ReferenceNotFoundExc
     ConnectionWrapper conn = borrowConnection();
     boolean failed = true;
     try {
+      @SuppressWarnings({"StreamResourceLeak", "MustBeClosedChecker"})
```
XN137, line 182: shouldnt this method also be annotated with `@MustBeClosed` ?
we are even returning `return intLog.onClose(conn::close);` so this method itself is adding on more "cleanup-responsibility" to the inner stream

same comment for all other suppression added to this file

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -473,19 +476,21 @@ private Hash refToHash(Ref ref) throws ReferenceNotFoundException {
   }
 
   @Override
+  @SuppressWarnings("MustBeClosedChecker")
```
XN137, line 479: instead of suppressing we should further delegate the `@MustBeClosed` on this method, no?

if there is a good reason for why we dont, then we should add this as a comment after the suppression

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferenceNotFound.java
```diff
@@ -76,6 +76,7 @@ void run(VersionStore<BaseContent, CommitMessage, BaseContent.Type> store)
     }
   }
 
+  @SuppressWarnings({"StreamResourceLeak", "MustBeClosedChecker"})
```
XN137, line 79: does this really need `StreamResourceLeak` ?
according to https://errorprone.info/bugpattern/StreamResourceLeak this is only for using the `Files` class.

regarding `MustBeClosedChecker` we should comment on when it is objectively not needed versus we are skipping it due to convenience (i assume that is what we do here, since these are for tests?)

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1918,4 +1868,26 @@ public void deleteAttachments(Stream<ContentAttachmentKey> keys) {
...
+    @MustBeClosed
+    Stream<R> process(ConnectionWrapper conn) throws E;
+  }
+
+  private <R, E extends Exception> Stream<R> withConnectionWrapper(ThrowingStreamResult<R, E> x)
+      throws E {
+    ConnectionWrapper conn = borrowConnection();
+    boolean failed = true;
+    try {
+      @SuppressWarnings("MustBeClosedChecker")
```
XN137, line 1883: am i missing something or why are we suppressing this check here when we intentionally added the `@MustBeClosed` above? this is the only place where `process` is getting called, right?

did we forget to annotate this helper-method with `@MustBeClosed` itself?

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -175,18 +176,7 @@ public Map<Key, ContentAndState<ByteString>> values(
...
-      Stream<CommitLogEntry> intLog = readCommitLogStream(conn, offset);
-
-      failed = false;
-      return intLog.onClose(conn::close);
-    } finally {
-      if (failed) {
-        conn.close();
-      }
-    }
+    return withConnectionWrapper(conn -> readCommitLogStream(conn, offset));
```
XN137, line 179: `withConnectionWrapper` is a nice solution to reduce the redundant code!

however it feels like `commitLog` and the other methods should still be annotated with `@MustBeClosed` in this file?

or is this correctly "inherited" through the `AbstractDatabaseAdapter` base class?
if yes, this seems to disagree with what we are doing in `NonTransactionalDatabaseAdapter`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -863,6 +886,7 @@ public final CommitLogEntry fetchFromCommitLog(OP_CONTEXT ctx, Hash hash) {
   protected abstract CommitLogEntry doFetchFromCommitLog(OP_CONTEXT ctx, Hash hash);
 
   @Override
+  @SuppressWarnings("MustBeClosedChecker")
   public Stream<CommitLogEntry> scanAllCommitLogEntries() {
```
XN137, line 890: isnt this missing `@MustBeClosed` ?
because again we are adding additional cleanup logic to the returned stream

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
snazy: Not sure how errorprone exactly determines that. The detection feels a bit "broken" - but OTOH it's due to how annotations are inherited (they're not) So it seems taht errorprone requies an explicit `@MustBeClosed` in some situations, even if the declaring base method already has a  `@MustBeClosed`.

Another actual bug is that errorrpone requires an `@SuppressWarnings("MustBeClosedChecker")` for e.g. `Stream.flatMap()` - returning a closed the stream from `flatMap()` is just ... wrong.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: the interface has it

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
snazy: yup

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
XN137: in the other comment you mention that annotations are not inherited... but for interfaces this works?
idk, i would learn torwards just adding them here as well to be on the safe side... but up to you.

XN137 (approved):
thanks for the updates, should be a nice improvement overall

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: call-sites usually refer to the interface - so they "use" the annotation

PR #4857, opened 2022-08-03
Use constructors instead of builder-instance

snazy:
Immutables allows the generation of `.of(...)` methods that directly
invoke the constructor, bypassing a temporary builder-object instance
and the required copying.

The simple use cases have been changed in this PR, leaving the use
cases that actually leverage having a "composable builder".

### DIFF
--- a/model/src/main/java/org/projectnessie/model/BaseMergeTransplant.java
+++ b/model/src/main/java/org/projectnessie/model/BaseMergeTransplant.java
@@ -66,8 +66,10 @@ public interface BaseMergeTransplant {
   @JsonSerialize(as = ImmutableMergeKeyBehavior.class)
   @JsonDeserialize(as = ImmutableMergeKeyBehavior.class)
   interface MergeKeyBehavior {
+    @Value.Parameter(order = 1)
     ContentKey getKey();
 
+    @Value.Parameter(order = 2)
     MergeBehavior getMergeBehavior();
 
     static ImmutableMergeKeyBehavior.Builder builder() {
@@ -75,7 +77,7 @@ public interface BaseMergeTransplant {
     }
 
     static MergeKeyBehavior of(ContentKey key, MergeBehavior mergeBehavior) {
-      return builder().key(key).mergeBehavior(mergeBehavior).build();
+      return ImmutableMergeKeyBehavior.of(key, mergeBehavior);
     }
   }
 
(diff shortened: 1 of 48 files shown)


model/src/main/java/org/projectnessie/model/BaseMergeTransplant.java
```diff
@@ -66,16 +66,18 @@ public interface BaseMergeTransplant {
...
     ContentKey getKey();
 
+    @Value.Parameter(order = 2)
     MergeBehavior getMergeBehavior();
 
     static ImmutableMergeKeyBehavior.Builder builder() {
       return ImmutableMergeKeyBehavior.builder();
     }
 
     static MergeKeyBehavior of(ContentKey key, MergeBehavior mergeBehavior) {
```
ajantha-bhat, line 79: Question: Do we really need one more wrapper on top of the existing ImmutableMergeKeyBehavior.of()? why can't we directly use ImmutableMergeKeyBehavior.of() ?

model/src/main/java/org/projectnessie/model/Branch.java
```diff
@@ -33,8 +35,15 @@
...
+  @Override
+  @NotBlank
+  @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
+  @Value.Parameter(order = 1)
+  String getName();
+
   @Override
   @Nullable
+  @Value.Parameter(order = 2)
   String getHash();
```
dimas-b, line 47: IMHO, using `of` method parameters of the same type (`String` in this case) is prone to mistakes... Are the runtime savings great enough to risk that?

To clarify: I'm ok with `of` methods where parameters are of different types.

dimas-b: In this particular case, this is a moot point since we already have an `of(String, String)` method ;)

model/src/main/java/org/projectnessie/model/IcebergTable.java
```diff
@@ -56,24 +56,34 @@
...
+  @Value.Parameter(order = 4)
   public abstract int getSchemaId();
 
   /** Corresponds to Iceberg's {@code defaultSpecId}. */
+  @Value.Parameter(order = 5)
   public abstract int getSpecId();
 
   /** Corresponds to Iceberg's {@code defaultSortOrderId}. */
+  @Value.Parameter(order = 6)
   public abstract int getSortOrderId();
```
dimas-b, line 87: How will generated `of` method behave when parameters are added/removed? I guess that would be addition risk of confusion in API classes since `javac` will only validate the total number of args and that the types match.

PR #4867, opened 2022-08-04
Fix issue when looking up key in an open-addressing-key-list

snazy:
A rather sad/stupid bug: `FetchValuesUsingOpenAddressing#segment`
used the _modulo_ operation to calculate the open-addressing bucket
index, but `KeyListBuildState#finish` uses binary-_and_.

Updating the code in `FetchValuesUsingOpenAddressing#segment`
fixes the lookup issue.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -40,9 +40,9 @@ final class FetchValuesUsingOpenAddressing {
 
   final int[] keyListSegmentOffsets;
   final int keyListCount;
+  final int segmentMask;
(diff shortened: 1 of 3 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -69,9 +69,7 @@ int segmentForKey(Key key) {
   }
 
   private int segment(int num) {
-    // This is compatible to 'num & (segmentCount-1)' (if segmentCount is a power of 2).
-    int mod = num % segmentCount;
-    return num >= 0 ? mod : segmentCount + mod;
+    return num & segmentMask;
```
ajantha-bhat, line 72: nit: This can be inlined in the callers?

ajantha-bhat (approved):
Didn't fully explored the `open-addressing-key-list` feature. 

Manually verified the testcase without the fix and reviewed the fix.

PR #4876, opened 2022-08-05
Adopt error-prone-rules to v2.15.0

snazy:
Add new rules to errorprone-rules.properties.

Disables new `CanIgnoreReturnValueSuggester`, which causes a lot of false-positive warnings.


XN137 (approved):
lgtm

> Disables new CanIgnoreReturnValueSuggester, which causes a lot of false-positive warnings.

really false positives or just suggestions we do not care about right now?

PR #4877, opened 2022-08-05
CI: Read-only cache usage for main-native + jackson

snazy:


.github/workflows/main.yml
```diff
@@ -133,6 +134,7 @@ jobs:
       - name: Jackson compatibility tests / 2.6.4 for Spark 2.4.4
         uses: gradle/gradle-build-action@v2
         with:
+          cache-read-only: true
```
XN137, line 137: seems unintentional that all other jackson versions below still write to the cache?

XN137 (commented):
whats the rationale ?
too large cache size currently ?

.github/workflows/main.yml
snazy: Only the first usage of the gradle-action pulls the cache

snazy:
> whats the rationale ?

otherwise (without this change) the native + jackson jobs also push their new cache - which conflicts with the cache pushed by the "main" job.

.github/workflows/main.yml
XN137: ok, adding it would be redundant but maybe also more consistent?

XN137 (approved):

PR #4878, opened 2022-08-05
IDE: Also detect Eclipse IDE

snazy:
IDE "detection" currently works fine for IntelliJ, but not for
Eclipse and related (VSCode) IDEs.

Gradle "command line" builds run for all Spark + Scala combinations,
reusing the same source folders for all Scala versions. But IDEs are
quite allergic when one source folder is used for multiple projects.
Therefore only one Scala version (usually 2.12) is "exposed" to IDEs.
This worked fine for IntelliJ, but did not for Eclipse.

This change adds Eclipse to the IDE "detection". (Sadly it has to be
duplicated in a few places.)

### DIFF
--- a/buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
@@ -49,6 +49,7 @@ plugins {
   id("org.projectnessie.buildsupport.spotless")
   id("org.projectnessie.buildsupport.errorprone")
   id("org.projectnessie.buildsupport.publishing")
+  `eclipse`
 }
 
 configureJava()
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -146,13 +146,16 @@ fun loadProjects(file: String) {
 
 loadProjects("gradle/projects.main.properties")
 
-val ideaSyncActive = System.getProperty("idea.sync.active").toBoolean()
+val ideSyncActive =
+  System.getProperty("idea.sync.active").toBoolean() ||
+    System.getProperty("eclipse.product") != null ||
+    gradle.startParameter.taskNames.any { it.startsWith("eclipse") }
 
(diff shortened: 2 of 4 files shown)


buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
```diff
@@ -49,6 +49,7 @@ plugins {
   id("org.projectnessie.buildsupport.spotless")
   id("org.projectnessie.buildsupport.errorprone")
   id("org.projectnessie.buildsupport.publishing")
+  `eclipse`
```
XN137, line 52: just wondering: what does adding this plugin here do for us?

settings.gradle.kts
```diff
@@ -146,13 +146,16 @@ fun loadProjects(file: String) {
 
 loadProjects("gradle/projects.main.properties")
 
-val ideaSyncActive = System.getProperty("idea.sync.active").toBoolean()
+val ideSyncActive =
+  System.getProperty("idea.sync.active").toBoolean() ||
+    System.getProperty("eclipse.product") != null ||
+    gradle.startParameter.taskNames.any { it.startsWith("eclipse") }
```
XN137, line 152: just wondering:
could the IDEA detection be part of https://github.com/projectnessie/gradle-build-plugins somehow?

XN137 (approved):
lgtm

i bet this was fun to figure out :grimacing:

buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
snazy: It generates the Eclipse files per project (during a `./gradlew eclipse`)

settings.gradle.kts
snazy: Nope - well, maybe.
Would need to build two plugins (one for a build.gradle + one for a settings.gradle).

Issue #3832, opened 2022-04-04
Sophisticated content-id validation


snazy:
Nessie should perform a couple more checks and some checks in a different way:

* every `Put` operation for a _new_ content key must have an _empty_/no `expectedContent`
* every `Put` operation for an _existing_ content key must have a _non-empty_ `expectedContent`
* verify that the content id of the `content` in a `Put` operation is equal to the content id of a _non-empty_ `expectedContent`

Depending on the content type, Nessie should perform this check as well:
* for a non-empty `expectedContent`, verify that the content id already exists

Also move checks from `PersistVersionStore.commit()` to `AbstractDatabaseAdapter`

dalaro:
I'm assuming this issue describes checks on `Put` operations within commits, otherwise my comment here is probably way off the mark.

I think some or all of this may have since been implemented in [`PersistVersionStore#commit`](https://github.com/projectnessie/nessie/blob/main/versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java#L119-L154) by 1654cf5fe51d969fabe6c071d1a080856240e744.

While writing some tests to exercise these checks in [an experimental branch](https://github.com/dalaro/nessie/tree/put-content-id-checks) in my fork, I found at least one case where I was unsure about the target behavior.  When a content key already exists, a new commit containing a `Put` operation on that key can specify an arbitrary value for both the expected content-id and the content-id to be written by the op.  As long as this arbitrary value is specified exactly the same as both expectation and to-write, it need not match the content-id of whatever's currently resident at the key.  This is a tricky case for me to articulate, so maybe [these fifteen lines](https://github.com/dalaro/nessie/blob/put-content-id-checks/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRestInvalidWithHttp.java#L597-L611) better convey it:

```java
    /*
     * Creates an Operations singleton.  The sole element is a Put.
     * The expected and to-be-written content IDs are both "fake id".
     * This intentionally mismatches the server-assigned UUID of the existing content at this key.
     */
    Operations putWithMatchingFakeContentIds =
        getPutTableOpSingleton(
            "fake id", "fake id", contentKey, "put with matching fake content ids");
    // TODO should this fail?
    getHttpClient()
        .newRequest()
        .path("trees/branch/{branchName}/commit")
        .resolveTemplate("branchName", branchName)
        .queryParam("expectedHash", getHeadHash(branchName))
[...]

PR #4890, opened 2022-08-08
Minor open-addressing key-list hashing changes

dalaro:
Hi @snazy.  Sorry I didn't review #4867 in time.  I saw the review ping, but I was too slow on the turnaround.

The code you wrote for open-addressing hashing of key-lists is clean and fascinating.  To understand the fix you pinged me about, I also looked the hash implementation generally (#4536) for background.  This is very nice.

#4867 looks good to me.  I may have found some minor tangential fixes to complement it.  I think there's basically one meaningful fix in here at most, which is described the first bullet point below.  The rest is either related to nomenclature consistency (bucket vs segment), testing or fixing extremely unlikely edge cases involving severe misconfiguration, or documentation and test additions.

I apologize for the size.  It's mostly tests and comments.  Furthermore, the only (intentional) substantive functionality changes are confined to `FetchValuesUsingOpenAddressing`, where the file-level diffstat is 60 lines.

_(With that, let me switch over to the PR branch's sole commit message, reproduced below)_

This commit makes a cluster of intertwined changes related to storage of keylists using open-addressing hashing:

* Fixes lookups of keys which hash to the highest-indexed segment of an  open-addressed key-list, but which physically reside in another   segment due to collisions (overwhelmingly likely to reside in the  embedded segment).  Also adds a unit test and integration test  exercising this case.  The unit test is  `TestFetchValuesUsingOpenAddressing#segmentWrapAroundWithPresentKey`,  which fails on `FetchValuesUsingOpenAddressing` from main (expect to  fix a method rename when mixing revs though: `getKeyListSegmentCount`  becomes `getKeyListBucketCount` in this commit).

* Attempts to make the word "segment" consistently refer to a collection  of contiguous hashtable cells, corresponding to either the embedded  keylist or a non-embedded keylist entity
* Attempts to make the word "bucket" consistently refer to a single cell  in the virtual keylist hashtable spread across all of the segments.  This extends to persist.proto, where I renamed a field, but did not  alter any tags, so it should remain wire-compatible.

* Made `FetchValuesUsingOpenAddressing#segmentForKey(bucket, round)` stop  clamping segment indices using the open addressing bucket index mask.  Instead, it sums the segment index and round, then applies the  remainder operator using the total segment count (embedded + not) as  the divisor.

* Added tests exercising some pathological edge cases.  Although these  elicited some failures against main, i.e. by checking out  `FetchValuesUsingOpenAddressing` from main, the importance is debatable.  Getting into these states generally requires major misconfiguration.  One case involving segment index overrun exercised by the new  `segmentWrapAround` test method might have come up, but the rest seem  low-probability.

* Made javadoc additions and changes on the read and write paths.

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -16,6 +16,7 @@
 package org.projectnessie.versioned.persist.adapter.spi;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -86,6 +87,14 @@ final class FetchValuesUsingOpenAddressing {
    * for {@code remainingKeys}.
    */
   List<Hash> entityIdsToFetch(int round, int prefetchEntities, Collection<Key> remainingKeys) {
+    // prefetchEntities is user-configurable.  Throw an exception if it is negative, otherwise clamp
+    // the nonnegative value to the range [0, keyListCount-1] = [0, keyListArray.length-1].
+    Preconditions.checkArgument(
+        0 <= prefetchEntities,
+        "Key-list segment prefetch parameter %s cannot be negative",
+        prefetchEntities);
+    prefetchEntities = Math.min(prefetchEntities, keyListCount - 1);
+
     // Identify the key-list segments to fetch
     List<Hash> entitiesToFetch = new ArrayList<>();
     for (Key key : remainingKeys) {
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -94,8 +94,6 @@ final class FetchValuesUsingOpenAddressing {
 
       for (int prefetch = 0; ; prefetch++) {
         if (segment > 0) {
-          checkSegmentIndexValidity(segment, key);
-
           // Decrement accounts for the embedded key-list segment
           int entitySegment = segment - 1;
 
@@ -148,8 +146,6 @@ final class FetchValuesUsingOpenAddressing {
         }
       }
 
-      checkSegmentIndexValidity(segment, key);
-
       KeyListEntry[] keys = keyListsArray[segment];
       for (int i = offsetInSegment; ; i++) {
         KeyListEntry keyListEntry = i < keys.length ? keys[i] : null;
@@ -166,17 +162,5 @@ final class FetchValuesUsingOpenAddressing {
       }
     }
     return keysForNextRound;
-  }
-
-  private void checkSegmentIndexValidity(final int segment, final Key key) {
-    // We expect segmentForKey to wrap in case the current round number is large enough to overrun
-    // the segment list, so this shouldn't happen
-    if (keyListsArray.length <= segment) {
-      throw new IllegalStateException(
-          String.format(
-              "Expected open-addressing key-list segment index between "
-                  + "0 (inclusive) and %s (exclusive), but computed invalid index %s for key %s",
-              keyListsArray.length, segment, key));
-    }
   }
 }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.versioned.persist.adapter.spi;
 
+import com.google.common.annotations.VisibleForTesting;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -32,15 +33,16 @@ import org.projectnessie.versioned.persist.adapter.KeyListEntry;
  * Processes {@link CommitLogEntry#getKeyList() embedded} and {@link CommitLogEntry#getKeyListsIds()
  * external} key-lists written by {@link KeyListBuildState}.
  *
- * <p>This implementation can process arbitrary numbers of {@link
- * CommitLogEntry#getKeyListSegmentCount() segments}, the number of segments does not need to be a
- * power of 2 here.
+ * <p>This implementation can process arbitrary numbers of segments, the number of segments does not
+ * need to be a power of 2 here.
  */
 final class FetchValuesUsingOpenAddressing {
 
   final int[] keyListSegmentOffsets;
   final int keyListCount;
-  final int segmentMask;
+  final int openAddressingMask;
+  // The major dimension is sized to match the number of segments (including embedded)
+  // The zeroth element corresponds with the embedded segment
   final KeyListEntry[][] keyListsArray;
   final Object2IntHashMap<Hash> entityIdToSegment;
   final List<Hash> keyListIds;
@@ -58,28 +60,25 @@ final class FetchValuesUsingOpenAddressing {
 
     entityIdToSegment = new Object2IntHashMap<>(keyListCount, Hashing.DEFAULT_LOAD_FACTOR, -1);
 
-    segmentMask = entry.getKeyListSegmentCount() - 1;
+    openAddressingMask = entry.getKeyListBucketCount() - 1;
 
     keyListIds = entry.getKeyListsIds();
   }
 
   /** Calculates the open-addressing bucket for a key. */
-  int segmentForKey(Key key) {
-    return segment(key.hashCode());
-  }
-
-  private int segment(int num) {
-    return num & segmentMask;
+  int bucketForKey(Key key) {
+    return key.hashCode() & openAddressingMask;
   }
 
   /**
    * Identifies the segment for a bucket. Segment 0 is the embedded key list, segment 1 is the first
    * key-list-entity.
    */
+  @VisibleForTesting
   int segmentForKey(int bucket, int round) {
     int binIdx = Arrays.binarySearch(keyListSegmentOffsets, bucket);
     int segment = binIdx >= 0 ? binIdx + 1 : -binIdx - 1;
-    return segment(segment) + round;
+    return (segment + round) % keyListsArray.length;
   }
 
   /**
@@ -90,17 +89,19 @@ final class FetchValuesUsingOpenAddressing {
     // Identify the key-list segments to fetch
     List<Hash> entitiesToFetch = new ArrayList<>();
     for (Key key : remainingKeys) {
-      int keyBucket = segmentForKey(key);
+      int keyBucket = bucketForKey(key);
       int segment = segmentForKey(keyBucket, round);
 
       for (int prefetch = 0; ; prefetch++) {
         if (segment > 0) {
+          checkSegmentIndexValidity(segment, key);
+
+          // Decrement accounts for the embedded key-list segment
           int entitySegment = segment - 1;
-          if (keyListIds.size() > entitySegment) {
-            Hash entityId = keyListIds.get(entitySegment);
-            if (entityIdToSegment.put(entityId, segment) == -1) {
-              entitiesToFetch.add(entityId);
-            }
+
+          Hash entityId = keyListIds.get(entitySegment);
+          if (entityIdToSegment.put(entityId, segment) == -1) {
+            entitiesToFetch.add(entityId);
           }
         }
         if (prefetch >= prefetchEntities) {
@@ -131,8 +132,13 @@ final class FetchValuesUsingOpenAddressing {
   Collection<Key> checkForKeys(
       int round, Collection<Key> remainingKeys, Consumer<KeyListEntry> resultConsumer) {
     List<Key> keysForNextRound = new ArrayList<>();
+    // If we've already examined every segment (implying a miss on a completely full hashmap),
+    // then return an empty collection early
+    if (keyListsArray.length <= round) {
+      return keysForNextRound;
+    }
     for (Key key : remainingKeys) {
-      int keyBucket = segmentForKey(key);
+      int keyBucket = bucketForKey(key);
       int segment = segmentForKey(keyBucket, round);
       int offsetInSegment = 0;
       if (round == 0) {
@@ -142,9 +148,7 @@ final class FetchValuesUsingOpenAddressing {
         }
       }
 
-      if (segment >= keyListsArray.length) {
-        continue;
-      }
+      checkSegmentIndexValidity(segment, key);
 
       KeyListEntry[] keys = keyListsArray[segment];
       for (int i = offsetInSegment; ; i++) {
@@ -162,5 +166,17 @@ final class FetchValuesUsingOpenAddressing {
       }
     }
     return keysForNextRound;
+  }
+
+  private void checkSegmentIndexValidity(final int segment, final Key key) {
+    // We expect segmentForKey to wrap in case the current round number is large enough to overrun
+    // the segment list, so this shouldn't happen
+    if (keyListsArray.length <= segment) {
+      throw new IllegalStateException(
+          String.format(
+              "Expected open-addressing key-list segment index between "
+                  + "0 (inclusive) and %s (exclusive), but computed invalid index %s for key %s",
+              keyListsArray.length, segment, key));
+    }
   }
 }
--- a/versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
@@ -1,0 +1,294 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.persist.adapter.spi;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
+
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.versioned.Hash;
+import org.projectnessie.versioned.Key;
+import org.projectnessie.versioned.persist.adapter.CommitLogEntry;
+import org.projectnessie.versioned.persist.adapter.ImmutableCommitLogEntry;
+import org.projectnessie.versioned.persist.adapter.ImmutableContentId;
+import org.projectnessie.versioned.persist.adapter.ImmutableKeyList;
+import org.projectnessie.versioned.persist.adapter.ImmutableKeyListEntity;
+import org.projectnessie.versioned.persist.adapter.ImmutableKeyListEntry;
+import org.projectnessie.versioned.persist.adapter.KeyListEntity;
+import org.projectnessie.versioned.persist.adapter.KeyListEntry;
+
+public class TestFetchValuesUsingOpenAddressing {
+
+  /**
+   * Test finding non-existent keys pushed out of the final (highest-indexed) segment by collisions.
+   *
+   * <p>This implies sufficient collisions to cause an insertion's bucket index to overrun the total
+   * bucket count, effectively "wrapping" around the segment and bucket indices to the beginning
+   * (i.e. the embedded segment). We'll also simulate exceptionally unlikely cases involving probing
+   * through multiple segments before finding an open slot, though this generally implies a
+   * misconfigured load factor and target size.
+   */
+  @Test
+  public void segmentWrapAroundWithAbsentKey() {
+    // Create four segments with two buckets each
+    // The hashes of the non-embedded/external segments will be 01, 02, and 03 (as hex strings)
+    FetchValuesUsingOpenAddressing fetch =
+        new FetchValuesUsingOpenAddressing(getCommitFixture(2, 4));
+
+    // Make a key that will hash to the largest value representable within the bucket mask
+    Key nonExistentKey =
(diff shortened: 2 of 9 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -90,17 +89,19 @@ List<Hash> entityIdsToFetch(int round, int prefetchEntities, Collection<Key> rem
...
           int entitySegment = segment - 1;
-          if (keyListIds.size() > entitySegment) {
-            Hash entityId = keyListIds.get(entitySegment);
-            if (entityIdToSegment.put(entityId, segment) == -1) {
-              entitiesToFetch.add(entityId);
-            }
+
+          Hash entityId = keyListIds.get(entitySegment);
+          if (entityIdToSegment.put(entityId, segment) == -1) {
+            entitiesToFetch.add(entityId);
```
dalaro, line 104: I'm removing the `keyListIds.size() > entitySegment` condition because I'm now relying on `segmentForKey` to ensure that property (wrapping if necessary), and confirming it just above in `checkSegmentIndexValidity`.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -58,28 +60,25 @@ final class FetchValuesUsingOpenAddressing {
...
   /**
    * Identifies the segment for a bucket. Segment 0 is the embedded key list, segment 1 is the first
    * key-list-entity.
    */
+  @VisibleForTesting
   int segmentForKey(int bucket, int round) {
     int binIdx = Arrays.binarySearch(keyListSegmentOffsets, bucket);
     int segment = binIdx >= 0 ? binIdx + 1 : -binIdx - 1;
-    return segment(segment) + round;
+    return (segment + round) % keyListsArray.length;
```
dalaro, line 81: I'm very curious about this line and whether you think this looks valid, since I know it just went through an edit in #4867.  However, the old behavior of ANDing the segment index against the bucket index mask seemed like a no-op under sane configurations, and could actually trigger an erroneous miss in a pathologically bad configuration (exercised by one of the new ITs).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -163,4 +167,16 @@ Collection<Key> checkForKeys(
     }
     return keysForNextRound;
   }
+
+  private void checkSegmentIndexValidity(final int segment, final Key key) {
+    // We expect segmentForKey to wrap in case the current round number is large enough to overrun
+    // the segment list, so this shouldn't happen
+    if (keyListsArray.length <= segment) {
+      throw new IllegalStateException(
```
snazy, line 175: Nit: `Preconditions.checkState` (provides string-formatting)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -58,28 +60,25 @@ final class FetchValuesUsingOpenAddressing {
...
   }
 
   /** Calculates the open-addressing bucket for a key. */
-  int segmentForKey(Key key) {
-    return segment(key.hashCode());
-  }
-
-  private int segment(int num) {
-    return num & segmentMask;
+  int bucketForKey(Key key) {
```
snazy, line 69: Nice rename!

versioned/persist/serialize-proto/src/main/proto/persist.proto
```diff
@@ -43,7 +43,7 @@ message CommitLogEntry {
   // Open-addressing bucket offsets. Same number of elements as key_list_ids.
   repeated int32 key_list_entity_offsets = 13;
   optional float key_list_load_factor = 14;
-  optional int32 key_list_segment_count = 15;
+  optional int32 key_list_bucket_count = 15;
```
snazy, line 46: Nice rename!

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestKeyListBuildState.java
```diff
@@ -154,6 +155,25 @@ static Stream<Arguments> openAddressing() {
...
+    // used for an int's shift distance, so the effective shift distance is 0.
+    assertThat(KeyListBuildState.nextPowerOfTwo(0)).isEqualTo(1);
+
+    assertThat(KeyListBuildState.nextPowerOfTwo(1)).isEqualTo(1);
+    assertThat(KeyListBuildState.nextPowerOfTwo((1 << 29) + 1)).isEqualTo(1 << 30);
+    assertThat(KeyListBuildState.nextPowerOfTwo((1 << 30) - 1)).isEqualTo(1 << 30);
+    assertThat(KeyListBuildState.nextPowerOfTwo(1 << 30)).isEqualTo(1 << 30);
+    assertThat(KeyListBuildState.nextPowerOfTwo((1 << 30) + 1)).isEqualTo(1 << 31);
+    assertThat(KeyListBuildState.nextPowerOfTwo(1 << 31)).isEqualTo(1 << 31);
+    // TODO might just want to throw an exception early in the following cases, in nextPowerOfTwo
```
snazy, line 171: Duh, yes! Please let it either throw an exception, before it becomes a negative value.

snazy (commented):
Nice PR!
WIll do a another thorough review, but the changes look good and correct from a brief look.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
snazy: Yea, you're right here - also your assumption "b/c of bad nomenclature" is also right.
`(segment + round) % keyListsArray.length` looks correct to me now - and this mod-operation is probably the reason why it's incorrectly been a mod-operation in the above methods above.
TL;DR very good catch!

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
snazy, line 175: Is there any way we can easily trigger this in a test? No worries if not.

snazy (commented):
The change looks really good!
Couple of nits (not fighting for those) and a few comments.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
dalaro: This was a thought-provoking comment.  The only ways I could induce this exception failure involve modifying this particular file.  After looking at it for a bit, I couldn't do it by limiting myself to only caller behavior, even with an arbitrarily bad/corrupt commit log entry.  Although I like defensive preconditions in general, what I've added here seems like overkill, since any potential violation of this invariant is likely to happen in the same (relatively compact) file where it is checked.  I've just deleted the precondition entirely, relying instead on `segmentForKey`'s new remainder operation with `keyListsArray.length` as divisor.  Though I can certainly bring it back the precondition, if you prefer.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -90,17 +89,19 @@ List<Hash> entityIdsToFetch(int round, int prefetchEntities, Collection<Key> rem
     // Identify the key-list segments to fetch
     List<Hash> entitiesToFetch = new ArrayList<>();
```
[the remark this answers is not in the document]
dalaro: Good idea.  I've added:
* a range clamp for `prefetchEntities` on the interval `[0, keyListCount-1]`
* an `IllegalArgumentException` when `prefetchEntities` is negative
* tests to exercise nonnegative cases (including "wraparound" `prefetchEntities` that exercises the clamp) and the exception

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractManyKeys.java
```diff
@@ -343,4 +350,168 @@ void pointContentKeyLookups(
...
+   * int)} computing the wrong segment index for the final key, since it clamps a segment index to
+   * the bucket index interval. Applying the open-addressing bucket mask to the segment index is
+   * normally a no-op, except in the extreme edge case described above.
+   *
+   * <p>It was possible to induce this bug without setting the goal sizes all the way down to zero.
+   * It just required values small enough to generate a size-zero embedded key list and a series of
+   * size-one external key lists after that.
+   *
+   * <p>This case is covered with greater specificity in {@linkplain
+   * TestFetchValuesUsingOpenAddressing#segmentWrapAround}.
```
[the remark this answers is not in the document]
dalaro: I actually didn't realize IDEA would handle this case.  I saw that I couldn't reference on `TestFetchValuesUsingOpenAddressing` from `AbstractManyKeys` given the current dependency tree, then settled for a shorter name, but I see IDEA has no problem searching rest of the project index to find it correctly once I qualify it.  Thanks!

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestKeyListBuildState.java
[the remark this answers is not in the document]
dalaro: Agreed, check and parameterized test (`nextPowerOfTwoBucketCountException`) added with some negative values and values exceeding 2^30

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+        .keyList(embeddedKeyList)
+        .keyListBucketCount(segmentCount * segmentSize)
+        .keyListsIds(keyListIds)
+        .hash(Hash.of("c0ffee"))
+        .build();
+  }
+
+  /** Convert an int to a hex string, left-padding with 0 until it has even length. */
+  private static String intToPaddedHexString(int i) {
+    String hexString = Integer.toHexString(i);
```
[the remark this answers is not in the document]
dalaro: Agreed, that's simpler and more legible.  I've added a precondition to make sure the input is in unsigned byte range, and added that to the javadoc.  That's more than enough for all of my uses already.

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+   * various +1 and -1 offsets in play for nonexact matches. This just helped me prove that it was
+   * correct to begin with, and should localize our search if we accidentally break it in the
+   * future.
+   */
+  @Test
+  public void segmentForKey() {
+    FetchValuesUsingOpenAddressing fetch =
+        new FetchValuesUsingOpenAddressing(getCommitFixture(2, 4));
+
+    assertEquals(0, fetch.segmentForKey(0, 0));
```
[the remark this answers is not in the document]
dalaro: Can do.  I've rolled up both this loop and the next one in the file, which had a similar comment.

After rolling up the implied loops in both of these two tests, the two tests are nearly identical.  The next natural step is to merge them into a single parameterized test, where the sole parameter is the size of the segments as measured in buckets (1 or 2).  I've also done that.

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+   * <p>This should only happen if a user misunderstands what the open-addressing-related config
+   * options actually do, or if they manage to serialize abnormally huge {@code CommitLogEntry}
+   * instances individually exceeding the segment size limit.
+   */
+  @Test
+  public void singletonSegment() {
+    FetchValuesUsingOpenAddressing fetch =
+        new FetchValuesUsingOpenAddressing(getCommitFixture(1, 4));
+
+    assertEquals(0, fetch.segmentForKey(0, 0));
```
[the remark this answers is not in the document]
dalaro: Yep (merged into previous test per comment just above).

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+
+    // Look for a non-existent key.
+    // The fake segments here are all full, so this will probe through the entire hashtable.
+    // Expect the final segment in round 0, then no segments (implying embedded) in round 1,
+    // the first non-embedded segment in round 2, etc.
+    // This has the side-effect of adding entityIdToSegment mappings, which we'll need to call
+    // entityLoaded in the next part of this test.
+    Hash hashOfFinalSegment = Hash.of("03");
+    List<Key> singleton = ImmutableList.of(nonExistentKey);
+    assertEquals(ImmutableList.of(hashOfFinalSegment), fetch.entityIdsToFetch(0, 0, singleton));
```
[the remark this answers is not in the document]
dalaro: Noted, thanks for the pointer.

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+    Key nonExistentKey =
+        new KeyWithConfigurableHash(ImmutableList.of("foo"), el -> Integer.MAX_VALUE);
+
+    // Look for a non-existent key.
+    // The fake segments here are all full, so this will probe through the entire hashtable.
+    // Expect the final segment in round 0, then no segments (implying embedded) in round 1,
+    // the first non-embedded segment in round 2, etc.
+    // This has the side-effect of adding entityIdToSegment mappings, which we'll need to call
+    // entityLoaded in the next part of this test.
+    Hash hashOfFinalSegment = Hash.of("03");
```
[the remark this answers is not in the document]
dalaro: I've inlined both instances of this

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -0,0 +1,294 @@
...
+    if (1 == hexString.length() % 2) {
+      hexString = "0" + hexString;
+    }
+    return hexString;
+  }
+
+  /**
+   * Builds a fake {@linkplain KeyListEntity}.
+   *
+   * <p>All keys appearing in this entity have acustom hashCode() that returns {@linkplain
```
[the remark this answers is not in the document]
dalaro: Good catch

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestKeyListBuildState.java
dalaro: Sounds good.  I've set the range of valid inputs to `[0, 2^30]`, with a `checkArgument` precondition, javadoc, and test (the same new parameterized test that exercises a negative-parameter-value exception, commented on further down).

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
dalaro: Agreed, I should use the variadic `checkState` overload instead of this verbose construct.  However, after reflecting on your comment about the testability of this exception elsewhere in the review, I've removed this check (with more on the thought process in that comment's thread).

snazy (approved):
LGTM!

dalaro:
I've just discovered a latent bug fixed by this PR.  While doing unrelated work in #3832 which exercises open-addresing key-list hashing, I ran batteries of put-heavy tests against an old commit which did not include this PR's changes.

One of these test configurations could throw exceptions in the following form (again, this is effectively before the PR merged):

```
java.lang.ArrayIndexOutOfBoundsException: Index 1976 out of bounds for length 1976
        at org.projectnessie.versioned.persist.adapter.spi.FetchValuesUsingOpenAddressing.checkForKeys(FetchValuesUsingOpenAddressing.java:147)
        at org.projectnessie.versioned.persist.adapter.spi.AbstractDatabaseAdapter.fetchValuesHandleOpenAddressingKeyList(AbstractDatabaseAdapter.java:1553)
        at org.projectnessie.versioned.persist.adapter.spi.AbstractDatabaseAdapter.lambda$keysForCommitEntry$45(AbstractDatabaseAdapter.java:1370)
        at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:271)
        at org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUtil$1.lambda$tryAdvance$0(DatabaseAdapterUtil.java:206)
        at org.projectnessie.versioned.persist.adapter.spi.AbstractDatabaseAdapter$1.tryAdvance(AbstractDatabaseAdapter.java:1061)
        at org.projectnessie.versioned.persist.adapter.spi.DatabaseAdapterUtil$1.tryAdvance(DatabaseAdapterUtil.java:199)
        at java.base/java.util.Spliterator.forEachRemaining(Spliterator.java:326)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
        at org.projectnessie.versioned.persist.store.PersistVersionStore.lambda$commit$1(PersistVersionStore.java:211)
        at org.projectnessie.versioned.persist.store.PersistVersionStore.lambda$commit$2(PersistVersionStore.java:247)
        at org.projectnessie.versioned.persist.adapter.spi.AbstractDatabaseAdapter.commitAttempt(AbstractDatabaseAdapter.java:215)
        at org.projectnessie.versioned.persist.tx.TxDatabaseAdapter.lambda$commit$15(TxDatabaseAdapter.java:425)
        at org.projectnessie.versioned.persist.tx.TxDatabaseAdapter.opLoop(TxDatabaseAdapter.java:881)
        at org.projectnessie.versioned.persist.tx.TxDatabaseAdapter.commit(TxDatabaseAdapter.java:417)
        at org.projectnessie.versioned.persist.adapter.spi.TracingDatabaseAdapter.commit(TracingDatabaseAdapter.java:133)
        at org.projectnessie.versioned.persist.store.PersistVersionStore.commit(PersistVersionStore.java:253)
        at org.projectnessie.versioned.TracingVersionStore.lambda$commit$3(TracingVersionStore.java:107)
        at org.projectnessie.versioned.TracingVersionStore.callWithTwoExceptions(TracingVersionStore.java:352)
        at org.projectnessie.versioned.TracingVersionStore.commit(TracingVersionStore.java:101)
        at org.projectnessie.versioned.MetricsVersionStore.lambda$commit$1(MetricsVersionStore.java:88)
        at org.projectnessie.versioned.MetricsVersionStore.delegate2ExR(MetricsVersionStore.java:269)
        at org.projectnessie.versioned.MetricsVersionStore.commit(MetricsVersionStore.java:87)
        at org.projectnessie.services.impl.TreeApiImpl.commitMultipleOperations(TreeApiImpl.java:668)
        at org.projectnessie.services.impl.TreeApiImplWithAuthorization.commitMultipleOperations(TreeApiImplWithAuthorization.java:287)
        at org.projectnessie.services.rest.RestTreeResource.commitMultipleOperations(RestTreeResource.java:154)
        at org.projectnessie.services.rest.RestTreeResource_Subclass.commitMultipleOperations$$superforward1(Unknown Source)
        at org.projectnessie.services.rest.RestTreeResource_Subclass$$function$$1.apply(Unknown Source)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.proceed(AroundInvokeInvocationContext.java:53)
        at io.quarkus.hibernate.validator.runtime.interceptor.AbstractMethodValidationInterceptor.validateMethodInvocation(AbstractMethodValidationInterceptor.java:68)
        at io.quarkus.hibernate.validator.runtime.jaxrs.JaxrsEndPointValidationInterceptor.validateMethodInvocation(JaxrsEndPointValidationInterceptor.java:35)
        at io.quarkus.hibernate.validator.runtime.jaxrs.JaxrsEndPointValidationInterceptor_Bean.intercept(Unknown Source)
        at io.quarkus.arc.impl.InterceptorInvocation.invoke(InterceptorInvocation.java:41)
        at io.quarkus.arc.impl.AroundInvokeInvocationContext.perform(AroundInvokeInvocationContext.java:40)
        at io.quarkus.arc.impl.InvocationContexts.performAroundInvoke(InvocationContexts.java:32)
        at org.projectnessie.services.rest.RestTreeResource_Subclass.commitMultipleOperations(Unknown Source)
        at org.projectnessie.services.rest.RestTreeResource_ClientProxy.commitMultipleOperations(Unknown Source)
```

I had not seen this before.  I wanted to understand the root cause, just in case the implementation was still susceptible to it despite the changes here.  I reran tests with a debugger attached and caught this situation with a conditional breakpoint.  After tripping, I found this case involved a `hashCode()` triggering an edge case in what was [formerly called `segment(int)`](https://github.com/projectnessie/nessie/blob/36a22ed87e9e2c4f4df4fb6d5460f72a4111ea1d/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java#L74):

```java
  private int segment(int num) {
    // This is compatible to 'num & (segmentCount-1)' (if segmentCount is a power of 2).
    int mod = num % segmentCount;
    return num >= 0 ? mod : segmentCount + mod;
  }
```

In the exception case, I had:

* `key.hashCode()` = -300679168
* `segmentCount`, which really described the number of buckets in the entire hash at this revision = 262144

This results in `mod = -300679168 % 262144 = 0`.  The problem is that we then return `segmentCount + 0 = segmentCount`.  The caller, `checkForKeys`, then treats this value as a 0-based indexed into the total virtual bucket space, causing it to run off the end by exactly one (accessing one index beyond the final array-element in the final segment/entity).

I think that this PR already fixed this case.  After merge, `checkForKeys` now calls [`bucketForKey`](https://github.com/projectnessie/nessie/blob/8efaf76f8ce1ceea9eebb9221fb2731e4143bdf1/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java#L64-L71) to get this same index, and that uses bitwise-and against a mask generated by taking ((total bucket count in the entire hash) - 1).

dalaro:
In closer examination, I think @snazy probably fixed this bug before I did, via https://github.com/projectnessie/nessie/pull/4867.  The version I found affected was older than his PR.  I haven't tried to reproduce the problem against 4867 specifically, but source reading suggests that 4867 should fix it, because that changeset replaced the problematic remainder operation with bitwise-and against a safe mask.

PR #4891, opened 2022-08-08
Fix missing version from content-generator tool

snazy:
The content generator tool has a command to print the Nessie version,
which didn't work anymore. This change fixes this.

### DIFF
--- a/tools/content-generator/build.gradle.kts
+++ b/tools/content-generator/build.gradle.kts
@@ -68,3 +68,10 @@ tasks.named<ShadowJar>("shadowJar") {
     attributes["Main-Class"] = "org.projectnessie.tools.contentgenerator.cli.NessieContentGenerator"
   }
 }
+
+tasks.named<Test>("intTest") { systemProperty("expectedNessieVersion", project.version) }
+
+tasks.named<ProcessResources>("processResources") {
+  inputs.property("nessieVersion", project.version)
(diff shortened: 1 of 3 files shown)


tools/content-generator/src/main/resources/org/projectnessie/tools/contentgenerator/cli/version.properties
```diff
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
-nessie.version=${nessie.version}
+nessie.version=${nessie_version}
```
XN137, line 17: nit: elsewhere we are using `nessieVersion`

tools/content-generator/build.gradle.kts
```diff
@@ -68,3 +68,5 @@ tasks.named<ShadowJar>("shadowJar") {
     attributes["Main-Class"] = "org.projectnessie.tools.contentgenerator.cli.NessieContentGenerator"
   }
 }
+
+tasks.named<ProcessResources>("processResources") { expand("nessie_version" to project.version) }
```
XN137, line 72: since this was failing missing silently, could it be worth adding an integration test?

XN137 (dismissed):
nice catch, maybe add basic smoke test to prevent this in the future

XN137 (approved):

PR #4913, opened 2022-08-12
Expose name and full key to "get entries" filters

### LINKED ISSUE #4912: Need entry's name to be available for the CEL filter.
**Description**:
I see two calls are available for entries. namespace and type. But need the entry's 'name' to be available for the nessie's entry.
For example, if the name of the entry is 'table1' in the namespace 'folder1/folder2,' it would be great if we filter the entry by the name not by the types and namespaces.

**Expected Use Cases**:
RequestBuilder.filter("entry.name == \'name\'")

dimas-b:
Expose name and full key to "get entries" filters
    
This will expose all of the available entry details to filters except
the content ID, which is an internal ID and probably should not be used
for filtering.
    
Rework the scrpt args in TreeApiImpl.filterEntries() to use CEL
wrappers.
    
Add proper CEL types for entry filter scrtips.
    
Drop VAR_NAMESPACE and VAR_CONTENT_TYPE decrarations that did not
have values even before this change.
    
Fixes #4912

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
+++ b/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
@@ -32,6 +32,7 @@ import org.projectnessie.model.Operation.Put;
 import org.projectnessie.model.RefLogResponse;
 import org.projectnessie.model.Reference;
 import org.projectnessie.model.ReferenceMetadata;
+import org.projectnessie.versioned.KeyEntry;
 
 /** A utility class for CEL declarations and other things. */
 public final class CELUtil {
@@ -45,8 +46,6 @@ public final class CELUtil {
   public static final String VAR_REF_META = "refMeta";
   public static final String VAR_COMMIT = "commit";
   public static final String VAR_ENTRY = "entry";
-  public static final String VAR_NAMESPACE = "namespace";
-  public static final String VAR_CONTENT_TYPE = "contentType";
   public static final String VAR_PATH = "path";
   public static final String VAR_ROLE = "role";
   public static final String VAR_OP = "op";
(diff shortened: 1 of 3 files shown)


servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
```diff
@@ -45,6 +45,8 @@ public final class CELUtil {
   public static final String VAR_REF_META = "refMeta";
   public static final String VAR_COMMIT = "commit";
   public static final String VAR_ENTRY = "entry";
+  public static final String VAR_CONTENT_KEY = "contentKey";
+  public static final String VAR_NAME = "name";
```
ajantha-bhat, line 49: Do we need to add them to `ENTRIES_DECLARATIONS`? Line 70

```
public static final List<Decl> ENTRIES_DECLARATIONS =
      ImmutableList.of(
          Decls.newVar(VAR_ENTRY, Decls.newMapType(Decls.String, Decls.String)),
          Decls.newVar(VAR_NAMESPACE, Decls.String),
          Decls.newVar(VAR_CONTENT_TYPE, Decls.String));
```

I know testcases are passing. I need to dig more to understand the implications if we don't add them to declarations.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -627,12 +629,17 @@ protected Stream<KeyEntry<Content.Type>> filterEntries(
           // currently this is just a workaround where we put EntriesResponse.Entry into a hash
           // structure.
           // Eventually we should just be able to do "script.execute(Boolean.class, entry)"
+          ContentKey contentKey = fromKey(entry.getKey());
           Map<String, Object> arguments =
               ImmutableMap.of(
                   VAR_ENTRY,
                   ImmutableMap.of(
+                      VAR_CONTENT_KEY,
```
snazy, line 637: Maybe expose these attributes using the same "structure" that `OperationForCel` already has, with different properties for the full string plus individual elements.

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
dimas-b: refactored.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: Good point. thx! I've refactored the entry filter script invocation to use CEL wrappers.

snazy (approved):

PR #4914, opened 2022-08-13
Add generic adapter type to DatabaseAdapterFactory

dimas-b:


compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/DatabaseAdapters.java
```diff
@@ -56,7 +59,10 @@ private DatabaseAdapters() {}
...
 
     DatabaseAdapterFactory.Builder<
-            DatabaseAdapterConfig, DatabaseAdapterConfig, DatabaseConnectionProvider<?>>
+            DatabaseAdapter,
+            DatabaseAdapterConfig,
+            DatabaseAdapterConfig,
+            DatabaseConnectionProvider<?>>
         builder = factory.newBuilder();
 
     TestConnectionProviderSource<DatabaseConnectionConfig> providerSource =
```
XN137, line 68: this is not from your change but reading through this logic here i was wondering:
factory creates a builder and from the builder we get the config and then `findCompatibleProviderSource` (internally `isCompatibleWith`) needs both the config and factory to determine whether it is compatible...

so if the factory determines the builder type and the builder the config type and compatibility is never based on values inside the config, then maybe this compatibilty check could be simplified to just use the factory.
(note that the values inside the config seem to get populated afterwards not before checking compatibilty)

XN137 (approved):
code changes lgtm so +1 assuming there was no good design reason for keeping the concrete adapter type out of the factory api.

snazy (approved):

compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/DatabaseAdapters.java
dimas-b: I agree. I was that discrepancy before, just never got around to fixing it :)

PR #4929, opened 2022-08-16
Register KeyEntryForCel for reflection inside Quarkus

dimas-b:
This is to fix native test failures from #4913

### DIFF
--- a/servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
+++ b/servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
@@ -26,5 +26,11 @@ import org.projectnessie.services.cel.CELUtil;
  * <p>In particular, this list is about Nessie model classes.
  */
 @RegisterForReflection(
-    targets = {NessieError.class, CELUtil.OperationForCel.class, CELUtil.ContentForCel.class})
+    targets = {
+      NessieError.class,
+      CELUtil.KeyedEntityForCel.class,
+      CELUtil.OperationForCel.class,
+      CELUtil.ContentForCel.class,
+      CELUtil.KeyEntryForCel.class
+    })
 public abstract class ModelReflections {}


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
```diff
@@ -26,5 +26,11 @@
  * <p>In particular, this list is about Nessie model classes.
  */
 @RegisterForReflection(
-    targets = {NessieError.class, CELUtil.OperationForCel.class, CELUtil.ContentForCel.class})
+    targets = {
+      NessieError.class,
+      CELUtil.KeyedEntityForCel.class,
```
XN137, line 31: just wondering:
`KeyedEntityForCel` is also new here, but mentioning int he commit msg is not important? :sweat_smile:

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
```diff
@@ -26,5 +26,11 @@
...
+      CELUtil.OperationForCel.class,
+      CELUtil.ContentForCel.class,
+      CELUtil.KeyEntryForCel.class
```
XN137, line 34: seems like we could prevent this from happening in the future by automatically labeling PRs when one of these files changes i.e. `CELUtil.java`...
similar to:
https://github.com/apache/iceberg/blob/master/.github/labeler.yml

wdyt?

XN137 (approved):

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
dimas-b: It is. I just mentioned the top think that was added in #4913. `KeyedEntityForCel` was a result of local refactoring, but it still has to be available for introspection.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
dimas-b: SGTM :+1: :)

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
dimas-b: updated commit message for this.

snazy (approved):

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/reflect/ModelReflections.java
dimas-b: #4931

PR #269, opened 2020-10-05
add NessieError to reflection list for native image

rymurr:
This change is [Reviewable](https://reviewable.io/reviews/projectnessie/nessie/269)

PR #4931, opened 2022-08-16
Automatically add labels for common cases requiring native tests

dimas-b:
Apparently the `pull_request_target` trigger type requires that we merge the change before it can be used (it uses the base of the pull request for running the job)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,21 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Pull Request Labeler GitHub Action Configuration: https://github.com/marketplace/actions/labeler
+pr-native:
+  # Run native tests on any quarkus-related changes
+  - servers/*quarkus*/**/*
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Pull Request Labeler GitHub Action Configuration: https://github.com/marketplace/actions/labeler
+pr-native:
+  # Run native tests on any model or quarkus-related changes (may affect java reflection operation)
+  - model/**/*
```
snazy, line 18: Why's model needed here?

snazy (commented):
Not sure whether it's necessary to run the native tests for all these changes.
The probability that a change breaks the native image isn't that high and native tests are always run on the main branch.

.github/labeler.yml
dimas-b: Mostly as a precaution. At some point we had to flag `NessieError` for reflection in native builds (#269)... It was a non-trivial dependency. I suppose it might be worth spending CI time on native build for model changes in general, because they are somewhat likely to be inspected via reflection.

dimas-b: I'd be ok to just limit the model regex to `**/NessieError.java`

XN137 (commented):
for me having a labeler encodes our "tribal knowledge" when to add which PR label into code.. so i think its worth trying out?
if we are too unhappy with the granularity we can revisit or remove it again later.

could there be a bunch of patterns we could use for the `pr-jackson` label as well?

what about the new integration tests?

dimas-b:
> could there be a bunch of patterns we could use for the pr-jackson label as well?

I do not know off the top of my head :shrug:

@snazy : Do you have any suggestions?

XN137 (approved):
looks useful to me, so +1

snazy:
Let’s try this

PR #4933, opened 2022-08-16
Make `AbstractRestRefLog.testReflog` repeatable

snazy:
REST-API based tests do not delete the underlying Nessie repository.
Since the Postgres container can be reused,
`AbstractRestRefLog.testReflog` will fail when run more than once.
Making the tag-name being asserted (mostly) unique, fixes the issue.


dimas-b (approved):
#4930 also fixes this by not reusing Postgres containers

PR #4941, opened 2022-08-17
Remove a lot of unnecessary Java type parameters

snazy:
Generics removed from:
* `StoreWorker`, uses `Content`, `Content.Type`, `CommitMeta` directly
* `VersionStore`, uses `Content`, `Content.Type`, `CommitMeta` directly
* `Commit`, uses `Content`, `CommitMeta` directly
* `Operation` + `Delete`/`Put`/`Unchanged`, uses `Content` directly
* `Diff`, uses `Content` directly
* `KeyEntry`, uses `Content.Type` directly

Changes to `Content` + `Content.Type`:
* `Content.Type` refactored to a "plain" Java interface
* Add ability to register `Content.Type`s via Java service-loader
  mechanism
* Updated custom `Content`/`Content.Type`/`CommitMeta` variants
  in `nessie-versioned-tests`, add a content-type bundle for these

Note: Weld/CDI extension for compatibility tests had to be updated
to allow (old) parameterized-type signature of `VersionStore` _and_
(new) non-parameterized-type signature.

This PR is split into two commits: one with the actual changes and another with the "boring stuff/changes" implied by the actual changes.

Follow-up PRs: #4942 #4943 #4944 #4945

Noteworthy changes in the follow-up PRs:
* `Content.Type.payload()` will be removed in a follow-up PR
* Many uses of `StoreWorker` will be eliminated

See issue #4940

Issue #4940, opened 2022-08-17
Clean up use of generics and "alien" `Content` types only for tests


snazy:
Central interfaces like `VersionStore`, `StoreWorker` and also `DatabaseAdapter` and a bunch of related Java types have an overly complex use of Java generics. The concrete Java types for content, content type and commit metadata are abstracted away. The only code that somehow requires this abstraction is test code, but that does not warrant the excessive complexity.

Aside from the above, there was always the idea to make it possible to "plug in" custom content types. Although this idea may sound unrelated, it is technically practically very related to the above cleanup.

The work for this issue will:
* Remove all the unnecessary generics from the above Java types
* Make all content objects implement `Content`
* Make `Content.Type` and content serialization extensible (there will be two bundles: one for "production" content types and one for additional test-only content types)
* Introduce a central registry for all content types (via Java's service loader mechanism)
* Remove unnecessary baggage from `StoreWorker`, which is only present due to unnecessary abstraction
* Finally completely eliminate the ability to persist new global state

Implemented with the following PR in this exact order:
1. [x] #4941
2. [x] #4942 
3. [x] #4943 
4. [x] #4944 
5. [ ] #4945

PR #4943, opened 2022-08-17
Re-add `payload` to `StoreWorker` to identify the content-type

snazy:
Previously we did not really enforce that the `payload` (a byte value
corresponding to a particular `Content.Type`) was correctly
persisted and as a workaround completely removed the handling of the
`payload` value. This is rather inefficient, because the
`StoreWorker` implementation has to deserialize the whole persisted
value to identify the content type.

Beside the above inefficiency having the right `payload` value is
also a prerequisite to make content-types completely pluggable.

Depends on PRs: #4941 #4942

Follow-up PRs: #4944 #4945

See issue #4940


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -250,6 +250,15 @@ protected CommitLogEntry commitAttempt(
     Function<Hash, CommitLogEntry> currentCommit =
         h -> h.equals(branchHead) ? currentBranchEntry : null;
 
+    for (KeyWithBytes put : commitParams.getPuts()) {
+      if (put.getPayload() <= (byte) 0) {
```
dimas-b, line 254: Do we actually expect this condition to be `true`? We validate payload bytes to be > 0 at the type level, do we not?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -632,12 +641,16 @@ protected Stream<Difference> buildDiff(
               if (f.equals(t)) {
                 return null;
               }
+              byte payload =
+                  fromVal != null
+                      ? fromVal.getPayload()
+                      : (toVal != null ? toVal.getPayload() : (byte) 0);
```
dimas-b, line 647: Should we check that payloads match if both are set?

servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java
```diff
@@ -246,7 +247,7 @@ public Content applyId(Content content, String id) {
   }
 
   @Override
-  public Content.Type getType(ByteString onRefContent) {
+  public Content.Type getType(byte payload, ByteString onRefContent) {
     ObjectTypes.Content parsed = parse(onRefContent);
```
dimas-b, line 251: Can we determine type based on `payload` now?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: We do - probably overly cautious here.

servers/store/src/main/java/org/projectnessie/server/store/TableCommitMetaStoreWorker.java
snazy: **Should** work - but probably not guaranteed for old values.
The follow-up PRs will actually handle the `payload` value and deal with the case when `payload==0`.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Actually a different problem.
I'll add a TODO there

snazy: #4949

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/adapter/serialize/ProtoSerialization.java
```diff
@@ -197,15 +198,17 @@ public static ContentIdAndBytes protoToContentIdAndBytes(AdapterTypes.ContentIdW
     return ContentIdAndBytes.of(ContentId.of(proto.getContentId().getId()), proto.getValue());
   }
 
-  public static AdapterTypes.ContentIdWithType toProto(ContentId x) {
-    return AdapterTypes.ContentIdWithType.newBuilder()
-        .setContentId(AdapterTypes.ContentId.newBuilder().setId(x.getId()))
-        .setTypeUnused(0)
+  public static AdapterTypes.ContentIdWithPayload toProto(ContentIdAndPayload x) {
```
dimas-b, line 201: This method appears to be called only from test code :thinking:

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/adapter/serialize/ProtoSerialization.java
```diff
@@ -197,15 +198,17 @@ public static ContentIdAndBytes protoToContentIdAndBytes(AdapterTypes.ContentIdW
...
+    return AdapterTypes.ContentIdWithPayload.newBuilder()
+        .setContentId(AdapterTypes.ContentId.newBuilder().setId(x.getContentId().getId()))
+        .setPayload(x.getPayload())
         .build();
   }
 
-  public static ContentId protoToContentId(AdapterTypes.ContentIdWithType proto) {
-    return ContentId.of(proto.getContentId().getId());
+  public static ContentIdAndPayload protoToContentIdAndPayload(
```
dimas-b, line 208: This method appears to be called only from test code :thinking:

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ContentIdAndPayload.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.persist.adapter;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ContentIdAndPayload {
```
dimas-b, line 21: Is this class required? I can see usage only in test code :thinking:

dimas-b (commented):
LGTM, just on concern about the use of `ContentIdWithPayload`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/ContentIdAndPayload.java
snazy: Weird... could swear, that it's being used.

dimas-b (approved):

PR #4944, opened 2022-08-17
Make content-type backend serialization pluggable

snazy:
This also removes the necessity to instantiate a `StoreWorker`,
implicitly cleaning up a bunch of related code.

Also finally removes the technical ability to store Iceberg content
with global state.

Depends on PRs: #4941 #4942 #4943

Follow-up PRs: #4945

See issue #4940


versioned/spi/src/main/java/org/projectnessie/versioned/store/ContentSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+import org.projectnessie.model.Content;
+import org.projectnessie.versioned.ContentAttachment;
+import org.projectnessie.versioned.ContentAttachmentKey;
+
+/**
+ * Content serializers provide persistence layer (de)serialization functionality for a specific
+ * content type.
+ */
+public interface ContentSerializer<C extends Content> {
+  String contentType();
```
dimas-b, line 32: Why not `ContentType` object?

versioned/spi/src/main/java/org/projectnessie/versioned/store/DefaultStoreWorker.java
```diff
@@ -0,0 +1,148 @@
...
+    return Lazy.INSTANCE;
+  }
+
+  private static final class Lazy {
+    private static final DefaultStoreWorker INSTANCE = new DefaultStoreWorker();
+  }
+
+  private static final class Registry {
+    private static final ContentSerializer<?>[] BY_PAYLOAD;
+    private static final Map<Class<? extends Content>, ContentSerializer<?>> BY_TYPE;
```
dimas-b, line 53: Why not `Map<ContentType, ContentSerializer<?>>`?

versioned/spi/src/main/java/org/projectnessie/versioned/store/ContentSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+
+  C applyId(C content, String id);
+
+  default boolean requiresGlobalState(C content) {
+    return false;
+  }
+
+  Content.Type getType(byte payload, ByteString onReferenceValue);
+
+  boolean requiresGlobalState(byte payload, ByteString onReferenceValue);
```
dimas-b, line 44: Could you move it one method up to be together with `requiresGlobalState(C content)`?

servers/store/src/main/java/org/projectnessie/server/store/IcebergTableSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+public final class IcebergTableSerializer extends BaseSerializer<IcebergTable> {
+
+  @Override
+  public String contentType() {
+    return Content.Type.ICEBERG_TABLE.name();
+  }
+
+  @Override
+  public Content.Type getType(byte payload, ByteString onReferenceValue) {
+    return Content.Type.ICEBERG_TABLE;
```
dimas-b, line 33: What is the purpose of the args now? If only for test, could we validate payload to value match some other way, without overloading the main production call path with extra args?

dimas-b: Ah, thx, I missed that... However, would it make sense to rename `UnknownSerializer` to `LegacySerializer` perhaps?.. just for the sake of clarity... I'd even call it `LegacyDeerializer`.

snazy: The type‘s named `UNKNOWN`, that’s why

dimas-b: But them we have type misalignment between `contentType()` and `getType(byte payload, ByteString onReferenceValue)`.

`contentType()` appears to be required only for serialization, which is not supported by `UnknownSerializer`... looks like of odd.

What do you think about making the `0` payload handling a special case in `DefaultStoreWorker`? Then, we may not need `UnknownSerializer` at all :thinking:

snazy: `DefaultStoreWorker` shouldn't know anything about the "main" content types.
We can likely remove the functionality in `UnknownSerializer` when there are no more commits with payload==0 - the state where we'll eventually be.

snazy: I.e. `UnknownSerializer` doesn't need to be perfect - it'll go away.

snazy: ... and the `getType()` function will then disappear as well

dimas-b: yes, but it still looks awkward. How about using a separate type for `0` payload, i.e. `LEGACY`, which would _not_ have a constant in `Content.Type`. The `UNKNOWN` type's serialize would throw exceptions on ser/deser attempts and the `LEGACY` serializer would be used only for reading old entries with `0` payload. WDYT?

dimas-b: Actually, we can probably resolve this after #4945

snazy: `UNKNOWN` itself is useful - i.e. when there's a payload for which you no longer have a serializer/content.

it's just that accidentally existing persisted data has payload==0, which should have never happened, but probably did.

dimas-b: Actually, we can probably resolve this after #4945

dimas-b (approved):

PR #4945, opened 2022-08-17
Remove payload from `Content.Type`

### LINKED ISSUE #4940: Clean up use of generics and "alien" `Content` types only for tests

snazy:
The payload value is an internal concern and must not be visible to
users of the API.

Depends on PRs: #4941 #4942 #4943 #4944

Fixes #4940

### DIFF
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/store/DefaultStoreWorker.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/store/DefaultStoreWorker.java
@@ -42,17 +42,22 @@ public class DefaultStoreWorker implements StoreWorker {
   }
 
   public static byte payloadForContent(Content c) {
-    return Registry.BY_TYPE.get(c.getType()).payload();
+    return serializer(c).payload();
   }
 
   public static byte payloadForContent(Content.Type contentType) {
-    return Registry.BY_TYPE.get(contentType).payload();
+    return serializer(contentType).payload();
   }
 
   public static Content.Type contentTypeForPayload(byte payload) {
-    return payload >= 0 && payload < Registry.BY_PAYLOAD.length
-        ? Registry.BY_PAYLOAD[payload].contentType()
-        : null;
+    Content.Type contentType =
+        payload >= 0 && payload < Registry.BY_PAYLOAD.length
+            ? Registry.BY_PAYLOAD[payload].contentType()
+            : null;
+    if (contentType == null) {
+      throw new IllegalArgumentException("Unknown payload " + payload);
+    }
+    return contentType;
   }
 
   private static final class Lazy {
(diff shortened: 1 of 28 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/store/DefaultStoreWorker.java
```diff
@@ -44,47 +42,63 @@ public static StoreWorker instance() {
     return Lazy.INSTANCE;
   }
 
+  public static byte payloadForContent(Content c) {
```
dimas-b, line 45: May be move the serializer registry stuff into a new top-level class to avoid confusion with "worker" names?

snazy: Hm, it's tightly coupled and I'm not so much a fan of opening that `Registry` class

versioned/spi/src/main/java/org/projectnessie/versioned/store/DefaultStoreWorker.java
```diff
@@ -43,6 +41,20 @@ public static StoreWorker instance() {
...
+  }
+
+  public static byte payloadForContent(Content.Type contentType) {
+    return Registry.BY_TYPE.get(contentType).payload();
+  }
+
+  public static Content.Type contentTypeForPayload(byte payload) {
+    return payload >= 0 && payload < Registry.BY_PAYLOAD.length
+        ? Registry.BY_PAYLOAD[payload].contentType()
+        : null;
```
dimas-b, line 55: Should we throw in this case? Where might `null` be used in practice?

dimas-b (approved):

PR #4946, opened 2022-08-18
Fix OpenAPI spec re-generation

snazy:
Changes to the API interfaces in `:nessie-model` do not correctly
lead to a re-generation of the `openapi.{json|yaml}` files.

This change makes the whole `src/main` tree an input to the Gradle
task generating the OpenAPI spec files.

### DIFF
--- a/model/build.gradle.kts
+++ b/model/build.gradle.kts
@@ -63,7 +63,9 @@ smallryeOpenApi {
   infoLicenseUrl.set("http://www.apache.org/licenses/LICENSE-2.0.html")
   schemaFilename.set("META-INF/openapi/openapi")
   operationIdStrategy.set("METHOD")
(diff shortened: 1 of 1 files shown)


model/build.gradle.kts
```diff
@@ -63,7 +63,9 @@ smallryeOpenApi {
   infoLicenseUrl.set("http://www.apache.org/licenses/LICENSE-2.0.html")
   schemaFilename.set("META-INF/openapi/openapi")
```
XN137, line 64: just wondering:
why is this not `META-INF/openapi.yaml` or `src/main/resources/META-INF/openapi.yaml` or the like?

XN137 (approved):

model/build.gradle.kts
snazy: There's already a generated openapi.yaml w/o the samples - that would conflict

PR #4948, opened 2022-08-18
Fix Quarkus logging during build

snazy:
Quarkus log messages are currently not properly formatted and look like this:
```
Detected a split package usage which is considered a bad practice and should be avoided. Following packages were detected in multiple archives: %s
```
but should really look like this:
```
WARN: Detected a split package usage which is considered a bad practice and should be avoided. Following packages were detected in multiple archives:
- "org.projectnessie.versioned.persist.mongodb" found in [org.projectnessie:nessie-versioned-persist-mongodb::jar, org.projectnessie:nessie-versioned-persist-mongodb:tests:jar]
```

Turns out that configuring the JBoss LogManager class was wrong, removing the configuration fixes the issue.

### DIFF
--- a/servers/lambda/build.gradle.kts
+++ b/servers/lambda/build.gradle.kts
@@ -105,8 +105,7 @@ tasks.withType<Test>().configureEach {
   systemProperty("quarkus.container-image.build", useDocker)
   systemProperty("quarkus.smallrye.jwt.enabled", "true")
   // TODO requires adjusting the tests - systemProperty("quarkus.http.test-port", "0") -  set this
-  // property in application.properties
-  systemProperty("java.util.logging.manager", "org.jboss.logmanager.LogManager")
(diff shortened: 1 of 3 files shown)


servers/lambda/build.gradle.kts
```diff
@@ -105,8 +105,7 @@ tasks.withType<Test>().configureEach {
   systemProperty("quarkus.container-image.build", useDocker)
   systemProperty("quarkus.smallrye.jwt.enabled", "true")
   // TODO requires adjusting the tests - systemProperty("quarkus.http.test-port", "0") -  set this
-  // property in application.properties
-  systemProperty("java.util.logging.manager", "org.jboss.logmanager.LogManager")
```
snazy, line 109: This is the actual "fix"

dimas-b (approved):

PR #4958, opened 2022-08-18
WIP: testing autolabeler

dimas-b:
Testing how well the autolebler from #4931 works


dimas-b:
Labels added under the `GITHUB_TOKEN` apparently do _not_ trigger other workflows... https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow

PR #4962, opened 2022-08-19
Adopt compatibility tests for Nessie 0.40.0..0.41.0

snazy:
Main purpose is^H^Hwas to just add Nessie versions 0.40.3 + 0.41.0
to the list of tested versions. That however unveiled a bunch of
issues, which are all resolved with this PR.

* `o.p.tools.compatibility.internal.TranslatingVersionNessieApi` did
  not handle the objects returned via `Stream`s, which means that
  the results returned via the new functionality
  `o.p.client.api.PagingBuilder#stream` was not properly handled.
* The published pom's had a couple of issues. Maven poms with bad
  dependency declarations do **not** error out by default (that's
  a Maven default!!) - so errors were not detected.
  * A new test that runs against the published pom's has been added
    to the build-tools-integration-tests to detect those issues in
    the future.
  * The coordinates for `jakarta.annotation:jakarta.annotation-api`
    were managed with the wrong group ID in the top-level
    build.gradle.kts file, so that project pom's declaring this
    dependency had no version to resolve.
  * `nessie-client`'s pom did not mention the `jackson-bom` in its
    `dependencyManagement` section.
  * The `junit-bom` version wasn't available from the pom of
    `nessie-jaxrs-testextension`.
  * The `junit-bom` version wasn't available from the pom of
    `nessie-jaxrs-tests`.
  * Nessie versions 0.40.x and 0.41.0 return two conflicting
    versions of `io.opentracing:opentracing-*` artifacts - 0.30.0
    _and_ 0.33.0 - which are incompatible to each other. Those
    are managed to 0.33.0 for these Nessie versions.
    (Current `main` branch is not affected by this one.)
* To mitigate the above issues in the backwards compatibility tests,
  a couple more changes have been made in
  `nessie-compatibility-common` to handle the above pom dependency
  issues in the new class `OldNessie` - plus a couple related
  changes.


build-tools-integration-tests/src/test/java/org/projectnessie/buildtools/TestPublishedPoms.java
```diff
@@ -0,0 +1,155 @@
...
+    repositories =
+      singletonList(
+        new RemoteRepository.Builder(
+          "maven-central", "default", "https://repo1.maven.org/maven2/")
+          .build());
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {
+    "nessie-client",
```
snazy, line 68: Cannot read the properties files in ../gradle/, because those contain _all_ Nessie projects, not just only the ones publishing jars.

build.gradle.kts
```diff
@@ -136,7 +136,7 @@ dependencies {
     api("com.google.protobuf:protobuf-java:$versionProtobuf")
     api("info.picocli:picocli:$versionPicocli")
     api("io.quarkus:quarkus-smallrye-opentracing:$versionQuarkus")
-    api("jakarta.validation:jakarta.annotation-api:$versionJakartaAnnotationApi")
```
snazy, line 139: No idea why this worked before

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/TranslatingVersionNessieApi.java
```diff
@@ -210,6 +215,9 @@ Object translateObject(Object o, ClassLoader classLoader, ClassLoader reverseCla
               .map(e -> translateObject(e, classLoader, reverseClassLoader))
               .collect(Collectors.toSet());
     }
+    if (o instanceof Stream) {
+      return ((Stream<?>) o).map(e -> translateObject(e, classLoader, reverseClassLoader));
```
snazy, line 219: This is the actual fix to handle `o.p.client.api.PagingBuilder#stream`

compatibility/compatibility-tests/src/test/java/org/projectnessie/tools/compatibility/tests/ITRollingUpgradeScenario.java
```diff
@@ -48,7 +48,7 @@
 
 @ExtendWith(RollingUpgradesExtension.class)
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
-@VersionCondition(minVersion = "0.31.0")
+@VersionCondition(minVersion = "0.40.0")
```
snazy, line 51: Just in case we do a 0.31....

servers/jax-rs-testextension/build.gradle.kts
```diff
@@ -29,6 +29,7 @@ description = "JUnit Jupyter Extension to run tests against an \"embedded\" Ness
 dependencies {
   api(platform(rootProject))
   api(project(":nessie-jaxrs"))
+  api(platform(project(":nessie-deps-testing")))
```
snazy, line 32: Needed to get the version of `junit-bom`

dimas-b (approved):

PR #4971, opened 2022-08-22
Handle merge response in nessie CLI

### LINKED ISSUE #4800: nessie --json merge does not return any data
**Description**:

As of 0.30.0 the `nessie --json merge` does not return any information, such as the resulting hash of the commit that has just been added to the target branch. Such information is vital if you want to also tag that commit (and not the one in the source branch, which is different).

Looking at the server-side code looks like the info is returned in the terms of a `MergeResponse` which contains such info. Why isn't it returned by the CLI too?

- Changes in nessie cli

kamalbhavyam:


python/pynessie/client/_endpoints.py
```diff
@@ -310,12 +310,14 @@ def merge(
...
     :param expected_hash: expected hash of HEAD of branch
     :param ssl_verify: ignore ssl errors if False
+    :return: json dict of a merge response
     """
     url = "/trees/branch/{}/merge".format(branch)
     params = {}
     if expected_hash:
         params["expectedHash"] = expected_hash
-    _post(base_url + url, auth, json=merge_json, ssl_verify=ssl_verify, params=params)
+    ret_dict = cast(dict, _post(base_url + url, auth, json=merge_json, ssl_verify=ssl_verify, params=params))
```
XN137, line 319: nit: lets not do too much on a single line
instead have:
```
response = _post(...)
return cast(dict, response)
```

python/pynessie/client/nessie_client.py
```diff
@@ -259,7 +261,7 @@ def merge(self, from_ref: str, onto_branch: str, from_hash: Optional[str] = None
             from_hash = from_hash_ref
 
         merge_json = MergeSchema().dump(Merge(from_ref, str(from_hash)))
-        merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify)
+        return MergeResponseSchema().load(merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify))
```
XN137, line 264: nit: same as above

@snazy wdyt is our preferred strategy in terms of compatibility of nessie CLI against different nessie versions?
i.e. do we need to gracefully handle getting an empty response here?

python/pynessie/commands/merge.py
```diff
@@ -70,5 +71,12 @@ def merge(ctx: ContextObject, ref: str, force: bool, expected_hash: str, hash_on
             """Either condition or force must be set. Condition should be set to a valid hash for concurrency
             control or force to ignore current state of Nessie Store."""
         )
-    ctx.nessie.merge(from_ref, ref, hash_on_ref, expected_hash)
-    click.echo()
+    merge_response = ctx.nessie.merge(from_ref, ref, hash_on_ref, expected_hash)
+    if ctx.json:
```
XN137, line 75: we should add tests to both code paths i.e. verify the output of the merge command with and without the json flag

also since the VCR recording still returns an empty string:
https://github.com/projectnessie/nessie/blob/ad496c3334cee883f4eed949dd666739c06a6368/python/tests/cassettes/test_nessie_cli/test_merge.yaml#L295-L319

most likely the existing test will still print nothing when calling merge, so we need to adjust this

XN137 (changes requested):
looks like a good start, i assume you tested the CLI against a locally running server?
we'll need to adjust the tests accordingly

also wdyt about renaming the PR like so:
```
Handle merge response in nessie CLI

fix #4800
```
since we are not only doing something in the `--json` case

python/pynessie/client/nessie_client.py
snazy: Nessie CLI should also work against older versions, if possible.

python/pynessie/client/_endpoints.py
kamalbhavyam: updated

python/pynessie/client/nessie_client.py
kamalbhavyam: _endpoints passes an empty dict as response for old servers that don't return any response on merge request. This means normal output will remain same and --json output will be a json with each field null.
`{"targetCommits": null, "sourceCommits": null, "targetBranch": null, "details": null, "wasApplied": false, "expectedHash": null, "wasSuccessful": false, "resultantTargetHash": null, "commonAncestor": null, "effectiveTargetHash": null}`

kamalbhavyam: Updated message for legacy servers

python/pynessie/commands/merge.py
kamalbhavyam: Added test cases for both json and normal output and updated cassettes for these

python/pynessie/commands/merge.py
```diff
@@ -70,5 +73,23 @@ def merge(ctx: ContextObject, ref: str, force: bool, expected_hash: str, hash_on
...
+            click.echo(json.dumps(hint))
+        else:
+            click.echo("Merge succeeded but legacy server did not respond with additional details.")
+    else:
+        if ctx.json:
+            click.echo(MergeResponseSchema().dumps(merge_response))
+        else:
+            commits = merge_response.source_commits
+            click.echo(
+                "The following {N} commits were merged onto {target_branch}:".format(
```
XN137, line 89: we should mention the commit hash for the merge? i.e. the hash on the target branch after this operation was done
(think about many operations happening concurrently and you need to decipher the order of things from CLI log output afterwards)

also since you mentioned that commits could be squashed, maybe this info is also available in the response somehow and we should surface it?

python/pynessie/model.py
```diff
@@ -458,6 +458,45 @@ class Merge:
...
+    effective_target_hash: str = attr.ib(metadata=desert.metadata(fields.Str(data_key="effectiveTargetHash")))
+    source_commits: List[LogEntry] = attr.ib(
+        metadata=desert.metadata(fields.List(fields.Nested(LogEntrySchema()), data_key="sourceCommits"))
+    )
+    details: List[MergeResponseDetails] = attr.ib(
+        metadata=desert.metadata(fields.List(fields.Nested(MergeResponseDetailsSchema()), data_key="details"))
+    )
+    resultant_target_hash: str = attr.ib(metadata=desert.metadata(fields.Str(allow_none=True, data_key="resultantTargetHash")))
+    common_ancestor: str = attr.ib(metadata=desert.metadata(fields.Str(allow_none=True, data_key="commonAncestor")))
+    expected_hash: str = attr.ib(metadata=desert.metadata(fields.Str(allow_none=True, data_key="expectedHash")))
```
XN137, line 489: are the `allow_none=True` parts still required now that we return `Optional[MergeResponse]` ?

python/pynessie/client/nessie_client.py
```diff
@@ -259,7 +263,10 @@ def merge(self, from_ref: str, onto_branch: str, from_hash: Optional[str] = None
             from_hash = from_hash_ref
 
         merge_json = MergeSchema().dump(Merge(from_ref, str(from_hash)))
-        merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify)
+        merge_response = merge(self._base_url, self._auth, onto_branch, merge_json, old_hash, self._ssl_verify)
+        if merge_response:
+            return MergeResponseSchema().load(merge_response)
+        return None
```
XN137, line 269: we should describe with a comment and/or in the method docs when `None` can be returned i.e. we only expect this to happen against legacy servers... starting from a specific server version it would be an error to get an empty response

python/tests/test_nessie_cli.py
```diff
@@ -315,7 +316,36 @@ def test_merge() -> None:
     # Passing detached commit-id plus a _different_ hash-on-ref --> error
     execute_cli_command(["merge", f"dev@{dev_hash}", "-c", main_hash, "-o", main_hash], ret_val=1)
 
-    execute_cli_command(["merge", "dev", "-c", main_hash])
+    assert_that(execute_cli_command(["merge", "dev", "-c", main_hash])).starts_with("The following 1 commits were merged onto main:")
```
XN137, line 319: nit:
```
stdout = execute_cli_command(..)
assert_that(stdout).starts_with(..)
```

python/tests/test_nessie_cli.py
```diff
@@ -315,7 +316,36 @@ def test_merge() -> None:
...
+    logs = simplejson.loads(execute_cli_command(["--json", "log"]))
+    # we don't check for equality of hashes here because a merge
+    # produces a different commit hash on the target branch
```
XN137, line 322: could we maybe verify things like the commit msg still being present or the like?

python/tests/test_nessie_cli.py
```diff
@@ -315,7 +316,36 @@ def test_merge() -> None:
...
+    assert len(logs) == 1
+    logs = simplejson.loads(execute_cli_command(["--json", "log", "dev"]))
+    assert len(logs) == 1
```
XN137, line 325: wondering: if we ran the same merge command afterwards, what kind of error would we be getting now?
maybe this could be part of the test WDYT ?

python/pynessie/commands/merge.py
```diff
@@ -70,5 +73,23 @@ def merge(ctx: ContextObject, ref: str, force: bool, expected_hash: str, hash_on
             """Either condition or force must be set. Condition should be set to a valid hash for concurrency
             control or force to ignore current state of Nessie Store."""
         )
-    ctx.nessie.merge(from_ref, ref, hash_on_ref, expected_hash)
-    click.echo()
+    merge_response = ctx.nessie.merge(from_ref, ref, hash_on_ref, expected_hash)
+    if merge_response is None:
```
XN137, line 77: would it be worthwhile to keep a VCR recording for the old server response and add a test around it?

XN137 (commented):
looks good, almost there i think

python/pynessie/model.py
kamalbhavyam: allow_none is only set to parameters that are nullable in MergeResponse

python/pynessie/client/nessie_client.py
kamalbhavyam: Added to method docs.

python/tests/test_nessie_cli.py
kamalbhavyam: Added the same

python/tests/test_nessie_cli.py
```diff
@@ -324,6 +337,62 @@ def test_merge() -> None:
...
+@pytest.mark.vcr
+def test_merge_legacy_json() -> None:
+    """Test merge operation."""
+    execute_cli_command(["branch", "dev"])
+    make_commit("merge.foo.bar", _new_table("test_merge"), "dev")
+    main_hash = ref_hash("main")
+    dev_hash = ref_hash("dev")
+
+    # Passing detached commit-id plus a _different_ hash-on-ref --> error
+    execute_cli_command(["merge", f"dev@{dev_hash}", "-c", main_hash, "-o", main_hash], ret_val=1)
```
XN137, line 390: is this part important for the legacy usecase?

maybe both `test_merge_legacy_json` and `test_merge_legacy` could just use a single vcr recording if we remove it, wdyt?

python/tests/test_nessie_cli.py
```diff
@@ -314,8 +315,20 @@ def test_merge() -> None:
...
+    assert_that(merge_output).is_equal_to(
+        "The following 1 commits were merged onto main:"
+        '\n{} "test message" \nResultant hash on main after merge: {}\n'.format(refs["dev"], refs["main"])
+    )
+
+    # if we try to merge again from dev to main we get an error
+    # this is because there is now a conflict as the same commit exists on both branches
+    merge_error = execute_cli_command(["merge", "dev", "-c", refs["main"]], ret_val=1)
+    assert "Client Error Conflict: The following keys have been changed in conflict:" in merge_error
     logs = simplejson.loads(execute_cli_command(["--json", "log"]))
```
XN137, line 332: nit: verifying the logs is still part of the happy path test... so should come before triggering the conflict error imo

python/tests/test_nessie_cli.py
XN137: nit: if the log cmd constains the commit msg maybe verify them as well here

XN137 (approved):
thanks for the updates, lgtm overall with a few suggestions to improve/simplify things

python/tests/test_nessie_cli.py
XN137: but possible we want to keep it to verify that not only do we get an error also nothing was written server-side, so keeping as is is fine... maybe add newline inbetween

python/tests/test_nessie_cli.py
```diff
@@ -335,8 +398,49 @@ def test_merge_detached() -> None:
...
+    assert_that(merge_output).is_equal_to(
+        "The following 1 commits were merged onto main:"
+        '\n{} "test message" \nResultant hash on main after merge: {}\n'.format(refs["dev"], refs["main"])
+    )
+
+    logs = simplejson.loads(execute_cli_command(["--json", "log"]))
+    assert_that(logs[0]["message"]).is_equal_to("test message")
+    # we don't check for equality of hashes here because a merge
+    # produces a different commit hash on the target branch
+    assert len(logs) == 1
```
XN137, line 415: asserting the length needs to come before accessing `logs[0]`

python/tests/test_nessie_cli.py
```diff
@@ -314,16 +315,78 @@ def test_merge() -> None:
...
     execute_cli_command(["merge", f"dev@{dev_hash}", "-c", main_hash, "-o", main_hash], ret_val=1)
+    merge_output = execute_cli_command(["merge", "dev", "-c", main_hash])
+
+    branches = ReferenceSchema().loads(execute_cli_command(["--json", "branch"]), many=True)
+    refs = {i.name: i.hash_ for i in branches}
+
+    expected_output_line1 = "The following 1 commits were merged onto main:\n"
+    expected_output_line2 = '{} "test message" \n'.format(refs["dev"])
+    expected_output_line3 = "Resultant hash on main after merge: {}\n".format(refs["main"])
+    assert_that(merge_output).is_equal_to(expected_output_line1 + expected_output_line2 + expected_output_line3)
```
XN137, line 326: sorry i should have been clearer, i suggested to use `merge_output.splitlines()` so that we dont need to use `\n` in these expected messages but can use a list of lines, but you can also leave it as is

XN137 (approved):
looks good now :+1: 

feel free to ignore the code coverage check (it is not super accurate)... as long as the python tests are passing this should be fine to merge

PR #4972, opened 2022-08-22
Add pylint plugin to check license headers

### LINKED ISSUE #2586: [CLI] Add a pylint plugin to check if the license header exists in the Python files
Currently, we don't have a mechanism to check if the Python files contain the required license header, therefore it will be good if we have such a check in place. This can be written as form of Pylint plugin that checks every file whenever Pylint checks are triggered.

kamalbhavyam:
Added [pylintfileheader](https://github.com/HaaLeo/pylint-file-header) plugin to enforce license header checks on files.

The plugin is configured using file-header-path variable which is set to escaped_license.txt.
This file is used as regex by the plugin, so special characters must be escaped.

PR also fixes current incorrect file headers and removes unnecessary shebangs and comments for old python versions adhering to PEP-263.

Fixes #2586


python/tests/test_nessie_cli_error.py
```diff
@@ -1,18 +1,17 @@
...
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Authentication tests for Nessi CLI."""
```
XN137, line 14: nit: maybe change to "Error handling tests for Nessie CLI." while we are at it, since the existing comment seems copy-paste from the other file.

python/pynessie/utils/expression_util.py
```diff
@@ -1,20 +1,19 @@
-# -*- coding: utf-8 -*-
```
XN137, line 1: this is still present in many more files while PEP-263 isnt needed for python3 at all, wdyt about removing them as well?

python/.pylintrc
```diff
@@ -45,7 +45,9 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python module names) to load,
 # usually to register additional checkers.
-load-plugins=
+load-plugins=pylintfileheader
+
+file-header=(#.*\n)*?# Copyright \(C\) 2020 Dremio\n#\n# Licensed under the Apache License, Version 2\.0 \(the \"License\"\);\n# you may not use this file except in compliance with the License\.\n# You may obtain a copy of the License at\n#\n# http://www\.apache\.org/licenses/LICENSE-2\.0\n#\n# Unless required by applicable law or agreed to in writing, software\n# distributed under the License is distributed on an \"AS IS\" BASIS,\n# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied\.\n# See the License for the specific language governing permissions and\n# limitations under the License\.
```
XN137, line 50: is it intentional that this regex seems to allow multiple lines in front of the Copyright?
do we need to support anything else besides a single shebang line?

python/.pylintrc
```diff
@@ -45,7 +45,9 @@ limit-inference-results=100
 
 # List of plugins (as comma separated values of python module names) to load,
 # usually to register additional checkers.
-load-plugins=
+load-plugins=pylintfileheader
```
XN137, line 48: nice find!

XN137 (commented):
looks good overall! some minor questions/suggestions

also lets use this PR title to be more python specific:
```
Check license header in python files

fixes #4972
```
and have the fixes link in the commit body on merge

python/tests/test_nessie_cli_error.py
kamalbhavyam: Changed it to the same.

python/pynessie/utils/expression_util.py
kamalbhavyam: Removed this comment from other files.

python/.pylintrc
kamalbhavyam: There was a flake8 noqa comment in one of the files- moved it. Changed the regex to only allow a shebang.

python/tests/test_nessie_client.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 #
```
XN137, line 2: nit: still 2 lines before copy right

python/tests/test_nessie_cli.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 #
```
XN137, line 2: nit: still 2 lines before copy right... shouldnt the check be catching that now?

python/setup.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 #
```
XN137, line 2: nit: still 2 lines before copy right

python/docs/conf.py
```diff
@@ -1,6 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
-# flake8: noqa
 #
 # pynessie documentation build configuration file, created by
```
XN137, line 3: this file seems to have no license header at all?
do we know why the check is not enforced here?

python/tests/test_nessie_error.py
```diff
@@ -1,18 +1,17 @@
...
-#  distributed under the License is distributed on an "AS IS" BASIS,
-#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-#  See the License for the specific language governing permissions and
-#  limitations under the License.
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Authentication tests for Nessi CLI."""
```
XN137, line 14: fix typo: `Nessie` while we are at it

XN137 (commented):
thanks for the quick update, added few more comments

also please update PR title (see previous review) and description (we no longer need to support PEP-XYZ)

python/docs/conf.py
kamalbhavyam: pylint checks are being run on pynessie, tools, and tests folders. This file is in docs and not checked.

python/tests/test_nessie_cli.py
kamalbhavyam: The test ensures there is a commented empty line between shebang and license. Should I change this?

python/docs/conf.py
XN137: hmm thats odd, because `# flake8: noqa` means that flake8 supposedly runs against this file... so if it does probably also pylint should

python/tests/test_nessie_cli.py
XN137: in all the other files you have removed this empty line... only in 3 lines you have kept it.

which made me wonder if the regexp is working as intended...

generally both would be fine with me but if we remove the `coding` line we should do it uniformally across all the files imo

python/docs/conf.py
kamalbhavyam: The folders are mentioned explicitly in the pylint command, should I change it to run on docs as well?

python/tests/test_nessie_cli.py
kamalbhavyam: These 3 files have the shebang while the others didn't. The regex expects that either no comment exists before license header or a shebang followed by a commented empty line.

XN137: ah i see, sorry i missed that earlier.

but taking a step back:
none of these files need to be executable directly from a shell (there is nothing special about them compared to the other py files)... so we should just remove the shebang here, wdyt?

python/docs/conf.py
XN137: hmm no its fine then, but then, since we do not run flake8 on the docs folder, we can just remove the `# flake8: noqa` (probably a leftover from a time when docs folder was still checked)

python/tests/test_nessie_cli.py
kamalbhavyam: removed shebangs from files and regex

python/.pylintrc
```diff
@@ -45,7 +45,9 @@ limit-inference-results=100
...
+
+file-header=# Copyright \(C\) 2020 Dremio\n#\n# Licensed under the Apache License, Version 2\.0 \(the \"License\"\);\n# you may not use this file except in compliance with the License\.\n# You may obtain a copy of the License at\n#\n# http://www\.apache\.org/licenses/LICENSE-2\.0\n#\n# Unless required by applicable law or agreed to in writing, software\n# distributed under the License is distributed on an \"AS IS\" BASIS,\n# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied\.\n# See the License for the specific language governing permissions and\n# limitations under the License\.
```
XN137, line 50: nit: we could now go back to using `file-header-path` here most likely, wdyt?

XN137 (dismissed):
thanks for the update, looks good to me +1 with minor suggestion

please make sure to craft a condensed commit msg when merging

python/.pylintrc
dimas-b: +1 to using file for the actual header text

python/tests/test_nessie_client.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env python
```
dimas-b, line 1: Did we not want to keep the shebang lines?

Why would we remove it from this file?

python/.pylintrc
kamalbhavyam: switched to file

XN137 (dismissed):
+1 when CI passes (with another minor comment, sorry :sweat_smile:)

XN137 (approved):
+1 after passing CI, friendly reminder about commit msg on merge.

Issue #4694, opened 2022-07-13
GC: Require functionality to purge completely unreferenced files


snazy:
Nessie GC, even with #3885, has no functionality to identify and prune completely unreferenced files - i.e. files that have been written, but were never committed to Nessie.

ajantha-bhat:
Duplicate of https://github.com/projectnessie/nessie/issues/855

snazy:
Implemented  via #4991

Issue #4626, opened 2022-07-01
`gc-base` module must not depend on Iceberg


snazy:
`gc-base` is intended to contain the common Nessie-GC code for any table-format. Iceberg gets its implementation in a separate module.

References to Iceberg code must therefore not be present in the `gc-base` module.

ajantha-bhat:
Currently, `gc-base` module needs to write its output to the spark table. 
It can be any table format but it has to use existing nessie catalog. (Don't want to force user to configure another non-Nessie catalog while working on Nessie gc)

And with the Nessie catalog, only Iceberg tables are better handled. So, chose Iceberg format.

I can move this `IdentifiedResultsRepo` code to `gc-iceberg` module. But `gc-base` module becomes incomplete or will have a dependency on `gc-iceberg` module. 

I will check about having a default implementation as spark table with spark default catalog and open a `gcParams` to provide className of the repo impl.

ajantha-bhat:
I tired removing iceberg dependency on gc-base module here
https://github.com/projectnessie/nessie/commit/f08ae70309e30f53cbcc645483c69d57b3015744

Instead of the Iceberg table, I am trying to store it as a normal parquet table. But I am facing issues as to which catalog to use?
  I used `session.sessionState().catalogManager().currentCatalog();`
  It gives me many errors while creating a table. ()
  I have only Nessie catalog configured in spark conf and really don't want to force user to have another catalog while working on Nessie GC.
  
  I will work on this once other things are handled for gc.

snazy:
Implemented via #4991

Issue #2255, opened 2021-10-05
Implement new Nessie GC


snazy:
(pulled this issue out of #1977)

Issue #706, opened 2021-01-20
Performance characteristics of GC Algo


rymurr:
There are a few things open from #684 on the topic of performance:

1. parallelisation of the store access. This involves, at a guess, moving the recursive L1 search to a Spark `DataSource`
2. potentially moving `java.util.function.Predicate` usages to Spark filters
3. Larger scale GC test with 10000s of commits/branches/objects

snazy:
Done via #4991

PR #4991, opened 2022-08-24
WIP! Nessie GC - mark & sweep

### LINKED ISSUE #4695: GC: Abstract distributed computing
Current GC-base implementation and upcoming Iceberg-GC implementations require Spark to distribute some processing tasks.

Those processing tasks are:
* Walk commit logs + identify live + expired contents
* Read metadata + manifest files

These tasks are I/O and potentially memory intensive operations, but likely not CPU intensive operations. Especially mostly I/O intensive operations would work more efficiently if those run in a single or few processes - think: multiple threads.

Task:
* Implement a "processing engine independent" API to distribute these tasks
* Provide an implementation using Spark

snazy:
**STILL WORK IN PROGRESS**

See also the readme in `./gc/README.md`.

GC is implemented as a two-phase approach:
1. Identify all refetences to live content versions by walking all
   named references. Content versions for Iceberg are tuples of
   commit-ID + snapshot ID + content ID + metadata reference).
   These `ContentReference`s are and stored in an instance of
   `LiveContentSetsStore`. This is the "mark phase". The set of
   content-references is called "live contents set".
2. For each content-ID, resolve all `ContentReferences` to file
   objects and populate a bloom-filter with those. Then traverse
   the base storage locations for each content-ID and delete all
   files that are not contained in the bloom filter, but do not
   delete files that are "too new" (created since the "mark phase"
   started).

A command line tool `nessie-gc-shell` is provided as well. It
supports running mark-and-sweep, mark only, sweep only plus a few
supplemental commands.

All pieces are pluggable, there are Java interfaces (or abstract
classes) providing the API for all pieces. This makes the code much
easier to test and helps using alternative implementations.

The base functionality is contained in the `nessie-gc-base` project.
This has default implementations to identify live content
references, an in-memory live-content-sets-storage and a default
implementation to run expire per content.

Support for content types, like Iceberg tables, implements
the functionality to expand content-references to actual
file-objects, list file-objects from base locations and delete
files. The "sweep" phase for Iceberg therefore also acts as the
"Nessie aware delete orphan files".

A live-content-sets-storage implementation using JDBC is present as
well.  Small Nessie repositories using a "one-off" mark-and-sweep
run can get away with the in-memory contents-storage. But big Nessie
repositories would require too much heap.

Another reason to eventually use a persistent
live-content-sets-storage is to allow the deletion of the files
occupied by completely unreferenced tables - tables that are not
referenced by any live Nessie commit.  This functionality is not
implemented, but would work by collecting the base-locations for
all content-IDs that are no longer present in the most recent
live-contents-set.

This PR still has a few non-`Test*` classes that were only added to
prove that the concept works quick and does not require much heap
and does not use a lot of CPU. Those "tests" in `nessie-gc-base` and
`nessie-gc-iceberg-inttest` will be removed, because those add no
real value and were only needed to validate the concept. The "final"
proof showed that a mark-and-sweep (deleting ~1.7M files, collected
from "many" branches, commits, content objects) alone can complete
in round about one minute with a 2GB Java heap.

Fixes #4695 #4694 #4626 #2255 #706


snazy:
Closing this PR (was used as an integration-PR of multiple, smaller PRs). Will keep the branch with the "workload tests" for a while.

PR #5041, opened 2022-08-29
quarkus-cli: Disable dev-services

snazy:


servers/quarkus-cli/src/main/resources/application.properties
```diff
@@ -31,6 +31,8 @@ quarkus.datasource.db-kind=postgresql
 #quarkus.datasource.username=<your username>
 #quarkus.datasource.password=<your password>
 #quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/my_database
+# Nessie tests manage PostgreSQL test containers explicitly. Datasource dev services are not necessary.
+quarkus.datasource.devservices.enabled=false
```
XN137, line 35: does it mean we can remove `quarkus.mongodb.devservices.enabled=false` below ?
same for `servers/quarkus-server/src/main/resources/application.properties` ?

for my quarkus understanding:
this configures the application but also its dev-mode as run by tests?
and this is why the `application.properties` file is kept below `src/main/resoursces` and not as part of the test resources?

XN137 (approved):
+1 as seems consistent with `servers/quarkus-server/src/main/resources/application.properties` though i probably dont understand all implications :sweat_smile:

servers/quarkus-cli/src/main/resources/application.properties
dimas-b: `quarkus.mongodb.devservices.enabled=false` can be removed now, AFAIK.

dimas-b: In my limited understanding of quarkus, there's only one `application.properties` file, which is used both for main runtime and for tests.

However, test can usually configure their stuff via java annotations.

dimas-b (approved):

PR #5043, opened 2022-08-29
Update `DatabaseAdapter` with functionality for export/import

snazy:
* Fetch multiple commit-log-entries
* Write multiple commit-log-entries
* Update (multiple) commit-log-entries
* Add aggregated key-list to a commit-log-entry

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
@@ -118,7 +118,7 @@ public interface DatabaseAdapter {
   Stream<CommitLogEntry> commitLog(Hash offset) throws ReferenceNotFoundException;
 
   /**
-   * Loads a single commit log entry.
+   * Loads commit log entries.
    *
    * @return the loaded {@link CommitLogEntry}s, non-existing entries will not be returned.
    */
@@ -370,6 +370,11 @@ public interface DatabaseAdapter {
    * Write multiple new commit-entries, the given commit entries are to be persisted as is. All
    * values of the given {@link CommitLogEntry} can be considered valid and consistent.
    *
+   * <p>Callers must call {@link #updateMultipleCommits(List)} for already existing {@link
+   * CommitLogEntry}s and {@link #writeMultipleCommits(List)} for new {@link CommitLogEntry}s.
+   * Implementations can rely on this assumption (think: SQL {@code INSERT} + {@code UPDATE}
+   * compared to a "simple put" for NoSQL databases).
+   *
    * <p>Implementations however can enforce strict consistency checks/guarantees, like a best-effort
    * approach to prevent hash-collisions but without any other consistency checks/guarantees.
    */
@@ -377,15 +382,28 @@ public interface DatabaseAdapter {
       throws ReferenceConflictException;
 
   /**
-   * Updates multiple new commit-entries, the given commit entries are to be persisted as is. All
-   * values of the given {@link CommitLogEntry} can be considered valid and consistent.
+   * Updates multiple commit-entries, the given commit entries are to be persisted as is. All values
+   * of the given {@link CommitLogEntry} can be considered valid and consistent.
+   *
+   * <p>Callers must call {@link #updateMultipleCommits(List)} for already existing {@link
+   * CommitLogEntry}s and {@link #writeMultipleCommits(List)} for new {@link CommitLogEntry}s.
+   * Implementations can rely on this assumption (think: SQL {@code INSERT} + {@code UPDATE}
+   * compared to a "simple put" for NoSQL databases).
    *
    * <p>Implementations however <em>can</em> enforce strict consistency checks/guarantees.
    */
   void updateMultipleCommits(List<CommitLogEntry> commitLogEntries)
       throws ReferenceNotFoundException;
 
-  CommitLogEntry modifyCommitLogEntryWithKeyList(
+  /**
+   * Populates the aggregated key-list for the given {@code entry} and returns it.
+   *
+   * @param entry the {@link CommitLogEntry} to build the aggregated key list for
+   * @param inMemoryCommits when
+   * @return commit-log-entry with the aggregated key-list. The returned {@link CommitLogEntry} has
+   *     not been persisted.
+   */
+  CommitLogEntry rebuildKeyList(
       CommitLogEntry entry, @Nonnull Function<Hash, CommitLogEntry> inMemoryCommits)
       throws ReferenceNotFoundException;
 }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
@@ -19,10 +19,12 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.errorprone.annotations.MustBeClosed;
 import com.google.protobuf.ByteString;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Stream;
+import javax.annotation.Nonnull;
 import org.projectnessie.versioned.ContentAttachment;
 import org.projectnessie.versioned.ContentAttachmentKey;
 import org.projectnessie.versioned.Diff;
@@ -114,6 +116,14 @@ public interface DatabaseAdapter {
    */
   @MustBeClosed
   Stream<CommitLogEntry> commitLog(Hash offset) throws ReferenceNotFoundException;
+
+  /**
+   * Loads a single commit log entry.
+   *
+   * @return the loaded {@link CommitLogEntry}s, non-existing entries will not be returned.
+   */
+  @MustBeClosed
+  Stream<CommitLogEntry> fetchCommitLogEntries(Stream<Hash> hashes);
 
   /**
    * Retrieve the content-keys that are "present" for the specified commit.
@@ -355,4 +365,27 @@ public interface DatabaseAdapter {
 
   @VisibleForTesting
   void assertCleanStateForTests();
+
+  /**
+   * Write multiple new commit-entries, the given commit entries are to be persisted as is. All
+   * values of the given {@link CommitLogEntry} can be considered valid and consistent.
+   *
+   * <p>Implementations however can enforce strict consistency checks/guarantees, like a best-effort
+   * approach to prevent hash-collisions but without any other consistency checks/guarantees.
+   */
+  void writeMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceConflictException;
+
+  /**
+   * Updates multiple new commit-entries, the given commit entries are to be persisted as is. All
+   * values of the given {@link CommitLogEntry} can be considered valid and consistent.
+   *
+   * <p>Implementations however <em>can</em> enforce strict consistency checks/guarantees.
+   */
+  void updateMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceNotFoundException;
+
+  CommitLogEntry modifyCommitLogEntryWithKeyList(
+      CommitLogEntry entry, @Nonnull Function<Hash, CommitLogEntry> inMemoryCommits)
+      throws ReferenceNotFoundException;
 }
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
@@ -399,7 +399,7 @@ public interface DatabaseAdapter {
    * Populates the aggregated key-list for the given {@code entry} and returns it.
    *
    * @param entry the {@link CommitLogEntry} to build the aggregated key list for
-   * @param inMemoryCommits when
+   * @param inMemoryCommits function to retrieve not-yet-written commit-log-entries
    * @return commit-log-entry with the aggregated key-list. The returned {@link CommitLogEntry} has
    *     not been persisted.
    */
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
@@ -186,6 +186,41 @@ public abstract class AbstractDatabaseAdapter<
     return NO_ANCESTOR;
   }
 
+  @Override
+  public Stream<CommitLogEntry> fetchCommitLogEntries(Stream<Hash> hashes) {
+    OP_CONTEXT ctx = borrowConnection();
+
+    BatchSpliterator<Hash, CommitLogEntry> commitFetcher =
+        new BatchSpliterator<>(
+            config.getParentsPerCommit(),
+            hashes,
+            h -> fetchMultipleFromCommitLog(ctx, h, ignore -> null).spliterator(),
+            0);
+
+    return StreamSupport.stream(commitFetcher, false)
+        .onClose(
+            () -> {
+              try {
+                ctx.close();
+              } catch (Exception e) {
+                throw new RuntimeException(e);
+              }
+            });
+  }
+
+  @Override
+  public CommitLogEntry modifyCommitLogEntryWithKeyList(
+      CommitLogEntry entry, @Nonnull Function<Hash, CommitLogEntry> inMemoryCommits)
+      throws ReferenceNotFoundException {
+    try (OP_CONTEXT ctx = borrowConnection()) {
+      return buildKeyList(ctx, entry, h -> {}, inMemoryCommits);
+    } catch (ReferenceNotFoundException e) {
+      throw e;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
   // /////////////////////////////////////////////////////////////////////////////////////////////
   // DatabaseAdapter subclass API (protected)
   // /////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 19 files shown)


XN137 (commented):
a few questions + suggestions

it seems like we miss tests for the newly added adapter methods, especially around `updateMultipleCommits` throwing a `ReferenceNotFoundException`

i know there are kind of a special case since mostly used for import/export but still its part of the general `DatabaseAdapter` contract so worth "enforcing" via tests WDYT?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -113,6 +115,14 @@ Map<Key, ContentAndState> values(Hash commit, Collection<Key> keys, KeyFilterPre
   @MustBeClosed
   Stream<CommitLogEntry> commitLog(Hash offset) throws ReferenceNotFoundException;
 
+  /**
+   * Loads a single commit log entry.
```
dimas-b, line 119: suggestion: `Loads individual commit log entries`

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -353,4 +363,27 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+   * values of the given {@link CommitLogEntry} can be considered valid and consistent.
+   *
+   * <p>Implementations however can enforce strict consistency checks/guarantees, like a best-effort
+   * approach to prevent hash-collisions but without any other consistency checks/guarantees.
+   */
+  void writeMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceConflictException;
+
+  /**
+   * Updates multiple new commit-entries, the given commit entries are to be persisted as is. All
```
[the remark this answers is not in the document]
dimas-b: I agree with @XN137 's comment

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/DatabaseAdapter.java
```diff
@@ -353,4 +363,27 @@ void updateRepositoryDescription(Function<RepoDescription, RepoDescription> upda
...
+   * values of the given {@link CommitLogEntry} can be considered valid and consistent.
+   *
+   * <p>Implementations however <em>can</em> enforce strict consistency checks/guarantees.
+   */
+  void updateMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceNotFoundException;
+
+  CommitLogEntry modifyCommitLogEntryWithKeyList(
```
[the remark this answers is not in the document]
dimas-b: Simply `rebuildKeyList` perhaps?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -186,6 +186,41 @@ public Hash noAncestorHash() {
...
 
+  @Override
+  public Stream<CommitLogEntry> fetchCommitLogEntries(Stream<Hash> hashes) {
+    OP_CONTEXT ctx = borrowConnection();
+
+    BatchSpliterator<Hash, CommitLogEntry> commitFetcher =
+        new BatchSpliterator<>(
+            config.getParentsPerCommit(),
+            hashes,
+            h -> fetchMultipleFromCommitLog(ctx, h, ignore -> null).spliterator(),
```
dimas-b, line 197: nit: `h` -> `hashes` for clarity?

dimas-b, line 197: Maybe add constant for `ignore -> null`? It appears to be used in a few places.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/TracingDatabaseAdapter.java
```diff
@@ -102,6 +104,13 @@ public Map<Key, ContentAndState> values(
     }
   }
 
+  @Override
+  public Stream<CommitLogEntry> fetchCommitLogEntries(Stream<Hash> hashes) {
+    try (Traced ignore = trace("fetchCommitLogEntries")) {
```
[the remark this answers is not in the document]
dimas-b: +1 to consistency. However, I'm not sure our overall approach to tracing stream processing is the best we could have under the OpenTracing framework. Substantial work can be performed by inner code during stream operations, but currently it will be attributed to the outer span, AFAIK. This is beyond the scope of this PR, of course. I would like to revisit this at some point, though.

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
```diff
@@ -632,6 +633,30 @@ protected void doWriteMultipleCommits(
...
+                        new Document("$set", d),
+                        new UpdateOptions().upsert(false)))
+            .collect(Collectors.toList());
+
+    BulkWriteResult result = client.getCommitLog().bulkWrite(requests);
+
+    verifyAcknowledged(result, client.getCommitLog());
+
+    if (result.getMatchedCount() != entries.size()) {
+      throw new ReferenceNotFoundException("");
```
dimas-b, line 656: Not sure this is the right exception type for this kind of failure. I think we ought to introduce a more specific exception for missing commits. I suppose we should expose this as an "Internal Server Error" as opposed to `NessieReferenceNotFoundException`, which would currently be the case, AFAIK.

versioned/persist/dynamodb/src/main/java/org/projectnessie/versioned/persist/dynamodb/DynamoDatabaseAdapter.java
```diff
@@ -252,6 +252,16 @@ protected void unsafeWriteGlobalPointer(
   @Override
   protected void doWriteMultipleCommits(
       NonTransactionalOperationContext ctx, List<CommitLogEntry> entries) {
+    persistMultipleCommits(entries);
+  }
+
+  @Override
+  protected void doUpdateMultipleCommits(
+      NonTransactionalOperationContext ctx, List<CommitLogEntry> entries) {
+    persistMultipleCommits(entries);
```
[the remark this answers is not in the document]
snazy: This not - other database adapters do.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: Clashes with `hashes`

snazy: Better not - no need for constants there (findbugs/checkstyle)

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: Why not? It means that commits to be updated do not exist.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractWriteUpdateCommits.java
```diff
@@ -0,0 +1,122 @@
...
+    List<Hash> hashes =
+        IntStream.rangeClosed(1, 10).mapToObj(this::intAsHash).collect(Collectors.toList());
+    List<CommitLogEntry> commits =
+        hashes.stream()
+            .mapToInt(h -> Integer.parseInt(h.asString(), 16))
+            .mapToObj(this::helloCommit)
+            .collect(Collectors.toList());
+
+    databaseAdapter.writeMultipleCommits(commits.subList(0, 2));
+    assertThatThrownBy(() -> databaseAdapter.writeMultipleCommits(commits.subList(0, 2)))
```
XN137, line 116: are we intentionally only writing the sublist?
would an exception still be thrown if we wrote the first 3 elements? i.e. it included 1 non existing ?
if this behavior has certain limitations that we need to respect in this test, then maybe explain them in a comment here.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractWriteUpdateCommits.java
```diff
@@ -0,0 +1,122 @@
...
+        .isInstanceOf(ReferenceConflictException.class);
+
+    assertThatThrownBy(() -> databaseAdapter.updateMultipleCommits(commits))
+        .isInstanceOf(ReferenceNotFoundException.class);
```
XN137, line 120: do the first 2 get updated? or does the whole batch fail atomically?

i would assume the latter to be true for the TX adapters for example.
would this be worth testing explicitly?

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractDatabaseAdapterTest.java
```diff
@@ -33,6 +33,10 @@
 public abstract class AbstractDatabaseAdapterTest {
   @NessieDbAdapter protected static DatabaseAdapter databaseAdapter;
 
+  protected boolean commitWritesValidated() {
```
XN137, line 36: personally not a fan of the iceberg style naming...(i.e. a method starting with "commit" not being an action)
how about `hasExistenceCheckOnCommitWrite` ?

and maybe add short javadoc what this means for a DB adapter...

but this makes me wonder, should this method maybe exist on DB-adapters directly? :thinking:
in a way it describes whether the commit-write methods of a DB-adapter really can throw the `ReferenceNotFoundException` as claimed by the method signature.
WDYT?

XN137 (commented):
thanks for the updates, especially for adding tests.
looks good overall but still a few questions...
maybe its overly pedantic but i feel like clearly defining the DB-adapter contract will be beneficial since its a central piece.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
dimas-b: fair enough

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
snazy: `Deteched` is also a reference.

dimas-b: Yes, but my point is: will this error be based on a user-provided reference?

dimas-b: We do not even indicate what is not found :thinking: 

Could we have a message that would allow interpreting the error in the right way if it ever shows up in API output?

snazy: No human interaction for this call

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractDatabaseAdapterTest.java
dimas-b: As for me the name sounds fine ("commit" can be a noun too).

Moving the method to the `DatabaseAdapter` interface sounds like a good idea to me because it's the adapter implementation that controls this behaviour. Are there any adverse effects of that?

snazy: Users of DA should _never_ assume anything derived from that fact

dimas-b: If not `DatabaseAdapter`, maybe `DatabaseAdapterFactory`?

snazy: No - whether the database is transactional or cannot "insert" the same key again is an internal concern

versioned/persist/mongodb/src/main/java/org/projectnessie/versioned/persist/mongodb/MongoDatabaseAdapter.java
dimas-b: It still does not feel right to me to throw a `ReferenceNotFoundException` here. 

WDYT about a different exception type? This failure may occur when some commits are updated and some not, which is still different from the "normal" cases of "reference not found".

All in all, I do not feel very strongly about this. I'd be ok with current code if there no simple alternative.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractDatabaseAdapterTest.java
dimas-b: fair enough.

dimas-b (commented):
Overall the PR LGTM. I just have minor concerns about the new exception throwing places (e.g. in the Mongo Adapter). I'd be ok with merging if @XN137 is ok :)

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -1657,6 +1685,9 @@ protected final void writeMultipleCommits(OP_CONTEXT ctx, List<CommitLogEntry> e
   protected abstract void doWriteMultipleCommits(OP_CONTEXT ctx, List<CommitLogEntry> entries)
       throws ReferenceConflictException;
 
+  protected abstract void doUpdateMultipleCommits(OP_CONTEXT ctx, List<CommitLogEntry> entries)
```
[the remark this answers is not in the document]
XN137: consistency would be nice but i guess it should not block the PR

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractDatabaseAdapterTest.java
XN137: to some degree that internal concern is forced onto the user of the adapter by having to handle `ReferenceNotFoundException` even though there are no good rules on when it happens, what exactly the problem was nor how to appropriately handle it.
i guess its just not easy to model this nicely with the many db adapters that we have.

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractWriteUpdateCommits.java
XN137: so i guess we dont want to test the TX adapter as strictly as possible here, fine i guess when we dont want users to rely on this exact behavior

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -691,6 +691,28 @@ private static ContentIdAndBytes globalContentFromRow(ResultSet rs) throws SQLEx
...
+  }
+
+  @Override
+  public void updateMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceNotFoundException {
+    try (ConnectionWrapper conn = borrowConnection()) {
+      doUpdateMultipleCommits(conn, commitLogEntries);
+      conn.commit();
+    } catch (ReferenceNotFoundException e) {
+      throw e;
```
XN137, line 712: nit: seems redundant

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -691,6 +691,28 @@ private static ContentIdAndBytes globalContentFromRow(ResultSet rs) throws SQLEx
...
   }
 
+  @Override
+  public void writeMultipleCommits(List<CommitLogEntry> commitLogEntries)
+      throws ReferenceConflictException {
+    try (ConnectionWrapper conn = borrowConnection()) {
+      doWriteMultipleCommits(conn, commitLogEntries);
+      conn.commit();
+    } catch (ReferenceConflictException e) {
+      throw e;
```
XN137, line 701: nit: seems redundant

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1132,23 +1176,44 @@ protected <T> void writeMany(
...
+          ps.setBytes(3, serializer.apply(e));
+        }
         ps.addBatch();
         cnt++;
         if (cnt == config.getBatchSize()) {
-          ps.executeBatch();
+          int[] result = ps.executeBatch();
+          if (update) {
+            for (int i : result) {
+              if (i != 1) {
```
XN137, line 1200: nit: according to the docs this can also return `java.sql.Statement#SUCCESS_NO_INFO`, but i guess we can assume all the driver+db combinations provide this info here

XN137 (approved):

PR #5057, opened 2022-08-30
Update python transplant test for CLI

### LINKED ISSUE #5040: Python transplant test fails for cli
The transplant test in `test_nessie_cli` expects commits to be individually committed to target branch, however they are being squashed into one causing the test to fail.

The API implementations have a _keepIndividualCommits_ parameter which dictates whether or not transplant squashes commits, but this seems to be false for cli operations by default.

kamalbhavyam:
As the default behaviour for cli transplant is to squash commits into one before committing to target branch, transplant test has been updated to expect the same behaviour.
The test also checks that the message now has the formatted version of messages from all squashed commits.

Temporarily fixes #5040 
Full fix would include adding a flag to allow users to define if commits should be made individually.

### DIFF
--- a/python/tests/test_nessie_cli.py
+++ b/python/tests/test_nessie_cli.py
@@ -348,16 +348,18 @@ def test_merge_detached() -> None:
 def test_transplant() -> None:
     """Test transplant operation."""
     execute_cli_command(["branch", "dev"])
-    make_commit("transplant.foo.bar", _new_table("test_transplant_1"), "dev")
-    make_commit("bar.bar", _new_table("test_transplant_2"), "dev")
-    make_commit("foo.baz", _new_table("test_transplant_3"), "dev")
+    make_commit("transplant.foo.bar", _new_table("test_transplant_1"), "dev", message="commit 1")
+    make_commit("bar.bar", _new_table("test_transplant_2"), "dev", message="commit 2")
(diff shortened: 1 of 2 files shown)


python/tests/test_nessie_cli.py
```diff
@@ -348,16 +348,18 @@ def test_merge_detached() -> None:
...
     main_hash = ref_hash("main")
     logs = simplejson.loads(execute_cli_command(["--json", "log", "dev"]))
     first_hash = [i["hash"] for i in logs]
     execute_cli_command(["cherry-pick", "-c", main_hash, "-s", "dev", first_hash[1], first_hash[0]])
 
+    # The default behaviour for cherry-pick is to squash commits into one
     logs = simplejson.loads(execute_cli_command(["--json", "log"]))
-    assert len(logs) == 2  # two commits were transplanted into an empty `main`
+    assert len(logs) == 1  # two commits were squashed into one and transplanted into an empty `main`
+    assert_that(logs[0]["message"]).is_equal_to("commit 2\n---------------------------------------------\ncommit 3")
```
XN137, line 362: why is `commit 1` is missing?

kamalbhavyam: 2 of the 3 commits are cherry-picked. The first commit is left out.

kamalbhavyam:
Closing in favour of #5059

PR #5059, opened 2022-08-30
Fix default behaviour for transplant in CLI

### LINKED ISSUE #5040: Python transplant test fails for cli

kamalbhavyam:
Added a keepIndividualCommit parameter (set to true) to Transplant model, fixing default squash behaviour for transplant command.
- Default squash behaviour for transplant/merge was changed by PR [#3813](https://github.com/projectnessie/nessie/pull/3813) but the CLI tests still expected transplanted commits to be present as individual commits on target branch.
- Transplants should maintain individual commits for CLI commands so an explicit parameter has been added to the model that ensures this behaviour by setting it explicitly in the API call.

Possible Enhancements: 
- Fix API to default keepIndividualCommits as True for transplant.
- Add flag to allow user to squash commits in transplant.

Fixes #5040

### DIFF
--- a/python/pynessie/model.py
+++ b/python/pynessie/model.py
@@ -442,6 +442,9 @@ class Transplant:
 
     from_ref_name: str = attr.ib(metadata=desert.metadata(fields.Str(data_key="fromRefName")))
     hashes_to_transplant: List[str] = attr.ib(metadata=desert.metadata(fields.List(fields.Str(), data_key="hashesToTransplant")))
+    keep_individual_commits: bool = attr.ib(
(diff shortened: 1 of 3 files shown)


XN137 (approved):
lgtm +1

maybe add something like this to the PR description for additional context:
```
- default squash behavior for transplant/merge was changed in #<PRnumber> but the tests were not updated
- after discussing with the team we think the CLI cmd that do transplants should still default to not squashing, so we need to pass an explicit param to the http api now
```

PR #5075, opened 2022-09-01
Workaround OOMs during Quarkus test runs

snazy:
A recent PR CI run timed out, because of OOMs during
`:nessie-quarkus-cli:intTest`. This is reproducible locally, and
investigating some heap and thread dumps showed that the heap
pressure is actually caused by jacoco/class-instrumentation, which
reads many jar files. This might be a combination of Quarkus class
loading and jacoco instrumentation.

TL;DR the code added to `:nessie-quarkus-cli` caused more code to
instrumented, which led to the increased heap pressure, which in
turn caused the OOMs.

The default heap sizes for Quarkus tests is 256m..1024m. This
change bumps those values to 512m..1536m and moves that
configuration to `nessie-conventions`.

Note: the heap size for tests can also be controlled manually
using for example `-PtestHeapSize=2g` on the command line.

### DIFF
--- a/buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
+++ b/buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
@@ -101,10 +101,6 @@ fun Project.testTasks() {
       if (testJvmArgs != null) {
         jvmArgs((testJvmArgs as String).split(" "))
       }
-      if (testHeapSize != null) {
-        setMinHeapSize(testHeapSize)
-        setMaxHeapSize(testHeapSize)
-      }
 
       systemProperty("file.encoding", "UTF-8")
       systemProperty("user.language", "en")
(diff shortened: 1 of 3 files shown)


snazy:
Failed PR CI run: https://github.com/projectnessie/nessie/runs/8114528819?check_suite_focus=true

buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
```diff
@@ -127,6 +123,19 @@ fun Project.testTasks() {
...
+        jvmArgs("--add-opens=java.base/java.util=ALL-UNNAMED")
+        systemProperty("quarkus.log.level", testLogLevel())
+        systemProperty("quarkus.log.console.level", testLogLevel())
+        systemProperty("http.access.log.level", testLogLevel())
+
+        minHeapSize = if (testHeapSize != null) testHeapSize as String else "512m"
+        maxHeapSize = if (testHeapSize != null) testHeapSize as String else "1536m"
+      } else if (testHeapSize != null) {
+        setMinHeapSize(testHeapSize)
+        setMaxHeapSize(testHeapSize)
```
XN137, line 137: nit: is there a difference between `setMaxHeapSize(x)` and `maxHeapSize = x` ?

buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
```diff
@@ -127,6 +123,19 @@ fun Project.testTasks() {
...
       }
+
+      if (plugins.withType<QuarkusPlugin>().isNotEmpty()) {
+        jvmArgs("--add-opens=java.base/java.util=ALL-UNNAMED")
+        systemProperty("quarkus.log.level", testLogLevel())
+        systemProperty("quarkus.log.console.level", testLogLevel())
+        systemProperty("http.access.log.level", testLogLevel())
+
+        minHeapSize = if (testHeapSize != null) testHeapSize as String else "512m"
+        maxHeapSize = if (testHeapSize != null) testHeapSize as String else "1536m"
```
XN137, line 134: nit: might be a use-case for https://kotlinlang.org/docs/null-safety.html#elvis-operator

XN137 (approved):

buildSrc/src/main/kotlin/nessie-conventions.gradle.kts
snazy: no, just different ways

PR #5076, opened 2022-09-01
picocli loses output to stdout

snazy:
The default `PrintWriter`s created by picocli are derived from
`System.out/err`, but sadly not flushed when a command finishes.

This change adds "safety flush".


XN137 (approved):

dimas-b (commented):
Nice catch :+1:

PR #5077, opened 2022-09-01
pynessie: add default timeout to http calls

XN137:
- defaults to 60 seconds
- controllable by env var `PYNESSIE_HTTP_TIMEOUT_SEC`
- values <= 0 mean no timeout

### DIFF
--- a/python/pynessie/client/_endpoints.py
+++ b/python/pynessie/client/_endpoints.py
@@ -15,6 +15,7 @@
 
 """Direct API operations on Nessie with requests."""
 
+import os
 from typing import Any, Optional, Union, cast
 
 import requests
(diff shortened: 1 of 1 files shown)


XN137:
whats the benefit do you see of reusing the other one?
as is we merge 1 of the PRs and close the other, which one is which shouldnt matter imo

XN137:
> miss the version-bump handling for the release notes

i dont think its particularly interesting for end users to know what tools versions we use for linting our code.
but release notes are a good point in that we should mention this behavior change explicitly,
so i've changed this PR to just be about adding the default http timeouts.

we can then later rebase the dependabot PR

snazy (approved):

PR #5081, opened 2022-09-05
SQLExtension: Fix show reference command with use reference@hash

ajantha-bhat:
Whenever SQL "USE REFERENCE @ HASH" is used, It only updates the [NessieIcebergClient](https://github.com/apache/iceberg/blob/master/nessie/src/main/java/org/apache/iceberg/nessie/NessieIcebergClient.java) 's `updatableReference`. So, there is no change in the actual Nessie reference in the backend.

But when current reference info is fetched, It fetches from the backend which is not aware of "USE REFERENCE @ HASH".  So, this is confusing for the users and also it is not the current state of the reference.  Hence, this fix.

Also, assign ref, create ref from, and merge ref will honor the value set by "use reference @hash" now.

### DIFF
--- a/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
+++ b/clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
@@ -22,6 +22,7 @@ import org.projectnessie.client.StreamingUtil
 import org.projectnessie.client.api.NessieApiV1
 import org.projectnessie.client.http.HttpClientBuilder
 import org.projectnessie.error.NessieNotFoundException
+import org.projectnessie.model.Reference.ReferenceType
 import org.projectnessie.model.{
   Branch,
   ImmutableBranch,
@@ -246,7 +247,16 @@ object NessieUtils {
   ): Reference = {
     val refName = getCurrentRefName(currentCatalog, catalog)
     try {
-      api.getReference.refName(refName).get
+      var ref = api.getReference.refName(refName).get
+      val refHash = getCurrentRefHash(currentCatalog, catalog);
+      if (refHash != null) {
+        if (ref.getType == ReferenceType.BRANCH) {
+          ref = Branch.of(ref.getName, refHash)
+        } else {
+          ref = Tag.of(ref.getName, refHash)
+        }
+      }
+      ref
     } catch {
       case e: NessieNotFoundException =>
         throw new NessieNotFoundException(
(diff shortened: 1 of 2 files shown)


clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java
```diff
@@ -374,7 +374,7 @@ void useShowReferencesAtTimestampWithTimeZone(DateTimeFormatter dateTimeFormatte
     assertThat(sql("USE REFERENCE %s AT `%s` IN nessie ", refName, timeWithZone))
         .containsExactly(row("Branch", refName, lastCommitBeforeTimePredicate.getHash()));
     assertThat(sql("SHOW REFERENCE IN nessie"))
-        .containsExactly(row("Branch", refName, lastRefHashGlobally));
+        .containsExactly(row("Branch", refName, lastCommitBeforeTimePredicate.getHash()));
```
snazy, line 377: Not sure why this change is necessary, because the logic for timestamps shouldn't change, right?
And there's no more check against the reason mentioned in the comment - and the variable is now unused.

clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
```diff
@@ -246,7 +247,16 @@ object NessieUtils {
   ): Reference = {
     val refName = getCurrentRefName(currentCatalog, catalog)
```
snazy, line 248: Why not update the existing getCurrentRefName to return a tuple of name + optional-hash?

clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java
ajantha-bhat: > Not sure why this change is necessary, because the logic for timestamps shouldn't change, right?

`USE ref AT hash or timestamp` has the same behaviour (just that timestamp will be mapped to a hash -> uses `NessieUtils#calculateRef`). 

Without this PR, after setting `USE reference at oldHashTimestamp`, show reference was always giving the head commit hash of that reference. Hence, need to modify it now as the show reference will return the current hash under use.

> and the variable is now unused.

Oh, I will remove it.

clients/spark-extensions-base/src/main/scala/org/apache/spark/sql/execution/datasources/v2/NessieUtils.scala
ajantha-bhat: originally thought of changing but didn't change as I had to explore some Scala syntaxes if I change.

Updated now.

snazy (approved):

PR #5101, opened 2022-09-05
Upgrade to codecov-action v3

XN137:
see https://github.com/codecov/codecov-action/releases


snazy (dismissed):
Still the same error :(

But let's bump the action anyway - just merge with a different PR title

snazy (approved):

PR #5103, opened 2022-09-05
Deps: Remove `versionJacksonSpark3`

snazy:
There was a bug in some older versions which required us to bump
the Spark 3 related Jackson dependencies to a "sub-patch" version.
This workaround is no longer necessary and dep-mgmt can use the
`jackson-bom`.


XN137 (approved):
just wondering:
do we know what made the workaround obsolete?

snazy:
> just wondering:
> do we know what made the workaround 

It‘s long ago, many Jackson releases

dimas-b (approved):

PR #5110, opened 2022-09-06
Enable dependabot for github actions

XN137:
basically copied from:
https://github.com/apache/iceberg/blob/1017bb11a98f075bcf5b7e1cf99bf32adcc797cf/.github/dependabot.yml#L28-L32

it should notify us about new major versions like here:
https://github.com/apache/iceberg/commit/e5d771b300e7ec5ed0a6da9e476697088a2978b8

we only run on sundays to avoid interference with the workweek

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,10 @@
 version: 2
 updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: sunday
 - package-ecosystem: gradle
   directory: "/"
   schedule:


snazy (commented):
Can you try it on the main branch of your fork first?

XN137:
doesnt work on the fork:
```
Dependabot failed to update your dependencies because there was an error parsing the main.yml found at /.github/workflows/main.yml.

Dependabot encountered the following error:

Dependabot::DependencyFileNotParseable
```
at https://github.com/XN137/nessie/network/updates

i cant anything related to that error in the logs or communities etc.
i am guessing it is related to our usage of `uses: ./.github/actions/patch-git` which requires a previous `checkout`... i guess dependabot cant do that and thus cant exactly parse the workflow definition.

thus closing.

PR #5142, opened 2022-09-09
Testing pre-requisites for Nessie GC: Two S3 testing projects

snazy:
* s3minio - JUnit extension to provide a Minio instance
* s3mock - S3 endpoint serving data via functions, no persistence


testing/s3mock/src/main/java/org/projectnessie/s3mock/data/Bucket.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableBucket.class)
+@JsonDeserialize(as = ImmutableBucket.class)
+public interface Bucket {
```
ajantha-bhat, line 26: Can we just depend on this library (https://github.com/adobe/S3Mock), instead of we implementing or maintaining the similar/same code?

snazy: If you look closely at both implementations, you see that those do _different_ things.

ajantha-bhat: I do find the data folder in this PR to be similar to https://github.com/adobe/S3Mock/tree/main/server/src/main/java/com/adobe/testing/s3mock/dto

But yeah, I do have to look closely as you mentioned.

XN137: imo this is a legitimate question, so my suggestion would be to briefly summarize the reasoning on why we can not use this library in a README file inside the module folder, WDYT ?
this way we dont need to expect everyone to look at both implementations closely to tell what the exact difference is and why we had to roll our own.

snazy: Again - the Adobe S3Mock does something very different. This one does intentionally _NOT_ serve "real" content.

snazy: Not sure why comparing with another library, that serves a different purpose, would make sense.
This one serves content and listings from _functions_ not a real file system incl put operations.

ajantha-bhat: Thanks for the explanation. I may need a while to understand all these object store interface implementations as I am not very familiar with them. 

If no one else reviews, I will dig deep to understand these.

ajantha-bhat (approved):
Compared the DTO with other standard implementations. This looks like a subset of it (only required functionalities).
Util and resource files I have reviewed as per my knowledge! (which may be very limited)

As these are just test code, I don't want to block other dependent PRs. Hence, I am ok in merging if no one else reviews this week.

PR #5143, opened 2022-09-09
Nessie GC: Remove current gc-base

snazy:


dimas-b (approved):
Is it going to be a complete rewrite?

PR #5144, opened 2022-09-09
Nessie GC: mark & sweep (`gc-base` module only)

### LINKED ISSUE #4695: GC: Abstract distributed computing

snazy:
See also the readme in `./gc/README.md`.

GC is implemented as a two-phase approach:
1. Identify all refetences to live content versions by walking all
   named references. Content versions for Iceberg are tuples of
   commit-ID + snapshot ID + content ID + metadata reference).
   These `ContentReference`s are and stored in an instance of
   `LiveContentSetsStore`. This is the "mark phase". The set of
   content-references is called "live contents set".
2. For each content-ID, resolve all `ContentReferences` to file
   objects and populate a bloom-filter with those. Then traverse
   the base storage locations for each content-ID and delete all
   files that are not contained in the bloom filter, but do not
   delete files that are "too new" (created since the "mark phase"
   started).

A command line tool `nessie-gc-shell` is provided as well. It
supports running mark-and-sweep, mark only, sweep only plus a few
supplemental commands.

All pieces are pluggable, there are Java interfaces (or abstract
classes) providing the API for all pieces. This makes the code much
easier to test and helps using alternative implementations.

The base functionality is contained in the `nessie-gc-base` project.
This has default implementations to identify live content
references, an in-memory live-content-sets-storage and a default
implementation to run expire per content.

Support for content types, like Iceberg tables, implements
the functionality to expand content-references to actual
file-objects, list file-objects from base locations and delete
files. The "sweep" phase for Iceberg therefore also acts as the
"Nessie aware delete orphan files".

A live-content-sets-storage implementation using JDBC is present as
well.  Small Nessie repositories using a "one-off" mark-and-sweep
run can get away with the in-memory contents-storage. But big Nessie
repositories would require too much heap.

Another reason to eventually use a persistent
live-content-sets-storage is to allow the deletion of the files
occupied by completely unreferenced tables - tables that are not
referenced by any live Nessie commit.  This functionality is not
implemented, but would work by collecting the base-locations for
all content-IDs that are no longer present in the most recent
live-contents-set.

This PR still has a few non-`Test*` classes that were only added to
prove that the concept works quick and does not require much heap
and does not use a lot of CPU. Those "tests" in `nessie-gc-base` and
`nessie-gc-iceberg-inttest` will be removed, because those add no
real value and were only needed to validate the concept. The "final"
proof showed that a mark-and-sweep (deleting ~1.7M files, collected
from "many" branches, commits, content objects) alone can complete
in round about one minute with a 2GB Java heap.

Fixes #4695 #4694 #4626 #2255 #706


ajantha-bhat (commented):
Just did a first level of review (didn't go through testcases in depth yet)

Nice PR. Here are my concerns and questions on the implementation. 

1.  Simple and impressive design of expiry (compute live content files and add into bloom filter, do list files at table path, check the list files result against bloom filter and delete it).

Iceberg also would have done the same instead of providing the two methods, one for `remove_orphan_files` and one for `expire_snapshots`. But the main reason that Iceberg has two implementation is because orphan files are rare and listing files is very slow when number of files grows. So, `expire_snapshots` works without a list file (frequently called gc api).

So, my concern is doing listFiles() for everytime gc run (as the data file and metadata file count grows in table)

2. One bloom filer per content id may be expensive (I remember getting same comment to my PR), may be we can keep a global bloom filter for each run.

3. Time travel for dropped reference is not handled. We should also traverse unreferenced heads while computing the live contents.

4. `PersistenceSpi`, even though we have in-memory impl and end to end PR mentions about postgres db based impl. I think forcing users to use postgres when user is on non-transactional db like mongo is not a good idea. Can we make use the user's version store itself to store this information?

5. Last but not least, concern about using multithreaded identify and expire instead of distributed jobs. But , if the real time test passes with good number then this may not be the concern.

gc/gc-base/src/main/java/org/projectnessie/gc/repository/Nessie.java
```diff
@@ -0,0 +1,130 @@
...
+    this.api = api;
+  }
+
+  public static RepositoryConnector nessie(NessieApiV1 api) {
+    return new Nessie(api);
+  }
+
+  @Override
+  public Stream<Reference> allReferences() throws NessieNotFoundException {
+    return api.getAllReferences().stream();
```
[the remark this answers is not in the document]
snazy: Sure, an already dropped reference might have been live at cut-off timestamp.
But does it matter anymore, if the reference is dropped?

gc/gc-base/src/main/java/org/projectnessie/gc/identify/CutoffPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      @Override
+      public String toString() {
+        return "cutoff at timestamp " + cutoffTimestamp;
+      }
+    };
+  }
+
+  static CutoffPolicy numCommits(int commits) {
```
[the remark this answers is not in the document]
snazy: Number-of-commits can be handy for a couple of use cases:
* for testing purposes
* for development/experiment branches
* for the explicit setting (e.g. tags): only keep the most recent commit

gc/gc-base/src/main/java/org/projectnessie/gc/identify/CutoffPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+        public String toString() {
+          return "no cutoff (keep everything)";
+        }
+      };
+
+  /**
+   * The timestamp of the "cutoff point" in a reference or {@link #NO_TIMESTAMP} for "no timestamp",
+   * never {@code null}.
+   */
+  default Instant timestamp() {
```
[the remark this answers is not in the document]
snazy: Nope. This function is used by the identify functionality to get that timestamp.

gc/gc-base/src/main/java/org/projectnessie/gc/files/NessieFileIOException.java
```diff
@@ -0,0 +1,31 @@
...
+
+  public NessieFileIOException(String message) {
+    super(message);
+  }
+
+  public NessieFileIOException(String message, Throwable cause) {
+    super(message, cause);
+  }
+
+  public NessieFileIOException(Throwable cause) {
```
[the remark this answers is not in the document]
snazy: Nope.

gc/gc-base/src/main/java/org/projectnessie/gc/expire/PerContentDeleteExpired.java
```diff
@@ -0,0 +1,245 @@
...
+    @CanIgnoreReturnValue
+    Builder contentId(String contentId);
+
+    PerContentDeleteExpired build();
+  }
+
+  /** Returns a stream of files that can be expired. */
+  @SuppressWarnings("UnstableApiUsage")
+  public DeleteSummary expire() {
+    BloomFilter<URI> filter = createBloomFilter();
```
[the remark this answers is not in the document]
snazy: The number of concurrently processed content-ids is restricted (default = 4), so the number of concurrently used BFs is restricted as well.

gc/gc-base/src/main/java/org/projectnessie/gc/contents/ContentReference.java
```diff
@@ -0,0 +1,73 @@
...
+  Long snapshotId();
+
+  static ContentReference icebergTable(
+      @NotNull String contentId,
+      @NotNull String commitId,
+      @NotNull ContentKey contentKey,
+      @NotNull String metadataLocation,
+      long snapshotId) {
+    return ImmutableContentReference.of(
+        contentId, commitId, contentKey, Content.Type.ICEBERG_TABLE, metadataLocation, snapshotId);
```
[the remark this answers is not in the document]
snazy: Let's start with Iceberg tables for now.

snazy:
> 2. One bloom filer per content id may be expensive (I remember getting same comment to my PR), may be we can keep a global bloom filter for each run.

As mentioned above: the number of BFs is restricted to the number of concurrently processed contents (by content-id), so the heap pressure is deterministic and constrained.

> 3. Time travel for dropped reference is not handled. We should also traverse unreferenced heads while computing the live contents.

I don't see a legit reason to traverse dropped references. If a named reference's dropped, it's gone.

> 4. `PersistenceSpi`, even though we have in-memory impl and end to end PR mentions about postgres db based impl. I think forcing users to use postgres when user is on non-transactional db like mongo is not a good idea. Can we make use the user's version store itself to store this information?

We can add other persistence implementations later, if needed. It's easy to setup a PostgreSQL or H2 instance and potentially hibernate it as long as it's unused.

> 5. Last but not least, concern about using multithreaded identify and expire instead of distributed jobs. But , if the real time test passes with good number then this may not be the concern.

What's the reason for the concern? Just because?

gc/gc-base/src/main/java/org/projectnessie/gc/contents/ContentReference.java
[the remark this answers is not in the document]
ajantha-bhat: SGTM. But do we need an issue to track this?

gc/gc-base/src/main/java/org/projectnessie/gc/identify/CutoffPolicy.java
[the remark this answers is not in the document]
ajantha-bhat: Thinking more about this. 
with other catalogs, expired snapshots won't be present in the current table metadata file and the user can only time travel to the live snapshots.  
But in Nessie, all the commits are visible and users can time travel to any of them. Only that query will fail with a FileNotFoundException (if that commit is expired and gc has cleaned its content). Not so graceful experience may be.

gc/gc-base/src/main/java/org/projectnessie/gc/repository/Nessie.java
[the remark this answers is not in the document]
ajantha-bhat: ok. we can have time travel support only for the live references.

gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
```diff
@@ -112,7 +112,7 @@ public interface Builder {
     Builder liveContentSetsRepository(LiveContentSetsRepository liveContentSetsRepository);
 
     @CanIgnoreReturnValue
-    Builder contentToReference(ContentToReference contentToReference);
+    Builder contentToReference(ToContentReference toContentReference);
```
ajantha-bhat, line 115: nit: better to update the method name too?

gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
```diff
@@ -402,7 +402,7 @@ void set(LogResponse.LogEntry logEntry) {
 
   abstract LiveContentSetsRepository liveContentSetsRepository();
 
-  abstract ContentToReference contentToReference();
+  abstract ToContentReference contentToReference();
```
ajantha-bhat, line 405: nit: better to update the method name too?

gc/gc-base/src/main/java/org/projectnessie/gc/expire/PerContentDeleteExpired.java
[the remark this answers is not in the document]
ajantha-bhat: got it. Deep dived on the code today.

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
```diff
@@ -0,0 +1,320 @@
...
+                  public boolean test(Content.Type type) {
+                    return ICEBERG_TABLE == type;
+                  }
+
+                  @Override
+                  public Set<Content.Type> validTypes() {
+                    return singleton(ICEBERG_TABLE);
+                  }
+                })
+            .cutOffPolicySupplier(r -> atTimestamp(markAndSweep.cutOff))
```
ajantha-bhat, line 187: can we a testcase for number of commits cutoff policy?

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
```diff
@@ -0,0 +1,320 @@
...
+            .contentToReference(
+                (content, commitId, key) ->
+                    ContentReference.icebergTable(
+                        content.getId(),
+                        commitId,
+                        key,
+                        ((IcebergTable) content).getMetadataLocation(),
+                        ((IcebergTable) content).getSnapshotId()))
+            .liveContentSetsRepository(repository)
+            .repositoryConnector(
```
ajantha-bhat, line 197: the user who supplied `repositoryConnector` has to close() it after the usage? I didn't see close() in the identify step.

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
```diff
@@ -0,0 +1,320 @@
...
+  @MethodSource("markAndSweep")
+  public void markAndSweep(MarkAndSweep markAndSweep) throws Exception {
+
+    InMemoryPersistenceSpi storage = new InMemoryPersistenceSpi();
+    LiveContentSetsRepository repository =
+        LiveContentSetsRepository.builder().persistenceSpi(storage).build();
+    IdentifyLiveContents identify =
+        IdentifyLiveContents.builder()
+            .contentTypeFilter(
+                new ContentTypeFilter() {
```
ajantha-bhat, line 176: I hope we provide some default implementations for `ContentTypeFilter`, `contentToReference()` and `contentToFiles`() to the users, right now it looks like big code needed for calling Identify() and expire()

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
```diff
@@ -0,0 +1,320 @@
...
+                (content, commitId, key) ->
+                    ContentReference.icebergTable(
+                        content.getId(),
+                        commitId,
+                        key,
+                        ((IcebergTable) content).getMetadataLocation(),
+                        ((IcebergTable) content).getSnapshotId()))
+            .liveContentSetsRepository(repository)
+            .repositoryConnector(
+                new RepositoryConnector() {
```
ajantha-bhat, line 198: why are we not using the `NessieRepositoryConnector` for this end to end test?

ajantha-bhat:
> We can add other persistence implementations later, if needed. It's easy to setup a PostgreSQL or H2 instance and potentially hibernate it as long as it's unused.

Agree that postgres is easy to setup. I was thinking more in terms of getting approvals to use one more DB in production while the user is already using some other DB (say mongodb). Yeah, we can check about using the Nessie backend store itself to store this information.

gc/gc-base/src/main/java/org/projectnessie/gc/contents/LiveContentSet.java
```diff
@@ -0,0 +1,154 @@
...
+
+  @Nullable
+  public abstract Instant expiryCompleted();
+
+  public abstract Status status();
+
+  @Nullable
+  public abstract String errorMessage();
+
+  public void delete() {
```
ajantha-bhat, line 69: I don't see caller for this. We should call this when expiry() is success?

ajantha-bhat:
I am mostly done with this PR review. PR looks good to me 👍🏻 
(Few minor comments and a performance concern about doing `listFiles()` every time for GC. But we don't have an alternate solutions that covers all the scenario. So, appreciate the current implementation)

@dimas-b, @XN137 : Do you guys have bandwidth to take a look at this PR?

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
snazy: Yes? It's a noop in the test.

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
snazy: _Users_ should really _not_ code directly against GC.

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
snazy: It's a unit test?

gc/gc-base/src/main/java/org/projectnessie/gc/contents/LiveContentSet.java
snazy: See markdown in #4991

gc/gc-base/src/main/java/org/projectnessie/gc/contents/LiveContentSet.java
```diff
@@ -0,0 +1,154 @@
...
+    return persistenceSpi().fetchContentReferences(id(), contentId);
+  }
+
+  public void associateBaseLocations(String contentId, Collection<URI> baseLocations) {
+    // TODO detect duplicate base locations for different content-IDs
+    persistenceSpi().associateBaseLocations(id(), contentId, baseLocations);
+  }
+
+  @MustBeClosed
+  public Stream<URI> fetchAllBaseLocations() {
```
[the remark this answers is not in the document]
snazy: See #4991

gc/gc-base/src/test/java/org/projectnessie/gc/roundtrip/TestMarkAndSweep.java
ajantha-bhat: As we saw in the mailing list, some people want to use API instead of CLI.

ajantha-bhat (approved):

gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
```diff
@@ -0,0 +1,431 @@
...
+              "live-set#{}: Checking commit {} with {} operations via {}.",
+              addContents.id(),
+              commitHash,
+              operations.size(),
+              namedReference);
+          lastCommitId = commitHash;
+          numContents +=
+              addContents.addLiveContent(
+                  operations.stream()
+                      .filter(operation -> operation instanceof Operation.Put)
```
ajantha-bhat, line 287: It just occurred to me that we are not cleaning the "dropped table" files when a table dropped time is newer than cutoff time /cutoff of commit. Which is not same as "dropped reference" behaviour. 

It may cause confusion to the users. Better not add dropped table's contents to the live-content-references . 

As there was no table level end to end testcase, it was hard to catch.

ajantha-bhat: Created https://github.com/projectnessie/nessie/issues/5210 for tracking/discussions.

PR #5147, opened 2022-09-10
No longer write ref-log entries for commit/merge/transplant

snazy:
Still writes for create/assign/drop reference


ajantha-bhat:
From the PR description/changes, I understood what are the changes. But still not clear about why these changes? 
Why not remove the whole reflog (as it is already deprecated)?

snazy:
> Why not remove the whole reflog (as it is already deprecated)?

We discussed it offline a few weeks ago, to first remove these and then maybe(!) the others later.

dimas-b (approved):
IIRC, we want to keep reflog for CREATE / DELETE ref operations for now so that end users could recover accidentally deleted references (until we figure out a better solution).

PR #5148, opened 2022-09-11
Do not cache versioned-persist-bench shadow jar

snazy:
The shadow jar takes a couple seconds to cache locally, which is wasted time, because it's rarely restored from the cache, and it is relatively fast to re-create it.


snazy:
I don’t. It says „packing build cache“ or so

PR #5149, opened 2022-09-11
Fix repositories in build-tools-ITs

snazy:


XN137 (approved):
lgtm

where/how did we notice they are broken?

PR #5145, opened 2022-09-09
Fix key list lookup with hash collisions

dimas-b:
This change fixes key lookup when hash collisions cause the key to be stored
in the same segments, in front of its natural position.

PR #5174, opened 2022-09-13
Fix key list lookup when the embedded segment is empty.

dimas-b:
This change fixes key lookup when hash collisions cause the empty embedded
segment to be processed during hash search wraparound.

----

Includes #5145

### DIFF
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -168,6 +168,11 @@ final class FetchValuesUsingOpenAddressing {
         KeyListEntry keyListEntry = i < keys.length ? keys[i] : null;
         if (keyListEntry == null) {
           // key _not_ found
+          if (keys.length == 0 && segment == 0) {
+            // ... however, the first (embedded) segment may be empty due to size restrictions.
+            // This means its keys had to be moved to key list entities - keep searching.
+            keysForNextRound.add(key);
+          }
           break;
         } else if (keyListEntry.getKey().equals(key)) {
           resultConsumer.accept(keyListEntry);
--- a/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
+++ b/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
@@ -140,8 +140,16 @@ final class FetchValuesUsingOpenAddressing {
       int round, Collection<Key> remainingKeys, Consumer<KeyListEntry> resultConsumer) {
     List<Key> keysForNextRound = new ArrayList<>();
     // If we've already examined every segment (implying a miss on a completely full hashmap),
-    // then return an empty collection early
-    if (keyListsArray.length <= round) {
+    // then return an empty collection early. Note that the initial segment may need to be examined
+    // twice when the key is moved in front of its natural position due to key collisions. This
+    // is achieved by the `less than` comparison with `round`. On the first round the initial
+    // segment is scanned from the natural key position. On the last round the initial segment is
+    // scanned from index zero. This may incur some extra key comparisons if the segment is full and
+    // the scanning overtakes the "natural" key position on the last round. However, this is
+    // probably a rare case and not worth complicating the lookup logic to avoid it. In any case,
+    // the double lookup in the initial segment does not cause extra I/O because the segment is
+    // pre-fetched by the time it is reused.
+    if (keyListsArray.length < round) {
       return keysForNextRound;
     }
     for (Key key : remainingKeys) {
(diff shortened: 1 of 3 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
```diff
@@ -165,6 +165,13 @@ Collection<Key> checkForKeys(
 
       KeyListEntry[] keys = keyListsArray[segment];
       for (int i = offsetInSegment; ; i++) {
+        if (keys.length == 0 && segment == 0) {
```
snazy, line 168: Move out of the loop?

snazy (commented):
LGTM - minor comment. Can you let @dalaro take a look at the test cases?

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/FetchValuesUsingOpenAddressing.java
dimas-b: The logic becomes more complicated, IMHO, if we do that check out of the loop. We have to add an `else` around the `for`... Another `break` case aligns nicely with the previous `break` cases. In principle, the condition here is about handling a "not found" case in addition to the other not found cases.

I can move it out, if you prefer, though.

dimas-b: that sgmt :+1:

dimas-b: updated

snazy (approved):

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -301,28 +402,29 @@ private static KeyListEntity getKeyListEntity(int segmentIndex, int entryCount)
...
         .key(k)
         .payload((byte) 99)
-        .contentId(ImmutableContentId.of(keyString + "_id"))
+        .contentId(ImmutableContentId.of(k + "_id"))
         .build();
   }
 
   /** All instances return {@linkplain Integer#MAX_VALUE} from {@linkplain #hashCode()}. */
-  private static class KeyWithMaxIntHash extends Key {
+  private static class KeyWithSpecificHash extends Key {
```
dalaro, line 414: Before simplifying to `KeyWithMaxIntHash` in the final squash and merge, I had something similar to this, which got me wondering at the time about accidental equals-hashCode violation.  I don't think there's any such violation here now; it looks good.

The scope might be sufficiently restricted that this remains safe over the life of the test, with whatever edits might still lie in the future. On the other hand, if you wanted, you could maybe add an integrity check to the `equals` implementation, e.g. instead of

```
return Objects.equals(elements, that.getElements());
```

perhaps

```
boolean elementsEqual = Objects.equals(elements, that.getElements());
if (elementsEqual && hashCode != that.hashCode()) {
  throw new IllegalStateException(String.format(
    "Equal key instances (path %s) found with different hashcodes: %s vs %s",
    Joiner.on(".").join(elements), hashCode, that.hashCode()));
}
return elementsEqual;
```

This could just as well be a precondition on fewer lines.  I used an explicit if-throw just to usually avoid the Joiner.  I can't properly contextualize this as a line-comment since it lies too far outside github's diff settings.

This whole comment is pretty subjective, and it's up to you.

dalaro (dismissed):
Nice find, and thanks for the detailed comment in `FetchValuesUsingOpenAddressing`.  I spent some time stepping through the new `keyCollision` test with a debugger, particularly on the `2, 0, 5, true` case.  The test and the fix LGTM.

Borrowing some of your test nomenclature, in 8efaf76f8ce1ceea9eebb9221fb2731e4143bdf1, I failed to account for the possibility you identified here, where an element insertion can probe through every other segment besides its natural segment, find them all completely full, and eventually locate an empty bucket at a lower-than-natural bucket index within the natural segment.

My only real comment here relates to using hex base when converting segment hashes back into integers; the rest is subjective.

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
```diff
@@ -199,6 +165,136 @@ public void segmentPrefetching() {
...
+      FetchValuesUsingOpenAddressing fetch = new FetchValuesUsingOpenAddressing(entry);
+      AtomicReference<KeyListEntry> hit = new AtomicReference<>();
+      Collection<Key> remaining = ImmutableList.of(key);
+      // Simulate loading key lists in multiple rounds
+      for (int r = 0; r < segmentCount + 1; r++) {
+        // Note: Do not preload segments for each of validation
+        List<Hash> hashes = fetch.entityIdsToFetch(r, 0, ImmutableList.of(key));
+        hashes.forEach(
+            h -> {
+              int segment = Integer.parseInt(h.asString());
```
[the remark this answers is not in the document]
dimas-b: ah, yes, I missed the hex base... will fix

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
dimas-b: Implemented this check in a slightly different way.

versioned/persist/adapter/src/test/java/org/projectnessie/versioned/persist/adapter/spi/TestFetchValuesUsingOpenAddressing.java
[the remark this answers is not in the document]
dimas-b: fixed

dalaro (approved):

PR #5208, opened 2022-09-17
Nessie GC: JDBC Live-Set-Repository

snazy:
Adds functionality to persist the live-content-sets from #5144 (#4991) in a JDBC database (H2 or PostgreSQL or compatible).


ajantha-bhat (approved):
LGTM

PR #5212, opened 2022-09-19
Unsupport 0.30.x versions

snazy:


README.md
```diff
@@ -91,16 +91,12 @@ Nessie Iceberg's integration is compatible with Iceberg as in the following tabl
 | Nessie version | Iceberg version | Spark version                                                        | Hive version | Flink version  | Presto version            |
 |----------------|-----------------|----------------------------------------------------------------------|--------------|----------------|---------------------------|
 | 0.43.0         | 0.14.1          | 3.1.X (Scala 2.12), 3.2.X (Scala 2.12+2.13), 3.3.X (Scala 2.12+2.13) | n/a          | 1.14.x, 1.15.x | 0.274.x, 0.275.x, 0.276.x |
-| 0.30.0         | 0.13.2          | 3.1.X, 3.2.X (all: Scala 2.12)                                       | n/a          | 1.12.1         | 0.274.x                   |
```
ajantha-bhat, line 94: Even the latest Iceberg release (https://github.com/apache/iceberg/blob/apache-iceberg-0.14.1/versions.props#L28) uses `0.30.0`

Maybe once an Iceberg release uses a newer version of Nessie, we can unsupport it?

ajantha-bhat: I know that Iceberg is still compatible with the latest Nessie version. But the Iceberg code and documentation still point to 0.30.0 and some users might try based on that.

ajantha-bhat: Also, demos are still pointing to 0.30.0

dimas-b: Iceberg uses the 0.30.0 Nessie _client_, which should work fine with Nessie 0.43.0 _servers_, AFAIK.

It might be worth mentioning that in README, but it's not super critical, IMHO.

dimas-b (approved):

PR #5285, opened 2022-09-29
Nessie client tests: Replace JDK's HTTP server w/ Jetty

snazy:
Original intent was to just add a test that exercises bigger request + response bodies, but the test `TestHttpClient.testWriteWithVariousSizes` failed with JDK's `HttpServer`. Therefore, this change adds the test and migrates to Jetty using a version that still works with Java 8.

### DIFF
--- a/clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
+++ b/clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
@@ -121,34 +121,46 @@ public class TestHttpClient {
 
           ArrayBean inputBean = ArrayBean.construct(10_000, num);
 
-          soft.assertThat(
-                  get(server.getAddress(), disableCompression)
-                      .queryParam("len", Integer.toString(len))
-                      .queryParam("num", Integer.toString(num))
-                      .get()
-                      .readEntity(ArrayBean.class))
+          soft.assertThatCode(
+                  () ->
+                      assertThat(
+                              get(server.getAddress(), disableCompression)
+                                  .queryParam("len", Integer.toString(len))
+                                  .queryParam("num", Integer.toString(num))
+                                  .get()
+                                  .readEntity(ArrayBean.class))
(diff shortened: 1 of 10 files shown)


clients/client/src/test/java/org/projectnessie/client/util/TestServer.java
```diff
@@ -40,48 +48,53 @@ public TestServer(String context, HttpHandler handler) throws IOException {
    * @param init init method (optional)
    * @throws IOException maybe
```
XN137, line 49: nit: now outdated?

clients/client/src/test/java/org/projectnessie/client/http/TestHttpClient.java
```diff
@@ -100,16 +196,14 @@ void testReadTimeout() {
...
-          try (InputStream in = new GZIPInputStream(h.getRequestBody())) {
+    TestServer.RequestHandler handler =
+        (req, resp) -> {
+          Assertions.assertEquals("PUT", req.getMethod());
+          try (InputStream in = req.getInputStream()) {
             Object bean = MAPPER.readerFor(ExampleBean.class).readValue(in);
             Assertions.assertEquals(inputBean, bean);
           }
-          h.sendResponseHeaders(200, 0);
+          emptyResponse(resp);
```
XN137, line 206: imo this should be named `writeEmptyResponse`, as is this statement look odd
also for consistency with `writeResponseBody`

XN137 (approved):
nice, lgtm overall with minor comments

snazy:
Addressed the comments, but also renamed two `Test*` classes, that are actually not tests

XN137 (approved):

snazy:
Removed testReadTimeout(), it's hard to test actually, because it relies on the server's implementation and we cannot properly distinguish when the request actually started without freezing the test for many seconds and risking a flaky test.

XN137 (approved):

PR #5313, opened 2022-10-03
Run Native image and Integration test for Renovate PRs

dimas-b:
Sometimes bumping a lib version breaks the Native image.

Running Integration tests also adds assurance that auto-merged version bumps do not break Nessie in various environments.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -8,7 +8,11 @@
     ":preserveSemverRanges"
   ],
 
-  "labels": ["dependencies"],
+  "labels": [
+    "dependencies",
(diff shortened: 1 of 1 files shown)


dimas-b:
Related to this auto-merged version bump: https://github.com/projectnessie/nessie/commit/83e9365f59aa413338b7ef5a365b75d23797b8f0

snazy:
Keep in mind that only failed _required_ checks prevent merges.

dimas-b:
I guess I assumed that if you run the `native` build _and_ it fails, it will prevent merging... Apparently I was wrong... Closing.

PR #5314, opened 2022-10-03
Isolate http-level test from java client-level tests

dimas-b:
Some tests in AbstractRestInvalidWithHttp used to construct HTTP requests directly (by using an HttpClient). However, the majority of tests in the "AbstractRest" suite rely on the Java client implementation to perform actual HTTP requests.

Those tests are moved to AbstractResteasyTest for the sake of clarity and uniformity of test approach and dependencies.

The now unused getHttpClient() is removed from AbstractRest. The unused getUri() method is removed too.


ajantha-bhat (approved):
LGTM.

uri field was added in `AbstractRest` for extending these tests in `gc-base` module with separate hierarchy. But now that code removed. So, no need to expose uri now. 

And `HttpClient` was present from long time in this test. Thanks for cleaning it up.

snazy (approved):

PR #5339, opened 2022-10-06
Extract a multi-env test engine into a module.

dimas-b:
Add a custom sub-interface of JUnit5 `Extension` for communications between the new test engine and test extensions that actually manage test execution environments.

The new test engine is meant to be reused for API v2 tests.

However, this engine is fairly self-contained and it makes sense to keep it in a dedicated module.

These changes also fix the filter that was meant to exclude multi-version tests from the Jupiter test engine. It was not sufficient to exclude intermediate class nodes, leaf nodes have to be excluded too.


gradle/projects.main.properties
```diff
@@ -1,5 +1,6 @@
 nessie-bom=bom
 nessie-client=clients/client
+nessie-compatibility-engine=compatibility/test-engine
```
snazy, line 3: Wouldn't this fit better in `testing/` - and called `nessie-multi-env-engine`, incl. package renames?

compatibility/test-engine/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+
+plugins {
+  `java-library`
+  jacoco
+  `maven-publish`
+  signing
+  `nessie-conventions`
+}
+
+extra["maven.name"] = "Nessie - Backward Compatibility - Test Engine"
```
snazy, line 25: needs another name

snazy (commented):
(Not an full review yet)
Briefly looking at the change: looks good. Just think that the new module would probably fit better in the `testing/` top-level directory and get another module name

gradle/projects.main.properties
dimas-b: sgmt - will move

dimas-b: moved

compatibility/test-engine/build.gradle.kts
dimas-b: renamed

compatibility/test-engine/src/main/java/org/projectnessie/tools/compatibility/engine/MultiEnvTestFilter.java
```diff
@@ -0,0 +1,86 @@
...
+    return Optional.empty();
+  }
+
+  private FilterResult filter(Class<?> testClass, UniqueId id) {
+    // Use the static extension data collected during the discovery phase.
+    // It is possible to reload extensions based of class objects from test descriptors,
+    // however that would add unnecessary overhead.
+    MultiEnvExtensionRegistry registry = registry();
+
+    Boolean isJupiter = id.getEngineId().map("junit-jupiter"::equals).orElse(false);
```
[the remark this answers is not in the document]
dimas-b: inlined

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/AbstractMultiVersionExtension.java
```diff
@@ -46,10 +49,44 @@
...
 
   private static final ConditionEvaluationResult PASS = enabled(null);
 
+  @Override
+  public String segmentType() {
+    return NESSIE_VERSION_SEGMENT_TYPE;
+  }
+
+  @Override
+  public SortedSet<String> allEnvironmentIds(ConfigurationParameters configuration) {
```
snazy, line 64: Not sure whether `SortedSet<String>` is correct here. While the `Version`s are `Comparable`, the set of sorted `Version.toString()` might be different. Maybe use a `Set<String>` as the result type and use a `LinkedHashSet` instead of `TreeSet` as the concrete type?

gradle/projects.main.properties
```diff
@@ -20,6 +20,7 @@ nessie-quarkus-tests=servers/quarkus-tests
 nessie-rest-services=servers/rest-services
 nessie-s3minio=testing/s3minio
 nessie-s3mock=testing/s3mock
+nessie-multi-env-test-engine=testing/test-engine
```
snazy, line 23: Can you rename the directory to `multi-env-test-engine`? "test-engine" is a bit too generic IMO.

snazy (commented):
Generally LGTM - but a suggestion to rename the new proejct's directory.

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/AbstractMultiVersionExtension.java
dimas-b: My intention was to ensure a stable iteration over test cases between runs. I'll probably change to `List<String>` with a javadoc comment instead. Implementations can chose their own order.

gradle/projects.main.properties
dimas-b: renamed

snazy (approved):

PR #5345, opened 2022-10-08
Upgrade desert to 2022.9.22

VladimirYushkevich:
Replaced `desert.ib` by `attr.ib` to fix `tox` lint error message.

### DIFF
--- a/python/pynessie/model.py
+++ b/python/pynessie/model.py
@@ -276,9 +276,9 @@ class OperationsSchema(OneOfSchema):
 class CommitMeta:
     """Dataclass for commit metadata."""
 
-    hash_: str = desert.ib(fields.Str(data_key="hash"), default=None)
-    commitTime: datetime = desert.ib(fields.DateTime(), default=None)
-    authorTime: datetime = desert.ib(fields.DateTime(), default=None)
+    hash_: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(data_key="hash")))
+    commitTime: datetime = attr.ib(default=None, metadata=desert.metadata(fields.DateTime()))
+    authorTime: datetime = attr.ib(default=None, metadata=desert.metadata(fields.DateTime()))
     committer: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(allow_none=True)))
     author: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(allow_none=True)))
     signedOffBy: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(allow_none=True)))
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -17,7 +17,7 @@ attrs==22.1.0
 botocore==1.27.87
 Click==8.1.3
 confuse==2.0.0
-desert==2020.11.18
+desert==2022.9.22
 marshmallow==3.18.0
 marshmallow_oneofschema==3.0.1
 python-dateutil>=2.8.0


VladimirYushkevich:
Looks like current fix helped to pass `python 3.8` lint check, but I don't really like it. Maybe someone can double look and suggest better solution.

python/pynessie/model.py
```diff
@@ -276,9 +276,9 @@ def get_obj_type(self, obj: Operation) -> str:
 class CommitMeta:
     """Dataclass for commit metadata."""
 
-    hash_: str = desert.ib(fields.Str(data_key="hash"), default=None)
-    commitTime: datetime = desert.ib(fields.DateTime(), default=None)
-    authorTime: datetime = desert.ib(fields.DateTime(), default=None)
+    hash_: str = desert.ib(fields.Str(data_key="hash"), default="")
+    commitTime: datetime = desert.ib(fields.DateTime(), default=datetime.now())
+    authorTime: datetime = desert.ib(fields.DateTime(), default=datetime.now())
     committer: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(allow_none=True)))
```
XN137, line 282:
```
  pynessie/model.py:279: error: Incompatible types in assignment (expression has type "None", variable has type "str")
  pynessie/model.py:280: error: Incompatible types in assignment (expression has type "None", variable has type "datetime")
  pynessie/model.py:281: error: Incompatible types in assignment (expression has type "None", variable has type "datetime")
```
for the problematic fields, have we tried using the same pattern as we do here in line 282?
i.e. `attr.ib(default=None, metadata=xyz)` ?
according to https://desert.readthedocs.io/en/stable/usage.html#use-with-attrs this would just be the "verbose form"

side note: also kind of surprising that in line 286 we are not hitting this issue because the `dict` type should also not be able to hold `None` as default.

XN137: following my original suggestion and changing these 3 lines to:
```
    hash_: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(data_key="hash")))
    commitTime: datetime = attr.ib(default=None, metadata=desert.metadata(fields.DateTime()))
    authorTime: datetime = attr.ib(default=None, metadata=desert.metadata(fields.DateTime()))
```
passed both linting and tests locally, can you confirm?
what errors are you getting exactly?

VladimirYushkevich: This works for me as well. I still tried to use `desert.ib` which can be replaced safely by `attr.ib`.

python/pynessie/model.py
```diff
@@ -276,9 +276,9 @@ def get_obj_type(self, obj: Operation) -> str:
 class CommitMeta:
     """Dataclass for commit metadata."""
 
-    hash_: str = desert.ib(fields.Str(data_key="hash"), default=None)
-    commitTime: datetime = desert.ib(fields.DateTime(), default=None)
-    authorTime: datetime = desert.ib(fields.DateTime(), default=None)
+    hash_: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(data_key="hash")))
```
XN137, line 279: looking at this again wondering we also need `allow_none=True` as we do below? :thinking: 
if we have a test that uses this class without a `hash_` it is fine i guess.

XN137 (approved):
please adjust PR title (+description?) to be more precise before merging
i.e. `pynessie: upgrade desert to 2022.9.22` or the like

lgtm otherwise

python/pynessie/model.py
VladimirYushkevich: This works as well:
```
hash_: str = attr.ib(default=None, metadata=desert.metadata(fields.Str(data_key="hash", allow_none=True)))
```

P.S. Updated PR name and description.

PR #5346, opened 2022-10-09
Update camelcase to v7.x

VladimirYushkevich:
This branch is based on https://github.com/projectnessie/nessie/pull/5305.

In this PR:
* Fixed import for new version of `camelcase` by moving it from top level into method body. This allows to suppress ERR_REQUIRE_ESM error.

### DIFF
--- a/ui/config/jest/fileTransform.js
+++ b/ui/config/jest/fileTransform.js
@@ -16,7 +16,6 @@
 "use strict";
 
 const path = require("path");
-const camelcase = require("camelcase");
 
 // This is a custom Jest transformer turning file imports into filenames.
 // http://facebook.github.io/jest/docs/en/webpack.html
@@ -28,6 +27,7 @@ module.exports = {
     if (filename.match(/\.svg$/)) {
       // Based on how SVGR generates a component name:
       // https://github.com/smooth-code/svgr/blob/01b194cf967347d43d4cbe6b434404731b87cf27/packages/core/src/state.js#L6
+      const camelcase = require("camelcase");
       const pascalCaseFilename = camelcase(path.parse(filename).name, {
         pascalCase: true,
       });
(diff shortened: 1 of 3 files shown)


ajantha-bhat (approved):

XN137 (commented):
the release notes of v7 have:

> This package is now pure ESM. Please [read this](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

wondering: have we tried the suggested approach already?

i am no JS/node expert but i am guessing the suggested "fix" might not actually work, we just happen to not process `.svg` files as part of our build, so the import doesnt get executed now.

VladimirYushkevich:
> wondering: have we tried the suggested approach already?
I tried this: 
```
import { camelCase } from "camelcase";

 import { camelCase } from "camelcase";
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module
```
> i am no JS/node expert but i am guessing the suggested "fix" might not actually work, we just happen to not process `.svg` files as part of our build, so the import doesnt get executed now.

Exactly. Looks like the code in `fileTransform.js` is not executed now.

XN137:
just to be clear:
with "the suggested fix doesnt work" i was talking about this PR, not the changes suggested in the gist.

imo it doesnt make sense to upgrade this library and fix new problems by avoiding its usage...

unless we find an easy way to fix the import "correctly" or we are planning to migrate the project to [ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) completely (which might not be a good use of our time right now), we should imo just not bother upgrading the library for now...
after all its just for camel casing a string?

XN137:
i think closing the original PR will mark it as ignored/blocked in the dependency dashboard:
https://github.com/projectnessie/nessie/issues/5255
and we would need to click a checkbox to reactivate it

snazy:
I'm okay with closing (and not yet bothering about the dependency update).

Issue #4015, opened 2022-04-21
Let tests in `nessie-quarkus` use dynamic ports


snazy:
Currently the Quarkus tests in `nessie-quarkus` always use the fixed, default port number 19121.

Quarkus can use a dynamically allocated port, by setting `quarkus.http.test-port=0`.

Just changing the configuration however is not enough, because the default port number 19121 is hard coded in tests, so the tests need to be adjusted to use the dynamically allocated port number instead of 19121.

PR #5352, opened 2022-10-11
quarkus-server tests use dynamic port from env

XN137:
adds `AbstractTestQuarkusRest` as common base class to configure the nessie api client with quarkus port

fixes https://github.com/projectnessie/nessie/issues/4015

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
@@ -62,10 +62,6 @@ public interface HttpClient {
 
     private Builder() {}
 
-    public URI getBaseUri() {
-      return baseUri;
-    }
-
     public Builder setBaseUri(URI baseUri) {
       this.baseUri = baseUri;
       return this;
--- a/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
+++ b/clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
@@ -62,6 +62,10 @@ public interface HttpClient {
 
     private Builder() {}
 
+    public URI getBaseUri() {
+      return baseUri;
+    }
+
     public Builder setBaseUri(URI baseUri) {
       this.baseUri = baseUri;
       return this;
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
@@ -22,7 +22,6 @@ import java.net.URI;
 import java.util.Locale;
 import java.util.Objects;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.http.HttpClientBuilder;
 import org.projectnessie.error.BaseNessieClientServerException;
@@ -62,11 +61,6 @@ public abstract class AbstractRest {
 
   public NessieApiV1 getApi() {
     return Objects.requireNonNull(api, "Tests need to call initApi in @BeforeEach");
-  }
-
-  @BeforeEach
-  public void setUp() {
-    initApi(URI.create("http://localhost:19121/api/v1"));
   }
 
   @AfterEach
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -109,7 +109,7 @@ quarkus.log.category."io.quarkus.http.access-log".level=${HTTP_ACCESS_LOG_LEVEL:
 
 ## Quarkus http related settings
 quarkus.http.port=19120
-quarkus.http.test-port=19121
+quarkus.http.test-port=0
 quarkus.http.access-log.enabled=true
 # fixed at buildtime
 quarkus.resteasy.path=/api/v1
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
@@ -18,10 +18,13 @@ package org.projectnessie.server;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import io.quarkus.test.security.TestSecurity;
+import java.net.URI;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.function.Executable;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.http.HttpClientBuilder;
@@ -34,9 +37,15 @@ import org.projectnessie.model.IcebergTable;
 import org.projectnessie.model.Operation.Delete;
 import org.projectnessie.model.Operation.Put;
 
+@ExtendWith(QuarkusNessieUriResolver.class)
 abstract class AbstractTestBasicOperations {
 
   private NessieApiV1 api;
+
+  @BeforeEach
+  void setUp(URI quarkusNessieUri) {
+    api = HttpClientBuilder.builder().withUri(quarkusNessieUri).build(NessieApiV1.class);
+  }
 
   @AfterEach
   void closeClient() {
@@ -47,10 +56,6 @@ abstract class AbstractTestBasicOperations {
   }
 
   void getCatalog(String branch) throws BaseNessieClientServerException {
-    api =
-        HttpClientBuilder.builder()
-            .withUri("http://localhost:19121/api/v1")
-            .build(NessieApiV1.class);
     if (branch != null) {
       api.createReference().reference(Branch.of(branch, null)).sourceRefName("main").create();
     }
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestQuarkusRest.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestQuarkusRest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Dremio
+ * Copyright (C) 2022 Dremio
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,10 +15,20 @@
  */
 package org.projectnessie.server;
 
-import io.quarkus.test.junit.QuarkusIntegrationTest;
-import io.quarkus.test.junit.TestProfile;
-import org.projectnessie.quarkus.tests.profiles.QuarkusTestProfilePostgres;
+import java.net.URI;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.projectnessie.jaxrs.AbstractTestRest;
 
-@QuarkusIntegrationTest
-@TestProfile(QuarkusTestProfilePostgres.class)
-class ITRestApiPostgres extends AbstractTestQuarkusRest {}
+/**
+ * Tests need to subclass this class and use @QuarkusIntegrationTest or @QuarkusTest, so that the
+ * quarkus context is available to resolve the nessie URI.
+ */
+@ExtendWith(QuarkusNessieUriResolver.class)
+public abstract class AbstractTestQuarkusRest extends AbstractTestRest {
+
+  @BeforeEach
+  public void setUp(URI quarkusNessieUri) {
+    initApi(quarkusNessieUri);
+  }
+}
--- a/servers/quarkus-server/src/test/java/org/projectnessie/server/BaseClientAuthTest.java
+++ b/servers/quarkus-server/src/test/java/org/projectnessie/server/BaseClientAuthTest.java
@@ -15,16 +15,28 @@
  */
 package org.projectnessie.server;
 
+import com.google.common.base.Preconditions;
+import java.net.URI;
 import java.util.function.Consumer;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.http.HttpClientBuilder;
 
 /** Base class for client-base authentication and authorization tests. */
+@ExtendWith(QuarkusNessieUriResolver.class)
 public abstract class BaseClientAuthTest {
+
+  private URI quarkusNessieUri;
 
   private NessieApiV1 api;
   private Consumer<HttpClientBuilder> customizer;
+
+  @BeforeEach
+  void setUp(URI quarkusNessieUri) {
+    this.quarkusNessieUri = quarkusNessieUri;
+  }
 
   @AfterEach
   void closeClient() {
@@ -35,6 +47,7 @@ public abstract class BaseClientAuthTest {
   }
 
   protected void withClientCustomizer(Consumer<HttpClientBuilder> customizer) {
+    Preconditions.checkState(api == null, "withClientCustomizer but api has already been created!");
     this.customizer = customizer;
   }
 
@@ -43,8 +56,7 @@ public abstract class BaseClientAuthTest {
       return api;
     }
 
-    HttpClientBuilder builder =
-        HttpClientBuilder.builder().withUri("http://localhost:19121/api/v1");
+    HttpClientBuilder builder = HttpClientBuilder.builder().withUri(quarkusNessieUri);
 
     if (customizer != null) {
       customizer.accept(builder);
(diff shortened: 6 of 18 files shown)


servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/TestJerseyRestNaiveClientInMemory.java
```diff
@@ -48,25 +46,22 @@
...
         HttpClientBuilder.builder()
             // Abuse the authentication callback a bit to inject the noAcceptFilter into
             // the java client.
             .withAuthentication(
                 (HttpAuthentication) client -> client.addRequestFilter(noAcceptFilter))
-            .withUri(httpClient.getBaseUri())
+            .withUri(nessieApiUri)
             .build(NessieApiV1.class);
 
-    super.init(api, httpClient);
```
XN137, line 70: `super.init` did not use `httpClient`

servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/TestJerseyRestNaiveClientInMemory.java
```diff
@@ -48,25 +46,22 @@
...
-    httpClient.addRequestFilter(noAcceptFilter);
 
-    api =
+    NessieApiV1 api =
         HttpClientBuilder.builder()
             // Abuse the authentication callback a bit to inject the noAcceptFilter into
             // the java client.
             .withAuthentication(
                 (HttpAuthentication) client -> client.addRequestFilter(noAcceptFilter))
-            .withUri(httpClient.getBaseUri())
```
XN137, line 67: we only used the base uri from the passed in `httpClient`... so we should instead override `initApi(URI)`

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestQuarkusRest.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.server;
+
+import java.net.URI;
+import java.util.Objects;
+import org.junit.jupiter.api.BeforeEach;
+import org.projectnessie.jaxrs.AbstractTestRest;
+
+public class AbstractTestQuarkusRest extends AbstractTestRest {
```
XN137, line 23: this is the equivalent to `AbstractTestJerseyRest`, both customize how their category of tests should be initializing their api client in `AbstractRest`

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
```diff
@@ -55,26 +50,17 @@ public abstract class AbstractRest {
...
   }
 
-  protected void init(NessieApiV1 api, @Nullable HttpClient.Builder httpClient) {
+  protected void initApi(NessieApiV1 api) {
     this.api = api;
   }
 
-  @BeforeEach
-  public void setUp() {
-    init(URI.create("http://localhost:19121/api/v1"));
```
XN137, line 77: this hardcoding of 19121 is quarkus specific even though the `jax-rs-tests` module shouldnt be...
all tests from `quarkus-server` depended on this hardcoding in the base class, while `AbstractTestJerseyRest` avoided this hard-coding by overriding `setUp`.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -109,7 +109,7 @@ quarkus.log.category."io.quarkus.http.access-log".level=${HTTP_ACCESS_LOG_LEVEL:
 
 ## Quarkus http related settings
 quarkus.http.port=19120
-quarkus.http.test-port=19121
+quarkus.http.test-port=0
```
XN137, line 112: these are the remains of the old port number:
```
~/code/nessie$ rg 19121
clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/SparkSqlTestBase.java
56:  protected static final int NESSIE_PORT = Integer.getInteger("quarkus.http.test-port", 19121);

clients/deltalake/src/test/java/org/projectnessie/deltalake/AbstractDeltaTest.java
52:  private static final int NESSIE_PORT = Integer.getInteger("quarkus.http.test-port", 19121);

site/docs/develop/java.md
40:  .withUri(URI.create("http://localhost:19121/api/v1"))
215:  .withUri(URI.create("http://localhost:19121/api/v1"))
224:  .withUri(URI.create("http://localhost:19121/api/v1"))
```

note that spark/delta use the nessie-apprunner where `quarkus.http.test-port` gets injected:
https://github.com/projectnessie/nessie-apprunner/blob/main/apprunner-gradle-plugin/src/main/java/org/projectnessie/quarkus/gradle/QuarkusAppExtension.java

i did let the `quarkus-server` tests print their system properties and `test.url` was the only thing that was set (also when running from intellij)

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
```diff
@@ -55,26 +50,17 @@ public abstract class AbstractRest {
...
-    HttpClient.Builder httpClient = HttpClient.builder().setBaseUri(uri).setObjectMapper(mapper);
-    httpClient.addResponseFilter(new NessieHttpResponseFilter(mapper));
-
-    init(api, httpClient);
+  protected void initApi(URI nessieApiUri) {
+    NessieApiV1 api = HttpClientBuilder.builder().withUri(nessieApiUri).build(NessieApiV1.class);
+    initApi(api);
   }
 
-  protected void init(NessieApiV1 api, @Nullable HttpClient.Builder httpClient) {
```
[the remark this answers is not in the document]
ajantha-bhat: good catch. But let us keep it in separate PR for easier review.

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
```diff
@@ -47,10 +48,8 @@ void closeClient() {
   }
 
   void getCatalog(String branch) throws BaseNessieClientServerException {
-    api =
-        HttpClientBuilder.builder()
-            .withUri("http://localhost:19121/api/v1")
-            .build(NessieApiV1.class);
+    String quarkusUrl = Objects.requireNonNull(System.getProperty("test.url"));
```
ajantha-bhat, line 51: I didn't get how this property name (`"test.url"`) is chosen.

Because in application.properties, `quarkus.http.test-port` is the one we are changing to dynamic.

I thought some code is translating it. But I couldn't find it.

servers/quarkus-server/src/test/java/org/projectnessie/server/BaseClientAuthTest.java
```diff
@@ -43,8 +44,8 @@ protected NessieApiV1 api() {
       return api;
     }
 
-    HttpClientBuilder builder =
-        HttpClientBuilder.builder().withUri("http://localhost:19121/api/v1");
+    String quarkusUrl = Objects.requireNonNull(System.getProperty("test.url"));
+    HttpClientBuilder builder = HttpClientBuilder.builder().withUri(quarkusUrl + "/api/v1");
```
ajantha-bhat, line 48: nit: I know we can't extend this class with `AbstractTestQuarkusRest`, but is it possible for all these classes (that have these duplicate codes) to call API building from a unified place instead of duplicating the same code?

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
ajantha-bhat: I saw this comment now https://github.com/projectnessie/nessie/pull/5352#discussion_r992110450

But still not sure how it is translated.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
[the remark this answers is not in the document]
XN137: split out into https://github.com/projectnessie/nessie/pull/5354 will rebase shortly

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
XN137: i printed all system properties when running these tests in intellij and saw that `test.url` seemed to be the only one injected automatically

i think it is set here by quarkus:
https://github.com/quarkusio/quarkus/blob/c7c6eab48e71ea782f98f0884ad624981cb1b802/test-framework/common/src/main/java/io/quarkus/test/common/LauncherUtil.java#L188-L196

according to my understanding of https://quarkus.io/guides/getting-started-testing#controlling-the-test-port
you set the `quarkus.http.test-port` in the quarkus application properties but it does not translate directly to a system property?
all the examples from that page use various quarkus annotations/injected resources etc... i assume internally they are reading stuff directly from the quarkus config (which is not available as system property?)

ajantha-bhat: when I was debugging for Sagar's PR, I remember seeing dynamic port value in `System.getProperty("quarkus.http.test-port")`. Hence, it triggered my doubts.

XN137: i re-checked and yeah it is there:
```
test.url -> http://localhost:38269
quarkus.http.test-port -> 38269
%test.quarkus.http.port -> 38269
quarkus.http.port -> 38269
```
not sure how i missed it originally... however `test.url` might still be the right thing to use?

servers/quarkus-server/src/test/java/org/projectnessie/server/BaseClientAuthTest.java
XN137: added a utils class

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestBasicOperations.java
XN137: i dont think it matters too much but ok will use that instead (for consistency with the other classes i.e. delta + spark)

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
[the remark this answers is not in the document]
dimas-b: Yes, I meant to do a similar cleanup wrt `HttpClient` in tests - thanks for taking care of this :)

servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
```diff
@@ -50,12 +51,12 @@
     QuarkusTestProfileInmemory.class) // use the QuarkusTestProfileInmemory, as it can be reused
 class TestNessieError {
 
-  static String baseURI = "http://localhost:19121/api/v1/nessieErrorTest";
-
   private static HttpClient client;
 
   @BeforeAll
   static void setup() {
+    String baseURI = QuarkusTestUtils.getNessieApiV1Url() + "/nessieErrorTest";
```
dimas-b, line 58: Technically the `nessieErrorTest` resource has nothing to do with Nessie API versions, it's a standalone service. Could we have a "get base URL" method of something like that?

servers/quarkus-server/src/test/java/org/projectnessie/server/QuarkusTestUtils.java
```diff
@@ -0,0 +1,33 @@
...
+  private QuarkusTestUtils() {
+    // utillity class
+  }
+
+  public static Integer getQuarkusTestPort() {
+    return Objects.requireNonNull(Integer.getInteger("quarkus.http.test-port"));
+  }
+
+  public static String getNessieApiV1Url() {
+    return String.format("http://localhost:%d/api/v1", getQuarkusTestPort());
```
dimas-b, line 31: I tend to think that it might be nicer to extract the Nessie URL discovery logic into a JUnit5 extension.... similar to how `@NessieUri` parameters are handled.

dimas-b: Basically, we could add something like a `NessieQuarkusTestExtension` to resolve `@NessieUri` parameters based on quarkus-managed system properties instead of `QuarkusTestUtils`.

`AbstractTestQuarkusRest` would declare `@ExtendWith(NessieQuarkusTestExtension.class)` and get the URL injected in a manner similar to `AbstractTestJerseyRest`. WDYT?

dimas-b: It it works well, we could even have a `@BeforeEach initApi(@NessieUri uri)` in `AbstractRest` and drop custom calls to `initApi(client)` overrides (except for `TestJerseyRestNaiveClientInMemory`)

XN137: i've added an extension that will resolve `URI quarkusNessieUri` parameters...

using `@NessieUri` would have been more difficult because it exists in a different module (jaxrs extension which is not realted to quarkus afaict)...
generally i think if we need to support different types like api versions we should overhaul the extensions when this api classes get introduced since then we have the full context and are not playing a guessing game, WDYT?

servers/quarkus-server/src/test/java/org/projectnessie/server/AbstractTestQuarkusRest.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server;
+
+import java.net.URI;
+import org.junit.jupiter.api.BeforeEach;
+import org.projectnessie.jaxrs.AbstractTestRest;
+
+public abstract class AbstractTestQuarkusRest extends AbstractTestRest {
```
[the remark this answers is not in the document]
XN137: actually it is not possible (see CI errors: https://github.com/projectnessie/nessie/actions/runs/3233984593/jobs/5296535131)
because sub-classes like `TestRestApiInMemory` use plain `@QuarkusTest`

servers/quarkus-server/src/test/java/org/projectnessie/server/error/TestNessieError.java
dimas-b: Well, my comment was just about the name of the method used for getting the URL :)

servers/quarkus-server/src/test/java/org/projectnessie/server/QuarkusTestUtils.java
dimas-b: Yes, extensions will have to be adjucted, of course

clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
```diff
@@ -62,10 +62,6 @@ class Builder {
 
     private Builder() {}
 
-    public URI getBaseUri() {
```
dimas-b, line 65: This may have dependencies outside of Nessie. I do not think it's justified to remove this in a test refactoring PR :thinking:

servers/quarkus-server/src/test/java/org/projectnessie/server/QuarkusNessieUriResolver.java
```diff
@@ -0,0 +1,60 @@
...
+        Integer.getInteger("quarkus.http.test-port"),
+        "System property not set correctly: quarkus.http.test-port");
+  }
+
+  private static String getNessieApiV1Url() {
+    return String.format("http://localhost:%d/api/v1", getQuarkusTestPort());
+  }
+
+  private boolean isQuarkusNessieUriParam(ParameterContext paramCtx) {
+    return paramCtx.getParameter().getName().equals("quarkusNessieUri")
```
dimas-b, line 38: Checking param names seems a bit too cryptic to me. I'd prefer an annotation, but chances are we'll have to refactor this for API v1/v2 tests, so I think current approach is ok.

servers/quarkus-server/src/test/java/org/projectnessie/server/QuarkusNessieUriResolver.java
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  @Override
+  public Object resolveParameter(ParameterContext paramCtx, ExtensionContext extensionCtx)
+      throws ParameterResolutionException {
+    if (isQuarkusNessieUriParam(paramCtx)) {
+      return URI.create(getNessieApiV1Url());
+    }
+    throw new ParameterResolutionException(
+        "Unsupported annotation on parameter "
```
dimas-b, line 55: Why `Unsupported annotation`? Annotations are not processed by this extension at all :thinking:

dimas-b (commented):
LGTM overall. Added some nit-picking comments

clients/client/src/main/java/org/projectnessie/client/http/HttpClient.java
ajantha-bhat: As it is a public interface and at least we should deprecate it before directly removing it.

Also, better to deprecate in separate PR and revert the change here!

dimas-b: TBH, I'm not sure we should deprecate this :) Why not keep it as a convenience method?

XN137: i think we removed the only user in https://github.com/projectnessie/nessie/pull/5354 but OK i have added it back

servers/quarkus-server/src/test/java/org/projectnessie/server/QuarkusNessieUriResolver.java
XN137: was copy paste error from other extension, have adjusted the message

dimas-b (approved):

PR #5354, opened 2022-10-11
Remove httpClient param from AbstractRest.init

XN137:
- the httpClient builder was effectively unused after https://github.com/projectnessie/nessie/pull/5314
- also rename to initApi

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
@@ -17,18 +17,14 @@ package org.projectnessie.jaxrs;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import com.google.common.collect.ImmutableMap;
 import java.net.URI;
 import java.util.Locale;
-import javax.annotation.Nullable;
+import java.util.Objects;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.projectnessie.client.api.NessieApiV1;
-import org.projectnessie.client.http.HttpClient;
 import org.projectnessie.client.http.HttpClientBuilder;
-import org.projectnessie.client.rest.NessieHttpResponseFilter;
 import org.projectnessie.error.BaseNessieClientServerException;
 import org.projectnessie.error.NessieConflictException;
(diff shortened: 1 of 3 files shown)


servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/TestJerseyRestNaiveClientInMemory.java
```diff
@@ -48,25 +46,22 @@
...
 
-    api =
+    NessieApiV1 api =
         HttpClientBuilder.builder()
             // Abuse the authentication callback a bit to inject the noAcceptFilter into
             // the java client.
             .withAuthentication(
                 (HttpAuthentication) client -> client.addRequestFilter(noAcceptFilter))
-            .withUri(httpClient.getBaseUri())
+            .withUri(nessieApiUri)
```
XN137, line 62: we only used the base uri from the passed in httpClient... so we should instead override initApi(URI)

servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/TestJerseyRestNaiveClientInMemory.java
```diff
@@ -48,25 +46,22 @@
...
             .build(NessieApiV1.class);
 
-    super.init(api, httpClient);
+    super.initApi(api);
```
XN137, line 65: `super.init` did not use httpClient

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/AbstractRest.java
```diff
@@ -55,26 +51,22 @@ public abstract class AbstractRest {
...
-    HttpClient.Builder httpClient = HttpClient.builder().setBaseUri(uri).setObjectMapper(mapper);
-    httpClient.addResponseFilter(new NessieHttpResponseFilter(mapper));
-
-    init(api, httpClient);
+  protected void initApi(URI nessieApiUri) {
+    NessieApiV1 api = HttpClientBuilder.builder().withUri(nessieApiUri).build(NessieApiV1.class);
+    initApi(api);
   }
 
-  protected void init(NessieApiV1 api, @Nullable HttpClient.Builder httpClient) {
```
XN137, line 71: `httpClient` param is ignored here (it used to be stored into a field previously)

ajantha-bhat (approved):

dimas-b (approved):
I'm thinking to move the Nessie client construction code into a JUnit5 extension and make it automatic for v1 and v2 API tests (based on #5339), but for now this looks like a reasonable refactoring. Thanks for cleaning this up :)

PR #5377, opened 2022-10-17
Automatically update release text files

snazy:
Automatically update the README.md + SECURITY.md + server-upgrade.md files, replacing the previously released version with the currently released version.


.github/workflows/release-create.yml
```diff
@@ -103,6 +103,17 @@ jobs:
...
         !
 
+    - name: Update README.md + SECURITY.md + server-upgrade.md files
+      run: |
+        LAST_VERSION="${LAST_TAG#nessie-}"
+        LAST_VERSION_PATTERN="$(echo ${LAST_VERSION} | sed 's/\([^.]*\)[.]\([^.]*\)[.].*/\1\\.\2/')\\."
+
+        for fn in SECURITY.md README.md site/docs/try/server-upgrade.md ; do
+          cp ${fn} /tmp/temp-release-file.md
+          sed "s/${LAST_VERSION_PATTERN}[^ ]*/${RELEASE_VERSION}/" < /tmp/temp-release-file.md > ${fn}
```
XN137, line 113: cant we use `sed -i` here for inplace updates?

.github/workflows/release-create.yml
```diff
@@ -103,6 +103,17 @@ jobs:
         | Before release Git HEAD    | \`$(git rev-parse HEAD)\` |
         !
 
+    - name: Update README.md + SECURITY.md + server-upgrade.md files
```
XN137, line 106: nit: imo the name is bound to get outdated with the script so i would use a more generic one maybe

XN137 (dismissed):
i checked the files and it seems like the script would do the right thing, so +1 but maybe it will be fragile in the future

XN137 (approved):
thanks for the quick update

PR #5385, opened 2022-10-18
GH workflows: Add missing `cache-read-only: true`

snazy:
Gradle action should not "pollute" GitHub's build cache with PR builds and builds from release workflows.

### DIFF
--- a/.github/workflows/pull-request-integ.yml
+++ b/.github/workflows/pull-request-integ.yml
@@ -88,6 +88,7 @@ jobs:
       - name: Iceberg Nessie test
         uses: gradle/gradle-build-action@v2
         with:
+          cache-read-only: true
           arguments: :iceberg:iceberg-nessie:test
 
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request-integ.yml
```diff
@@ -88,6 +88,7 @@ jobs:
       - name: Iceberg Nessie test
         uses: gradle/gradle-build-action@v2
         with:
+          cache-read-only: true
```
XN137, line 91: > By default, the gradle-build-action will only write to the cache from Jobs on the default (main/master) branch. 

https://github.com/gradle/gradle-build-action#using-the-caches-read-only

so isnt it redundant to do this on PR workflows?

XN137 (approved):

.github/workflows/pull-request-integ.yml
snazy: ah, right - always forget about that.

PR #5392, opened 2022-10-18
Fix version in the Nessie Helm Chart

dimas-b:
Use the latest actually released version.


snazy:
Hm - looks like `.github/actions/bump-versions/action.yml` needs a flag whether the helm-chart's version should be bumped or not (default to not bump) and `.github/workflows/release-create.yml` set that flag for the release version.

snazy (approved):
LGTM! THanks for the WF change!

PR #5408, opened 2022-10-24
Build UI against Nessie API v1 from 0.44.0

dimas-b:
This is a temporary pin in preparation for API V2.

Newer builds will not publish OpenAPI YAML, which is why it is
necessary to refer to an older Nessie version in the UI build script.

The build uses a checked-in copy of Nessie 0.44.0 OpenAPI YAML file
at ui/src/openapi/

UI is to be updated to the upcoming V2 Nessie API eventually.

### DIFF
--- a/ui/build.gradle.kts
+++ b/ui/build.gradle.kts
@@ -32,10 +32,6 @@ plugins {
   `nessie-conventions`
 }
 
-val openapi by configurations.creating { description = "Used to reference OpenAPI spec files" }
-
-dependencies { openapi(project(":nessie-model", "openapi")) }
-
 node {
   download.set(true)
   version.set("16.14.2")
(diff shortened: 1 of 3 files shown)


XN137 (approved):
lgtm since it is a temporary thing and i assume you tried out the UI locally.

do we have a rough timeline when we could be updating the UI to v2 api?
we should file a ticket to not forget about it

dimas-b:
I've reworked this PR to use a check-in copy of Nessie API v1 from 0.44.0 to avoid ivy repository workarounds at build time.

The API v1 YAML is not expected to evolve.

dimas-b:
And, yes, I did manually verify that the UI still works :)

ui/src/openapi/openapi.yaml
```diff
@@ -0,0 +1,2324 @@
+---
+# A copy of OpenAPI definition of Nessie 0.44.0, a.k.a. API v1
```
XN137, line 2: shouldnt the filename reflect this i.e. `openapi-nessie044.yaml` or the like?
imo then its clearer this is a temporary thing potentially, not the "leading" openapi file

XN137 (dismissed):

ui/src/openapi/openapi.yaml
dimas-b: renamed

XN137 (approved):

PR #5412, opened 2022-10-25
Extract service-side interfaces for client-facing services

dimas-b:
This is in preparation for API v2.

Move `MergeBehavior` and `MergeKeyBehavior` to the top level.

The extracted interfaces (in the `services.spi` package) are intended to be reused between API v1 and v2 REST endpoints.

Parameter holders are expanded into list of individual parameter values to simplify reuse. Most of those parameter classes are going to be different in API v2.

Return values types are preserved. V2 changes to `model` classes are intended to be backward-compatible.


snazy (approved):
LGTM - can you check whether there are backwards-compatibility tests exercising these? Maybe it's good enough to add the moved types to the existing test. I think, it shouldn't break, because a) the compat-tests re-serialize using JSON and b) we do not have any clients that depend on these types (without re-building their code).

dimas-b:
@snazy : If you mean `MergeBehavior` - it is not (yet) exposed in the java client. Only plain REST clients can use it. We could probably add test to the "Reseasy" suite.

dimas-b:
I'll add tests in a follow-up PR. I believe existing `ITOlderClients` would have actually broken on this move of `MergeBehavior` were it exposed to java clients. Luckily, it is not exposed, so the move should not affect any clients.

dimas-b:
Follow-up issue #5415

PR #5419, opened 2022-10-27
Move old API classes into the api.v1 package

dimas-b:
This is in preparation for API v2, which is going to have parallel set of HTTP interfaces and parameter classes using different annotations.

Classes used by java clients are not affected. In this respect, `FetchOption` could have been moved, but that would have broken binary compatibility with older java clients. In fact `FetchOption` probably belongs under the `model` package rather than the `params` package, but it is already exposed to external clients, which is why it remains in it old stop in this PR.

The moved classes are not expected to be referenced directly by external code.

---

Opening as "draft" to avoid merging it accidentally, but the changes should be reviewable. There will be a series of additional PRs that build on top of this one.


clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpUpdateNamespace.java
```diff
@@ -15,9 +15,9 @@
  */
 package org.projectnessie.client.http.v1api;
```
XN137, line 16: is this naming pattern contradicting `org.projectnessie.api.v1.http` to some degree?

XN137 (commented):
generally seems good, assuming we have confirmed that we have tests covering that old clients dont break.

wondering (pardon my ignorance :sweat_smile:): would it not have been an alternative to keep things where they are and add `org.projectnessie.apiv2` package (or the like) until the old one can go away?
is it just because `FetchOption` will be used in both api versions or are there other benefits from moving?

clients/client/src/main/java/org/projectnessie/client/http/v1api/HttpUpdateNamespace.java
dimas-b: It is, but I did not want to rename unnecessary packages :) In this PR I just wanted to make room for adding `api.v2`.

I can do the client renaming in a separate PR, if you think it might be worth the effort.

dimas-b:
> assuming we have confirmed that we have tests covering that old clients don't break.

Integration test cover that.

Plus I manually verified that `TreeApi` and similar interfaces are not used by clients under Nessie's control. Of course, there's risk that some external code refers to them, but I do not think we have to guarantee source-level backward compatibility at this stage.

snazy (approved):

dimas-b:
rebased

snazy (approved):

PR #5420, opened 2022-10-27
Introduce multi-version API test framework

dimas-b:
Currently only V1 is usable, but V2 can be added seamlessly (hopefully). API version contexts in tests are managed by MultiVersionApiTest.

Copy REST path annotations to implementation classes with version- specific path values. This allows generating path in the OpenAPI YAML using only parth that follow the `/v1` segment, while allowing Quarkus servers to define the base URI one level higher. The latter will be handy when `/v2` paths are introduced.

Legacy tests (including external code) get v1 Nessie URIs injected for backward compatibility.

Nessie tests get the latest API URI injected by default. In-memory tests will get all API versions tests. This should ensure coverage for all API versions, while not wasting CI resources by duplicating tests using persistent storage. API version differences are handled before requests reach storage.


clients/client-testextension/build.gradle.kts
```diff
@@ -29,5 +29,7 @@ dependencies {
   api(libs.junit.jupiter.api)
 
   implementation(project(":nessie-client"))
+  api(project(":nessie-multi-env-test-engine"))
```
snazy, line 32: Isn't `Implementation` enough?

clients/client-testextension/build.gradle.kts
```diff
@@ -29,5 +29,7 @@ dependencies {
...
+  api(libs.junit.jupiter.engine)
```
snazy, line 33: Isn't `Implementation` enough?

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/CurrentNessieServer.java
```diff
@@ -54,7 +54,7 @@ static CurrentNessieServer currentNessieServer(
 
   @Override
   public URI getUri() {
-    return jersey.getUri();
+    return jersey.getUri().resolve("v1");
```
snazy, line 57: Don't we need something similar (in the future?) for the old servers? And have a "mapping" which Nessie versions support with API versions?

servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
```diff
@@ -149,7 +149,9 @@ public Object resolveParameter(
 
     if (parameterContext.isAnnotated(NessieUri.class)) {
       // Backward compatibility with older (external) tests
-      return getBaseUri(extensionContext);
+      // Inject v1 URIs in this case. Version-aware test should use
+      // `org.projectnessie.client.ext.NessieUri`.
```
snazy, line 153: Should `org.projectnessie.client.ext.NessieUri` be renamed then to avoid ambiguouty?

snazy (commented):
Generally LGTM - it's a test thing, so can be changed at any time.
Just thinking that having two `NessieUri` annotations is probably not good - better rename the new one to maybe `NessieClientUri` or so to avoid issues.

clients/client-testextension/build.gradle.kts
dimas-b: I'll double check, but IIRC transitive dependencies are required for `javac` in downstream modules.

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/CurrentNessieServer.java
dimas-b: Yeah, I have not thought in details about the older clients / old servers tests yet, but I suppose we'd use `v1` until the switch over version, then `v2` (`v1` tests in current servers always run during normal CI). Older clients would always use `v1` when talking even to new servers.

servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
dimas-b: will do

clients/client-testextension/build.gradle.kts
dimas-b: I stand corrected. `implementation` is sufficient here.

I guess `api` was a remnant of earlier experiments when test code referenced `MultiVersionApiTest` directly.

clients/client-testextension/build.gradle.kts
dimas-b: fixed

servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
dimas-b: renamed

snazy (approved):

PR #5435, opened 2022-11-03
Move Merge/Transplant classes to the api.v1.params package

dimas-b:
These classes are not used in user-facing java client methods.

They are used only as parameters to the REST API interfaces, so it
makes sense to keep them under the params package. This will also
improve isolation between API v1 and v2.

This change is expected to be backward-compatible to existing java
and REST API clients.

----

This PR builds on top of #5419 but unlike #5419 is contains some non-trivial changes, so I thought having a separate PR for it could be helpful for reviews.


dimas-b:
rebased

snazy (approved):

PR #5445, opened 2022-11-07
Move some tests from db-adapter test code to version-store test code

snazy:
This change only moves tests from database-adapter tests to version-store tests. No functional change.


XN137 (approved):
nit: commit msg / PR title contains a typo: `verstion-store`

can you maybe also add why we are moving things?
is it because in `AbstractDatabaseAdapterVersionStoreTest` the only new testing dimension is the tracing tests?

dimas-b (approved):

PR #5480, opened 2022-11-15
python: fix lint with new mypy version

XN137:
mypy was pulled in through pytest-mypy without version constraints, thus we got a hidden upgrade causing linting to fail, due to: https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

-> No Implicit Optional Types for Arguments

we explicitly pin mypy now and fix all related violations

### DIFF
--- a/python/pynessie/model.py
+++ b/python/pynessie/model.py
@@ -68,7 +68,7 @@ class Content:
 
     def pretty_print(self) -> str:
         """Print out for cli."""
-        pass
+        raise NotImplementedError
 
 
 @attr.dataclass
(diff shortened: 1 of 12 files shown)


python/pynessie/model.py
```diff
@@ -68,7 +68,7 @@ def requires_expected_state() -> bool:
 
     def pretty_print(self) -> str:
         """Print out for cli."""
-        pass
+        raise NotImplementedError
```
XN137, line 71: see `Safe Handling of Empty Function Bodies`
https://mypy-lang.blogspot.com/2022/11/mypy-0990-released.html

snazy (approved):

PR #5482, opened 2022-11-15
Suppress a few checkstyle warnings

snazy:


gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/MarkOptions.java
```diff
@@ -36,6 +36,7 @@
 import picocli.CommandLine.ParameterException;
 
 public class MarkOptions {
+  @SuppressWarnings("JavaTimeDefaultTimeZone")
```
XN137, line 39: nit: maybe add a comment to explain that we are intentionally using the CLI users timezone here (and it will not get mixed with server timezone later on)

gc/gc-tool/src/test/java/org/projectnessie/gc/tool/cli/commands/TestMarkOptions.java
```diff
@@ -148,6 +148,7 @@ public void cutoffPoliciesPerRef() {
         .isEqualTo(CutoffPolicy.numCommits(1));
   }
 
+  @SuppressWarnings("TypeParameterUnusedInFormals")
```
XN137, line 151: these are errorprone (not checkstyle) warnings:
http://errorprone.info/bugpattern/TypeParameterUnusedInFormals

wondering: couldnt we have fixed this as suggested?

snazy: err - yes

versioned/persist/tests/src/main/java/org/projectnessie/versioned/persist/tests/AbstractManyKeys.java
```diff
@@ -537,7 +537,8 @@ public static Stream<List<String>> progressivelyManyKeyNames() {
         .mapToObj(
             i ->
                 IntStream.range(1, i)
-                    .mapToObj(j -> Long.toString(Math.abs(random.nextLong()), Character.MAX_RADIX))
+                    .mapToObj(
+                        j -> Long.toString(random.nextLong() & Long.MAX_VALUE, Character.MAX_RADIX))
```
dimas-b, line 541: Ideally, we need to choose the seed (line 533) again in this case. The changed names are not guaranteed to expose the bug, whose handling that the test is meant to validate.

XN137: sorry i missed reading the context of this change (that we were testing an edge condition with explicitly crafted test data)

snazy: Yea - let's have a separate issue for that.

dimas-b: well, to be fair, the comment was hidden outside of the GH "changed lines" window :)

dimas-b: I ran a quick test with the new name generator and old code base: compared to the old generator the number of failed test cases decreased from 40 to 33 (out of 200). I think the test data is still good enough. I do not think we need to change the seed.

PR #5483, opened 2022-11-15
Move tests from o.p.jaxrs to o.p.jaxrs.tests

snazy:
... and rename two base classes to indicate those use the DatabaseAdapter.

### DIFF
--- a/servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/tests/AbstractTestDatabaseAdapterRest.java
+++ b/servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/tests/AbstractTestDatabaseAdapterRest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Dremio
+ * Copyright (C) 2022 Dremio
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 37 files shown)


dimas-b (approved):

servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/tests/AbstractTestDatabaseAdapterRest.java
```diff
@@ -24,7 +24,7 @@
 
 /** See {@link AbstractTestRest} for details about and reason for the inheritance model. */
 @ExtendWith(DatabaseAdapterExtension.class)
-abstract class AbstractTestJerseyRest extends AbstractRestSecurityContext {
+abstract class AbstractTestDatabaseAdapterRest extends AbstractRestSecurityContext {
```
XN137, line 27: imo both old and new name read a bit odd in that `Test` is somewhere in the middle... `Test` prefix or suffix would be more obvious

servers/jax-rs-tests/src/test/java/org/projectnessie/jaxrs/tests/TestJerseyResteasyH2.java
```diff
@@ -22,4 +22,4 @@
 
 @NessieDbAdapterName(H2DatabaseAdapterFactory.NAME)
 @NessieExternalDatabase(H2TestConnectionProviderSource.class)
-class TestJerseyResteasyH2 extends AbstractTestJerseyResteasy {}
+class TestJerseyResteasyH2 extends AbstractTestDatabaseAdapterResteasy {}
```
XN137, line 25: why is the `Jersey`-part still relevant for this concrete subclass but no longer for the base class?
in this file itself there seems to be nothing Jersey-specific

XN137 (approved):
+1 i guess but some naming still seems inconsistent imo

PR #5486, opened 2022-11-15
Refactor Nessie-Jax-RS extension a little bit

snazy:
Uses a "factory method" instead of a constructor, making it less dependent on the "database adapter" code.

### DIFF
--- a/gc/gc-base/src/test/java/org/projectnessie/gc/identify/TestIdentifyLiveContents.java
+++ b/gc/gc-base/src/test/java/org/projectnessie/gc/identify/TestIdentifyLiveContents.java
@@ -16,6 +16,7 @@
 package org.projectnessie.gc.identify;
 
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.projectnessie.jaxrs.ext.NessieJaxRsExtension.jaxRsExtensionForDatabaseAdapter;
 
 import java.time.Instant;
 import java.util.Set;
@@ -31,7 +32,6 @@ import org.projectnessie.gc.contents.LiveContentSetsRepository;
 import org.projectnessie.gc.contents.inmem.InMemoryPersistenceSpi;
 import org.projectnessie.gc.repository.NessieRepositoryConnector;
 import org.projectnessie.gc.repository.RepositoryConnector;
-import org.projectnessie.jaxrs.ext.NessieJaxRsExtension;
 import org.projectnessie.model.Content;
 import org.projectnessie.model.IcebergTable;
 import org.projectnessie.versioned.persist.adapter.DatabaseAdapter;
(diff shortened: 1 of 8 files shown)


gc/gc-base/src/test/java/org/projectnessie/gc/identify/TestIdentifyLiveContents.java
```diff
@@ -51,7 +51,7 @@ public class TestIdentifyLiveContents {
 
   @RegisterExtension
   static org.projectnessie.jaxrs.ext.NessieJaxRsExtension server =
```
dimas-b, line 53: nit: why not keep the import and use the simple name for this type?

gc/gc-base/src/test/java/org/projectnessie/gc/repository/TestNessieRepositoryConnector.java
```diff
@@ -39,7 +40,6 @@
 import org.projectnessie.client.ext.NessieClientFactory;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
-import org.projectnessie.jaxrs.ext.NessieJaxRsExtension;
```
dimas-b, line 42: nit: same here: why not keep the import?

dimas-b (approved):
nit: Just an idea - another way is to share data between `DatabaseAdapterExtension` and `NessieJaxRsExtension` via the JUnit5 `Store` using a well-known constant key. This way we would not have to declare a variable for `NessieJaxRsExtension`, but could add it to `ExtendWith`.

servers/jax-rs-testextension/src/main/java/org/projectnessie/jaxrs/ext/NessieJaxRsExtension.java
```diff
@@ -82,10 +83,15 @@ public NessieJaxRsExtension() {
     throw new UnsupportedOperationException();
   }
 
-  public NessieJaxRsExtension(Supplier<DatabaseAdapter> databaseAdapterSupplier) {
+  private NessieJaxRsExtension(Supplier<DatabaseAdapter> databaseAdapterSupplier) {
```
dimas-b, line 86: This might break Iceberg Nessie tests during NesQuEIT runw... WDYT?

gc/gc-base/src/test/java/org/projectnessie/gc/repository/TestNessieRepositoryConnector.java
dimas-b: to use a simple class name on line 70... same in `TestIdentifyLiveContents.java`

snazy: It's in the static-import - that's why the method name's that verbose.

dimas-b: I mean the `sever` variable type.

snazy: oh - no clue why IJ changed that

dimas-b (approved):

PR #5517, opened 2022-11-17
Docs: Update Spark Python/Java API docs

### LINKED ISSUE #5462: Update the document in https://projectnessie.org/tools/iceberg/spark/#writing
https://projectnessie.org/tools/iceberg/spark/#writing

looks like java and python code uses 'url' keyword instead of 'uri' keyword for Nessie catalog properties. 
Need to recheck these examples with the latest versions and update the docs.

ajantha-bhat:
`url` is not a catalog property, it is `uri`. 

Fixes #5462

### DIFF
--- a/site/docs/tools/iceberg/spark.md
+++ b/site/docs/tools/iceberg/spark.md
@@ -167,9 +167,10 @@ addressing jvm objects directly in Python (until the python library for iceberg 
     // first instantiate the catalog
     NessieCatalog catalog = new NessieCatalog();
     catalog.setConf(sc.hadoopConfiguration());
+    // other catalog properties can be added based on the requirement. For example, "io-impl","authentication.type", etc.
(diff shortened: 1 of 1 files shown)


ajantha-bhat:
@keithgchapman: As you were recently looking into it, let me know if anything else need an update for python API example.

site/docs/tools/iceberg/spark.md
```diff
@@ -204,8 +205,9 @@ addressing jvm objects directly in Python (until the python library for iceberg
     # first instantiate the catalog
```
ajantha-bhat, line 205: Spark-3.x currently says python code is the same as spark-2.x and doesn't have an example for that. 

I have the plan to drop spark-2.x mention in the document in a follow-up PR. That time I will move this example under spark-3.x

dimas-b (approved):

PR #5540, opened 2022-11-21
Nessie CLI: rename `call()` to `callWithDatabaseAdapter()`

snazy:
+ nit: print erase-repository notification

### DIFF
--- a/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/BaseCommand.java
+++ b/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/BaseCommand.java
@@ -16,6 +16,7 @@
 package org.projectnessie.quarkus.cli;
 
 import java.util.concurrent.Callable;
+import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
 import org.projectnessie.quarkus.config.VersionStoreConfig;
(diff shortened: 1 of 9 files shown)


servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/BaseCommand.java
```diff
@@ -29,8 +30,29 @@ public abstract class BaseCommand implements Callable<Integer> {
   @Inject VersionStoreConfig versionStoreConfig;
   @Inject ServerConfig serverConfig;
 
+  @Inject Instance<DatabaseAdapter> databaseAdapterInstance;
```
dimas-b, line 33: Should `@Inject` on line 29 be removed?

snazy: Nope, that's the "quarkus specific" one

dimas-b: ... but we overwrite it from `databaseAdapterInstance.get()` in `call()` :thinking: 

What is the difference between these two adapter objects?

snazy: err, wrong line 29

dimas-b (approved):

PR #5542, opened 2022-11-21
Nessie CLI: Update workding for in-memory warning

snazy:


dimas-b (approved):
nit: typo in commit message: `workding` (extra `k`)

PR #5551, opened 2022-11-23
Add RefreshContent command

dimas-b:
The `content-refresh` command of the Content Generator tool can be to reinsert old content objects without modification to cause them to be stored in the most up-to-date and efficient way.

This command is mostly helpful with migrating old "global state" contents to "on reference state".


snazy:
Isn't export+import "good enough"?

dimas-b:
export+import requires downtime, `content-refresh` does not.

snazy (approved):

PR #5553, opened 2022-11-23
Remove dependency to database-adapter type from `ImportResult`

snazy:


versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ImportResult.java
```diff
@@ -18,6 +18,7 @@
 import org.immutables.value.Value;
 import org.projectnessie.versioned.persist.adapter.HeadsAndForkPoints;
```
dimas-b, line 19: Should we remove this `import` too?

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ImportResult.java
```diff
@@ -32,9 +33,9 @@ public interface ImportResult {
    * optimization after a repository import.
    *
    * <p>Note: the HEADS of the named references in an export may not match the heads in {@link
-   * HeadsAndForkPoints} when commits happened while the export has been created.
+   * HeadsAndForks} when commits happened while the export has been created.
    */
-  HeadsAndForkPoints headsAndForkPoints();
+  HeadsAndForks headsAndForks();
```
dimas-b, line 38: Why do we remove `HeadsAndForkPoints` here, but keep it in `CommitLogOptimization`, which is in the same java package as this class?

dimas-b: It looks like `CommitLogOptimization` always gets heads and forks from `ImportResult`. I tend to think it might be preferable to change `CommitLogOptimization.headsAndForks` to `HeadsAndForks` and add a utility conversion method to it. WDYT?

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ImportResult.java
snazy: Will go away later - needed for the javadoc (which isn't currently incorrect)

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ImportResult.java
snazy: CLO is only useful for `DatabaseAdapter`, so fine to keep it in CLO

dimas-b: ok

dimas-b (approved):

PR #5564, opened 2022-11-25
upgrade workflows to ubuntu-22.04

XN137:
the image is out of beta and planned to become `ubuntu-latest` later this year

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -43,7 +43,7 @@ on:
 jobs:
   create-release:
     name: Bump version
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       BUMP_ON_BRANCH: ${{ github.event.inputs.bumpBranch }}
       BUMP_TYPE: ${{ github.event.inputs.bumpType }}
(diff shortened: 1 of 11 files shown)


snazy:
Do the Nessie Quarkus native images built with/on the 22.04 GH runner work on older Linux distros?
I.e. build the Nessie Quarkus native image on Ubuntu 22.04, run it on Ubuntu 18.04.
IIRC there were issues a while ago, but it might be fine with the 22.04 GH runner, because the Quarkus native image's built in a container. But would be good to check.

XN137:
did we give any strong support guarantees or why would we care about 18.04 still ? (it seems to go EOL next year)

imo even if it did break, users could just:
- stick to a previous nessie version
- upgrade their machines to 20.04 or 22.04 (which they should do anyway?)
- build their own images from source however they like

snazy:
It is not about Ubuntu, it's about the libc version and e.g. current RH dsitros

snazy:
@XN137 shall we get this one done?

XN137:
i only did this because i thoughts it's simple and risk free.
if there is the risk of breaking things and requires extensive testing then since this is neither important nor urgent, i think we can just close this for now (until the version we are using goes EOL).

PR #5004, opened 2022-08-25
Add REST API v2

dimas-b:
Add v2 REST endpoints.

Add java client implementation for v2 REST API.
Since namespaces are treated as ordinary content objects in API v2
the old java client API for dealing with namespaces is implemented
on the client side on top of v2 "get entries" and "commit" APIs.

Run in-memory REST server tests for both v1 and v2 APIs.

REST test using persistent backend only use v2 API.

Resteasy test remain on v1 API since they construct REST requests
according to Nessie API v1 specs.

OpenAPI YAML is built and published only for v2. OpenAPI for v1
should be obtained from older Nessie publications.

This is the main contribution towards #5112

----

This PR builds on top of #5419, #5420, #5435, #5436.

PR #5584, opened 2022-11-29
Fix OpenAPI spec for DiffApi

dimas-b:
* Avoid using fake {f} blocks in the query path spec.
  Both Swagger and OpenAPI Generator flag that as a validation error:
  "Declared path parameter "f" needs to be defined as a path
  parameter at either the path or operation level"
    
* Move @PathParam annotations for DiffParams from fields to setters
  and use custom parsing code for breaking "ref with hash" parameters
  into the ref name + hash.

### DIFF
--- a/model/src/main/java/org/projectnessie/api/v1/http/HttpDiffApi.java
+++ b/model/src/main/java/org/projectnessie/api/v1/http/HttpDiffApi.java
@@ -38,8 +38,7 @@ public interface HttpDiffApi extends DiffApi {
 
   @GET
   @Produces(MediaType.APPLICATION_JSON)
-  @Path(
-      "{fromRef : [^*]+}{f : [*]?}{fromHashOnRef : ([^.]*)?}...{toRef : [^*]+}{t : [*]?}{toHashOnRef : ([^.]*)?}")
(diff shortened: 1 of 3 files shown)


dimas-b:
Re-implemented slightly to fix tests.

snazy (approved):
LGTM!

Can we change the `*` separator here to the common one in the v2 API?

snazy:
Hm - CI still says:
```
	-attribute paths.'/diffs/{fromRef}{f}{fromHashOnRef}...{toRef}{t}{toHashOnRef}'. Declared path parameter fromRef}{f}{fromHashOnRef needs to be defined as a path parameter in path or operation level
```

dimas-b:
> Can we change the * separator here to the common one in the v2 API?

V2 will use its own syntax: https://github.com/projectnessie/nessie/pull/5004/files#diff-d4a16766b66828e9d383d99375bf5dfb9a8c5df0c568414df241859978498638R48

dimas-b:
The CI error message in the UI build happens because the UI uses a fixed, check-in old YAML version. I'll fix that in a separate PR (or may be under #5004 if/when we combine v1 and v2 in the same yaml)

PR #5599, opened 2022-12-02
Deprecate commit lists in merge-results

snazy:
The commits lists in the merge results can become quite big, possibly "too big", and the value isn't clear. Deprecating those for removal.

### DIFF
--- a/model/src/main/java/org/projectnessie/model/MergeResponse.java
+++ b/model/src/main/java/org/projectnessie/model/MergeResponse.java
@@ -59,14 +59,11 @@ public interface MergeResponse {
   @Nullable
   String getExpectedHash();
 
-  /** List of commit-IDs to be merged or transplanted. */
+  @Deprecated // for removal and replaced with something else
   List<LogEntry> getSourceCommits();
 
-  /**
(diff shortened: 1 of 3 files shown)


model/src/main/java/org/projectnessie/model/MergeResponse.java
```diff
@@ -59,14 +59,11 @@ default boolean wasSuccessful() {
   @Nullable
   String getExpectedHash();
 
-  /** List of commit-IDs to be merged or transplanted. */
+  @Deprecated // for removal and replaced with something else
```
dimas-b, line 62: Could we also add deprecation flag to the OpenAPI via `@Schema` annotations (cf. `RefLogResponse`)?

dimas-b (approved):

model/src/main/java/org/projectnessie/model/MergeResponse.java
snazy: Duh - somehow I thought you did that in the v2 PR

dimas-b (approved):

PR #5600, opened 2022-12-02
Factory methods for EntriesResponse.Entry + DiffResponse.Entry + MergeResult.KeyDetails

snazy:


model/src/main/java/org/projectnessie/model/EntriesResponse.java
```diff
@@ -43,9 +43,15 @@ static ImmutableEntry.Builder builder() {
...
 
     @NotNull
+    @Value.Parameter(order = 2)
     Content.Type getType();
 
     @NotNull
+    @Value.Parameter(order = 1)
     ContentKey getName();
+
+    static Entry entry(ContentKey name, Content.Type type) {
```
dimas-b, line 53: nit: why not the usual `of(...)`?

dimas-b (approved):

model/src/main/java/org/projectnessie/model/EntriesResponse.java
snazy: Sooooo many `of()` methods - it's personal taste, but I like more verbose method names, becasue those can be statically imported.

PR #5604, opened 2022-12-02
Minor version-store commit-log test refactoring

snazy:


versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractCommitLog.java
```diff
@@ -114,23 +113,26 @@ public void commitLogExtendedNoGlobalState() throws Exception {
...
+              .put("k" + i, onRef("v" + i, "c" + i))
+              .put("key" + i, onRef("value" + i, "cid" + i))
+              .delete("delete" + i)
+              .toBranch(branch);
+      hashes.add(head);
+    }
+
     List<Hash> parentHashes =
-        Stream.concat(Stream.of(firstParent), hashes.subList(0, 9).stream())
+        Stream.concat(Stream.of(firstParent), hashes.subList(0, numCommits + 1).stream())
```
dimas-b, line 135: `9 == numCommits - 1` :thinking: was is wrong before?

snazy: It's "shifted by 1" below now, and one more hash added above.

dimas-b: ah, I see... I was thrown off balance by "minor refactoring" ;)

dimas-b: Do we need `subList` now?

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractCommitLog.java
```diff
@@ -144,24 +146,23 @@ public void commitLogExtendedNoGlobalState() throws Exception {
...
-                  .containsExactly(
-                      CommitMeta.fromMessage("Commit #" + i),
-                      hashes.get(i - 1),
-                      parentHashes.get(i - 1),
-                      Arrays.asList(
-                          Delete.of(Key.of("delete" + i)),
-                          Put.of(Key.of("k" + i), onRef("v" + i, "c" + i)),
-                          Put.of(Key.of("key" + i), onRef("value" + i, "cid" + i))));
-            });
+    for (int i = 1; i <= numCommits; i++) {
```
dimas-b, line 149: Why not validate the "initial" commit too (i.e. at index `0`)?.. Maybe in outside this loop, though.

snazy: Let's not make this one too complicated.

dimas-b: This one is a nit (optional), I agree, but how about `subList` above? I think it makes sense to remove it now since the intention is to process the whole list, right?

dimas-b: Still I find having `firstParent` in `parentHashes` confusing if we do not validate it... might as well use `null`... Sorry about nitpicking, just hoping for a more readable code :)

dimas-b (approved):

PR #5607, opened 2022-12-02
Switch nessie-client to use OpenTelemetry

snazy:


clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
-                      try {
-                        inner.setTag(
-                            "http.status_code", responseContext.getResponseCode().getCode());
-                      } catch (IOException e) {
-                        // There's not much we can (and probably should) do here.
-                      }
+            Span span =
+                tracer
+                    .spanBuilder("Nessie-HTTP")
+                    .startSpan()
```
dimas-b, line 102: We should probably set Span kind to `CLIENT`.

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
-              HashMap<String, String> headerMap = new HashMap<>();
-              TextMap httpHeadersCarrier = new TextMapAdapter(headerMap);
-              tracer.inject(inner.context(), Builtin.HTTP_HEADERS, httpHeadersCarrier);
-              headerMap.forEach(context::putHeader);
-            }
+                  }
+                  if (exception != null) {
+                    span.setStatus(StatusCode.ERROR, exception.toString());
+                  }
+                  scope.close();
```
dimas-b, line 126: Are we sure the callback is invoked on the same thread? If not, making this `scope` current (above) would be incorrect.

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
+            Span span =
+                tracer
+                    .spanBuilder("Nessie-HTTP")
+                    .startSpan()
+                    .setAttribute(SemanticAttributes.HTTP_URL, context.getUri().toString())
+                    .setAttribute(SemanticAttributes.HTTP_METHOD, context.getMethod().name())
+                    .setAttribute("component", "http");
+
+            @SuppressWarnings({"resource", "MustBeClosedChecker"})
+            Scope scope = span.makeCurrent();
```
dimas-b, line 108: This is probably not necessary since we propagate headers manually (line 111). It might be simpler to avoid making the span current and inject `Context.current().with(span)` on line 111.

The "current" span concept is only about thread local stuff. It is not critical here since we do not expect any nested traceable events on this java thread during the request.

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
+
+            W3CTraceContextPropagator.getInstance()
+                .inject(Context.current(), context, RequestContext::putHeader);
+
+            context.addResponseCallback(
+                (responseContext, exception) -> {
+                  if (responseContext != null) {
+                    try {
+                      span.setAttribute(
+                          "http.status_code", responseContext.getResponseCode().getCode());
```
dimas-b, line 118: `SemanticAttributes.HTTP_STATUS_CODE`

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
+                    } catch (IOException e) {
+                      // There's not much we can (and probably should) do here.
```
dimas-b, line 120: How about `span.recordException(...)`?

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
-                  .setTag("http.method", context.getMethod().name());
-
-              HashMap<String, String> headerMap = new HashMap<>();
-              TextMap httpHeadersCarrier = new TextMapAdapter(headerMap);
-              tracer.inject(inner.context(), Builtin.HTTP_HEADERS, httpHeadersCarrier);
-              headerMap.forEach(context::putHeader);
-            }
+                  }
+                  if (exception != null) {
+                    span.setStatus(StatusCode.ERROR, exception.toString());
```
dimas-b, line 124: Could you also do `span.recordException(...)`?

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
-                      } catch (IOException e) {
-                        // There's not much we can (and probably should) do here.
-                      }
+            Span span =
+                tracer
+                    .spanBuilder("Nessie-HTTP")
+                    .startSpan()
+                    .setAttribute(SemanticAttributes.HTTP_URL, context.getUri().toString())
+                    .setAttribute(SemanticAttributes.HTTP_METHOD, context.getMethod().name())
+                    .setAttribute("component", "http");
```
dimas-b, line 105: If `component` is not a standard attribute, should we use some prefix?.. e.g. `nessie.component`?

clients/client/src/main/java/org/projectnessie/client/http/NessieHttpClient.java
```diff
@@ -93,43 +91,41 @@ private static void addTracing(HttpClient.Builder httpClient) {
...
-                            "http.status_code", responseContext.getResponseCode().getCode());
-                      } catch (IOException e) {
-                        // There's not much we can (and probably should) do here.
-                      }
+            Span span =
+                tracer
+                    .spanBuilder("Nessie-HTTP")
+                    .startSpan()
+                    .setAttribute(SemanticAttributes.HTTP_URL, context.getUri().toString())
+                    .setAttribute(SemanticAttributes.HTTP_METHOD, context.getMethod().name())
```
dimas-b, line 104: nit: Would it be an overkill to add the client's version and build number?

dimas-b (commented):
Great to see us moving to OTel :)

clients/client/src/main/java/org/projectnessie/client/http/OpentelemetryTracing.java
```diff
@@ -0,0 +1,75 @@
...
+            Span span =
+                tracer
+                    .spanBuilder("Nessie-HTTP")
+                    .setSpanKind(SpanKind.CLIENT)
+                    .startSpan()
+                    .setAttribute(SemanticAttributes.HTTP_URL, context.getUri().toString())
+                    .setAttribute(SemanticAttributes.HTTP_METHOD, context.getMethod().name())
+                    .setAttribute("nessie.version", NessieVersion.NESSIE_VERSION);
+
+            W3CTraceContextPropagator.getInstance()
```
dimas-b, line 51: I just realized that we should probably take the propagator from the env.... that is use whatever is configured in the "container" running the Nessie client... I do not recall off the top of my head what API we need to use on the OTel side, but I hope something is available to let us get "active" propagators :thinking:

dimas-b (commented):
Just one late comment about propagators

clients/client/src/main/java/org/projectnessie/client/http/OpentelemetryTracing.java
snazy: We had the Jaeger specific one before, but that seems to have disappeared, so there's the W3C one here.

I'd assume, that if tracing is explicitly enabled, it's properly setup in the calling code?

dimas-b (approved):
We can add support for other propagators later

PR #5610, opened 2022-12-02
Use soft-assertions in jaxrs-tests

snazy:
Soft-assertions were really helpful while working on the storage-stuff, so I think, it makes sense to add them to the main code base.

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRest.java
@@ -19,6 +19,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import com.google.common.collect.ImmutableMap;
 import java.util.Locale;
+import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.projectnessie.client.api.NessieApiV1;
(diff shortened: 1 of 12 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRest.java
```diff
@@ -60,6 +65,10 @@ public NessieApiV1 getApi() {
 
   @AfterEach
   public void tearDown() throws Exception {
+    // Cannot use @ExtendWith(SoftAssertionsExtension.class) + @InjectSoftAssertions here, because
+    // of Quarkus class loading issues. See https://github.com/quarkusio/quarkus/issues/19814
+    soft.assertAll();
```
dimas-b, line 70: Will this prevent test branch cleanup (below) if it throws? Could we put it in a separate `@AfterEach` method instead?

dimas-b (commented):
LGTM overall. Just one concern with test data cleanup.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRest.java
snazy: Ah, good catch

snazy: try-finally works, too

dimas-b (approved):

PR #5631, opened 2022-12-03
Export/import: explicit tests for exporter/importer implementations

snazy:
Adds explicit tests for directory+zip export/import backends - plus some code style + validation additions. No functional change.

### DIFF
--- a/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/files/FileExporter.java
+++ b/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/files/FileExporter.java
@@ -15,12 +15,20 @@
  */
 package org.projectnessie.versioned.transfer.files;
 
-import com.google.common.base.Preconditions;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static java.nio.file.Files.createDirectories;
+import static java.nio.file.Files.exists;
+import static java.nio.file.Files.isDirectory;
+import static java.nio.file.Files.newOutputStream;
+
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.stream.Stream;
+import javax.annotation.Nonnull;
 import org.immutables.value.Value;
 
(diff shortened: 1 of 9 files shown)


versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/files/FileExporter.java
```diff
@@ -39,23 +47,37 @@ public interface Builder {
...
     }
   }
 
   @Override
-  public OutputStream newFileOutput(String fileName) throws IOException {
-    return Files.newOutputStream(targetDirectory().resolve(fileName));
+  @Nonnull
+  public OutputStream newFileOutput(@Nonnull String fileName) throws IOException {
+    checkArgument(
+        fileName.indexOf('/') == -1 && fileName.indexOf('\\') == -1, "Directories not supported");
```
dimas-b, line 74: nit: would it make sense to do this check after a `Path` is constructed from the name so that we do not have to worry about OS-specific separators?

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/files/ZipArchiveImporter.java
```diff
@@ -44,10 +48,15 @@ ZipFile zipFile() throws IOException {
...
-  public InputStream newFileInput(String fileName) throws IOException {
+  @Nonnull
+  public InputStream newFileInput(@Nonnull String fileName) throws IOException {
     @SuppressWarnings("resource")
     ZipFile zip = zipFile();
     ZipEntry entry = zip.getEntry(fileName);
+    if (entry == null) {
+      throw new FileNotFoundException(fileName);
+    }
+    checkState(!entry.isDirectory(), "%s is a directory, expect a file", fileName);
```
dimas-b, line 59: nit: `expect` -> `expected`?

dimas-b (approved):

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/files/FileExporter.java
snazy: \\ isn't recognized by Path on Unix (and vice versa)

PR #5639, opened 2022-12-05
Fix GetReferenceParams description for API v2

dimas-b:


model/src/main/java/org/projectnessie/api/v2/params/GetReferenceParams.java
```diff
@@ -31,8 +29,21 @@
 public class GetReferenceParams {
 
   @Parameter(
-      description = REF_NAME_DESCRIPTION,
-      examples = {@ExampleObject(ref = "ref")})
+      description =
+          "A reference to a particular version of the contents tree (a point in history).\n"
```
snazy, line 33: Should we maybe move the text `ApiDoc` near `REF_PARAMETER_DESCRIPTION` as `REF_NAME_PARAMETER_DESCRIPTION`?

model/src/main/java/org/projectnessie/api/v2/params/GetReferenceParams.java
```diff
@@ -31,8 +29,21 @@
...
+              + "\n"
+              + "This reference can be specification in these forms:\n"
+              + "- \\- (literal minus character) - identifies the HEAD of the default branch.\n"
+              + "- name - Identifies the HEAD commit of a branch or tag.\n"
+              + "- name@hash - Identifies the 'hash' commit on a branch or tag.\n"
+              + "\n"
+              + "Currently, the hash part is permitted to be specified in the parameter,"
```
snazy, line 40: Maybe rephrase a little ("Currently" is actually "Permanently" ;) ) or even remove this sentence entirely.

dimas-b: I thought, we'd validate the hash if given... rephrasing

model/src/main/java/org/projectnessie/api/v2/params/GetReferenceParams.java
dimas-b: moved

snazy (approved):

PR #5641, opened 2022-12-05
New Nessie storage model

snazy:
Implementation of the new Nessie storage model, driven by some inefficiencies in the current ("database adapter") storage model.

* Storage space efficient index structures, with layered / striped indexes, supporting N * 100,000 keys
* Flexible named references
* Flexible store-keys to allow multiple "key universes" and "key variants" (latter: per "content key")
* Simpler code


versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -0,0 +1,154 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.storage.cache;
+
+import static org.projectnessie.versioned.storage.common.persist.ObjId.EMPTY_OBJ_ID;
+import static org.projectnessie.versioned.storage.serialize.ProtoSerialization.serializeObj;
+
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
```
XN137, line 22: should this import from caffeine?

snazy: This PR isn't ready for review, let's not add review comments for now.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * Lists references, starting at the given reference name, only available from <em>{@link
+   * #isTransactional() transactional}</em> implementations.
+   *
+   * <p>Note: this lists all references with reference names greater than or equal to {@code begin},
+   * which is different from {@link ReferenceLogic#queryReferences(ReferencesQuery)}, which performs
+   * a <em>prefix</em> check.
+   */
+  default CloseableIterator<Reference> listReferences(String begin) {
+    throw new UnsupportedOperationException(
```
dimas-b, line 173: Would it be worth introducing a sub-interface for Tx persistence implementations?

dimas-b (commented):
Posting some initial, superficial comments as a token that I started reviewing :)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
snazy: Not sure yet - two reasons:
* I did not want to have a "interface hierarchy nightmare" (as we have with the `*Config` interfaces in DA)
* Don't really know whether it's really a good idea to have a different model for Tx (up for discussion)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
```diff
@@ -0,0 +1,607 @@
...
+ * reference is accessed:
+ *
+ * <ul>
+ *   <li>{@link Persist#findReference(String)} (or {@link Persist#findReferences(String[])}) is used
+ *       to fetch a reference by name.
+ *   <li>If the reference is found and {@link Reference#deleted()} is {@code false}, the reference
+ *       has been found and can be returned.
+ *   <li>If the the {@link Reference#deleted()} flag is {@code true}, it is possible that a previous
+ *       delete-reference operation failed in the middle. The implementation must resume the
+ *       reference-deletion process described above.
```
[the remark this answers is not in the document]
snazy: I'd like to avoid that in the ref-logic-impl, because it would also be tricky for "tools" (what could that do/be?) w/ the reference-row and references-index.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
dimas-b: > Don't really know whether it's really a good idea to have a different model for Tx (up for discussion)

If the distinction is mostly for a transactional list of references vs. a commit log-based list of references (not sure now - I've not reviewed all of the code yet). I'd think it might be preferable to have the same model for both backend types.

snazy: I could imagine that _listing_ all references should be quicker w/ the non-tx model - i.e. having our own index at hand from a single row feels quicker / more efficient than listing many database rows. IDK

snazy: I could imagine that _listing_ all references should be quicker w/ the non-tx model - i.e. having our own index at hand from a single row feels quicker / more efficient than listing many database rows. IDK

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -0,0 +1,740 @@
...
+    }
+  }
+
+  private Supplier<CommitObj> supplyRefHeadForNamedRefs(
+      RefMapping refMapping, GetNamedRefsParams params) throws ReferenceNotFoundException {
+    NamedRef baseReference = params.getBaseReference();
+    if (baseReference == null) {
+      return () -> null;
+    }
+    Reference ref = refMapping.resolveNamedRef(baseReference);
```
dimas-b, line 273: Why do we resolve the ref immediately, but load the commit in a supplier? What's the rationale for this distinction?

snazy: yea, not really necessary, but also not overkill.. updated.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
XN137: since you now seem to take early review comments:
this should still import from caffeine instead of guava :sweat_smile:

snazy: fixed

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
```diff
@@ -0,0 +1,619 @@
...
+
+            return new CommitReferenceResult(reference(name, pointer, false), ADDED_TO_INDEX);
+          });
+    } catch (CommitConflictException e) {
+      checkState(e.conflicts().size() == 1, "Unexpected amount of conflicts %s", e.conflicts());
+
+      CommitConflict conflict = e.conflicts().get(0);
+      checkState(conflict.conflictType() == KEY_EXISTS, "Unexpected conflict type %s", conflict);
+
+      Supplier<StoreIndex<CommitOp>> indexSupplier = createRefsIndexSupplier();
```
dimas-b, line 412: Should this be using the `REF_REFS` state from the last try (line 381) as opposed to "current" HEAD?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
```diff
@@ -0,0 +1,619 @@
...
+      @Nonnull @jakarta.annotation.Nonnull List<String> references) {
+    Reference[] refs = persist.findReferences(references.toArray(new String[0]));
+    List<Reference> r = new ArrayList<>(refs.length);
+
+    Supplier<StoreIndex<CommitOp>> refsIndexSupplier = createRefsIndexSupplier();
+
+    for (int i = 0; i < refs.length; i++) {
+      Reference ref = refs[i];
+
+      ref = maybeRecover(references.get(i), ref, refsIndexSupplier);
```
dimas-b, line 189: I might have mentioned this before (not sure). WDYT about using a unique ID for each reference and tracking that ID in the index (commit against internal refs). Normal commits on a ref would do CAS keyed by that ID as opposed to ref name. This should be simplier is dealing with same names in different reference "generations" (create/delete cycles), I guess. We would not have to assume that recovery happened before the next ref generation is used.

dimas-b (commented):
A loose collection of unrelated comments :) The main one this time is about CAS on ref IDs as opposed to names.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
snazy: IMO tracking an ID  for a reference makes it overly complicated and also eliminates the ability to do range-queries on reference names, and the recovery logic would "explode" (or become impossible). TL;DR all the benefits that the ref-name index brings would be gone.

I doubt there's a real use case by which users re-use reference names that quickly.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
```diff
@@ -0,0 +1,619 @@
...
+ *
+ * <p>Note: Updates to a named reference are <em>not</em> tracked in {@link InternalRef#REF_REFS}.
+ *
+ * <h3>Delete reference / Logical approach</h3>
+ *
+ * <ol>
+ *   <li>The {@link Reference} is marked as {@link Reference#deleted()} using {@link
+ *       Persist#markReferenceAsDeleted(Reference)}.
+ *   <li>Then the reference deletion is tracked by creating a commit in {@link InternalRef#REF_REFS}
+ *       using a {@link Action#REMOVE} operation. Implementations must only commit, if the reference
```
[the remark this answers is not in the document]
snazy: It would need _2_ commits - one that writes the updated `RefObj` and one that deletes it. (A REMOVE+ADD is a re-add).

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBBackendFactory.java
```diff
@@ -0,0 +1,54 @@
...
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  public MongoDBBackendConfig parseConfig(
+      @Nullable @jakarta.annotation.Nullable MongoDBBackendConfig baseConfig,
+      @Nonnull @jakarta.annotation.Nonnull Function<String, String> config) {
+    ImmutableMongoDBBackendConfig.Builder v = MongoDBBackendConfig.builder();
+    if (baseConfig != null) {
+      v.from(baseConfig);
+    }
+    return v.build();
```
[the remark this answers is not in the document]
snazy: (function gone)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/BackendFactory.java
```diff
@@ -0,0 +1,38 @@
...
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  String name();
+
+  /** Helper to construct backend configurations from for example property bags. */
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  C parseConfig(
+      @Nullable @jakarta.annotation.Nullable C baseConfig,
+      @Nonnull @jakarta.annotation.Nonnull Function<String, String> config);
```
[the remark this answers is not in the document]
snazy: Yea - that was probably intended to be configured via system properties or so. But it's unused, so I removed it for now.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
```diff
@@ -0,0 +1,1272 @@
...
+    abstract O deserialize(ResultSet rs, ObjId id) throws SQLException;
+
+    abstract int storeNone(PreparedStatement ps, int idx) throws SQLException;
+
+    abstract int store(
+        PreparedStatement ps, int idx, O obj, int incrementalIndexLimit, int maxSerializedIndexSize)
+        throws SQLException, ObjTooLargeException;
+  }
+
+  static Map<ObjType, StoreObjDesc<?>> STORE_OBJ_TYPE = new EnumMap<>(ObjType.class);
```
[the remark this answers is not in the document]
snazy: Sounds good - shall we leave that as a follow-up?

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -0,0 +1,1019 @@
...
+    Document inner = new Document();
+    doc.put(ID_PROPERTY_NAME, idObjDoc(id));
+    doc.put(COL_OBJ_TYPE, type.shortName());
+    int incrementalIndexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIncrementalIndexSizeLimit();
+    int indexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIndexSegmentSizeLimit();
+    //noinspection unchecked
+    storeObj.objToDoc(obj, inner, incrementalIndexSizeLimit, indexSizeLimit);
+    doc.put(storeObj.typeName, inner);
```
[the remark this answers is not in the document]
snazy: It's somewhat the idea to have some kind of a schema here, but without "looong" attribute names.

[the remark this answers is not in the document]
dimas-b: Interesting :thinking: I wonder whether Mongo handles nested stuff as efficiently as flat objects, but I do not have any real world observations for this :)

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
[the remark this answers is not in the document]
dimas-b: Sure, but for now can it be `private` or at least have declared type of `EnumMap` to claim certain expected behaviours?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/SetupLogicImpl.java
```diff
@@ -0,0 +1,214 @@
...
+          RepositoryDescription.builder()
+              .oldestPossibleCommitTime(now)
+              .repositoryCreatedTime(now)
+              .defaultBranchName(defaultBranchName);
+
+      repositoryDescription.accept(repoDesc);
+
+      ByteString text = unsafeWrap(SHARED_OBJECT_MAPPER.writeValueAsBytes(repoDesc.build()));
+
+      StringObj string = stringData("application/json", Compression.NONE, null, emptyList(), text);
```
dimas-b, line 175: nit: content type constant

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
```diff
@@ -0,0 +1,536 @@
...
+    List<IndexStripe> commitStripes = commit.referenceIndexStripes();
+    if (!commitStripes.isEmpty()) {
+      checkState(
+          referenceIndexId == null,
+          "Commit %s: must not have both pointer to a reference index and stripes",
+          commit.id());
+      StoreIndex<CommitOp> referenceIndex = referenceIndexFromStripes(commitStripes, commit.id());
+      index = layeredIndex(referenceIndex, incremental);
+    }
+    if (referenceIndexId != null) {
```
dimas-b, line 198: Should this be `else if`? The `index` value from line 196 would be overwritten below.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * This is an <em>internal</em> utility class, which provides an unsynchronized, not thread-safe
+ * {@link Supplier} that memoizes returned value but also a thrown exception.
+ */
+public final class SupplyOnce {
+
+  private SupplyOnce() {}
+
+  public static <T> Supplier<T> supplyOnce(Supplier<T> loader) {
```
dimas-b, line 28: This is neat, but at the time time slightly confusing :) Makes me wonder (at call sites) what to expect on the second `.get()` call... Maybe simply `memoize(Supplier<T> loader)`?

dimas-b (commented):
Another batch of loose comments

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,341 @@
...
+   *
+   * @return the object with the requested type, {@code null} for {@link ObjId#EMPTY_OBJ_ID}
+   * @throws ObjNotFoundException with the ID for which no matching {@link Obj objects} exist,
+   *     either the no object with the given ID exists or that object is not of the requested type
+   * @see #fetchObjType(ObjId)
+   * @see #fetchObj(ObjId)
+   * @see #fetchObjs(ObjId[])
+   */
+  <T extends Obj> T fetchTypedObj(
+      @Nonnull @jakarta.annotation.Nonnull ObjId id, ObjType type, Class<T> typeClass)
```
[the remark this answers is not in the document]
snazy: Neccessary for type-safety / generics

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/SetupLogicImpl.java
snazy: Meh - only used once

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
snazy: done

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/SetupLogic.java
```diff
@@ -0,0 +1,37 @@
...
+  void initialize(@Nonnull @jakarta.annotation.Nonnull String defaultBranchName);
+
+  void initialize(
+      @Nonnull @jakarta.annotation.Nonnull String defaultBranchName,
+      boolean createDefaultBranch,
+      Consumer<RepositoryDescription.Builder> repositoryDescription);
+
+  @Nullable
+  @jakarta.annotation.Nullable
+  RepositoryDescription fetchRepositoryDescription();
```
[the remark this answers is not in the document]
snazy: renamed

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: Why not make it a type parameter on `ObjType`? smth like: ` <T extends Obj> T fetchTypedObj(..., ObjType<T> type)`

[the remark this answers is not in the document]
dimas-b: IMHO, having a `typeClass` as a parameter that is passed in, but not used, does not add type safety, because the caller can use any class as `typeClass` and still get a `ClassCastException` in runtime :thinking:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,341 @@
...
+
+  /**
+   * Erases the whole repository.
+   *
+   * <p>It is possible that databases have to scan all rows/items in the tables/collections, which
+   * can lead to a <em>very</em> long runtime of this method.
+   */
+  void erase();
+
+  boolean isCaching();
```
dimas-b, line 340: Is it still used?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
dimas-b: Range queries use our internal index (from the commit on `int/refs`), right? They do not rely on the backend database, do they?

My thinking was to allocate a unique ID for a ref, then store it in the database, then commit to `int/refs`. The commit will act as the source of truth for listing the references (getting the set of names). Active recovery of partially created / deleted references will not be necessary, only passive cleanup of failed create attempts. Even in case the cleanup fails, it won't invalidate the state of the repository.

An extra hop through the commit at `inf/refs` will be required for reference lookup at the API level, but caching can probably mitigate that. WDYT?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -0,0 +1,735 @@
...
+import org.projectnessie.versioned.storage.common.logic.PagingToken;
+import org.projectnessie.versioned.storage.common.logic.ReferenceLogic;
+import org.projectnessie.versioned.storage.common.objtypes.CommitObj;
+import org.projectnessie.versioned.storage.common.objtypes.CommitOp;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.Persist;
+import org.projectnessie.versioned.storage.common.persist.Reference;
+import org.projectnessie.versioned.storage.versionstore.BaseCommitHelper.CommitterSupplier;
+
+public class VersionStoreImpl implements VersionStore {
```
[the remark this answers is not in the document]
snazy: Yes, sure

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,341 @@
...
+
+  /**
+   * Low-level, find a reference.
+   *
+   * <p><em>Do not use this function from service implementations, use {@link ReferenceLogic}
+   * instead!</em>
+   *
+   * @return the reference or {@code null}, if it does not exist
+   */
+  Reference findReference(@Nonnull @jakarta.annotation.Nonnull String name);
```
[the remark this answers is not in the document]
snazy: --> `fetchReference` (similar to `fetchObj*`)

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/RefMapping.java
```diff
@@ -0,0 +1,313 @@
...
+      return referenceLogic.getReference(refName);
+    } catch (RefNotFoundException e) {
+      throw referenceNotFound(namedRef);
+    }
+  }
+
+  public Reference resolveNamedRef(@Nonnull @jakarta.annotation.Nonnull String refName)
+      throws ReferenceNotFoundException {
+    Reference[] refs =
+        persist.findReferences(new String[] {asBranchName(refName), asTagName(refName)});
```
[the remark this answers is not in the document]
snazy: should be ReferenceLogic here

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,640 @@
...
+      CommitReferenceResult commitToIndex = commitCreateReference(name, pointer);
+      Reference reference = commitToIndex.reference;
+
+      switch (commitToIndex.kind) {
+        case ADDED_TO_INDEX:
+          checkState(!reference.deleted(), "internal error");
+          try {
+            return persist.addReference(reference);
+          } catch (RefAlreadyExistsException e) {
+            if (!e.reference().deleted()) {
```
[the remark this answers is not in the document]
snazy: Nice catch!
So yes, `RefAlreadyExistsException` can return `null` and that's correct, because a concurrent reference-delete might have happened. So all call sites of `org.projectnessie.versioned.storage.common.exceptions.RefAlreadyExistsException` must use fetchReference()

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
```diff
@@ -0,0 +1,183 @@
...
+      sleepAndBackoff(totalElapsed, attemptElapsed);
+
+      return true;
+    }
+
+    private void sleepAndBackoff(long totalElapsed, long attemptElapsed) {
+      long lower = lowerBound;
+      long upper = upperBound;
+      long sleepMillis =
+          lower == upper ? lower : ThreadLocalRandom.current().nextLong(lower, upper);
```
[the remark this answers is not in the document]
snazy: nah

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,341 @@
...
+  @jakarta.annotation.Nonnull
+  String name();
+
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  StoreConfig config();
+
+  default void sleep(long millis) {
+    try {
+      Thread.sleep(millis);
```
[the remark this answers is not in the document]
snazy: Testing ;)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,640 @@
...
+
+  @VisibleForTesting // needed to simulate recovery scenarios
+  CommitReferenceResult commitCreateReference(String name, ObjId pointer)
+      throws RetryTimeoutException {
+    try {
+      return commitRetry(
+          persist,
+          (p, retryState) -> {
+            Reference refRefs = p.findReference(REF_REFS.name());
+            long created = p.config().currentTimeMicros();
```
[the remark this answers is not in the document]
snazy: can do when it'll be needed

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -0,0 +1,1096 @@
...
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  public PagedResult<ObjId, ObjId> commitIdLog(
+      @Nonnull @jakarta.annotation.Nonnull CommitLogQuery commitLogQuery) {
+    ObjId startCommitId =
+        commitLogQuery
+            .pagingToken()
+            .map(PagingToken::token)
+            .map(ObjId::objIdFromBytes)
+            .orElse(commitLogQuery.commitId());
```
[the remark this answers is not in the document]
snazy: The idea here is to move the whole paging-token logic into the actual logic implementations, have signing mechanisms and such. But that cannot be done at this stage w/ the existing `DatabaseAdapter` implementations.

snazy:
> Range queries use our internal index (from the commit on int/refs), right? They do not rely on the backend database, do they?
>
> My thinking was to allocate a unique ID for a ref, then store it in the database, then commit to int/refs. The commit will act as the source of truth for listing the references (getting the set of names). Active recovery of partially created / deleted references will not be necessary, only passive cleanup of failed create attempts. Even in case the cleanup fails, it won't invalidate the state of the repository.
> 
> An extra hop through the commit at inf/refs will be required for reference lookup at the API level, but caching can probably mitigate that. WDYT?

Not sure whether that's a good enough argument to add an additional database roundtrip. But recovery would still be needed, because both the CL entry and the database for for the ref are still necessary.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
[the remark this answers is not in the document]
snazy: #6183

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/exceptions/CommitConflictException.java
```diff
@@ -0,0 +1,42 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.storage.common.exceptions;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import org.projectnessie.versioned.storage.common.logic.CommitConflict;
+
+public class CommitConflictException extends CommitException {
```
adutra, line 22: Most, if not all, exceptions in this package look more like runtime exceptions. Any reason why you preferred to make them all checked? This choice forces you to introduce things like `SQLRunnableExceptions` which might not have been required were these exceptions unchecked in the first place.

adutra (commented):
I have just started reviewing. Here is a tiny first round of (mostly boring) comments. Will continue later today.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/exceptions/CommitConflictException.java
snazy: Those are checked exceptions, to make callers explicitly aware that these can happen (or not). Similar to `NessieConflictException`

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/CassandraBackendBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.projectnessie.quarkus.config.VersionStoreConfig.VersionStoreType.CASSANDRA;
+
+import javax.enterprise.context.Dependent;
+import org.projectnessie.versioned.storage.common.persist.Backend;
+
+@StoreType(CASSANDRA)
+@Dependent
+public class CassandraBackendBuilder implements BackendBuilder {
+
+  // @Inject CompletionStage<QuarkusCqlSession> client;
```
adutra, line 27: If this is your only injection point of the driver session, maybe you don't need the Cassandra extension. Just create the session manually, and read the required properties (contact points and local DC) from some config properties like you are doing below to retrieve the keyspace.

snazy:
Note: I've "effectively disabled" the C* store-type in Quarkus, because at this point we cannot be sure that there will be a version of the C*-quarkus-driver that works with Quarkus 3.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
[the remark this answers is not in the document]
dimas-b: noted

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: `fetchReference` SGTM :)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: But shouldn't testing use a custom clock impl?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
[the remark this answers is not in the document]
dimas-b: that's fine, but data based on this (old) code will have accumulated :)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,644 @@
...
+          try {
+            return persist.addReference(reference);
+          } catch (RefAlreadyExistsException e) {
+            Reference existing = e.reference();
+            if (existing != null) {
+              if (!existing.deleted()) {
+                // Might happen in a rare race
+                throw e;
+              }
+              maybeRecover(name, reference, createRefsIndexSupplier());
```
dimas-b, line 289: Can we actually recover anything in this case? On `ADDED_TO_INDEX` the `reference` is non-null and not deleted, so `maybeRecover` will not do anything, it seems... and the same holds about `existing`.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
snazy: It's a Java language necessity.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
snazy: Testing overloads sleep() to _not_ sleep but check the arg

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,644 @@
...
+          (p, retryState) -> {
+            Reference refRefs = requireNonNull(p.fetchReference(REF_REFS.name()));
+            long created = p.config().currentTimeMicros();
+            RefObj ref = ref(name, pointer, created);
+            ObjId refObjId;
+            try {
+              refObjId = p.storeObj(ref);
+              if (refObjId == null) {
+                // Oops, such a RefObj already exists
+                throw new RetryException();
```
dimas-b, line 394: It looks like this try will run with the same input data and hit the same object ID collision (we do not expect objects to be removed from the store, right?)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
```diff
@@ -0,0 +1,341 @@
...
+   * @return the non-null object ID if the object was stored as a new record or {@code null} if an
+   *     object with the same ID already exists.
+   * @throws ObjTooLargeException thrown when a hard database row/item size limit has been hit, or,
+   *     if {@code ignoreSoftSizeRestrictions} is {@code false}, a "soft" size restriction in {@link
+   *     #config()}
+   * @see #storeObjs(Obj[])
+   */
+  @Nullable
+  @jakarta.annotation.Nullable
+  ObjId storeObj(@Nonnull @jakarta.annotation.Nonnull Obj obj, boolean ignoreSoftSizeRestrictions)
```
dimas-b, line 259: Most (all?) implementations return the `obj.id()` from this method on success and return `null` on exceptions. Replacing an already available ID with `null` to indicate errors looks odd. Would it make sense to return a `boolean` instead?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: I do not see any test code calling this method. All call paths appear to go through `CommitRetry`

dimas-b:
(apparenly IJ reviews interfere with GH web UI reviews :sweat: Submitting a comment in GH Web also submits all in progress IJ comments :shrug: )

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
snazy: There was a case that needed it. Guess some test fails, if that line's deleted.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
snazy: handles races (concurrent, same ref-creation)

snazy: maybve - it likely never ever happens

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: `ObjType` could be a generic class or itnerface. It does not have to be an `enum`, I guess.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
dimas-b: Since `RefObj` is immutable, do we have to re-try in this case? Why not simply carry on?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
dimas-b: We can iron this out later, of course.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
snazy: feels also doable - probably better.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
snazy: refactored

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
snazy: org.projectnessie.versioned.storage.common.logic.TestCommitRetry#doesNotSleepLongerThanMax - last line for example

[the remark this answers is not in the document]
dimas-b: That's exactly what I mean :) This call hits `MonotonicClock` in `CommitRetry`. It does not hit `Persist.sleep()`.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
snazy: Why? It handles the race when another party created the same reference - and if that's the case, it just returns the created one.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java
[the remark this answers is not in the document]
snazy: Huh...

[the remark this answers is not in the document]
snazy: Eh - why did it show me the wrong diff in IJ ....

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -0,0 +1,454 @@
...
+      String datacenters =
+          metadata.getNodes().values().stream()
+              .map(Node::getDatacenter)
+              .distinct()
+              .map(dc -> format("'%s': %d", dc, replicationFactor))
+              .collect(Collectors.joining(", "));
+
+      session.execute(
+          format(
+              "CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'NetworkTopologyStrategy', %s};",
```
adutra, line 380: This makes me uncomfortable as we are using the same replication factor for all dcs, but it might not be achievable in all of them. If we really need to create the schema on the fly, may I suggest that we require from user to provide the RFs on a per-dc basis.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/AbstractCassandraBackendTestFactory.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Optional;
+import org.projectnessie.versioned.storage.testextension.BackendTestFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.CassandraContainer;
+import org.testcontainers.containers.ContainerLaunchException;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+import org.testcontainers.utility.DockerImageName;
+
+abstract class AbstractCassandraBackendTestFactory implements BackendTestFactory {
```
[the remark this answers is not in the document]
snazy: It's shared w/ other modules for integration tests. Database-adapters code had a separate project just for that single purpose, but that felt "overly abstracted". It's nothing that "hurts" prod code.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -0,0 +1,454 @@
...
+          new BiFunction<AsyncResultSet, Throwable, Object>() {
+            @Override
+            public Object apply(AsyncResultSet rs, Throwable ex) {
+              if (ex != null) {
+                noteException(ex);
+                terminate.accept(query);
+              } else {
+                try {
+                  // Needed to ensure that writes into the 'result' get populated
+                  // TODO verify that this the implementation is correct in terms of the JMM:
```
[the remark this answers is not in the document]
snazy: :facepalm: thanks for that - will do

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicNonTxImpl.java
dimas-b: This piece of code was refactored... I do not recall its original form :)

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: Actually, users should provide an already created keyspace. This one's rather a "damn stupid" default behavior. The Quarkus prod code does (did) require a preexisting keyspace IIRC.

Shall we defer all C* related stuff to later in a follow-up when we re-add C*?

snazy: I suspect, it was enforced by the C*-quarkus-extension?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/AbstractCassandraBackendTestFactory.java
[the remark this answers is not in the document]
snazy: Or better: [c|sh]ould be

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: BUt I've adopted the code to explicitly require the keyspace (and removed the "CREATE KEYSPACE" from prod code)

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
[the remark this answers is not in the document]
adutra: UPDATE: After re-checking this once more, it seems the code is fine, and I apologize for the false alert. Although, this is definitely pushing the JMM to its limits :-)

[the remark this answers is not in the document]
snazy: Atomic*Array is more intuitive though - and it's already refactored ;)

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/CassandraBackendBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+    throw new UnsupportedOperationException(
+        "The CASSANDRA version store type is currently unsupported, because there is no Quarkus 3 "
+            + "compatible version of com.datastax.oss.quarkus:cassandra-quarkus-client.");
+    // TODO   CassandraBackendFactory factory = new CassandraBackendFactory();
+    //    try {
+    //      CassandraBackendConfig c =
+    //          CassandraBackendConfig.builder()
+    //              .client(client.toCompletableFuture().get())
+    //              .keyspace(keyspace)
+    //              .replicationFactor(1)
```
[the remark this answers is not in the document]
snazy: That stuff's moved to "test only" now

adutra:
Re: Cassandra, the ball is in their court:

https://github.com/datastax/cassandra-quarkus/pull/222

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,636 @@
...
+      switch (commitToIndex.kind) {
+        case ADDED_TO_INDEX:
+          checkState(!reference.deleted(), "internal error");
+          try {
+            return persist.addReference(reference);
+          } catch (RefAlreadyExistsException e) {
+            Reference existing = e.reference();
+            if (existing != null && !existing.deleted()) {
+              // Might happen in a rare race
+              throw e;
```
dimas-b, line 286: Not sure about the exceptopn here. I believe in most cases `existing` would be recovered by other Nessie processes and hence would be the same as `reference`... so it might be preferable to check equality and return success. WDYT?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,636 @@
...
+            }
+            // try again
+            break;
+          }
+        case REF_ROW_MISSING:
+          checkState(!reference.deleted(), "internal error");
+          // Note: addReference() may or may not throw a ReferenceAlreadyExistsException
+          reference = persist.addReference(reference);
+          throw new RefAlreadyExistsException(reference);
```
dimas-b, line 295: Why throw in this case? We've just created the requested reference - it's both in the index and has a HEAD row.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,636 @@
...
+        ref =
+            persist.fetchTypedObj(
+                requireNonNull(el.content().value(), "Reference commit operation has no value"),
+                REF,
+                RefObj.class);
+      } catch (ObjNotFoundException ex) {
+        throw new RuntimeException("Internal error getting reference creation object", e);
+      }
+      return new CommitReferenceResult(
+          reference(name, ref.initialPointer(), false), REF_ROW_MISSING);
```
dimas-b, line 442: It seems to me that moving the two "ref row *" sub-cases to `createReference` would make the overall logic easier to follow. This method would also contain the commit login, `createReference` would deal with finding and inserting other rows.  WDYT?

dimas-b (commented):
Still trying to wrap my head around ref logic... not sure all my comments made sense, but here they are.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -0,0 +1,636 @@
...
+          });
+    } catch (CommitConflictException e) {
+      checkState(e.conflicts().size() == 1, "Unexpected amount of conflicts %s", e.conflicts());
+
+      CommitConflict conflict = e.conflicts().get(0);
+      checkState(conflict.conflictType() == KEY_EXISTS, "Unexpected conflict type %s", conflict);
+
+      Supplier<StoreIndex<CommitOp>> indexSupplier = createRefsIndexSupplier();
+      StoreIndexElement<CommitOp> el = indexSupplier.get().get(key(name));
+      checkNotNull(el, "Key %s missing in index", name);
```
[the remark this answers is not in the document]
snazy: > not be the same commit that caused the conflict

Yes. the intent here is to _handle_ the concurrent-create case, and otherwise throw.

What you describe is a create _and_ delete happening concurrent to "our" create. It'll be handled, but not gracefully.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
snazy: Let's tackle that outside that code - not duplicate the same cases.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
snazy: Sounds rather cosmetic to me.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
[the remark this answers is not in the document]
dimas-b: I mean that `el` may be `null` under normal concurrent modifications... Could we handle that gracefully, as opposed to throwing from `checkNotNull`?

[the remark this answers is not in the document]
snazy: Not without a ton of code and tons of `if`s checking for all likely and unlikely race conditions.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
```diff
@@ -0,0 +1,535 @@
...
+      firstLastKeys.add(s.firstKey());
+      firstLastKeys.add(s.lastKey());
+    }
+    if (stripes.size() == 1) {
+      return stripes.get(0);
+    }
+
+    IndexLoader<CommitOp> indexLoader =
+        indexesToLoad -> {
+          checkArgument(indexesToLoad.length == loaded.length);
```
dimas-b, line 336: This (bulk) loader and the loader inside the lazy stripes essentially load the same data just on different call paths but their "scope of responsibility" is essentially the same. WDYT about manging stripe their loading via one  `IndexLoader`? I suppose we could have the lazy index load its data via the same loader interface.

dimas-b (commented):
I'm pretty much done with reviewing the `Persist` stuff and the various "logics". I guess I cannot really cover the whole change set in details in reasonable time :)

Is there any particular area of the code, to which you'd like me to pay closer attention, @snazy ?

versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoDBPersist.java
```diff
@@ -0,0 +1,1161 @@
...
+    checkArgument(storeObj != null, "Cannot serialize object type %s ", type);
+
+    Map<String, AttributeValue> item = new HashMap<>();
+    Map<String, AttributeValue> inner = new HashMap<>();
+    item.put(KEY_NAME, objKey(id));
+    item.put(COL_OBJ_TYPE, fromS(type.shortName()));
+    int incrementalIndexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIncrementalIndexSizeLimit();
+    int indexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIndexSegmentSizeLimit();
```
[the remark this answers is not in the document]
snazy: Indexes alone, sure. But serialized indexes are embedded into (database specific) objects - and the size of the whole thing counts.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -0,0 +1,1094 @@
...
+    StoreIndex<CommitOp> index;
+    StoreIndex<CommitOp> fullIndex;
+    if (parent != null) {
+      List<ObjId> parentTail = parent.tail();
+      int amount = Math.min(config.parentsPerCommit() - 1, parentTail.size());
+      for (int i = 0; i < amount; i++) {
+        c.addTail(parentTail.get(i));
+      }
+
+      IndexesLogic indexesLogic = indexesLogic(persist);
```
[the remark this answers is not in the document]
snazy: Hm - not sure (yet). But we can revisit later. Both objects are short-lived.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -0,0 +1,1094 @@
...
+    // use halt of the max as the initial size for _new_ segments/splits
+    int newSegmentSize = maxSize / 2;
+
+    List<StoreIndex<CommitOp>> stripes = new ArrayList<>();
+
+    StoreIndex<CommitOp> current = newStoreIndex(COMMIT_OP_SERIALIZER);
+
+    IndexesLogic indexesLogic = indexesLogic(persist);
+
+    for (StoreIndexElement<CommitOp> el : indexesLogic.incrementalIndexFromCommit(commitObj)) {
```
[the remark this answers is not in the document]
snazy: Yup - true. Not sure whether it's worth to let the blow for the "special" case though.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
```diff
@@ -0,0 +1,535 @@
...
+  // structure are rather inefficient, when elements need to be added "in the middle" of the
+  // 'elements' j.u.ArrayList.
+
+  StoreIndexImpl(ElementSerializer<V> serializer) {
+    this(new ArrayList<>(), 1, serializer, false);
+  }
+
+  StoreIndexImpl(
+      List<StoreIndexElement<V>> elements,
+      int deserializedSize,
```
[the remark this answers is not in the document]
snazy: Depends on the point of view ;)

The field means, that the serialized index, from which this `StoreIndexImpl` object has been constructed, had that size - so its serialized size before it's been modified.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
```diff
@@ -0,0 +1,535 @@
...
+        parts,
+        size);
+    int partSize = size / parts;
+    int serializedMax = deserializedSize + estimatedSerializedSizeDiff;
+
+    List<StoreIndex<V>> result = new ArrayList<>(parts);
+    int index = 0;
+    for (int i = 0; i < parts; i++) {
+      int end = i < parts - 1 ? index + partSize : elems.size();
+      List<StoreIndexElement<V>> partElements = new ArrayList<>(elements.subList(index, end));
```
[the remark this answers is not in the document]
snazy: The real serialized size can be completely off (due to the key-prefix-encoding), so it's pretty hard to do that.
But I hope the code / object-format is flexible enough to allow such changes in the future.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
```diff
@@ -0,0 +1,535 @@
...
+      StoreIndexImpl<V> part = new StoreIndexImpl<>(partElements, serializedMax, serializer, true);
```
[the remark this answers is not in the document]
snazy: True, but it doesn't really matter much at this point. The value's rather useful _before_ splitting.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -0,0 +1,1094 @@
...
+            batch = emptyIterator();
+            next = null;
+          } else if (!b.hasNext()) {
+            CommitObj obj;
+            try {
+              obj = fetchCommit(c);
+            } catch (ObjNotFoundException e) {
+              throw new NoSuchElementException("Commit '" + c + "' not found");
+            }
+            if (obj == null) {
```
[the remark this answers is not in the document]
snazy: Updated `Persist.fetchObj*` function to not return `null`.

Just not happy with having a non-null `CommitObj` for `EMPTY_OBJ_ID`, because that would "materialize" the "nothing commit" and require us to update the logic to handle empty parent commit list.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
snazy: (Not sure I completely understood your suggesting)

Usually, the call paths load the required index stripes eagerly, based on the requested store-keys, so that lazy loading should practically never happen.

IIRC there were some rare scenarios, in which it could be theoretically possible (don't really recall which TBH).

Duplicate loads of the same obj-IDs are mitigated, if the cache layer's present.

snazy:
> I'm pretty much done with reviewing the `Persist` stuff and the various "logics". I guess I cannot really cover the whole change set in details in reasonable time :)
> 
> Is there any particular area of the code, to which you'd like me to pay closer attention, @snazy ?

I think, we've covered the most important pieces (`Persist` and other interfaces, the various logic implementations, the most relevant database specific implementations, and the `store` module).

IMO the cache and telemetry modules are worth to look at.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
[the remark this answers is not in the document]
dimas-b: Worth a comment?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
dimas-b: My suggestion is to use the same `IndexLoader` impl both for the bulk load case and for the single lazy load case.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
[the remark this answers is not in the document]
dimas-b: agreed

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/PersistProvider.java
```diff
@@ -0,0 +1,162 @@
...
+    b.setupSchema();
+
+    LOGGER.info("Creating/opening version store {} ...", versionStoreType);
+
+    BiFunction<Persist, String, Persist> wrapPersistTracing = (p, name) -> p;
+    String tracingInfo = "without tracing";
+    if (versionStoreConfig.isTracingEnabled()) {
+      if (opentelemetryTracer.isUnsatisfied()) {
+        LOGGER.warn(
+            "OpenTelemetry is enabled, but not available, forgot to add quarkus-opentelemetry? Will fall back to OpenTracing, if enabled.");
```
dimas-b, line 127: Not sure I understand... Where is the fallback to OpenTracing happening? Why OpenTracing?

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
```diff
@@ -0,0 +1,405 @@
...
+    this.persist = persist;
+    this.tracerSupplier = tracerSupplier;
+  }
+
+  @SuppressWarnings("resource")
+  Traced traced(String spanName) {
+    String repo = persist.config().repositoryId();
+    Traced traced = tracerSupplier.apply(spanName);
+    return repo != null && !repo.isEmpty()
+        ? traced.attribute("repo", persist.config().repositoryId())
```
dimas-b, line 54: nit: repo ID is already in a local var.

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
```diff
@@ -0,0 +1,405 @@
...
+  @Override
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  public Reference addReference(@Nonnull @jakarta.annotation.Nonnull Reference reference)
+      throws RefAlreadyExistsException {
+    try (Traced trace = traced("addReference")) {
+      try {
+        return persist.addReference(reference);
+      } catch (RefAlreadyExistsException e) {
+        trace.attribute("error", "already exists");
```
dimas-b, line 67: I think it's worth adding the exception to the span too... for trouble-shooting

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/Traced.java
```diff
@@ -0,0 +1,43 @@
...
+  Traced attribute(String tag, String value);
+
+  Traced attribute(String tag, boolean value);
+
+  Traced attribute(String tag, int value);
+
+  Traced attribute(String tag, long value);
+
+  default RuntimeException unhandledError(RuntimeException e) {
+    event("Unhandled error: " + e);
```
dimas-b, line 40: How about doing this too: `span.setStatus(StatusCode.ERROR).recordException(exception);
`

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
```diff
@@ -0,0 +1,405 @@
...
+        throw e;
+      } catch (RuntimeException e) {
+        throw trace.unhandledError(e);
+      }
+    }
+  }
+
+  @Override
+  public Reference fetchReference(@Nonnull @jakarta.annotation.Nonnull String name) {
+    try (Traced trace = traced("findReference")) {
```
dimas-b, line 137: nit: span name does not match API name

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
```diff
@@ -0,0 +1,405 @@
...
+          .forEach((t, c) -> trace.attribute("type." + t.name() + ".count", c));
+      try {
+        boolean[] result = persist.storeObjs(objs);
+        int successes = 0;
+        for (boolean b : result) {
+          if (b) {
+            successes++;
+          }
+        }
+        trace.attribute("result.nonNull", successes);
```
dimas-b, line 271: nit: `created.count`?

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
```diff
@@ -0,0 +1,405 @@
...
+      @SuppressWarnings("resource")
+      CloseableIterator<Obj> iter = persist.scanAllObjects(returnedObjTypes);
+      ok = true;
+      return new CloseableIterator<Obj>() {
+        @Override
+        public void close() {
+          try {
+            iter.close();
+          } finally {
+            trace.close();
```
dimas-b, line 345: I think this is not correct. Since the `span` is made "current" we should not return from this method without closing the current `scope`. The caller may close its `scope` before us and that would be a violation of OTel spec, AFAIK.

dimas-b: In this case I'd propose closing the `scope` upon returning from this method, but closing / ending the `span` when the stream is exhausted. This way any nested spans will be properly linked to this `span`, while the caller will be able to exit or start sibling spans correctly. Out `span` will become "detached" (may stretch longer than the parent span). To be strictly correct in this situation I think we ought to have a pair or "producer" and "consumer" spans - the producer would close on return, the consumer would track the lifetime of the stream... but even with one "non-current" span it might be good enough.

dimas-b: An even more correct approach would be to use two spans and activate/deactivate the inner (stream) span of every stream methods... but that would be an overkill... Do we really need to track the stream's lifetime as a span?

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/OpenTelemetryTraced.java
```diff
@@ -0,0 +1,72 @@
...
+
+final class OpenTelemetryTraced implements Traced {
+
+  private final Span span;
+  private final Scope scope;
+
+  @SuppressWarnings("MustBeClosedChecker")
+  OpenTelemetryTraced(Tracer tracer, String spanName) {
+    span = tracer.spanBuilder(spanName).startSpan();
+    scope = span.makeCurrent();
```
dimas-b, line 32: I wonder if Quarkus is able to automagically instrument the application at build time based on `@WithSpan` annotations (I did not read the docs :) ... yet)

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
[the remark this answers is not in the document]
dimas-b: fair enough :+1:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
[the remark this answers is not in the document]
dimas-b: (maybe for follow-up). As far as I understand index splitting may occur only after the stripes have been serialized. Since Deserializaion (for splitting) follows the natural key order, we know the side "before" each key. Splitting also follows the natural key order, so we could use the size marks from the (de-)serialization step for splitting by size (I hope I did not miss anything :) )

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
```diff
@@ -0,0 +1,543 @@
...
+    previousKey.clear();
+    keyBuf.position(keyPos);
+    previousKey.put(keyBuf);
+    previousKey.flip();
+
+    return previousKey;
+  }
+
+  static <V> StoreIndex<V> deserializeStoreIndex(
+      ByteBuffer serialized, ElementSerializer<V> ser, BiFunction<StoreKey, V, V> updater) {
```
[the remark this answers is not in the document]
snazy: nope, not anymore, removed

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
snazy: Hm - it's a rather maintenance-only operation, so probably not worth to add a ton of code here.

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/TelemetryPersist.java
snazy: nah - it can only happen, if the reference already exists in the `refs` table

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
```diff
@@ -0,0 +1,296 @@
...
+  @Override
+  @Nonnull
+  @jakarta.annotation.Nonnull
+  public <T extends Obj> T fetchTypedObj(
+      @Nonnull @jakarta.annotation.Nonnull ObjId id, ObjType type, Class<T> typeClass)
+      throws ObjNotFoundException {
+    Obj o = cache.get(id);
+    if (o != null) {
+      if (o.type() != type) {
+        throw new ObjNotFoundException(id);
```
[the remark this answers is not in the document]
snazy: Depends on the approach - I think of it as a "SELECT * FROM objs WHERE id = ? AND type = ?", which is what we actually actually do against some DBs - so we'd need another "SELECT" to find the actual cause.

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/OpenTelemetryTraced.java
snazy: But... this module does not depend on Quarkus

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -0,0 +1,154 @@
...
+
+  private int repositoryKey(String repositoryId) {
+    return repositoryIdMap.computeIfAbsent(repositoryId, k -> repositoryKey.incrementAndGet());
+  }
+
+  static final class CacheKey {
+
+    public static final int HEAP_OVERHEAD =
+        JAVA_OBJ_HEADER + JAVA_INT + JAVA_LONG + JAVA_OBJ_HEADER;
+    final int repoKey;
```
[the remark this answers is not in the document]
snazy: Simplified the code a bit, at the expense of a bit larger cache-key

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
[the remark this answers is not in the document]
dimas-b: Right, but does `AND type = ?` help that much? Could it be a post-select check in java code?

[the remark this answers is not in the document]
dimas-b: the full key still does not include `type`, does it?... and we do not expect (allow) more than one types obj with the same ID, right?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
snazy: the "load N" and "load single" implementations are quite different - unifying those is likely a bit of an over-generalization

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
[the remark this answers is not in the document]
snazy: What would be the benefit of a "cast exception"?
We can certainly change things later, but is it really worth now?

versioned/storage/telemetry/src/main/java/org/projectnessie/versioned/storage/telemetry/OpenTelemetryTraced.java
dimas-b: Missed that :facepalm: We can keep this with custom instrumentation and revisit later. Perhaps there's a way to run standard OTel instrumentors as a build step (they can certainly instrument in runtime, but I think build-time is preferable for Nessie in this case).

dimas-b (approved):
Great work on the new model, @snazy ! :)

PR #5644, opened 2022-12-06
Validate ID for explicitly created Namespaces

dimas-b:


ajantha-bhat (dismissed):

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRestNamespace.java
```diff
@@ -108,6 +108,7 @@ public void testNamespacesRetrieval() throws BaseNessieClientServerException {
     Namespace four = createNamespace.apply("one.two");
     for (Namespace namespace : Arrays.asList(one, two, three, four)) {
       soft.assertThat(namespace).isNotNull();
+      soft.assertThat(namespace.getId()).isNotNull();
```
snazy, line 111: Maybe also add a similar check to the getMultipleNamespaces() validating that the returned IDs stay the same?

dimas-b: `containsExactlyInAnyOrder` on line 116 uses `Namespace.equals()`, I assume, which takes ID into account.

snazy: hm - yea - you're probably right ;)

snazy: hm - yea - you're probably right ;)

dimas-b: However, I added that check to the "update" test case and rebased.

snazy (approved):

PR #5656, opened 2022-12-08
Add key range parameters to REST API v2

### LINKED ISSUE #5591: Add start/end key parameters to v2 API methods
With some incoming storage model improvements, Nessie can return results not only for a collection of fixed keys but also for key ranges.

Start/end key boundaries are meaningful in these `TreeApi` v2 methods:
* `getEntries`
* `getDiff`

Potentially later:
* `getReferences` (?)

Likely not happening soon:
* `getSeveralContents`
* `getMultipleContents`

dimas-b:
* In "diff" API

* In "get entries"

* New "get contents" endpoint for range queries

This is just an API-level change to reserve REST paths and query 
parameters. Server-side implementation to be done separately.

Closes #5591

### DIFF
--- a/model/src/main/java/org/projectnessie/api/v2/TreeApi.java
+++ b/model/src/main/java/org/projectnessie/api/v2/TreeApi.java
@@ -256,6 +256,27 @@ public interface TreeApi {
       throws NessieNotFoundException;
 
   /**
+   * Similar to {@link #getContent(ContentKey, String)}, but takes min and max {@link ContentKey}s
+   * and returns the {@link Content} objects having their keys in that range in a named-reference (a
+   * {@link org.projectnessie.model.Branch} or {@link org.projectnessie.model.Tag}).
+   *
+   * @param ref named-reference to retrieve the content for
+   * @param minKey the lower bound of the content key range to retrieve (inclusive).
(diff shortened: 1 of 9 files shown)


dimas-b:
@snazy : do you want range params for reference endpoints too?

snazy:
> @snazy : do you want range params for reference endpoints too?

Yes, that would be good. It would also be nicer to the new storage stuff to change the `maxKey` parameter from _exclusive_ to _inclusive. We can technically encode a prefix-query with the min and max parameters:

* _from...to_ : `minKey` defines the start (inclusive) and `maxKey` the end (inclusive)
* _from_ : `minKey` defines the start (inclusive)
* _to_ : `maxKey` defines the end (inclusive)
* _prefix_ : `minKey == maxKey` define the _prefix_ for the returned keys. "min equal to max" doesn't really make sense for a query and should be performed via a point-query (get-content) - so we _could_ abuse this case. From a usability perspective if would probably make sense to have a separate parameter for this though.

snazy:
Shall we continue on this one?
We can omit `New "get contents" endpoint for range queries` for now, but have range-queries for `getAllReferences`

snazy:
Closing with its successor #6743

PR #5665, opened 2022-12-08
Fix OpenAPI docs for the required "expected hash" parameters

### LINKED ISSUE #5465: Update OpenAPI spec doc to make expectedHash as required in certain APIs
**What happened**: 
While trying to use merge branch or delete branch API call, API spec says expectedHash is optional, not required. So I tried passing it as optional and API threw error.
{
    "status": 400,
    "reason": "Bad Request",
    "message": "deleteReference.expectedHash: Hash must consist of the hex representation of 4-32 bytes",
    "errorCode": "BAD_REQUEST",
    "serverStackTrace": null
}
image (1)

API shouldn't have thrown error. It should have worked as documented.
or Doc should have mentioned it as required parameter instead of Optional

Try calling the API using any tools such as POSTMAN:
DELETE /trees/branch/branch1 

It will throw below error:
{
    "status": 400,
    "reason": "Bad Request",
    "message": "deleteReference.expectedHash: Hash must consist of the hex representation of 4-32 bytes",
    "errorCode": "BAD_REQUEST",
    "serverStackTrace": null
}

**Details**:
- Nessie server type (docker/built from source) and version: 0.44.0

dimas-b:
* Add `required` annotation flags in V1

* Update reference format descriptions in V2. V2 does not have a separate parameter for the expected hash.

* This is just a spec/doc level change. Existing behaviour is not affected.

Fixes #5465


model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -84,19 +93,14 @@ public interface ApiDoc {
...
           + "- name@hash - Identifies the 'hash' commit on a branch or tag.\n"
           + "\n"
-          + "If both 'name' and 'hash' are given, 'hash' must be the current HEAD of the branch or tag. It will be "
-          + "used to validate that at execution time the reference points to the same hash that the caller expected "
-          + "when the operation arguments were constructed.\n"
-          + "\n"
-          + "Not specifying the 'hash' value relaxes server-side checks and can lead to unexpected side effects if "
-          + "multiple changes to the same reference are executed concurrently. It is recommended to always specify the "
-          + "'hash' value when assigning or deleting a reference.\n";
+          + "The specified 'hash' must be the current HEAD of the branch or tag. It will be used to validate that "
```
snazy, line 101:
```suggestion
          + "The specified 'hash' must be the value of the current HEAD of the branch or tag known by the client. It will be used to validate that "
```

model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -58,6 +58,15 @@ public interface ApiDoc {
           + "\n"
           + FULL_REF_INFO;
 
+  String CHECKED_BRANCH_DESCRIPTION =
+      "A reference to a specific version of the contents tree (a point in history) on a branch.\n"
+          + "This reference is specified in this form:\n"
+          + "- name@hash - Identifies the 'hash' commit on the named branch.\n"
+          + "\n"
+          + "The 'hash' commit must be reachable from the current HEAD of the branch.\n"
+          + "In this case 'hash' indicates the state of contents that should be used for validating incoming changes\n"
```
snazy, line 67:
```suggestion
          + "In this case 'hash' indicates the state of contents used for validating incoming changes\n"
```

model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -58,6 +58,15 @@ public interface ApiDoc {
...
+          + "(commits / merges / transplants).\n";
```
snazy, line 68:
```suggestion
          + "(commits / merges / transplants). This is usually the value of the HEAD of the branch known by the client.\n";
```

dimas-b: Not sure about the "usually" part. I think it's a common use case to commit with a reference to an old hash that was retrieved at the start of a table change process (Iceberg transaction). As long as there are not conflicts the hash can be quite old.

I'll rephrase that it should be as recent as reasonably possible from the client's perspective. WDYT?

snazy: "recent" in the client's sense, yes

dimas-b: rephrased and rebased

snazy (approved):

PR #5668, opened 2022-12-09
Add README.md to `model`

dimas-b:
... with API v2 highlights

### DIFF
--- a/model/README.md
+++ b/model/README.md
@@ -1,0 +1,77 @@
+# Nessie Model and API
+
+The `model` module contains java class that define the user-visible model for data exchanging data between Nessie
+clients and servers, as well as corresponding APIs.
+
+Note: not all classes in this module are meant to be referenced by external client-side code. See the following
+sections for details.
+
+# Nessie Model
+
+The java classes in the `org.projectnessie.model` and `org.projectnessie.error` packages are meant to be used
+by java clients to communicate with Nessie Servers. These classes can be used to compile custom client-side code.
+
+The utilities provided by the `clients/client` module can help building Nessie clients in `java`.
+
+# Nessie API
+
+The java classes in the `org.projectnessie.api` packages represent the Nessie API conceptually. These classes are
+used to generate the Nessie OpenAPI specification.
+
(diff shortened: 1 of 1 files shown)


.idea/codeStyles/codeStyleConfig.xml
```diff
@@ -1,5 +1,5 @@
 <component name="ProjectCodeStyleConfiguration">
   <state>
-    <option name="PREFERRED_PROJECT_CODE_STYLE" value="GoogleStyle" />
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="dremio" />
```
ajantha-bhat, line 3: nit: Can it be a separate PR? As the changes are independent. 

Also, I remember spotless uses GoogleStyle. So, it makes sense for Idea also to use GoogleStyle?

dimas-b: oops - my mistake - removed this change from PR. Thanks for spotting! :+1:

model/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+However, `org.projectnessie.api` classes are not expected to be used for compiling client-side code outside the scope
+of this repository.
+
+Clients built in `java` should build against the API of the `clients/client` module.
+
+Clients build in other languages can generate language-specific bindings from the published OpenAPI specification of
+the Nessie API.
+
+The published version of the Nessie OpenAPI specification is available on [SwaggerHub](https://app.swaggerhub.com/apis/projectnessie/nessie).
```
ajantha-bhat, line 29: This still points to v1 and in the next release it will be updated to V2 right?
https://app.swaggerhub.com/apis/projectnessie/nessie/0.45.0

ajantha-bhat (approved):
Overall LGTM. 

Just a question about Swagger hub API version.

model/README.md
dimas-b: v2 OpenAPI has not been released yet. The next release will have both v1 and v2 in Nessie's OpenAPI yaml.

ajantha-bhat: I thought so. Thanks for the clarification.

snazy (approved):

PR #5670, opened 2022-12-09
Expose updated reference data in java client APIs

### LINKED ISSUE #5595: Return updated Reference data from java client methods that modify references
When a reference is creates/changed, it would be nice to return the updated data from the related java client APIs.

dimas-b:
For assign and delete operations.

This is supported only in API v2 because API v1 does not return the required information from the server.

Fixes #5595


snazy (dismissed):

dimas-b:
rebased and fixed conflicts

snazy (approved):

PR #5682, opened 2022-12-12
Expose MergeBehavior to java clients

### LINKED ISSUE #5415: Expose MergeBehavior to java clients
`MergeBehavior` was added to the REST API under #3834.

It makes sense to add helper methods to the java client for the related new merge/transplant parameters.

Tests should be added to `AbstractCompatibilityTests`.

dimas-b:
Pure REST clients already have the ability to specify per-key merge behaviour. This change exposes the same functionality to java clients.

Available in both API v1 and v2.

Fixes #5415


snazy (approved):
The change LGTM.

However, I doubt that we should keep `MergeResponse` as it currently is. I've opened #5689 for that.

PR #5686, opened 2022-12-13
Support merge/transplant message overrides

### LINKED ISSUE #4430: Review and update commit information (e.g. message) created for merge + transplant operations
See discussion in https://github.com/projectnessie/nessie/pull/4333#issuecomment-1148805078

dimas-b:
Commit message may be overridden when the operation results in exactly one commit on the target branch.

That is overriding messages in unsquashed transplants is not supported.

Merge message overrides are supported only in API v2 at this point because v1 does not have the commit message in merge HTTP payloads.

Closes #4430

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRestMergeTransplant.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRestMergeTransplant.java
@@ -22,6 +22,7 @@ import static org.assertj.core.api.InstanceOfAssertFactories.map;
 import static org.assertj.core.data.MapEntry.entry;
 
 import com.google.common.collect.ImmutableList;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
 import org.assertj.core.api.InstanceOfAssertFactories;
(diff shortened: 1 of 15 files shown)


servers/services/src/main/java/org/projectnessie/services/impl/BaseApiImpl.java
```diff
@@ -155,13 +161,21 @@ public CommitMeta squash(List<CommitMeta> metadata) {
...
                 .authorTime(now);
         StringBuilder newMessage = new StringBuilder();
+
+        if (messageOverride != null) {
+          newMessage.append(messageOverride);
+        }
+
         Map<String, String> newProperties = new HashMap<>();
         for (CommitMeta commitMeta : metadata) {
           newProperties.putAll(commitMeta.getProperties());
```
snazy, line 171: I think, we should skip this step for v2 as well, actually skipping the whole loop for v2, eventually changing the above `if` to 
```
if (messageOverride != null) { return newMeta.message(messageOverride).build(); }
```

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -425,13 +425,19 @@ public MergeResponse transplantCommitsIntoBranch(
         transplants = s.collect(Collectors.toList());
       }
 
+      if (Boolean.TRUE.equals(keepIndividualCommits) && transplants.size() > 1) {
+        // Message overrides are not meaningful when transplanting more than one commit.
```
snazy, line 429: Note: I don't mind to change the the default message(s) in a Nessie release. The current default message isn't great, and it's not a documented behavior.

snazy (dismissed):

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
snazy: Since the commit messages (at least those from Iceberg) are usually somewhat meaningless, i'm not sure it's worth to keep even those.

servers/services/src/main/java/org/projectnessie/services/impl/BaseApiImpl.java
dimas-b: yes, but for now I'm going to merge just the message changes :)

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: good point. #5694

dimas-b:
rebased and fixed conflicts

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRestMergeTransplant.java
```diff
@@ -364,6 +365,126 @@ private void conflictExceptionReturnedAsMergeResult(
...
+                .returnConflictAsResult(returnConflictAsResult)
+                .transplant(),
+        // Note: the expected messages are given in the commit log order (newest to oldest)
+        ImmutableList.of("test-commit-2", "test-commit-1"));
+  }
+
+  private void testMergeTransplantMessage(
+      MergeTransplantActor actor, Collection<String> expectedMessages)
+      throws BaseNessieClientServerException {
+    Branch target = createBranch("merge-transplant-msg-target");
```
snazy, line 456: Please please add a common ancestor commit for this test

dimas-b: Will do a fix here. Also: https://github.com/projectnessie/nessie/pull/5696

dimas-b: fixed

snazy: thanks!

snazy (approved):

PR #5691, opened 2022-12-14
Properly exclude junit:junit

snazy:
The old junit dependency sneaks as an (unfortunately) required compile dependency via testcontainers. But as soon the old junit dependency's on the class path, assertj uses the old assertion- exception, which isn't recognized by junit 5. This was worked around with the `JUnit5Assumptions` configuration for assertj, but that does not work in Quarkus tests, thanks to class loader "magic"...

This change leaves junit4 as a compile-only dependency, but excludes it at runtime.


dimas-b (approved):
Thanks for sorting this out! :tada:

snazy:
Aw - bummer. Totally ignored, that `TestRule` is an interface, which is needed during runtime as well ....

snazy:
GenericContainer implements TestRule

dimas-b:
We could just add a fake `TestRule` interface to make the JVM happy ;)

PR #5701, opened 2022-12-15
Allow publishing Gradle build scans from CI

snazy:
But not automatically for every build to protect sensitve workflows, like releasw workflows that have access to sensitive secrets. Gradle build scans are available for free (Gradle Enterprise is not required). The scans help identifying build issues, dependency issues, build performance issues. Also interesting: it shows the result of each test of a CI run.

Build scans are not published for every run - especially sensitive workflows (release workflows w/ access to secrets) are _not_ scanned/published.

Links to the build scans are available from the _Summary_ page of the workflow runs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,18 +42,18 @@ jobs:
     - name: Gradle / compile
       uses: gradle/gradle-build-action@v2
       with:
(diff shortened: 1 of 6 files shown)


dimas-b (approved):
nice feature :)

PR #5706, opened 2022-12-15
Add v2 attributes to `CommitMeta`

### LINKED ISSUE #5424: Enhance `CommitMeta` in API v2
From https://github.com/projectnessie/nessie/pull/5004#discussion_r981049824 : 

We should change the CommitMeta for v2 to be more generic, so these attributes:

* commit message
* map of property name to list of property values
* commit ID
* list of parent commit IDs
* Pretty much what Git does. With that, there could be for example a list of authors (think: squashing & merging).

dimas-b:
Support multiple authors and "signed off by" names.

Add explicit parent hash getter.

Support list properties.

Support deserializing old v1 JSON form in all contexts.

Generate v1 (old) JSON in API v1 responses.

Generate v2 (new) JSON in API v2 responses.

Use v1 JSON for storage. We can switch to v2 JSON in storage in the next releases to make older versions capable of reading metadata JSON during rolling upgrades.

This means that the new `CommitMeta` capabilities are added to the API, but are not yet supported by servers.

Note: `@JsonView` annotations have to be both on the HTTP interface and the impl. class to allow both Quarkus and Jersey servers to work properly. Quarkus appears to see the annotation the interface, which Jersey appears to see the annotation on the impl. class.

Closes #5424

### DIFF
--- a/model/src/main/java/org/projectnessie/model/CommitMeta.java
+++ b/model/src/main/java/org/projectnessie/model/CommitMeta.java
@@ -29,6 +29,8 @@ import com.fasterxml.jackson.databind.ser.std.StdSerializer;
 import java.io.IOException;
 import java.time.Instant;
 import java.time.format.DateTimeFormatter;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import javax.annotation.Nullable;
@@ -141,12 +143,23 @@ public abstract class CommitMeta {
    * hostnames etc.
    */
   @NotNull
-  public abstract Map<String, String> getProperties();
+  @Value.Derived
+  @JsonView(ApiAttributesV1.class)
+  public Map<String, String> getProperties() {
+    HashMap<String, String> firstElements = new HashMap<>();
+    for (Map.Entry<String, List<String>> entry : getAllProperties().entrySet()) {
+      List<String> list = entry.getValue();
+      if (!list.isEmpty()) {
+        firstElements.put(entry.getKey(), list.get(0));
+      }
+    }
+    return firstElements;
+  }
 
   @NotNull
   @JsonView(ApiAttributesV2.class)
   @JsonInclude(JsonInclude.Include.NON_EMPTY)
-  public abstract Map<String, List<String>> getListProperties();
+  public abstract Map<String, List<String>> getAllProperties();
 
   @NotNull
   @JsonView(ApiAttributesV2.class)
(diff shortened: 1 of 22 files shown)


model/src/main/java/org/projectnessie/model/CommitMeta.java
```diff
@@ -170,4 +203,16 @@ public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOE
       return Instant.parse(p.getText());
     }
   }
+
+  private static final class CoauthorsList {
```
snazy, line 207: Nit: rename

snazy (commented):
The apporach LGTM. I suspect the boilerplate in the Rest*TreeResource classes is just technically necessary.

model/src/main/java/org/projectnessie/model/CommitMeta.java
dimas-b: Suggestion? :)

dimas-b:
> I suspect the boilerplate in the Rest*TreeResource classes is just technically necessary.

This is apparently required to run in Jersey, which does not recognize annotations on interface methods.

model/src/main/java/org/projectnessie/model/CommitMeta.java
dimas-b: I may steer is the direction of deriving `getAuthor()` from the list, so this may disappear :)

versioned/transfer/src/test/java/org/projectnessie/versioned/transfer/AbstractExportImport.java
```diff
@@ -127,7 +127,7 @@ public void manyExportFiles(@TempDir Path targetDir) throws Exception {
           .exportNessieRepository();
     }
 
-    int expectedCommitsFileCount = 126;
+    int expectedCommitsFileCount = 144;
```
dimas-b, line 130: the total size is bigger due to extra symbols in `CommitMeta` JSON, specifically the `parentCommitHashes` attribute.

dimas-b: fix in the latest commit

dimas-b:
Squashed. One minor change remains for list properties, but otherwise the PR should be reviewable now.

snazy (approved):
LGTM! The JsonView stuff looks really nice.
Just a few comments for follow-ups.

model/src/test/java/org/projectnessie/api/v1/params/TestParamObjectsSerialization.java
```diff
@@ -35,6 +35,7 @@ public class TestParamObjectsSerialization extends TestModelObjectsSerialization
   static List<Case> goodCases() {
     final String branchName = "testBranch";
 
+    // TODO: add test for CommitMeta v1
```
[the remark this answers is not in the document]
dimas-b: I think I meant v2... v1 serialization is tested by existing model serializations tests... I guess this comment is actually in the wrong place... also, general v2 serialization is covered by the "resteasy" and "deserializer" tests in this PR, but yes, I'll add more specific tests in a follow-up PR.

versioned/spi/src/main/java/org/projectnessie/versioned/CommitMetaSerializer.java
```diff
@@ -32,7 +33,9 @@ public class CommitMetaSerializer implements Serializer<CommitMeta> {
   @Override
   public ByteString toBytes(CommitMeta value) {
     try (Output out = ByteString.newOutput()) {
-      MAPPER.writeValue(out, value);
+      // Store commit metadata using v2 format. This is mostly to avoid duplicate data from v1
+      // attributes in the serialized form.
+      MAPPER.writerWithView(ApiAttributesV1.class).writeValue(out, value);
```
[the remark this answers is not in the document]
dimas-b: I fixed this comment (and removed the one above). Since it's a comment-only change I'll merge once CI passes.

PR #5731, opened 2022-12-20
Add serialized form tests for `CommitMeta` v2

dimas-b:
Following up on #5706


dalaro (approved):
Reviewing as requested.  I'm not familiar with the context and have no deep insight here, but the narrow change -- generalizing `TestModelObjectsSerialization` to have one case for V1 and another for V2 -- LGTM.

PR #5732, opened 2022-12-20
Make generated OpenAPI type names more readable

dimas-b:
Rename JSON view classes and make them inner classes with short and readable names.

If the view class is an inner classes, the OpenAPI generator apparently uses its simple name as a suffix for related type names in the YAML.

Also move view classes under the `model` package since they are referenced from model classes.

Following up on #5424

----
Example type name before: `Branch_ApiAttributesV1`, after: `Branch_V1`.


versioned/spi/src/test/java/org/projectnessie/versioned/TestCommitMetaSerializer.java
```diff
@@ -42,9 +42,7 @@ void testCommitSerde() throws JsonProcessingException {
 
     ByteString expectedBytes =
         ByteString.copyFrom(
-            new ObjectMapper()
-                .writerWithView(ApiAttributesV1.class)
-                .writeValueAsBytes(expectedCommit));
+            new ObjectMapper().writerWithView(Views.V1.class).writeValueAsBytes(expectedCommit));
```
dalaro, line 45: Hi, I'll do my best to review as requested, though I'd suggest assigning low value to my input.  I have no opinion on the specific name changes (ApiAttributesV1/2 -> Views.V1/V2).  My sole question as about Jackson views themselves.  My naive and possibly misinformed understanding is that Jackson views are a runtime-only concept, and view names/classes don't enter serialized representations in any way, so that this change has no compat implications.  Is that right?

I'm tentatively approving as long as that's true, along with the caveat to give my input low weight here.

dalaro (approved):

versioned/spi/src/test/java/org/projectnessie/versioned/TestCommitMetaSerializer.java
dimas-b: Thanks for the rewview, @dalaro !

Correct. View classes do not appear in the JSON serialized form. However, their names affect the generated OpenAPI type definitions. It's the latter part that this PR attempts to improve.

I suppose the OpenAPI type names are likely going to affect the generated clients for UI, for example, which I why I wanted to make then more human-readable.

This PR does not change actual JSON in REST API input/output.

dimas-b:
rebased and fixed conflicts

PR #5740, opened 2022-12-23
Add validation annotations to backend service interfaces

dimas-b:
Also relax test checks to account to differences in validation parameter and method names.

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
+++ b/servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
@@ -18,6 +18,10 @@ package org.projectnessie.services.spi;
 import java.util.Collection;
 import java.util.List;
 import javax.annotation.Nullable;
+import javax.validation.Valid;
+import javax.validation.constraints.NotBlank;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 import org.projectnessie.error.NessieConflictException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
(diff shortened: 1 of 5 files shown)


ajantha-bhat:
please rebase or re-trigger the build. Python failures are been fixed in #5749

ajantha-bhat:
closing and re-opening the PR to re-trigger the build.

servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
```diff
@@ -41,68 +46,126 @@ public interface TreeService {
...
+          String branchName,
+      @Valid @NotNull @Pattern(regexp = Validation.HASH_REGEX, message = Validation.HASH_MESSAGE)
+          String expectedHash,
       String message,
       List<String> hashesToTransplant,
-      String fromRefName,
+      @Valid
+          @NotBlank
+          @NotNull
+          @Pattern(regexp = Validation.REF_NAME_REGEX, message = Validation.REF_NAME_MESSAGE)
```
ajantha-bhat, line 119: a) does `REF_NAME_REGEX` handle the blanks too? Do we need this to be `NotBlank` or just `NotNull` is enough?

b) do we need validations for `hashesToTransplant` too?

dimas-b: I used the same validation rules as in v1 `HttpTreeApi` and related classes.

`NotBlank` may actually produce a more intuitive error message,

dimas-b: We could validate `hashesToTransplant`, but we do not do that in API v1 and did not want to change behaviour in this PR.

ajantha-bhat: I thought it is freshly added validations. Ok to be inline with v1 as of now and improve in the follow up.

ajantha-bhat (approved):

PR #5793, opened 2023-01-09
Nit: Suppress ClassCanBeStatic warnings

snazy:


XN137 (approved):
lgtm

justFYI: there is an errorprone PR that will do this out of the box in future versions

PR #5799, opened 2023-01-09
Proper Path handling in GC tool CLI

snazy:


gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/BaseRepositoryCommand.java
```diff
@@ -140,16 +141,22 @@ protected LiveContentSet identify(
       CommandSpec commandSpec) {
 
     // Sanity check, whether that file can be written
-    if (markOptions.getLiveSetIdFile() != null) {
+    Path liveSetIdFile = markOptions.getLiveSetIdFile();
+    if (liveSetIdFile != null) {
       try {
-        Files.createDirectories(markOptions.getLiveSetIdFile().getParent());
-        Files.write(markOptions.getLiveSetIdFile(), new byte[0]);
+        Files.createDirectories(liveSetIdFile.getParent());
```
dimas-b, line 147: Just wondering how does splitting the `try` blocks help? Only one of the calls will throw and we do not apparently validate specific failure cases (messages) in tests.

dimas-b (approved):
LGTM with minor comment

gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/BaseRepositoryCommand.java
snazy: B/c testing on macOS + Windows + Linux also lead to very different error messages and exception types.

PR #5705, opened 2022-12-15
CI: cross-check macOS + Windows

snazy:
Adds CI jobs to test Nessie on macOS and Windows (non-WSL/2). Unfortunately, both testing on macOS and Windows is far from "perfect" in the sense that all tests run on both platforms.
    
Setting up the GH runners takes a couple extra steps (setting up podman for macOS, setting up a local Hadoop on Windows), which can take some time.
    
The issues are:
* Docker on Windows (non-WSL) is not supported by testcontainers, so the tests that need testcontainers are disabled on Windows.
* testcontainers against podman (required on macOS) has issues inquiring information about the _fetched_ (locally present) image (workaround in the Nessie code base).
* Probably not related to macOS or Windows, but the npm-tests in `:nessie-ui` seem to be flaky. This podman & nessie-ui issue may require re-running CI jobs.
* Build times seem to vary a lot, from "normal" times as used w/ the Linux based runners to magnitudes of that.
    
The workflows shall run daily against `main`, on demand against `main` and against PRs having the `pr-macos-win` label.

PR #5803, opened 2023-01-09
Prevent using the Minio Extension on non-Linux OS

snazy:
Background is that S3 URLs encode the bucket name in front of the S3 target's FQDN, which works fine for `bucket-name.localhost` on Linux, but not on other OS'es. S3 requests encode the bucket name in the DNS FQDN - e.g. `my-bucket.my-minio.com`, which translates to `my-bucket.localhost` for local testing. Linux supports these wildcards out of the box, but macOS + Windows do not.

### DIFF
--- a/testing/s3minio/src/main/java/org/projectnessie/minio/MinioExtension.java
+++ b/testing/s3minio/src/main/java/org/projectnessie/minio/MinioExtension.java
@@ -15,12 +15,18 @@
  */
 package org.projectnessie.minio;
 
+import static java.lang.String.format;
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.disabled;
+import static org.junit.jupiter.api.extension.ConditionEvaluationResult.enabled;
 import static org.junit.platform.commons.util.AnnotationUtils.findAnnotatedFields;
 import static org.junit.platform.commons.util.ReflectionUtils.makeAccessible;
 
 import java.lang.reflect.Field;
+import org.junit.jupiter.api.condition.OS;
 import org.junit.jupiter.api.extension.BeforeAllCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.jupiter.api.extension.ParameterContext;
 import org.junit.jupiter.api.extension.ParameterResolutionException;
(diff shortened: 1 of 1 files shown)


testing/s3minio/src/main/java/org/projectnessie/minio/MinioExtension.java
```diff
@@ -35,10 +40,20 @@
...
   private static final ExtensionContext.Namespace NAMESPACE =
       ExtensionContext.Namespace.create(MinioExtension.class);
 
+  @Override
+  public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+    if (OS.current() == OS.LINUX) {
+      return enabled("Running on Linux");
+    }
+    return disabled(
+        "Not running on Linux, because neither macOS nor Windows support wildcard localhost FQDNs.");
```
dimas-b, line 54: Hmmm... I'm not sure I follow the meaning of this message :thinking:

snazy: `foobar.localhost` cannot be resolved

dimas-b: If we disable Minio tests on Mac+Win, how could we make sure our S3-related code still works there?

dimas-b: (to clarify: I'm fine with this change as long as we fix the message, but I wonder if we actually need some other test env. for S3 on Mac/Win)

snazy: They already do not work on those platforms, this change only let's CI pass on those platforms (see #5705).
The _mocked_ S3 impl is still exercised - Minio is not.

dimas-b: Could we add a local host wildcards in a CI action?

snazy: Too complicated and doesn't work - already tried.

dimas-b: ok

dimas-b: Could you just fix the "disabled" message? It still looks awkward :)

snazy: It's a wildcard fully-qualified-domain-name?

dimas-b: No, why `not running on Linux` when the OS is _not_ Linux?

dimas-b: I'd say `Disabled on a non-Lixus OS, because....`

dimas-b: I tend to interpret the disabled message as belonging to the test :) so in my mind  `the test is` `not running on Linux, becase...` looks odd :)

snazy: Proposal?

dimas-b (approved):

PR #5805, opened 2023-01-09
Disable compatibility tests on macOS

snazy:
Fails on macOS with `libc++abi: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`

### DIFF
--- a/compatibility/compatibility-tests/build.gradle.kts
+++ b/compatibility/compatibility-tests/build.gradle.kts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


compatibility/compatibility-tests/build.gradle.kts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import org.apache.tools.ant.taskdefs.condition.Os
```
dimas-b, line 17: ant?

snazy: yes

PR #5830, opened 2023-01-11
Actually assign `main` in AbstractContentGeneratorTest

dimas-b:


tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/AbstractContentGeneratorTest.java
```diff
@@ -53,7 +53,7 @@ public class AbstractContentGeneratorTest {
   void emptyRepo() throws Exception {
     try (NessieApiV1 api = buildNessieApi()) {
       Branch defaultBranch = api.getDefaultBranch();
-      api.assignBranch().branch(defaultBranch).assignTo(Detached.of(NO_ANCESTOR));
+      api.assignBranch().branch(defaultBranch).assignTo(Detached.of(NO_ANCESTOR)).assign();
```
XN137, line 56: makes me wonder if our api methods that only build but not yet execute an operation should be annotated with:
https://errorprone.info/bugpattern/CheckReturnValue

dimas-b: Yes, that'd be nice.

I had an IJ rule for that at some point, but it proved difficult to maintain to match API changes. An annotation would work better, I think.

PR #5831, opened 2023-01-11
Testcontainers: retry container launch

snazy:
Testcontainers against podman (required on macOS) has issues inquiring information about the _fetched_ (locally present) image. The built-in retry mechanism in testcontainers does not properly handle the situation. This change adds a retry mechanism when the image-fetch failed.

### DIFF
--- a/versioned/persist/tx-test/src/main/java/org/projectnessie/versioned/persist/tx/postgres/ContainerTestConnectionProviderSource.java
+++ b/versioned/persist/tx-test/src/main/java/org/projectnessie/versioned/persist/tx/postgres/ContainerTestConnectionProviderSource.java
@@ -18,6 +18,8 @@ package org.projectnessie.versioned.persist.tx.postgres;
 import org.projectnessie.versioned.persist.tx.local.GenericJdbcTestConnectionProviderSource;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.ContainerFetchException;
+import org.testcontainers.containers.ContainerLaunchException;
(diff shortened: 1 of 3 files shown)


versioned/persist/tx-test/src/main/java/org/projectnessie/versioned/persist/tx/postgres/ContainerTestConnectionProviderSource.java
```diff
@@ -34,9 +36,23 @@ public void start() throws Exception {
...
-    container.start();
+    for (int retry = 0; ; retry++) {
+      container =
+          createContainer().withLogConsumer(new Slf4jLogConsumer(LOGGER)).withStartupAttempts(5);
+      try {
+        container.start();
+        break;
+      } catch (ContainerLaunchException e) {
+        container.close();
+        if (e.getCause() != null && e.getCause() instanceof ContainerFetchException && retry < 3) {
```
dimas-b, line 47: Could we factor this code into a utility method and pass a lambda parameter from specific source implementations?

snazy: The project don't have much in common that would be suitable for a utility class. Not sure, whether it's worth to add a new project just for this one.

dimas-b: Meant the catch/retry logic, but the copy-paste is not terrible so +1

dimas-b (approved):

PR #5879, opened 2023-01-17
Add content-ID to REST API v2 entries endpoint

### LINKED ISSUE #5878: Add content ID to `EntriesResponse` in API v2
Content ID can be used to distinguish entities with the same name (same key) in different commits (branches, tags).

Having content ID in `EntriesResponse` can help to avoid an extra call to the `getContent` API.

snazy:
Fixes #5878

### DIFF
--- a/model/src/main/java/org/projectnessie/model/EntriesResponse.java
+++ b/model/src/main/java/org/projectnessie/model/EntriesResponse.java
@@ -15,11 +15,14 @@
  */
 package org.projectnessie.model;
 
+import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.List;
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 import org.immutables.value.Value;
+import org.projectnessie.model.ser.Views;
 
 @Value.Immutable
 @JsonSerialize(as = ImmutableEntriesResponse.class)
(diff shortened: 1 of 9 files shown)


model/src/main/java/org/projectnessie/model/EntriesResponse.java
```diff
@@ -50,8 +53,13 @@ static ImmutableEntry.Builder builder() {
...
     ContentKey getName();
 
-    static Entry entry(ContentKey name, Content.Type type) {
-      return ImmutableEntry.of(name, type);
+    @JsonView(Views.V2.class)
+    @Value.Parameter(order = 3)
+    @Nullable // for V1 backwards compatibility
+    String getContentId();
+
+    static Entry entry(ContentKey name, Content.Type type, String contentId) {
```
dimas-b, line 61: Would it make sense to keep the old override in case some old client rely on it? The new attribute is nullable, so we can maintain a compatible implementation of the old `entry` method.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractResteasyV1Test.java
```diff
@@ -121,6 +122,17 @@ public void testBasic() {
...
+    EntriesResponse entries =
+        rest()
+            .get("trees/tree/{branch}/entries", newReference.getName())
+            .then()
+            .statusCode(200)
+            .extract()
+            .as(EntriesResponse.class);
+    assertThat(entries.getEntries())
+        .hasSize(1)
+        .allSatisfy(e -> assertThat(e.getContentId()).isNull());
```
dimas-b, line 134: Do you think a plain JSON string test like `testCommitMetaAttributes` for the new content ID attribute is worth adding (in addition to this test, or replacing this test)? My thinking is that java code blurs the case of having a `null` attribue and not having an attribute.

dimas-b (commented):
Thanks for the quick fix. LGTM overall. Minor comments.

model/src/main/java/org/projectnessie/model/EntriesResponse.java
snazy: not sure whether clients have a reason to call this method, but good point.

snazy:
Hum - something's broken :(

snazy:
Damn - if `namespaceDepth > 0` is specified, the "flattening" doesn't work ATM.

snazy:
> Damn - if `namespaceDepth > 0` is specified, the "flattening" doesn't work ATM.

Ah - not an issue at all - that's only used from the namespace API, which doesn't get the content-ID anyway - so all good :)

dimas-b:
> if namespaceDepth > 0 is specified [...]

`namespaceDepth` is not supported in API v2... Does this help?

dimas-b (approved):

PR #5880, opened 2023-01-17
Update V2 commit-response to include generated content-IDs

snazy:
This changes `org.projectnessie.model.CommitResponse` to include the content-IDs of the _newly_ added `Content` objects - for REST API v2.

The Nessie API requires people to _not_ set the `Content.id` for _new_ content (think: `CREATE TABLE`), Nessie will generate the content-ID in that case. This behavior will be strictly enforced with #5641. So this change adds a new `List<AddedContent>` to `CommitResponse`, which contains the generated content-IDs, so use cases that need the content-ID can just inspect the `CommitResponse` and do not have to issue a `getContents()` to get the generated content-ID.

### DIFF
--- a/versioned/spi/src/test/java/org/projectnessie/versioned/TestMetricsVersionStore.java
+++ b/versioned/spi/src/test/java/org/projectnessie/versioned/TestMetricsVersionStore.java
@@ -132,7 +132,8 @@ class TestMetricsVersionStore {
                         Optional.empty(),
                         CommitMeta.fromMessage("metadata"),
                         Collections.emptyList(),
-                        () -> null),
+                        () -> null,
+                        (k, c) -> {}),
                 () -> Hash.of("cafebabedeadbeef"),
                 refNotFoundAndRefConflictThrows),
             new VersionStoreInvocation<>(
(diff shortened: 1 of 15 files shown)


versioned/spi/src/test/java/org/projectnessie/versioned/TestMetricsVersionStore.java
```diff
@@ -132,7 +132,8 @@ public CommitMeta squash(List<CommitMeta> metadata) {
                         Optional.empty(),
                         CommitMeta.fromMessage("metadata"),
                         Collections.emptyList(),
-                        () -> null),
+                        () -> null,
+                        (k, c) -> {}),
```
dimas-b, line 136: nit: is this necessary? Could we use the default (old) `commit` method?

versioned/spi/src/test/java/org/projectnessie/versioned/TestTracingVersionStore.java
```diff
@@ -121,7 +121,8 @@ public CommitMeta squash(List<CommitMeta> metadata) {
                             Optional.empty(),
                             CommitMeta.fromMessage("metadata"),
                             Collections.emptyList(),
-                            () -> null),
+                            () -> null,
+                            (k, c) -> {}),
```
dimas-b, line 125: nit: is this necessary? Could we use the default (old) `commit` method?

dimas-b (approved):

versioned/spi/src/test/java/org/projectnessie/versioned/TestMetricsVersionStore.java
snazy: IIRC that was necessary that time (to use the non-default convenience override)

PR #5905, opened 2023-01-19
Return latest state of deleted reference

### LINKED ISSUE #5772: Make TreeService.deleteReference() return the latest state of the deleted reference
API v2 requires returning the latest state of the deleted reference to the client.

Current impl. achieves that by alway validating that the HEAD is strictly equal to the expected HEAD of the reference. 

However, `DatabaseAdapter`s allow deleting references without specifying the expected hash. 

In order to make REST API consistent with `DatabaseAdapter` SPI, `TreeService.deleteReference()` should return the latest reference hash from the adapter layer.

snazy:
Fixes #5772

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
+++ b/servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
@@ -34,6 +34,7 @@ import org.projectnessie.model.MergeKeyBehavior;
 import org.projectnessie.model.MergeResponse;
 import org.projectnessie.model.Operations;
 import org.projectnessie.model.Reference;
+import org.projectnessie.model.Reference.ReferenceType;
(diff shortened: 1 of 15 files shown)


servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
```diff
@@ -80,8 +81,8 @@ Reference assignReference(
       @Valid Reference assignTo)
       throws NessieNotFoundException, NessieConflictException;
 
-  void deleteReference(
-      Reference.ReferenceType referenceType,
+  String deleteReference(
```
dimas-b, line 84: nit: add javadoc that the returned `String` is a hash value?

dimas-b (approved):

PR #5920, opened 2023-01-23
Nit: Just rename test classes

snazy:


XN137 (approved):
do we have a strict convention? because there seem to be other classes with a `Test` suffix as well

dimas-b (approved):

PR #5828, opened 2023-01-10
Validate runtime parameters in service implementations.

dimas-b:
This is to follow up on #5740, which introduced validation annotations to Nessie backend service interfaces, but did not enforce them.

This change introduced proper CDI beans for backend services in order for Quarkus and Weld to automatically validate method parameters in runtime.

Quarkus does the right validation out of the box.

Weld needs an explicit dependency on a CDI validator lib, for which the Hibernate Validator is used. The `@ValidateOnExecution` annotations are also required for this use case.

Namespace tests are adjusted to provide proper expected hashes during commit operations. These tests used to work fine over API v1, which is lenient in its NamespaceApi implementation, but using strict expected hashes is less ambiguous, matches old test behaviour in API v1, and works in both API versions.

PR #5921, opened 2023-01-23
Namespace API over REST v2 - fetch HEAD if hashOnRef is not specified

snazy:


dimas-b (approved):
Did you hit a test failure for this somewhere?

clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideCreateNamespace.java
```diff
@@ -74,11 +74,16 @@ public Namespace create()
     }
 
     try {
+      String expectedHash = hashOnRef;
+      if (expectedHash == null) {
+        expectedHash = api.getReference().refName(refName).get().getHash();
```
dimas-b, line 79: Could you add a test case for this?

snazy: THat's coming w/ #5924

dimas-b (changes requested):
Just realized that we should probably have a test case for this because I updated old tests to provide expected hashes in #5828

clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideCreateNamespace.java
dimas-b: cool :+1:

Issue #5926, opened 2023-01-23
Re-add batch authz checks


snazy:
With #5901 we changed the (internal) interfaces to separate concerns:
* let service-level implementation _push_ entries so that
* resource implementation (think: REST resources) can properly care about page sizes

The change in 5901 was already big-ish and made it necessary to temporarily remove some batch-authz check.

There need to be two follow-ups:
1. unify `*ApiImpl` and `*ApiImplWithAuthorization` classes in `:nessie-services`
2. re-add batch access checks for entries et al
3. address the comments in #5927

PR #5927, opened 2023-01-23
Unify `*ApiImplWithAuthorization` and "base" `*ApiImpl` classes

snazy:
This change simplifies the code and makes authorization a "first class" citizen. Also prevents duplicate calls to `namedRefWithHashOrThrow` for `getContent`+`getContents`+`getDiff`. Introduces a `NoopAuthorizer` for the case that there's no authorization.


servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -228,6 +235,22 @@ public Reference getReferenceByName(String refName, FetchOption fetchOption)
...
+    BatchAccessChecker check =
+        startAccessCheck().canCreateReference(RefUtil.toNamedRef(type, refName));
+    try {
+      check.canViewReference(namedRefWithHashOrThrow(sourceRefName, targetHash).getValue());
+    } catch (NessieNotFoundException e) {
+      // If the default-branch does not exist and hashOnRef points to the "beginning of time",
+      // then do not throw a NessieNotFoundException, but re-create the default branch. In all
+      // cases, re-throw the exception.
+      if (!(Reference.ReferenceType.BRANCH.equals(type)
+          && refName.equals(getConfig().getDefaultBranch())
```
dimas-b, line 247: So we do not run the default branch re-creation requests through access checks? Is that intended?

Would it be preferable to call `namedRefWithHashOrThrow` (line 241) separately and always pass the branch name to `check.canViewReference()`?

dimas-b (approved):
LGTM. The comments apply to old code as well, so it's more of a general "should we fix this?" question.

snazy:
Added notes for your comments to #5926

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
snazy: The "hard check" is intentional.

PR #5935, opened 2023-01-26
Separate version catalog for `buildSrc/`

snazy:
Have a separate version catalog into one for `buildSrc/`.

This helps Gradle to not invalidate the build cache for _every_ dependency update.

Background: `buildSrc/` used the `libs` version catalog, but if that one is updated, the cached build artifacts for `buildSrc/` are invalidated, which then in turn causes _all_ projects' build caches to be invalidated.

### DIFF
--- a/buildSrc/src/main/kotlin/OpenApiPlugin.kt
+++ b/buildSrc/src/main/kotlin/OpenApiPlugin.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Dremio
+ * Copyright (C) 2023 Dremio
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


gradle/baselibs.versions.toml
```diff
@@ -0,0 +1,32 @@
...
+[versions]
+jandexPlugin = "1.86"
+junit = "5.9.2"
+nessieBuildPlugins = "0.2.15"
+protobufPlugin = "0.9.2"
+shadowPlugin = "7.1.2"
+spotlessPlugin = "6.13.0"
+
+[libraries]
+assertj-core = { module = "org.assertj:assertj-core", version = "3.24.2" }
```
XN137, line 12: there now also is a `assertj-bom` we might want to use it maybe

buildSrc/src/main/kotlin/OpenApiPlugin.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.projectnessie.buildtools.smallryeopenapi.SmallryeOpenApiPlugin
+
+/** Makes the generated sources available to IDEs, disables Checkstyle on generated code. */
+@Suppress("unused")
+class OpenApiPlugin : Plugin<Project> {
```
XN137, line 24: is this split-out intentionally included in this PR ?
there doesnt seem any mention of it in the commit msg

XN137 (approved):
lgtm +1 assuming:
- macos CI faliure is sporadic/unrelated
- renovate will pick up the new file as well

buildSrc/src/main/kotlin/OpenApiPlugin.kt
snazy: yea - it's completely going "away" w/ #5936

dimas-b (approved):

PR #5946, opened 2023-01-27
REST v2: Correct namespace key evaluation for delete + get-multiple namespaces

snazy:
The CEL expression used to find a namespace and its children (`entry.namespace.startsWith(%s)` using `ContentKey.toString()` is incorrect and leads to wrong results for two reasons:
* example: `foo.bar.meep` would match the above for the namespace `foo.bar.me` - it does not consider key element boundaries
* dots in key elements are not correctly respected - so a key consisting of `["a", "b.b", "c"]` would be "equal" to `["a", "b", "b", "c"]` - need to use the path-escaped representation

### DIFF
--- a/clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideDeleteNamespace.java
+++ b/clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideDeleteNamespace.java
@@ -67,7 +67,7 @@ public final class ClientSideDeleteNamespace extends BaseDeleteNamespaceBuilder 
           .getEntries()
           .refName(refName)
           .hashOnRef(hashOnRef)
-          .filter(String.format("entry.namespace.startsWith('%s')", key))
+          .filter(String.format("entry.encodedKey.startsWith('%s.')", namespace.name()))
           .stream()
           .findAny()
           .isPresent()) {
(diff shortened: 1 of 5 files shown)


clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideDeleteNamespace.java
```diff
@@ -67,7 +67,7 @@ public void delete()
           .getEntries()
           .refName(refName)
           .hashOnRef(hashOnRef)
-          .filter(String.format("entry.namespace.startsWith('%s')", key))
+          .filter(String.format("entry.encodedKey.startsWith('%s.')", namespace.name()))
```
dimas-b, line 70: nit: WRT `encodedKey`, would it be clearer to call `namespace.toPathString()`?

clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideGetMultipleNamespaces.java
```diff
@@ -51,9 +51,11 @@ public GetNamespacesResponse get() throws NessieReferenceNotFoundException {
     try {
       GetEntriesBuilder getEntries = api.getEntries().refName(refName).hashOnRef(hashOnRef);
 
-      if (namespace != null) {
+      if (namespace != null && !namespace.isEmpty()) {
+        String nsName = namespace.name();
         getEntries.filter(
-            String.format("entry.key.startsWith('%s')", ContentKey.of(namespace.getElements())));
+            String.format(
+                "entry.encodedKey == '%s' || entry.encodedKey.startsWith('%s.')", nsName, nsName));
```
dimas-b, line 58: Newer V2 clients will not be able to run this against older V2 servers, but it's probably ok at this stage of API v2 development.

dimas-b (approved):
Nice catch :+1:

clients/client/src/main/java/org/projectnessie/client/util/v2api/ClientSideGetMultipleNamespaces.java
snazy: yep, should be okay

PR #5952, opened 2023-01-27
Add PVC for RocksDB in Nessie Helm chart

### LINKED ISSUE #5937: [Helm chart] Add PVC for ROCKSDB path
**Description**:
When deploying Nessie using the helm chart with ROCKSDB enabled, there is no way to use a path on a PVC. Thus, on every restart of the deployment, the data is lost. 

To allow this with the helm chart, a PVC could be added to the template that is mounted and used as storage location for ROCKSDB.

- Changes in Helm chart

**Expected Use Cases**:
Persisting data in kubernetes environments.

dimas-b:
* Add a PVC for use with RocksDB storage backend.

* Remove /tmp prefix from the RocksDB directory path for clarity.

* Set RocksDB example storage size request to 1 GB.

Fixes #5937

### DIFF
--- a/helm/nessie/templates/deployment.yaml
+++ b/helm/nessie/templates/deployment.yaml
@@ -33,6 +33,11 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.Version }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if eq .Values.versionStoreType "ROCKS" }}
+          volumeMounts:
+            - name: rocks-storage
+              mountPath: {{ .Values.rocksdb.dbPath }}
+          {{- end }}
           env:
             - name: NESSIE_VERSION_STORE_TYPE
               value: {{ .Values.versionStoreType }}
@@ -185,6 +190,12 @@ spec:
             timeoutSeconds: 10
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+      {{- if eq .Values.versionStoreType "ROCKS" }}
+      volumes:
+        - name: rocks-storage
+          persistentVolumeClaim:
+            claimName: {{ include "nessie.fullname" . }}
+      {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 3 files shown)


snazy (approved):
Not a Helm expert, but this looks like it fits :)

dimas-b:
@julwin :

> The label selector is not supported for my system 

I believe it is a standard k8s feature. Are you able to put custom labels on the physical volume, which you mean to use for Nessie?

snazy:
@adutra might be able to help with this one.

adutra:
Hi, I think @julwin is right and we might have a problem indeed.

In addition to the storage class, we are currently adding a label selector to the PVC, see [here](https://github.com/projectnessie/nessie/blob/abf41e91b0e0ed6f6c22242faa964cdd51f6e44a/helm/nessie/templates/storage.yaml#L16). But according to the [K8s docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1):

> Currently, a PVC with a non-empty selector can't have a PV dynamically provisioned for it.

This means that the PV must be manually created (with matching labels) and won't be automatically provisioned. This won't work if the storage class has a dynamic provisioner like AWS EBS. If this is the case, the provisioning will fail with:

> claim.Spec.Selector is not supported for dynamic provisioning on AWS

@dimas-b what was the initial use case for adding the label selector to the PVC? 

In any case I can see two options:

1. Remove the label selector completely; or
2. Parameterize the label selector and let users add one if they need to.

julwin:
Yes, that is exactly the problem we're seeing when using a dynamic provisioner. See [this issue](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1200) for reference (solution requires patching the StorageClass, which we're unable to).

From my testing it seems like Nessie is operating perfectly fine even when the selector is removed.

adutra:
Yeah, usually if you specify the storage class, it's because you are enabling dynamic provisioning – in which case you shouldn't specify the label selector. I think the simplest fix is just to make the label selector configurable as well.

dimas-b:
> what was the initial use case for adding the label selector to the PVC?

Just my proactive (possibly misguided) thinking that it could be useful to allow the user to choose one among many existing volumes.

> I think the simplest fix is just to make the label selector configurable as well.

+1 @adutra : Would you open a PR for this? :)

PR #5953, opened 2023-01-27
Add `@JsonView` everywhere

snazy:
... just to be on the safe side.

### DIFF
--- a/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestContentResource.java
+++ b/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestContentResource.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.services.rest;
 
+import com.fasterxml.jackson.annotation.JsonView;
 import javax.enterprise.context.RequestScoped;
 import javax.inject.Inject;
 import org.projectnessie.api.v1.http.HttpContentApi;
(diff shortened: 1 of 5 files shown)


dimas-b (approved):

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestContentResource.java
```diff
@@ -50,12 +52,14 @@ private ContentService resource() {
   }
 
   @Override
+  @JsonView(Views.V1.class)
```
dimas-b, line 55: We should probably put them on V1 _interfaces_ too. IIRC, Jersey and Quarkus look for view annotations in different places :shrug:

snazy: Works w/ Quarkus this way

dimas-b: Right, does Jersey see these new `Views.V1` annotations?

dimas-b: Tests will not discover that since model classes are backward-compatible for the methods in this PR :thinking:

PR #5979, opened 2023-01-30
Fix `MathAbsoluteNegative` in `NonTransactionalDatabaseAdapter`

snazy:
see https://errorprone.info/bugpattern/MathAbsoluteNegative

### DIFF
--- a/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
+++ b/versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
@@ -1022,7 +1022,11 @@ public abstract class NonTransactionalDatabaseAdapter<
   }
 
   protected final int refLogStripeForName(String refName) {
-    return Math.abs(refName.hashCode()) % config.getRefLogStripes();
+    int h = refName.hashCode();
+    if (h == Integer.MIN_VALUE) {
+      h++;
+    }
+    return Math.abs(h) % config.getRefLogStripes();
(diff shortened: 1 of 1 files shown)


versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -1022,7 +1022,11 @@ protected final RefLogParents fetchRefLogParents(
   }
 
   protected final int refLogStripeForName(String refName) {
-    return Math.abs(refName.hashCode()) % config.getRefLogStripes();
+    int h = refName.hashCode();
+    if (h == Integer.MIN_VALUE) {
+      h++;
```
dimas-b, line 1027: Aren't we breaking backward compatibility at storage level with this change to stripe computation?

snazy: Nope. If the hash was ever `MIN_VALUE`, if would have errored out

dimas-b: Not sure about errors... I could not find a solid error path by reading the code :thinking:  Most paths lead to constructing database keys from the integer and those look fine wrt negative numbers :shrug: Nonetheless, I guess in general the probably of hitting this hash value is low and reflog is deprecated... so it's probably of to do this change.

dimas-b (approved):

PR #5981, opened 2023-01-30
Helm: Add storage class for RockDB PVC

### LINKED ISSUE #5937: [Helm chart] Add PVC for ROCKSDB path

dimas-b:
Default to "standard"

Fixes #5937

### DIFF
--- a/helm/nessie/values.yaml
+++ b/helm/nessie/values.yaml
@@ -14,6 +14,7 @@ versionStoreType: INMEMORY
 
 # this is needed when selecting ROCKS
 rocksdb:
+  storageClassName: standard
   storageSize: 1Gi
   dbPath: /rocks-nessie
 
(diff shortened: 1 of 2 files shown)


helm/nessie/values.yaml
```diff
@@ -14,6 +14,7 @@ versionStoreType: INMEMORY
 
 # this is needed when selecting ROCKS
 rocksdb:
+  storageClassName: standard
   storageSize: 1Gi
```
dimas-b, line 18: @julwin : This PR adds storage class as a settable parameter as you suggested. `storageSize` was made already available by the previous PR.

julwin: Sounds good 😄

snazy (approved):
LGTM if @julwin's happy ;)

PR #5984, opened 2023-01-31
Add effective-reference to get-entries, diff and modify-namespace responses

snazy:


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -724,64 +766,89 @@ public void namespaces() throws Exception {
...
         .containsExactlyInAnyOrder(
             namespace1WithId, namespace2update, namespace3WithId, namespace4WithId);
 
-    api()
-        .updateProperties()
-        .refName(main.getName())
-        .namespace(namespace2)
-        .removeProperty("foo")
-        .update();
+    if (isV2()) {
```
dimas-b, line 819: those version `if`s look awkward to me... Could we perhaps introduce a utility test method taking a builder and expected v2 data to do the actual API call and validate extra data only in case the API supports it? This way the main test method will be nice and linear... WDYT?

dimas-b: The util method could also validate exceptions on the "unsupported" v1 call path.

model/src/main/java/org/projectnessie/model/GetNamespacesResponse.java
```diff
@@ -15,16 +15,27 @@
...
+@Value.Immutable
 @JsonSerialize(as = ImmutableGetNamespacesResponse.class)
 @JsonDeserialize(as = ImmutableGetNamespacesResponse.class)
 public interface GetNamespacesResponse {
   @NotNull
   List<Namespace> getNamespaces();
+
+  @JsonView(Views.V2.class)
+  @Nullable
+  Reference getEffectiveReference();
```
dimas-b, line 36: `GetNamespacesResponse` is not used in API v2. Can clients actually get any data for this `Reference`?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
snazy: Let's not add unrelated stuff to this PR.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -630,7 +639,11 @@ public void entries() throws Exception {
                     .toArray(Operation[]::new))
             .commit();
 
-    List<EntriesResponse.Entry> notPaged = api().getEntries().reference(main).get().getEntries();
+    EntriesResponse response = api().getEntries().reference(main).get();
+    if (isV2()) {
+      soft.assertThat(response.getEffectiveReference()).isEqualTo(main);
```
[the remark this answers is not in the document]
snazy: Nah - not helpful IMHO

PR #5985, opened 2023-01-31
Allow retrieval of `Content` with `get-entries` for API v2

snazy:
Since the `get-entries` API call supports paging it is fine to optionally also return the `Content` objects for the returned keys.

### DIFF
--- a/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestTreeResource.java
+++ b/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestTreeResource.java
@@ -133,6 +133,7 @@ public class RestTreeResource implements HttpTreeApi {
             params.namespaceDepth(),
             params.filter(),
             params.pageToken(),
+            false,
             new PagedCountingResponseHandler<EntriesResponse, EntriesResponse.Entry>(maxRecords) {
               @Override
               public EntriesResponse build() {
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -729,6 +729,7 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
       Integer namespaceDepth,
       String filter,
       String pagingToken,
+      boolean withContent,
       PagedResponseHandler<R, Entry> pagedResponseHandler,
       Consumer<WithHash<NamedRef>> effectiveReference)
       throws NessieNotFoundException {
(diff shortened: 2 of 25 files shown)


servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -793,9 +794,12 @@ protected Set<Check> checksForEntry(KeyEntry entry) {
...
             }
 
+            Content c = key.getContent();
             EntriesResponse.Entry entry =
-                EntriesResponse.Entry.entry(
-                    fromKey(key.getKey()), key.getType(), key.getContentId());
+                c != null
+                    ? EntriesResponse.Entry.entry(fromKey(key.getKey()), key.getType(), c)
+                    : EntriesResponse.Entry.entry(
+                        fromKey(key.getKey()), key.getType(), key.getContentId());
```
dimas-b, line 802: nit: would it not be more straight-forward to use an `Entry` builder in this case to avoid conditional operators?

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -468,11 +468,30 @@ public PaginationIterator<KeyEntry> getKeys(Ref ref, String pagingToken)
...
+                      .get(entry.getKey());
+              if (cs != null) {
+                Content content = mapContentAndState(cs);
+                return KeyEntry.of(
+                    DefaultStoreWorker.contentTypeForPayload(entry.getPayload()),
+                    entry.getKey(),
+                    content);
+              }
+            } catch (ReferenceNotFoundException e) {
+              throw new RuntimeException(e);
```
dimas-b, line 487: nit: `IllegalStateException` is more concrete,  IMHO, since we expect the reference to be valid at this point.

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestTreeResource.java
```diff
@@ -133,6 +133,7 @@ public EntriesResponse getEntries(String refName, EntriesParams params)
             params.namespaceDepth(),
             params.filter(),
             params.pageToken(),
+            false,
```
dimas-b, line 136: Would it make sense to forbid (exception) `GetEntriesBuilder.withContent(true)` in API _v1_ for the sake of clarify?

dimas-b (commented):
LGTM :+1:  Minor comments.

dimas-b (approved):

PR #6000, opened 2023-02-02
Add support for API v2 to the compatibility test framework

### LINKED ISSUE #5998: Add support for API v2 to `OlderNessieServersExtension`
`OlderNessieServersExtension` and related test extensions should support API v2 (for server versions that have it).

dimas-b:
* Fix class loading for older clients

* Declare actual dependency on slf4j in :nessie-client

* Resolve `NessieApiV2` fields in test classes

* Load V2 REST resources when they are available in Jersey servers

Fixes #5998

### DIFF
--- a/compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/AbstractNessieApiHolder.java
+++ b/compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/AbstractNessieApiHolder.java
@@ -67,13 +67,13 @@ abstract class AbstractNessieApiHolder implements CloseableResource {
       Field field,
       Version version,
       Function<ExtensionContext, NessieServer> nessieServerSupplier) {
-    Map<String, String> configs = buildApiBuilderConfig(context, field, nessieServerSupplier);
-
     // This method is only called for fields that are annotated with NessieAPI.
     NessieAPI nessieAPI = field.getAnnotation(NessieAPI.class);
 
     @SuppressWarnings("unchecked")
     Class<? extends NessieApi> apiType = (Class<? extends NessieApi>) field.getType();
+    Map<String, String> configs =
+        buildApiBuilderConfig(context, field, apiType, nessieServerSupplier);
     ClientKey clientKey = new ClientKey(version, nessieAPI.builderClassName(), apiType, configs);
     return clientKey;
   }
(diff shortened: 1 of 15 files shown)


compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/Util.java
```diff
@@ -61,4 +64,9 @@ static <T> T withClassLoader(ClassLoader classLoader, Callable<T> callable) thro
       Thread.currentThread().setContextClassLoader(appClassLoader);
     }
   }
+
+  static URI resolve(URI base, Class<? extends NessieApi> apiType) {
```
snazy, line 68: `resolve` is a bit too generic - maybe `resolveNessieBaseUri` (with that you could use a static import for this method, too)

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/OldNessie.java
```diff
@@ -86,6 +86,19 @@ static ClassLoader oldNessieClassLoader(Version version, String mainArtifactId)
                                     "io.opentracing", artifactId, "jar", opentracingVersion),
                                 "runtime")));
           };
+    } else if (Version.parseVersion("0.46.0").isLessThanOrEqual(version)
```
snazy, line 89: Why not have constants for this in `Version`, like the other ones added, too?

compatibility/common/src/main/java/org/projectnessie/tools/compatibility/internal/AbstractNessieApiHolder.java
```diff
@@ -129,7 +133,12 @@ protected static AutoCloseable createNessieClient(ClassLoader classLoader, Clien
           };
       builderInstance = fromConfigMethod.invoke(builderInstance, getCfg);
 
-      Class<?> targetClass = classLoader.loadClass(clientKey.getType().getName());
+      Class<?> targetClass;
+      try {
+        targetClass = classLoader.loadClass(clientKey.getType().getName());
+      } catch (ClassNotFoundException e) {
```
snazy, line 139: Is this one to just have `null` for `NessieApiV2` w/ older clients? Should it be guarded using a check of `clientKey.getVersion()` against a constant in `Version`?

dimas-b: Yes. Since injection is filed-based we have to provide a value for `NessieApiV2` fields in test classes even when the test env. is set up for V1.

I'll add a version check.

dimas-b: I've refactored to a non-null proxy producing specific exceptions when/if the unsupported API is _accessed_ (in c3a1ee881d967c2ef244e1d11fd4a6de24ca2f33)... How does that look?

snazy (approved):
LGTM +1

PR #6005, opened 2023-02-02
CI: fix helm-testing action refs

snazy:
... do not use `main` and do not cache the minikube tarball (it's big and pollutes the actions cache).


XN137 (approved):
thx for the quick followup, lgtm

PR #6012, opened 2023-02-03
Refactor API v2 reference resolution logic into a utility class

dimas-b:
Put the new utility class under `model`.

This is to keep the code that interprets ref strings in REST path elements closer to the REST interface definitions.

### DIFF
--- a/model/src/main/java/org/projectnessie/api/v2/params/ReferenceResolver.java
+++ b/model/src/main/java/org/projectnessie/api/v2/params/ReferenceResolver.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


model/src/main/java/org/projectnessie/api/v2/params/ReferenceResolver.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.api.v2.params;
+
+import java.util.function.Supplier;
+import org.projectnessie.model.Branch;
+import org.projectnessie.model.NessieConfiguration;
+import org.projectnessie.model.Reference;
+
+public class ReferenceResolver {
```
snazy, line 23: `final` + private c'tor

model/src/main/java/org/projectnessie/api/v2/params/ReferenceResolver.java
```diff
@@ -0,0 +1,38 @@
...
+
+  public static Reference toReference(String refPathString, Supplier<NessieConfiguration> config) {
+    if (DEFAULT_REF_IN_PATH.equals(refPathString)) {
+      return Branch.of(config.get().getDefaultBranch(), null);
+    }
+
+    return toReference(refPathString, Reference.ReferenceType.BRANCH);
+  }
+
+  public static Reference toReference(String refPathString, Reference.ReferenceType type) {
```
snazy, line 35: Maybe `resolveReferencePath` ?

dimas-b: Renamed to `resolveReferencePathElement`... hope the name is not too long :)

snazy: I like such names - and methods/fields with such names can be nicely used with static imports.

snazy (approved):

snazy (approved):

PR #6020, opened 2023-02-03
Produce NessieBadRequestException when paging is not available

dimas-b:
Affected APIs:
* getAllReferences()
* getEntries()
* getDiff()

Previously, if the page has to be cut short in the responses from these APIs (e.g. if `maxRecords` is set). The response would have `hasMore == true` but `pageToken == null`. This can be confusing to clients.

With this change the service with cause a `NessieBadRequestException` when it has to limit the response size, but cannot generate a paging token to clearly indicate that paging is not supported in these situations.

Clients may still observe that requests with a large `maxRecords` value succeed, but with a smaller `maxRecords` value fail. This is because the decision to limit the page size is made at the REST API level, but the token is generated at the storage level and exact behaviour depends on the available data.

Full pagination support is to be added with the new model (#5641).


snazy (commented):
I'd prefer to make the tests conditional based on the result of `pagingSupported()` `protected` in `BaseTestNessieApi`, just let it return `false` by default instead of taking a callable.

dimas-b:
Rebased and refactored `pagingSupported()`. Please have a look at the second commit. WDYT?

snazy (approved):
Let's go ahead with this one - I can update it in #5641 later

PR #6021, opened 2023-02-03
Expose definitive Nessie API version to tests

dimas-b:
Use the injected `NessieApiVersion` enum value for version checks to avoid relying on client implementation details.

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
@@ -116,7 +116,7 @@ public abstract class BaseTestNessieApi {
   }
 
   public boolean isV2() {
-    return NessieApiVersion.V2.equals(apiVersion);
+    return NessieApiVersion.V2 == apiVersion;
   }
 
   @AfterEach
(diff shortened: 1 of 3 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -115,7 +116,7 @@ public NessieApiV1 api() {
   }
 
   public boolean isV2() {
-    return api instanceof NessieApiV2;
+    return NessieApiVersion.V2.equals(apiVersion);
```
snazy, line 119: NIt: `==` (enum)

snazy (dismissed):

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
dimas-b: fixed (and rebased)

ajantha-bhat:
@dimas-b: 
Just wondering whether it is better to define `apiVersion()` inside  `NessieApi` interface? 

Now the changes in the PR are sufficient but the `NessieClientFactory` will only be available in the constructor and the user has to store it to a variable and use it instead of checking with the api object itself.

dimas-b:
> Just wondering whether it is better to define apiVersion() inside NessieApi interface?

Currently the `NessieApiVersion` class exists in test code only. I wanted to keep the PR small. However, in general it may be possible to expose a version enum in `NessieApi`. I think this needs to be considered carefully though.

At this point the idea is that clients _request_ the API version they want by providing an API class (contract) to the client builder. In the main use case the client has a clear idea of what to expect from the API so no conditional execution (based on the version getter) should be necessary.

Test code, on the other hand, often needs to validate difference in behaviour across API versions.

If you imply the Iceberg `NessieCatalog` tests, ideally they should not depends on the API version, but of course, the path to achieving that may have to cross some rough terrain.

PR #6047, opened 2023-02-07
python: upgrade flake8 to v6.0.0

XN137:
replaces https://github.com/projectnessie/nessie/pull/6031

flake8 6.0.0 Requires-Python >=3.8.1

so we move it to the lint config, which we only run on 3.8 anyway:
https://github.com/projectnessie/nessie/blob/80997890e19072ad5d69634729e201a29ed7fc21/python/tox.ini#L20-L25

### DIFF
--- a/python/requirements_basedev.txt
+++ b/python/requirements_basedev.txt
@@ -20,7 +20,6 @@ assertpy==1.1
 bump2version==1.0.1
 codecov==2.1.12
 coverage==7.1.0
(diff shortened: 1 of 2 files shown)


snazy (approved):
Sounds like reasonable approach! +1

dimas-b (approved):

ajantha-bhat (approved):

PR #6048, opened 2023-02-07
Cleanup InmemoryStore after tests

snazy:
This was an attempt to fix an OOM in `:nessie-quarkus:test`. Although this change did not fix the issue, it is not a bad idea to be nice and release unused objects in Quarkus (triggered earliest after a test class).

### DIFF
--- a/versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryStore.java
+++ b/versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryStore.java
@@ -49,7 +49,17 @@ public class InmemoryStore implements DatabaseConnectionProvider<InmemoryConfig>
   public void initialize() {}
 
   @Override
-  public void close() {}
+  public void close() {
+    globalStateLog.clear();
+    commitLog.clear();
+    keyLists.clear();
+    refLog.clear();
+    refHeads.clear();
+    refNames.clear();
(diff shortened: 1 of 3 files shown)


versioned/persist/inmem/src/main/java/org/projectnessie/versioned/persist/inmem/InmemoryStore.java
```diff
@@ -49,7 +49,17 @@ public void configure(InmemoryConfig config) {}
   public void initialize() {}
 
   @Override
-  public void close() {}
+  public void close() {
+    globalStateLog.clear();
```
dimas-b, line 53: Will GC not reclaim them anyway after the Store itself is no longer referenced?

snazy: maybe - maybe not - depends on the tests being run

dimas-b: Do you mean the `dispose` may be invoked, but the Store object still referenced after that?

snazy: I mean that the store may survive because it's still referenced - the _contents_ of the store not cleaned up

dimas-b (approved):

PR #6062, opened 2023-02-09
Remove `@CheckForNull` annotation, no `jakarta.*` counterpart

snazy:


versioned/spi/src/main/java/org/projectnessie/versioned/paging/FilteringPaginationIterator.java
```diff
@@ -42,7 +41,6 @@ public FilteringPaginationIterator(
     this.sourcePredicate = sourcePredicate;
   }
 
-  @CheckForNull
```
dimas-b, line 45: nit: Should we have `@Nullable` instead?

dimas-b (approved):

PR #6063, opened 2023-02-09
Remove `@ThreadSafe` annotation, no `jakarta.*` counterpart

snazy:


adutra:
Could we consider using `net.jcip.annotations.ThreadSafe` ?

https://jcip.net/annotations/doc/index.html

snazy:
> Could we consider using `net.jcip.annotations.ThreadSafe` ?

Could, but for a single class it's maybe not necessary (we don't have similar annotations elsewhere)

dimas-b (approved):

PR #6068, opened 2023-02-09
Gatling simulation for a mixed workload

snazy:
Current Gatling simulations target only one specific use case and are always focued on "commit rate".

The simulation in this PR supports a mixed workload.

"Workflow" of the whole simulation
1. Setup phase:
    1. Create a starting branch with many Iceberg tables, allows creating multiple tables in a single commit to reduce the setup time.
    2. Create additional branches, if more than one branch is configured, based on the HEAD of the branch from step 1.1.
2. "Mixed workload" phase with three use cases. All can be tweaked with the number of simulated users and the rate per second for each of the simulated users.
    1. UI user: 1. List all tables 2. Get current content of one tables 3. Update that table (Nessie commit)
    2. "Readers" (read from N tables at some fixed rate):
        1. Get current content of one or more tables
    2. "Writer" (think: a scheduled job committing at some fixed rate): 1. Get current content of one tables 1. Update that table (Nessie commit)

The "global" set of tables consists of a "pool" of tables and a subset of that "pool" as "active" tables to simulate the (probably common) scenario that people have many 10000s tables, but only actively use a fraction of those.

### DIFF
--- a/perftest/simulations/src/gatling/scala/org/projectnessie/perftest/gatling/MixedWorkloadsParams.scala
+++ b/perftest/simulations/src/gatling/scala/org/projectnessie/perftest/gatling/MixedWorkloadsParams.scala
@@ -1,0 +1,158 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.perftest.gatling
+
+import io.gatling.core.session.Session
+import java.lang.Integer.getInteger
+import java.lang.System.getProperty
+import java.util.concurrent.ThreadLocalRandom
+import org.projectnessie.model.{ContentKey, Namespace}
+import scala.concurrent.duration.{Duration, FiniteDuration, HOURS, NANOSECONDS}
+
+case class SessionsAndRate(
+    users: Int,
+    opRate: Double
+) {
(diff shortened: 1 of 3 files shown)


perftest/simulations/src/gatling/scala/org/projectnessie/perftest/gatling/MixedWorkloadsParams.scala
```diff
@@ -0,0 +1,158 @@
...
+    activeTables: Int,
+    suffix: String = ""
+) {
+  def contentKey(t: Int, session: Session): ContentKey = {
+    val namespace = session("namespace").as[Namespace]
+    ContentKey.of(namespace, s"table-${t}$suffix")
+  }
+
+  def randomActiveTable(): Int = {
+    ThreadLocalRandom.current().nextInt(activeTables)
```
dimas-b, line 62: This will always cause active keys to be in front (lexicographically) of the total list of tables. Would it be preferable to choose a random but deterministic sub-set of active tables spanning the whole list?

snazy: Maybe as a follow-up

perftest/simulations/src/gatling/scala/org/projectnessie/perftest/gatling/MixedWorkloadsSimulation.scala
```diff
@@ -0,0 +1,379 @@
...
+
+  private def prepareInitialReference: NessieActionBuilder =
+    nessie(s"prepare - Create initial branch")
+      .execute { (client, session) =>
+        // create the branch (errors will be ignored)
+        val branch = client
+          .createReference()
+          .reference(Branch.of(params.branches.branchOf(0), null))
+          .create()
+        System.err.println(s"Created initial $branch")
```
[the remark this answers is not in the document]
snazy: no special reason

perftest/simulations/src/gatling/scala/org/projectnessie/perftest/gatling/MixedWorkloadsParams.scala
dimas-b: sure

dimas-b (approved):

Issue #6094, opened 2023-02-14
Nessie native image for arm64


tmnd1991:
**Description**:
Right now nessie docker image is published only for amd64 architecture. 
It would be good to publish the image also for arm64.

**Requested Changes in public API**:
- ❌Changes in REST API
- ❌Changes in Java Client API
- ❌Changes in Python Client API
- ❌Changes in SQL extentions
- ✅Changes in build/publish

**Expected Use Cases**:
Run Nessie natively on arm64 machines.

tmnd1991:
I dig a bit in the build and I understood that Nessie is using `jib` to build the container image.

[quarkus.jib.platforms](https://quarkus.io/guides/container-image#quarkus-container-image-jib_quarkus.jib.platforms) could be used to set the arm64 architecture, but unfortunately the "multi" platform feature of jib does not work for native image, so unfortunately two different steps/workflows would be needed to make it work.

Would you accept a PR on this?

dimas-b:
Thanks for the PR :eye:

snazy:
Closed after discussion in and in favor of #6099

snazy:
Closed after discussion in and in favor of #6099 / #6110

snazy:
Oh, and what I haven't really realized before: GitHub offers only offers runners on x86_64, none for amd64/aarch64. In other words, we'd have had no chance to build the amd64/aarch64 native binaries in CI :(

tmnd1991:
Yeah, that sucks :( you probably need custom runners to do so. I know the guys at typelevel managed to have them, but never investigated how

PR #6098, opened 2023-02-14
Add support to native arm64 build

tmnd1991:
PR for #6094.
I did not add any workflow change because I wasn't sure.
The arm64 build worked locally without any problem running:

`./gradlew nessie-quarkus:quarkusBuild -Pnative -Pdocker -Parm64`

Standard amd64 build works when passing -Pamd64 or even when neither `arm64` and `amd64` are defined.

### DIFF
--- a/buildSrc/src/main/kotlin/Utilities.kt
+++ b/buildSrc/src/main/kotlin/Utilities.kt
@@ -280,3 +280,8 @@ class SparkScalaVersions(
   val sparkVersion: String,
   val scalaVersion: String
 ) {}
+
+enum class Arch(val value: String) {
+  AMD64("amd64"),
+  ARM64("arm64")
+}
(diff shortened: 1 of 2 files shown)


dimas-b (commented):
@tmnd1991 : Thanks for the contribution!

The changes set LGTM, but please run `./gradlew spotlessApply` to let the build finish :)

As for CI changes, I think we ought to include workflows to actually run ARM image tests the same way we test ADM64. Do you know whether GH supports that OOTB?

tmnd1991:
sure, just run spotless.
I will look into arm64 support for GH actions

snazy:
> arm64 support for GH actions

Just noting: these are the workflows `main.yml`, `pull-request-native.yml`, `release-publish.yml` and `snapshot-publish.yml`.

Wonder whether it would be "good enough" to publish "pure Java" images. I've opened #6099 to discuss this.

snazy:
Closed after discussion in and in favor of #6099 / #6110

PR #6103, opened 2023-02-15
Simplify Quarkus build configuration

snazy:
... and prepare for multiplatform builds


servers/lambda/build.gradle.kts
```diff
@@ -55,25 +55,39 @@ dependencies {
 buildForJava11()
 
 val useDocker = project.hasProperty("docker")
-val packageType = quarkusPackageType()
-val quarkusBuilderImage = libs.versions.quarkusBuilderImage.get()
+val useNative = project.hasProperty("native")
+var jibPlatforms: String = System.getProperty("quarkus.jib.platforms", "linux/amd64")
```
dimas-b, line 59: just wondering: how is it different from project property? Why do we use a project property for `native` by java system property for `platforms`?

dimas-b (commented):
LGTM, just one question

servers/lambda/build.gradle.kts
snazy: THose are used more than once.

BTW: "project property for native " --> that's a Kotlin `val`, not a _project_ property.

dimas-b (approved):

PR #6106, opened 2023-02-15
Disable :nessie-compatibility-common tests on macOS

snazy:
RocksDB JNI core dumps...


compatibility/common/build.gradle.kts
```diff
@@ -77,3 +79,11 @@ tasks.withType<Test>().configureEach {
     excludeTestsMatching("TestNessieCompatibilityExtensions\$*")
   }
 }
+
+// Compatibility tests fail on macOS with a SIGSEGV via rocksdb:
```
XN137, line 83: shouldnt we just skip the rocksdb tests on macos?
or do the other tests provide no value in terms of OS test coverage (i.e. in-memory)?

XN137 (approved):

XN137 (approved):
thanks for the update

tbh i was originally thinking that maybe we can only include the rocksdb tests conditionally here:
https://github.com/projectnessie/nessie/blob/ad7cfddc003c4c8d7c7fc9e4d7d041eabd1b2f9f/compatibility/common/build.gradle.kts#L62-L63

but i am not familiar with how we discover the tests to run, so maybe this wouldnt work.

generally speaking i am not sure how much value testing on macos provides, so the patch as is looks fine to me, even though it no longer seems rocksdb specific.

Issue #6114, opened 2023-02-16
ITSparkExtensions fail w/ Spark 3.3 + latest Iceberg


snazy:
Example CI run: https://github.com/projectnessie/query-engine-integration-tests/actions/runs/4190567702/jobs/7264100795

Example failure: https://scans.gradle.com/s/f3ywj2or6e6e6/tests/overview

ajantha-bhat:
ACK. Let me check it.

ajantha-bhat:
Analysis from the code review:

https://github.com/projectnessie/nessie/blob/31357d3f10e50515f7e40a0d75225b4f339ca1af/clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java#L631

Yesterday this PR got merged (https://github.com/apache/iceberg/pull/6801) which introduces one more argument. Hence, the strict check fails. 

@snazy: We need to update the testcase when we bump the Iceberg version or shall I modify the test case now to loosen the check?

ajantha-bhat:
I have fixed all 4 test failures from the above build in https://github.com/projectnessie/nessie/pull/6117

But when I test locally (with the iceberg master), I do get some timeouts for a few tests (even without my fix).  Not sure about the cause yet. But the above build doesn't seem to have those errors.
<img width="1705" alt="image" src="https://user-images.githubusercontent.com/5889404/219437660-2a6c1862-54e7-45a1-99db-74ec77243555.png">

ajantha-bhat:
Closing the ticket as the PR is merged and build is passing
https://github.com/projectnessie/query-engine-integration-tests/actions/runs/4200564588

PR #6117, opened 2023-02-16
Test: Update the validation to unblock query-engine-integration-tests

ajantha-bhat:
Yesterday this PR got merged (https://github.com/apache/iceberg/pull/6801) which introduces one more output value. Hence, the strict check fails.

This PR is to unblock `query-engine-integration-tests` 

Part of fixing #6114

### DIFF
--- a/clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java
+++ b/clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java
@@ -628,7 +628,8 @@ public abstract class AbstractNessieSparkSqlExtensionTest extends SparkSqlTestBa
       throws NessieNotFoundException {
     List<Object[]> result = executeCompaction(tableName);
     // re-written files count is 2 and the added files count is 1
-    assertThat(result).hasSize(1).containsExactly(row(2, 1));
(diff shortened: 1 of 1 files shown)


clients/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/AbstractNessieSparkSqlExtensionTest.java
```diff
@@ -628,7 +628,8 @@ void executeAndValidateCompaction(String branchName, String branchHash, String t
       throws NessieNotFoundException {
     List<Object[]> result = executeCompaction(tableName);
     // re-written files count is 2 and the added files count is 1
-    assertThat(result).hasSize(1).containsExactly(row(2, 1));
+    assertThat(result.get(0)[0]).isEqualTo(2);
+    assertThat(result.get(0)[1]).isEqualTo(1);
```
ajantha-bhat, line 632: I did try `.extracting()` but it became too messy/complex as `result.get(0)` can have a bigger array in the latest iceberg.

dimas-b: nit: `.first()` might help. Also, would just dropping `hasSize(1)` and using `.containt(row(2, 1))` work?

dimas-b (approved):

PR #6127, opened 2023-02-17
GH workflow to smoke test Docker images

snazy:
Exercises Docker image build and push and lets both the native and Java image run for 1 minute to validate those start up fine.

### DIFF
--- a/.github/workflows/pull-request-docker.yml
+++ b/.github/workflows/pull-request-docker.yml
@@ -1,0 +1,181 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Projectnessie GitHub Pull-Request / Docker build and publishing
+
+name: Docker images test
+
+on:
+  pull_request:
+    types: [labeled, opened, synchronize, reopened]
+  schedule:
+    # Run daily on week days
+    - cron:  '0 4 * * 1-5'
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docker-testing:
+    name: Docker build and publishing
+    runs-on: ubuntu-22.04
+    if: (github.repository == 'projectnessie/nessie' && github.event_name != 'pull_request') || contains(github.event.pull_request.labels.*.name, 'pr-docker')
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3.3.0
+        with:
+          fetch-depth: 0
+      - name: Setup runner
+        uses: ./.github/actions/setup-runner
+        with:
+          more-memory: 'true'
+      - name: Setup Java, Gradle
+        uses: ./.github/actions/dev-tool-java
+
+      - name: Setup docker-registry
+        run: |
+          sudo apt-get install -y docker-registry apache2-utils 
+          cat <<! > config.yml
+          version: 0.1
+          log:
+            fields:
+              service: registry
+          storage:
+            cache:
+              blobdescriptor: inmemory
+            filesystem:
+              rootdirectory: /var/lib/docker-registry
+            delete:
+              enabled: true
+          http:
+            addr: 127.0.0.1:5000
+            headers:
+              X-Content-Type-Options: [nosniff]
+          auth:
+            htpasswd:
+              realm: basic-realm
+              path: /etc/docker/registry/htpasswd
+          health:
+            storagedriver:
+              enabled: true
+              interval: 10s
+              threshold: 3
+          !
+          sudo mv config.yml /etc/docker/registry/config.yml
+          sudo htpasswd -cBb /etc/docker/registry/htpasswd micky mouse 
+          
+          sudo service docker-registry restart
+          
+          echo mouse | docker login -u micky --password-stdin localhost:5000
+
+          VERSION="$(cat version.txt)"
+          DOCKER_VERSION="${VERSION%-SNAPSHOT}"
+          echo "DOCKER_VERSION=${DOCKER_VERSION}" >> ${GITHUB_ENV}
+          echo "DOCKER_IMAGE=localhost:5000/nessie-testing" >> ${GITHUB_ENV}
+
+      - name: Gradle / prepare
+        uses: gradle/gradle-build-action@v2
+        with:
+          cache-read-only: true
+          # Just fetch cache, defer build to "later"
+          arguments: projects
+
+      - name: Docker images publishing
+        env:
+          ARTIFACTS: ../build-artifacts
+        run: |
+          rm -rf "${ARTIFACTS}"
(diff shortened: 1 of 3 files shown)


snazy:
Example run https://github.com/snazy/nessie/actions/runs/4202720836

.github/workflows/pull-request-docker.yml
```diff
@@ -0,0 +1,182 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docker-testing:
+    name: Docker build and publishing
+    runs-on: ubuntu-22.04
+    # if: (github.repository == 'projectnessie/nessie' && github.event_name != 'pull_request') || contains(github.event.pull_request.labels.*.name, 'pr-docker')
+    if: (github.repository == 'snazy/nessie' && github.event_name != 'pull_request') || contains(github.event.pull_request.labels.*.name, 'pr-docker')
```
XN137, line 37: forgot to replace with previous line?

.github/workflows/pull-request-docker.yml
```diff
@@ -0,0 +1,182 @@
...
+          arguments: projects
+
+      - name: Docker images publishing
+        env:
+          ARTIFACTS: ../build-artifacts
+        run: |
+          rm -rf "${ARTIFACTS}"
+          mkdir -p "${ARTIFACTS}"
+
+          tools/dockerbuild/build-push-images.sh \
```
XN137, line 105: is there really value in exercising the push part in this workflow?
couldnt we just be building and testing the images locally without the local docker registry?

.github/workflows/pull-request-docker.yml
```diff
@@ -0,0 +1,182 @@
...
+          !
+
+      - name: Check if Docker native image works
+        run: |
+          docker run --detach --name nessie ${DOCKER_IMAGE}:latest-native
+          echo "Let native Nessie Docker image run for one minute (to make sure it starts up fine)..."
+          for i in {1..60}; do
+            STATUS="$(docker container inspect nessie | jq -r '.[0].State.Status')"
+            if [[ ${STATUS} != "running" ]] ; then
+              echo "Nessie native Docekr image stopped on its own ... a bug?" > /dev/stderr
```
XN137, line 141: typo: Docekr

.github/workflows/pull-request-docker.yml
```diff
@@ -0,0 +1,182 @@
...
+          sudo htpasswd -cBb /etc/docker/registry/htpasswd micky mouse 
+          
+          sudo service docker-registry restart
+          
+          echo mouse | docker login -u micky --password-stdin localhost:5000
+
+          VERSION="$(cat version.txt)"
+          DOCKER_VERSION="${VERSION%-SNAPSHOT}"
+          echo "DOCKER_VERSION=${DOCKER_VERSION}" >> ${GITHUB_ENV}
+          echo "DOCKER_IMAGE=localhost:5000/nessie-testing" >> ${GITHUB_ENV}
```
XN137, line 89: does the `build-push-images.sh` script support an image name that includes the registry?

should we only use the registry name for the pull commands?

.github/workflows/pull-request-docker.yml
snazy: oops, yea

.github/workflows/pull-request-docker.yml
snazy: The images are only _built_, not _added_ to the "local" thingy.

.github/workflows/pull-request-docker.yml
snazy: same would be true for `quay.io/blah/blah`

.github/workflows/pull-request-docker.yml
XN137: sorry i dont understand, why do we need the local registry then if you claim we are not pushing images to it?

afaict we need the local registry because the script not only builds the images but also unconditionally pushes them:
https://github.com/projectnessie/nessie/blob/aeb8ed725fdc77a0094e72eaef427e7cea09ce32/tools/dockerbuild/build-push-images.sh#L171-L173

if we would only push them conditionally (i.e. additional scirpt param) then they are just build locally and we can use them in this workflow without needing a local registry

snazy: >is there really value in exercising the push part in this workflow?

It is _mandatory_ to be able to _run_ the image, right?

snazy: `docker buildx build ...` does _not_ make the built images available locally - e.g. via `docker images`.

snazy: The local Docker repository is platform specific - a registry is not.

XN137: according to https://docs.docker.com/engine/reference/commandline/buildx_build/#load

```
Push the build result to a registry (--push)

Shorthand for [--output=type=registry](https://docs.docker.com/engine/reference/commandline/buildx_build/#registry). Will automatically push the build result to registry.
```

```
Load the single-platform build result to docker images (--load)

Shorthand for [--output=type=docker](https://docs.docker.com/engine/reference/commandline/buildx_build/#docker). Will automatically load the single-platform build result to docker images.
```

so if we used `--load` instead of `--push` we wouldnt need the local registry imo, also not for running the images afterwards

snazy: But: why?
This WF is meant to exercise the whole roundtrip, so why add extra complexity into the tested code for the sake of not configuring a dummy local registry?

XN137: > The local Docker repository is platform specific - a registry is not.

when we push to the local registry and later pull down the images... are we not running them from the "local Docker repository" as well ?

> why add extra complexity into the tested code for the sake of not configuring a dummy local registry?

because imo configuring the local registry in this workflow is much more complexity that deciding between `--push` or `--load`

> This WF is meant to exercise the whole roundtrip

this was exactly my original question `is there really value in exercising the push part in this workflow?`
but first you claimed we are not pushing anything to the local registry...

from my POV with the extra step we are mostly testing if buildx can push to a registry, so i didnt see that much value in it..
if you think its worth keeping then "OK"

.github/workflows/pull-request-docker.yml
```diff
@@ -0,0 +1,181 @@
...
+          echo "Nessie native Docker image smoke test: PASSED"
+          docker stop nessie
+          docker rm nessie
+
+      - name: Check if Docker Java image works
+        run: |
+          docker run --detach --name nessie ${DOCKER_IMAGE}:latest-java
+          echo "Let Nessie Java Docker image run for one minute (to make sure it starts up fine)..."
+          for i in {1..60}; do
+            STATUS="$(docker container inspect nessie | jq -r '.[0].State.Status')"
```
XN137, line 163: could there be value to run a HTTP GET with curl against our API endpoints as well?
maybe there could be reasons the container is still running but the application or http endpoints are dead

maybe our Dockerfile could contain a https://docs.docker.com/engine/reference/builder/#healthcheck for this purpose, but potentially as a followup

XN137 (approved):

.github/workflows/pull-request-docker.yml
snazy: > you claimed we are not pushing anything to the local registry...

Nope - i did not. The script _does_ push to the local _registry_ - not the local "thingy" on which `docker images` works.

.github/workflows/pull-request-docker.yml
snazy: Wanna take a stab on that?

XN137: time probably doesnt allow it anytime soon... do we already have an endpoint that would be always available without auth?

snazy: Heath checks probably.

PR #6128, opened 2023-02-17
PR Auto labeler

snazy:
A little bit more safety when opening PRs.

See https://github.com/actions/labeler

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,32 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+pr-docker:
+- 'tools/dockerbuild/**'
+- 'gradle/wrapper/**'
+- '.github/**'
+
+pr-helm:
+- 'helm/**'
+- '.github/**'
+
+pr-native:
+- 'gradle/wrapper/**'
+- 'servers/**'
+- '.github/**'
+
+pr-integrations:
+- 'gradle/wrapper/**'
+- 'clients/**'
+- '.github/**'
--- a/.github/workflows/pull-request-labeler.yml
+++ b/.github/workflows/pull-request-labeler.yml
@@ -1,0 +1,29 @@
+# Copyright (C) 2020 Dremio
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,32 @@
...
+
+pr-native:
+- 'gradle/wrapper/**'
+- 'servers/**'
+- '.github/**'
+
+pr-integrations:
+- 'gradle/wrapper/**'
+- 'clients/**'
+- '.github/**'
```
XN137, line 32: could these be more specific i.e. `.github/workflows/pull-request-helm.yml` or is there a good reason why we want to add all these labels whenever any workflow changes?

.github/workflows/pull-request-labeler.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/labeler@v4
+        with:
+          repo-token: "${{ secrets.CI_REPORTS_TOKEN }}"
```
XN137, line 29: should we set `sync-labels: true` ?

```
Whether or not to remove labels when matching files are reverted or no longer changed by the PR
```

iceberg is using it:
https://github.com/apache/iceberg/blob/master/.github/workflows/labeler.yml

.github/workflows/pull-request-labeler.yml
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Auto Labeler
+
+on:
+  pull_request_target:
```
XN137, line 18: trailing `.` looks odd, should we use the suggestion from the action README?
```
on:
- pull_request_target
```

XN137 (approved):
good idea!
change seems OK as is but please consider the suggestions/questions

.github/labeler.yml
snazy: TL;DR easier to reason about

There are also custom actions and such

.github/workflows/pull-request-labeler.yml
snazy: Was thinking about it, but what it you open a PR, add the label yourself, and just want to make sure that things you changed elsewhere still works.

.github/workflows/pull-request-labeler.yml
snazy: Just style from everywhere else.

snazy:
Side note: also thinking about making the "optional" checks mandatory. But there's no good solution from GH. If a check is mandatory but not executed ("Skipped"), then the "required check" will fail. GH's workaround is to add another WF with the same name, that runs, but does nothing, and has its filter configured in the "opposite way". https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

.github/labeler.yml
```diff
@@ -0,0 +1,32 @@
...
+pr-helm:
+- 'helm/**'
+- '.github/**'
+
+pr-native:
+- 'gradle/wrapper/**'
+- 'servers/**'
+- '.github/**'
+
+pr-integrations:
```
dimas-b, line 29: We should probably add `pr-integrations` also in case of `model` changes, WDYT?

snazy: Sure, mind opening a PR?

dimas-b: #6132

PR #6129, opened 2023-02-17
Fix typo in release notes

snazy:
A backtick needs to be escaped, otherwise it's interpreted to be a _command_ to be executed (shell script stuff).

The typo made the release notes say something wrong: "Multiplatform Java image (amd64, arm64, ppc64le, s390x): projectnessie/nessie:0.49.0-native (only recommended for short running Nessie instances, amd64 only)"


snazy:
image

XN137 (approved):

PR #6143, opened 2023-02-19
Rename jobs for mac+win GH workflows

snazy:
Otherwise the names clash with the job names of the "main" CI jobs and cause confusion in the set of the "required steps".


snazy:
image

The Windows + macOS checks should not appear as "required"

XN137 (approved):

PR #6148, opened 2023-02-20
Disable intTest using containers/podman for macOS CI

snazy:
there are issues "pulling" the images - but that's likely more an issue in GH workflows, likely memory issues

### DIFF
--- a/servers/services/build.gradle.kts
+++ b/servers/services/build.gradle.kts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import org.apache.tools.ant.taskdefs.condition.Os
+
 plugins {
   `java-library`
   jacoco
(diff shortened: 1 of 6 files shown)


servers/services/build.gradle.kts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import org.apache.tools.ant.taskdefs.condition.Os
```
XN137, line 17: does gradle bundle ant? kinda surprising that gradle has no api for this

XN137 (approved):

servers/services/build.gradle.kts
snazy: Yea - ant's included.
There is an API `org.gradle.api.Project#ant(org.gradle.api.Action<? super org.gradle.api.AntBuilder>)` et al - also a "groovy builder" within kotlin scripts for it.

PR #6152, opened 2023-02-21
mac-os/win CI - add retry for steps that regularly timeout/hang

snazy:
* macOS CI: `podman machine start`
* win CI: pulling the Hadoop tarball

Add a GH retry action with a timeout of 5 minutes + 10 attempts

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -51,20 +51,23 @@ jobs:
         path: hadoop-winutils
         ref: d018bd9c919dee1448b95519351cc591a6338a00
 
-    - name: Setup Hadoop
+    - name: Pull Hadoop
       # A local Hadoop setup is required on Windows, very sad...
-      timeout-minutes: 30
       run: |
         cd $HOME
-        Invoke-WebRequest -Uri https://archive.apache.org/dist/hadoop/common/hadoop-3.3.2/hadoop-3.3.2.tar.gz -OutFile hadoop-3.3.2.tar.gz
+        Invoke-WebRequest -Uri https://dlcdn.apache.org/hadoop/common/hadoop-3.3.2/hadoop-3.3.2.tar.gz -TimeoutSec 60 -MaximumRetryCount 10 -OutFile hadoop-3.3.2.tar.gz
 
+    - name: Setup Hadoop
+      # A local Hadoop setup is required on Windows, very sad...
+      run: |
         echo "::group::Extract hadoop-3.3.2.tar.gz"
+        cd $HOME
         mkdir hadoop-3.3.2
         cd hadoop-3.3.2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-win.yml
```diff
@@ -51,20 +51,27 @@ jobs:
...
         mkdir hadoop-3.3.2
         cd hadoop-3.3.2
         7z e ..\hadoop-3.3.2.tar.gz
         echo "::endgroup::"
-
         echo "HADOOP_HOME=$HOME\hadoop-3.3.2" >> $env:GITHUB_ENV
+
     - name: Add Hadoop winutils
       run: |
         cd hadoop-winutils\hadoop-3.2.2\bin
```
XN137, line 77: just wondering: is this intentionally referring to `3.2.2` while we download hadoop `3.3.2` ?

.github/workflows/ci-win.yml
```diff
@@ -51,20 +51,27 @@ jobs:
...
+    - name: Pull Hadoop
+      # A local Hadoop setup is required on Windows, very sad...
+      uses: nick-fields/retry@v2
+      with:
+        max_attempts: 10
+        timeout_minutes: 5
+        retry_wait_seconds: 3
+        command: |
+          cd $HOME
+          Invoke-WebRequest -Uri https://archive.apache.org/dist/hadoop/common/hadoop-3.3.2/hadoop-3.3.2.tar.gz -OutFile hadoop-3.3.2.tar.gz
```
XN137, line 63: we could use these here instead:
```
      [-MaximumRetryCount <Int32>]
      [-RetryIntervalSec <Int32>]
```
https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.3

also maybe we should prefer to download from the dlcdn instead of the archive:
https://dlcdn.apache.org/hadoop/common/hadoop-3.3.2/hadoop-3.3.2.tar.gz

XN137 (dismissed):
+1 but with suggestions/questions worth considering imo

.github/workflows/ci-win.yml
snazy: that's hadoop-winutils (see above) - no real idea, but that works

.github/workflows/ci-win.yml
snazy: updated

XN137: hmm sorry seems like the new url doesnt work ... even though `-MaximumRedirection` is 5 by default :thinking:

snazy: probably a missing `cd $HOME`

snazy: that was it

XN137 (approved):

PR #6153, opened 2023-02-21
Redesign telemetry support in Helm chart

### LINKED ISSUE #6105: Rethink telemetry support in Helm chart
Currently the Helm chart exposes the following values:

```yaml
jaegerTracing:
  # -- Specifies whether jaeger tracing for the nessie server should be enabled.
  enabled: false
  # -- The traces endpoint, in case the client should connect directly to the Collector, e.g. http://jaeger-collector:14268/api/traces
  endpoint: ""
  # -- The Jaeger service name.
  serviceName: nessie
  # -- Whether metrics are published if tracing is enabled.
  publishMetrics: true
  # -- The sampler type (const, probabilistic, ratelimiting or remote).
  samplerType: ratelimiting
  # -- The request sampling probability. 1=Sample all requests. Set samplerParam to somewhere between 0 and 1, e.g. 0.50, if you do not wish to sample all requests.
  samplerParam: 1
```

But these settings date back to pre-Open Telemetry times, and might not even function properly anymore.

adutra:
This commit removes the old jaegerTracing section from the values.yaml file. Indeed, OpenTracing support has been replaced in 0.46.0 with OpenTelemetry support, and properties like quarkus.jaeger.service.name are not honored anymore.

Instead, this commit creates a new tracing section that maps to OpenTelemetry configuration properties.

This commit also introduces some doc changes:

- In helm/nessie/README.md to explain how to test Nessie with OpenTelemetry enabled, and with custom Docker images;
- In site/docs/try/configuration.md to create a new Traces chapter with a few troubleshooting tips.

Fixes #6105

### DIFF
--- a/helm/nessie/README.md
+++ b/helm/nessie/README.md
@@ -73,12 +73,6 @@ $ helm uninstall --namespace nessie-ns nessie
 | ingress.enabled | bool | `false` | Specifies whether an ingress should be created. |
 | ingress.hosts | list | `[{"host":"chart-example.local","paths":[]}]` | A list of host paths used to configure the ingress. |
 | ingress.tls | list | `[]` | A list of TLS certificates; each entry has a list of hosts in the certificate, along with the secret name used to terminate TLS traffic on port 443. |
-| jaegerTracing.enabled | bool | `false` | Specifies whether jaeger tracing for the nessie server should be enabled. |
-| jaegerTracing.endpoint | string | `""` | The traces endpoint, in case the client should connect directly to the Collector, e.g. http://jaeger-collector:14268/api/traces |
-| jaegerTracing.publishMetrics | bool | `true` | Whether metrics are published if tracing is enabled. |
-| jaegerTracing.samplerParam | int | `1` | The request sampling probability. 1=Sample all requests. Set samplerParam to somewhere between 0 and 1, e.g. 0.50, if you do not wish to sample all requests. |
-| jaegerTracing.samplerType | string | `"ratelimiting"` | The sampler type (const, probabilistic, ratelimiting or remote). |
-| jaegerTracing.serviceName | string | `"nessie"` | The Jaeger service name. |
 | logLevel | string | `"INFO"` | The default logging level for the nessie server. |
 | mongodb.connectionString | string | `"mongodb://localhost:27017"` | The MongoDB connection string. |
 | mongodb.name | string | `"nessie"` | The MongoDB database name. |
@@ -92,7 +86,7 @@ $ helm uninstall --namespace nessie-ns nessie
 | postgres.secret.name | string | `"postgres-creds"` | The secret name to pull Postgres credentials from. |
 | postgres.secret.password | string | `"postgres_password"` | The secret key storing the Postgres password. |
 | postgres.secret.username | string | `"postgres_username"` | The secret key storing the Postgres username. |
-| replicaCount | int | `1` | The number of replicas to deploy (horizontal scaling). Beware that replicas are stateless; don't set this number > 1 when using ROCKS version store type. |
+| replicaCount | int | `1` | The number of replicas to deploy (horizontal scaling). Beware that replicas are stateless; don't set this number > 1 when using INMEMORY or ROCKS version store types. |
 | resources | object | `{}` | Configures the resources requests and limits for nessie pods. We usually recommend not to specify default resources and to leave this as a conscious choice for the user. This also increases chances charts run on environments with little resources, such as Minikube. If you do want to specify resources, uncomment the following lines, adjust them as necessary, and remove the curly braces after 'resources:'. |
 | rocksdb.selectorLabels | object | `{}` | Labels to add to the persistent volume claim spec selector; a persistent volume with matching labels must exist. Leave empty if using dynamic provisioning. |
 | rocksdb.storageClassName | string | `"standard"` | The storage class name of the persistent volume claim to create. |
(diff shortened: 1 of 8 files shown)


helm/nessie/templates/deployment.yaml
```diff
@@ -142,28 +142,27 @@ spec:
...
+              value: {{ .Values.tracing.endpoint | quote }}
+            - name: QUARKUS_OPENTELEMETRY_TRACER_RESOURCE_ATTRIBUTES
+            {{- if .Values.tracing.attributes }}
+              value: "{{- include "nessie.dictToString" .Values.tracing.attributes }}"
+            {{- else }}
+              value: "service.name={{ include "nessie.name" . }}"
+            {{- end }}
+            {{- if .Values.tracing.sampler }}
+            - name: QUARKUS_OPENTELEMETRY_TRACER_SAMPLER
+              value: {{ .Values.tracing.sampler | toString | replace "true" "on" | replace "false" "off" | quote }}
```
adutra, line 158: The calls to `replace` are required because in Yaml, `on` and `off` are actually booleans, but the server expects the strings `on` or `off`, not `true` or  `false`.

helm/nessie/README.md
```diff
@@ -189,6 +188,83 @@ This is broadly following the example from https://kubernetes.io/docs/tasks/acce
...
+```
+
+### Custom Docker images for Nessie with Minikube
+
+You can modify Nessie's code and deploy it to Minikube. Once you've made your changes, build the
+Docker image and deploy it as follows:
+
+```bash
+eval $(minikube docker-env)
+docker build -f ./tools/dockerbuild/docker/Dockerfile-jvm -t nessie-test:latest ./servers/quarkus-server
```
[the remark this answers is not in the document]
snazy: Nice to see that that these Dockerfiles help w/ k8s stuff as well.

helm/nessie/values.yaml
```diff
@@ -1,5 +1,5 @@
 # -- The number of replicas to deploy (horizontal scaling).
-# Beware that replicas are stateless; don't set this number > 1 when using ROCKS version store type.
+# Beware that replicas are stateless; don't set this number > 1 when using INMEMORY or ROCKS version store types.
```
snazy, line 2: :+1:

helm/nessie/values.yaml
```diff
@@ -97,19 +97,26 @@ authorization:
...
-  publishMetrics: true
-  # -- The sampler type (const, probabilistic, ratelimiting or remote).
-  samplerType: ratelimiting
-  # -- The request sampling probability. 1=Sample all requests. Set samplerParam to somewhere between 0 and 1, e.g. 0.50, if you do not wish to sample all requests.
-  samplerParam: 1
+  # -- The collector endpoint URL to connect to (required).
+  # The endpoint URL must have either the http:// or the https:// scheme.
+  # The collector must talk the OpenTelemetry protocol (OTLP) and the port must be its gRPC port (by default 4317).
+  endpoint: "http://otlp-collector:4317"
+  # -- The sampler to use for tracing. Valid values are: off (never sample); on (always sample);
```
snazy, line 107: Should we change it to true/false as suggested by the linter, given that the boolean's converted to `"on"` + `"off"` above?

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -217,3 +221,7 @@ quarkus.micrometer.binder.http-server.match-patterns=\
 # endless 'Retried waiting for GCLocker too often allocating * words' messages instead of a
 # "proper OutOfMemoryException" happen.
 %test.quarkus.micrometer.binder.jvm=false
+
+
+quarkus.log.category."io.smallrye.config".level=DEBUG
```
snazy, line 226: left-overs?

site/docs/try/configuration.md
```diff
@@ -128,6 +128,36 @@ The following configurations are advanced configurations to configure how Nessie
...
+### Traces
+
+Since Nessie 0.46.0, traces are published using OpenTelemetry. See [Using
+OpenTelemetry](https://quarkus.io/guides/opentelemetry) in the Quarkus documentation.
+
+In order for the server to publish its traces, the
+`quarkus.opentelemetry.tracer.exporter.otlp.endpoint` property _must_ be set. Its value must be a
+valid collector endpoint URL, with either `http://` or `https://` scheme. The collector must talk
+the OpenTelemetry protocol (OTLP) and the port must be its gRPC port (by default 3417), e.g.
+"http://otlp-collector:4317".
```
snazy, line 140: Guess, a link to https://quarkus.io/guides/opentelemetry could be helpful.

servers/quarkus-server/src/main/resources/application.properties
adutra: Whoops, yes. Good catch.

helm/nessie/values.yaml
adutra: I think you are right. The other option is to force users to always quote "on" and "off":

Option 1: `true`, `false`, `"ratio"`
Option 2: `"on"`, `"off"`, `"ratio"`

None of these are ideal, but I think option 1 is probably easier for users.

There is actually an option 3:

Option 3: `true`, `false`, `<double>`

In this case, if  `tracing.sampler` has a double value, it resolves to `tracing.sampler=ratio` internally + `tracing.ratio=<...>`.

Wdyt? I think I can live with `true`, `false`, `"ratio"`.

dimas-b: How about `sample: "all"` | `"none"` | `"0.42"`?

snazy: Either 2 or 3 are okay - not sure, if `"ratio"` and `<double>` both work.
(We need tsyaml :disappear:)

helm/nessie/README.md.gotmpl
```diff
@@ -129,6 +129,90 @@ This is broadly following the example from https://kubernetes.io/docs/tasks/acce
...
+
+If the above command fails with "failed to call webhook [...] connection refused", then the Jaeger
+Operator was not yet ready. Wait a few seconds and try again.
+
+* Install Nessie Helm chart with OpenTelemetry Collector enabled:
+
+```bash
+helm install nessie -n nessie-ns helm/nessie \
+  --set tracing.enabled=true \
+  --set tracing.endpoint=http://jaeger-collector:4317
```
dimas-b, line 171: Would it work with the `otlp/2` protocol (default port `55680`)?

snazy (dismissed):
LGTM - if @dimas-b is happy, too

helm/nessie/values.yaml
adutra: OK I will give @dimas-b suggestion a try, it does require some more parsing but should be doable.

Fun fact: obviously, this on/off thing hit some people before us:

https://github.com/quarkusio/quarkus/issues/30780#issuecomment-1416135542

:-)

adutra: Done, it's much better now, thanks @dimas-b !

helm/nessie/README.md.gotmpl
adutra: I confess I don't know, I am going to run a few tests now to investigate.

adutra: Here is what I found:

According to this issue: https://github.com/open-telemetry/opentelemetry-specification/issues/1148, it seems that in its early stages, OpenTelemetry chose to use ports 55680 and 55681 for OTLP/gRPC and OTLP/HTTP respectively.

However, as the discussion in the issue reveals, these choices weren't good ones, because these ports are in the ephemeral range, and consequently, are not suitable to be used as service ports.

So, they asked IANA to reserve instead two other ports: 4317 and 4318, the common ports that we use today for OTLP/gRPC and OTLP/HTTP respectively.

If you look at [the specs in 2020](https://github.com/open-telemetry/opentelemetry-specification/blob/cb25fe94d8d9378c04ee37d7802e7552cf5b6cc0/specification/protocol/exporter.md#configuration-options), the default endpoint for OTLP/gRPC was `localhost:55680`.

But if you look at [the specs now](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md), the default endpoint for OTLP/gRPC was changed to `http://localhost:4317`.

As a conclusion, I don't think there is any other alternate port that we should be supporting, only 4317 is nowadays being used for OTLP/gRPC.

dimas-b: nice :+1: thanks for the detailed explanation!

helm/nessie/values.yaml
```diff
@@ -97,19 +97,23 @@ authorization:
...
+  # The endpoint URL must have either the http:// or the https:// scheme.
+  # The collector must talk the OpenTelemetry protocol (OTLP) and the port must be its gRPC port (by default 4317).
+  # See https://quarkus.io/guides/opentelemetry for more information.
+  endpoint: "http://otlp-collector:4317"
+  # -- Which requests should be sampled. Valid values are: "all", "none", or a ratio between 0.0 and
+  # 1.0 (inclusive). E.g. 0.5 means that 50% of the requests will be sampled.
+  sample: all
+  # -- Resource attributes to identify the nessie service among other tracing sources.
+  # See https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#service.
+  # If left empty, the following attribute will be automatically added: service.name=nessie.
```
adutra, line 113: I am having second thoughts about this.

If `service.name` is not provided at all, it seems that Quarkus will use the value from `quarkus.application.name`, which in our case is `Nessie`. This might be enough, wdyt?

dimas-b (dismissed):

helm/nessie/values.yaml
dimas-b: Defaulting to `quarkus.application.name` SGTM.

adutra: OK I made the change. You will need to re-approve, sorry.

snazy (approved):

PR #6155, opened 2023-02-21
Minor fixes to the build-push-images.sh script

adutra:
The build-push-images.sh script has a usage function, but the current usage text is outdated. This commit makes it up-to-date with the current supported parameters.


snazy (approved):
Thanks :)

snazy:
Re-triggered the WF run - it got stuck

PR #6164, opened 2023-02-22
Disable versioned-transfer ITs in macOS CI

snazy:


XN137 (approved):

XN137:
maybe mention what the exact issues was for future reference

Issue #4066, opened 2022-04-25
Bump Delta to 1.2.0


snazy:
See #3937

dimas-b:
While refreshing the integration with DeltaLake, we ought to review the packaging of the DelataLake client, which currently apparently includes some irrelevant jars/classes, as discussed in https://github.com/projectnessie/nessie/pull/6197/files#r1123283815

snazy:
Replaced with new umbrella #6815

PR #6197, opened 2023-03-01
Maven group-ID refactoring

### LINKED ISSUE #5019: Idea: revisit Maven group IDs for published Nessie artifacts
Having Gradle projects like nessie-spark-extensions... and everything else that "requires Iceberg" (aka the projects listed in `gradle/projects.iceberg.properties`).

This would simplify the build code in NesQuEIT and make it possible to eventually release e.g. Spark SQL extensions separately in a separate build, potentially having us produce artifacts for a broader version matrix, potentially also to allow having the Nessie code in Iceberg released separately.

This issue is by far completely thought through, but I think it does not hurt to at least change the Maven group IDs for certain projects, to make a potential later change possible.

The above is also not strictly restricted to "just Iceberg projects", but maybe other projects as well.

snazy:
The current Maven group-ID "clashes" with other projects (CEL, etc), and separating included builds in NesQuEIT is tricks with the "Nessie-Iceberg" projects and the "core" Nessie projects.

This PR moves the "main" Nessie projects from `org.projectnessie` to `org.projectnessie.nessie`, and the "integrations" Nessie projects to `org.projectnessie.nessie-integrations`.

This is a somewhat breaking change and requires updating Maven corrdinates, relocation pom's will be published though.

Related/requires: https://github.com/projectnessie/query-engine-integration-tests/pull/178

Fixes #5019


snazy:
Looks like Maven relocation in this PR is working fine. Tested w/ a locally published snapshot w/ this PR and just bumping the version in Iceberg's `versions.props` + `./gradlew :iceberg-nessie:test`. Also works with the group-ID change in Iceberg's `build.gradle`+`versions.props`.

site/macros.py
```diff
@@ -39,7 +39,7 @@ def iceberg_spark_runtime(spark="3.3", scala="2.12", version=__latest_iceberg_ve
     @env.macro
     def nessie_spark_extensions(spark="3.3", scala="2.12", version=__latest_nessie_version):
         return __maven_artifact(
-            "org.projectnessie",
+            "org.projectnessie.nessie-integrations",
```
dimas-b, line 42: This will probably break docs until we release... Would it be worth moving doc change into a separate PR to be merged after the next release?

buildSrc/src/main/kotlin/Utilities.kt
```diff
@@ -123,8 +121,10 @@ fun isIntegrationsTestingEnabled() =
 fun Project.nessieClientForIceberg(): Dependency {
   val dependencyHandlerScope = DependencyHandlerScope.of(dependencies)
   if (!isIntegrationsTestingEnabled()) {
+    val clientVersion = libsRequiredVersion("nessieClientVersion")
     return dependencies.create(
-      "org.projectnessie:nessie-client:${libsRequiredVersion("nessieClientVersion")}"
+      if (clientVersion >= "0.50.1") "org.projectnessie.nessie:nessie-client:$clientVersion"
```
dimas-b, line 126: Does `>=` just work here? Will it work for `0.100.0`?

However, chances are by the time we get to 0.100 we won't have to support 0.50 and earlier :thinking:

clients/deltalake/build.gradle.kts
```diff
@@ -75,5 +75,8 @@ nessieQuarkusApp {
 forceJava11ForTests()
 
 tasks.named<ShadowJar>("shadowJar") {
-  dependencies { include(dependency("org.projectnessie:.*:.*")) }
+  dependencies {
+    include(dependency("org.projectnessie.nessie:.*:.*"))
+    include(dependency("org.projectnessie.nessie-integrations:.*:.*"))
```
dimas-b, line 80: This will include Iceberg stuff, I guess. Would it make sense to limit this to deltalake? I know it matches old behaviour, but since we're repackaging artifacts, I guess it might be worth isolating unrelated artifacts? WDYT?

dimas-b (commented):
LGTM overall. Comments below. Mostly wondering about partitioning integrations by kind.

clients/deltalake/build.gradle.kts
snazy: Idea is exactly to move everything, especially the Iceberg-related stuff, to the `org.projectnessie.nessie-integrations` group-ID

buildSrc/src/main/kotlin/Utilities.kt
snazy: 0.100 is quite some time away. But yes, otherwise it would be bad.

site/macros.py
snazy: _sigh_ yes, let me factor this to a separate PR.

snazy: #6207

clients/deltalake/build.gradle.kts
dimas-b: Is Iceberg stuff relevant to the deltalake client? Why do we bundle it into the client jar?

snazy: It's an integration as well

.github/workflows/pull-request-integ.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       NESSIE_DIR: included-builds/nessie
       NESSIE_PATCH_REPOSITORY: ''
       NESSIE_PATCH_BRANCH: ''
+      # TODO NESQUEIT_REPOSITORY: projectnessie/query-engine-integration-tests
+      # TODO NESQUEIT_BRANCH: main
```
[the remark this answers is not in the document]
snazy: Ah, sorry, nope. I've removed that change.
Just that the pr-integrations run won't work for this PR.

clients/deltalake/build.gradle.kts
dimas-b: Added a note of this to #4066 for follow-up.

dimas-b (approved):

PR #6202, opened 2023-03-02
Fix: `Content.Type` is a string

### LINKED ISSUE #6199: Typescript: Incorrect type for Entries "type" field
**What happened**:

The type property for the EntriesV1/V2 model is generated as an object.
image

Typescript example:
image

Discussion with @dimas-b indicates that this should be a string.

This can be seen from the generated Typescript client here: https://app.swaggerhub.com/apis/projectnessie/nessie/0.48.2

**Details**:
- Nessie server type (docker/built from source) and version: 0.48.2+
- Client type (Ex: UI/Spark/pynessie ...) and version: Typescript/UI

snazy:
The OpenAPI spec incorrectly exposes `Content.Type` as an object, but it is a string representing the name of the content's type. It is intentionally not declared as/restricted to an enum.

Before:
```yaml
    Type_V1:
      type: object
    Type_V2:
      type: object
```

Now:
```yaml
    Type_V1:
      type: string
    Type_V2:
      type: string
```

Fixes #6199

### DIFF
--- a/model/src/main/java/org/projectnessie/model/Content.java
+++ b/model/src/main/java/org/projectnessie/model/Content.java
@@ -48,6 +48,12 @@ public abstract class Content {
 
   @JsonDeserialize(using = Util.ContentTypeDeserializer.class)
   @JsonSerialize(using = Util.ContentTypeSerializer.class)
+  @Schema(
+      type = SchemaType.STRING,
+      description =
+          "Declares the type of a Nessie content object, which is currently one of "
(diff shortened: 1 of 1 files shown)


XN137 (approved):
lgtm

any ideas how we could catch such mistakes in our build in the future?

snazy:
> any ideas how we could catch such mistakes in our build in the future?

No, I don't have a idea.

PR #6219, opened 2023-03-04
Symlink gradle-wrapper.jar, unignore gradle-wrapper.jar

snazy:
Since gradle-wrapper.jar is in Git, it should be actively tracked as well


XN137 (approved):
do we need to commit the latest jar as well?

snazy:
It's already the latest jar

PR #6224, opened 2023-03-05
Make V2 ref path-param parsing independent of `Reference`

### LINKED ISSUE #6223: queryParam 'type' in DELETE /v2/trees/{ref} is nonnull while the swagger says it is optional.
According to the Swagger API, the type param should be optional for DELETE /v2/trees/{ref} , but in actual implementation in Util.java, it forces to have non null value. Can I get some clarification on this? Thanks!
Screen Shot 2023-03-04 at 3 33 43 PM
Screen Shot 2023-03-04 at 3 33 31 PM

snazy:
`Reference` is strongly types, it can only be exactly one of the types `Branch`, `Tag` or `Detached`. When parsing the path parameter in V2 API, the actual reference type (branch or tag or detached) is usually unknown, it _can_ be typed by the called by passing the _optional_ reference type.

This change introduces a new V2-internal tuple `ParsedReference` to allow parsing a reference path parameter without knowing the reference type. The backend service interface (`TreeService`) and implementation (`TreeApiImpl`) already allows the reference-type to be optional. A follow-up ticket might refactor the `TreeService` functions to use the new `ParsedReference` type, possibly renaming it and handling the various validation rules.

Updates `TestAuthorizationRules` and adopts the authz check for delete-reference to work like the other functions, which prefer throwing a NessieReferenceNotFoundException over NessieForbiddenException. This behavior (not-found over forbidden) is undefined though - we might change it in the future, but it would require updating the authz check interface and contract as well, because the reference-type would become optional).

Fixes #6223

### DIFF
--- a/model/src/main/java/org/projectnessie/api/v2/params/ParsedReference.java
+++ b/model/src/main/java/org/projectnessie/api/v2/params/ParsedReference.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


dimas-b (approved):
Thanks for the fix, @snazy !

PR #6243, opened 2023-03-07
New DynamoDB storage - do not pull in the Apache HTTP client

snazy:
The Apache HTTP client pulls in the (very old) commons-logging:commons-logging and log4j:log4j dependencies, which we do not want. Compare: dependencies of `nessie-versioned-persist-dynamodb`.

Otherwise the native image build w/ Quarkus 3 fails with `Caused by: java.lang.ClassNotFoundException: org.apache.log4j.Priority`

### DIFF
--- a/versioned/storage/dynamodb/build.gradle.kts
+++ b/versioned/storage/dynamodb/build.gradle.kts
@@ -46,7 +46,8 @@ dependencies {
   implementation(libs.protobuf.java)
   implementation(platform(libs.awssdk.bom))
   implementation(libs.awssdk.dynamodb) { exclude("software.amazon.awssdk", "apache-client") }
-  implementation(libs.awssdk.apache.client)
+  implementation(libs.awssdk.netty.nio.client)
+  implementation(libs.awssdk.url.connection.client)
 
   compileOnly(libs.testcontainers.testcontainers)
   compileOnly(libs.docker.java.api)
--- a/versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoClientProducer.java
+++ b/versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoClientProducer.java
@@ -18,7 +18,7 @@ package org.projectnessie.versioned.storage.dynamodb;
 import java.net.URI;
 import org.immutables.value.Value;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
-import software.amazon.awssdk.http.apache.ApacheHttpClient;
+import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClientBuilder;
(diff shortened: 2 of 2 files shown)


versioned/storage/dynamodb/build.gradle.kts
```diff
@@ -46,7 +46,8 @@ dependencies {
   implementation(libs.protobuf.java)
   implementation(platform(libs.awssdk.bom))
   implementation(libs.awssdk.dynamodb) { exclude("software.amazon.awssdk", "apache-client") }
-  implementation(libs.awssdk.apache.client)
+  implementation(libs.awssdk.netty.nio.client)
+  implementation(libs.awssdk.url.connection.client)
```
XN137, line 50: currently on `main`:
```
➜  nessie git:(main) rg awssdk.apache.client
gradle/libs.versions.toml
83:awssdk-apache-client = { module = "software.amazon.awssdk:apache-client" }

versioned/storage/dynamodb/build.gradle.kts
49:  implementation(libs.awssdk.apache.client)

servers/quarkus-common/build.gradle.kts
58:  implementation(libs.awssdk.apache.client) { exclude("commons-logging", "commons-logging") }

servers/lambda/build.gradle.kts
42:  implementation(libs.awssdk.apache.client)
43:  implementation(libs.awssdk.apache.client) { exclude("commons-logging", "commons-logging") }
```

a) line 42 in lambda module should be removed probably?

b) `exclude(x, x)` -> `exclude(x)`

c) why is it OK for the other modules to keep pulling in `libs.awssdk.apache.client` but only `dynamodb` needs to change? to me it would make more sense to add the same exclusions to `dynamodb` as in the other modules and keep using the more advanced http client (especially since this PR says we do this for quarkus, but `quarkus-common` keeps pulling in the apache http client

c) wondering if in gradle versions catalog we could manage the exclusions for `libs.awssdk.apache.client` in a single place instead of at every usage site?
this would then probably also affect transitive usage automatically

for example we use this in one of our maven builds:
```
    <dependencyManagement>
      <dependency>
        <groupId>org.apache.httpcomponents</groupId>
        <artifactId>httpclient</artifactId>
        <version>4.5.14</version>
        <exclusions>
          <exclusion>
            <artifactId>commons-logging</artifactId>
            <groupId>commons-logging</groupId>
          </exclusion>
        </exclusions>
      </dependency>
      ...
    </dependencyManagement>
```
and it affects all direct and transitive usage

snazy: This PR is _only_ to align this module - not to update any other module. That's #6245

XN137: could you answer my question in b) ?
to me it's not immediately how this aligns things when other modules keep pulling the apache client

XN137 (approved):
+1 if we only want to align with `versioned/persist/dynamodb/build.gradle.kts`

but if we want to avoid pulling the apache http client for awssdk as suggested in the PR description then we are missing a few things imo

versioned/storage/dynamodb/build.gradle.kts
snazy: I did?

> b) why is it OK for the other modules to keep pulling in libs.awssdk.apache.client 

That's about to be tackled in #6245 - would appreciate if you can take over that PR.

And in the PR description

> Otherwise the native image build w/ Quarkus 3 fails with

XN137: no time atm but if you can file a ticket with a clear description of the goal (i.e. is it just to get rid of commons-logging/log4j or of the apache http client across all nessie modules?) then maybe in 1-2 weeks i can take a look

snazy: See description of that PR #6245 : `Let AWS SDK only use the URL connection client`

PR #6253, opened 2023-03-08
Fix `AbstractDatabaseAdapter.removeKeyCollisionsForNamespaces`

snazy:
The (current) database-adapter storage code has functionality that ignores merge/transplant conflicts for namespaces (#3572). The function that "removes" conflicting keys for namespaces was always called with the wrong source-commit-id, it was called with the first commit after the common ancestor of the source branch, not with the HEAD of the branch to merge.

A test will be implicitly added via #6246, which discovered this bug.


adutra (approved):
LGTM assuming that a test will cover this later on.

PR #6256, opened 2023-03-08
Add helper functions to `Namespace` and `ContentKey()`

snazy:
* `ContentKey.getElementCount()`
* `ContentKey.getParent()`
* `Namespace.getElementCount()`
* `Namespace.getParent()`
* `Namespace.of(ContentKey)`
* `Namespace.toContentKey()`

### DIFF
--- a/model/src/main/java/org/projectnessie/model/ContentKey.java
+++ b/model/src/main/java/org/projectnessie/model/ContentKey.java
@@ -61,6 +61,12 @@ public abstract class ContentKey implements Comparable<ContentKey> {
     return getElements().toArray(new String[0]);
   }
 
+  @JsonIgnore
+  @Value.Redacted
+  public int getElementCount() {
+    return getElements().size();
+  }
+
   /**
    * Returns the namespace that is always consisting of the first <b>N-1</b> elements from {@link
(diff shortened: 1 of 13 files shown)


model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -173,9 +179,17 @@ protected void check() {
...
       throw new IllegalStateException("Key too long, max allowed length: " + MAX_LENGTH);
     }
-    if (getElements().size() > MAX_ELEMENTS) {
+    if (getElementCount() > MAX_ELEMENTS) {
       throw new IllegalStateException(
           "Key too long, max allowed number of elements: " + MAX_ELEMENTS);
     }
   }
+
+  public ContentKey toParent() {
```
adutra, line 188: Nit: `toParent` gives the impression that the ContentKey is being transformed into its parent. I think `getParent` is a bit more suitable.

adutra (dismissed):
The PR is still in draft state, but I'm leaving an early review. Thanks for separating the commits, it makes the review much easier!

snazy:
Only rebased, not changed

PR #6263, opened 2023-03-10
Test changes for 'Ensure parent namespace(s) for commited content objects exist' PR

snazy:
Test code changes for #6246

### DIFF
--- a/clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
+++ b/clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
@@ -18,6 +18,7 @@ package org.apache.iceberg.nessie;
 import static org.apache.iceberg.types.Types.NestedField.required;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.projectnessie.model.CommitMeta.fromMessage;
 
 import com.google.common.collect.ImmutableMap;
 import java.io.IOException;
@@ -44,6 +45,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.projectnessie.error.NessieConflictException;
+import org.projectnessie.error.NessieNamespaceAlreadyExistsException;
 import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Branch;
 import org.projectnessie.model.CommitMeta;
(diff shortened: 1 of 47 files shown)


clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
```diff
@@ -75,7 +80,19 @@ public TestNessieIcebergViews() {
...
+  public void beforeEach() throws Exception {
+    try {
+      api.commitMultipleOperations()
+          .branchName(branch)
+          .commitMeta(fromMessage(CREATE_REQUIRED_NAMESPACES))
+          .operation(Put.of(ContentKey.of("db"), Namespace.of("db")))
+          .operation(Put.of(ContentKey.of("nessie"), Namespace.of("nessie")))
+          .operation(Put.of(ContentKey.of("nessie", "db"), Namespace.of("nessie", "db")))
+          .commit();
+    } catch (NessieNamespaceAlreadyExistsException e) {
```
adutra, line 92: Since we are creating 3 namespaces in the same commit: what happens if `db` exists but not `nessie` nor `nessie.db`? Don't we risk leaving the catalog in an inconsistent state?

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/AbstractContentGeneratorTest.java
```diff
@@ -117,6 +121,11 @@ int getExitCode() {
     List<String> getStdOutLines() {
       return Arrays.asList(stdOut.split("\n"));
     }
+
+    @Override
+    public String toString() {
+      return "ProcessResult{" + "exitCode=" + exitCode + ", stdOut='" + stdOut + '\'' + '}';
```
adutra, line 127: Nit: unnecessary string concatenations.

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITDeleteContent.java
```diff
@@ -53,7 +53,7 @@ void deleteContent() throws NessieNotFoundException {
             CONTENT_KEY.getElements().get(0),
             "--key",
             CONTENT_KEY.getElements().get(1));
-    assertThat(proc.getExitCode()).isEqualTo(0);
```
adutra, line 56: I see this change in several places, but I'm failing to see why we need it.

adutra (changes requested):

tools/content-generator/src/test/java/org/projectnessie/tools/contentgenerator/ITDeleteContent.java
snazy: `proc` has a nice `toString()` - so the whole information is printed when it fails

clients/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
snazy: The tests would then fail. For prod code, it should be much safer, sure, but not really for test code w/ a known clean state.

adutra (approved):

PR #6266, opened 2023-03-10
Allow some integration tests to be executed on macOS

adutra:
These tests work perfectly well on modern macs.


snazy:
Or we introduce a new env-var `CI_MAC`  or so in our CI workflows for Mac and check that one.

testing/s3minio/src/main/java/org/projectnessie/minio/MinioExtension.java
```diff
@@ -51,9 +51,15 @@ public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext con
     if (OS.current() == OS.LINUX) {
       return enabled("Running on Linux");
     }
+    if (OS.current() == OS.MAC) {
+      if (System.getenv("CI_MAC") == null) {
```
snazy, line 55: Nit: collapse the 2 `if`s

snazy (approved):

snazy (approved):

PR #6286, opened 2023-03-13
Build/CI: Allow renovate to update the Maven wrapper

snazy:
Renovate cannot update the Maven Wrapper automatically, because the built-tools-integration-tests/pom.xml requires the `nessie.version` system property to be set. This change adds some Nessie version to let Renovate do its job.

### DIFF
--- a/build-tools-integration-tests/pom.xml
+++ b/build-tools-integration-tests/pom.xml
@@ -28,6 +28,7 @@
     <maven.compiler.release>8</maven.compiler.release>
     <versionMavenResolver>1.7.3</versionMavenResolver>
     <versionMaven>3.8.7</versionMaven>
+    <nessie.version>0.51.1</nessie.version>
   </properties>
 
   <dependencyManagement>


build-tools-integration-tests/pom.xml
```diff
@@ -28,6 +28,7 @@
     <maven.compiler.release>8</maven.compiler.release>
     <versionMavenResolver>1.7.3</versionMavenResolver>
     <versionMaven>3.8.7</versionMaven>
+    <nessie.version>0.51.1</nessie.version>
```
XN137, line 31: where does this maven build property get translated to a system property at test runtime?

i can see it being done in `build.gradle` but not clear for maven:
https://github.com/projectnessie/nessie/blob/01baaef484bd6f17be3688a6a32e2fa530dbe250/build-tools-integration-tests/build.gradle#L53-L54

also wondering: worth to find a maven plugin that does this as well?
https://github.com/projectnessie/nessie/blob/01baaef484bd6f17be3688a6a32e2fa530dbe250/build-tools-integration-tests/build.gradle#L29

XN137 (approved):

PR #6288, opened 2023-03-13
Core reorg: use `api/` and `integrations/` directories

snazy:
Moves `nessie-model` + `nessie-client` (+ test extension) to `api/`, moves remaining modules in `clients/` to `integrations/`. This makes it clear that `integrations/` contains the Nessie integrations and client/public-API facing modules in `api/`.

As a follow-up it's possible to factor-out the HTTP/REST specific parts from `nessie-client` into e.g. `nessie-client-rest` and open the door for other formats.

(No code change in this PR.)


dimas-b (commented):
Looks reasonable to me :+1:

dimas-b (approved):

PR #6299, opened 2023-03-15
Workaround UI issues in ghcr.io + quay.io

snazy:
Both ghcr.io + quay.io show "unknown/unknown" architectures for Nessie's multi-arch images. quay.io one for each "real" architecture, ghcr.io one "unknown" architecture. This is not a big deal, but rather confusing to see. This change shall work around that issue.


snazy:
quay.io:
image

ghcr.io:
image

dimas-b (approved):

PR #6305, opened 2023-03-16
Replace Docker Hub with ghcr.io

### LINKED ISSUE #6293: Move away from Docker Hub
Docker terminates all free plans soon.

Current options:
* Github registry
* quay.io

Both options are free to use. Maybe we should just publish to both, but only "advertise" one of those in the Readme.

Migration can be easily done using the [regctl](https://github.com/regclient/regclient) tool  - just need "some" network bandwidth to actually copy the images around.

Either each image individually using `regctl image copy <source_image:tag> <target_image:tag>` or, probably much easier, via [`regsync`](https://github.com/regclient/regclient#regsync-features).

Should also update:
* Docker badge in the `README.md`
* Links on the website

snazy:
... and mirror to quay.io + docker.io (as long as it works)

Fixes #6293

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -129,13 +129,13 @@ jobs:
         mv servers/quarkus-cli/build/nessie-quarkus-cli-${RELEASE_VERSION}-runner.jar "${ARTIFACTS}"
         echo "::endgroup::"
 
-        echo '${{ secrets.DOCKER_TOKEN }}' | docker login -u '${{ secrets.DOCKER_USERNAME }}' --password-stdin
+        echo '${{ secrets.GITHUB_TOKEN }}' | docker login ghcr.io -u $ --password-stdin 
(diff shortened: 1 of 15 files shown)


.github/workflows/release-publish.yml
```diff
@@ -129,13 +129,13 @@ jobs:
         mv servers/quarkus-cli/build/nessie-quarkus-cli-${RELEASE_VERSION}-runner.jar "${ARTIFACTS}"
         echo "::endgroup::"
 
-        echo '${{ secrets.DOCKER_TOKEN }}' | docker login -u '${{ secrets.DOCKER_USERNAME }}' --password-stdin
+        echo '${{ secrets.GITHUB_TOKEN }}' | docker login ghcr.io -u $ --password-stdin
```
dimas-b, line 132: Just `$` for the `-u` option?

snazy: yep

PR #6313, opened 2023-03-16
Forbid special ASCII chars in content keys

### LINKED ISSUE #6311: Possible sorting mismatch between `ContentKey` and `StoreKey`
Test case:
```
    ContentKey k1 = ContentKey.of("b", "b", "c");
    ContentKey k2 = ContentKey.of("b", "b\1a");
    int c1 = k1.compareTo(k2);
    int c2 = keyToStoreKey(k1).compareTo(keyToStoreKey(k2));
```
Result:
```
c1 = -2
c2 = 2
```

dimas-b:
This is to ensure correctness with conversion to/from StoreKey.

Fixes #6311

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/ContentKey.java
+++ b/api/model/src/main/java/org/projectnessie/model/ContentKey.java
@@ -16,8 +16,7 @@
 package org.projectnessie.model;
 
 import static org.projectnessie.model.Util.DOT_STRING;
-import static org.projectnessie.model.Util.GROUP_SEPARATOR_STRING;
-import static org.projectnessie.model.Util.ZERO_BYTE_STRING;
+import static org.projectnessie.model.Util.FIRST_ALLOWED_KEY_CHAR;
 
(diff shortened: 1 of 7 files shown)


api/model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -171,10 +172,12 @@ protected void validate() {
         throw new IllegalArgumentException(
             String.format("Content key '%s' must not contain a null element.", elements));
       }
-      if (e.contains(ZERO_BYTE_STRING) || e.contains(GROUP_SEPARATOR_STRING)) {
```
snazy, line 174: Wonder whether we should forbid all chars < 0x20 - guess, those don't make any sense at all.

dimas-b: updated

snazy (approved):

PR #6332, opened 2023-03-20
New storage: fix writetamp collision in integration tests

snazy:
We must ensure that the system clock advances a little, so that C*'s next write-timestamp does not collide with the write-timestamps from an "erase repository" operation.  In C*, if a DELETE and another INSERT/UPDATE have the same write-timestamp, the DELETE wins. This makes Nessie tests fail on machines that are "fast enough".

### DIFF
--- a/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraPersist.java
+++ b/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraPersist.java
@@ -539,6 +539,16 @@ public class CassandraPersist implements Persist {
         requests.submitted(backend.executeAsync(ERASE_OBJ, repoId, objId));
       }
     }
+    // We must ensure that the system clock advances a little, so that C*'s next write-timestamp
+    // does not collide with the write-timestamps of the DELETE statements above. Otherwise, the
+    // above DELETEs will silently "overrule" a following INSERT/UPDATE statement. In C*, if a
+    // DELETE and another INSERT/UPDATE have the same write-timestamp, the DELETE wins. This makes
+    // Nessie tests fail on machines that are "fast enough".
+    try {
+      Thread.sleep(2L);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+    }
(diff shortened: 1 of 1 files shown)


versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraPersist.java
```diff
@@ -539,6 +539,16 @@ public void erase() {
         requests.submitted(backend.executeAsync(ERASE_OBJ, repoId, objId));
       }
     }
+    // We must ensure that the system clock advances a little, so that C*'s next write-timestamp
+    // does not collide with the write-timestamps of the DELETE statements above. Otherwise, the
+    // above DELETEs will silently "overrule" a following INSERT/UPDATE statement. In C*, if a
+    // DELETE and another INSERT/UPDATE have the same write-timestamp, the DELETE wins. This makes
+    // Nessie tests fail on machines that are "fast enough".
+    try {
+      Thread.sleep(2L);
```
dimas-b, line 548: Would it make sense to spin a bit until the clock advances?... just for the sake of clarity.

snazy: Don't think so. We can't "sync" to "the database's system clock" anyway.

dimas-b: Well, is 2ms enough to compensate for clock differences? 

Anyway, this kind of collision use case is not very probable outside of tests, I think, and as long as 2ms fixes tests, I'm ok :)

dimas-b (approved):

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraPersist.java
snazy: Yea - this is mostly a test-fix (and 2ms shouldn't harm any real use case).

PR #6346, opened 2023-03-21
Properly return `ParamConverterProvider` errors as HTTP/400

### LINKED ISSUE #6333: Invalid refrence type specification in REST API leads to 500 responses instead of 400
```
Request method:	DELETE
Request URI:	http://localhost:38481/v2/trees/test%402e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d?type=X
Proxy:			<none>
Request params:	<none>
Query params:	<none>
Form params:	<none>
Path params:	<none>
Headers:		Accept=application/json, application/javascript, text/javascript, text/json
				Content-Type=application/json
Cookies:		<none>
Multiparts:		<none>
Body:			<none>

HTTP/1.1 500 Internal Server Error
Vary: Accept-Encoding
Content-Encoding: gzip
Content-Type: application/json
Connection: close
Content-Length: 236

{
    "status": 500,
    "reason": "Internal Server Error",
    "message": "org.glassfish.jersey.server.ParamException$QueryParamException: HTTP 404 Not Found, caused byjava.lang.IllegalArgumentException: No enum constant org.projectnessie.model.Reference.ReferenceType.X",
    "errorCode": "UNKNOWN",
    "serverStackTrace": null
}
```

This is clearly a user error - the response should have a 4xx status code instead,

snazy:
Fixes #6333

### DIFF
--- a/servers/rest-services/src/main/java/org/projectnessie/services/restjakarta/NessieExceptionMapper.java
+++ b/servers/rest-services/src/main/java/org/projectnessie/services/restjakarta/NessieExceptionMapper.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.JsonMappingException;
 import com.google.common.base.Throwables;
 import jakarta.inject.Inject;
 import jakarta.ws.rs.NotSupportedException;
+import jakarta.ws.rs.WebApplicationException;
 import jakarta.ws.rs.core.Response;
 import jakarta.ws.rs.ext.Provider;
 import java.security.AccessControlException;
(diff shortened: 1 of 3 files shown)


servers/rest-services/src/main/java/org/projectnessie/services/restjakarta/NessieExceptionMapper.java
```diff
@@ -82,13 +83,18 @@ public Response toResponse(Exception exception) {
     } else if (exception instanceof NotSupportedException) {
       errorCode = ErrorCode.UNSUPPORTED_MEDIA_TYPE;
       message = exception.getMessage();
+    } else if (exception instanceof WebApplicationException
```
snazy, line 86: Sadly a "complicated" condition here, so it works with both Quarkus (RESTeasy) and Jersey.
Didn't want to have different code paths for javax + jakarta.

dimas-b (approved):
Thanks for the quick fix, @snazy ! :+1:

PR #6355, opened 2023-03-22
Use relocated protobuf runtime

snazy:
Generated protobuf code must be run with the protobuf runtime using the exact same version as the generator used. This is currently a soft constraint, but it is likely that protobuf will enforce that in future protobuf versions. Using a different version of the protobuf runtime can (and does) lead to issues at runtime, see [this issue for an example](https://github.com/protocolbuffers/protobuf/issues/11986).

This change adds a new Nessie artifact `nessie-protobuf-relocated` which contains a relocated protobuf runtime that is used by all other Nessie artifacts.

Also:
* Move `*.proto` files into the directories corresponding to the Java package names.
* Remove the `ProtobufHelperPlugin`, as the underlying issue (not exposing the generated java source folders in IDEs) has been fixed.
* Do not apply checkstyle to Gradle projects that just generate protobuf code (the `ProtobufHelperPlugin` did that before).

Side note: CEL-Java heavily depends on protobuf itself via its API and not just as a runtime/compile-only dependency. As long as no generated protobuf code is being used, it _should_ be okay to use a different protobuf version at runtime, but there is and will be no guarantee.


snazy:
Review notes: The PR contains multiple commits:
* `Use relocated protobuf runtime` contains the "actual change"
* The other 4 commits rather "boring" changes

snazy:
Another note: Quarkus itself (or better: some extensions) also pull in protobuf in the exact same version that Quarkus requires via the `quarkus-bom`.

tools/protobuf-relocated/build.gradle.kts
```diff
@@ -0,0 +1,53 @@
...
+  `java-library`
+  `maven-publish`
+  signing
+  id("org.projectnessie.buildsupport.reflectionconfig")
+  `nessie-conventions`
+}
+
+apply<NessieShadowJarPlugin>()
+
+extra["maven.name"] = "Nessie - Relocated Protobuf ${libs.protobuf.java.get().version}"
```
dimas-b, line 29: nit: Shall we state the full protobuf's Maven coordinates here?

buildSrc/src/main/kotlin/Checkstyle.kt
```diff
@@ -29,6 +29,10 @@ import org.gradle.kotlin.dsl.withType
 class NessieCheckstylePlugin : Plugin<Project> {
   override fun apply(project: Project): Unit =
     project.run {
+      if (project.name.endsWith("-proto")) {
+        return
```
dimas-b, line 33: Why not run checkstyle on proto modules?

snazy: it's generated code

snazy: (Try it, if your terminal's scroll buffer is huuuuuge ;) ;) )

tools/protobuf-relocated/build.gradle.kts
snazy: hm - it's just a name though

buildSrc/src/main/kotlin/Checkstyle.kt
dimas-b: > it's generated code

Could you add a comment about that?

What about `.proto` files themselves?

tools/protobuf-relocated-gen/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+  // which contains the the "com.google.protobuf:protobuf-java" dependency.
+  includeConfigurations.set(listOf("runtimeClasspath"))
+  relocations.put("com[.]google[.]protobuf[.]", "org.projectnessie.nessie.relocated.protobuf.$1")
+}
+
+val shadowJar =
+  tasks.named<ShadowJar>("shadowJar") {
+    dependsOn("generateReflectionConfig")
+    relocate("com.google.protobuf", "org.projectnessie.nessie.relocated.protobuf")
+    manifest { attributes["Google-Protobuf-Version"] = libs.protobuf.java.get().version }
```
dimas-b, line 54: How about something like this?
`Specification-Title` = `Google Protobuf`,
`Specification-Version` = `libs.protobuf.java.get().version`,
`Implementation-Title` = `Nessie Protobuf Relocated`
`Implementation-Version` = nessie version

buildSrc/src/main/kotlin/Checkstyle.kt
snazy: Checkstyle is Java only

dimas-b (approved):

snazy:
Ugh - turned out that the published Gradle module metadata produced via the recommended way is broken. Created #6361 to fix it.

dimas-b (dismissed):

PR #6363, opened 2023-03-23
Mongo/storage: use `replaceOne` instead of `findAndReplaceOne`

snazy:
Difference is to _not_ use a "find" (aka "scan") since we have the full key available.


adutra (approved):
LGTM – although I had to look up the docs for the `replaceOne` method to make sure it doesn't insert a new doc when zero docs match the filter – which is the case.

PR #6384, opened 2023-03-24
Persist: replace `updateObj()` with `upsertObj()`

snazy:
The only purpose of `updateObj()` is to allow the Nessie repository import to enhance the freshly imported `CommitObj`s, which are marked with `incompleteIndex==true`, with properly populated incremental and reference indexes.

During the import operation, we do not really care, whether the object to be updated (the commit) does not exist, because that will not be the case, because the import already stored all the commits. TL;DR the behavior to check whether the `Obj` passed to `updateObj()` already exists is rather meaningless.

Using "upsert" semantics has the big benefit that those can be batched, which is not only nice but a big performance improvement during import.

### DIFF
--- a/versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
+++ b/versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
@@ -91,6 +91,8 @@ import com.mongodb.bulk.BulkWriteResult;
 import com.mongodb.client.FindIterable;
 import com.mongodb.client.MongoCursor;
 import com.mongodb.client.model.InsertOneModel;
+import com.mongodb.client.model.ReplaceOneModel;
+import com.mongodb.client.model.ReplaceOptions;
 import com.mongodb.client.model.WriteModel;
 import com.mongodb.client.result.DeleteResult;
 import com.mongodb.client.result.UpdateResult;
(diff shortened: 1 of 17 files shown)


versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -579,26 +581,49 @@ public void deleteObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids) {
...
+    UpdateResult result =
+        backend.objs().replaceOne(eq(ID_PROPERTY_NAME, idObjDoc(id)), doc, options);
+    if (!result.wasAcknowledged()) {
+      throw new RuntimeException("Upsert not acknowledged");
     }
   }
 
+  private static ReplaceOptions upsertOptions() {
+    // A `ReplaceOneModel` with the default replace options (upsert==false) silently does just
+    // nothing.
```
dimas-b, line 600: nit: nothing when there's no matching row?

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -579,26 +581,49 @@ public void deleteObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids) {
...
-    // MongoDB does not support bulk conditional updates
+  public void upsertObjs(@Nonnull @jakarta.annotation.Nonnull Obj[] objs)
+      throws ObjTooLargeException {
+    ReplaceOptions options = upsertOptions();
+
+    List<WriteModel<Document>> docs = new ArrayList<>(objs.length);
     for (Obj obj : objs) {
-      updateObj(obj);
+      if (obj != null) {
+        ObjId id = obj.id();
```
dimas-b, line 614: nit: Shall we `checkArgument(id != null, ...)` as in `upsertObj()`?

dimas-b (approved):

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
snazy: You can't have an `Obj` with a `null` `id`

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
snazy: No, Mongo does litereally nothing with upsert==false - just comes back with an "empty" result

PR #6385, opened 2023-03-24
New storage: add "batch write" facade

snazy:
Adds a `Persist` implementation that batches "store" and "update" operations.

This change leverages the "upsert" capability introduced by #6384 for "Nessie's repository import" to perform bulk writes, replaces the previous `Obj` bulk write and introduces bulk-updates for the new storage model's finalization phase. This change, in combination with a "large-ish" value (e.g. 500) for the import's `--commit-batch-size` parameter, brings a extremely huge improvement. For a local test case, the duration to "finalize" 1000 commits was ~5.2 seconds, with this change down to <0.2 seconds per 1000 commits.


snazy:
Coding done - PR depends on #6384

dimas-b (approved):

PR #6431, opened 2023-03-29
Nit: fix IntelliJ import warning for nessie-perftest-simulations

snazy:
IJ complains about Scala-library not present (it's there for the 'gatling' source set).

### DIFF
--- a/perftest/simulations/build.gradle.kts
+++ b/perftest/simulations/build.gradle.kts
@@ -28,6 +28,10 @@ plugins {
 extra["maven.name"] = "Nessie - Perf Test - Simulations"
 
 dependencies {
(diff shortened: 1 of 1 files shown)


perftest/simulations/build.gradle.kts
```diff
@@ -28,6 +28,9 @@ plugins {
 extra["maven.name"] = "Nessie - Perf Test - Simulations"
 
 dependencies {
+  if (System.getProperty("idea.sync.active").toBoolean()) {
```
adutra, line 31: Maybe add a comment to explain why we are doing this?

adutra (approved):

PR #6433, opened 2023-03-29
Testing: add `-XX:+HeapDumpOnOutOfMemoryError`

snazy:


buildSrc/src/main/kotlin/Testing.kt
```diff
@@ -87,6 +87,7 @@ class NessieTestingPlugin : Plugin<Project> {
       tasks.withType<Test>().configureEach {
         val testJvmArgs: String? by project
         val testHeapSize: String? by project
+        jvmArgs("-XX:+HeapDumpOnOutOfMemoryError")
```
dimas-b, line 90: Will the dump be saved in CI runs? Is the expected benefit for local runs only?

snazy: It is (was) for local use at the moment.

dimas-b (approved):

PR #6434, opened 2023-03-29
JDK 11 HTTP Client - use common ForkJoinPool by default

snazy:
Lets Nessie's JDK11 HTTP client by default use the common fork-join-pool to prevent tests from piling up thread-pools, which is just not necessary. Usually, the common fork-join-pool is "good enough" and the use of a separate pool is not necessary.


snazy:
Actually, that's not an issue - even in tests. Closing this one.

PR #6438, opened 2023-03-29
Remove the need to pass `Put.expectedContent()`

snazy:
`expectedContent` was originally introduced for global-state, which has
been removed a while ago.

The field `expectedContent` is still present and used for REST API v1,
but excluded for REST API v2.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Operation.java
+++ b/api/model/src/main/java/org/projectnessie/model/Operation.java
@@ -18,6 +18,7 @@ package org.projectnessie.model;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import javax.annotation.Nullable;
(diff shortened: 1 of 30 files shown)


api/model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -30,6 +30,12 @@
 @Schema(
     type = SchemaType.OBJECT,
     title = "Operation",
+    description =
+        "Describes an operation to be performed against one content object.\n"
+            + "\n"
+            + "The Nessie backend will perform validations against the correctness of the "
+            + "operations. Note that the amount and strictness of validations depends on the Nessie "
```
dimas-b, line 37: nit: `The Nessie backend will validate the correctness of the operations.`

api/model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -82,6 +105,15 @@ static Put of(ContentKey key, Content content, Content expectedContent) {
...
 
+  @Schema(
+      type = SchemaType.OBJECT,
+      title = "Delete-'Content'-operation for a 'ContentKey'.",
+      description =
+          "Used to delete an existing content key.\n"
+              + "\n"
+              + "If the key for a content shall change (aka a rename), then use a `Delete` "
+              + "operation using the current (old) key and a `Put` operation using the new key "
+              + "populating both the `value` and `expectedValue` fields. See `Put` operation.")
```
dimas-b, line 116: The `populating both the `value` and `expectedValue` fields` part seem outdated now that `expected value` is deprecated?

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractCommits.java
```diff
@@ -535,10 +527,10 @@ public void commitConflictingOperations() throws Exception {
             () ->
                 commit("Conflicting Commit")
                     .fromReference(initialCommit)
-                    .put("t3", V_3_2.withId(t3), t3)
+                    .put("t3", V_3_2.withId(t3))
                     .toBranch(branch))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage("Key 't3' does not exist, but Put-operation has expectedValue");
+        .hasMessage("New value for new must not have a content iD");
```
dimas-b, line 533: nit: the message looks odd to me :shrug:

dimas-b (approved):
LGTM overall :+1: Minor comments, can be addressed in a follow-up PR.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractCommits.java
dimas-b: Maybe quote the key in the message? This particular message is just hard to read :)

dimas-b: Also `iD` starts with a lower case ;)

dimas-b: #6450

PR #6441, opened 2023-03-30
Re-enable CassandraBackend (fixes #6389)

### LINKED ISSUE #6389: Re-enable C*/Scylla in Nessie Quarkus
The code to use C* or Scylle in Nessie has been disabled since it was not clear whether the C* integration in Quarkus will still be available in Quarkus 3. Once it's clear that Quarkus 3 still has C* support, we can re-enable the code to use C*/Scylla.

adutra:


servers/quarkus-tests/build.gradle.kts
```diff
@@ -43,6 +43,8 @@ dependencies {
   implementation(libs.testcontainers.mongodb)
   implementation(libs.docker.java.api)
 
+  implementation(libs.cassandra.quarkus.tests)
```
snazy, line 46: Maybe exclude the testcontainers-cassandra dependency from this one and add 
```
implementation(libs.testcontainers.cassandra)
```
to avoid potential version conflicts.

snazy:
Hm - looks like `:nessie-quarkus:intTest` is somehow broken now - it complains about "port already used" (although we do not have static port assignments). Maybe because `CassandraTestResource` does not respect `io.quarkus.test.common.DevServicesContext.ContextAware#setIntegrationTestContext`?

servers/quarkus-tests/build.gradle.kts
adutra: In this case I think we should use the cassandra-quarkus BOM artifact and also exclude the driver, wdyt?

snazy: Quarkus-BOM's already in there as an enforcedPlatform dependency.

gradle/libs.versions.toml
```diff
@@ -94,8 +95,9 @@ bouncycastle-bcpkix = { module = "org.bouncycastle:bcpkix-jdk15on", version.ref
 bouncycastle-bcprov = { module = "org.bouncycastle:bcprov-jdk15on", version.ref = "bouncycastle" }
 cassandra-driver-bom = { module = "com.datastax.oss:java-driver-bom", version.ref = "cassandraDriver" }
 cassandra-driver-core = { module = "com.datastax.oss:java-driver-core" }
-cassandra-quarkus = { module = "com.datastax.oss.quarkus:cassandra-quarkus-client", version = "1.1.3" }
-cassandra-quarkus-tests = { module = "com.datastax.oss.quarkus:cassandra-quarkus-test-framework", version = "1.1.3" }
+cassandra-quarkus-bom = { module = "com.datastax.oss.quarkus:cassandra-quarkus-bom", version.ref = "cassandraQuarkus" }
```
snazy, line 98: Oh, that bom, okay

adutra:
Closing in favor of #6440.

PR #6445, opened 2023-03-30
IntelliJ: Always use Gradle to run tests

snazy:
IJ sometimes gets configurations and classpathes for tests wrong from the Gradle configuration, so better let IJ always run tests via Gradle, which will always work.


adutra (approved):
A long awaited improvement, afaic :-)

dimas-b:
In my experience running / debugging individual test methods from parameterized tests via the IJ Gradle runner is not always possible (unlike the "native" IJ test runner).

PR #6451, opened 2023-03-31
python: move safety check to separate tox env

XN137:
by not installing safety alongside the other dependencies, we are not constrained by safety's overly restrictive dependency version ranges (see inline comment)

this unblocks renovate to upgrade other linting dependencies

### DIFF
--- a/python/tox.ini
+++ b/python/tox.ini
@@ -15,13 +15,13 @@
 #
 
 [tox]
-envlist = py37, py38, py39, py310, format, lint, docs, docs_sync_check
+envlist = py37, py38, py39, py310, format, lint, safety, docs, docs_sync_check
 
 [gh-actions]
 python =
     3.7: py37
     # docs needs to run before docs_sync_check
-    3.8: py38, lint, docs, docs_sync_check
+    3.8: py38, lint, safety, docs, docs_sync_check
     3.9: py39
     3.10: py310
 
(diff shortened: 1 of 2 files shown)


python/tox.ini
```diff
@@ -43,12 +43,19 @@ commands =
     # flake8 includes black check due to flake8-black
     # flake8 includes isort check which checks for import order due to flake8-isort
     flake8 pynessie tests tools
-    # ignore https://pyup.io/v/51457/f17 -> https://github.com/pytest-dev/py/issues/287
```
XN137, line 46: no longer needed since https://github.com/pytest-dev/pytest/releases/tag/7.2.0
because pytest no longer depends on py

python/tox.ini
```diff
@@ -43,12 +43,19 @@ commands =
...
     mypy --install-types --non-interactive -p pynessie -p tests -p tools
 
 
+[testenv:safety]
+basepython = python
+deps = safety
+commands =
+    # using separate env because of https://github.com/pyupio/safety/issues/455
+    # note that requirements_lint.txt imports all other requirement files
+    safety check --file requirements_lint.txt
```
XN137, line 56: from CI run:
```
     Safety v2.3.5 is scanning for Vulnerabilities...
    Scanning dependencies in your files:
  
    -> requirements_lint.txt
  
    Using non-commercial database
    Found and scanned 39 packages
    Timestamp 2023-03-31 07:50:58
    0 vulnerabilities found
    0 vulnerabilities ignored
  +==============================================================================+
  
   No known security vulnerabilities found. 
```

this might exclude scanning of transitive dependencies, but i guess its still better than flat out removing safety until they manage to release their new version (that blocks our other upgrade PRs)

PR #6469, opened 2023-04-03
Bring back logback logging

snazy:
After the bump to SLF4J v2, logging via the "old" logback-classic no longer worked due to https://www.slf4j.org/codes.html#ignoredBindings. This change adds the recommended slf4j-simple dependency to re-enable the "old" logback-classic.

### DIFF
--- a/api/client/build.gradle.kts
+++ b/api/client/build.gradle.kts
@@ -76,7 +76,7 @@ dependencies {
   testFixturesApi(libs.awssdk.auth)
   testFixturesApi(libs.undertow.core)
   testFixturesApi(libs.undertow.servlet)
-  testFixturesImplementation(libs.logback.classic)
+  testFixturesImplementation(libs.bundles.logback.logging)
 }
 
 jandex { skipDefaultProcessing() }
--- a/compatibility/common/build.gradle.kts
+++ b/compatibility/common/build.gradle.kts
@@ -34,7 +34,7 @@ dependencies {
 
   implementation(platform(libs.jersey.bom))
   api(libs.slf4j.api)
-  api(libs.logback.classic)
+  api(libs.bundles.logback.logging)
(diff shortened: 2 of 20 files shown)


XN137:
afaik `slf4j-simple` is its own (trivial) logging backend... using it does not re-enable logging through logback.

```
The current actively developed version of logback supporting Java EE (java.* namespace) is 1.3.6. It requires SLF4J version 2.0.4 and JDK 8.
```
https://logback.qos.ch/download.html

i think we would need to upgrade logback to 1.3.x in order to let its binding be discovered by slf4j 2.0 ?

imo when we upgraded to slf4j 2.0 we did not consider all "downstream" implications ...

snazy:
It works in practice, so?

XN137:
> It works in practice, so?

logback and `slf4j-simple` are both logging backends for slf4j...
using 2 logging backends in the same project does not make sense imo?

https://www.slf4j.org/manual.html#swapping

```
In your code, in addition to slf4j-api-2.0.7.jar, you simply drop one and only one binding of your choice onto the appropriate class path location.
Do not place more than one binding on your class path.
```

as mentioned before the reason why logback logging broke with the slf4j api 2.0 upgrade is because we did not upgrade logback to a slf4j-2.0-compatible version (i.e. 1.3.x).
so this imo does not justify adding a different backend (even though of course, logging then still works, it's just not going through logback now)

snazy:
So, what's the right way then? Can you provide a solution?

XN137:
> So, what's the right way then?

if we want to stick with slf4j 2.0, then the right way would be to upgrade logback to `1.3.6`

snazy:
I mean, can you come up with a PR - that would help a lot

dimas-b (approved):

PR #6472, opened 2023-04-03
Don't retain a squashed PUT-DELETE sequence

adutra:
This commit fixes a small inconsistency between old and new storage models during merge and transplant: in the old model, if the commits to squash contained a PUT followed by a DELETE for the same key, the squashed commit would mistakenly retain the DELETE operation.

The tests included in this PR do not pass for the old model, without the fix.

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
@@ -22,6 +22,7 @@ import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.tuple;
 import static org.assertj.core.api.InstanceOfAssertFactories.list;
 import static org.junit.jupiter.api.Assumptions.assumeFalse;
 import static org.projectnessie.model.CommitMeta.fromMessage;
@@ -1240,30 +1241,51 @@ public abstract class BaseTestNessieApi {
 
     ContentKey key1 = ContentKey.of("ns", "t1");
     ContentKey key2 = ContentKey.of("ns", "t2");
-    ContentKey key3 = ContentKey.of("ns", "t3");
-    ContentKey key4 = ContentKey.of("ns", "t4");
+    ContentKey key3a = ContentKey.of("ns", "t3a");
+    ContentKey key3b = ContentKey.of("ns", "t3b");
 
-    IcebergTable t1a = IcebergTable.of("t1", 1, 1, 1, 1);
-    IcebergTable t2a = IcebergTable.of("t2", 1, 1, 1, 1);
-    IcebergTable t2b = IcebergTable.of("t2", 2, 2, 2, 2);
-    IcebergTable t3a = IcebergTable.of("t3", 1, 1, 1, 1);
+    IcebergTable t1 = IcebergTable.of("t1", 1, 1, 1, 1);
+    IcebergTable t2a = IcebergTable.of("t2a", 1, 1, 1, 1);
+    IcebergTable t2b = IcebergTable.of("t2b", 2, 2, 2, 2);
+    IcebergTable t3 = IcebergTable.of("t3", 1, 1, 1, 1);
 
     // Add 3 commits to the source branch:
-    // C1: PUT t1=v1.1, PUT t2=v2.1, PUT t3=v3.1
-    // C2: DELETE t1, DELETE t2, DELETE t3, PUT t4=v3.1 (rename t3 => t4)
-    // C3: PUT t2=v2.2
 
+    // C1: PUT key1=t1, PUT key2=t2a, PUT key3a=t3
     source =
-        prepCommit(source, "C1", Put.of(key1, t1a), Put.of(key2, t2a), Put.of(key3, t3a)).commit();
+        prepCommit(source, "C1", Put.of(key1, t1), Put.of(key2, t2a), Put.of(key3a, t3)).commit();
+
+    Map<ContentKey, Content> contents =
+        api().getContent().refName(source.getName()).key(key3a).get();
+    assertThat(contents).hasSize(1).containsKey(key3a);
+    t3 = (IcebergTable) contents.get(key3a);
+
+    // C2:
+    // - DELETE key1
+    // - DELETE key2 (to be re-created in the next commit)
+    // - DELETE key3a, PUT key3b=t3 (rename key3a => key3b, same payload t3)
     source =
         prepCommit(
-                source, "C2", Delete.of(key1), Delete.of(key2), Delete.of(key3), Put.of(key4, t3a))
+                source, "C2", Delete.of(key1), Delete.of(key2), Delete.of(key3a), Put.of(key3b, t3))
             .commit();
+
+    // C3: PUT key2=t2b (re-create key2)
     source = prepCommit(source, "C3", Put.of(key2, t2b)).commit();
 
     api().mergeRefIntoBranch().fromRef(source).branch(target).keepIndividualCommits(false).merge();
 
-    // Expected operations in the squashed commit: PUT t2=v2.2, PUT t4=v3.1 (rename t3 => t4)
+    contents = api().getContent().refName(source.getName()).keys(Arrays.asList(key2, key3b)).get();
+    assertThat(contents).hasSize(2).containsKeys(key2, key3b);
+    IcebergTable t2final = (IcebergTable) contents.get(key2);
+    IcebergTable t3final = (IcebergTable) contents.get(key3b);
+
+    assertThat(t2final).usingRecursiveComparison().ignoringFields("id").isEqualTo(t2b);
+    assertThat(t3final).usingRecursiveComparison().ignoringFields("id").isEqualTo(t3);
+
+    // Expected operations in the squashed commit:
+    // PUT key2=t2b (same key re-created with different content)
+    // PUT key3b=t3 (renamed to different key, same content)
+
     LogEntry logEntry =
         api()
             .getCommitLog()
@@ -1273,24 +1295,11 @@ public abstract class BaseTestNessieApi {
             .get()
             .getLogEntries()
             .get(0);
+
     soft.assertThat(logEntry.getOperations())
         .extracting("key", "content")
         .hasSize(2)
-        .satisfiesExactlyInAnyOrder(
-            t -> {
-              soft.assertThat(t.toArray()[0]).isEqualTo(key2);
-              soft.assertThat(t.toArray()[1])
-                  .usingRecursiveComparison()
-                  .ignoringFields("id")
-                  .isEqualTo(t2b);
-            },
-            t -> {
-              soft.assertThat(t.toArray()[0]).isEqualTo(key4);
-              soft.assertThat(t.toArray()[1])
-                  .usingRecursiveComparison()
-                  .ignoringFields("id")
-                  .isEqualTo(t3a);
-            });
+        .containsExactlyInAnyOrder(tuple(key2, t2final), tuple(key3b, t3final));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -2091,8 +2091,9 @@ protected CommitLogEntry squashCommits(
       CommitLogEntry source = commitsToMergeChronological.get(i);
       for (ContentKey delete : source.getDeletes()) {
         if (includeKeyPredicate.test(delete)) {
-          deletes.add(delete);
-          puts.remove(delete);
+          if (puts.remove(delete) == null) {
+            deletes.add(delete);
```
dimas-b, line 2095: What about a `Put/Delete` pair in the same (single) commit? It may be meaningful when content IDs in the deleted and added objects differ.

dimas-b: I suppose we ought to retain both the `Delete` and the `Put` operations in this case (according to the latest OpenAPI descriptions updates in #6438).

Would you mind adding a test case to see how it works?

adutra: Hmm how can you achieve that? I tried:

1) DELETE and PUT in the same source commit:

```java
    commit("C1").put(key1, V_1_1).put(key2, V_2_1).put(key3, V_3_1).toBranch(source);
    commit("C2").delete(key1).delete(key2).delete(key3).put(key3, V_3_1.withId("changed")).toBranch(source);
    Hash sourceHead = commit("C3").put(key2, V_2_2).toBranch(source);
```

This gives me the error (new model):

```
java.lang.IllegalArgumentException: Duplicate key in commit operations: t3
```

2) DELETE and PUT in two subsequent source commits, with a manually assigned content ID:

```java
    commit("C1").put(key1, V_1_1).put(key2, V_2_1).put(key3, V_3_1).toBranch(source);
    commit("C2").delete(key1).delete(key2).delete(key3).toBranch(source);
    Hash sourceHead = commit("C3").put(key2, V_2_2).put(key3, V_3_1.withId("changed")).toBranch(source);
```

This gives me the error (new model):

```
java.lang.IllegalArgumentException: New value for key 't3' must not have a content ID
```

dimas-b: Re: Snipped 1: perhaps the old model does not follow (or cannot follow?) the latest API spec?

adutra: Oooh are you referring to this?

> "If the key for a content shall change (aka a rename), then use a `Delete` "
              + "operation using the current (old) key and a `Put` operation using the new key "
              + "with the `value` having the correct content-id. Both operations must happen "
              + "in the same commit.\n"

I think this is the opposite case: it's a rename. I can add a test case for that.

adutra: @dimas-b I added a rename test case in 294345c.

dimas-b: That's a good test/use case to keep :+1: 

However, I meant this:
```
              + "A content object can be replaced (think: `DROP TABLE xyz` + `CREATE TABLE xyz`) "
              + "with a `Delete` operation and a `Put` operation for a content using a `value`"
              + "representing a new content object, so without a content-id, in the same commit.")
```

adutra: Ah that's clearly not possible, neither in the old nor the new model, regardless of the changes in this PR. The PUT must follow the DELETE in a separate commit.

I think we should amend the API specs, wdyt?

dimas-b: Yeah, let's fix the API spec (in a separate PR maybe)?

dimas-b: The `Duplicate key in commit operations` error looks like a bug in the new model to me (`CommitImpl`). I think API permits `Put/Delete` pairs for the same key in a single commit.

dimas-b: Looking at the current API spec/docs I think we should actually keep `Put/Delete` pairs in the squashed commit, just maybe use the most recent `Put` in this case. WDYT?

adutra: > The Duplicate key in commit operations error looks like a bug in the new model to me (CommitImpl). I think API permits Put/Delete pairs for the same key in a single commit.

The old model does not allow it either, see [here](https://github.com/projectnessie/nessie/blob/nessie-0.54.0/versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java#L322-L327).

Looking at the Git history of this file, we see that this check was added in #2311. The PR description by @snazy is enlightening:

    While thinking about how Nessie could actually squash multiple commits, I realized
    that this can result in multiple commit-operations for the same content-key.

    While commit-operations are ordered on the (REST) API level (list of Operations),
    database adapters split those into separate collections of delete and put operations,
    which makes it impossible to determine the original order of the squashed operations.

    The Nessie spec does not define the behavior of a Nessie commit with multiple
    operations for the same key in one commit.

    The spec now clearly states that there must be no more than [one] operation (put / delete)
    for each content key.

The above PR introduced [this sentence](https://github.com/projectnessie/nessie/blob/nessie-0.54.0/site/docs/develop/spec.md?plain=1#L144) in `site/docs/develop/spec.md`: "A Content Key must only occur once in a Nessie commit." The sentence is still there.

> Looking at the current API spec/docs I think we should actually keep Put/Delete pairs in the squashed commit, just maybe use the most recent Put in this case. WDYT?

How would we know in which order the operations should be applied? PUT then DELETE, or DELETE then PUT? Allowing the same key to appear more than once would only be feasible if we could determine the order in which operations in the same commit should be applied.

All in all, I think the sentence introduced recently [here](https://github.com/projectnessie/nessie/blob/nessie-0.54.0/api/model/src/main/java/org/projectnessie/model/Operation.java#L75-L77) is wrong and should be amended.

>  It might be worth adding API-level tests for this too (BaseTestNessieApi)

Sure I will add some.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -1221,4 +1221,73 @@ public void fetchEntriesByNamelessReference() throws BaseNessieClientServerExcep
             immutableEntry(a, Content.Type.ICEBERG_TABLE),
             immutableEntry(b, Content.Type.ICEBERG_VIEW));
   }
+
+  /**
+   * This test verifies that a merge with squash strategy correctly merges operations targeting the
+   * same key, and that the result commit does not contain duplicate keys or spurious operations.
+   */
+  @Test
+  public void commitMergeSquashSameKeys() throws Exception {
```
adutra, line 1230: I added a test for merge, as requested. I wanted to also add a test for transplant, but realized that the API does not allow transplants with squash strategy. The version store layer does accept that though.

adutra: Done.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
snazy: There are two scenarios:
1. rename (Delete + Put w/ previous content-ID, must not change payload)
2. re-create (Detlete + Put w/o content-ID, can change payload)

There are test cases on the VersionStore level that validate that though (I _think_ I added tests for these cases).

Yes, the sentence that a single key must only appear once in a commit is correct - except for the rename & re-create cases.

snazy:
Haven't looked into the code yet (so apologies if my question is wrong):
For these two commits, the "key" would still be deleted @ commit 2, correct?
1. PUT("key")
2. PUT("key") + DELETE("key")

snazy:
Oh, damn

snazy:
> > Haven't looked into the code yet (so apologies if my question is wrong): For these two commits, the "key" would still be deleted @ commit 2, correct?
> > 
> > 1. PUT("key")
> > 2. PUT("key") + DELETE("key")
> 
> PUT("key") + DELETE("key") in the same commit is currently forbidden as per #2311.

I think, we have to allow DELETE + PUT, because that's how renames (and squashed re-creates) work.
I'm actually surprised, because there are tests that exercise rename scenarios (in Iceberg, too).

adutra:
@dimas-b and @snazy are there any issues remaining to be addressed?

snazy:
Let me think about the scenario tomorrow with a fresh brain - too much context switching lately :(

snazy:
Not sure whether the actual change in `AbstractDatabaseAdapter` is correct. 

The following sequence seems to work:
```
  "main"
    |
    |
  Commit 1: PUT "key"
    |\
    | \
    |  --------------------
    |                      \
    |                       |
    |                    Commit 1: DELETE "key"
    |                       |
    |                      /
    |  --------------------
    | /
    |/
  Merge operation
  (correct) merge-commit content DELETE "key"
```

The following sequence seems incorrect, because the added `if (puts.remove(delete) == null)` yields `false`, but the intent of the merged changes is to DELETE "key":
```
  "main"
    |
    |
  Commit 1: PUT "key"
    |\
    | \
    |  --------------------
    |                      \
    |                       |
    |                    Commit 2: PUT "key"
    |                       |
    |                       |
    |                    Commit 3: DELETE "key"
    |                       |
    |                      /
    |  --------------------
    | /
    |/
  Merge operation
  (incorrect) merge-commit content (empty)
```

adutra:
You are correct, I did not think about this situation.

But unfortunately, without the change in this PR, the below situation becomes incorrect with the old model (but not with the new one):

```
  "main"
    |
    |
  Commit 1: unrelated
    |\
    | \
    |  --------------------
    |                      \
    |                       |
    |                    Commit 2: PUT "key"
    |                       |
    |                       |
    |                    Commit 3: DELETE "key"
    |                       |
    |                      /
    |  --------------------
    | /
    |/
  Merge operation
  (incorrect) merge-commit content DELETE "key"
```

We would need to check whether `main` contains the key "key" in order to decide whether to keep the operation DELETE "key" in the merge commit.

I don't see a simple way to do this in the old storage model. I suggest that we close this PR and acknowledge that old and new storage models do not yield the same merge commit contents in this situation.

I guess we can live with it? A spurious DELETE "key" operation in the commit history of main is certainly wrong, but is it harmful after all? We are deleting a non-existent content so ¯\_(ツ)_/¯

snazy:
> ...
>
> I don't see a simple way to do this in the old storage model. I suggest that we close this PR and acknowledge that old and new storage models do not yield the same merge commit contents in this situation.
> 
> I guess we can live with it? A spurious DELETE "key" operation in the commit history of main is certainly wrong, but is it harmful after all? We are deleting a non-existent content so ¯_(ツ)_/¯

I think a "stale DELETE" is fine. It's effectively a no-op, so all good.

adutra:
Fair enough. Closing then.

PR #6481, opened 2023-04-03
CI-Workflow updates

snazy:
* Use `github.ref` for the `concurrency` group, hoping that previous WF run works with that (see comment in this PR below).
* The `Ci Success` step doesn't need to be a matrix job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ on:
 concurrency:
   # PRs: 1 CI run concurrently
   # main branch: 1 CI run per commit (so every commit is checked)
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.head_commit.id }}-${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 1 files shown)


snazy:
GitHub concurrency group cancelled the _wrong_ runs - GH did not cancel "older" runs, but the newer ones. This might also be a bug at GitHub though.

image

dimas-b (approved):

snazy:
Oh my, GH handling of `concurrency`/`cancel-in-progress` is ... weird. If GH can submit a job (aka not bound by the max number of concurrent WF jobs), it's (most of the time) correct and cancels _older_ jobs.
However, it seems that either when the concurrent WF job limit is reached, or the WF runs are submitted "too quick" (same timestamp?), GH cancels the _wrong_ WFs/jobs, leaving PRs without checks and lettings the WFs run with the wrong state (PR labels).

Background: Each label addition emits an event, which then triggers a WF run. So if a PR is created and auto-labeler decides to add 4 labels, the PR workflow(s) get triggered 5 times - and only the latest should run. This is most often the case, but sadly not always.

Moving this PR back to "draft" to come up with a properly working solution.

PR #6485, opened 2023-04-04
Bump jaxb-impl from hadoop-common

snazy:
Fixes this error:
```
java.lang.reflect.InaccessibleObjectException: Unable to make protected final java.lang.Class java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int) throws java.lang.ClassFormatError accessible: module java.base does not "opens java.lang" to unnamed module @19e4653c
...
    at com.sun.xml.bind.v2.runtime.reflect.opt.Injector$1.run(Injector.java:177)
```

### DIFF
--- a/gc/gc-iceberg-files/build.gradle.kts
+++ b/gc/gc-iceberg-files/build.gradle.kts
@@ -42,6 +42,9 @@ dependencies {
     exclude("org.eclipse.jetty")
     exclude("org.apache.zookeeper")
   }
+  // Bump the jabx-impl version 2.2.3-1 via hadoop-common to make it work with Java 17+
+  implementation(libs.jaxb.impl)
+
   implementation(libs.iceberg.core)
   implementation(libs.iceberg.aws)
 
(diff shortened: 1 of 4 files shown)


gc/gc-iceberg-files/build.gradle.kts
```diff
@@ -42,6 +42,9 @@ dependencies {
     exclude("org.eclipse.jetty")
     exclude("org.apache.zookeeper")
   }
+  // Bump the jabx-impl version 2.2.3-1 via hadoop-common to make it work with Java 17+
```
adutra, line 45: The comment seems wrong? In gradle/libs.versions.toml, the version is 2.3.8, not 2.2.3-1.

adutra: Oh I see, you probably meant "bump from" not "bump to" :-)

snazy: well, the dependency comes "via" hadoop-common ;)

adutra (approved):

PR #6480, opened 2023-04-03
Add min-version + spec-version-bag to `NessieConfiguration`

snazy:

PR #6486, opened 2023-04-04
Let errorprone not check code generated by APTs

snazy:
Errorprone fails for a recent change for #6480, complaining about an unnecessary null-check in generated immutables code (something like `int x; Map.put(Objects.requireNonNull(x), ...)`). The check itself is fine, however the code is generated, so there's not much we can do.

This change just adds the temporary directory in which errorprone works.

### DIFF
--- a/buildSrc/src/main/kotlin/Errorprone.kt
+++ b/buildSrc/src/main/kotlin/Errorprone.kt
@@ -73,7 +73,7 @@ class NessieErrorpronePlugin : Plugin<Project> {
             )
 
         options.errorprone.checks.putAll(checksMapProperty)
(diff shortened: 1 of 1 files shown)


adutra:
I am not able to reproduce the issue using #6480. In which module/class is it happening for you?

snazy:
> I am not able to reproduce the issue using #6480. In which module/class is it happening for you?

Yea - the code in that PR has changed :disappear:
Errorprone checks the `build/tmp/compile*Java/compileTransaction/*` directories as well, because that's where APTs operate.

adutra (approved):

PR #6505, opened 2023-04-05
CI: Save 15m in CI NesQuEIT

snazy:
Disables the unused local-repo publication


snazy:
CI is fine - GH just messed up the WF run association :disappointed: 
(explanation here https://github.com/projectnessie/nessie/pull/6481#issuecomment-1495861141)

PR #6515, opened 2023-04-06
CI: Cache everything that's needed

snazy:
The new CI workflow accidentally does not cache dependencies, generated jars and more, only the build-cache was saved.

Not having the dependencies available leads to unnecessary network traffic and sporadic build issues (dependencies not downloaded).


snazy:
Reference on my fork: https://github.com/snazy/nessie/actions/runs/4629040327/jobs/8189213688

dimas-b (approved):

PR #6516, opened 2023-04-06
Gradle: eliminate some more `doFirst`/`doLast` script references

snazy:


integrations/spark-antlr-grammar/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/NessieSqlExtensions.g4
```diff
@@ -15,6 +15,10 @@
  */
 grammar NessieSqlExtensions;
 
+@header {
+// PACKAGE_PLACEHOLDER
```
snazy, line 19: This makes the `ReplaceInFiles` idempotent

gc/gc-iceberg-files/build.gradle.kts
```diff
@@ -78,9 +78,9 @@ dependencies {
 
 tasks.withType(Test::class.java).configureEach {
   systemProperty("aws.region", "us-east-1")
+  val tmpdir = project.buildDir.resolve("tmpdir")
```
snazy, line 81: Moves the reference to `project` out of _task execution_, which is mandatory to enable Gradle's configuration cache.

buildSrc/src/main/kotlin/Testing.kt
```diff
@@ -81,12 +81,9 @@ class NessieTestingPlugin : Plugin<Project> {
       }
       if (plugins.hasPlugin("io.quarkus")) {
         // This directory somehow disappears... Maybe some weird Quarkus code.
-        tasks.named("quarkusGenerateCodeTests") {
```
snazy, line 84: Such changes elimniate the use of script references, which do not work with Gradle's configuration cache (not serializable)

dimas-b (approved):

PR #6520, opened 2023-04-06
CI: finally the correct artifacts

snazy:
Only add the really needed files to the per-job-cache-diff artifacts.

Gradle cache content is finally correct and cacheable/cached data is properly used and speeds up the builds significantly.


snazy:
See https://github.com/snazy/nessie/actions/runs/4631791927/usage

PR #6527, opened 2023-04-07
OAuth2 authentication provider for Nessie client

adutra:


api/client/src/main/java/org/projectnessie/client/http/impl/RequestContextImpl.java
```diff
@@ -72,15 +72,30 @@ public Method getMethod() {
...
+
   @Override
   public Optional<Object> getBody() {
     return body == null ? Optional.empty() : Optional.of(body);
   }
 
+  @Override
+  public boolean isFormEncoded() {
+    return getMethod() == Method.POST
+        && getContentType().isPresent()
```
snazy, line 91: Nit: `getOrElse("").equals(..."`

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -44,6 +44,44 @@ public final class NessieConfigConstants {
    * authentication.
    */
   public static final String CONF_NESSIE_AUTH_TOKEN = "nessie.authentication.token";
+
+  /**
+   * Required. The URL of the OAuth2 token endpoint, including the path to the token endpoint when
```
snazy, line 49: Can you add the `{@value #CONF_...}` javadoc content? Helps to have the config key handy in docs (if we ever have some automatic way to generate a page out of this one).

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -44,6 +44,44 @@ public final class NessieConfigConstants {
...
+  /** Required. The client ID to use when using OAuth2 authentication. */
+  public static final String CONF_NESSIE_OAUTH_CLIENT_ID = "nessie.authentication.oauth.client-id";
+
+  /** Required. The client secret to use when using OAuth2 authentication. */
+  public static final String CONF_NESSIE_OAUTH_CLIENT_SECRET =
+      "nessie.authentication.oauth.client-secret";
+
+  /**
+   * The default token validity when using OAuth2 authentication. Defaults to PT1H (one hour). Must
+   * be a valid ISO-8601 duration.
```
snazy, line 65: Shall we link to `https://en.wikipedia.org/wiki/ISO_8601#Durations` ?

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -44,6 +44,44 @@ public final class NessieConfigConstants {
...
+
+  /** Required. The client ID to use when using OAuth2 authentication. */
+  public static final String CONF_NESSIE_OAUTH_CLIENT_ID = "nessie.authentication.oauth.client-id";
+
+  /** Required. The client secret to use when using OAuth2 authentication. */
+  public static final String CONF_NESSIE_OAUTH_CLIENT_SECRET =
+      "nessie.authentication.oauth.client-secret";
+
+  /**
+   * The default token validity when using OAuth2 authentication. Defaults to PT1H (one hour). Must
```
snazy, line 64: *sigh* I suspected, that the standard(s) require that an expiration time is sent with the token - but it's only recommended :(

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
```diff
@@ -0,0 +1,242 @@
...
+ * tokens, using pre-defined client credentials.
+ *
+ * <p>This client also supports refreshing access tokens using both the refresh token method defined
+ * in RFC 6749 and the token exchange method defined in RFC 8693.
+ *
+ * <p>If you don't need to refresh access tokens, you can use the {@link
+ * org.projectnessie.client.auth.BearerAuthenticationProvider BearerAuthenticationProvider} instead
+ * and provide the token to use directly through configuration.
+ */
+public class OAuth2Client implements OAuth2Authenticator, Closeable {
```
snazy, line 49: I suspect we should let `NessieAuthenticator` extend `AutoCloseable` w/ a default noop and close associated authenticators when the client's closed

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2ClientParams.java
```diff
@@ -0,0 +1,161 @@
...
+    // authorization server MUST support the HTTP Basic authentication scheme
+    // for authenticating clients that were issued a client password.
+    HttpAuthentication authentication =
+        BasicAuthenticationProvider.create(
+            getClientId(), new String(getClientSecret(), StandardCharsets.UTF_8));
+    authentication.applyToHttpClient(builder);
+    return builder;
+  }
+
+  default void check() {
```
snazy, line 103: Add `@Value.Check` and omit the call in `OAuth2Client`?

snazy (commented):
Really nice PR!
Left a bunch of comment - got too curious ;)

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2ClientParams.java
```diff
@@ -0,0 +1,161 @@
...
+          thread.setDaemon(true);
+          thread.setName("oauth-client");
+          return thread;
+        });
+  }
+
+  @Value.Default
+  default HttpClient.Builder getHttpClient() {
+    HttpClient.Builder builder =
+        HttpClient.builder()
```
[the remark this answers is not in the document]
adutra: >  if the user forces the use of the "old" URLConnection client.

That's already supported: if the user forces the old client, the old client will be used here as well.

> configured with additional options for TLS/SSL and/or timeouts

First there would be a chicken-and-egg problem here. But besides that, my fear is that the options won't necessarily be the same for this client and the main client:

1. Base uri and authentication are not the same;
2. Timeouts could be different;
3. SSL could have different cipher suites, etc.

I think we will need a whole bunch of new settings under `nessie.authentication.oauth2.*` if we want to fine-tune the OAuth client.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
adutra: That was my first intention, but then I backed off because the changes are a bit invasive: you need to "bubble up" the call to `close()` up to all `NessieApi` impls (they are already `AutoCloseable`, but nothing was enforcing that). But if you are OK with that, let's do it.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
```diff
@@ -0,0 +1,242 @@
...
+    objectMapper = params.getObjectMapper();
+  }
+
+  @Override
+  public AccessToken authenticate() {
+    return token;
+  }
+
+  public void start() {
+    Runtime.getRuntime().addShutdownHook(new Thread(this::close));
```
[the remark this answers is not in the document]
adutra: I'm going down the "AutoCloseable" path, so I believe we won't need this hook anymore, which is way more elegant.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2ClientParams.java
```diff
@@ -0,0 +1,161 @@
...
+    // for authenticating clients that were issued a client password.
+    HttpAuthentication authentication =
+        BasicAuthenticationProvider.create(
+            getClientId(), new String(getClientSecret(), StandardCharsets.UTF_8));
+    authentication.applyToHttpClient(builder);
+    return builder;
+  }
+
+  default void check() {
+    if (getClientId().isEmpty()) {
```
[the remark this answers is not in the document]
adutra: There is a few methods already in `HttpUtils` btw.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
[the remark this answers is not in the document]
adutra: It must be visible by `OAuth2AuthenticationProvider` which is in the parent package. I will move the provider here then.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
```diff
@@ -0,0 +1,242 @@
...
+      LOGGER.error("Failed to fetch new access token", e);
+    }
+    return null;
+  }
+
+  public AccessToken refreshAccessToken(AccessToken oldToken) {
+    AccessToken newToken;
+    if (oldToken != null) {
+      // Try the RFC 6749 refresh token method first.
+      newToken = refreshAccessTokenRfc6749(oldToken);
```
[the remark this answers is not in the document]
adutra: I don't have experience enough to tell. I will play a bit with keycloak and revisit this later.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
snazy: Yea, `NessieApi` instances should always be properly closed from call sites. I'm (mostly) sure, that we always do that.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk8/UrlConnectionClient.java
```diff
@@ -54,4 +54,9 @@ public HttpRequest newRequest() {
   public URI getBaseUri() {
     return config.getBaseUri();
   }
+
+  @Override
```
snazy, line 58: If it makes sense, feel free to make `HttpClient` an abstract base class

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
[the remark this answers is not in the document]
snazy: _sigh_ Just noting: I realized that Iceberg's `Catalog` itself does not extend `(Auto)Closeable`, but a bunch of `Catalog` implementations do implement `(Auto)Closeable` (e.g. `NessieCatalog` or `RESTCatalog`). I suspect that we have to bite the bullet and live with the fact that `Catalog`s and in turn OAuth clients do not get properly closed.

Maybe we need to implement something in a follow-up - just "brain dumping" (no need to take it seriously):
* Track the last use time of the OAuth token
* It some time elapsed (couple minutes?), stop refreshing and stop the thread pool
* It the OAuth client is used again, restart the thread pool if necessary

I suspect, we need some mechanism like that in the future, but it's probably out of scope of this PR.

api/client/src/test/resources/logback-test.xml
```diff
@@ -27,4 +27,6 @@
     <level value="${test.log.level:-INFO}"/>
     <appender-ref ref="console"/>
   </root>
+  <logger name="com.github.dockerjava" level="WARN"/>
+  <logger name="org.projectnessie.client.auth.oauth2" level="DEBUG"/>
```
dimas-b, line 31: nit: Do we still need debug logs in STDOUT?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
```diff
@@ -0,0 +1,81 @@
...
+      this.authenticator = authenticator;
+    }
+
+    @Override
+    public void applyToHttpClient(HttpClient.Builder client) {
+      client.addRequestFilter(
+          ctx -> {
+            AccessToken token = authenticator.authenticate();
+            ctx.putHeader(
+                "Authorization", capitalize(token.getTokenType()) + " " + token.getPayload());
```
dimas-b, line 62: Would it be possible to add a test case for this to `AbstractOpenIdAuthentication` (if token type can be something other than `Bearer`)?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  static String capitalize(String tokenType) {
+    if (tokenType == null || tokenType.isEmpty()) {
+      return "";
+    }
+    if (tokenType.length() == 1) {
+      return tokenType.toUpperCase();
+    }
+    return tokenType.substring(0, 1).toUpperCase() + tokenType.substring(1).toLowerCase();
```
dimas-b, line 79: I suppose it is safer to use `.toNNNCase(Locale.ROOT)`. We do not expect non-ASCII chars there, do we? At the same time some fancy locales can take chars out of the ASCII range on case conversion.

dimas-b (commented):
Still not done with a full review... Posting some preliminary comments :)

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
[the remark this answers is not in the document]
adutra: That was my suspicion, and it's unfortunate indeed. In the meantime, I removed the shutdown hook completely. If that's OK let's indeed explore solutions in a follow-up PR as this one is already quite big.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk8/UrlConnectionClient.java
adutra: For now it's just one method that needs implementation in each concrete impl: `close()`, so I think it's not (yet) justified to make this change.

api/client/src/test/resources/logback-test.xml
adutra: Sorry that was a leftover.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
adutra: To be fair, I don't think it can be something else than `(B|b)earer`. I wonder if we shouldn't just remove this `capitalize` stuff and print `Bearer` all the time.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
adutra: It's only ASCII as per RFC 7230. Thus `Locale.ROOT` is fine.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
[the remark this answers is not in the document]
snazy: Sure, let's do that independently of this PR. Mind opening an issue for it?

gradle/libs.versions.toml
```diff
@@ -210,9 +210,11 @@ spark-sql-v33-v212 = { module = "org.apache.spark:spark-sql_2_12", version = { s
 spark-sql-v33-v213 = { module = "org.apache.spark:spark-sql_2_13", version = { strictly = "[3.3, 3.4[", prefer = "3.3.1"}}
 testcontainers-cassandra = { module = "org.testcontainers:cassandra", version.ref = "testcontainers" }
 testcontainers-cockroachdb = { module = "org.testcontainers:cockroachdb", version.ref = "testcontainers" }
+testcontainers-junit = { module = "org.testcontainers:junit-jupiter", version.ref = "testcontainers" }
 testcontainers-mongodb = { module = "org.testcontainers:mongodb", version.ref = "testcontainers" }
 testcontainers-postgresql = { module = "org.testcontainers:postgresql", version.ref = "testcontainers" }
 testcontainers-testcontainers = { module = "org.testcontainers:testcontainers", version.ref = "testcontainers" }
+testcontainers-keycloak = { module = "com.github.dasniko:testcontainers-keycloak", version = "2.5.0" }
```
snazy, line 217: Nit: move up a bit (alphabetical order)

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
snazy: (Except that Iceberg `Catalog`s do not implement `AutoCloseable`, but some implementations do *sigh*)

api/client/src/main/java/org/projectnessie/client/http/impl/jdk8/UrlConnectionClient.java
snazy: Okay for me

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -0,0 +1,277 @@
...
+   * token lifespans and with/without sending the refresh token on the initial request.
+   *
+   * <p>For 15 seconds, the OAuth2 client will strive to keep the access token valid; in the
+   * meantime, another HTTP client, acting as a typical resource server, will attempt to use the
+   * access token to obtain a UMA (User Management Access) ticket from Keycloak, symbolizing the
+   * fact that Keycloak authorizes the client to access its resources.
+   */
+  @ParameterizedTest
+  @MethodSource
+  void testOAuth2Client(
```
adutra, line 95: FYI I'm not entirely satisfied with this test that is admittedly very long (9*15 seconds). Today I'm going to try to create other tests focusing on details of the refresh cycle and maybe reduce this one to just one run of 15 seconds. (I think it's important to at least have one test exercising the full refresh cycle.)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
snazy: I'd omit `capitalize()` completely.
For example, [RFC 6749 section 7.1](https://www.rfc-editor.org/rfc/rfc6749.html#section-7.1) mentions a `MAC` authentication scheme (work in process). Not saying we should think of that at this time.
But overall, it can't probably be anything else than `Bearer` - maybe we should add a runtime check like `tokenType.toLowerCase(ROOT).equals("bearer")`?

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
snazy: Yea - that's very long. Maybe one quick IT integration round plus a couple with a mocked HTTP backend?

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -255,50 +274,61 @@ public HttpClientBuilder withDisableCompression(boolean disableCompression) {
...
   @Override
   public <API extends NessieApi> API build(Class<API> apiVersion) {
     Objects.requireNonNull(apiVersion, "API version class must be non-null");
 
+    if (tracing) {
+      // Do this at the last possible moment because once added, tracing cannot be removed.
+      builder.addTracing();
+    }
+
+    HttpClient httpClient = builder.build();
```
snazy, line 323: O O! Do not move this here, the Json-View setting won't be applied.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
adutra: I like your suggestion, let's remove the capitalize stuff and replace with a check.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2ClientParams.java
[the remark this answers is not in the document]
snazy: Could be. Let's defer to a follow-up, if necessary.

api/client/build.gradle.kts
```diff
@@ -77,6 +77,12 @@ dependencies {
   testFixturesApi(libs.undertow.core)
   testFixturesApi(libs.undertow.servlet)
   testFixturesImplementation(libs.logback.classic)
+
+  intTestImplementation(libs.testcontainers.testcontainers)
+  intTestImplementation(libs.testcontainers.junit)
+  intTestImplementation(libs.testcontainers.keycloak) {
+    exclude(group = "org.slf4j") // uses SLF4J 2.x, we are not ready yet
```
adutra, line 84: @snazy let me know if there is a better way to handle this.

api/client/src/main/java/org/projectnessie/client/auth/oauth/OAuth2Client.java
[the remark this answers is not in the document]
adutra: There you go: https://github.com/projectnessie/nessie/issues/6586.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -0,0 +1,324 @@
...
+          Thread thread = Executors.defaultThreadFactory().newThread(runnable);
+          thread.setDaemon(true);
+          thread.setName("oauth2-client");
+          return thread;
+        });
+  }
+
+  @Override
+  public AccessToken authenticate() {
+    return getCurrentTokens().getAccessToken();
```
snazy, line 107: Would it be a lot of effort to have this covered in a unit test? Does it make sense?
(I see, it's covered in the IT, so not really an issue.)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -0,0 +1,324 @@
...
+   */
+  public void start() {
+    started.complete(null);
+  }
+
+  @Override
+  public void close() {
+    try {
+      currentTokensStage.toCompletableFuture().cancel(true);
+      ScheduledFuture<?> tokenRefreshFuture = this.tokenRefreshFuture;
```
snazy, line 145: Nit: Should `this.tokenRefreshFuture` be set to `null` in a `finally`?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -0,0 +1,324 @@
...
+      }
+      if (shouldCloseExecutor) {
+        if (!executor.isShutdown()) {
+          executor.shutdown();
+          if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
+            executor.shutdownNow();
+          }
+        }
+      }
+      if (password != null) {
```
snazy, line 157: Hm - if the same password's used in multiple clients/authentictors  (probably never happens though, but tests maybe...), a `close()` would corrupt the password for other instances.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/JwtToken.java
```diff
@@ -0,0 +1,216 @@
...
+   * claim is generally application specific. The "iss" value is a case-sensitive string containing
+   * a StringOrURI value. Use of this claim is OPTIONAL.
+   */
+  @Nullable
+  public String getIssuer() {
+    return (String)
+        claims.computeIfAbsent(
+            ISS_CLAIM,
+            k ->
+                payload.has(ISS_CLAIM) && payload.get(ISS_CLAIM).isTextual()
```
snazy, line 94: Would something like this make it overall simpler?

```java

  private <T> T payloadFieldValue(String field, Predicate<JsonNode> tester, Function<JsonNode, T> mapper) {
    JsonNode value = payload.get(field);
    if (value == null || !tester.test(value)) {
      return null;
    }
    return mapper.apply(value);
  }
```

and for this `getIssuer`:
```java
          k -> payloadFieldValue(ISS_CLAIM, JsonNode::isTextual, JsonNode::asText));
```

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public void applyToHttpClient(HttpClient.Builder client) {
+      client.addRequestFilter(this::addAuthHeader);
+    }
+
+    void addAuthHeader(RequestContext ctx) {
+      AccessToken token = authenticator.authenticate();
+      if (!token.getTokenType().toLowerCase(Locale.ROOT).equals("bearer")) {
+        throw new IllegalArgumentException(
+            "OAuth2 token type must be 'Bearer', but was: " + token.getTokenType());
```
snazy, line 67:
```suggestion
            "OAuth2 token type returned from the authenticating server must be 'Bearer', but was: " + token.getTokenType());
```

snazy (commented):
(Only finished a first, incomplete round, will continue tomorrow)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientParams.java
```diff
@@ -53,8 +58,17 @@ interface OAuth2ClientParams {
 
   byte[] getClientSecret();
 
+  Optional<String> getUsername();
+
+  Optional<byte[]> getPassword();
```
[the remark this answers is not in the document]
adutra: In fact I just changed that. My thinking is that this object is meant to be short-lived, so there is really no need to worry about passwords here.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: Good catch, but that doesn't happen anymore, because in `OAuth2ClientParams` we are using plain strings now. This class creates a new array when reading the password.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -0,0 +1,324 @@
...
+            delay.toMillis(),
+            TimeUnit.MILLISECONDS);
+  }
+
+  private CompletionStage<Tokens> renewTokens() {
+    return currentTokensStage
+        // try refreshing the current tokens
+        .thenApply(this::refreshTokens)
+        // if that fails, try fetching brand-new tokens
+        .exceptionally(error -> fetchNewTokens());
```
snazy, line 189: Does it make sense to log the error here or any (runtime) exception in `refreshTokens`?
Guess, it can only be either the HTTP request failing or the response not being deserializable in `refreshTokens` (and `exchangeTokens`).

snazy (commented):
Looks really great!

Just a few comments, nothing dramatic.

And great job on test coverage!

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticationProvider.java
```diff
@@ -0,0 +1,77 @@
...
+      }
+      ctx.putHeader("Authorization", "Bearer " + token.getPayload());
+    }
+
+    @Override
+    public void close() {
+      authenticator.close();
```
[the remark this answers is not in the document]
adutra: Done in ce91cbd37 and 30fbbe1e4, although I had to move `TestNessieHttpClient.java` to the `org.projectnessie.client.http` package for visibility reasons. But it makes sense to have that test in that package imho since the class under test is also in that package.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: That should be possible with some mocking. Good idea.

adutra: Done in 91b5ca7c6.

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -44,15 +47,24 @@
...
-  private HttpAuthentication authentication;
+  private static final ObjectMapper MAPPER =
+      new ObjectMapper()
+          .enable(SerializationFeature.INDENT_OUTPUT)
+          .disable(SerializationFeature.FAIL_ON_EMPTY_BEANS);
+
+  private final HttpClient.Builder builder =
+      HttpClient.builder()
+          .setObjectMapper(MAPPER)
+          .addResponseFilter(new NessieHttpResponseFilter());
```
adutra, line 66: I wonder: we should probably not add this filter by default. For the OAuth2 client, it doesn't make sense for instance (it's not harmful either, just conceptually wrong since the client is not talking to Nessie at all).

snazy: Yes, doesn't make sense for OAuth. Probably safer to not add it for OAuth and only for Nessie REST.

adutra: Actually scratch that, the OAuth client uses `HttpClient.Builder`, not `HttpClientBuilder`... naming is hard :-)

snazy (approved):
Great stuff!
Ship it (once CI's happy)!

PR #6546, opened 2023-04-09
CI: Skip Helm CI, if version is not available

snazy:
Fix a slight annoyance when creating a release the Helm-CI-job fails, because the already commited Helm chart version in `main` is not yet available on `charts.projectnessie.org`.


.github/workflows/ci.yml
```diff
@@ -660,6 +660,15 @@ jobs:
         uses: actions/checkout@v3.5.0
         with:
           fetch-depth: 0
+
+      # Skip remaining steps, if version is available
```
dimas-b, line 664: `... if version is NOT available`?

.github/workflows/ci.yml
```diff
@@ -660,6 +660,15 @@ jobs:
...
+      - name: Precheck
+        run: |
+          EXPECT_VERSION="$(cat helm/nessie/Chart.yaml | grep -E '^version: ')"
+          curl https://charts.projectnessie.org/index.yaml 2>/dev/null | grep -q "${EXPECT_VERSION}" || (
+            echo "WF_EXEC=false" >> ${GITHUB_ENV}
```
dimas-b, line 669: nit: maybe add a note to the build log as well?

dimas-b (approved):

PR #6548, opened 2023-04-10
Support `@NessieServerProperty` annotations in nested tests

dimas-b:
* `@NessieServerProperty` annotations are processed on all test classes starting from the root context.

* Properties in inner / nested contexts take precedence over properties in outer contexts.


snazy (commented):
The changes overall look good to me.
But why the change to `ITImplicitNamespaces`?

PR #6577, opened 2023-04-12
CI: Replace auto-labeler with changed-files checks

snazy:
The auto-labeler approach is nice, but has some disadvantages. Each label addition triggers a new workflow run. This is usually fine for manually added labels, but if multiple labels are added in a short period of time, there is a non-neglectible chance that GitHub then cancels the wrong workflow run (think: `concurrency` in `ci.yml`).

This change replaces the auto-labeler approach with an additional job in `ci.yml` that determines the jobs to run.

For pull requests, the corresponding label is checked first, then the list of changed files is checked against the same globs that were used for the auto-labeling approach.


.github/workflows/ci.yml
```diff
@@ -331,14 +331,136 @@ jobs:
...
+
+            ///////////////////////////////////////////////////////////////////////////////////////
+            // "Non PR" case:
+            // Behavior of the conditional jobs when the event that triggered the workflow is not
+            // for a pull request (usually a push-to-main).
+            //
+            if (!context.payload || !context.payload.pull_request) {
+              core.info("Not a pull-request, enabling all jobs.");
+
+              // Yield 'true' for all event payload types, except pull_request
```
dimas-b, line 410: but `nesqueit` is `false`?

snazy: yes, handled in a scheduled job - updated comments

dimas-b (approved):

PR #6581, opened 2023-04-13
Quarkus: Fix HTTP compression parameters

snazy:
... and also disable file-uploads, which we do not use, as a safety thingy.

### DIFF
--- a/servers/quarkus-server/src/main/resources/application.properties
+++ b/servers/quarkus-server/src/main/resources/application.properties
@@ -129,6 +129,9 @@ quarkus.http.access-log.enabled=true
 # fixed at buildtime
 quarkus.resteasy.path=/api
 quarkus.resteasy.gzip.enabled=true
+quarkus.http.enable-compression=true
+quarkus.http.enable-decompression=true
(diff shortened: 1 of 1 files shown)


servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -128,7 +128,9 @@ quarkus.http.test-port=0
 quarkus.http.access-log.enabled=true
 # fixed at buildtime
 quarkus.resteasy.path=/api
-quarkus.resteasy.gzip.enabled=true
+quarkus.http.enable-compression=true
```
adutra, line 131: According to [this page](https://quarkus.io/guides/resteasy#gzip-support):

> The configuration property `quarkus.http.enable-compression` has no effect on compression support of RESTEasy Classic endpoints.

So if I'm understanding this correctly, this change will effectively disable compression for all REST endpoints, but enable it for static resources. Is that the intent?

snazy: Strange. IntelliJ shows that property as "not mapped". I've re-added it.

PR #6584, opened 2023-04-13
JAX-RS: Properly "pass through" `WebApplicationException`

### LINKED ISSUE #6569: API v2 requests sent to v1 base URL lead to 500 errors but should be 404
```
$ curl http://localhost:19120/api/v1/trees/-

{
  "status" : 500,
  "reason" : "Internal Server Error",
  "message" : "javax.ws.rs.NotFoundException: RESTEASY003210: Could not find resource for full path: http://localhost:19120/api/v1/trees/-",
  "errorCode" : "UNKNOWN",
  "serverStackTrace" : null
}
```

Note that the `trees/-` part is an API v2 sub-path, while the base URL is `v1`.

The servers should return a 404 or 400 response in this case, not an "internal error"

snazy:
fixes #6569

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
@@ -77,7 +77,6 @@ public abstract class BaseTestNessieRest extends BaseTestNessieApi {
     RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
   }
 
-  @SuppressWarnings("JUnitMalformedDeclaration")
   @BeforeEach
   void setupRestUri(@NessieClientUri URI uri) {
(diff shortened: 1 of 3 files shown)


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
```diff
@@ -146,6 +145,20 @@ private Branch createBranchV1(String name) {
...
+  @ParameterizedTest
+  @ValueSource(
+      strings = {
+        "trees/not-there-using",
+        "contents/not-there-using",
+        "something/not-there-using",
+        ""
+      })
+  public void testNotFoundUrls(String path) {
+    System.err.println(RestAssured.baseURI);
```
dimas-b, line 157: `System.err`?

dimas-b (dismissed):

PR #6595, opened 2023-04-14
Docs for the new OAuth2 authentication provider

adutra:
...and a few javadoc improvements.

### DIFF
--- a/site/docs/tools/client_config.md
+++ b/site/docs/tools/client_config.md
@@ -107,35 +107,112 @@ environment-specific prefixes for brevity. Nonetheless, in practice the property
 given appropriate prefixes (as in the example above) for them to be recognized by the tools and Nessie
 code.
 
-The value of the `authentication.type` property can be one of the following:
+The value of the `nessie.authentication.type` property can be one of the following:
 
 * `NONE` (default)
 * `BEARER`
+* `OAUTH2`
 * `AWS`
 * `BASIC` (deprecated)
 
 ## Authentication Type `NONE`
 
-For the Authentication Type `NONE` only the `authentication.type` property needs to be set.
+For the Authentication Type `NONE` only the `nessie.authentication.type` property needs to be set.
 
 This is also the default authentication type if nothing else is configured.
 
 ## Authentication Type `BEARER`
 
-For the `BEARER` Authentication Type the `authentication.token` property should be set to a valid
-[OpenID token](https://openid.net/specs/openid-connect-core-1_0.html).
+For the `BEARER` Authentication Type the `nessie.authentication.token` property should be set to a
+valid [OpenID token](https://openid.net/specs/openid-connect-core-1_0.html).
+
+This authentication type is recommended only when the issued access token has a lifespan large
+enough to cover the duration of the entire Nessie client's session. Once the token is expired, the
+Nessie client will not be able to refresh it and will have to be restarted, with a different token.
+If the token needs to be refreshed periodically, then the `OAUTH2` authentication type should be
+preferred to this one.
+
+## Authentication Type `OAUTH2`
+
+The `OAUTH2` Authentication Type is able to authenticate against an OAuth2 server and obtain a valid
+access token. The access token is then used to authenticate against Nessie. The client will
+automatically refresh the access token. This authentication type is recommended when the access
+token has a lifespan shorter than the Nessie client's session lifespan.
+
+Note that the Nessie server must be configured to accept OAuth2 tokens from the same server. For
+example, if the OAuth2 server is Keycloak, this can be done by defining the following properties in
+the `application.properties` file of the Nessie server:
+
+```properties
+nessie.server.authentication.enabled=true
+quarkus.oidc.auth-server-url=https://<keycloak-server>/realms/<realm-name>
+```
+
+The following properties are available for the `OAUTH2` authentication type:
+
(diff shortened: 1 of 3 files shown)


site/docs/tools/client_config.md
```diff
@@ -107,35 +107,112 @@ environment-specific prefixes for brevity. Nonetheless, in practice the property
...
 * `NONE` (default)
 * `BEARER`
+* `OAUTH2`
 * `AWS`
 * `BASIC` (deprecated)
 
 ## Authentication Type `NONE`
 
-For the Authentication Type `NONE` only the `authentication.type` property needs to be set.
+For the Authentication Type `NONE` only the `nessie.authentication.type` property needs to be set.
```
dimas-b, line 120: Just wondering: why do we need the `nessie` prefix here?

site/docs/tools/client_config.md
```diff
@@ -107,35 +107,112 @@ environment-specific prefixes for brevity. Nonetheless, in practice the property
...
+This authentication type is recommended only when the issued access token has a lifespan large
+enough to cover the duration of the entire Nessie client's session. Once the token is expired, the
+Nessie client will not be able to refresh it and will have to be restarted, with a different token.
+If the token needs to be refreshed periodically, then the `OAUTH2` authentication type should be
+preferred to this one.
+
+## Authentication Type `OAUTH2`
+
+The `OAUTH2` Authentication Type is able to authenticate against an OAuth2 server and obtain a valid
+access token. The access token is then used to authenticate against Nessie. The client will
```
dimas-b, line 138: Would it make sense that only `Bearer` tokens are currently supported?

dimas-b (commented):
LGMT :+1: Minor comment below

site/docs/tools/client_config.md
adutra: Well I just looked at `NessieConfigConstants`: all constants start with `nessie.` so I thought the docs were wrong. Are we supposed to remove that prefix when running the client?

dimas-b: Yeah, it can be misleading, I guess :thinking: IIRC, this doc applies to Spark/Iceberg catalog configuration where the `nessie.` prefix is replaced with the actual catalog name (there could be multiple names).

Cf. https://github.com/apache/iceberg/blob/522480dc4dff7305808403281980f6afdedf9cf1/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java#L86

dimas-b: I tend to think that presenting an Iceberg-centric picture here is preferable because the doc page is about "tools" so it should hopefully be easier for end users to understand.

dimas-b: Example for Spark:
```
bin/spark-shell [...] --conf spark.sql.catalog.mycatalog.catalog-impl=org.apache.iceberg.nessie.NessieCatalog --conf spark.sql.catalog.mycatalog.uri=$NESSIE_URI  --conf spark.sql.catalog.mycatalog.authentication.type=BEARER --conf spark.sql.catalog.mycatalog.authentication.token=$NESSIE_TOKEN
```

adutra: I see, thanks for the explanation. It's really confusing for newcomers like me especially because at the top of the same page the options have the `nessie.` prefix:

https://github.com/projectnessie/nessie/blob/nessie-0.57.0/site/docs/tools/client_config.md?plain=1#L11

But OK, I will revert the prefix changes.

site/docs/tools/client_config.md
```diff
@@ -125,6 +126,82 @@ This is also the default authentication type if nothing else is configured.
...
+  and user must be properly configured with appropriate permissions in the OAuth2 server for the
+  authentication to succeed.
+
+* `authentication.oauth2.client-id`: the client ID to use when authenticating against the OAuth2
+  server. Required.
+
+* `authentication.oauth2.client-secret`: the client secret to use when authenticating against the
+  OAuth2 server. Required.
+
+* `authentication.oauth2.username`: the username to use when authenticating against the OAuth2
```
adutra, line 173: @dimas-b I also wanted to mention this: I created 2 new options for username/password in the context of OAuth2, rather than reusing the 2 existing (deprecated) settings: `authentication.username` and `authentication.password`.

Are we OK with this?

site/docs/tools/client_config.md
dimas-b: Good point. We have this in the `Authentication Settings` section:
```
The property names are shown without environment-specific prefixes for brevity. Nonetheless, in practice the property names should be given appropriate prefixes (as in the example above) for them to be recognized by the tools and Nessie code.
```
.. maybe we could have another clarifying statement around the top-level table too... WDYT?

site/docs/tools/client_config.md
dimas-b: Yes, these are fine, I think because they are specific to the (standard) OAuth2 protocol. The deprecated `authentication.password` comes from the time when Nessie itself was validating passwords.

dimas-b: It might be ok to remove `authentication.password` and `authentication.username` now, I guess :)

dimas-b (approved):

site/docs/tools/client_config.md
adutra: Hmm I think that's enough in fact. I should have read the document more carefully.

PR #6597, opened 2023-04-14
Enhance ITOAuth2Client

adutra:
This commit enhances the existing IT tests for OAuth2Client by starting a true Nessie server and validating the obtained tokens against it, instead of using a custom validating HTTP client.


api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -292,51 +332,29 @@ void testOAuth2ClientUnauthorizedBadPassword() {
...
-      // expire.
-      revokeAccessToken(validatingClient, tokens.getAccessToken());
-      soft.assertThatThrownBy(() -> tryUseAccessToken(validatingClient, tokens.getAccessToken()))
-          .isInstanceOf(HttpClientException.class)
-          .hasMessageContaining("401");
+      token.set(tokens.getAccessToken());
+      // Revoking the token is not enough because Nessie only
+      // validates the JWT token, it doesn't query Keycloak,
+      // so we need to wait until the token is expired :-(
+      Thread.sleep(11000);
```
adutra, line 345: This is slightly frustrating. Another option is to simply get rid of this test, it doesn't really test the OAuth2 client features.

snazy: Yea, then remove it, if it doesn't test anything that's under our control.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -44,27 +44,50 @@
...
+      new KeycloakContainer()
+          .withNetwork(network)
+          .withNetworkAliases("keycloak")
+          .withEnv("KC_HOSTNAME_URL", "http://keycloak:8080")
+          .withFeaturesEnabled("preview", "token-exchange");
+
+  @Container
+  @SuppressWarnings("resource")
+  private static final GenericContainer<?> NESSIE =
+      new GenericContainer<>("ghcr.io/projectnessie/nessie:latest")
```
dimas-b, line 76: Why not run the end-to-end test under the quarkus server module as another sub-class of `BaseClientAuthTest`?

adutra: Why not... but I'm not sure as we are testing here some internals of `OAuth2Client` and besides, `OAuth2Client` is package-private. It might not be so easy to move this to the server module. But I can look into adding another test there, maybe something less involved. Would that work for you?

dimas-b: I mean we could probably keep the majority of tests for `OAuth2Client` here and validate against Keycloak. Then, just have a basic end-to-end test via Quarkus under the quarkus module so that we would not have to depend on Nessie docker images.

snazy: Yea - better some higher-level tests in `:nessie-quarkus`. The `latest` tag refers to the latest Nessie release, not the current state of the code, so it's tricky. BTW: we have the apprunner-plugin to run a Nessie using the in-tree Nessie code. But pulling that into `:nessie-client` would implicitly pull `:nessie-quarkus` for ITs into `:nessie-client`, not strictly a cyclic dependency, but not a "nice" depeendency cycle.

TL;DR it's probably better to have high-level in `:nessie-quarkus`.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -194,9 +195,7 @@ private CompletionStage<Tokens> renewTokens() {
 
   private void log(Tokens tokens, Throwable error) {
     if (error != null) {
-      LOGGER.error("Failed to renew tokens", error);
-    } else {
-      LOGGER.debug("Successfully renewed tokens: {}", tokens);
```
adutra, line 199: Not a good idea to log tokens, they may contain sensitive information.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/JacksonSerializers.java
```diff
@@ -54,8 +54,8 @@ public SecondsToInstantDeserializer() {
 
     @Override
     public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
-      int seconds = p.getValueAsInt(0);
-      if (seconds > 0) {
+      if (p.currentToken().isNumeric()) {
```
adutra, line 57: Small improvement to parse any value, including zero and negative (zero happens sometimes).

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: Alright, I added 2 classes in nessie-quarkus:

* `ITOAuth2Authentication` uses a true Keycloak server and a newly-introduced test resource for Keycloak.
* `TestOAuth2Authentication` uses Wiremock and `OidcWiremockTestResource`, just like the Bearer tests.

Let me know what you think. If you prefer that I rollback the changes in `ITOAuthClient`, no problem. Now that we have tests in nessie-quarkus, it's not so much of a problem if we don't bring Nessie in in `ITOAuthClient`

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/KeycloakTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,218 @@
...
+import org.keycloak.representations.idm.RoleRepresentation;
+import org.keycloak.representations.idm.RolesRepresentation;
+import org.keycloak.representations.idm.UserRepresentation;
+
+/**
+ * This is a copy of, and a drop-in replacement for, the Keycloak test resource {@code
+ * io.quarkus.test.keycloak.server.KeycloakTestResourceLifecycleManager} from {@code
+ * io.quarkus:quarkus-test-keycloak-server}.
+ *
+ * <p>This class exists because the original one does not work with recent versions of Keycloak. It
```
snazy, line 41: Does it make sense to open a Quarkus PR for it?

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/KeycloakTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,218 @@
...
+    realm.getClients().add(createWebAppClient(KEYCLOAK_WEB_APP_CLIENT));
+
+    realm.getUsers().add(createUser("alice", getUserRoles()));
+    realm.getUsers().add(createUser("admin", getAdminRoles()));
+    realm.getUsers().add(createUser("jdoe", Arrays.asList("user", "confidential")));
+
+    return realm;
+  }
+
+  //  private static String getAdminAccessToken() {
```
snazy, line 144: left-over?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
snazy: Or maybe only remove the tokens but keep the message?

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
snazy: Yea, I'd prefer to remove Nessie from the IT in `:nessie-client`

snazy (approved):
LGTM - consider my comments as suggestions

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/KeycloakTestResourceLifecycleManager.java
adutra: I don't think so unfortunately, someone already asked: https://github.com/quarkusio/quarkus/issues/30072 and the answer was: use Dev Services instead.

adutra:
Interesting. The new IT tests are interfering somehow with the existing `ITOpenIdAuthentication`, which are now failing. I think it's better to also migrate `ITOpenIdAuthentication` to use a true Keycloak server. This way both will share the same configuration.

snazy (approved):

PR #6615, opened 2023-04-17
Disable IT-Auth on WIn/Mac CI

snazy:
Recent runs ([here](https://github.com/projectnessie/nessie/actions/runs/4717326974) and [here](https://github.com/projectnessie/nessie/actions/runs/4717302209)) failed.

### DIFF
--- a/api/client/build.gradle.kts
+++ b/api/client/build.gradle.kts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import org.apache.tools.ant.taskdefs.condition.Os
+
 plugins {
   `java-library`
   jacoco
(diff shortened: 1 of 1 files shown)


adutra (approved):
LGTM although the error is a `ContainerFetchException` so I wonder if this isn't going to happen also under Linux eventually :-(

snazy:
> LGTM although the error is a `ContainerFetchException` so I wonder if this isn't going to happen also under Linux eventually :-(

Nah - that's something that sadly happens only in these environments (both are not really supported by testcontainers / docker-java). It's not really an error that happens during a "fetch", but somewhere around it - and it doesn't recover.

Issue #5363, opened 2022-10-12
Nessie model classes from the SQL extensions jar interfere with Nessie classes in Iceberg jars (JSON parsing)


dimas-b:
Iceberg relocates `jackson` classes in its own artifacts. Nessie SQL extensions do not perform a compatible relocation.

As a result with the Nessie Iceberg Catalog and Nessie SQL extensions are used on the same class path, the resulting behaviour is not well defined and can lead to linkage errors (depending on jar processing order).

For example, with Spark 3.3:
```
$ bin/spark-sql \
                               --packages \
                              org.projectnessie:nessie-spark-extensions-3.3_2.12:0.43.0,org.apache.iceberg:iceberg-spark-runtime-3.3_2.12:0.14.0 \
                               --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions  \
                               --conf spark.sql.catalog.nessie=org.apache.iceberg.spark.SparkCatalog \
                               --conf spark.sql.catalog.nessie.warehouse=$PWD/data2 \
                               --conf spark.sql.catalog.nessie.catalog-impl=org.apache.iceberg.nessie.NessieCatalog \
                               --conf spark.sql.catalog.nessie.uri=http://localhost:19120/api/v1 \
                               --conf spark.sql.catalog.nessie.ref=main \
                               --conf spark.sql.catalog.nessie.cache-enabled=false \
                              --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions,org.projectnessie.spark.extensions.NessieSparkSessionExtensions
```

Executing `use nessie` followed by `CREATE TABLE demo7 (id bigint, data string);` leads to:
```
java.lang.NoSuchMethodError: 'org.projectnessie.model.GenericMetadata org.projectnessie.model.GenericMetadata.of(java.lang.String, org.apache.iceberg.shaded.com.fasterxml.jackson.databind.JsonNode)'
	at org.apache.iceberg.nessie.NessieTableOperations.doCommit(NessieTableOperations.java:174)
	at org.apache.iceberg.BaseMetastoreTableOperations.commit(BaseMetastoreTableOperations.java:133)
	at org.apache.iceberg.BaseMetastoreCatalog$BaseMetastoreCatalogTableBuilder.create(BaseMetastoreCatalog.java:172)
	at org.apache.iceberg.spark.SparkCatalog.createTable(SparkCatalog.java:188)
[...]

PR #6616, opened 2023-04-17
Nessie: Generic information for operations and content results

### LINKED ISSUE #6593: Figure out "the way" to pass relevant information "through" Nessie
Consumers of Nessie events are likely interested in "numbers" that represent for example the "size" of a change to an Iceberg table - so the number of added files, the amount of bytes written, etc.

Nessie might not need to actually store this information, because it is already available via the current snapshot in table-metadata.

The idea is to add information to `Operation.Put`, something like this:
```java
interface Put extends Operation {
    @JsonInclude(Include.NON_EMPTY)
    List<ContentMetadata> getMetadata();
}
```

```java
@Value.Immutable
public interface ContentMetadata {
  String getVariant();
  JsonNode getMetadata();
}
```

Related issue: #6592 which would somehow need to include the metadata as well.

Also, just in case, add `List<ContentMetata> getMetadata()` to `ContentResponse` and `GetMultipleContentsResponse.ContentWithKey`.

snazy:
Adds `List<ContentMetadata>` to `Operation.Put`, `ContentResponse` and its multiple-get conterpart.

"Metadata" can be a lot of different things. If and how a Nessie server handles a particular "metadata variant" (think: type) depends on the Nessie server (configuration) and of course the variant itself.

For example, one metadata variant might contain the Iceberg snapshot summary to be passed through via Nessie events, or passed through and also stored in Nessie.

Removes the already unused `GenericMetadata` and its unused usages.

Fixes #6593

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
+++ b/api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Dremio
+ * Copyright (C) 2023 Dremio
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,35 +15,19 @@
  */
 package org.projectnessie.model;
 
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import com.fasterxml.jackson.databind.annotation.JsonSerialize;
-import javax.annotation.Nullable;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver;
 import javax.validation.constraints.NotEmpty;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
-import org.immutables.value.Value;
+import org.projectnessie.model.metadata.ContentMetadataResolver;
 
-@Value.Immutable
-@JsonSerialize(as = ImmutableContentMetadata.class)
-@JsonDeserialize(as = ImmutableContentMetadata.class)
+@Schema(type = SchemaType.OBJECT, title = "ContentMetadata", discriminatorProperty = "variant")
+@JsonTypeIdResolver(ContentMetadataResolver.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CUSTOM, property = "variant", visible = true)
 public interface ContentMetadata {
 
   @NotEmpty
   @jakarta.validation.constraints.NotEmpty
-  @Value.Parameter(order = 1)
   String getVariant();
-
-  @Nullable
-  @jakarta.annotation.Nullable
-  @Schema(type = SchemaType.OBJECT)
-  @Value.Parameter(order = 2)
-  @JsonInclude(Include.NON_NULL)
-  JsonNode getMetadata();
-
-  static ContentMetadata of(String variant, JsonNode metadata) {
-    return ImmutableContentMetadata.of(variant, metadata);
-  }
 }
--- a/api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
+++ b/api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
@@ -15,27 +15,35 @@
  */
 package org.projectnessie.model;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import javax.annotation.Nullable;
 import javax.validation.constraints.NotEmpty;
 import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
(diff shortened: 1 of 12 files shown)


snazy:
@dimas-b @adutra : I could need your opinion on this one.

This PR is meant to have "something" in the REST v2 APi that allows us to send "additional information" or "metadata" along with a put-operation. A Nessie server would chose on its own what exactly it does does with each individual metadata object - it could only pass it "through" to Nessie events, or store it, or something else:tm:. Whether and how metadata could be returned as part of a "get-content(s)" can be defined later.

Primary purpose of this PR is to have it defined in REST API v2 before we finalize it.

api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
```diff
@@ -15,27 +15,35 @@
...
   @jakarta.validation.constraints.NotEmpty
+  @Value.Parameter(order = 1)
   String getVariant();
 
+  @Nullable
+  @jakarta.annotation.Nullable
   @Schema(type = SchemaType.OBJECT)
+  @Value.Parameter(order = 2)
+  @JsonInclude(Include.NON_NULL)
   JsonNode getMetadata();
```
dimas-b, line 44: I suppose we have to fix #5363 before we can use this.

api/model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -91,6 +94,15 @@ interface Put extends Operation {
...
     Content getExpectedContent();
 
+    /**
+     * Additional information about the operation and/or content object. If and how a Nessie server
+     * uses and handles the information depends on the server version and type of metadata (called
+     * variant).
+     */
+    @JsonInclude(Include.NON_EMPTY)
+    @JsonView(Views.V2.class)
+    List<ContentMetadata> getMetadata();
```
dimas-b, line 104: Should we perhaps have a new `FetchOption` constant to control the loading of metadata in commit log (for example)?

api/model/src/main/java/org/projectnessie/model/ContentResponse.java
```diff
@@ -44,6 +49,11 @@ static ImmutableContentResponse.Builder builder() {
   @Value.Parameter(order = 2)
   Reference getEffectiveReference();
 
+  /** Additional content related information, if any. */
+  @JsonInclude(Include.NON_EMPTY)
+  @JsonView(Views.V2.class)
+  List<ContentMetadata> getMetadata();
```
dimas-b, line 55: Could it be useful to have a list of metadata types in the corresponding _request_ object?

Is it meaningful for a client to fetch (some) metadata _without_ a `Content` object?

api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
```diff
@@ -15,27 +15,35 @@
...
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.immutables.value.Value;
 
 @Value.Immutable
-@JsonSerialize(as = ImmutableGenericMetadata.class)
-@JsonDeserialize(as = ImmutableGenericMetadata.class)
-public interface GenericMetadata {
+@JsonSerialize(as = ImmutableContentMetadata.class)
+@JsonDeserialize(as = ImmutableContentMetadata.class)
+public interface ContentMetadata {
```
dimas-b, line 32: Why do we have custom server-side serializers for `Content`, but no for `ContentMetadata`?

dimas-b: I mean if plain Json is ok for metadata, why not for `Content`?

api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
snazy: Ah, damn!

snazy: But there are ways around it ;)

api/model/src/main/java/org/projectnessie/model/ContentMetadata.java
snazy: I'm actually thinking about whether it makes sense for `Content`. At least clients could then receive unknown content-types without crashing during deserialization.

api/model/src/main/java/org/projectnessie/model/ContentResponse.java
snazy: Not sure whether the `getMetadata()` in `ContentResponse` will be of any use short-term, but might be mid-term. Just would like to be prepared for the unlikely case.

dimas-b: Well, should we support fetching content without the metadata (overhead) then?

api/model/src/main/java/org/projectnessie/model/Operation.java
dimas-b: Yes, but so that we can have a stable v2 API, shouldn't we also prepare new `FetchOption` constants (or other means) to allow limiting the about of (unneeded) data in API responses?

api/model/src/main/java/org/projectnessie/model/metadata/GenericContentMetadataSerialization.java
```diff
@@ -0,0 +1,72 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.jsontype.TypeSerializer;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Map.Entry;
+
+public class GenericContentMetadataSerialization {
+
+  static final class GenericContentMetadataSerializer
+      extends JsonSerializer<ContentMetadataUnknown> {
```
dimas-b, line 32: Maybe rename `ContentMetadataUnknown` -> `GenericContentMetadata`

dimas-b (commented):
LGTM overall :+1:

snazy:
Removed the attributes in the get-content-responses

dimas-b (approved):

PR #6618, opened 2023-04-17
Ensure custom content-types work

### LINKED ISSUE #6591: Verify that adding new content types to Nessie works "everywhere"
Nessie internal code is built to support an extensible set of different content types. The (REST) API however is mostly not able to leverage "additional" content types or fails when those would be added (think: Java enum, current OpenAPI spec).

This issue is meant to ensure and empirically prove:
* adding new content types works on all levels (storage level serialization, model-API level serialization, REST-API level server & clients).
* add tests that validate this behavior

snazy:
* Add some more valdiations when registering content-type-bundles.
* Add a page on the project web site for registered content-types.
* Add tests in `:nessie-model` + `:nessie-versioned-spi` to verify that
  custom content-type-bundles and custom content-serializer-bundles
  work.

Fixes #6591

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Content.java
+++ b/api/model/src/main/java/org/projectnessie/model/Content.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import javax.annotation.Nullable;
@@ -34,7 +35,13 @@ import org.projectnessie.model.types.ContentTypes;
 @Schema(
     type = SchemaType.OBJECT,
     title = "Content",
-    oneOf = {IcebergTable.class, DeltaLakeTable.class, IcebergView.class, Namespace.class},
+    anyOf = {
+      IcebergTable.class,
+      DeltaLakeTable.class,
+      IcebergView.class,
+      Namespace.class,
+      Map.class
+    },
     discriminatorMapping = {
       @DiscriminatorMapping(value = "ICEBERG_TABLE", schema = IcebergTable.class),
       @DiscriminatorMapping(value = "DELTA_LAKE_TABLE", schema = DeltaLakeTable.class),
--- a/api/model/src/main/java/org/projectnessie/model/types/ContentTypeRegistry.java
+++ b/api/model/src/main/java/org/projectnessie/model/types/ContentTypeRegistry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022 Dremio
+ * Copyright (C) 2023 Dremio
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,10 +15,9 @@
  */
 package org.projectnessie.model.types;
 
-/**
- * Used to provide custom {@link org.projectnessie.model.Content} implementations via the Java
- * {@link java.util.ServiceLoader service loader} mechanism.
- */
-public interface ContentTypeBundle {
-  void register(ContentTypeRegistry contentTypeRegistry);
+import org.projectnessie.model.Content;
+
+/** An implementation of this interface is passed to {@link ContentTypeBundle}s. */
+public interface ContentTypeRegistry {
+  void register(Class<? extends Content> type);
 }
--- a/api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
+++ b/api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
@@ -15,12 +15,10 @@
  */
 package org.projectnessie.model.types;
 
-import org.projectnessie.model.types.ContentTypes.Registrar;
-
 public class CustomTestContentTypeBundle implements ContentTypeBundle {
 
   @Override
-  public void register(Registrar registrar) {
-    registrar.register("TEST_CUSTOM_CONTENT_TYPE", CustomTestContent.class);
+  public void register(ContentTypeRegistry contentTypeRegistry) {
+    contentTypeRegistry.register(CustomTestContent.class);
   }
 }
--- a/api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
+++ b/api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model.types;
+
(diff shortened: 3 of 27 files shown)


api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+package org.projectnessie.model.types;
+
+import org.projectnessie.model.types.ContentTypes.Registrar;
+
+public class CustomTestContentTypeBundle implements ContentTypeBundle {
+
+  @Override
+  public void register(Registrar registrar) {
+    registrar.register("TEST_CUSTOM_CONTENT_TYPE", CustomTestContent.class);
```
dimas-b, line 24: nit: use name constant from `CustomTestContent`?

api/model/src/test/java/org/projectnessie/model/types/CustomTestContentTypeBundle.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model.types;
+
+import org.projectnessie.model.types.ContentTypes.Registrar;
+
+public class CustomTestContentTypeBundle implements ContentTypeBundle {
```
dimas-b, line 20: Suggestion: Do we really need the `ContentTypeBundle` interface? Why not load each custom type class via the service loader and then register them all automatically based on `@JsonTypeName` (which we validate to match the registered type name anyway)?

IIRC, service loader resources can list more than one service class per file.

With current approach it looks like type names leak across class boundaries without any good code-level reason :thinking: 

If that works, we might as well follow the same approach for `ContentSerializerBundle`, I guess.

snazy: In theory, yes.
But we need `Class<? extends Content>` for the registration. Service loaders however return instantiated instances of the class.

snazy:
Updated the PR a bit.

* Introduced "top level" registry interfaces for content-type + content-serializer registration.
* Removed the `name` parameter

versioned/spi/src/main/java/org/projectnessie/versioned/store/ContentSerializerBundle.java
```diff
@@ -15,9 +15,12 @@
...
-/** Implementations of this {@link java.util.ServiceLoader service} register content serializers. */
+/**
+ * Custom content serializers need to implement this interface and call {@link
+ * ContentSerializerRegistry#register(ContentSerializer)} for each serializer.
+ *
+ * <p>Instances of this class are loaded via Java's {@link java.util.ServiceLoader}.
+ */
 public interface ContentSerializerBundle {
-  void register(Consumer<ContentSerializer<?>> registry);
+  void register(ContentSerializerRegistry registry);
```
dimas-b, line 25: Well, unlike `Content.Type`, here we register serializer instances... Could you simply do `ServiceLoader.load(ContentSerializer.class)`? On the receiving side, we do not know specific type parameters anyway.

snazy: But why have another approach here?

dimas-b: Do remove an extra interface... but this is a minor point, optional.

api/model/src/main/java/org/projectnessie/model/types/ContentTypeRegistry.java
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model.types;
+
+import org.projectnessie.model.Content;
+
+/** An implementation of this interface is passed to {@link ContentTypeBundle}s. */
+public interface ContentTypeRegistry {
+  void register(Class<? extends Content> type);
```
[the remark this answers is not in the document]
dimas-b: The Json part is fine, I think. My concern is only about the Open API spec (which has its own discriminators in the schema annotation on `Content`).

Since we now expect custom types to be present, is it worth declaring a sub-set of possible types as `oneOf` in our Open API spec?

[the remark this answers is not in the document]
snazy: Define "possible types" ;)

The problem is that we cannot dynamically extend that `oneOf` + the OpenAPI discriminators. OTOH, it's only "an issue" for new/completely-custom-private types. New "core" types will naturally be added there.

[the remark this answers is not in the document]
dimas-b: Why would we want to keep `oneOf` in the Open API spec at all? With pluggable types, the spec is guaranteed to be skewed sooner or later. 

I'm proposing to replace it with a generic `Object` type (or the Open API) equivalent.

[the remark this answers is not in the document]
adutra: I think this could become a problem for the future Event API. The generated JSON schema for this API currently has this:

```json
      "Content" : {
        "title" : "Content",
        "type" : "object",
        "properties" : {
          "id" : {
            "type" : "string"
          }
        },
        "oneOf" : [ {
          "$ref" : "#/components/schemas/IcebergTable"
        }, {
          "$ref" : "#/components/schemas/DeltaLakeTable"
        }, {
          "$ref" : "#/components/schemas/IcebergView"
        }, {
          "$ref" : "#/components/schemas/Namespace"
        } ],
        "discriminator" : {
          "propertyName" : "type",
          "mapping" : {
            "ICEBERG_TABLE" : "#/components/schemas/IcebergTable",
            "DELTA_LAKE_TABLE" : "#/components/schemas/DeltaLakeTable",
            "ICEBERG_VIEW" : "#/components/schemas/IcebergView",
            "NAMESPACE" : "#/components/schemas/Namespace"
          }
        }
      },
```

The `oneOf` approach will be problematic if someone uses this schema to attempt a polymorphic deserialization of components of different types, for example when using `KafkaJsonSchemaDeserializer` to deserialize events from a Kafka topic.

Even if the consumer has the jar containing the custom content implementation, the schema won't allow it, and deserializing such a content is likely to fail imho.

(However, I _think_ this could work when using a schema-less deserializer like `KafkaJsonDeserializer`.)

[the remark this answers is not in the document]
snazy: Hm - custom content-types aren't new though. Not sure why events infra needs to know the whole model.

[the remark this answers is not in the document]
snazy: > perhaps replace with anyOf(coreType1, coreType2, ..., Object).

`anyOf` seems to work

api/model/src/main/java/org/projectnessie/model/Content.java
```diff
@@ -34,7 +35,13 @@
 @Schema(
     type = SchemaType.OBJECT,
     title = "Content",
-    oneOf = {IcebergTable.class, DeltaLakeTable.class, IcebergView.class, Namespace.class},
+    anyOf = {
+      IcebergTable.class,
+      DeltaLakeTable.class,
+      IcebergView.class,
+      Namespace.class,
+      Map.class
```
dimas-b, line 43: nit: this does not have a name in Swagger / Open API yaml:
```
    Content_V2:
      title: Content
      type: object
      properties:
        id:
          type: string
      anyOf:
      - $ref: '#/components/schemas/IcebergTable_V2'
      - $ref: '#/components/schemas/DeltaLakeTable_V2'
      - $ref: '#/components/schemas/IcebergView_V2'
      - $ref: '#/components/schemas/Namespace_V2'
      - type: object
```
not a big deal, but perhaps we can name it "Custom Content" (latger)

dimas-b (approved):

api/model/src/main/java/org/projectnessie/model/Content.java
snazy: Can't put a "name" for it :(

PR #6621, opened 2023-04-17
Adjust testNotFoundUrls parameters

dimas-b:
This is a follow-up to #6584

Use only truly unmapped paths for this test.

`contents/not-there-using` URLs are mapped in API v1 as requests for the key `not-there-using`.

`trees/not-there-using` URLs are mapped in API v2 as requests for the reference named `not-there-using`.

Even though the above two case do produce 404 response, they do not validate the code changed in `NessieExceptionMapper` for #6569.

Also the contents request lacks some required parameters (ref name).

### DIFF
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
@@ -145,13 +145,7 @@ public abstract class BaseTestNessieRest extends BaseTestNessieApi {
   }
 
   @ParameterizedTest
-  @ValueSource(
-      strings = {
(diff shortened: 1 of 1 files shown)


snazy:
Yea - I added those intentionally. The first isn't mapped in v2 and the 2nd not mapped in v1.

snazy (approved):

dimas-b:
To clarify:  with the partially specified requests the behaviour is not well-defined.... e.g. for `contents/not-there-using` is it not strictly defined with the response should be a 404 or a validation error due to missing reference name / hash parameters.

dimas-b:
Actually, the `refName` parameter for API v1 `GET contents/{key}` is not required, so the request is correct.

dimas-b:
Make this use case explicit in #6623

PR #6630, opened 2023-04-19
Null out references to java.net.http.HttpClient

adutra:
HttpClient maintains a thread pool and a single
selector thread. These resources cannot be freed
if a strong reference to the HttpClient instance
exists.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaRequest.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaRequest.java
@@ -55,7 +55,7 @@ final class JavaRequest extends BaseHttpRequest {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(JavaRequest.class);
 
-  private final HttpClient client;
+  private HttpClient client;
 
   JavaRequest(JavaHttpClient client) {
     super(client.config);
@@ -131,6 +131,7 @@ final class JavaRequest extends BaseHttpRequest {
       response = null;
       return config.responseFactory().make(responseContext, config.getMapper());
     } finally {
+      client = null;
       if (response != null) {
         try {
           LOGGER.debug(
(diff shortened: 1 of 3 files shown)


api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaRequest.java
```diff
@@ -55,7 +55,7 @@ final class JavaRequest extends BaseHttpRequest {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(JavaRequest.class);
 
-  private final HttpClient client;
+  private HttpClient client;
```
snazy, line 58: Do we care for a single request? Mean, a request is a short lived object.

dimas-b: In terms of GC - probably not. However, it is has a thread pool, would this cause overhead when we create a new client for each request?

snazy: Could, yes.
Maybe not for the current impls. But other implementations (like Apache HttpClient) do have their own pool.

adutra: It's meant to be short lived, but you don't know how users are going to sue it: a user could very well store an instance of `JavaRequest` in a field of some object and forget to null it out. In this case, the reference to the client would leak until the container object is garbage-collected.

I can remove this part of the change, but I think we are safer if we proactively null it out everywhere we can.

adutra: > In terms of GC - probably not. However, if it has a thread pool, will this cause overhead when we create a new client for each request?

We don't create a new client for every request, it's the same client that is passed to each instance of `JavaRequest`.

snazy (approved):

snazy (approved):

PR #6634, opened 2023-04-19
Add "fallback cases" for relevant enums in `:nessie-model`

snazy:
Adds custom deserializers for all enums in `:nessie-model` to make client & servers "safe" against new enum values. All relevant enums have a `static ENUMTYPE parse(String)` method, which falls back to a hard coded "safe default" value.

This apporach is meant to be prepared for emergency cases, when it is strictly necessary to add a new enum value. Or potentially also to let older clients continue to work "longer" against newer API/spec versions.

This approach is _NOT_ meant as a workaround to add arbitrary enum values, it should still be avoided wherever and whenever possible to "just add enum value".


snazy:
@dimas-b WDYT about this one? It is not a great one, but it might "safe our a.." at some point.

api/model/src/main/java/org/projectnessie/model/MergeBehavior.java
```diff
@@ -15,11 +15,36 @@
...
+  DROP;
+
+  public static MergeBehavior parse(String mergeBehavior) {
+    try {
+      if (mergeBehavior != null) {
+        return MergeBehavior.valueOf(mergeBehavior.toUpperCase(Locale.ROOT));
+      }
+      return null;
+    } catch (IllegalArgumentException e) {
+      return NORMAL;
```
dimas-b, line 39: I'd add an `UNDEFINED` case for this. It is not likely that client gets a merge behaviour value in a response that it did not request, but in case it happens, I guess we should not try to make it look like it was `NORMAL`.

dimas-b: `UNDEFINED` could also make sense as a value for `BaseMergeTransplant.getDefaultKeyMergeMode()` to indicate that the client expects all affected keys to be covered by `getKeyMergeModes()` and the server should error out if it has to fallback to `getDefaultKeyMergeMode()` for some reason.

api/model/src/main/java/org/projectnessie/model/Reference.java
```diff
@@ -109,6 +117,26 @@ static String toPathString(String name, String hash) {
...
+    TAG;
+
+    public static ReferenceType parse(String referenceType) {
+      try {
+        if (referenceType != null) {
+          return ReferenceType.valueOf(referenceType.toUpperCase(Locale.ROOT));
+        }
+        return null;
+      } catch (IllegalArgumentException e) {
+        return BRANCH;
```
dimas-b, line 129: I'd default to `TAG` because it is the simplest form of a named reference. A `BRANCH`, for example, can be used in commits/merges. Some new ref type (if it ever appears) will probably have some new use cases too.

dimas-b (commented):
I think this makes sense. Let's implement :)

api/model/src/main/java/org/projectnessie/model/MergeBehavior.java
snazy: Hm - in that case it's maybe better to not add a "fallback value" to `MergeBehavior`, because it's first sent to the server. If the server doesn't know about it, it should definitely error out.

dimas-b: If we're sure the server will not use new values in merge responses - then yes.

snazy: Yes, it replies with what the client sent.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
```diff
@@ -399,8 +399,7 @@ public void testInvalidTag(String invalidTagName) {
                         .extract()
                         .as(NessieError.class)
                         .getMessage())
-                .contains(
-                    "Could not resolve type id 'FOOBAR' as a subtype of `org.projectnessie.model.Reference`"));
+                .contains("assignReference.expectedHash: must not be null"));
```
dimas-b, line 402: Shouldn't this still be something like "Reference type FOOBAR is not supported"? That is to say that we should probably report the same "unsupported" error on the server side regardless of the API method (`assignReference.expectedHash: must not be null` is specific to ref. assignment, but the real problem is that the ref. type is bogus).

api/model/src/main/java/org/projectnessie/model/Reference.java
```diff
@@ -95,6 +105,7 @@ default void checkHash() {
 
   @JsonIgnore
   @Value.Redacted
+  @JsonDeserialize(using = Deserializer.class)
```
adutra, line 108:
```suggestion
  @JsonDeserialize(using = ReferenceType.Deserializer.class)
```

api/model/src/main/java/org/projectnessie/model/types/ReferenceTypeIdResolver.java
```diff
@@ -0,0 +1,90 @@
...
+import com.fasterxml.jackson.databind.jsontype.impl.TypeIdResolverBase;
+import java.util.Locale;
+import org.projectnessie.model.Branch;
+import org.projectnessie.model.Content;
+import org.projectnessie.model.Detached;
+import org.projectnessie.model.Reference;
+import org.projectnessie.model.Reference.ReferenceType;
+import org.projectnessie.model.Tag;
+
+/** Dynamic {@link Content} object (de)serialization for <em>Jackson</em>. */
```
adutra, line 30: Nit: javadoc is not accurate.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
snazy: It's ambiguous with this change to default to `TAG`

api/model/src/main/java/org/projectnessie/model/Reference.java
```diff
@@ -109,6 +120,27 @@ static String toPathString(String name, String hash) {
...
+
+    public static ReferenceType parse(String referenceType) {
+      try {
+        if (referenceType != null) {
+          referenceType = referenceType.toUpperCase(Locale.ROOT);
+          return "DETACHED".equals(referenceType) ? null : ReferenceType.valueOf(referenceType);
+        }
+        return null;
+      } catch (IllegalArgumentException e) {
+        return TAG;
```
adutra, line 133: I'm trying to apply your ideas to #6646... Is the intent here to return a "random" enum constant when the text does not match any known enum constant?

snazy: Not a random one, but the "likely best match" / least-common-denominator

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
dimas-b: Ah, right... I guess that default is meaningful only on the client side. If the server received a reference with a `type` that it does not know about, that is probably a "bad request" regardless of how JSON serialization is done... WDYT?

snazy: That's difficult. Opted to remove the change for `ReferenceType`.

dimas-b (approved):

PR #6635, opened 2023-04-19
Expose Nessie repository information (preparation)

### LINKED ISSUE #4993: REST API to fetch the Date of Repository Initialization or Date of First Commit
**Description**: Requesting the ability to know either the date of the first commit or when the repository was initialized.

- Changes in REST API

**Expected Use Cases**: In a UI when showing a date picker for selecting a commit before a certain time, this would help to disable irrelevant dates. E.g. if the repo was initialized on Aug 24, 2022, we can grey out/disable dates before that time so that the user cannot select them.

### LINKED ISSUE #5810: Common way to refer NO_ANCESTOR hash from clients
**Description**:
Nessie uses a hard-coded hash `2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d` to refer to the root, empty commit with no ancestor. In Java code, this is referred to as NO_ANCESTOR in a few places.

It can be hard to remember the actual hash (although it happens to be the SHA256 hash of the string `empty`). It would be nice to have a programmatic way to refer to it without needing to know the hash itself.

- Changes in REST API
- Changes in Java Client API
- Changes in SQL extentions

Primary request is for the Java API as shown below. Optionally it would also be nice to have similar functionality for Python, SQL, and possibly REST.

**Expected Use Cases**:
For example, to reset the default branch to the NO_ANCESTOR commit

Today, from Java client:
```
api.assignBranch()
  .branch(defaultBranch)
  .assignTo(Detached.of("2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d"));
```

One possibility, courtesy of @dimas-b :
```
api.assignBranch()
  .branch(defaultBranch)
  .assignTo(Detached.root());
```

snazy:
Enhances the REST endpoint `v2/config` with repository specific information:
* repository creation timestamp (new storage model only)
* oldest possible commit timestamp (new storage model only)
* no-ancestor-hash value
* default branch name (currently hard-wired to the configuration value)

Prepares the API changes for #4993 and #5810.

Since this changes the existing model for v2-beta, the added fields will be populated in a follow-up-PR.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/NessieConfiguration.java
+++ b/api/model/src/main/java/org/projectnessie/model/NessieConfiguration.java
@@ -15,15 +15,27 @@
  */
 package org.projectnessie.model;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.time.Instant;
+import java.util.Map;
 import javax.annotation.Nullable;
 import javax.validation.constraints.Size;
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
 import org.immutables.value.Value;
+import org.projectnessie.model.CommitMeta.InstantDeserializer;
+import org.projectnessie.model.CommitMeta.InstantSerializer;
 import org.projectnessie.model.ser.Views;
 
 /** configuration object to tell a client how a server is configured. */
+@Schema(
+    type = SchemaType.OBJECT,
+    title = "NessieConfiguration",
+    description = "Configuration object to tell a client how a server is configured.")
 @Value.Immutable
 @JsonSerialize(as = ImmutableNessieConfiguration.class)
 @JsonDeserialize(as = ImmutableNessieConfiguration.class)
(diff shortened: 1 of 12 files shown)


api/model/src/main/java/org/projectnessie/model/RepositoryInformation.java
```diff
@@ -0,0 +1,71 @@
...
+        "Exposes some information about the Nessie repository.\n\n"
+            + "All attributes may or may not be present, depending on the server version and repository.")
+@Value.Immutable
+@JsonSerialize(as = ImmutableRepositoryInformation.class)
+@JsonDeserialize(as = ImmutableRepositoryInformation.class)
+@JsonView(Views.V2.class)
+public interface RepositoryInformation {
+  @Nullable
+  @jakarta.annotation.Nullable
+  String getDefaultBranch();
```
dimas-b, line 51: If we intend repo info to be the source of truth for the default branch name, should we deprecate it in `NessieConfiguration` perhaps?

api/model/src/main/java/org/projectnessie/api/v2/http/HttpConfigApi.java
```diff
@@ -42,7 +43,7 @@ public interface HttpConfigApi extends ConfigApi {
   @Produces(MediaType.APPLICATION_JSON)
   @jakarta.ws.rs.Produces(jakarta.ws.rs.core.MediaType.APPLICATION_JSON)
   @Operation(
-      summary = "Returns repository and server settings relevant to clients.",
+      summary = "Returns repository and server settings relevant to clients.\rSince Nessie 0.59.0.",
```
dimas-b, line 46: the `getConfig` method is unchanged, right? Why add `Since ...`?

snazy:
I've changed this PR to not expose a new REST endpoint but extend the existing `NessieConfiguration`.

This PR adds new attributes to `NessieConfiguration`, but leaves those empty. A follow-up PR will then populate the new attributes in a Nessie release later than the one that includes this commit.

api/model/src/main/java/org/projectnessie/model/NessieConfiguration.java
```diff
@@ -57,11 +69,57 @@ public int getMinSupportedApiVersion() {
...
+   * commit-ID. A commit with this value does never exist.
+   */
+  @JsonView(Views.V2.class)
+  @JsonInclude(Include.NON_NULL)
+  @Nullable
+  @jakarta.annotation.Nullable
+  public abstract String getNoAncestorHash();
+
+  /**
+   * Timestamp when the repository has been created.
```
dimas-b, line 90: nit: `was created`?

servers/services/src/main/java/org/projectnessie/services/impl/ConfigApiImpl.java
```diff
@@ -19,12 +19,15 @@
...
+import org.projectnessie.versioned.VersionStore;
 
 public class ConfigApiImpl implements ConfigService {
 
+  private final VersionStore store;
   private final ServerConfig config;
 
-  public ConfigApiImpl(ServerConfig config) {
+  public ConfigApiImpl(ServerConfig config, VersionStore store) {
+    this.store = store;
```
dimas-b, line 30: nit: This can pprobably be in #6640

dimas-b (approved):

Issue #5342, opened 2022-10-07
Multiple repository on same Nessie instance


tmnd1991:
**Description**: as of now a Nessie instance/server can handle 1 repository. Therefore it forces to handle the datalake as a monorepo. I see use cases (data mesh) where such thing is not desired and would like to be able to handle groups of tables as different repositories, so that their “log” do not intertwine. Right now I’m forced to spin multiple Nessie servers, but I think this could be supported natively by Nessie.

**Requested Changes in public API**:
- Changes in REST API✅
- Changes in Java Client API✅
- Changes in Python Client API ✅
- Changes in SQL extentions ✅

**Expected Use Cases**: data mesh, data domains

keithgchapman:
@tmnd1991 Thanks for raising this issue. At this point we don't have any plans to built multi-tenancy in Nessie. We would welcome your ideas around this though and would be more than happy to accept code contributions in this regard.

Having said that [Dremio's Arctic service](https://docs.dremio.com/cloud/getting-started/arctic-architecture/) has the ability to support multiple catalogs within a single organization. It is powered by a managed version of Nessie.

tmnd1991:
Interesting, so the feature is there in the managed version, am I right? That's good to hear.

I will try to think about a model for multi-tenancy and maybe provide a PR in that regard :)

dimas-b:
@tmnd1991 : It may manifest differently at the API level. I'm interested in how you envision that.

tmnd1991:
I finally might have time to work on this.
I would like to start from the Curren openapi spec of nessie. Should I build the project in order to obtain it or is there an easier way? (like a link?)

tmnd1991:
Found the link in the release page, sorry for the spam.

PR #6640, opened 2023-04-20
Enables the extended information in `NessieConfiguration`

### LINKED ISSUE #4993: REST API to fetch the Date of Repository Initialization or Date of First Commit

### LINKED ISSUE #5810: Common way to refer NO_ANCESTOR hash from clients

snazy:
Builds on the API changes from #6635

Fixes #4993
Fixes #5810

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/impl/ConfigApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/ConfigApiImpl.java
@@ -19,6 +19,7 @@ import org.projectnessie.model.ImmutableNessieConfiguration;
 import org.projectnessie.model.NessieConfiguration;
 import org.projectnessie.services.config.ServerConfig;
 import org.projectnessie.services.spi.ConfigService;
+import org.projectnessie.versioned.RepositoryInformation;
 import org.projectnessie.versioned.VersionStore;
 
 public class ConfigApiImpl implements ConfigService {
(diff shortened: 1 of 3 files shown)


dimas-b (commented):
+1

servers/services/src/main/java/org/projectnessie/services/impl/ConfigApiImpl.java
```diff
@@ -33,9 +34,18 @@ public ConfigApiImpl(ServerConfig config, VersionStore store) {
 
   @Override
   public NessieConfiguration getConfig() {
```
adutra, line 36: Is `RepositoryInformation` supposed to change throughout the lifecycle of a Nessie instance? It seems that it is not, except maybe for the method `getAdditionalProperties()`.

If this information doesn't change, then `NessieConfiguration` could be cached in application scope. I would therefore like to suggest something: `NessieConfiguration` could be exposed as a singleton CDI bean, then injected here – rather than having one instance created here, per request.

The reason why I'm suggesting this is that this bean is likely to be useful in the events workflow as well.

dimas-b: `RepositoryInformation` is a function of the repo ID, which is currently fixed, but may not always be fixed (cf. #5342).

dimas-b (approved):

Issue #6626, opened 2023-04-18
Merging unrelated branches gives Http Status 404


mthirani2021:
While trying to merge two unrelated branches, it throws Http 404.

**What happened**:
Spinned up a Nessie server with default branch "main". Created a child branch out of it "dev" when there was no commit in main branch.
While trying to merge the "dev" in "main", got the expected error "No common ancestor" but Http status code was 404.
<img width="1039" alt="image" src="https://user-images.githubusercontent.com/96448504/232813680-a54d19e9-8f66-4b97-a8c7-b0897928f8cb.png">

**What you expected to happen**:
To throw the Http Status code as 409.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create a branch "branch1" from the parent branch "main" when there is no commit in parent branch.
2. Try to merge "branch1" to "main"

**Details**:
- Nessie server type (docker/built from source) and version: 0.57.0
- Client type (Ex: UI/Spark/pynessie ...) and version:

snazy:
That's a duplicate of #6447?

dimas-b:
Not a duplication IMO... #6447 is about having the "no common ancestor" error in principle. I think the plan is to resolve that with the `allow-unrelated-histories` (new) merge option.

This issue is about the HTTP status code and the ability to detect "no common ancestor" errors in a "dry run" merge.

PR #6646, opened 2023-04-21
Events notification system for Nessie - API module

adutra:
This PR is the first of a series of PRs related to the events notification system for Nessie.

See design doc: https://github.com/projectnessie/nessie/blob/main/design/events.md.

In this PR:

* New module `events/api` containing the event types defined in [section 2.1](https://github.com/projectnessie/nessie/blob/main/design/events.md#21-events) of the design doc.
* Type hierarchy annotated with Jackson and OpenAPI annotations, to facilitate serialization to Json. (Note that Nessie itself does not need to serialize these objects, but users will likely have to do that.)


events/api/build.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+
+extensions.configure<SmallryeOpenApiExtension> {
+  scanDependenciesDisable.set(false)
+  infoVersion.set(project.version.toString())
+  schemaFilename.set("META-INF/openapi/openapi")
+  operationIdStrategy.set(OperationIdStrategy.METHOD)
+  scanPackages.set(listOf("org.projectnessie.events.spi"))
+  scanClasses.set(
+    listOf(
+      "org.projectnessie.model.Content",
```
adutra, line 67: These are the model classes referenced by the events API. Basically, `Content` and its subtypes.

I wonder though:

1. Should we make this module completely independent from `nessie-model`?
2. Will this work in the presence of custom content types? I think so, but we can re-evaluate this after #6618. The only issue for custom content types would be for users: to deserialize a custom serialized content, they would need the concrete implementation.

events/api/src/main/java/org/projectnessie/events/api/MergeEvent.java
```diff
@@ -0,0 +1,57 @@
...
+@Schema(
+    type = SchemaType.OBJECT,
+    description =
+        "An event that is emitted when a merge is performed. "
+            + "This event is emitted after the merge has been persisted.")
+@Value.Immutable
+@JsonTypeName("MERGE")
+@JsonSerialize
+@JsonDeserialize
+public interface MergeEvent extends CommittingEvent {
```
adutra, line 39: Note that, as agreed, `MergeEvent` and `TransplantEvent` do not include the individual commits that were added to the target branch. This is to prevent huge event payloads, but the resulting objects are rather tiny now.

I wonder if we could improve this.

Assume we have a merge operation that appended 10 commits to the target branch, containing 50 operations in total.

We currently would emit:

1. One `MergeEvent`;
2. 50 `ContentStoredEvent` or `ContentRemovedEvent` events;

That is, 51 events total.

We could instead emit:

1. One `MergeEvent`;
2. 10 `CommitEvent`s;
2. 50 `ContentStored` or `ContentRemoved` events.

That is, 61 events total.

The advantage of this is that individual commit events would be sent along will their `CommitMeta` objects.

The drawback is the "write amplification" phenomenon that gets even more exacerbated.

events/api/src/main/java/org/projectnessie/events/api/Event.java
```diff
@@ -0,0 +1,146 @@
...
+  @JsonSubTypes.Type(name = "MERGE", value = ImmutableMergeEvent.class),
+  @JsonSubTypes.Type(name = "TRANSPLANT", value = ImmutableTransplantEvent.class),
+  @JsonSubTypes.Type(name = "REFERENCE_CREATED", value = ImmutableReferenceCreatedEvent.class),
+  @JsonSubTypes.Type(name = "REFERENCE_UPDATED", value = ImmutableReferenceUpdatedEvent.class),
+  @JsonSubTypes.Type(name = "REFERENCE_DELETED", value = ImmutableReferenceDeletedEvent.class),
+  @JsonSubTypes.Type(name = "CONTENT_STORED", value = ImmutableContentStoredEvent.class),
+  @JsonSubTypes.Type(name = "CONTENT_REMOVED", value = ImmutableContentRemovedEvent.class)
+})
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+public interface Event {
```
adutra, line 78: Should we have `Event` extend `Serializable`? Or is it too old-school by now?

events/api/build.gradle.kts
snazy: They wouldn't need the concrete implementation to deserialize an unknown content-type - they'd get a `GenericContent` object as a fallback (that particular `GenericContent` class shouldn't be used directly though)

events/api/build.gradle.kts
```diff
@@ -0,0 +1,89 @@
...
+    isCanBeResolved = false
+    description = "Source OpenAPI spec files, containing the examples"
+  }
+
+val generateOpenApiSpec =
+  tasks.named<SmallryeOpenApiTask>("generateOpenApiSpec") {
+    inputs.files("src/main").withPathSensitivity(PathSensitivity.RELATIVE)
+  }
+
+artifacts { add(openapiSource.name, file("src/main/resources/META-INF")) }
```
snazy, line 89: I guess this would need to be published to swaggerhub as well - and added as a release artifact.
Aka: changes to the `release-publish.yml` workflow

events/api/src/main/java/org/projectnessie/events/api/ReferenceEvent.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.projectnessie.events.api;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
+/** Event that is emitted when a reference is created, updated or deleted. */
+public interface ReferenceEvent extends Event {
+
+  /** The name of the reference. */
+  String getReferenceName();
```
snazy, line 24: Don't we have the "full reference name" here?

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
```diff
@@ -0,0 +1,60 @@
...
+ * <ul>
+ *   <li>{@link CommitEvent}: for commits;
+ *   <li>{@link MergeEvent}: for merges;
+ *   <li>{@link TransplantEvent}: for transplants.
+ * </ul>
+ */
+public interface CommittingEvent extends Event {
+
+  /** The source branch where the committed operations come from. */
+  String getSourceBranch();
```
snazy, line 34: Source doesn't need to be a branch - can be anything - even a "detached" commit.

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.projectnessie.events.api;
+
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+/** An enum of all possible reference types. */
+@Schema(
+    description = "An enum of all possible reference types.",
+    enumeration = {"BRANCH", "TAG"})
+public enum ReferenceType {
```
snazy, line 24: Probably worth to add the same "fallback" logic as introduced in #6634

events/api/src/main/java/org/projectnessie/events/api/ContentStoredEvent.java
```diff
@@ -0,0 +1,60 @@
...
+@JsonDeserialize
+public interface ContentStoredEvent extends ContentEvent {
+  @Override
+  @Value.Default
+  default EventType getType() {
+    return EventType.CONTENT_STORED;
+  }
+
+  /** The content that was stored. */
+  Content getContent();
```
snazy, line 49: Probably worth to ensure that content (de)serialization happens in the same way as elsewhere, so allowing to (de)serialize a `GenericContent`

events/api/build.gradle.kts
adutra: I need to test, but my fear is that a Kafka consumer may fail to deserialize a `GenericContent` because `GenericContent` is not part of the schema.

More specifically, [here](https://github.com/confluentinc/schema-registry/blob/v7.5.0-532/json-schema-serializer/src/main/java/io/confluent/kafka/serializers/json/AbstractKafkaJsonSchemaDeserializer.java#L178) the Kafka deserializer attempts to validate the schema. Even worse: [here](https://github.com/confluentinc/schema-registry/blob/v7.5.0-532/json-schema-serializer/src/main/java/io/confluent/kafka/serializers/json/AbstractKafkaJsonSchemaDeserializer.java#L249), it will attempt to locate the exact subtype among the `oneOf` alternatives available. These operations will fail if the subtype is not explicitly listed.

(This Kafka deserializer doesn't honor annotations such as `@JsonTypeIdResolver`.)

I think it would be safer to cut the dependency to nessie:model completely and provide a simplified `Content` interface here. Wdyt?

events/api/src/main/java/org/projectnessie/events/api/ReferenceEvent.java
adutra: I completely forgot it :-)

events/api/build.gradle.kts
snazy: > cut the dependency to nessie:model completely

SGTM

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
adutra: Do we really need that though? Nessie does not need to deserialize this from json. Only users will. I'd say it's their business :-)

events/api/build.gradle.kts
adutra: Ah good point.

Taking a step back, I think I'm going to remove the OpenAPI bits. All I wanted was for users to have a Json schema to use, e.g. in Kafka. But we don't need a full-blown OpenAPI spec here, since there is no operations in this package (no REST endpoints.)

Searching a bit, it seems that many Kafka users are solving this problem using this project: https://github.com/mbknor/mbknor-jackson-jsonSchema . It can generate Json schemas from POJOs and honors all Jackson annotations, including `@JsonSubTypes`.

So I suggest that we remove OpenAPI for now, and when I get to the reference implementation, we can reassess the need for a Json schema here.

events/api/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+The `Event` interface has a `getType()` method that returns the type of the event. The type is an
+enum that uniquely identifies the event type. This enum serves as a discriminator for the
+event type and allows subscribers to filter events by type. It also allows subscribers to
+easily serialize and deserialize events to and from JSON.
+
+The following concrete event implementations are defined:
+
+* `CommitEvent` - This event is published when a commit is created.
+* `MergeEvent` - This event is published when a merge is created.
```
dimas-b, line 21: nit: `when a merge is created`... maybe `when a merge is committed`? A merge is not an entity in Nessie... not sure about "creating a merge" :thinking:

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Value.Lazy
+  @JsonIgnore
+  default String getName() {
+    return getElements().get(getElements().size() - 1);
+  }
+
+  @Value.Lazy
+  @JsonIgnore
+  default Optional<ImmutableContentKey> getParent() {
```
dimas-b, line 47: Why not return `Optional<ContentKey>` or `Optional<? extends ContentKey>`?

events/api/src/main/java/org/projectnessie/events/api/ContentRemovedEvent.java
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Event that is emitted when a content is removed. This event corresponds to a DELETE operation in
+ * a commit, merge or transplant. This event is emitted after the content has been removed.
+ */
+@Value.Immutable
+@JsonTypeName("CONTENT_REMOVED")
+@JsonSerialize
+@JsonDeserialize
+public interface ContentRemovedEvent extends ContentEvent {
```
dimas-b, line 31: The design doc calls for a `Content type` attribute here... Did I miss it? Would it be meaningful at the `ContentEvent` level?

events/api/src/main/java/org/projectnessie/events/api/CustomContent.java
```diff
@@ -0,0 +1,44 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonTypeName("CUSTOM")
+@JsonSerialize
+@JsonDeserialize
+public interface CustomContent extends Content {
```
dimas-b, line 27: Is this type supposed to report add actual content data as `Content.getAttributes()`? If so, could you add javadoc to clarify?

events/api/src/main/java/org/projectnessie/events/api/Event.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /** The type of the event. */
+  EventType getType();
+
+  /**
+   * The id of the event. The UUID is generated by the event emitter. No assumptions should be made
+   * about the version of the UUID. Most UUIDs will be version 4 (random).
```
dimas-b, line 58: tangential: Would it make sense to use time UUIDs since we already have `getCreatedAt()`. Would time UUIDs be preferable to reduce the strain of the source of randomness in the server?

dimas-b: For that matter, WDYT about simply removing `Most UUIDs will be version 4 (random)` from the javadoc? :)

events/api/src/main/java/org/projectnessie/events/api/MergeEvent.java
dimas-b: API v2 does not permit unquashed merges (and API v1 is supposed to be phasing out of use), so the end result will be very similar: 1 merge event, 1 commit, 50 contents. So, I'd say let's do that.

Multiple commits are possible with transplants, but I would not call it write "amplification" in that case as it would be mostly equivalent to manually re-committing the same changes on the target branch.

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
dimas-b: If we do not serialize it but expose internal names, should we have a mapping from `ReferenceEvent.getFullReferenceName()` to `ReferenceType`?

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
```diff
@@ -0,0 +1,63 @@
...
+ *   <li>{@link CommitEvent}: for commits;
+ *   <li>{@link MergeEvent}: for merges;
+ *   <li>{@link TransplantEvent}: for transplants.
+ * </ul>
+ */
+public interface CommittingEvent extends Event {
+
+  /**
+   * The source reference where the committed operations came from. This is usually a branch, but
+   * not always (e.g. it could be a detached reference).
```
dimas-b, line 35: nit: Could also be a tag.

events/api/src/test/java/org/projectnessie/events/api/TestCommitMeta.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.events.api;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
dimas-b, line 18: Most of Nessie code uses `AssertJ` ;)

events/api/build.gradle.kts
snazy: That project looks unmaintained, the project license is okay though.

But yes, removing the openapi bits for events is fine.
Guess it makes sense to move the "schema concerns" to when we get to the reference implementation.

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
snazy: I agree that it's a "user concern" similar to Git, where the "simple name mapping" is rather a tool's thing.
But not opposed to have convenience methods (not "real" attributes) on the data types.

events/api/src/main/java/org/projectnessie/events/api/Event.java
snazy: Not sure how well time-UUIDs work in a container environment. They get their "unique" part from the encoded ethernet address, which was supposed to be a hardware thing. But with containers, you don't have that - or cannot assume that some ethernet address is "uniquely assigned by a hardware vendor". So pure-random UUIDs are probably fine. Further, it requires a lot of synchronisation to get time-UUIDs right - not just within a process, but strictly speaking across processes (think: _time_-UUID).

But +1 on removing any assumption on the UUID type.

events/api/README.md
adutra: You are right, how about "performed"?

events/api/src/main/java/org/projectnessie/events/api/ContentRemovedEvent.java
adutra: Good catch! That said I think the design doc is wrong... we can't know the content type of a DELETE. If you agree, I will amend the design doc in a separate PR.

events/api/src/main/java/org/projectnessie/events/api/CustomContent.java
adutra: Yes that's the intent. I will add some javadocs.

events/api/src/test/java/org/projectnessie/events/api/TestCommitMeta.java
adutra: Yeah, I was reluctant to add the dependency just for 3 cases so far: assert equals, assert true and assert false. But OK to add it.

events/api/src/main/java/org/projectnessie/events/api/MergeEvent.java
adutra: FYI https://github.com/projectnessie/nessie/pull/6672

events/api/src/main/java/org/projectnessie/events/api/ContentRemovedEvent.java
adutra: FYI https://github.com/projectnessie/nessie/pull/6672

events/api/src/main/java/org/projectnessie/events/api/Event.java
adutra: ...and for completeness, this is one of the few generators that handle version 6 and 7:

https://github.com/f4b6a3/uuid-creator

events/api/src/test/java/org/projectnessie/events/api/TestCommitMeta.java
dimas-b: I'd say let's add it - it's not a heavy artifacts and the build as a whole pulls it anyway - and we get uniformity of test style across the codebase (TBH, I do not often remember how to write assertions in the plain JUnit way anymore :) )

events/api/src/main/java/org/projectnessie/events/api/Event.java
dimas-b: As for the MAC part, I believe the spec allows generating it randomly (which we'd have to do at most once per start).

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
dimas-b: Not quite, I'm suggesting to have something like `isBranch()` derived from `getFullReferenceName()` (not sure it is possible, though)

events/api/README.md
dimas-b: `performed` SGTM :+1:

events/api/src/main/java/org/projectnessie/events/api/ContentRemovedEvent.java
dimas-b: Yes, #6672 makes sense

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
adutra: Ah OK I can add that too, but it will likely be derived from `getReferenceType()`.

events/api/src/main/java/org/projectnessie/events/api/CustomContent.java
dimas-b: I wonder whether we could have `Content.getAttributes()` simply duplicate all regular attributes _and_ allow it to include extra attributes... I guess that should make the handling `CustomContent` more straight-forward (i.e. avoid the `attributes` map in JSON). WDYT?

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
dimas-b: Now, I'm confused :) I thought we were going to drop `ReferenceType` from JSON, not so?

events/api/src/main/java/org/projectnessie/events/api/Content.java
```diff
@@ -0,0 +1,50 @@
...
+
+  ContentType getType();
+
+  /**
+   * Unique id for this object.
+   *
+   * <p>This id is unique for the entire lifetime of this Content object and persists across
+   * renames. Two content objects with the same key will have different ids.
+   */
+  String getId();
```
adutra, line 43: I think we can enforce `UUID` here, this seems to be the case for every new content stored in Nessie with both the old and the new storage models.

dimas-b: Yes, but the main model API defines it as a `String`: https://github.com/projectnessie/nessie/blob/main/api/model/src/main/java/org/projectnessie/model/Content.java#L78

If we changes it, I think we ought to change it in both places.

@snazy , WDYT?

events/api/src/main/java/org/projectnessie/events/api/CustomContent.java
adutra: > I wonder whether we could have Content.getAttributes() simply duplicate all regular attributes and allow it to include extra attributes...

I am not sure that this would make things easier tbh.

I have a code and tests in a separate branch that will handle this. Basically the mapping goes as follows:

```java
public static Content map(org.projectnessie.model.Content content) {
    org.projectnessie.model.Content.Type type = content.getType();
    if (type == org.projectnessie.model.Content.Type.NAMESPACE) {
      ...
    } else if (type == org.projectnessie.model.Content.Type.ICEBERG_TABLE) {
      ...
    } else if (type == org.projectnessie.model.Content.Type.ICEBERG_VIEW) {
      ...
    } else if (type == org.projectnessie.model.Content.Type.DELTA_LAKE_TABLE) {
      ...
    } else if (content instanceof GenericContent) {
      GenericContent genericContent = (GenericContent) content;
      ImmutableCustomContent.Builder builder =
          ImmutableCustomContent.builder()
              .id(UUID.fromString(Objects.requireNonNull(content.getId())))
              .customType(genericContent.getType().name());
      if (genericContent.getAttributes() != null) {
        builder.putAllAttributes(genericContent.getAttributes());
      }
      return builder.build();
    } else {
      Map<String, ?> map = MAPPER.convertValue(content, MAP_TYPE);
      return ImmutableCustomContent.builder()
          .id(UUID.fromString((String) map.remove("id")))
          .customType((String) map.remove("type"))
          .putAllAttributes(map)
          .build();
    }
  }
```

With that in mind, and assuming that `CustomAbcTable` is a non-built-in content type, e.g.:

```java
class CustomAbcTable extends org.projectnessie.model.Content {
      static final Type TYPE = new Type() {
            public String name() { return "CUSTOM_ABC_TABLE"; }
            public Class<? extends org.projectnessie.model.Content> type() { return CustomAbcTable.class; }
      };
      public Type getType() { return TYPE; }
      public String getId() { return UUID.randomUUID().toString();}
      public String getText() { return "foo"; }
      public int getNumber() { return 123; }
      public boolean isBoolean() { return true; }
}
```

Then the last `else` branch above will be used and will indeed produce a `CustomContent` that will be serialized as follows:

```json
{
  "type": "CUSTOM",
  "customType": "CUSTOM_ABC_TABLE",
  "id": "3171c949-b6f6-488a-9484-bd247c393223",
  "attributes": {
    "number": 123,
    "text": "foo",
    "boolean": true
  }
}
```

That is for non-built-in types. For the built-in ones, their serialization is straightforward and the `getAttributes()` map will be empty for now. Do you think we should duplicate their attributes there?

My thinking about placing the `getAttributes()` method in the parent `Content` interface was simply that it could be useful in the future if e.g. `IcebergTable` exposed a `getExtendedMetadata()` map or stg in the style.

events/api/src/main/java/org/projectnessie/events/api/Content.java
adutra: Whoops I went ahead and modified it, but I can revert if necessary.

events/api/src/main/java/org/projectnessie/events/api/ReferenceType.java
adutra: > Do we really need that though?

I was referring to the "fallback logic mentioned by @snazy :-)

dimas-b: Well, I think if we add `@JsonSerialize` we have to support schema/sub-type evolution too. Clients will probably be able to use a lenient attribute parsing mode, but will they be able to handle new sub-types if we do have a dedicated serializer? I guess the serializer has to be defined in our code. Not sure if clients can add that on top of our jars. WDYT?

adutra: Yes, you are both right. In the case where a newer event type is created and sent by Nessie, a consumer that has an older version of the Events API will not be able to deserialize it.

I will rework this.

adutra: Done in 66a19a2.

adutra: I pushed one more commit and now I think everything is taken care of. The 3 enums in this package were handled as follows:

1. `EventType` is handled by a special (de)serializer that deserializes unknown events as `CustomEvent`;
2. `ContentType` is handled by a special (de)serializer that deserializes unknown contents as `CustomContent`; 
3. `ReferenceType` was deleted completely. I was convinced by @dimas-b :-)

Let me know what you think.

events/api/src/main/java/org/projectnessie/events/api/CustomContent.java
adutra: Update: the most recent changes modify slightly how a `CustomContent` is serialized; here is the updated example:

```json
{
  "type": "CUSTOM_ABC_TABLE",
  "id": "3171c949-b6f6-488a-9484-bd247c393223",
  "number": 123,
  "text": "foo",
  "boolean": true
}
```

events/api/src/main/java/org/projectnessie/events/api/Event.java
dimas-b: Certainly a separate issue for time UUIDs :+1:

adutra: There you go: https://github.com/projectnessie/nessie/issues/6689.

events/api/src/main/java/org/projectnessie/events/api/Content.java
dimas-b: Deviating from the `model` API on the type of content ID looks risky to me. I'd prefer reverting for now.

I'm also ok with migrating to `UUID` as the java type for content IDs everywhere (in a separate PR).

adutra: You are right, I just reverted that.

events/api/src/main/java/org/projectnessie/events/api/MergeEvent.java
```diff
@@ -0,0 +1,40 @@
...
+@JsonDeserialize(as = ImmutableMergeEvent.class)
+public interface MergeEvent extends CommittingEvent {
+  @Override
+  @Value.Default
+  default EventType getType() {
+    return EventType.MERGE;
+  }
+
+  /** The hash of the common ancestor of the two merged branches. */
+  String getCommonAncestorHash();
```
dimas-b, line 39: We are going to allow merging unrelated histories (cf. #6626). Should we perhaps mention that this value may be the virtual "no ancestor" hash (cf. #6635)?

events/api/src/test/java/org/projectnessie/events/api/TestJsonSerde.java
```diff
@@ -0,0 +1,373 @@
...
+            .sentBy("Nessie")
+            .putProperty("string", "foo")
+            .putProperty("number", 123)
+            .putProperty("boolean", true)
+            .build();
+    assertThat(deserialize(serialize(event), Event.class)).isEqualTo(event);
+  }
+
+  @Test
+  void customContent() throws Exception {
```
dimas-b, line 255: nit: maybe add a test for `Namespace` too... or did I miss it?

events/api/src/test/java/org/projectnessie/events/api/TestJsonSerde.java
```diff
@@ -0,0 +1,373 @@
...
+
+  @Test
+  void unknownContentSerialization() throws Exception {
+    CustomContent content =
+        ImmutableCustomContent.builder()
+            .id("id")
+            .customType("weird")
+            .putProperty("string", "foo")
+            .putProperty("number", 123)
+            .putProperty("boolean", true)
```
dimas-b, line 330: Could you add a test case for a custom type with some data in `Content.getProperties()`?

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * The name of the key.
+   *
+   * <p>The key name is the last element of the key. For example, the name of the key {@code ["a",
+   * "b", "c"]} is {@code "c"}.
+   */
+  @Value.Lazy
+  @JsonIgnore
+  default String getName() {
```
dimas-b, line 52: Should we add a getter for the encoded full name string (as we have in `Namespace`)?

events/api/src/main/java/org/projectnessie/events/api/json/ContentTypeIdResolver.java
```diff
@@ -0,0 +1,64 @@
...
+
+  @Override
+  public JavaType typeFromId(DatabindContext context, String id) {
+    Class<?> subtype;
+    try {
+      subtype = ContentType.valueOf(id.toUpperCase(Locale.ROOT)).getSubtype();
+    } catch (Exception e) {
+      subtype = ContentType.CUSTOM.getSubtype();
+    }
+    assert baseType.getRawClass().isAssignableFrom(subtype);
```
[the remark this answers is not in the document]
adutra: This is really a documenting assertion, nothing more: remind the reader that `subtype` is necessarily a subtype of `baseType.getRawClass()`. Do you prefer to have it removed?

events/api/src/main/java/org/projectnessie/events/api/Event.java
```diff
@@ -0,0 +1,97 @@
...
+   * The user that created the event. The user is set by the event emitter and is based on the
+   * authenticated user that performed the action that caused the event to be emitted.
+   *
+   * <p>If authentication is disabled, this will be empty.
+   */
+  Optional<String> getCreatedBy();
+
+  /**
+   * A distinctive identifier for the entity that emitted the event. This is configured on a per-
+   * instance basis and usually contains the host name and the deployed Nessie version.
```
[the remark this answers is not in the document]
adutra: > Why would we want to expose it in events?

I understand the security concern, but events are more secure than API cals. Do you think we should refrain from indicating the version?

> Would it make sense to use specVersion in events too?

Yes, I think so. I will make sure we add that – but keep reading :-)

> would it be preferable to separate the version/spec part from the actual instance/host ID?

Keeping in mind that this will be sent along with _every_ event, I think it's important to keep this lightweight. I suggest the following:

1. Keep this `getSentBy` attribute as just the server hostname, nothing more.
2. In the events SPI package (not reviewable yet but already implemented):
    1. leverage the `onSubscribe(EventSubscription)` method that is invoked once at startup for each subscriber;
    2. use the `EventSubscription` type to communicate to each subscriber the full server configuration, including but not limited to: server version, server spec version, default branch, etc. We can even pass the whole `NessieConfiguration` object, and maybe even more: the whole Quarkus configuration as a `Map<String,String>`. 
    3. If they wish, subscribers can store this info and augment the received events with all the info they need, e.g. when using Kafka, this can be easily done by setting headers on the `ProducerRecord`.

Wdyt?

In fact, if we go this route, we should even remove `repositoryId` from the Event interface as it is a global setting too.

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
```diff
@@ -0,0 +1,45 @@
...
+ *   <li>{@link TransplantEvent}: for transplants.
+ * </ul>
+ */
+public interface CommittingEvent extends Event {
+
+  /**
+   * The source reference where the committed operations came from. This is usually a branch, but
+   * not always (e.g. it could be a tag or a detached reference).
+   */
+  String getSourceReference();
```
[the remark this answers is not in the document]
adutra: Always identical to the `targetBranch`.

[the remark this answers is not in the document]
adutra: I will add a note in the javadoc.

events/api/src/test/java/org/projectnessie/events/api/TestJsonSerde.java
adutra: Good catch!

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
adutra: I gave this a try in d0f908b, let me know what you think.

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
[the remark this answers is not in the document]
dimas-b: Thx (it was not obvious to me :) )

events/api/src/main/java/org/projectnessie/events/api/Event.java
```diff
@@ -0,0 +1,100 @@
...
+  @JsonSubTypes.Type(name = "CONTENT_STORED", value = ImmutableContentStoredEvent.class),
+  @JsonSubTypes.Type(name = "CONTENT_REMOVED", value = ImmutableContentRemovedEvent.class),
+  @JsonSubTypes.Type(name = "CUSTOM", value = ImmutableCustomEvent.class),
+})
+@JsonTypeIdResolver(EventTypeIdResolver.class)
+@JsonTypeInfo(use = JsonTypeInfo.Id.CUSTOM, property = "type", visible = true)
+public interface Event {
+
+  /** The root, empty commit with no ancestor. */
+  String NO_ANCESTOR_HASH = "2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d";
```
dimas-b, line 57: Not sure about pinning it in a public constant. PR #6635 allows servers some latitude in choosing the "no ancestor" hash value.

Can we simply refer to `NessieConfiguration.getNoAncestorHash()`?

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * The source reference where the committed operations came from. This is usually a branch, but
+   * not always (e.g. it could be a tag or a detached reference).
+   *
+   * <p>For commits, this is always the same as the {@linkplain #getTargetBranch() target branch}.
+   */
+  String getSourceReference();
+
+  /** The target branch where the committed operations were applied to. */
+  String getTargetBranch();
```
snazy, line 40: Should we use the full-ref-name here to allow for future changes?
For example, if we wanna have a separate reference type for transactions.

events/api/src/main/java/org/projectnessie/events/api/ReferenceEvent.java
```diff
@@ -0,0 +1,65 @@
...
+public interface ReferenceEvent extends Event {
+
+  /** The reference type name for branches. */
+  String BRANCH = "BRANCH";
+
+  /** The reference type name for tags. */
+  String TAG = "TAG";
+
+  /** The name of the reference, e.g. "branch1". */
+  String getReferenceName();
```
snazy, line 41: Nit: rename to `getSimpleReferenceName()` and mention in the javadoc that this might be ambiguous?

events/api/src/main/java/org/projectnessie/events/api/ReferenceEvent.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /** The full name of the reference, e.g. "refs/heads/branch1". */
+  String getFullReferenceName();
+
+  /**
+   * The type of the reference. This is usually either {@value #BRANCH} or {@value #TAG}, but more
+   * types may be added in the future.
+   */
+  String getReferenceType();
```
snazy, line 50: Nit: maybe mention that the reference type can also be derived from the full reference name?

events/api/src/main/java/org/projectnessie/events/api/Event.java
```diff
@@ -0,0 +1,100 @@
...
+ * @see MergeEvent
+ * @see TransplantEvent
+ * @see ReferenceCreatedEvent
+ * @see ReferenceUpdatedEvent
+ * @see ReferenceDeletedEvent
+ * @see ContentStoredEvent
+ * @see ContentRemovedEvent
+ */
+@JsonSubTypes({
+  @JsonSubTypes.Type(name = "COMMIT", value = ImmutableCommitEvent.class),
```
snazy, line 42: Nit: remove `Immutable` from the `value` attributes.

events/api/src/main/java/org/projectnessie/events/api/Content.java
```diff
@@ -0,0 +1,59 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver;
+import java.util.Collections;
+import java.util.Map;
+import org.immutables.value.Value;
+import org.projectnessie.events.api.json.ContentTypeIdResolver;
+
+/** An object stored in Nessie, such as a table or a view. */
+@JsonSubTypes({
+  @JsonSubTypes.Type(name = "ICEBERG_TABLE", value = ImmutableIcebergTable.class),
```
snazy, line 28: Nit: remove `Immutable` from the types specified via `value`

events/api/src/main/java/org/projectnessie/events/api/Event.java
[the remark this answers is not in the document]
snazy: > we should even remove repositoryId from the Event interface

`repositoryId` would be needed for multi-tenancy.

> Would it make sense to use specVersion in events too?

IMO yes. A _spec_ version is fine. Not a fan of a particular server version, because then you have to know the exact _release_ version that introduces a particular change. A _spec_ version is IMO a better "grouping". It doesn't need to be the same spec version that we expose via `NessieConfiguration` though.

Why do we need the server / pod hostname? What's the use case for consumers?

Subscribers might need some configuration (could think of retry limits or so). But not the full configuration, which does include secrets. Let's discuss on the SPI PR.

events/api/src/main/java/org/projectnessie/events/api/json/ContentTypeIdResolver.java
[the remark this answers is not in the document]
snazy: Similar to `org.projectnessie.model.types.ContentTypeIdResolver#typeFromId` - the baseType doesn't necessarily need to be assignable from subtype. Note that baseType is actually the type you pass to Jackson to read an object.

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
snazy: Wonder whether it makes sense to introduce a new module that's used by `:nessie-model` and `:nessie-events-api` for all the relevant conversions, so we don't have to maintain those in multiple places. WDYT? Or even have a `BaseContentKey` class used by both?

events/api/src/main/java/org/projectnessie/events/api/Event.java
snazy: Not sure whether we need to expose it for events at all.

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
snazy: Maybe also for the actual `Content` and derived types. If, then it's probably worth to do it in a separate PR?

events/api/src/main/java/org/projectnessie/events/api/Event.java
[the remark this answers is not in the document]
adutra: > repositoryId would be needed for multi-tenancy.

OK, let's keep it.

> Why do we need the server / pod hostname? What's the use case for consumers?

I thought it is important for consumers to know who / what created the event.

That said, server / pod hostname is also global to the Nessie instance, so we could remove `getSentBy` completely from the Event body and pass it to consumers using the `onSubscribe(EventSubscription)` method described above.

To get the ball rolling, I suggest that we remove `getSentBy` completely for now, and re-evaluate the best way to communicate this information + spec version to consumers, in the SPI PR.

events/api/src/main/java/org/projectnessie/events/api/json/ContentTypeIdResolver.java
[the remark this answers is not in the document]
adutra: Good catch, I will amend this and add tests specifically for that.

events/api/src/main/java/org/projectnessie/events/api/Event.java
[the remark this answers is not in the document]
adutra: Done in f00e395.

events/api/src/main/java/org/projectnessie/events/api/json/ContentTypeIdResolver.java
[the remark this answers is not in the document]
adutra: Done in 5ee72af.

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
adutra: You are right, but at this point, I think we need to introduce a separate `Reference` type. I will try that.

On a side note, I think I will also rename all occurrences of the word "Branch" as in this method with "Reference" which sounds more future-proof, wdyt? E.g. `getTargetReference()`.

events/api/src/main/java/org/projectnessie/events/api/MergeEvent.java
```diff
@@ -0,0 +1,55 @@
...
+   * this will be the so-called {@link #NO_ANCESTOR_HASH} hash. You can use the {@link
+   * #hasCommonAncestorHash()} method to check if the two branches share a common history.
+   */
+  String getCommonAncestorHash();
+
+  /** Returns {@code true} if the two branches share a common history, {@code false} otherwise. */
+  @Value.Derived
+  @JsonIgnore
+  default boolean hasCommonAncestorHash() {
+    return !NO_ANCESTOR_HASH.equals(getCommonAncestorHash());
```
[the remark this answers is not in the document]
adutra: Done in c6a42aa14.

events/api/src/main/java/org/projectnessie/events/api/Event.java
adutra: Actually, it seems that `@JsonSubTypes` is not needed anymore: we are using a custom strategy instead to resolve the type ids.

events/api/src/main/java/org/projectnessie/events/api/Event.java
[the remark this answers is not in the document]
dimas-b: SGTM :+1:

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
dimas-b: +1 to extracting common code into a separate module.

events/api/src/main/java/org/projectnessie/events/api/CommittingEvent.java
dimas-b: `getTargetReference()` SGTM

adutra:
@snazy and @dimas-b we are getting close to a mergeable state and I'd like to have your opinion on some naming matters before we merge:

1. The API is using the word "custom" for type ids that are not known at runtime. e.g. `CustomEvent` and `CustomContent`. Is this the best word? Maybe "generic" or "unknown"?
2. The `Event` class has some attributes names that might be confusing for users, what about:
    1. `getCreatedAt`  => `getEventCreationTimestamp()` (to avoid confusion with the commit timestamp)
    2. `getCreatedBy()` => `getEventCreationUsername()` (to avoid confusion with commit author or committer)
    
Wdyt?

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
snazy: Validation and such can be specific to the "concrete" `Content` (et al) types in `:nessie-model` and `:nessie-event-api`. I was (maybe naively) thinking of a common abstract base class or interface in a new module and have concrete `@Immutable`s with the "right annotations".

snazy:
> 1. The API is using the word "custom" for type ids that are not known at runtime. e.g. `CustomEvent` and `CustomContent`. Is this the best word? Maybe "generic" or "unknown"?

Yea, I agree. "Generic" sounds good to me.

> 2. The `Event` class has some attributes names that might be confusing for users, what about:
>    1. `getCreatedAt`  => `getEventCreationTimestamp()` (to avoid confusion with the commit timestamp)
>    2. `getCreatedBy()` => `getEventCreationUsername()` (to avoid confusion with commit author or committer)

+1 to `getEventCreation*()`

Agree to @dimas-b - pretty close, just a few things left (I'm often distracted and not good w/ proper reviews ATM).

dimas-b:
> The API is using the word "custom" for type ids that are not known at runtime. e.g. CustomEvent and CustomContent. Is this the best word? Maybe "generic" or "unknown"?

My preference is "generic".

>  getCreatedAt => getEventCreationTimestamp() (to avoid confusion with the commit timestamp)

+1 to `getEventCreationTimestamp()`

>  getCreatedBy() => getEventCreationUsername() (to avoid confusion with commit author or committer)

I'd avoid "username". In some contexts it may be just an email or some ID, I think.

How about `getCommitter()`? The committer data is the same for the actual commit and event, so there should be no confusion, I guess.

adutra:
> How about getCommitter()? 

Wouldn't it be confusing if the event is not a `CommitEvent` though?

adutra:
How about `getEventCreator()`? Or `getEventInitiator()`?

events/api/src/main/java/org/projectnessie/events/api/ContentKey.java
adutra: Ah yes, that could work 👍

dimas-b:
`getEventInitiator()` LGTM.

dimas-b (approved):
LGTM :+1: Thanks @adutra !

snazy (approved):
+1, but let's tackle the "code deduplication" work next

PR #6647, opened 2023-04-21
Events notification system for Nessie - VersionStore changes

adutra:
This PR is part of a series of PRs related to the events notification system for Nessie.

See design doc: https://github.com/projectnessie/nessie/blob/main/design/events.md.

This PR focuses on changes required to the existing `VersionStore` API and implementations.

Summary of changes:

* New `EventsVersionStore` implementation following the delegate pattern already used for metrics and tracing. **Not used for now**.
* Most methods in `VersionStore` now return more detailed results:
  * `commit()` returns `CommitResult`
  * `assign()` returns `ReferenceAssignedResult`
  * `create()` returns `ReferenceCreatedResult`
  * `delete()` returns `ReferenceDeletedResult`
* The existing `MergeResult` now has more methods:
  * `BranchName getSourceBranch();`
  * `List<COMMIT> getAddedCommits();` **this is the most tricky addition**
* The new `LazyPut` interface: contrary to `Put`, it does not deserialize the content value until it's necessary to do so. (The goal is to avoid deserializing contents on Nessie's write path, if the REST API doesn't need it.)

**All these additions are meant to be consumed by the events system, not by the REST API which does not change at all, as required by the design doc.**


versioned/spi/src/main/java/org/projectnessie/versioned/LazyPut.java
```diff
@@ -0,0 +1,63 @@
...
+import org.projectnessie.model.ContentKey;
+import org.projectnessie.nessie.relocated.protobuf.ByteString;
+import org.projectnessie.versioned.store.DefaultStoreWorker;
+
+/**
+ * A PUT operation that has been retrieved from the version store. The content value is kept in its
+ * raw (serialized) form; deserialization happens lazily, on first access to {@link #getValue()}.
+ */
+@Value.Immutable
+public interface LazyPut extends Operation {
```
snazy, line 30: Having `Put` and `LazyPut` looks a bit dangerous to me. Both serve the same operation ("put"), but `instanceof` checks against either excludes the other one.

Maybe make `Put` a "bare" interface and have concrete eager and lazy implementations so all users of `Put` can stay as they currently are and no user needs to know what particular implementation will be there.

Potentially also with factory methods only in `Put` (so no top-level type `LazyPut`)

versioned/spi/src/main/java/org/projectnessie/versioned/CommitResult.java
```diff
@@ -0,0 +1,35 @@
...
+
+@Value.Immutable
+public interface CommitResult<COMMIT> extends Result {
+
+  @Override
+  default ResultType getResultType() {
+    return ResultType.COMMIT;
+  }
+
+  COMMIT getCommit();
```
snazy, line 28: Nit: add a `getCommitHash` convenience method (to avoid all the added `.getCommit().getHash()` chains)

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -110,6 +110,7 @@ default Hash commit(
    * to concurrent readers/writers. The sequence to transplant must be contiguous, in order and
    * share a common ancestor with the target branch.
    *
+   * @param sourceBranch The branch we're transplanting from
```
snazy: Seems this new parameter's only there to populate the field in `MergeResult`.
I suspect, we can just add 
```java
MergeResult<T> withSourceBranch(BranchName sourceBranch);

    MergeResult<T> withResultType(ResultType resultType);
```
to `MergeResult` and call it from the call sites of `transplant()` and `merge()`. WDYT?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
```diff
@@ -647,4 +649,13 @@ MergeResult<Commit> mergeTransplantSuccess(
 
     return mergeResult.resultantTargetHash(objIdToHash(newHead)).wasApplied(true).build();
   }
+
+  Commit commitObjToCommit(CommitObj newCommit) throws ReferenceNotFoundException {
+    try {
+      ContentMapping contentMapping = new ContentMapping(persist);
+      return contentMapping.commitObjToCommit(true, newCommit);
+    } catch (ObjNotFoundException e) {
+      throw referenceNotFound(e);
```
snazy, line 658: This (or the only call size) could just re-throw it as a `RuntimeException`. Hitting this exception means that something went terribly wrong: content value stored, but cannot be retrieved - that's a bad situation.
Whereas `ReferenceNotFoundExcepiton` tends to indicate a user-error

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/ContentMapping.java
```diff
@@ -132,19 +132,17 @@ public Commit commitObjToCommit(
...
           if (c.action().exists()) {
-            commit.addOperations(
-                Put.of(
-                    key,
-                    contentMapping.fetchContent(
-                        requireNonNull(c.value(), "Required value pointer is null"))));
+            ObjId objId = requireNonNull(c.value(), "Required value pointer is null");
+            ContentValueObj contentValue =
+                persist.fetchTypedObj(objId, VALUE, ContentValueObj.class);
+            commit.addOperations(LazyPut.of(key, contentValue.payload(), contentValue.data()));
```
snazy, line 145: I know, it's not changed here, but probably worth to first collect all `objId`s to load and then to a bulk-load instead of one `fetchTypedObj` for each `objId` individually.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
```diff
@@ -412,6 +417,6 @@ private void commitAddPut(
 
   private String contentIdFromContent(@Nonnull @jakarta.annotation.Nonnull ObjId contentValueId)
       throws ObjNotFoundException {
-    return new ContentMapping(persist).fetchContent(contentValueId).getId();
+    return contentMapping.fetchContent(contentValueId).getId();
```
snazy, line 420: Can you do me a favor and open an issue indicating that we should pre-load these objects, so they are bulk-loaded and in turn available via the cache here?

snazy (commented):
Not a full review (and not done yet), but went though a bunch of files.

Adding all the `*Result` objects to the `VersionStore` functions is fine.
Not so much a fan of the new "top-leve" `LazyPut` (commented). 

I suspect, with some of the suggestions applied, the size of the PR will reduce a bit.

versioned/spi/src/main/java/org/projectnessie/versioned/CommitResult.java
adutra: It's a bit tricky because `COMMIT` can be anything, but I'll find a way.

versioned/spi/src/main/java/org/projectnessie/versioned/LazyPut.java
adutra: So, I managed to reunite `Put` and `LazyPut`. However I had to turn it into an abstract class because I needed a custom equals() method.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
adutra: Here you are: https://github.com/projectnessie/nessie/issues/6673.

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestIndexesLogicImpl.java
```diff
@@ -76,7 +78,7 @@ public void completeIndexesInCommitChain(boolean discrete) throws Exception {
     }
     ObjId secondaryHeadId = secondary.get(0);
     CommitObj secondaryHead = commitLogic.fetchCommit(secondaryHeadId);
-    soft.assertThat(secondaryHead).isNotNull();
+    assertThat(secondaryHead).isNotNull();
```
snazy, line 81: Better leave the `soft.assertThat()` and add a `soft.assertAll()` after it.
Otherwise previous soft-assertions would get lost. (Same below)

versioned/spi/src/main/java/org/projectnessie/versioned/Put.java
```diff
@@ -15,24 +15,35 @@
...
    *
    * @return the value
    */
-  Content getValue();
+  @Value.Lazy
+  public Content getValue() {
+    return getValueSupplier().get();
+  }
+
+  protected abstract Supplier<Content> getValueSupplier();
```
snazy, line 43: Should this be annotated with `@Value.Auxiliary` ?

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -346,23 +364,31 @@ public MergeResult<CommitLogEntry> transplant(TransplantParams transplantParams)
   }
 
   @Override
-  public Hash commit(CommitParams commitParams)
+  public CommitResult<CommitLogEntry> commit(CommitParams commitParams)
       throws ReferenceConflictException, ReferenceNotFoundException {
     try {
-      return opLoop(
+      AtomicReference<ImmutableCommitResult.Builder<CommitLogEntry>> commitResultHolder =
```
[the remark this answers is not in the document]
adutra: Hehehe, I initially frowned upon this as well. But this is already being done exactly like this in other methods in this class, look at `merge()` and `transplant()` for instance. So, I decided to use the same idiom for `commit()`, `create()`, `assign()` and `delete()`.

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestIndexesLogicImpl.java
adutra: That was to calm down IntelliJ that cannot understand soft assertions. But OK, I will revert.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/ContentMapping.java
```diff
@@ -131,25 +132,31 @@ public Commit commitObjToCommit(
     CommitMeta commitMeta = toCommitMeta(commitObj);
 
     if (fetchAdditionalInfo) {
-      ContentKey key;
-      ContentMapping contentMapping = new ContentMapping(persist);
       IndexesLogic indexesLogic = indexesLogic(persist);
+      Map<ObjId, ContentKey> idsToKeys = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
adutra: Some tests fail if we switch to `HashMap` because they expect operations to appear in a strict order. OK to apply your suggested changes.

versioned/spi/src/main/java/org/projectnessie/versioned/Put.java
adutra: We *could*, but it wouldn't change the fact that we would still need to provide a custom `equals` and `hashCode` methods, because we need `getValue()` to be included in the computation, but `getValue()` is marked `@Lazy`.

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: Hm - yea - you're right. It didn't work with the original return type. I'd be fine w/ the holder since it's code that'll go away soon-ish.

versioned/spi/src/main/java/org/projectnessie/versioned/Put.java
snazy: Technically we don't need equals & hashCode for the operations in nessie-versioned-spi - if it helps to make this simpler.

adutra: Indeed, but sadly, tests rely heavily on equality of `Operation` instances. We'd have to revisit a lot of test code... Not sure what's the least worse :-)

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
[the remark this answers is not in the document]
adutra: You call, I can either revert that commit or keep it. Since the old storage model is going away, I'm not sure what makes the most sense.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractMerge.java
```diff
@@ -475,6 +493,134 @@ private void doMergeIntoEmpty(
...
   }
 
+  private void checkAddedCommits(
+      boolean individualCommits,
+      Hash targetHead,
+      MergeResult<Commit> result,
+      boolean expectFastForward) {
+    if (individualCommits) {
+      if (expectFastForward) {
+        // FIXME assertion is failing for new storage model since BatchingPersist introduction
```
adutra, line 503: Fix will be in #6736.

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractTransplant.java
```diff
@@ -153,19 +162,67 @@ private BranchName checkTransplantOnEmptyBranch(
...
+                Arrays.asList(firstCommit, secondCommit, thirdCommit),
+                commitMetaModify,
+                individualCommits,
+                Collections.emptyMap(),
+                MergeBehavior.NORMAL,
+                false,
+                false);
+
+    if (individualCommits) {
+      // FIXME assertion is failing for new storage model since BatchingPersist introduction
```
adutra, line 182: Fix will be in #6736.

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
snazy: _sigh_ fair enough

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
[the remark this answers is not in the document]
snazy: nah - let's keep it now

versioned/spi/src/main/java/org/projectnessie/versioned/Put.java
snazy: yea, then let's keep it - would be "surprising" if the Put-immutable behaves differently.

versioned/persist/adapter/src/main/java/org/projectnessie/versioned/persist/adapter/spi/AbstractDatabaseAdapter.java
```diff
@@ -588,29 +593,29 @@ protected Hash mergeTransplantCommon(
...
               commitsToMergeChronological,
               newKeyLists,
               params.getUpdateCommitMetadata(),
-              mergePredicate);
-
-      // Write commits
+              mergePredicate,
+              addedCommits);
+
+      if (!params.isDryRun()) {
```
snazy, line 609: Would it break anything if we leave this `if` where it was before?
I don't mind having the copyCommits doing that in the old model (because it eventually goes away), if it's a lot of work.

versioned/persist/nontx/src/main/java/org/projectnessie/versioned/persist/nontx/NonTransactionalDatabaseAdapter.java
```diff
@@ -828,25 +852,48 @@ protected Hash casOpLoop(
...
+                    ctx, ref, refHead.getRef(), mergeResult.getResultantTargetHash());
             break;
           default:
             throw new UnsupportedOperationException("Unknown opVariant " + opVariant);
         }
 
         if (!casSuccess) {
-          if (opVariant == CasOpVariant.COMMIT) {
+          if (opVariant == CasOpVariant.COMMIT || opVariant == CasOpVariant.MERGE) {
             cleanUpCommitCas(ctx, individualCommits, individualKeyLists);
```
snazy, line 897: Just noting (nothing to change here) - only for posterity why this is _not_ in the new model:
Initially I ported this behavior to the new storage model as well, but (manual, bad "heavy load") tests proved that this can go sideways - up to the point that two concurrent writers produce the exact same commit, one wins and bumps HEAD, the other one "loses" and deletes the commit -> HEAD points to a non-existing commit. It's a rather artificial and will probably never happen in real-live, but still, there's a chance.

versioned/spi/src/main/java/org/projectnessie/versioned/EventsVersionStore.java
```diff
@@ -0,0 +1,223 @@
...
+            targetBranch,
+            referenceHash,
+            sequenceToTransplant,
+            updateCommitMetadata,
+            keepIndividualCommits,
+            mergeKeyBehaviors,
+            defaultMergeBehavior,
+            dryRun,
+            fetchAdditionalInfo);
+    if (!dryRun) {
```
snazy, line 94: Shouldn't this check `result.wasApplied()` instead to only emit the event, when the merge was actually persisted? (same for `merge()` below)

versioned/spi/src/main/java/org/projectnessie/versioned/MergeResult.java
```diff
@@ -72,6 +75,22 @@ default boolean wasSuccessful() {
...
+   * resulted in a fast-forward merge or transplant, because no new commit is created in this case.
+   * Otherwise, if commits were squashed, the returned list will contain exactly one element: the
+   * squashed commit; conversely, if individual commits were preserved, the list will generally
+   * contain as many commits as there were {@linkplain #getSourceCommits() source commits} to rebase
+   * (unless some source commits were filtered out).
+   *
+   * <p>The REST API does not expose this property currently; it is used by the Nessie events
+   * notification system.
+   */
+  List<COMMIT> getAddedCommits();
```
snazy, line 92: Nit: should this be renamed to `getPersistedCommits()` or `getCreatedCommits()`

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractReferences.java
```diff
@@ -69,12 +69,13 @@ public void createAndDeleteBranch() throws Exception {
     soft.assertThat(hash).isNotNull();
 
     final BranchName anotherBranch = BranchName.of("bar");
-    final Hash createHash = store().create(anotherBranch, Optional.of(hash));
+    final Hash createHash = store().create(anotherBranch, Optional.of(hash)).getHash();
```
snazy, line 72: Can you add some assertion on the `ReferenceCreatedResult` here and on `ReferenceDeletedResult` to the (unasserted) `.delete()` in line 103? And similar for tags below?

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractAssign.java
```diff
@@ -115,13 +115,13 @@ public void assignReferenceToFreshMain()
     }
 
     BranchName testBranch = BranchName.of("testBranch");
-    Hash testBranchHash = store.create(testBranch, Optional.empty());
+    Hash testBranchHash = store.create(testBranch, Optional.empty()).getHash();
```
snazy, line 118: Can you add some assertion on the `ReferenceAssignedResult` here and below?

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractCommits.java
```diff
@@ -95,7 +95,8 @@ public void commitToBranch() throws Exception {
                 branch,
                 Optional.of(initialHash),
                 CommitMeta.fromMessage("Some commit"),
-                Collections.emptyList());
+                Collections.emptyList())
```
snazy, line 98: Can you add some assertion on the `CommitResult` here?

snazy (commented):
Veeeery close :)

Just a few more minor-ish comments and then we can ship it.

snazy (approved):
Boom! Ship it! :)

PR #6648, opened 2023-04-21
bugfix: namespaceDepth filter loses entry content

XN137:
implicit namespaces are somewhat deprecated and thus also the namespaceDepth filter

however as long as the filter exists, it should not lose already retrieved information about the content of real entries

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -776,11 +776,15 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
               continue;
             }
 
+            Content c = key.getContent();
             EntriesResponse.Entry entry =
-                EntriesResponse.Entry.entry(key.getKey(), key.getType(), key.getContentId());
+                c != null
+                    ? EntriesResponse.Entry.entry(key.getKey(), key.getType(), c)
+                    : EntriesResponse.Entry.entry(key.getKey(), key.getType(), key.getContentId());
 
-            entry = namespaceDepthMapping(entry, depth);
+            entry = maybeTruncateToDepth(entry, depth);
 
+            // add implicit namespace entries only once (single parent of multiple real entries)
             if (seen.add(entry.getName())) {
               if (!pagedResponseHandler.addEntry(entry)) {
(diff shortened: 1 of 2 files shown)


servers/services/src/testFixtures/java/org/projectnessie/services/impl/AbstractTestEntries.java
```diff
@@ -341,14 +342,37 @@ public void filterEntriesByNamespaceAndPrefixDepth(ReferenceMode refMode)
...
+    entries = entries(reference, 4, filterThatFindsTwoTables);
     soft.assertThat(entries)
         .hasSize(2)
         .map(e -> immutableEntry(e.getName(), e.getType()))
         .containsExactlyInAnyOrder(
             immutableEntry(ContentKey.of("a", "b", "c", "secondTable"), Content.Type.ICEBERG_TABLE),
             immutableEntry(ContentKey.of("a", "b", "c", "firstTable"), Content.Type.ICEBERG_TABLE));
 
+    // namespaceDepth filter always returns contentId of real entries
+    entries = entries(reference.getName(), reference.getHash(), 4, filterThatFindsTwoTables, false);
```
XN137, line 355: before the fix:
```
java.lang.AssertionError: 
Expecting all elements of:
  [Entry{type=ICEBERG_TABLE, name=a.b.c.firstTable, contentId=null, content=null},
    Entry{type=ICEBERG_TABLE, name=a.b.c.secondTable, contentId=null, content=null}]
to satisfy given requirements, but these elements did not:
```
see how both contentId and content were always null

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -776,11 +776,15 @@ protected Set<Check> checksForEntry(KeyEntry entry) {
               continue;
             }
 
+            Content c = key.getContent();
             EntriesResponse.Entry entry =
-                EntriesResponse.Entry.entry(key.getKey(), key.getType(), key.getContentId());
+                c != null
+                    ? EntriesResponse.Entry.entry(key.getKey(), key.getType(), c)
+                    : EntriesResponse.Entry.entry(key.getKey(), key.getType(), key.getContentId());
```
XN137, line 783: note this is the same logic as in the else branch of the line 768 if statement

dimas-b (approved):
Nice find! :+1:

PR #6672, opened 2023-04-24
Events design doc - minor evolutions

adutra:
3 minor changes to the design doc:

1. Don't include content type in content events (because the info is not available for content removed events, and already implicitly present for content stored events);
2. Change Merge and transplant events to emit commit events, which in turn will emit content events (see discussion: https://github.com/projectnessie/nessie/pull/6646/files#r1173378052);
3. Simplify event delivery mechanism and merge `EventListener` into `EventSubscriber` (this is because I received many offline remarks that the original desing was unnecessarily convoluted, and that having just one interface to implement will make user's lives easier).


design/events.md
```diff
@@ -370,19 +367,19 @@ The system does not provide any ordering guarantees. Events may be emitted in an
...
-* Methods in the `EventListener` interface operate in fire and forget mode: the method MUST return as soon as the
-  should not wait for read acknowledgements.
-* The server COULD impose a (configurable) timeout after which the thread calling the `EventListener` method is
-  interrupted and delivery is considered failed.
+* `onXyz()` methods in the `EventSubscriber` interface operate in fire and forget mode: the method MUST return
+  immediately; it should not wait for read acknowledgements from remote systems.
+* The server COULD impose a (configurable) timeout after which the thread calling the `EventSubscriber` `onXyz()` method
+  is interrupted and delivery is considered failed.
 * A `RuntimeException` MAY be thrown to indicate that event delivery wasn't possible, e.g. because the underlying
   pub/sub system is unreachable. But in this case, the exception MUST be thrown immediately.
```
dimas-b, line 381: off topic to this PR, but I think this contradicts the requirements that `onXyz()` methods must not do blocking I/O on the calling thread. How can they immediately determine whether the even delivery was possible without blocking? If the intention is to allow subscribers to report asynchronously determined problems as exceptions on the notification thread, each call still cannot reliably throw if there's an error. What is the benefit of specifying this failure-handling behaviour? Could we simply say that subscribers MAY throw `RuntimeExceptions` without going into details?

From my POV we might as well add that subscribers SHOULD NOT throw `RuntimeExceptions` if they can avoid it. What can Nessie do about those exceptions, except perhaps log them?

dimas-b (approved):

design/events.md
adutra: I agree, this doesn't make a lot of sense :-) I will simplify as you suggest.

dimas-b (approved):

adutra:
license/cla is stuck again...

PR #6681, opened 2023-04-25
switch to gradle java toolchains

XN137:


snazy:
Let's be careful with toolchains, it's sadly not as easy as it sounds.

XN137:
i am mostly using this PR to run some experiments atm

what specific limitations/problems are you aware of with toolchains that we cant deal with?

snazy:
* Gradle downloading a JDK
* Gradle using the wrong JDK (new Java version CI runs)
* And something else I don't recall

XN137:
closing since this was mostly replaced with https://github.com/projectnessie/nessie/pull/6726 ...

i guess the remaining difference can only be properly cleaned up once we target java 11 bytecode for non-client code parts

PR #6682, opened 2023-04-25
set javac release param to 8

XN137:


snazy (commented):
Using the `release` option should be done at some point.
However, it's not as easy as this. Nessie _servers_ are not supported running on Java 8, only clients are. Quarkus builds _require_ Java 11.
Also need to adjust all the logic to compile for Java 11.

If we do this at some point, we should default to Java 11 (or even better: 17, if possible), and only build for 8 for the client modules.

Side note: Quarkus will bump the minimum Java version from 11 to 17 soon.

snazy:
Included in #6730

PR #6685, opened 2023-04-25
Make dry run merge / transplant throw exceptions

dimas-b:
The `dryRun` API parameter is different from `returnConflictAsResult`.

If a dry run is requested it should produce the same client-side effect as a normal merge / transplant.

At the `VersionStore` level this means throwing the same exception regardless of the `dryRun` parameter because `returnConflictAsResult` does not exist at that level.


dimas-b:
I'll make another PR for testing `returnConflictAsResult`

snazy (approved):
WFM (assuming all tests pass ;) )

snazy (approved):

PR #6687, opened 2023-04-25
Let `:nessie-versioned-spi` use OpenTelemetry

### LINKED ISSUE #5608: Use OpenTelemetry in `:nessie-versioned-spi`
Migrate the existing code from OpenTracing & Micrometer to OpenTelemetry.
Once that's done, the shim-code in `:nessie-quarkus` can be removed, and the dependency references to micrometer + opentracing removed from `libs.gradle.toml`.

snazy:
Migrate from OpenTracing to OpenTelemetry.

Updates some build files to refer to "one of the right opentelemetry BOMs" (there are 4 or so, but all interchangeable).

Fixes #5608


versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
```diff
@@ -138,11 +147,12 @@ public MergeResult<Commit> transplant(
     return TracingVersionStore
         .<MergeResult<Commit>, ReferenceNotFoundException, ReferenceConflictException>
             callWithTwoExceptions(
```
dimas-b, line 149: I wonder whether we could use simple delegation code here with method-level `@WithSpan` annotations and instrument the bytecode with OTel at build time... WDYT?

Span attributes can be added via `Span.current().setAttribute(...)`

versioned/spi/src/test/java/org/projectnessie/versioned/TestTracingVersionStore.java
```diff
@@ -259,89 +263,125 @@ void versionStoreInvocation(
...
+      //                expectedLogs, tracer.getActiveSpan().getLogs(), "expected logs don't
+      // match"),
+      //        () ->
+      //            assertEquals(
+      //                new HashMap<>(expectedTags),
+      //                tracer.getActiveSpan().getTags(),
+      //                "expected tags don't match"),
+      //        () -> assertTrue(tracer.isParentSet(), "Span-parent not set"),
+      //        () -> assertTrue(tracer.isClosed(), "Scope not closed"),
+      //        () -> assertTrue(tracer.getActiveSpan().finished(), "Span did not finish"));
```
dimas-b, line 383: Do we want to assert this?

dimas-b (commented):
LGTM - the `@WithSpan` comment is optional :)

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
snazy: In Quarkus, probably better, yea. Let's get that done for Quarkus.

dimas-b: AFAIK, instrumentation can be done in a generic way with Quarkus providing runtime OTel impl.

dimas-b (approved):

versioned/spi/src/main/java/org/projectnessie/versioned/TracingVersionStore.java
snazy: Yep, there's a way to do that.

PR #6698, opened 2023-04-27
Add extraEnv to helm chart

### LINKED ISSUE #6670: Helm Chart: Add option to get MongoDB Connection String from Secret
Currently when using a MongoDB as a backend, only Username and Password can be specified as a secrets.
However, MongoDB supports many different authentication options, not only Username and Password - and QUARKUS does as well.

For the specific case of MongoDB it would be great to add the option to get the `QUARKUS_MONGODB_CONNECTION_STRING` from a secret instead of from the values directly.

Additionally, it would be great to add a general `extraEnv` section to the values.yaml, that allows users to use additional configuration options of QUARKUS for different databases, even when they are not explicitly supported in the chart.

c-thiel:
Fixes #6670


helm/nessie/templates/deployment.yaml
```diff
@@ -184,6 +184,16 @@ spec:
...
+            {{- range $k, $v := .Values.advancedConfig }}
+            - name: {{ $k | upper | replace "\"" "_" | replace "." "_"  | replace "-" "_" }}
+              value: {{ $v | quote }}
+            {{- end }}
+            
+            {{- if .Values.extraEnv }}
+            {{- toYaml .Values.extraEnv | nindent 12 }}
+            {{- end }}
+
             {{ list .Values.advancedConfig "" | include "nessie.propsToEnvVars" | nindent 12 }}
```
adutra, line 197: 1. The `{{- range $k, $v := .Values.advancedConfig }}` section is not needed;
2. I think you should put the `extraEnv` section after `advancedConfig` as the former is even more powerful than the latter, so imho it should go last.

```suggestion

            {{ list .Values.advancedConfig "" | include "nessie.propsToEnvVars" | nindent 12 }}

            {{- if .Values.extraEnv }}
            {{- toYaml .Values.extraEnv | nindent 12 }}
            {{- end }}
```

adutra (changes requested):
Nice addition, I like how it's flexible and applies to other possible use cases as well.

helm/nessie/values.yaml
```diff
@@ -93,6 +93,20 @@ advancedConfig:
 #      console.format: "%d{HH:mm:ss} %s%e%n"
 #      category."org.projectnessie".level: DEBUG
 
+# -- Advanced configuration via Environment Variables.
+# Extra environment variables to add to the Nessie server container.
+# You can pass here any valid Nessie or Quarkus configuration properties.
```
adutra, line 98:
```suggestion
# You can pass here any valid EnvVar object:
# https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#envvar-v1-core
```

snazy:
@c-thiel can you sign the CLA?

c-thiel:
I signed the CLA for my Github Account. I did some doc changes unfortunately on my iPad with a wrongly configured author.
Is that still fine for you?
Otherwise I would need to create a new PR.

adutra:
Maybe you could squash your commits and force-push the squashed commit with your verified GitHub account?

c-thiel:
Ok, I did some rebase. You need to check again though.

adutra (approved):

PR #6717, opened 2023-04-30
CI: capture test reports

snazy:


.github/workflows/ci.yml
```diff
@@ -138,6 +138,16 @@ jobs:
         with:
           arguments: test -x :nessie-quarkus:test --scan
 
+      - name: Capture Test Reports
+        uses: actions/upload-artifact@v3
```
dimas-b, line 142: Are gradle scans not enough? I used to get test reports from there.

dimas-b (approved):

.github/workflows/ci.yml
snazy: They usually are. But those do not include successful results from the same run.

The problem at hand was that some Quarkus ITs (the Quarkus 3 PR) failed and others did not, so I wanted to compare those. (Because of course: the issues were hardly reproducible locally)

PR #6146, opened 2023-02-20
Bump to Quarkus 3.2.0.Final

snazy:
Includes a workaround for Quarkus3/OTel/okio OOM in tests, see https://github.com/quarkusio/quarkus/issues/32238

Most of the changes are rather "boring", replacing `javax` with `jakarta`.

Bumps the Keycloak version used in ITs.

Updates also include OpenTelemetry related configuration updates (`quarkus.opentelementy.*` --> `quarkus.otel.*`), including the Helm chart.

PR #6719, opened 2023-04-30
Keycloak requires container-network

snazy:
Quarkus ITs require the use of a container network ID and have to rewrite URLs/hosts accordingly. Otherwise Quarkus will not be able to connect to the services.


dimas-b (approved):

adutra:
For my enlightenment, what were the symptoms of the issue that this PR solved? The tests were passing before afaict...

snazy:
> For my enlightenment, what were the symptoms of the issue that this PR solved? The tests were passing before afaict...

Yes, works w/ current Quarkus - however, it's rather a mystery why it works, because Quarkus runs ITs in Docker networks, which aren't accessible from/to outside that network.

This PR is a spin-off off #6146, where Quarkus ITs using Keycloak do not work w/o this change.

PR #6736, opened 2023-05-03
Reserve usage of BatchingPersist to dry-run mode only

adutra:
This commit changes the behavior of merges and transplants to only use BatchingPersist in dry-run mode.


snazy (approved):

dimas-b (commented):
+1

PR #6743, opened 2023-05-04
REST V2: Key ranges for entries + diff

### LINKED ISSUE #3957: ContentKey based filter in the Nessie Diff API
**Description**: The current diff API brings the differences in all content keys between the two perimeters. The request is to allows a filter that restricts the search operation for a specific ContentKey.
The DBAdapter implementation is already equipped t[o take KeyFilterPredicate parameter](https://github.com/projectnessie/nessie/blob/c55073e62a09e3924a745d5375d2d348abf1eb6c/versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java#L323). So, the ask to expose the same at upper layers (version store / APIs etc).

**Requested Changes in public API**: Extend diff API to support filter options on the content key
- Changes in REST API 
- Changes in Java Client API
- Changes in SQL extentions

**Expected Use Cases**: Skip compaction on unchanged tables

### LINKED ISSUE #5591: Add start/end key parameters to v2 API methods

snazy:
Builds upon the work in #5656.

However, the "get range of contents" has been removed, because it's tricky to add it to the current Nessie Java API and since it's probably not a very useful functionality. It would also require adding paging and min/max-keys to the `GetMultipleContentRequest`/`Response` types to leverage the existing paging infrastructure and/or add `ContentsParams` parameter object - neither works well with the existing code.

Adds the necessary `VersionStore` changes, which are pretty straight forward.

Fixes #3957
Fixes #5591

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/v2api/HttpGetDiff.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/v2api/HttpGetDiff.java
@@ -18,7 +18,9 @@ package org.projectnessie.client.http.v2api;
 import org.projectnessie.api.v2.params.DiffParams;
 import org.projectnessie.client.builder.BaseGetDiffBuilder;
 import org.projectnessie.client.http.HttpClient;
+import org.projectnessie.client.http.HttpRequest;
 import org.projectnessie.error.NessieNotFoundException;
+import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.DiffResponse;
 import org.projectnessie.model.Reference;
 
@@ -36,20 +38,38 @@ final class HttpGetDiff extends BaseGetDiffBuilder<DiffParams> {
         .fromRef(Reference.toPathString(fromRefName, fromHashOnRef))
         .toRef(Reference.toPathString(toRefName, toHashOnRef))
         .maxRecords(maxRecords)
+        .minKey(minKey)
+        .maxKey(maxKey)
+        .prefixKey(prefixKey)
+        .filter(filter)
+        .requestedKeys(keys)
         .build();
   }
 
   @Override
   public DiffResponse get(DiffParams params) throws NessieNotFoundException {
-    return client
-        .newRequest()
-        .path("trees/{from}/diff/{to}")
-        .resolveTemplate("from", params.getFromRef())
-        .resolveTemplate("to", params.getToRef())
-        .queryParam("max-records", params.maxRecords())
-        .queryParam("page-token", params.pageToken())
-        .unwrap(NessieNotFoundException.class)
-        .get()
-        .readEntity(DiffResponse.class);
+    HttpRequest req =
+        client
+            .newRequest()
+            .path("trees/{from}/diff/{to}")
+            .resolveTemplate("from", params.getFromRef())
+            .resolveTemplate("to", params.getToRef())
+            .queryParam("max-records", params.maxRecords())
+            .queryParam("page-token", params.pageToken())
+            .queryParam("filter", params.getFilter());
+    params.getRequestedKeys().forEach(k -> req.queryParam("key", k.toPathString()));
+    ContentKey k = params.minKey();
+    if (k != null) {
+      req.queryParam("min-key", k.toPathString());
+    }
+    k = params.maxKey();
+    if (k != null) {
+      req.queryParam("max-key", k.toPathString());
+    }
+    k = params.prefixKey();
+    if (k != null) {
+      req.queryParam("prefix-key", k.toPathString());
+    }
+    return req.unwrap(NessieNotFoundException.class).get().readEntity(DiffResponse.class);
   }
 }
--- a/api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
+++ b/api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
@@ -112,11 +112,36 @@ public interface ApiDoc {
           + "'type' parameter may be used to ensure the operation is performed on the same object that the user "
           + "expects.\n";
 
-  String KEY_PARAMETER_DESCRIPTION =
-      "The key to a content object.\n"
-          + "\n"
-          + "Key components (namespaces) are separated by the dot ('.') character. Dot ('.') characters that are not "
+  String KEY_ELEMENTS_DESCRIPTION =
+      "Key components (namespaces) are separated by the dot ('.') character. Dot ('.') characters that are not "
           + "Nessie namespace separators must be encoded as the 'group separator' ASCII character (0x1D).\n";
+
+  String KEY_PARAMETER_DESCRIPTION = "The key to a content object.\n\n" + KEY_ELEMENTS_DESCRIPTION;
+
+  String REQUESTED_KEY_PARAMETER_DESCRIPTION =
+      "Restrict the result to one or more keys.\n\n"
+          + "Can be combined with min/max-key and prefix-key parameters, however both predicates must match. "
+          + "This means that keys specified via this parameter that do not match a given min/max-key or prefix-key will not be returned.\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_MIN_PARAMETER_DESCRIPTION =
+      "The lower bound of the content key range to retrieve (inclusive). "
+          + "The content keys of all returned entries will be greater than or equal to the min-value. "
+          + "Content-keys are compared as a 'whole', unlike prefix-keys.\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_MAX_PARAMETER_DESCRIPTION =
+      "The upper bound of the content key range to retrieve (inclusive). "
+          + "The content keys of all returned entries will be less than or equal to the max-value. "
+          + "Content-keys are compared as a 'whole', unlike prefix-keys.\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_PREFIX_PARAMETER_DESCRIPTION =
+      "The content key prefix to retrieve (inclusive). "
+          + "A content key matches a given prefix, a content key's elements starts with all elements of the prefix-key. "
+          + "Key prefixes exactly match key-element boundaries.\n\n"
+          + "Must not be combined with min/max-key parameters.\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
 
   String DEFAULT_KEY_MERGE_MODE_DESCRIPTION =
       "The default merge mode. If not set, `NORMAL` is assumed.\n"
--- a/api/model/src/main/java/org/projectnessie/api/v2/params/KeyRangeParams.java
+++ b/api/model/src/main/java/org/projectnessie/api/v2/params/KeyRangeParams.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.api.v2.params;
+
+import static org.projectnessie.api.v2.doc.ApiDoc.KEY_MAX_PARAMETER_DESCRIPTION;
+import static org.projectnessie.api.v2.doc.ApiDoc.KEY_MIN_PARAMETER_DESCRIPTION;
+import static org.projectnessie.api.v2.doc.ApiDoc.KEY_PREFIX_PARAMETER_DESCRIPTION;
+
+import javax.annotation.Nullable;
+import javax.ws.rs.QueryParam;
+import org.eclipse.microprofile.openapi.annotations.media.ExampleObject;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.projectnessie.model.ContentKey;
+
+public abstract class KeyRangeParams<IMPL extends KeyRangeParams<IMPL>>
+    extends AbstractParams<IMPL> {
+
+  @Parameter(
+      description = KEY_MIN_PARAMETER_DESCRIPTION,
+      examples = @ExampleObject(ref = "ContentKeyGet"))
+  @QueryParam("min-key")
+  @Nullable
+  @jakarta.annotation.Nullable
+  private ContentKey minKey;
+
+  @Parameter(
+      description = KEY_MAX_PARAMETER_DESCRIPTION,
+      examples = @ExampleObject(ref = "ContentKeyGet"))
+  @QueryParam("max-key")
+  @Nullable
+  @jakarta.annotation.Nullable
+  private ContentKey maxKey;
+
+  @Parameter(
+      description = KEY_PREFIX_PARAMETER_DESCRIPTION,
+      examples = @ExampleObject(ref = "ContentKeyGet"))
+  @QueryParam("prefix-key")
+  @Nullable
+  @jakarta.annotation.Nullable
+  private ContentKey prefixKey;
+
+  protected KeyRangeParams() {}
+
+  public KeyRangeParams(
+      @Nullable @jakarta.annotation.Nullable Integer maxRecords,
+      @Nullable @jakarta.annotation.Nullable String pageToken,
+      @Nullable @jakarta.annotation.Nullable ContentKey minKey,
+      @Nullable @jakarta.annotation.Nullable ContentKey maxKey,
+      @Nullable @jakarta.annotation.Nullable ContentKey prefixKey) {
+    super(maxRecords, pageToken);
+    this.minKey = minKey;
+    this.maxKey = maxKey;
+    this.prefixKey = prefixKey;
+  }
+
+  @Nullable
+  @jakarta.annotation.Nullable
+  public ContentKey minKey() {
+    return minKey;
+  }
+
+  @Nullable
+  @jakarta.annotation.Nullable
+  public ContentKey maxKey() {
+    return maxKey;
+  }
+
+  @Nullable
+  @jakarta.annotation.Nullable
+  public ContentKey prefixKey() {
+    return prefixKey;
+  }
+}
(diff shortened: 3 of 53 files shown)


api/model/src/main/java/org/projectnessie/api/v2/params/KeyRangeParams.java
```diff
@@ -0,0 +1,69 @@
...
+  @jakarta.annotation.Nullable
+  private ContentKey minKey;
+
+  @Parameter(
+      description = KEY_MAX_PARAMETER_DESCRIPTION,
+      examples = @ExampleObject(ref = "ContentKeyGet"))
+  @QueryParam("max-key")
+  @Nullable
+  @jakarta.annotation.Nullable
+  private ContentKey maxKey;
```
dimas-b, line 44: Just wondering: Would it make sense to have a separate `key-prefix` query parameter? We could validate it to actually be the prefix of both min and max keys if all are set.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -494,8 +496,8 @@ public PaginationIterator<KeyEntry> getKeys(Ref ref, String pagingToken, boolean
     StoreKey begin =
         pagingToken != null && !pagingToken.isEmpty()
             ? keyFromString(fromString(pagingToken).token().toStringUtf8())
-            : null;
-    StoreKey end = null;
+            : (minKey != null ? keyToStoreKey(minKey) : null);
+    StoreKey end = maxKey != null ? keyToStoreKey(maxKey) : null;
```
dimas-b, line 500: Will this work when paginating over a "prefix" query? I mean that on the second page min and max keys will no longer be equals when passed to `index.iterator(...)` :thinking:

api/client/src/main/java/org/projectnessie/client/http/v2api/HttpGetDiff.java
```diff
@@ -36,20 +38,33 @@ protected DiffParams params() {
...
+            .path("trees/{from}/diff/{to}")
+            .resolveTemplate("from", params.getFromRef())
+            .resolveTemplate("to", params.getToRef())
+            .queryParam("max-records", params.maxRecords())
+            .queryParam("page-token", params.pageToken())
+            .queryParam("filter", params.getFilter());
+    keys.forEach(k -> req.queryParam("key", k.toPathString()));
+    ContentKey k = params.minKey();
+    if (k != null) {
+      req.queryParam("min-key", k.toPathString());
```
dimas-b, line 62: Should we have this in `HttpGetEntries` too?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
snazy: The max key will still be the same, and "min" would be "replaced" with the key from the paging token.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -109,12 +109,20 @@ public interface ApiDoc {
...
           + "Nessie namespace separators must be encoded as the 'group separator' ASCII character (0x1D).\n";
 
+  String KEY_PARAMETER_DESCRIPTION = "The key to a content object.\n\n" + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_MIN_PARAMETER_DESCRIPTION =
+      "[Future] The lower bound of the content key range to retrieve (inclusive).\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_MAX_PARAMETER_DESCRIPTION =
+      "[Future] The upper bound of the content key range to retrieve (exclusive).\n\n"
```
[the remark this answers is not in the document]
snazy: Since `maxKey` is inclusive, that wouldn't work.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
dimas-b: that's exactly what I mean. If we start with min == max for a prefix query, min != max in the second page... Just wondering if we want this to work for prefix queries as well as for normal ranges..

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestKeyRanges.java
```diff
@@ -0,0 +1,117 @@
...
+  public void max() {
+    ContentKey bar = ContentKey.of("bar");
+    KeyRanges keyRange = keyRanges(null, null, bar, null);
+
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKey(bar));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKeyVariant(bar, "Z"));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKeyVariant(bar, "A"));
+
+    soft.assertThat(keyRange.endStoreKey())
+        .isLessThanOrEqualTo(keyToStoreKey(ContentKey.of("bar", "baz")));
```
dimas-b, line 76: Why not simply `isLessThan`? We know it is not equal.

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -523,9 +524,18 @@ private BiConsumer<ImmutableCommit.Builder, CommitLogEntry> enhancerForCommitLog
...
+      boolean withContent,
+      ContentKey minKey,
+      ContentKey maxKey,
+      ContentKey prefixKey,
+      Predicate<ContentKey> contentKeyPredicate)
       throws ReferenceNotFoundException {
-    checkArgument(pagingToken == null, "Paging not supported");
+    checkArgument(pagingToken == null, "Paging not supported by the storage model in use");
+    checkArgument(
+        minKey == null && maxKey == null, "Key ranges not supported by the storage model in use");
```
dimas-b, line 538: I guess we ought to check `prefixKey` too (also in `getDiffs`).

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestKeyRanges.java
```diff
@@ -0,0 +1,117 @@
...
+    soft.assertThat(keyRange.beginStoreKey()).isEqualTo(keyToStoreKeyPrefixMin(bar));
+
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKey(bar));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKeyVariant(bar, "Z"));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKeyVariant(bar, "A"));
+
+    soft.assertThat(keyRange.endStoreKey())
+        .isGreaterThan(keyToStoreKey(ContentKey.of("bar", "baz")));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKey(ContentKey.of("bar", "0")));
+    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKey(ContentKey.of("barbaz")));
```
dimas-b, line 108: This fails, though:
```
    soft.assertThat(keyRange.endStoreKey()).isGreaterThan(keyToStoreKey(ContentKey.of("bar\u03C0")));
```

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
dimas-b: Sorry, my comment is not relevant to the latest code - resolving.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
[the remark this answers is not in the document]
dimas-b: Yes. Having a separate `prefixKey` parameter is preferable :+1:

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestKeyRanges.java
snazy: Yes - why should it not? It's `> "bar"`

dimas-b: But it has the right prefix.... why `barbaz` passes by `bar\u03C0` fails?

snazy: It's a negative `byte` value in that UTF char - and this one just ends in a string comparison.

dimas-b: but we support Unicode at the `ContentKey` level, right?.. and as a `ContentKey` is has the right prefix, so I guess it should be included in the range :thinking:

dimas-b: To put it another way, should we use `Character.MAX_VALUE` instead of 255 in `TypeMapping`?

snazy: Actually, it belongs to this PR. Just sad that this change requires such a change :(

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreKey.java
```diff
@@ -39,24 +41,54 @@ public final class StoreKey implements Comparable<StoreKey> {
...
+    }
+    if (diff >= key.length) {
+      if (prefixEnd && diff == key.length) {
+        return 1;
+      }
+      return -1;
+    }
+    if (diff >= that.key.length) {
+      return 1;
+    }
```
dimas-b, line 83: This makes comparisons non-commutative. It will bite us sooner or later, I think.
```
  @Test
  void comparePrefix() {
    StoreKey k1 = StoreKey.keyFromStringPrefixEnd("aaa");
    StoreKey k2 = StoreKey.keyFromStringPrefixEnd("aaaBBB");
    soft.assertThat(k1).isGreaterThan(k2);
    soft.assertThat(k2).isLessThan(k1); // this assert fails
  }
```

dimas-b: I think it might be preferable to actually encode the "prefix" flag in the key's byte form (to ensure consistency in case it ever gets into storage).

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreKey.java
```diff
@@ -39,24 +41,54 @@ public final class StoreKey implements Comparable<StoreKey> {
...
+    byte my = key[diff];
+    byte other = that.key[diff];
+    return Byte.compareUnsigned(my, other);
```
dimas-b, line 86: Is this always consistent with char-wise comparison?.. i.e. if we have two (unicode) strings, can we be sure that `StoreKeys` derived from those strings will compare the same way as the Strings?... more specifically - same as `ContentKey` objects?

snazy:
Reverted the `byte[]` change in `StoreKey` and updated `StoreKey.compareTo` to respect the `0`/`1` separated elements to not conflict with (unicode) chars.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
```diff
@@ -479,7 +479,23 @@ private ByteBuffer serializeKey(
...
   static <V> StoreIndex<V> deserializeStoreIndex(ByteBuffer serialized, ElementSerializer<V> ser) {
     byte version = serialized.get();
-    checkArgument(version == 1, "Unsupported serialized representation of KeyIndexSegment");
+    boolean needSort = false;
+    switch (version) {
+      case 1:
+        // In PR #6743, introducing min/max/prefix-key arguments for "get entries" and "get diff",
+        // StoreKey.compareTo() was updated, which might lead to a different order in the elements
+        // list. The "current" version has been bumped, version=1 serialized indexes are sorted
+        // after they have been loaded.
```
dimas-b, line 488: Do you have an example where the sort order actually changes? I could not figure one out mentally :)

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/KeyRanges.java
```diff
@@ -0,0 +1,88 @@
...
+  @Value.NonAttribute
+  PagingToken pagingTokenObj() {
+    String paging = pagingToken();
+    return paging != null ? fromString(paging) : null;
+  }
+
+  @Value.NonAttribute
+  StoreKey endStoreKey() {
+    ContentKey max = maxKey();
+    return max != null ? keyToStoreKeyMax(max) : null;
```
dimas-b, line 77: Why not derive a max key from `prefix` when it is set?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -491,15 +502,15 @@ public PaginationIterator<KeyEntry> getKeys(Ref ref, String pagingToken, boolean
...
 
-    StoreKey begin =
-        pagingToken != null && !pagingToken.isEmpty()
-            ? keyFromString(fromString(pagingToken).token().toStringUtf8())
-            : null;
-    StoreKey end = null;
-
-    Iterator<StoreIndexElement<CommitOp>> result = index.iterator(begin, end, false);
+    Iterator<StoreIndexElement<CommitOp>> result =
+        index.iterator(keyRanges.beginStoreKey(), keyRanges.endStoreKey(), false);
```
dimas-b, line 506: If we do not derive an end key from `prefix`, then prefix queries will have to scan all the way to the end of the index, will they not?

dimas-b (commented):
Mostly optimization concerns

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreKey.java
```diff
@@ -51,7 +51,70 @@ public String rawString() {
...
+    while (true) {
+      boolean blkEof1 = blkIdx1 >= blkLen1;
+      boolean blkEof2 = blkIdx2 >= blkLen2;
+      if (blkEof1) {
+        return blkEof2 ? 0 : -1;
+      } else if (blkEof2) {
+        return 1;
+      }
+
+      blkIdx1 = nextBlock(key, (char) 0, blkIdx1, blkLen1, blk1);
```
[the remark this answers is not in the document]
snazy: Um, totally right. Fixed.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/indexes/StoreIndexImpl.java
snazy: ouch - again

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/KeyRanges.java
snazy: That's not possible, sadly.
It would need something to compare to - but what?
"Incrementing the last char" doesn't really work there.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
snazy: Fixed

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -748,16 +785,35 @@ public PaginationIterator<Diff> getDiffs(Ref from, Ref to, String pagingToken)
...
+            ? d -> {
+              ContentKey key = storeKeyToKey(d.key());
+              return key != null && contentKeyPredicate.test(key);
+            }
+            : x -> true;
+
+    Predicate<DiffEntry> stopPredicate;
+    if (prefixKey != null) {
+      StoreKey prefix = keyToStoreKeyNoVariant(prefixKey);
+      stopPredicate = d -> !d.key().startsWithElementsOrParts(prefix);
```
dimas-b, line 812: `startsWithElementsOrParts` seems to contradict the doc text in `ApiDoc`... which way do we want it to work? :)

dimas-b (commented):
LGTM overall, but prefix handling seems to contradict (older) docs

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -109,12 +109,37 @@ public interface ApiDoc {
...
+  String KEY_MAX_PARAMETER_DESCRIPTION =
+      "The upper bound of the content key range to retrieve (inclusive). "
+          + "The content keys of all returned entries will be less than or equal to the max-value. "
+          + "Content-keys are compared as a 'whole', unlike prefix-keys.\n\n"
+          + KEY_ELEMENTS_DESCRIPTION;
+
+  String KEY_PREFIX_PARAMETER_DESCRIPTION =
+      "The content key prefix to retrieve (inclusive). "
+          + "A content key matches a given prefix, a content key's elements starts with all elements of the prefix-key. "
+          + "It is not restricted to key-element boundaries.\n\n"
```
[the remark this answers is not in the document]
snazy: fixed

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
snazy: doc's wrong, fixed

dimas-b (dismissed):

dimas-b (approved):

dimas-b (approved):

PR #6746, opened 2023-05-04
GC: Trim long failure messages in JDBC repo

snazy:
And bump SQL-DDL for metadata location pointer.


gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
```diff
@@ -44,7 +48,7 @@ private SqlDmlDdl() {}
           + "    commit_id VARCHAR(100), \n"
           + "    content_key VARCHAR(500), \n"
           + "    content_type VARCHAR(40), \n"
-          + "    metadata_location VARCHAR(500), \n"
+          + "    metadata_location VARCHAR(1000), \n"
```
adutra, line 51: Why this change?

adutra (approved):

PR #6747, opened 2023-05-04
Cleanup V2 MergeResponse

### LINKED ISSUE #5689: Refactor `MergeResponse` in API v2
Currently `MergeResponse` is too verbose but at the same time not informative enough.
The original idea behind e.g. `MergeResponse.getDetails()` was to give callers information especially about why a merge operation did not work - the "why" is quite binary at the moment: `ContentKeyConflict.NONE` or `ContentKeyConflict.UNRESOLVABLE` and far from being informative. The fields containing full commit objects and commit IDs have already been deprecated (and are already unused in v2) for the same reason (too verbose, but not informative).

I wonder whether we should either:
* deprecate and hide `MergeResponse.getDetails()` in API v2
* have a new, cleaned up `MergeResponse2` for API v2

snazy:
Ensure that all non-V2 attributes in `MergeResponse` (those that are present for V1, but deprecated) are properly annotated with `@JsonView(Views.V1.class)`.

On top of the already deprecated attributes,
`MergeResponse.ContentKeyDetails.getConflictType()` is deprecated and V1-only now as well. It already has a better replacement with `MergeResponse.getConflictType.getConflict()`. The corresponding enum is also deprecated now.

Fixes #5689

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/MergeResponse.java
+++ b/api/model/src/main/java/org/projectnessie/model/MergeResponse.java
@@ -73,12 +73,14 @@ public interface MergeResponse {
 
   @Deprecated // for removal and replaced with something else
   @Schema(deprecated = true, hidden = true)
+  @JsonView(Views.V1.class)
   List<LogEntry> getSourceCommits();
 
   @Nullable
   @jakarta.annotation.Nullable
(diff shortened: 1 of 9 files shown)


dimas-b (approved):
LGTM :+1:

PR #6758, opened 2023-05-05
New storage: allow deletion of multiple repositories

snazy:
This change only moves the logic from the `Persist.erase()` implementations to their `Backend.eraseRepositories(Set)` counterparts and makes the implementations delete multiple repos in one shot. No other changes, just moving some already existing code around.

### DIFF
--- a/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractBackendRepositoryTests.java
+++ b/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractBackendRepositoryTests.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (C) 2022 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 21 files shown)


versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractBackendRepositoryTests.java
```diff
@@ -0,0 +1,89 @@
...
+    soft.assertThat(repositoryLogic.repositoryExists()).isTrue();
+    repo1.erase();
+    soft.assertThat(repositoryLogic.repositoryExists()).isFalse();
+    try (CloseableIterator<Obj> scan = repo1.scanAllObjects(EnumSet.allOf(ObjType.class))) {
+      soft.assertThat(scan).isExhausted();
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(ints = {1, 3, 10})
```
dimas-b, line 66: Shouldn't we go above `MAX_CONCURRENT_DELETES`?

dimas-b (approved):

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractBackendRepositoryTests.java
snazy: oops - saw it too late - #6768

PR #6760, opened 2023-05-05
Events notification system for Nessie - Service module

adutra:
Next PR for the events notification system: the service layer.


events/service/src/main/java/org/projectnessie/events/service/util/ReferenceMapping.java
```diff
@@ -0,0 +1,49 @@
...
+    if (ref instanceof BranchName) return Reference.BRANCH;
+    if (ref instanceof TagName) return Reference.TAG;
+    if (ref instanceof DetachedRef) return "DETACHED";
+    else return "UNKNOWN";
+  }
+
+  private static String getFullReferenceName(NamedRef ref) {
+    if (ref instanceof BranchName) return "refs/heads/" + ref.getName();
+    if (ref instanceof TagName) return "refs/tags/" + ref.getName();
+    else return "refs/???/" + ref.getName();
```
snazy, line 47: Can (actually should) throw here.
If `ref` is an instanceof `DetachedRef`, there's no named reference - the result should be `null` then

events/service/src/main/java/org/projectnessie/events/service/EventService.java
```diff
@@ -0,0 +1,516 @@
...
+
+    protected void deliver(Event event) {
+      try {
+        MDC.put(SUBSCRIPTION_ID_MDC_KEY, subscriptionId);
+        if (subscriber.accepts(event)) {
+          LOGGER.debug("Received event: {}", event);
+          CTX context = currentContext();
+          Duration initialDelay = config.getRetryConfig().getInitialDelay();
+          if (subscriber.isBlocking()) {
+            deliverAsync(event, 1, initialDelay, null, context);
```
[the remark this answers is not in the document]
adutra: But if we implement it later, would it be implemented here, or somewhere else? Trying to figure out if it's best to implement it right now, or decide later if we want this or not.

adutra:
Switching to draft until we get a clear picture around retries and async delivery.

adutra:
Also I was about to push a simplification of `EventExecutor`: it's unnecessarily bloated with futures, etc.

events/service/src/main/java/org/projectnessie/events/service/EventService.java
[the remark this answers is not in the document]
adutra: OK I've given this a bit more thought, and I agree to remove the executor completely and the retry mechanism. This will make this service impl completely synchronous by default. But that's enough for this module. In the quarkus-specific module I'd keep the retry logic if you don't mind, but maybe with a default of 0 retries for safety. How does that sound to you?

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
```diff
@@ -0,0 +1,221 @@
...
+  public EventFactory(EventConfig config) {
+    this.config = config;
+  }
+
+  protected Event newCommitEvent(
+      Commit commit, BranchName targetBranch, String repositoryId, @Nullable Principal user) {
+    org.projectnessie.model.CommitMeta commitMeta = commit.getCommitMeta();
+    ImmutableCommitEvent.Builder builder =
+        ImmutableCommitEvent.builder()
+            .id(config.getIdGenerator().get())
```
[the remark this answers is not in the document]
adutra: I tried that initially, and while it makes our lives slightly easier when writing tests, I don't think these builders will be useful for end users, since they won't need to create event instances manually. That's why I decided to not include the builders.

events/service/src/main/java/org/projectnessie/events/service/util/ReferenceMapping.java
adutra: Ref full name is mandatory currently, should I change that to be optional?

adutra:
Ready for review again.

events/service/src/main/java/org/projectnessie/events/service/EventSubscribers.java
```diff
@@ -0,0 +1,154 @@
...
+
+  public void start(Function<EventSubscriber, EventSubscription> subscriptionFactory) {
+    if (!started) {
+      LOGGER.info("Starting subscribers...");
+      for (EventSubscriber subscriber : subscribers) {
+        try {
+          EventSubscription subscription = subscriptionFactory.apply(subscriber);
+          subscriber.onSubscribe(subscription);
+        } catch (Exception e) {
+          LOGGER.error("Error starting subscriber", e);
```
snazy, line 94:
```suggestion
          throw new RuntimeException("Error starting subscriber", e);
```
?

events/service/src/main/java/org/projectnessie/events/service/EventSubscribers.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+      }
+      LOGGER.info("Done starting subscribers.");
+      started = true;
+    }
+  }
+
+  @Override
+  public void close() {
```
snazy, line 103: Suggestion:

```java
  @Override
  public void close() {
    if (!closed) {
      LOGGER.info("Closing subscribers...");
      RuntimeException error = null;
      try {
        for (EventSubscriber subscriber : subscribers) {
          try {
            subscriber.close();
          } catch (Exception e) {
            if (error == null) {
              error = new RuntimeException("Error closing subscribers", e);
            } else {
              error.addSuppressed(e);
            }
          }
        }
      } finally{
        LOGGER.info("Done closing subscribers.");
        closed = true;
        if (error != null) {
          throw error;
        }
      }
    }
  }
```

events/service/src/main/java/org/projectnessie/events/service/EventService.java
```diff
@@ -0,0 +1,516 @@
...
+    this.config = config;
+    this.factory = factory;
+    this.subscribers = subscribers;
+    this.executor = executor;
+  }
+
+  /** Starts event delivery by activating the subscriber handlers. */
+  public synchronized void start() {
+    if (!started) {
+      LOGGER.info("Starting subscribers...");
```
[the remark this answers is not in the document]
snazy: Thinking a bit more about `EventSubscriber`'s `default void onSubscribe(EventSubscription subscription) {}`, I think your initial version in the other PR where `onSubscribe` is not a default-method was actually better.
WDYT if we require that `EventSubscriber`s must not initialize in the constructor, but in the `onSubscribe` implementation - and make it a non-default method so it has to be implemented?

events/service/src/main/java/org/projectnessie/events/service/EventService.java
```diff
@@ -0,0 +1,303 @@
...
+public class EventService implements AutoCloseable {
+
+  public static final String SUBSCRIPTION_ID_MDC_KEY = "nessie.events.subscription.id";
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(EventService.class);
+
+  protected final EventConfig config;
+  protected final EventFactory factory;
+  protected final EventSubscribers subscribers;
+  private final Map<EventSubscription, EventSubscriber> subscriptions = new ConcurrentHashMap<>();
```
snazy, line 74: Not sure that we need a CHM here - it might only be necessary during startup, but not during runtime.

Should we build a non-concurrent map during `start()` and just initialize this field when done?

events/service/src/main/java/org/projectnessie/events/service/EventService.java
```diff
@@ -0,0 +1,303 @@
...
+   *     ReferenceCreatedEvent}s, then it should only be invoked for {@link
+   *     ReferenceCreatedResult}s. Similarly, if this service is only interested in {@link
+   *     ContentStoredEvent}s, then it should be invoked for any result that might produce such
+   *     events, namely: {@link CommitResult} and {@link MergeResult}.
+   * @see ResultCollector#accept(Result)
+   * @see ResultCollector#shouldProcess(Result)
+   */
+  @SuppressWarnings("unchecked")
+  public void onVersionStoreEvent(VersionStoreEvent event) {
+    Result result = event.getResult();
```
snazy, line 118: WDYT about a `if (!started) return` guard here?

events/service/src/main/java/org/projectnessie/events/service/EventService.java
snazy, line 74: The map's key can probably be just the subscription ID. The config part is not used later on (after the onSubscribe).

events/service/src/main/java/org/projectnessie/events/service/EventService.java
```diff
@@ -0,0 +1,303 @@
...
+          ContentKey contentKey = ContentMapping.map(operation.getKey());
+          fireEvent(
+              factory.newContentRemovedEvent(targetBranch, hash, contentKey, repositoryId, user));
+        }
+      }
+    }
+  }
+
+  private boolean hasContentSubscribers() {
+    return subscribers.hasSubscribersFor(EventType.CONTENT_STORED)
```
snazy, line 231: Guess this can one and the function below can be memoized in `start()`.
Or at least have it end in `EnumSet.containsAll` which ends in a single bit-mask-check.

events/service/src/main/java/org/projectnessie/events/service/EventService.java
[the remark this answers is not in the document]
snazy: Yea - let's see how that looks in Quarkus

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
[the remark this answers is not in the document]
snazy: Okay.

events/service/src/main/java/org/projectnessie/events/service/util/ReferenceMapping.java
snazy: Optional, yes

events/service/src/main/java/org/projectnessie/events/service/EventSubscribers.java
adutra: Not sure... this means that the next subscriber won't be started either.

adutra: Well, I guess it's safer to prevent the application from starting, rather than having a half-initialized list of subscribers. So OK to do as you suggest.

events/service/src/main/java/org/projectnessie/events/service/EventService.java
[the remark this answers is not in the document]
adutra: Yes good idea, I will amend the SPI module.

events/service/src/main/java/org/projectnessie/events/service/EventConfig.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.UUID;
+import java.util.function.Supplier;
+import org.projectnessie.events.api.Event;
+import org.projectnessie.events.spi.EventSystemConfiguration;
+import org.projectnessie.events.spi.ImmutableEventSystemConfiguration;
+import org.projectnessie.model.NessieConfiguration;
+
+public interface EventConfig {
```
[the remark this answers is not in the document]
adutra: That would make it impossible to extend this interface by a `QuarkusEventConfig` for instance. More or less the same reason why `ServerConfig` is not immutable I guess.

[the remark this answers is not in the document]
snazy: Ah, that's true

events/service/src/main/java/org/projectnessie/events/service/util/ContentMapping.java
```diff
@@ -0,0 +1,101 @@
...
+      IcebergView view = (IcebergView) content;
+      return ImmutableIcebergView.builder()
+          .id(Objects.requireNonNull(content.getId()))
+          .metadataLocation(view.getMetadataLocation())
+          .versionId(view.getVersionId())
+          .schemaId(view.getSchemaId())
+          .sqlText(view.getSqlText())
+          .dialect(view.getDialect())
+          .build();
+    } else if (type == org.projectnessie.model.Content.Type.DELTA_LAKE_TABLE) {
```
snazy, line 70: Sorry, completely forgot about it ... can you add `UTF` here as well?

Actually, I wonder whether we should mirror the model types 1:1 at all in events, and rather handle them all as "generic". Would save us a bunch of code. Downstream consumer could potentially directly deserialize to the model-type. WDYT? Could be done in a follow-up though.

snazy (approved):

PR #6761, opened 2023-05-05
Add new content type UDF

snazy:


api/model/src/main/java/org/projectnessie/model/UDF.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @Deprecated
+  @Nullable
+  @jakarta.annotation.Nullable
+  @JsonInclude(Include.NON_NULL)
+  @JsonView(Views.V1.class)
+  // Left here in case an old Nessie client sends this piece of information.
+  // To be removed when API v1 gets removed.
+  public abstract Map<String, Object> getMetadata();
```
dimas-b, line 59: Not sure I understand how this relates to API versions :thinking: If the client uses new jars, won't it be able to sent normal `UDF` objects over API v1 _and_ v2 (whatever it chooses)?

If the client uses old jars, it will not see this method anyway.

servers/store/src/main/java/org/projectnessie/server/store/UDFSerializer.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public UDF applyId(UDF content, String id) {
+    return ((ImmutableUDF) content).withId(id);
+  }
+
+  @Override
+  public boolean requiresGlobalState(ByteString content) {
+    ObjectTypes.Content parsed = parse(content);
+    return !parsed.getIcebergViewState().hasMetadataLocation();
```
dimas-b, line 53: Should it be a constant `false` here?

dimas-b (approved):

PR #6796, opened 2023-05-11
Some errorprone mitigations (no functional change)

snazy:


gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/EnvironmentDefaultProvider.java
```diff
@@ -38,7 +39,7 @@ public String defaultValue(ArgSpec argSpec) {
       OptionSpec optionSpec = (OptionSpec) argSpec;
       return Arrays.stream(optionSpec.names())
           .filter(name -> name.startsWith("--"))
-          .map(name -> "NESSIE_GC_" + name.substring(2).replace('-', '_').toUpperCase())
+          .map(name -> "NESSIE_GC_" + name.substring(2).replace('-', '_').toUpperCase(Locale.ROOT))
```
XN137, line 42: should we set the new upper/lowercase locale check to ERROR ?

gc/gc-repository-jdbc/src/test/java/org/projectnessie/gc/contents/jdbc/TestJdbcPersistenceUnit.java
```diff
@@ -24,6 +24,7 @@
 
 public class TestJdbcPersistenceUnit {
   @Test
+  @SuppressWarnings("InlineMeInliner")
```
XN137, line 27: if this code is considered java11 then we should use the new api here instead probably

XN137 (approved):

gc/gc-repository-jdbc/src/test/java/org/projectnessie/gc/contents/jdbc/TestJdbcPersistenceUnit.java
snazy: Yea, but it's just a test

gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/EnvironmentDefaultProvider.java
snazy: Mind opening a PR?

PR #6800, opened 2023-05-11
increase severity of multiple errorprone checks

XN137:
note that we fix DifferentNameButSame with a minimal diff, but
not make it uniform across all files

### DIFF
--- a/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
+++ b/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
@@ -61,10 +61,8 @@ import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.IcebergTable;
 import org.projectnessie.model.LogResponse;
-import org.projectnessie.model.LogResponse.LogEntry;
 import org.projectnessie.model.Operation.Delete;
 import org.projectnessie.model.Operation.Put;
-import org.projectnessie.model.RefLogResponse;
 import org.projectnessie.model.RefLogResponse.RefLogResponseEntry;
 import org.projectnessie.model.Reference;
 import org.projectnessie.tools.compatibility.api.NessieAPI;
@@ -132,8 +130,8 @@ public class ITUpgradePath {
   }
 
   @SuppressWarnings("deprecation")
-  Stream<RefLogResponse.RefLogResponseEntry> refLog(
-      Function<GetRefLogBuilder, GetRefLogBuilder> configurer) throws NessieNotFoundException {
+  Stream<RefLogResponseEntry> refLog(Function<GetRefLogBuilder, GetRefLogBuilder> configurer)
+      throws NessieNotFoundException {
     if (version.isGreaterThan(Version.parseVersion("0.30.0"))) {
       return configurer.apply(api.getRefLog()).stream();
     } else {
@@ -221,7 +219,7 @@ public class ITUpgradePath {
               String commitMessage = "hello world " + versionFromRef;
               assertThat(commitLog)
                   .hasSize(1)
-                  .map(LogEntry::getCommitMeta)
+                  .map(LogResponse.LogEntry::getCommitMeta)
                   .map(CommitMeta::getMessage)
                   .containsExactly(commitMessage);
             })
--- a/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
+++ b/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
@@ -60,7 +60,7 @@ import org.projectnessie.model.Content;
 import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.EntriesResponse;
 import org.projectnessie.model.IcebergTable;
-import org.projectnessie.model.LogResponse;
+import org.projectnessie.model.LogResponse.LogEntry;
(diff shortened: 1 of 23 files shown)


snazy:
Running `./gradlew compileAll` locally could be more efficient than force-pushing?

compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/ITUpgradePath.java
```diff
@@ -221,7 +219,7 @@ void commitLog() throws NessieNotFoundException {
               String commitMessage = "hello world " + versionFromRef;
               assertThat(commitLog)
                   .hasSize(1)
-                  .map(LogEntry::getCommitMeta)
+                  .map(LogResponse.LogEntry::getCommitMeta)
```
dimas-b, line 222: why not keep the import for `LogEntry`?

gc/gc-iceberg-mock/src/main/java/org/projectnessie/gc/iceberg/mocks/MockManifestEntry.java
```diff
@@ -51,7 +50,7 @@ public Status status() {
 
   public abstract String filePath();
 
-  public abstract StructType partitionType();
+  public abstract Types.StructType partitionType();
```
dimas-b, line 53: Why do we sometimes import the inner class name, and sometime qualify it with its parent class name?

In this particular case, the inner class name seems to be non-ambiguous, so why do we change it to be qualified by the parent type?

XN137: as stated in the PR description we resolve `DifferentNameButSame` violations by doing the minimal amount of changes... so this file uses more `Types.StructType` than `StructType` right now, we we prefer to use the former for the fix.

if we can come up with a uniform convention on how to handle these things, we can also apply that kind of fix.

imo the first step is making things uniform on a per file basis, wdyt?

dimas-b: IMHO, shorter type references (i.e. importing the most specific name possible)  are preferable over smaller PR diff.

I'm open to other opinions too.

integrations/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/SparkSqlTestBase.java
```diff
@@ -299,14 +298,13 @@ protected List<SparkCommitLogEntry> commitAndReturnLog(String branch, String ini
         content != null
             ? ImmutableOperations.builder()
                 .addOperations(
-                    Operation.Put.of(
-                        key, IcebergTable.of("foo", 42, 42, 42, 42, content.getId()), content))
+                    Put.of(key, IcebergTable.of("foo", 42, 42, 42, 42, content.getId()), content))
```
XN137, line 301: is `Put` too short/common like `Builder` or not?

versioned/persist/serialize/src/main/java/org/projectnessie/versioned/persist/adapter/serialize/ProtoSerialization.java
```diff
@@ -34,14 +34,13 @@
 import org.projectnessie.versioned.persist.adapter.RefLog;
 import org.projectnessie.versioned.persist.adapter.RepoDescription;
 import org.projectnessie.versioned.persist.serialize.AdapterTypes;
```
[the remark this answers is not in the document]
dimas-b: SGTM

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -1440,7 +1439,8 @@ public void fetchEntriesByNamelessReference() throws BaseNessieClientServerExcep
             .operation(Put.of(b, tb))
             .commitMeta(CommitMeta.fromMessage("commit 1"))
             .commit();
-    List<Entry> entries = api().getEntries().hashOnRef(branch.getHash()).get().getEntries();
+    List<EntriesResponse.Entry> entries =
```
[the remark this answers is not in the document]
dimas-b: In the context of Nessie API (implied by this class), I think `Entry` has a specific meaning. I'd prefer keeping the import for it.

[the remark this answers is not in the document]
snazy: As mentioned, errorprone's `BadImport` rule applies.

versioned/persist/tx/src/main/java/org/projectnessie/versioned/persist/tx/TxDatabaseAdapter.java
```diff
@@ -1673,7 +1672,7 @@ private RefLogEntry writeRefLogEntry(
       NamedRef ref,
       RefLogHead refLogHead,
       Hash commitHash,
-      Operation operation,
+      RefLogEntry.Operation operation,
```
XN137, line 1675: should we prefer `Operation` in this file instead?

dimas-b: I think in this particular case using the longer name is justified to identify this as belonging to the deprecated ref log functionality.

dimas-b (approved):
LGTM, if @snazy is ok with these changes. Thanks for you effort, @XN137! :)

snazy (approved):

PR #6818, opened 2023-05-12
Add an API compatibility check to Nessie clients

### LINKED ISSUE #6805: GC Tool is mixing v1 and v2 URLs
**What happened**:

Since #5660 GC tool switched to V2 api by default, but it seems to be mixing V1 and V2 URLs:

```
GET /api/v1/trees/main%402e1cfa82035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be
8f10d/history?fetch=ALL HTTP/1.1" 405
```

The above v1 URL does not exist, but there is a matching v2 one. The request is simply trying to read the commitlog for the given reference, using v2 `getCommitLog` REST endpoint, but the URI prefix is wrong.

As a consequence the router is mistakenly matching the request against the v1 `assignReference` REST endpoint, which expects a PUT method, not GET – hence the 405 HTTP response (method not allowed).

Here is the DEBUG log showing that:

```
DEBUG [io.qua.mic.run.bin.RequestMetricInfo] (vert.x-eventloop-thread-1) Path /api/vi/trees/main%402e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d/history matched pattern /api/v1/trees/.*/ using /api/v1/trees/{referenceType}/{ref}
```

GC tool should use v2 URL prefixes by default.

I suggest that we:

1. Change the default value of `NessieOptions#nessieUri` to be a V2 URI prefix;
2. Add a warning if the value of this field does not end with `api/v2`.

I think this is happening by default if user uses the default URI, or passes a v1 URI using the `--uri` option.

adutra:
... and change default URI of GC tool to a v2 URI.

Fixes #6805.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
@@ -16,6 +16,7 @@
 package org.projectnessie.client.http;
 
 import static org.projectnessie.client.NessieConfigConstants.CONF_CONNECT_TIMEOUT;
+import static org.projectnessie.client.NessieConfigConstants.CONF_ENABLE_API_COMPATIBILITY_CHECK;
 import static org.projectnessie.client.NessieConfigConstants.CONF_FORCE_URL_CONNECTION_CLIENT;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_DISABLE_COMPRESSION;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_HTTP_2;
@@ -66,6 +67,8 @@ public class HttpClientBuilder implements NessieClientBuilder<HttpClientBuilder>
           .addResponseFilter(new NessieHttpResponseFilter());
 
   private boolean tracing;
+
+  private boolean enableApiCompatibilityCheck = true;
 
   protected HttpClientBuilder() {}
 
@@ -172,6 +175,11 @@ public class HttpClientBuilder implements NessieClientBuilder<HttpClientBuilder>
     s = configuration.apply(CONF_FORCE_URL_CONNECTION_CLIENT);
     if (s != null) {
       withForceUrlConnectionClient(Boolean.parseBoolean(s.trim()));
+    }
+
+    s = configuration.apply(CONF_ENABLE_API_COMPATIBILITY_CHECK);
+    if (s != null) {
+      withEnableApiCompatibilityCheck(Boolean.parseBoolean(s));
     }
 
     return this;
@@ -305,12 +313,17 @@ public class HttpClientBuilder implements NessieClientBuilder<HttpClientBuilder>
   }
 
   @CanIgnoreReturnValue
+  public HttpClientBuilder withEnableApiCompatibilityCheck(boolean enable) {
+    enableApiCompatibilityCheck = enable;
+    return this;
+  }
+
+  @CanIgnoreReturnValue
   public HttpClientBuilder withResponseFactory(HttpResponseFactory responseFactory) {
     builder.setResponseFactory(responseFactory);
     return this;
   }
 
-  @SuppressWarnings("unchecked")
   @Override
   public <API extends NessieApi> API build(Class<API> apiVersion) {
     Objects.requireNonNull(apiVersion, "API version class must be non-null");
@@ -320,16 +333,31 @@ public class HttpClientBuilder implements NessieClientBuilder<HttpClientBuilder>
       builder.addTracing();
     }
 
+    API api = buildNessieApi(apiVersion);
+
+    if (enableApiCompatibilityCheck) {
+      NessieApiCompatibility.checkApiCompatibility(api);
+    }
+
+    return api;
+  }
+
+  private <API extends NessieApi> API buildNessieApi(Class<API> apiVersion) {
+
     if (apiVersion.isAssignableFrom(HttpApiV1.class)) {
       builder.setJsonView(Views.V1.class);
       HttpClient httpClient = builder.build();
-      return (API) new HttpApiV1(new NessieHttpClient(httpClient));
+      @SuppressWarnings("unchecked")
+      API api = (API) new HttpApiV1(new NessieHttpClient(httpClient));
+      return api;
     }
 
     if (apiVersion.isAssignableFrom(HttpApiV2.class)) {
       builder.setJsonView(Views.V2.class);
       HttpClient httpClient = builder.build();
-      return (API) new HttpApiV2(httpClient);
+      @SuppressWarnings("unchecked")
+      API api = (API) new HttpApiV2(httpClient);
+      return api;
     }
 
     throw new IllegalArgumentException(
--- a/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
@@ -324,6 +324,7 @@ public class HttpClientBuilder implements NessieClientBuilder<HttpClientBuilder>
     return this;
   }
 
+  @SuppressWarnings("unchecked")
   @Override
   public <API extends NessieApi> API build(Class<API> apiVersion) {
     Objects.requireNonNull(apiVersion, "API version class must be non-null");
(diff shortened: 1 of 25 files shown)


api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibility.java
```diff
@@ -0,0 +1,58 @@
...
+              "API version %s is not supported by the server. "
+                  + "The server supports API versions from %d to %d inclusive.",
+              clientApiVersion, minServerApiVersion, maxServerApiVersion));
+    }
+    // Next, if the client is a V2 API, check that the URI prefix is indeed a V2 API root.
+    // Spec version is only returned in V2+, so we can use that as a heuristic.
+    // FIXME we need a better way to detect API vs URI prefix mismatches.
+    int serverApiVersion = specVersion != null && !specVersion.isEmpty() ? 2 : 1;
+    if (clientApiVersion != serverApiVersion) {
+      throw new IllegalArgumentException(
```
dimas-b, line 51: WDYT about using `NessieServiceException` instead? `NessieServiceException` is often reported when the URL is misconfigured in such as way that Nessie end points are not even reachable... Would it make sense to reuse it for this failure mode too?

snazy:
Can you move the fix for the GC tool from this PR?

dimas-b (commented):
The PR LGTM overall. Did you have a chance to test how the new error manifests on the Iceberg side (e.g. in Spark) when the Nessie Catalog URI is misconfigured? We may need a corresponding change in the Iceberg `NessieCatalog`.

adutra:
Switching to draft, sorry, because unsurprisingly, these changes are breaking a few compatibility tests, especially `ITOlderServers`. Obviously I could fix them by simply disabling the compatibility check, but I would like to first understand why and where the tests are failing.

adutra:
> Did you have a chance to test how the new error manifests on the Iceberg side (e.g. in Spark) when the Nessie Catalog URI is misconfigured? We may need a corresponding change in the Iceberg NessieCatalog.

Very good point, I will check.

adutra:
So, the compatibility tests were failing because of a combination of factors:

* The `NessieApi` instances are eagerly injected into the test classes, and so when e.g. the server doesn't support V2 at all, the `apiV2` field cannot be populated and the whole test class fails.
* A more interesting failure is e.g. when the server does support V2, but returns null for `specVersion`. This actually happens for many server versions, I believe because `specVersion` is a rather recent addition.

I temporarily fixed the issues by disabling the client compat check.

But this unfortunately uncovers a deeper problem: `specVersion` is not a good indicator of whether the server is replying with v1 or v2. I will think about this more and see if there is a better alternative.

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibility.java
dimas-b: `NessieServiceException` is also raised on the client side.

+1 to `NessieApiCompatibilityException`

adutra:
OK, I am proposing something different: test what the `/trees/tree/{defaultBranch}` endpoint returns. If it returns 200, then the server used v1 unambiguously, since this path cannot exist in v2. I _could_ have used `/trees/tree`, but in the unlikely case where the default branch is named `tree`, this path is ambiguous.

dimas-b:
`tree/main` is also a valid branch name :thinking:

dimas-b:
How about returning the "effective" API version from the `/config` endpoint (existing `NessieConfiguration` object or something new)?

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
```diff
@@ -323,12 +337,18 @@ public <API extends NessieApi> API build(Class<API> apiVersion) {
     if (apiVersion.isAssignableFrom(HttpApiV1.class)) {
       builder.setJsonView(Views.V1.class);
       HttpClient httpClient = builder.build();
+      if (enableApiCompatibilityCheck) {
+        NessieApiCompatibility.check(1, httpClient);
+      }
       return (API) new HttpApiV1(new NessieHttpClient(httpClient));
     }
 
     if (apiVersion.isAssignableFrom(HttpApiV2.class)) {
```
snazy, line 346: Can you move this `if` up (minor nit)

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibility.java
```diff
@@ -0,0 +1,65 @@
...
+      throw new NessieApiCompatibilityException(
+          clientApiVersion, minServerApiVersion, maxServerApiVersion, actualServerApiVersion);
+    }
+  }
+
+  private static NessieConfiguration fetchConfig(HttpClient httpClient) {
+    return httpClient.newRequest().path("config").get().readEntity(NessieConfiguration.class);
+  }
+
+  private static int fetchActualServerApiVersion(
```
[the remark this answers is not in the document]
snazy: Better to return the "effective" version (the one that was used on the server side) via `NessieConfiguration`

api/client/src/test/java/org/projectnessie/client/http/TestNessieHttpClient.java
```diff
@@ -184,13 +190,60 @@ void testUnauthorized() throws Exception {
...
       assertThatThrownBy(api::getConfig)
           .isInstanceOf(NessieNotAuthorizedException.class)
           .hasMessageContaining("Unauthorized");
     }
   }
 
+  @Test
+  void testApiCompatibility() {
+    // Good cases
+    assertThatCode(() -> testConfig(NessieApiV1.class, 1, 1, 1, true)).doesNotThrowAnyException();
```
snazy, line 204: Candidate for softly-assertions (`soft.assertThatCode...`)

compatibility/compatibility-tests/build.gradle.kts
```diff
@@ -55,6 +55,6 @@ tasks.withType<Test>().configureEach {
 // with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`
 //
 // Compatibility tests fail, because Windows not supported by testcontainers (logged message)
-if (Os.isFamily(Os.FAMILY_MAC) || Os.isFamily(Os.FAMILY_WINDOWS)) {
+if ((Os.isFamily(Os.FAMILY_MAC) || Os.isFamily(Os.FAMILY_WINDOWS)) && System.getenv("CI") != null) {
```
snazy, line 58: Why this one effectively disabling CI for this project?

adutra: This is actually _enabling_ the tests to run on my Mac. CI was already disabled previously.

(and btw the tests run just fine on Mac, I'm not sure about the comment line 55).

adutra:
> `tree/main` is also a valid branch name 🤔

Hmm right. There will always be a risk of ambiguity it seems.

> This check requires that authz works for you on that branch.

That is true as well, I didn't think of it so far. This may be a blocker.

> How about returning the "effective" API version from the /config endpoint

That is imho the proper way to solve this, but keep in mind that this won't solve the problem for existing Nessie versions, only for future ones.

I think this PR is getting stuck. As it stands, I see 2 options going forward:

1. Give up on this work and just acknowledge that there is no good way to detect URI vs API version mismatches.
2. Implement "returning the effective API version from the /config endpoint" and only run the compatibility check for versions that implemented this endpoint.

Wdyt?

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibility.java
```diff
@@ -0,0 +1,65 @@
...
+   * Checks if the API version of the client is compatible with the server's.
+   *
+   * @param clientApiVersion the API version of the client
+   * @param httpClient the underlying HTTP client.
+   * @throws NessieApiCompatibilityException if the API version is not compatible.
+   */
+  public static void check(int clientApiVersion, HttpClient httpClient)
+      throws NessieApiCompatibilityException {
+    NessieConfiguration config = fetchConfig(httpClient);
+    int minServerApiVersion = config.getMinSupportedApiVersion();
```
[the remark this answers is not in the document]
snazy: v2 is still in beta, we can somewhat ignore older versions w/ v2.

compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
```diff
@@ -115,9 +115,73 @@ void getDefaultBranchV2() throws Exception {
...
+            NessieConfiguration::getMinSupportedApiVersion,
+            NessieConfiguration::getMaxSupportedApiVersion,
+            NessieConfiguration::getActualApiVersion,
+            NessieConfiguration::getSpecVersion)
+        .containsExactly("main", 1, 2, 0, "2.0-beta.1");
+  }
+
+  @Test
+  @VersionCondition(minVersion = "0.59.0")
+  void getConfigV2WithSpecWithActualVersion() {
```
adutra, line 175: This test currently doesn't run, but will when 0.59.0 is released and included in the tested versions.

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilder.java
adutra: Actually this would break things:  `HttpApiV2` is a subclass of `NessieApiV1`.

dimas-b (approved):
The latest approach (actual API version in `NessieConfiguration`) looks good to me.

snazy (approved):

PR #6819, opened 2023-05-12
Remove the lambda module

### LINKED ISSUE #756: Decide about Lambda module

snazy:
Lambda hasn't been maintained for a long while and we are not aware of any usage of it. The source files indicate that it cannot work either, so removing it for now.

Fixes #756


dimas-b (approved):
Agreed

PR #6823, opened 2023-05-12
Allow exporting only the latest contents from a branch.

dimas-b:
This can be useful when migrating to the new Nessie data model in cases where the full previous change history do not have to be preserved.

Exporting only the latest state of contents can save time and storage (smaller export data files), when the number of content objects is significantly smaller than the number of commits.

Both old and new data models can be exported with `--only-contents-from-branch`, even though the primary use case is exporting from the old model and importing into the new model.

### DIFF
--- a/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/ExportRepository.java
+++ b/servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/ExportRepository.java
@@ -21,6 +21,8 @@ import java.nio.file.Path;
 import java.util.Locale;
 import java.util.function.Consumer;
 import javax.annotation.Nonnull;
+import org.projectnessie.versioned.persist.store.PersistVersionStore;
+import org.projectnessie.versioned.storage.versionstore.VersionStoreImpl;
 import org.projectnessie.versioned.transfer.ExportImportConstants;
 import org.projectnessie.versioned.transfer.NessieExporter;
 import org.projectnessie.versioned.transfer.ProgressEvent;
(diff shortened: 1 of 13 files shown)


versioned/transfer/src/test/java/org/projectnessie/versioned/transfer/TestContentMigrationToPersist.java
```diff
@@ -0,0 +1,149 @@
...
+  @InjectSoftAssertions protected SoftAssertions soft;
+  @NessieDbAdapter protected static DatabaseAdapter databaseAdapter;
+  @NessieDbAdapter protected static VersionStore srcStore;
+
+  @NessiePersist protected static Persist persist;
+
+  @TempDir Path dir;
+
+  private VersionStore prepareTargetRepo() {
+    // Initialize repository w/o a default branch
```
snazy, line 76: Not sure I understand - it erases and then initialized with a default branch.

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/ExportRepository.java
```diff
@@ -95,14 +97,37 @@ enum Format {
       })
   private boolean fullScan;
 
+  @CommandLine.Option(
+      names = {"--only-contents-from-branch"},
```
snazy, line 101: Nit: can you use constants for the names

snazy, line 101: Not sure the option name is understandable. Maybe: `--single-branch-current-content`

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportContents.java
```diff
@@ -0,0 +1,170 @@
...
+    try (PaginationIterator<KeyEntry> entries = store.getKeys(ref.getNamedRef(), null, false)) {
+      while (true) {
+        List<KeyEntry> batch = take(exporter.contentsBatchSize(), entries);
+        if (batch.isEmpty()) {
+          break;
+        }
+
+        ByteString meta =
+            CommitMetaSerializer.METADATA_SERIALIZER.toBytes(
+                CommitMeta.fromMessage("Export " + seq));
```
snazy, line 101: Maybe `Single branch export from '<branch>', part #<part-num>` - and `... last part` for the last one. And `seq + 1`?

versioned/transfer/src/test/java/org/projectnessie/versioned/transfer/TestContentMigrationToPersist.java
dimas-b: rephrased

servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/ExportRepository.java
dimas-b: renamed and moved to constant

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportContents.java
dimas-b: Updated the message... The `last part` looks like an overkill to me... skipped that ;)

PR #6840, opened 2023-05-16
Remove Content type hierarchy from Events API

### LINKED ISSUE #6820: Events & model types
Currently the events object model mirrors all (built in) content types from `:nessie-model` 1:1 in ':nessie-events-api`.

This issue is to think about whether it's better to do this or whether it's "enough" or maybe better (think: custom types) to handle all content types generically in the Nessie events code.

Pro "generic content" in events:
* less code
* content is "just passed through", not handled on a case-by-case bases
* feels more future proof - thinking of whether there could be differences in serialization between a "well known" and generic type - or type clashes in 3rd party event-SPI implementations.

adutra:
Fixes #6820.


events/api/src/main/java/org/projectnessie/events/api/Content.java
```diff
@@ -15,28 +15,37 @@
...
+   *   <li>{@code ICEBERG_TABLE};
+   *   <li>{@code ICEBERG_VIEW};
+   *   <li>{@code NAMESPACE};
+   *   <li>{@code UDF};
+   *   <li>{@code DELTA_LAKE_TABLE}.
+   * </ul>
+   *
+   * Other content types may be added in the future, or registered by end users.
+   */
+  String getType();
```
dimas-b, line 48: WDYT about adding something like `asModelObject()` for auto-conversion to model API java wrappers?

Clients that are sure to have the right custom types deployed, could benefit by not having to maintain the code do this conversion on their side.

dimas-b (commented):
LGTM overall.

events/api/src/main/java/org/projectnessie/events/api/Content.java
adutra: That would require a dependency on the `nessie-model` module, not sure if that is a good thing. We could maybe provide an utility class somewhere else. @snazy wdyt?

OTOH, I'm not sure how much support we want to provide for back-and-forth conversion between Event and Model. Smart users will certainly use their own schemas when processing events, because relying on Nessie's Event schema means that they are tying their business logic to Nessie's domain. So in my opinion if users need to retrieve the original Model object, they will probably have to do so by deserializing some user-specific domain object that was created from the original Event object from Nessie.

snazy (approved):
+1

dimas-b (approved):

PR #6845, opened 2023-05-17
Add ObjId.objIdFromByteBuffer()

snazy:


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjId.java
```diff
@@ -110,20 +118,43 @@ public static ObjId objIdFromBytes(ByteString bytes) {
     }
   }
 
+  /**
+   * Creates an {@link ObjId} from its bytes representation, assuming that all remaining data in
```
dimas-b, line 122: the `remaining` piece is a bit misleading... the byte array is treated whole, is it not? There is no moving pointer :thinking:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjId.java
```diff
@@ -98,6 +98,14 @@ public static ObjId objIdFromString(@Nonnull @jakarta.annotation.Nonnull String
...
```
dimas-b, line 102: the `remaining` piece is a bit misleading... `ByteString` does not have a "pointer" (unlike `ByteBuffer`), does it?

dimas-b (commented):
Code changes LGTM, just minor javadoc comments.

dimas-b (approved):

PR #6847, opened 2023-05-17
Remove native image

snazy:
The native image isn't used in the wild (10 total pulls for 0.58.0+0.58.1 vs ~500 for the Java image).

There are already issues w/ native images. Spanner + BigTable (effectively all Google services) don't work in native images due to code issues in Google code. Also there are "strange" native-test issues in the Quarkus 3 PR.

FaaS (lambda) style deployments do not really need native images. Sure, the startup is faster, but those instances cannot run for long, because SerialGC can easily cause STWs of many seconds. So keeping FaaS instances up longer (and maybe have a "spare one") should mitigate that issue.

Also, we can only build x86_64/Linux native images.


dimas-b (dismissed):

adutra:
`CI intTest Quarkus native` needs to be removed from the list of required jobs.

PR #6852, opened 2023-05-18
Quarkus ITs: Restrict Keycloak to tests using Keycloak

snazy:
Otherwise Keycloak is run for every test, not just the annotated test classes.


adutra (approved):

dimas-b (approved):
Why do we have to do this for `KeycloakTestResourceLifecycleManager` but not for other resources?

PR #6857, opened 2023-05-19
Enable compatibility tests on macOS

adutra:
This commit enables all the compatibility tests that are currently doable on macOS, namely, all tests except the ones relying on RocksDB.

Compatibility tests with RocksDB fail consistently with SIGSEGV on macOS.

### DIFF
--- a/compatibility/compatibility-tests/build.gradle.kts
+++ b/compatibility/compatibility-tests/build.gradle.kts
@@ -51,10 +51,7 @@ tasks.withType<Test>().configureEach {
   systemProperty("junit.jupiter.extensions.autodetection.enabled", "true")
 }
 
-// Compatibility tests fail on macOS with the following message: `libc++abi: terminating
-// with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`
-//
 // Compatibility tests fail, because Windows not supported by testcontainers (logged message)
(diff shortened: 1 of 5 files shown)


compatibility/compatibility-tests/build.gradle.kts
```diff
@@ -51,10 +51,7 @@ tasks.withType<Test>().configureEach {
   systemProperty("junit.jupiter.extensions.autodetection.enabled", "true")
 }
 
-// Compatibility tests fail on macOS with the following message: `libc++abi: terminating
-// with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument`
-//
 // Compatibility tests fail, because Windows not supported by testcontainers (logged message)
-if ((Os.isFamily(Os.FAMILY_MAC) || Os.isFamily(Os.FAMILY_WINDOWS)) && System.getenv("CI") != null) {
```
adutra, line 58: This will have the side-effect of enabling compatibility tests in CI. If they fail for whatever reason, I will re-disable them in CI. As long as I can run them locally it's fine :-)

adutra:
As expected CI has failed, so I just reverted the CI change. 

Podman usage in the macOS CI jobs needs some serious updates, which I will try to tackle next.

snazy (approved):

PR #6861, opened 2023-05-19
Add user guide for TLS ingress

dimas-b:


site/docs/guides/tls.md
```diff
@@ -0,0 +1,83 @@
...
+certificates in the client OS.
+
+To generate a fresh pair of a key and a (self-signed) certificate:
+
+```shell
+openssl req -new -subj "/CN=Nessie" -addext "subjectAltName = DNS:nessie.local" \
+  -newkey rsa:2048 -keyout nessie-key.pem -out nessie.crt -noenc
+```
+
+Note the `-noenc` option. It is used only for the sake of simplicity of this example deployment.
```
adutra, line 25: I see this option [here](https://www.openssl.org/docs/man3.1/man1/openssl-req.html), but on macOS, although I do have openssl 3.1.0, the option is not available:

```
openssl req -new -subj "/CN=Nessie" -addext "subjectAltName = DNS:nessie.local" \
  -newkey rsa:2048 -keyout nessie-key.pem -out nessie.crt -noenc
req: Unrecognized flag noenc
req: Use -help for summary.
```

It seems we must use `-nodes` instead. Maybe we should remove this option, if it's not strictly required for the tutorial?

site/docs/guides/tls.md
```diff
@@ -0,0 +1,83 @@
...
+
+Add the new certificate to the local set of CA certificates on the client host (where the Nessie clients,
+such as Nessie CLI / `curl` / browser, are going to run).
+
+The following example is for Linux:
+
+1. `sudo cp nessie.crt /usr/local/share/ca-certificates/nessie.crt`
+2. `sudo update-ca-certificates`
+
+This should output something like "Certificate added: CN=Nessie".
```
adutra, line 35: FYI on macOS the command should be:

```shell
security add-trusted-cert -d -r trustAsRoot -k /Library/Keychains/System.keychain nessie.crt
```

But it didn't work:

```
Password:
SecCertificateCreateFromData: Unknown format in import.
Error reading file nessie.crt
```

site/docs/guides/tls.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Creating a k8s Secret for the Nessie Certificate
+
+```shell
+kubectl -n nessie-ns create secret tls nessie-tls \
```
adutra, line 40:
```shell
kubectl -n nessie-ns create secret tls nessie-tls \
  --cert=nessie.crt --key=nessie-key.pem
error: tls: failed to find "CERTIFICATE" PEM block in certificate input after skipping PEM blocks of the following types: [CERTIFICATE REQUEST]
```

Indeed there is no CERTIFICATE section, only a CERTIFICATE REQUEST in the generated crt... the openssl command you used above is for generating a CSR (certificate signing request), not a CA certificate. Are you sure we are not missing one step here?

dimas-b: Let me double  check the commands... it looks like I missed the `-x509` options during copy-paste from my local history :facepalm:

site/docs/guides/tls.md
dimas-b: fixed

site/docs/guides/tls.md
dimas-b: Should be fixed by new instructions. Please give it another try.

site/docs/guides/tls.md
dimas-b: fixed

site/docs/guides/tls.md
```diff
@@ -0,0 +1,84 @@
...
+  --set ingress.tls[0].hosts[0]='nessie.local'
+```
+
+The deployment process may take some time. Use the following command to check its status and get the ingress IP address.
+
+```shell
+kubectl get ingress -n nessie-ns
+```
+
+Add an entry in the local hosts file (e.g. `/etc/hosts`) mapping that IP address to `nessie.local`, for example:
```
adutra, line 62: While this works, you could avoid fiddling with /etc/hosts by using one or both of the methods below:

1. for curl only: use the `--resolve` option, e.g. 

```shell
curl --resolve "nessie.local:443:$(minikube ip)" https://nessie.local/api/v2/config
```

2. for all tools and browsers: use a nip.io hostname, e.g. `nessie.$(minikube ip).nip.io`: such a host name will resolve to whatever IP `minikube ip` indicates.  E.g. 

```shell
ping nessie.$(minikube ip).nip.io
PING nessie.192.168.49.2.nip.io (192.168.49.2): 56 data bytes
64 bytes from 192.168.49.2: icmp_seq=0 ttl=64 time=0.108 ms
```

Then you would use `nessie.192.168.49.2.nip.io` as the hostname for your certificates and ingress rules.

site/docs/guides/tls.md
```diff
@@ -0,0 +1,84 @@
...
+
+```shell
+192.168.49.2	nessie.local
+```
+
+Use `curl` to verify that the server is accessible:
+
+```shell
+$ curl https://nessie.local/api/v2/config
```
adutra, line 71: I finally got this to work on macOS, so I'm approving the PR, but it required many changes:

1. First, ingress ports are not exposed on macOS, see https://github.com/kubernetes/minikube/issues/7332 for background. So mac users must run the following command after installing the ingress:

```shell
minikube tunnel
```

This will detect the ingress rule and open ports for it:

```
✅  Tunnel successfully started
📌  NOTE: Please do not close this terminal as this process must stay alive for the tunnel to be accessible ...
❗  The service/ingress nessie requires privileged ports to be exposed: [80 443]
🔑  sudo permission will be asked for it.
🏃  Starting tunnel for service nessie.
Password:
```

2. Once this is done, the service is accessible, but under 127.0.0.1, not `$(minikube ip)`. So you need to change all the commands above, e.g.

```shell
curl --resolve "nessie.local:443:127.0.0.1" https://nessie.local/api/v2/config
{
  "defaultBranch" : "main",
  "minSupportedApiVersion" : 1,
  "maxSupportedApiVersion" : 2,
  "actualApiVersion" : 2,
  "specVersion" : "2.0.0",
  "noAncestorHash" : "2e1cfa82b035c26cbbbdae632cea070514eb8b773f616aaeaf668e2f0be8f10d",
  "repositoryCreationTimestamp" : "2023-05-26T10:28:53.462334804Z",
  "oldestPossibleCommitTimestamp" : "2023-05-26T10:28:53.462334804Z"
}
```

adutra (dismissed):

site/docs/guides/tls.md
dimas-b: Yeah, but the certificate must have the exact host name, which is not known until we start / configure ingress... I guess that would create a chicken and egg problem.

Also, the java client still depends on OS-level host name resolution (although I did not put and example for it).

site/docs/guides/tls.md
dimas-b: @adutra : Could you update with Mac specifics after merging? I do not have a Mac to play with :)

site/docs/guides/tls.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Deploying Nessie with Helm
+
+```shell
+helm install nessie -n nessie-ns helm/nessie \
+  --set ingress.enabled=true \
+  --set ingress.hosts[0].host='nessie.local' \
+  --set ingress.hosts[0].paths[0]='/' \
+  --set ingress.tls[0].secretName=nessie-tls \
+  --set ingress.tls[0].hosts[0]='nessie.local'
```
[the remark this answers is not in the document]
dimas-b: Nice catch... will fix quotes presently

site/docs/guides/tls.md
dimas-b: updated

site/docs/guides/tls.md
adutra: Good idea, will do!

adutra (approved):

PR #6877, opened 2023-05-23
Don't use filters when checking API compatibility

adutra:


adutra:
This PR has some test failures because in case of HTTP 401 the error is not the same anymore, but instead of changing the tests, I think everything should come back to normal if #6878 is merged first, so let's do that.

PR #6856, opened 2023-05-19
Use nip.io domain in MinioExtension

adutra:
This commit also re-enables MinioExtension on macOS since the switch to nip.io fixes the localhost subdomain issue.

Note: this will have the side-effect of enabling MinIO tests on CI for macOS. If these fail, I will disable the tests, but for CI only.

PR #6885, opened 2023-05-24
Make nip.io usage resilient against lookup failures

snazy:
nip.io's "magic" DNS names are not always resolvable, or resolvable in every setup. This change first tries to resolve the "magic" nip.io address and falls back to the pre-#6856 behavior if the pre-check fails.


adutra:
Indeed nip.io is not resolvable if you are running a VPN with Tunnelblick for instance. But that's rather an issue with Tunnelblick imho. Do you know other cases where nip.io is not being resolved?

adutra (approved):

PR #6886, opened 2023-05-24
Build: cleanup `libs.versions.toml`

snazy:
* Move "version-less" declarations, that are not referenced by a bundle, into build files (no need for those)
* Inline single-usage version declarations, that are not referenced by build code
* Remove unnecessary bundles


gradle/libs.versions.toml
```diff
@@ -4,72 +4,38 @@
...
-guava = "31.1-jre"
 hadoop = "3.3.5"
 iceberg = "1.2.1" # While bumping this version, 'nessieClientVersion' must also be updated.
 immutables = "2.9.3"
 jacoco = "0.8.10"
 jandex = "3.1.1"
 jmh = "1.36"
 junit = "5.9.3"
-keycloak = "21.0.2"
 logback = "1.2.11"
```
adutra, line 19:
```suggestion
```

adutra (approved):

PR #6894, opened 2023-05-25
Fix `MergeKeyBehavior.DROP` in new storage model for merge

snazy:
The (legacy) behavior of `MergeKeyBehavior.DROP` is to completely ignore such content keys from the source branch to be merged. The implementation in the new model wrongly implemented it to ignore the value from the source branch, but only in case of a conflict. It must also ignore such content-keys without a conflict.

This was discovered while porting the compatibility-tests to the new storage model.

This PR adds test cases to validate the behavior (and updates an existing test).


versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseMergeTransplantSquash.java
```diff
@@ -132,9 +139,23 @@ private CreateCommit createSquashCommit(
       commitBuilder.addSecondaryParents(mergeFromId);
     }
 
+    Predicate<StoreKey> filter =
```
snazy, line 142: This filter is the central piece of this PR - it removes all `DROP`ped keys from the diff.

snazy (commented):
Note: this PR only updates the "merge squash" part. "merge unsquashed" is somewhat "deprecated".

adutra (approved):

PR #6899, opened 2023-05-25
Remove Jackson support in Events API

adutra:


snazy (commented):
Generally LGTM - but I suspect Jackson can also be removed from `nessie-events-service`

adutra:
> Generally LGTM - but I suspect Jackson can also be removed from `nessie-events-service`

Heh, I thought that too, but in fact we need Jackson there for `ContentMapping` where we convert a model `Content` into an events `Content` by serializing to and deserializing from JSON.

snazy (approved):

snazy:
> > Generally LGTM - but I suspect Jackson can also be removed from `nessie-events-service`
> 
> Heh, I thought that too, but in fact we need Jackson there for `ContentMapping` where we convert a model `Content` into an events `Content` by serializing to and deserializing from JSON.

Yea, true.

PR #6910, opened 2023-05-25
Compatibility tests: move to new storage model

snazy:
* Replaces the backend used by the compatibility tests from the "old" to the "new" storage model.
* Minimum tested Nessie version is 0.55.0, which has the new model _and_ compatibility testing support.


compatibility/compatibility-tests/src/intTest/resources/META-INF/nessie-compatibility.properties
```diff
@@ -14,31 +14,14 @@
...
-#   - Namespace properties
-# 0.28.0:
-#   - Namespaces with content-IDs
-# 0.40.0:
-#   - changed references data model
-#   - introduced open-addressing key-list
-# 0.51.1:
-#   - moved to new Maven group ID (with relocation poms for the old group ID)
+# 0.52.0 (use .3 version)
+#   - new storage model
```
dimas-b, line 18: Why bother commenting on this version if we do not test with it?

dimas-b (commented):
Changes LGTM. I'm just not sure it is time to drop support for the _old_ model in compatibility tests.

dimas-b (approved):

dimas-b (approved):

PR #6937, opened 2023-05-30
Add convenience `Content.withId(String)`

snazy:


api/model/src/main/java/org/projectnessie/model/Content.java
```diff
@@ -97,6 +97,8 @@ public interface Type {
   @JsonIgnore
   public abstract Type getType();
 
+  public abstract Content withId(String id);
```
adutra, line 100: Maybe:

```suggestion
  public abstract Content withId(@Nullable @jakarta.annotation.Nullable String id);
```

Conversely we could require `id` to be non null and introduce a `withoutId()` method, wdyt?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/ContentMapping.java
```diff
@@ -107,7 +107,7 @@ public ContentValueObj buildContent(
   @jakarta.annotation.Nonnull
   public Content assignContentId(
       @Nonnull @jakarta.annotation.Nonnull Content putValue, String contentId) {
-    return STORE_WORKER.applyId(putValue, contentId);
+    return putValue.withId(contentId);
```
adutra, line 110: Method `assignContentId` seems unused.

versioned/tests/src/main/java/org/projectnessie/versioned/testworker/OnRefOnly.java
```diff
@@ -52,10 +52,9 @@ public ByteString serialized() {
   }
 
   public OnRefOnly withId(Content c) {
-    return onRef(getOnRef(), c.getId());
+    return withId(c.getId());
```
adutra, line 55: Maybe inline?

api/model/src/main/java/org/projectnessie/model/CommitResponse.java
```diff
@@ -61,6 +61,24 @@ default Map<ContentKey, String> toAddedContentsMap() {
...
+  default <T extends Content> T contentWithId(ContentKey key, T content) {
+    List<AddedContent> addedContents = getAddedContents();
+    if (addedContents == null) {
+      return content;
+    }
+    return (T)
+        getAddedContents().stream()
+            .filter(added -> added.getKey().equals(key))
+            .findFirst()
+            .map(added -> content.withId(added.contentId()))
```
dimas-b, line 78: Not sure I understand the purpose of this method :thinking: why not simply return `added` to the caller?

api/model/src/main/java/org/projectnessie/model/CommitResponse.java
```diff
@@ -61,6 +61,24 @@ default Map<ContentKey, String> toAddedContentsMap() {
     return added.stream().collect(Collectors.toMap(AddedContent::getKey, AddedContent::contentId));
   }
 
+  /**
+   * If new content has been added to Nessie, updates the {@link Content#getId() content ID} of the
+   * given content with the generated content ID.
+   */
+  @SuppressWarnings("unchecked")
+  default <T extends Content> T contentWithId(ContentKey key, T content) {
```
adutra, line 69: Seems unused.

api/model/src/main/java/org/projectnessie/model/Content.java
snazy: Nope - ID must be nullable - otherwise you could never add new content

api/model/src/main/java/org/projectnessie/model/CommitResponse.java
dimas-b: right :facepalm:

api/model/src/main/java/org/projectnessie/model/CommitResponse.java
```diff
@@ -61,6 +61,24 @@ default Map<ContentKey, String> toAddedContentsMap() {
...
+    List<AddedContent> addedContents = getAddedContents();
+    if (addedContents == null) {
+      return content;
```
dimas-b, line 72: If the purpose of this method is for clients to obtain content IDs that have just been assigned, should we throw an exception here? If we get here that means the caller passed in a content that was either not part of the commit request, or was not "added". WDYT?

dimas-b: well, but why are we adding this method? Like @adutra mentioned, it does not appear to be used ATM :thinking:

dimas-b (approved):

PR #6943, opened 2023-05-31
Events notification system for Nessie - Reference Implementation

adutra:


events/ri/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `nessie.events.topic` - the name of the Kafka topic to publish events to. Defaults to `nessie-events`.
+- `nessie.events.repository-ids` - a comma-separated list of repository ids to watch. If not set, all repositories will
+  be watched. See below for more details.
+
+A template configuration file can be found in `src/main/resources/nessie-kafka.template.properties`.
+
+### Serialization and deserialization
+
+`KafkaEventSubscriber` uses the [Avro] library to define a domain model that is specific to the subscriber, and is not
```
tomekl007, line 43: Awesome that this version uses `Avro`.

events/ri/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `CommitEvent` is generated when a commit is made to a branch and corresponds to Nessie's `COMMIT` event type;
+- `ReferenceEvent` is generated when a branch or tag is created, updated, or deleted, and corresponds to Nessie's
+  `REFERENCE_CREATED`, `REFERENCE_UPDATED`, and `REFERENCE_DELETED` event types;
+- `OperationEvent` is generated when a PUT or a DELETE operation occurs, and corresponds to Nessie's
+  `CONTENT_STORED` and `CONTENT_REMOVED` event types.
+
+In this example, the destination Kafka topic is going to receive messages of three different types. This is a common
+scenario in Kafka, but requires some extra work to be done on both the producer and consumer sides, especially with
+respect to schema handling. The strategy we chose to use here is outlined in [Martin Kleppmann's excellent article] and
```
tomekl007, line 73: Can you elaborate more on the most important aspect in our context of why we picked this strategy? 
Was it `any events that need to stay in a fixed order must go in the same topic`

events/ri/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+The subscriber is further configured to optionally watch just a few repository ids, or all of them. This is done by
+setting the `nessie.events.repository-ids` configuration option, and is implemented by modifying the subscriber's
+`EventFilter`. Again, this is purely illustrative. Other ways of filtering events on the subscriber side are possible,
+for example one could configure the subscribe to watch only certain branches or tags.
+
+### Using headers efficiently
+
+The `KafkaEventSubscriber` implementation also illustrates how to use Kafka headers to pass additional information along
+with the message. In this case, the subscriber adds the repository id, the user and the event creation timestamp to the
```
tomekl007, line 116: Did we consider putting `creation timestamp` in the event itself? Instead of headers? It maybe beneficial in the situation when the end user would like to store the events offline for further analysis. If the `creation timestamp` is not there, users will need to create a new avro schema that contains the timestamp and map our event + timestamp to this new schema and finally save the event.

events/ri/README.md
```diff
@@ -0,0 +1,137 @@
...
+message headers, allowing consumers to filter or sort messages based on these values.
+
+### Other considerations
+
+The `KafkaEventSubscriber` implementation also illustrates how to handle events in a non-blocking way, by publishing
+messages without waiting for the broker's acknowledgement. This is achieved by using the `KafkaProducer#send` method in
+a fire-and-forget mode, along with the `acks=0` producer configuration option. If your implementation needs to wait for
```
tomekl007, line 123: As far as I know, using `acks=0` is not recommended in the majority of cases. I am not sure if using 0 in the reference implementation is the best solution.

events/ri/src/main/avro/NessieEvents.avdl
```diff
@@ -0,0 +1,39 @@
+@namespace("com.example.nessie.events.generated")
+protocol NessieEvents {
```
tomekl007, line 2: Does it model all the possible events or only a subset of them? 
If the latter, maybe we should document for end users what's the best way to extend it.

events/ri/src/main/java/org/projectnessie/events/ri/kafka/KafkaEventSubscriber.java
```diff
@@ -0,0 +1,379 @@
...
+      Properties props, Function<Properties, Producer<String, Object>> producerFactory) {
+    this.props = props;
+    this.producerFactory = producerFactory;
+  }
+
+  @Override
+  public boolean isBlocking() {
+    // We don't wait for the broker acknowledgement, so we can tell Nessie
+    // that we don't do blocking I/O.
+    return false;
```
tomekl007, line 180: how does the nessie handle the situation when `isBlocking==true`?

events/ri/src/main/java/org/projectnessie/events/ri/kafka/KafkaEventSubscriber.java
```diff
@@ -0,0 +1,379 @@
...
+   * in order.
+   *
+   * <p>Note: strict ordering of events is not guaranteed. If necessary, consider using Kafka
+   * Streams.
+   */
+  private static String eventKey(Event upstreamEvent, SpecificRecord downstreamEvent) {
+    String repositoryId = upstreamEvent.getRepositoryId();
+    String referenceName;
+    if (downstreamEvent instanceof CommitEvent) {
+      referenceName = ((CommitEvent) downstreamEvent).getTargetReference();
```
tomekl007, line 354: Why the naming inconsistency? `getTargetReference` vs `getReference` below.

events/ri/src/main/resources/META-INF/services/org.projectnessie.events.spi.EventSubscriber
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Console printing event subscriber
+org.projectnessie.events.ri.console.PrintingEventSubscriber
+
+# Kafka event subscriber
+org.projectnessie.events.ri.kafka.KafkaEventSubscriber
```
tomekl007, line 21: Is there a limit on the number of Subscribers that we can support?

events/ri/src/intTest/java/org/projectnessie/events/ri/kafka/ITKafkaEventSubscriber.java
```diff
@@ -0,0 +1,566 @@
...
+    return props;
+  }
+
+  private static void addCommonConfig(Properties props) {
+    props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, KAFKA.getBootstrapServers());
+    props.put(
+        AbstractKafkaSchemaSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG,
+        "http://localhost:" + SCHEMA_REGISTRY.getMappedPort(8081));
+    props.put(
+        AbstractKafkaSchemaSerDeConfig.VALUE_SUBJECT_NAME_STRATEGY, TopicRecordNameStrategy.class);
```
tomekl007, line 167: Where do we specify what part (which field) of the record is used by the `TopicRecordNameStrategy`?

tomekl007 (commented):
Awesome work!

events/ri/README.md
adutra: I chose this strategy for the reference impl because this strategy does not require to pre-configure the schema registry, which makes it easier to test the producer. 

IOW, when the topic is created and the first message comes in, the producer will upload the schema since it's not present in the registry yet. It will do this for each new message type, since the strategy is to have one schema per (topic + message type) pair. 

Other strategies, for example those outlined in Robert Yokota's article, require the registry to be set up will all the schemas prior to using the topic.

events/ri/README.md
adutra: Yes, the event timestamp can be part of the message if you prefer. Here I'm using headers, but it's just an example.

events/ri/src/main/avro/NessieEvents.avdl
adutra: It doesn't model merge and transplant events indeed, since in this example, we don't process them. I will add a note.

events/ri/src/main/java/org/projectnessie/events/ri/kafka/KafkaEventSubscriber.java
adutra: It defers the call to `onXYZ` to a worker thread, as opposed to calling the method on the event loop thread.

events/ri/src/main/java/org/projectnessie/events/ri/kafka/KafkaEventSubscriber.java
adutra: In Nessie's Events API, a `CommitEvent` inherits from `CommittingEvent` where 2 methods are defined: `getSourceReference()` and `getTargetReference()`. In the case of a commit, both references are always the same (but they are different for merges and transplants). 

Whereas a `ReferenceEvent` has a distinct `getReference()` method, because it's always about one single reference.

events/ri/src/main/resources/META-INF/services/org.projectnessie.events.spi.EventSubscriber
adutra: No hard limit, but the more subscribers, the more you mobilize Nessie's event loop and/or worker threads. I am planning some benchmarks later on.

events/ri/src/intTest/java/org/projectnessie/events/ri/kafka/ITKafkaEventSubscriber.java
adutra: In this strategy, the schema is recorded under the subject `<topic name>-<schema name>`, so you don't need to select a field in the record. E.g. for the Avro `ReferenceEvent` message, the schema name is `ReferenceEvent`.

events/ri/README.md
```diff
@@ -0,0 +1,137 @@
...
+- `nessie.events.topic` - the name of the Kafka topic to publish events to. Defaults to `nessie-events`.
+- `nessie.events.repository-ids` - a comma-separated list of repository ids to watch. If not set, all repositories will
+  be watched. See below for more details.
+
+A template configuration file can be found in `src/main/resources/nessie-kafka.template.properties`.
+
+### Serialization and deserialization
+
+`KafkaEventSubscriber` uses the [Avro] library to define a domain model that is specific to the subscriber, and is not
+tied in any way to Nessie's. This is indeed the recommended approach, as it allows the subscriber to define its own
```
omarsmak, line 44: I am not sure why tiding the implementation with Avro? Wouldn't make sense to have approach to let users definer their "choice" format? You could have some prefer using Avro, some may prefer protobuf and some may even prefer JSON :) . However, since is an example, could be fine though.

omarsmak (commented):
Question, I guess we can't have an example based on KafkaConnect since the producer runs in Nessie server, correct?

events/ri/README.md
adutra: I think this would go beyond the purpose of a reference implementation. As you said, this is an example and what matters most imho is that people can easily adapt it to their own needs. Here, it shouldn't be very hard to switch to protobuf for instance, so I don't see a lot of value in actually providing the protobuf bindings + another subscriber.

adutra:
> I guess we can't have an example based on KafkaConnect since the producer runs in Nessie server, correct?

That, and also the fact that providing a full-blown source connector for Nessie would certainly be an interesting project to do, but also a bit out of scope for a reference implementation of the Events SPI module (especially because it would be less portable / adaptable by a variety of users).

events/ri/README.md
adutra: Sorry for the late reply. You are right, and I had to open my Kafka in Action book to refresh my skills around acks and send methods :-)

In short, we have 3 send methods:

1. fire-and-forget: don't wait for the future to complete, don't provide any callback;
2. async: don't wait for the future to complete, but provide a completion callback;
3. sync: wait for the future to complete (synchronously).

With Nessie, 1 and 2 are fine, but 3 should be avoided (or at least, `isBlocking()` should return true).

The way these methods really work will further depend on the `acks` setting (and `retries`). In short, the returned future is completed when:

1. With acks=0 => immediately when the message is sent;
2. With acks=1 => when the broker acknowledged the write;
3. With acks=all => when the broker and all the in-sync replicas acknowledged the write.

Which brings us to the conclusion: the number of acks is relatively orthogonal to the send method. 

With all that in mind, I will indeed change the example to use the async send method AND ack=1, since both seem reasonable choices for this use case.

events/ri/src/main/avro/NessieEvents.avdl
adutra: Done. I added comments to all record types in the Avro schema, and also included a new record type to model Merges and Transplants (it's not used in the example, but the record definition is there). Let me know what you think.

adutra:
@tomekl007 or @omarsmak could I get another round of review please?

gradle/projects.main.properties
```diff
@@ -6,6 +6,7 @@ nessie-compatibility-tests=compatibility/compatibility-tests
 nessie-compatibility-jersey=compatibility/jersey
 nessie-events-api=events/api
 nessie-events-quarkus=events/quarkus
+nessie-events-ri=events/ri
```
snazy, line 9: Not in the nessie-bom?

adutra: None of the event modules is declared in `:nessie-bom`... I guess that's an oversight.

dimas-b (approved):
LGTM :)

PR #6945, opened 2023-05-31
Enhancements to the Events API

adutra:
This PR introduces two enhancements that seemed interesting to me while preparing #6943 :

1. Get rid of the awkward `CommitEvent.getSourceReference()` method by introducing a few more common interfaces.
2. Introduce a new `ContentEvent.getCommitCreationTimestamp()` that might be helpful to determine the order in which the operations were applied.

### DIFF
--- a/events/api/src/main/java/org/projectnessie/events/api/CommitMeta.java
+++ b/events/api/src/main/java/org/projectnessie/events/api/CommitMeta.java
@@ -53,11 +53,11 @@ public interface CommitMeta {
   /** The commit message. */
   String getMessage();
 
-  /** The commit time. */
-  Instant getCommitTime();
+  /** The commit timestamp, as set by the server. */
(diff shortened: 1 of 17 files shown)


events/api/src/main/java/org/projectnessie/events/api/CommitMeta.java
```diff
@@ -53,11 +53,11 @@ default Optional<String> getSignedOffBy() {
   /** The commit message. */
   String getMessage();
 
-  /** The commit time. */
-  Instant getCommitTime();
+  /** The commit timestamp, as set by the server. */
+  Instant getCommitTimestamp();
```
adutra, line 57: Renaming just to be consistent with `Event.getEventCreationTimestamp()`.

snazy (approved):

PR #6950, opened 2023-06-01
Let new-storage DynamoDb use Apache Http Client

snazy:


versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoClientProducer.java
```diff
@@ -18,7 +18,7 @@
 import java.net.URI;
 import org.immutables.value.Value;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
-import software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
```
XN137, line 21: whats the reasoning for only doing this change here but not also in `DynamoDatabaseClient.java` ?

versioned/storage/dynamodb/build.gradle.kts
```diff
@@ -44,11 +44,8 @@ dependencies {
   implementation(libs.slf4j.api)
 
   implementation(platform(libs.awssdk.bom))
-  implementation("software.amazon.awssdk:dynamodb") {
-    exclude("software.amazon.awssdk", "apache-client")
-  }
-  implementation("software.amazon.awssdk:netty-nio-client")
-  implementation("software.amazon.awssdk:url-connection-client")
```
XN137, line 51: we seem to be using `software.amazon.awssdk:url-connection-client` in lots of other places as well... wouldnt it make sense to switch all awssdk usage to the same apache http client?

XN137 (approved):
change looks right for new dynamodb storage but not sure i understand why other parts are not switched over as well

versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoClientProducer.java
snazy: That's code that'll go away.

versioned/storage/dynamodb/build.gradle.kts
snazy: Others are testing modules

XN137: if the apache client is the preferred choice for production modules, why not also for testing modules?

PR #6952, opened 2023-06-01
Add some version-store microbenchmarks

snazy:
This can serve as a tool to later optimize these parts in the Nessie code base:
* version store implementation
* database specific `Persist` implementations
* storate logic implementations

So far, the results look good enough, no concerning outliers w/ in-memory, mongo + cassandra.

Microbenchmarks use `BackendTestFactory` to locate and initialize a `Backend` (and in turn `Persist`) implementation. Microbenchmarks are intended to inspect the Nessie code down up to the database level, but not intended to optimize configurations for a particular database. In other words: only the Nessie code is interesting, but the performance of the database itself is not (that much).

The `BackendTestFactory` interface got a new function `getName()` and the corresponding service definitions have been added.

### DIFF
--- a/servers/services-bench/README.md
+++ b/servers/services-bench/README.md
@@ -1,28 +1,42 @@
 # Nessie VersionStore micro benchmarks
 
 Building:
+
 ```bash
 ./gradlew :nessie-services-bench:jmhJar
 ```
 
 Running:
+
 ```bash
 java -jar servers/services-bench/build/libs/nessie-services-bench-*-jmh.jar
 ```
 
 ## Async-profiler
 
-See the [Async Profiler repo](https://github.com/async-profiler/async-profiler) for a pre-built library or how to build
-it from source.
+See the [Async Profiler repo](https://github.com/async-profiler/async-profiler) for a pre-built library or how to build it from source and also how to add it to a
+JMH run. The following examples worked at the time of writing this README.
 
-Running (Linux):
+### Running on Linux
+
 ```bash
-LD_LIBRARY_PATH=(PATH-TO-LIBRARY)/ java \
+ASYNC_PROFILER_DLL=(path to async profiler libararies)
+LD_LIBRARY_PATH=${ASYNC_PROFILER_DLL} java \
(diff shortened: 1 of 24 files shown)


servers/services-bench/README.md
```diff
@@ -0,0 +1,28 @@
...
+Running (Linux):
+```bash
+LD_LIBRARY_PATH=(PATH-TO-LIBRARY)/ java \
+  -jar servers/services-bench/build/libs/nessie-services-bench-*-jmh.jar \
+  -prof async
+```
+
+## Linux Perf tools
+
+Install the appropriate `linux-tools` package for your distribution, or `make` it from the Linux sources in `tools/perf`
```
adutra, line 27: From what I could find, the `linux-tools` package is a Ubuntu thing. Maybe we could rephrase this paragraph and focus more on which commands need to be available, rather than which packages contain those commands.

servers/services-bench/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Async-profiler
+
+See the [Async Profiler repo](https://github.com/async-profiler/async-profiler) for a pre-built library or how to build
+it from source.
+
+Running (Linux):
+```bash
+LD_LIBRARY_PATH=(PATH-TO-LIBRARY)/ java \
```
adutra, line 20: FYI, the only command that worked for me on macOS was:

```shell
java -Djava.library.path=/path/to/async-profiler-2.9-macos/build/ -jar servers/services-bench/build/libs/nessie-services-bench-*-jmh.jar -prof async
```

Neither `LD_LIBRARY_PATH` nor `DYLD_LIBRARY_PATH` worked.

adutra (approved):

servers/services-bench/README.md
snazy: Rephrased

servers/services-bench/README.md
snazy: Updated README

adutra (approved):

PR #6954, opened 2023-06-01
Move to Quarkus provided observability

### LINKED ISSUE #6700: Evaluate: Replace use of `Tracing/MetricsVersionStore` in `:nessie-quarkus*`

snazy:
Uses delegates for `VersionStore` and `Persist` annotated for Quarkus metrics + tracing instead of the hand-written `Tracing`/`MetricsVersionStore` and `nessie-versioned-storage-telemetry`.

Fixes #6700


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ObservingVersionStore.java
```diff
@@ -0,0 +1,285 @@
...
+      @Nonnull BiConsumer<ContentKey, String> addedContents)
+      throws ReferenceNotFoundException, ReferenceConflictException {
+    return delegate.commit(branch, referenceHash, metadata, operations, validator, addedContents);
+  }
+
+  @WithSpan
+  @Override
+  @Counted(PREFIX)
+  @Timed(value = PREFIX, histogram = true)
+  public MergeResult<Commit> transplant(TransplantOp transplantOp)
```
adutra, line 127: Here, and in a few other places, we are losing the granularity that we used to have for span attributes. The old code had this:

```java
 b.setAttribute(TAG_TARGET_BRANCH, safeRefName(transplantOp.toBranch()))
                        .setAttribute(TAG_HASH, safeToString(transplantOp.expectedHash()))
                        .setAttribute(
                            TAG_TRANSPLANTS, safeSize(transplantOp.sequenceToTransplant())),
```

Are we OK with this? Alternatively, you could:

1. create the span manually (but I suppose you don't want to, because if the tracing extension is not present at runtime, this would result in a linkage error – annotations are much more flexible);
2. call a delegate method with parameters for all the fields in `TransplantOp` that we want span attributes for.

Wdyt?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java
```diff
@@ -58,19 +58,6 @@ public boolean isNewStorage() {
   @WithDefault("INMEMORY")
   VersionStoreType getVersionStoreType();
 
-  /**
-   * Whether calls against the version-store are traced with OpenTracing/OpenTelemetry (Jaeger),
-   * enabled by default.
-   */
-  @WithName("trace.enable")
-  @WithDefault("true")
-  boolean isTracingEnabled();
```
adutra, line 67: If I am following correctly, the idea is now to rely solely on which extensions are available at runtime, and let the annotations do their magic. Is that correct? One minor drawback of that is that if one extension is missing, traces or metrics won't be produced but we won't have any errors, not even warnings.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/PersistProvider.java
```diff
@@ -134,23 +129,8 @@ public Persist producePersist() {
 
     LOGGER.info("Creating/opening version store {} ...", versionStoreType);
 
-    BiFunction<Persist, String, Persist> wrapPersistTracing = (p, name) -> p;
-    String tracingInfo = "without tracing";
-    if (versionStoreConfig.isTracingEnabled()) {
-      if (opentelemetryTracer.isUnsatisfied()) {
-        LOGGER.warn(
-            "OpenTelemetry is enabled, but not available, forgot to add quarkus-opentelemetry?");
```
adutra, line 142: It's a bit sad that we are losing the ability to warn users in such cases. But I guess we can live with that.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
```diff
@@ -96,22 +87,6 @@ public VersionStore getVersionStore() {
...
-        }
-      }
-      if (storeConfig.isMetricsEnabled()) {
-        if (meterRegistry.isUnsatisfied()) {
-          LOGGER.warn("Metrics are enabled, but not available, forgot to add quarkus-micrometer?");
-        } else {
-          versionStore = new MetricsVersionStore(versionStore, meterRegistry.get(), Clock.SYSTEM);
-        }
-      }
       return versionStore;
```
adutra, line 90: The usage of the `@NotObserved` qualifier is certainly clever, but wouldn't it be much simpler to just:

```java
      return new ObservingVersionStore(versionStore);
```

Since the `ObservingVersionStore` bean will always be created anyways, and is the only one injectable on endpoints expecting a `@Default` bean?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
```diff
@@ -133,7 +108,7 @@ private VersionStore persistVersionStore() {
 
   private VersionStore databaseAdapterVersionStore() {
     try {
-      DatabaseAdapter da = databaseAdapter.select().get();
+      DatabaseAdapter da = databaseAdapter.select(WIthInitializedRepository.Literal.INSTANCE).get();
```
adutra, line 111: This doesn't look required. I suppose it's just an attempt to create a symmetry with how the new-model `Persist` instance is created and for correctness, since, after all, the old-model bean performs repo initialization.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java
adutra: Also, you probably need to remove the following lines as well:

https://github.com/projectnessie/nessie/blob/d17a9ed9e3ea29a8d8126f06886c1950527efbe2/events/quarkus/src/main/java/org/projectnessie/events/quarkus/config/QuarkusEventConfig.java#L39-L47

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
snazy: IIRC the annotated metrics/tracing would not work :(

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
snazy: Sadly not only for symmetry. Without the literal it barks "I got two beans!" :(

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ObservingVersionStore.java
snazy: Yea - the drawback is that we lose a few attributes. However, I think we didn't really use those attributes.

snazy: Sure, we can add delegate methods - if there's demand for that.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
snazy: The Nessie-Quarkus-CLI tool comes into play here as well.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java
snazy: Hm - metrics & tracing in for events already uses Quarkus stuff. I suspect, it's okay to leave it there.

adutra: I think you should remove them, and replace the occurrences with `true`. It should be a really minor change.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
adutra: Hmm OK the annotations must be on a managed bean indeed.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/versionstore/ConfigurableVersionStoreFactory.java
adutra: Ah true, now we have the `ObservingDatabaseAdapter` bean, so a qualifier is required to disambiguate 👍

adutra (approved):

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java
snazy: Yea, I tried, but there are indeed some conflicts. :(

PR #6958, opened 2023-06-02
Throw when v2 requested but v1 provided

### LINKED ISSUE #6946: Enabling withEnableApiCompatibilityCheck to true does not enforces v2 in URL
While trying to create the NessieApiV2 object with the nessie endpoint - http://localhost:19120/api/v1 and enabling the param 
`withEnableApiCompatibilityCheck` to `true`, I can create the object successfully with no issues but was expecting that it should throw error instead.

**What happened**:
Tried to enable the compatibility parameter `withEnableApiCompatibilityCheck` to make sure v1 is not part of URL and it throws error instead while building the NessieApiV2 instance.

Throwing an error for the invalid url

Spin up the latest OSS Nessie and pass the URL as http://localhost:19120/api/v1 and try to create the NessieApiV2 instance using below:
```
HttpClientBuilder.builder()
        .withUri(URI.create(nessieEndpoint));
        .withEnableApiCompatibilityCheck(true)
        .build(NessieApiV2.class);
```

**Details**:
- Nessie server type (docker/built from source) and version: 0.60.1

adutra:
Fixes #6946


snazy (approved):

dimas-b (commented):
+1

PR #6975, opened 2023-06-03
Micro-optimizations in ContentKey + TypeMapping

snazy:


api/model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -165,22 +166,35 @@ public static ContentKey of(@JsonProperty("elements") List<String> elements) {
...
+    int sum = 0;
     for (String e : elements) {
       if (e == null) {
         throw new IllegalArgumentException(
-            String.format("Content key '%s' must not contain a null element.", elements));
+            format("Content key '%s' must not contain a null element.", elements));
       }
-      if (e.isEmpty()) {
+      int l = e.length();
+      sum += l;
```
adutra, line 176: Since we are micro-optimizing, this statement could be moved to line 189.

api/model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -165,22 +166,35 @@ public static ContentKey of(@JsonProperty("elements") List<String> elements) {
...
+              format(
+                  "Content key '%s' must not contain characters less than 0x%2h.",
+                  elements, FIRST_ALLOWED_KEY_CHAR));
+        }
       }
     }
+    if (sum > MAX_LENGTH) {
+      throw new IllegalStateException("Key too long, max allowed length: " + MAX_LENGTH);
+    }
+    if (elements.size() > MAX_ELEMENTS) {
```
adutra, line 194: This if block could be moved to line 169: there is no point in iterating over the elements, if we know that there are too many elements.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TypeMapping.java
```diff
@@ -196,7 +196,11 @@ private static StringBuilder keyToStoreKeyPrepare(
     // Note: the relative values of outer and inner (key elements) separators affect the correctness
     // of StoreKey comparisons WRT to ContentKey comparisons. The inner separator must be greater
     // than the outer separator because longer ContentKeys are greater than shorter ContentKeys.
-    StringBuilder sb = new StringBuilder();
+    int len = 4;
```
adutra, line 199: Why 4?

api/model/src/main/java/org/projectnessie/model/ContentKey.java
snazy: That would optimize for the bad case though.

api/model/src/main/java/org/projectnessie/model/ContentKey.java
snazy: That would optimize for the bad case though.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TypeMapping.java
snazy: Added comment

PR #6985, opened 2023-06-05
Do not initialize new model repository for CLI tool

### LINKED ISSUE #6367: Import/new model: do not initialize the repository for import
The logic in `org.projectnessie.quarkus.providers.PersistProvider#producePersist` unconditionally initializes the Nessie repository, which lets the Nessie import tool complain that the repository already exists.

Expected behavior is to _not_ create the Nessie repository with the CLI tool.

snazy:
Fixes #6367


servers/quarkus-cli/src/main/java/org/projectnessie/quarkus/cli/EraseRepository.java
```diff
@@ -48,6 +48,11 @@ public class EraseRepository extends BaseCommand {
   protected Integer callWithPersist() {
     warnOnInMemory();
 
+    if (!repositoryLogic(persist).repositoryExists()) {
+      spec.commandLine().getErr().println("Nessie repository does not exist");
+      return 1;
```
dimas-b, line 53: Maybe use a different exit code in case shell tools need to distinguish this failure mode from error that happen during the actual repo deletion process?

dimas-b (approved):

PR #6994, opened 2023-06-06
DynamoDB: let Quarkus use the Apache HTTP client

snazy:


adutra:
What is the expected improvement of switching to Apache HTTP client?

snazy:
`(Http)URLConnection` is quite antique and no longer really maintained. It has a bunch of disadvantages. Basically the same reason why we're using the Java 11's new HTTP client in `nessie-client`.

Apache HTTP client (should) give better throughput and such things.
BTW: Iceberg's already moved to that one for S3 requests.

PR #7004, opened 2023-06-07
Micro-opt `TypeMapping`

snazy:
* Replacing implicit `Iterator`
* store-key to content-key: initialize w/ 10 to prevent `grow()` for every added element

### DIFF
--- a/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TypeMapping.java
+++ b/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TypeMapping.java
@@ -107,7 +107,7 @@ public final class TypeMapping {
       variant = raw.substring(idx1 + 1);
     } else {
 
-      keyElements = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TypeMapping.java
```diff
@@ -107,7 +107,7 @@ public static ContentKey storeKeyToKey(@Nonnull @jakarta.annotation.Nonnull Stor
       variant = raw.substring(idx1 + 1);
     } else {
 
-      keyElements = new ArrayList<>();
+      keyElements = new ArrayList<>(10);
```
dimas-b, line 110: Would it be worth counting the number of `1` bytes when we search for `idx2` so as to have an exact size for `keyElements`?

snazy: I doubt

PR #6890, opened 2023-05-24
Remove database adapter code

snazy:
Fixes #6701

PR #7035, opened 2023-06-13
Identify merge-base instead of common-ancestor

### LINKED ISSUE #4562: "Incremental merge" - allow merge of additional changes
When working on a branch, it is currently only possible to merge one branch into another once.

```
                ----------COMMIT----COMMIT----COMMIT----MORE----MORE----
               /                                 \                      \
              /                                   \                      \
---main----COMMIT---------------COMMIT-----------MERGE-----COMMIT------ANOTHER_MERGE----
```

Only the first `MERGE` works, but `ANOTHER_MERGE` cannot.
In the past we did not record the merge-parent at all. But since recently we do and merge could be more clever now.
While identifying the common-ancestor, it could look for the latest merge-parent on the target branch that is present on the source branch.

The `MERGE` would not find a merge-parent that is a commit on the source branch - therefore look for the common ancestor. 
The `ANOTHER_MERGE` would find a merge-parent that is a commit on the source branch - and then use the commits on the source branch _after_ the latest merged commit from the source branch.

snazy:
Enhances the existing logic to identify the common ancestor to respect merge parents.

Externalizes the logic into a new class `MergeBase`, which combines the logic for the common-ancestor and merge-base use cases.

Currently the logic for a merge operation is to identify the common ancestor and apply the diff between the common ancestor and the "from HEAD" onto the "target HEAD".

This change will use the "best merge base" instead of the "common ancestor".

This change also allows quite complex merge scenarios - see `TestMergeBase` for examples. With that the actually "merged commits" are extremely difficult to identify, because those can come from a potentially huge variety of "commit tree paths". This implies these user visible changes:
* Merges (with the new database model) _always_ produce a squash merge commit. Even for single-commit fast-forward merges.
* Merges (with the new database model) that request individual commits are silently squash-merged
* Merge commit messages changed from `Merged N commits from REF-NAME ...` to `Merged REF-NAME ...`
* Merge author is the committer (unless the merge-request passes a `CommitMeta` requesting different author(s))

Fixes #4562


servers/services/src/testFixtures/java/org/projectnessie/services/impl/AbstractTestMergeTransplant.java
```diff
@@ -210,19 +210,34 @@ private void mergeTransplant(
...
+      // commits to merge, reported as "successful".
+      soft.assertThat(actor.act(target, source, committed1, committed2, false))
+          .extracting(
+              MergeResponse::getCommonAncestor,
+              MergeResponse::getEffectiveTargetHash,
+              MergeResponse::getResultantTargetHash,
+              MergeResponse::wasApplied,
+              MergeResponse::wasSuccessful)
+          .containsExactly(committed2.getHash(), newHead.getHash(), newHead.getHash(), false, true);
+      ;
```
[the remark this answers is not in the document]
snazy: oh

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/MergeSquashImpl.java
```diff
@@ -52,7 +52,7 @@ final class MergeSquashImpl extends BaseMergeTransplantSquash implements Merge {
   public MergeResult<Commit> merge(Optional<?> retryState, MergeOp mergeOp)
       throws ReferenceNotFoundException, RetryException, ReferenceConflictException {
     ObjId fromId = hashToObjId(mergeOp.fromHash());
-    ObjId commonAncestorId = identifyCommonAncestor(fromId);
+    ObjId commonAncestorId = identifyMergeBase(fromId).id();
 
     SourceCommitsAndParent sourceCommits = loadSourceCommitsForMerge(fromId, commonAncestorId);
```
dimas-b, line 57: Not sure this will work correctly when the merge base is not a direct ancestor... `loadSourceCommitsForMerge` appears to assume that both hashes are part of the predecessor-based commit log.

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
```diff
@@ -0,0 +1,434 @@
...
+   *
+   * <p>Merge-base outcome for {@code F} onto {@code G} can be {@code D} or {@code E}
+   * (non-deterministic!).
+   */
+  @Test
+  void afterCrossMerge() {
+    CommitObj a = repo.add(initialCommit());
+    CommitObj b = repo.add(buildCommit("b").addTail(a.id()));
+    CommitObj c = repo.add(buildCommit("c").addTail(a.id()));
+    CommitObj d = repo.add(buildCommit("d").addTail(b.id()).addSecondaryParents(b.id()));
```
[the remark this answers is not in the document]
snazy: ah, yes, that's wrong

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
```diff
@@ -0,0 +1,434 @@
...
+    CommitObj g = repo.add(buildCommit("g").addTail(e.id()));
+
+    soft.assertThat(
+            MergeBase.builder()
+                .loadCommit(repo::loadCommit)
+                .targetCommitId(g.id())
+                .fromCommitId(f.id())
+                .build()
+                .identifyMergeBase())
+        .isIn(d, e);
```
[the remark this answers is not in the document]
snazy: ah, yes, that's wrong

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
```diff
@@ -0,0 +1,434 @@
...
+   * <p>Merge-base outcome for {@code F} onto {@code G} can be {@code D} or {@code E}
+   * (non-deterministic!).
+   */
+  @Test
+  void afterCrossMerge() {
+    CommitObj a = repo.add(initialCommit());
+    CommitObj b = repo.add(buildCommit("b").addTail(a.id()));
+    CommitObj c = repo.add(buildCommit("c").addTail(a.id()));
+    CommitObj d = repo.add(buildCommit("d").addTail(b.id()).addSecondaryParents(b.id()));
+    CommitObj e = repo.add(buildCommit("e").addTail(c.id()).addSecondaryParents(d.id()));
```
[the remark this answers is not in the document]
snazy: ah, yes, that's wrong

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/MergeSquashImpl.java
snazy: Ouch - yes.
As discussed offline, the commit-meta information will no longer be constructed from the "merged commits", because that would become a real PITA wrt to more complex merge scenarios (see the `TestMergeBase` test cases).

All new commits are there to serve the above purpose:
* the actual production code changes are relatively simple
* there are a bunch of necessary simplification and cleanup changes
* the necessary test code changes make the tests quite complex wrt to the test scenarios and data model being tested - #6890 will eventually clean this m... up

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseMergeTransplantSquash.java
```diff
@@ -164,7 +162,11 @@ private CreateCommit createSquashCommit(
     CommitLogic commitLogic = commitLogic(persist);
     PagedResult<DiffEntry, StoreKey> diff =
         commitLogic.diff(
-            diffQuery(sourceCommits.sourceParent, sourceCommits.mostRecent(), true, filter));
+            diffQuery(
+                mergeTransplantContext.baseCommit(),
+                mergeTransplantContext.headCommit(),
```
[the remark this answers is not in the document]
dimas-b: I tend to think a preferable approach would be to do the diff on the commit histories (similar to `git log main..dev`) as opposed to specific index states at the end of branches. Then we could aggregate incremental changes from commits not yet included into the target branch and use them to make the merge commit.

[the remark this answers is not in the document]
snazy: Collecting and applying "all the commits to be merged" is extremely difficult, and also not the expected/wanted behavior. That will eventually end up in the diff anyway, because the diff is already the result of "applying all the commits".
It is not as easy as `git log main..dev` or the like - it has to traverse _all_ possible graphs of all the merges that happened in between, which is way more complicated and not necessary.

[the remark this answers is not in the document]
snazy: >  I think conflict should not occur in this case, because B has already been merged into the history of G. WDYT?

That's a different issue - it*s applying the exact same thing again - that should not cause a conflict but just ignore that change.

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
```diff
@@ -0,0 +1,438 @@
...
+    CommitObj g = repo.add(buildCommit("g").addTail(e.id()));
+
+    soft.assertThat(
+            MergeBase.builder()
+                .loadCommit(repo::loadCommit)
+                .targetCommitId(g.id())
+                .fromCommitId(f.id())
+                .build()
+                .identifyMergeBase())
+        .isEqualTo(b);
```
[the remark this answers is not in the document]
snazy: Only `B` will be selected here, because of the `.get(0)` in the last line of `MergeBase.identifyMergeBase()`, which chooses the "best" merge base.

We should however add an ability to let a user chose the "right/best" merge-base - as a follow-up though. Not sure whether that particular test case is a legit prod use case anyway. Manually choosing the "right/best" merge-base is difficult for humans in complex graphs anyway. I cannot remember a case when I had to force Git to use a specific merge-base.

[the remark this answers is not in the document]
snazy: See added VS-test cases

snazy:
The added commits enforce "squash merges" + "individual transplants" for all API calls. This also simplifies the "historically grown" tests. The last commit also adds a new, hopefully much simpler way to check merge scenarios, currently just mirroring the test cases in `TestMergeBase`.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseMergeTransplantSquash.java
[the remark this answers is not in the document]
dimas-b: I guess, you're right :+1:

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
```diff
@@ -0,0 +1,479 @@
...
+    CommitObj j = repo.add(buildCommit("j").addTail(h.id()));
+
+    soft.assertThat(
+            MergeBase.builder()
+                .loadCommit(repo::loadCommit)
+                .targetCommitId(g.id())
+                .fromCommitId(j.id())
+                .build()
+                .identifyMergeBase())
+        .isEqualTo(b.id());
```
dimas-b, line 289: This assert contradicts the test's javadoc (line 268). I also think `D` is the best merge base in this case.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
```diff
@@ -0,0 +1,324 @@
...
+    List<ShallowCommit> result = new ArrayList<>();
+
+    commitA.setCommitA();
+    queue.add(commitA);
+
+    commitB.setCommitB();
+    queue.add(commitB);
+
+    while (queue.stream().anyMatch(ShallowCommit::isNotCandidate)) {
+      ShallowCommit commit = requireNonNull(queue.poll());
```
dimas-b, line 124: I'm not sure the iteraction order is ideal here. This is what I observed for `TestMergeBase.multiMerge2()`:

```
commit ID: 00000006, seq: 6
commit ID: 00000003, seq: 3
commit ID: 00000000, seq: 0
commit ID: 00000005, seq: 5
commit ID: 00000004, seq: 4
commit ID: 00000002, seq: 2
commit ID: 00000001, seq: 1
commit ID: 00000008, seq: 8
commit ID: 00000007, seq: 7
commit ID: 00000001, seq: 1
commit ID: 00000000, seq: 0
commit ID: 00000003, seq: 3
```

I think it'd be preferable to process from more recent commits to older commits (i.e. reversing the priority queue's comparator). Then, perhaps we won't need the extra `removeRedundant` step?.. WDYT?

dimas-b (commented):
Posting some preliminary thoughts.. I'll keep reviewing.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
dimas-b: Also, current iteration strategy will always go to the root, it seems, but we should be able to stop as soon as we reach the deepest merge base (which is not root in most cases).

versioned/spi/src/main/java/org/projectnessie/versioned/DefaultMetadataRewriter.java
```diff
@@ -90,14 +92,16 @@ private CommitMeta buildCommitMeta(
 
   @Override
   public CommitMeta rewriteSingle(CommitMeta metadata) {
-    return buildCommitMeta(CommitMeta.builder().from(metadata), metadata::getMessage);
+    return buildCommitMeta(
+        CommitMeta.builder().from(metadata).hash(null).parentCommitHashes(emptyList()),
```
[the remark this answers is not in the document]
snazy: No need to, it's a new commit.

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestMergeBase.java
snazy: bug in test code, fixed

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
snazy: It can "shoot over the common ancestor" a little bit, but won't traverse the whole history.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
```diff
@@ -0,0 +1,325 @@
...
+
+    return findMergeBases(fromCommit, targetCommit);
+  }
+
+  private List<ShallowCommit> findMergeBases(ShallowCommit commitA, ShallowCommit commitB) {
+    if (commitB.id().equals(commitA.id())) {
+      return newArrayList(commitA);
+    }
+
+    List<ShallowCommit> reachableCommits = flagReachableCommits(commitA, commitB);
```
dimas-b, line 106: returning `reachableCommits` at this point does not break `TestMergeBase`. I tend to think this list already contains best merge base as the first element, but if I missed something perhaps we need another test case... WDYT?

snazy: Seems we can at least short-cut when there's only one reachable commit.

dimas-b: The short-cut is fine, but I'm still not sure that `removeRedundant` is required anymore.

`flagReachableCommits` already returns results in sequence order (greatest first) because of the new deque sort order, which basically means the merge bases closer to the target/source HEADs will be closer to the start of the `reachableCommits` list.

The only uncertainty is when the history graph is symmetrical and we have more than one merge base with the same sequence number (for example the case in `TestMergeBase.afterCrossMerge()`). I think it might be preferable to resolve that by including commit time into the deque comparator (and commit ID to ensure total order). Then, the output of `flagReachableCommits` will be deterministic with the highest sequence number at the first element. WDYT?

Alternatively, we could prefer merge bases that are direct descendants of the "target" commit (add a new flag to mark them).

snazy: Not sure whether it's worth to change it. The logic can still return less results - just not sure how to construct a test case for that.

dimas-b: > The logic can still return less results [...]

Well, I'm not sure this is actually possible... or rather, it feels like we can adjust the comparator in `flagReachableCommits` to make the `removeRedundant` unnecessary and greatly simplify the code.

snazy: Do you insist?

dimas-b (approved):

PR #7037, opened 2023-06-13
[Nessie Catalog Server] - OAuth token endpoint implementation

adutra:
I've been able to manually validate this with Postman and spark-sql.

1. Nessie server and Keycloak running from the in-tree docker-compose file + `KC_HOSTNAME_URL` env var;
2. Manually created client-id `nessie-client` with appropriate permissions and `catalog` scope;
3. REST catalog server started with `-Dquarkus.oidc.auth-server-url=http://keycloak:8080/realms/master`
4. Spark SQL launch command:

```shell
spark-sql \
    --packages org.apache.iceberg:iceberg-spark-runtime-3.4_2.12:1.3.0 \
    --conf spark.sql.extensions=org.apache.iceberg.spark.extensions.IcebergSparkSessionExtensions \
    --conf spark.sql.catalog.nessie=org.apache.iceberg.spark.SparkCatalog \
    --conf spark.sql.catalog.nessie.catalog-impl=org.apache.iceberg.rest.RESTCatalog \
    --conf spark.sql.catalog.nessie.uri=http://127.0.0.1:19130/iceberg \
    --conf spark.sql.catalog.nessie.prefix=main \
    --conf spark.sql.catalog.nessie.io-impl=org.apache.iceberg.io.ResolvingFileIO \
    --conf spark.sql.catalog.nessie.credential=nessie-client:<client-secret> 
```

The `use nessie;` statement should succeed. This means that both the `/oauth/tokens` and the `/config` endpoints are working.

Next steps: make the `/api` endpoint work with authentication. It will be tricky because we need to inject the bearer token from the Iceberg client request into the `NessieApiV2` instance.


rest-catalog/rest-catalog-service/src/main/java/org/projectnessie/restcatalog/service/resources/IcebergV1ApiResource.java
```diff
@@ -97,6 +98,7 @@ public class IcebergV1ApiResource extends BaseIcebergResource implements Iceberg
 
   @Override
   public ListNamespacesResponse listNamespaces(String prefix, String parent) throws IOException {
+    ThreadLocalBearerAuthentication.setAuthHeaderValue(authorization);
```
snazy, line 101: Is there a way to avoid having these calls in every endpoint? If we miss one endpoint, it could use the authorization of another user.

docker/authn/docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
   nessie:
     image: ghcr.io/projectnessie/nessie:latest
     ports:
-      - 19120:19120
+      - "19120:19120"
```
snazy, line 22: Should the changes in this file be on `main`?

adutra: Yes I will cherry-pick these changes and create a PR.

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/auth/DynamicBearerAuthenticationFilter.java
```diff
@@ -0,0 +1,50 @@
...
+   * class.
+   *
+   * <p>This injection point does not fall in that category; it is therefore honored only when the
+   * JAX-RS implementation takes care of that. This is the case for RestEasy Reactive, because a
+   * bridge between JAX-RS and CDI/Arc has been implemented; it will NOT be honored by other JAX-RS
+   * impls, like RestEasy Classic. See <a
+   * href="https://github.com/quarkusio/quarkus/issues/10496#issuecomment-887468525">this
+   * comment</a> for context.
+   */
+  @Context HttpHeaders headers;
```
adutra, line 41: @snazy this fixes the thread-local problem.

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/json/IcebergObjectMapperProducer.java
```diff
@@ -0,0 +1,43 @@
...
+   * ContextResolver @Provider-annotated classes are a JAX-RS mechanism, and sometimes they are not
+   * picked up by the CDI container, for example in {@link
+   * io.quarkus.resteasy.reactive.jackson.runtime.serialisers.ServerJacksonMessageBodyReader}. So we
+   * also produce the mapper here, to make sure it is available everywhere.
+   *
+   * <p>The bean produced here will override the default bean produced by {@link
+   * io.quarkus.jackson.runtime.ObjectMapperProducer}.
+   */
+  @Singleton
+  public ObjectMapper produceObjectMapper() {
```
adutra, line 40: I am not sure if this has become necessary because of the switch to resteasy-reactive, or if things were already failing before. But without this bean, the default ObjectMapper would be used to deserialize requests and the objects would be empty.

rest-catalog/quarkus-rest-catalog/src/main/resources/application.properties
```diff
@@ -52,18 +52,20 @@ quarkus.log.category."io.quarkus.http.access-log".level=${HTTP_ACCESS_LOG_LEVEL:
 quarkus.http.port=19130
 quarkus.http.test-port=0
 quarkus.http.access-log.enabled=true
-# fixed at buildtime
-quarkus.resteasy.path=/
-quarkus.resteasy.gzip.enabled=true
```
adutra, line 57: resteasy-reactive does not have an equivalent property, but rather relies on `quarkus.http.enable-compression`. I manually verified that responses are being compressed if the user agent includes an `Accept-Encoding: gzip` header in the request.

rest-catalog/quarkus-rest-catalog/build.gradle.kts
```diff
@@ -59,10 +59,8 @@ dependencies {
 
   implementation(enforcedPlatform(libs.quarkus.bom))
   implementation(enforcedPlatform(libs.quarkus.amazon.services.bom))
-  implementation("org.jboss.resteasy:resteasy-core-spi")
-  implementation("io.quarkus:quarkus-resteasy")
-  implementation("io.quarkus:quarkus-resteasy-jackson")
-  implementation("io.quarkus:quarkus-reactive-routes")
+  implementation("io.quarkus:quarkus-resteasy-reactive")
```
adutra, line 62: As stated elsewhere the switch to resteasy-reactive allows to inject request context anywhere, not only on resource classes. So far I haven't seen any inconvenience or regression, apart maybe from the fact that the Iceberg ObjectMapper bean now must be explicitly produced.

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/ITNessieRestCatalogQuarkus.java
```diff
@@ -100,10 +100,6 @@ protected RESTCatalog catalog() {
               uri(),
               CatalogProperties.FILE_IO_IMPL,
               "org.apache.iceberg.io.ResolvingFileIO",
-              OAuth2Properties.TOKEN,
```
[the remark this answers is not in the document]
snazy: okay

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/auth/DynamicBearerAuthenticationFilter.java
snazy: okay

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/json/IcebergObjectMapperProducer.java
snazy: This works for me

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/NessieApiProvider.java
```diff
@@ -47,10 +50,12 @@ public NessieApiV2 produceNessieApi() {
     return clientBuilder
         .fromSystemProperties()
         .fromConfig(cfg -> config.nessieClientConfig().get(cfg))
+        // FIXME: should we allow to configure the authentication method?
```
snazy, line 53: I suspect, we'll need OAuth here, but this state is fine for me at the moment.

rest-catalog/quarkus-rest-catalog/src/main/resources/application.properties
snazy: okay

snazy (commented):
I've updated your branch to see whether CI is happy now

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/NessieApiProvider.java
adutra: Yes, the other way to do this is to configure an auth provider via config, for example the OAuth2 provider. I'm not sure it's better because this means that we'll have different credentials: one for Iceberg Client -> Iceberg REST Server communication, and another one for Iceberg REST Server -> Nessie Server. But I think we should let users decide, so I propose the following:

1. If an auth provider is provided via config, use it;
2. Otherwise, use the injected bean. By default, it should be `DynamicBearerAuthentication`, but if some other bean is present, use it.

adutra:
This is not ready as in "production-ready", but I think it is stable enough for a review. 

I will summarize here a few pain points regarding how the Iceberg client interacts with the `/oauth/tokens` endpoint. I am not sure if these should be considered defects, or if they happen just because I'm using Keycloak with default values everywhere. I have no idea if other backends would work better, of if the REST Catalog server should handle the requests in a different way. 

In fact, there is no formal specification for this endpoint, and `CatalogTests`, which serve as a TCK for the REST Catalog in general, does not exercise authentication at all, to my best knowledge. We are thus left to guess what's the desired behavior.

Anyways, here are the items I'd like to stress:

1. Iceberg ignores the (optional) `refresh_token` field from a `client_credentials` response, and so cannot
make use of it to refresh the access token using the method described in [RFC 6749 section 6](https://www.rfc-editor.org/rfc/rfc6749#section-6).
2. When refreshing tokens, a token exchange request is sent according to RFC 7523, but since this request follows an initial `client_credentials` request, there is no `issued_token_type` field available,
which causes an error [here](https://github.com/apache/iceberg/blob/c07fa0f9c208ecb2e0824b63ebd1ffc9e260f2bb/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java#L213-L214). The stack trace is:
   ```
    23/06/14 19:07:16 WARN Tasks: Retrying task after failure: Invalid token type: null
    java.lang.IllegalArgumentException: Invalid token type: null
            at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:220)
            at org.apache.iceberg.rest.auth.OAuth2Util.tokenExchangeRequest(OAuth2Util.java:213)
            at org.apache.iceberg.rest.auth.OAuth2Util.tokenExchangeRequest(OAuth2Util.java:204)
            at org.apache.iceberg.rest.auth.OAuth2Util.refreshToken(OAuth2Util.java:139)
            at org.apache.iceberg.rest.auth.OAuth2Util.access$000(OAuth2Util.java:55)
            at org.apache.iceberg.rest.auth.OAuth2Util$AuthSession.refreshCurrentToken(OAuth2Util.java:473)
            at org.apache.iceberg.rest.auth.OAuth2Util$AuthSession.lambda$refresh$1(OAuth2Util.java:447)
            at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413)
            at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:219)
            at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:203)
            at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:196)
            at org.apache.iceberg.rest.auth.OAuth2Util$AuthSession.refresh(OAuth2Util.java:447)
    ```
3. After 3 failed retries, Iceberg is supposed to fall back to a new `client_credentials` request according to the comment [here](https://github.com/apache/iceberg/blob/c07fa0f9c208ecb2e0824b63ebd1ffc9e260f2bb/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java#L436-L437), but this is not what is being done currently; instead, a fourth token exchange request is sent, which fails obviously, and prints the message [here](https://github.com/apache/iceberg/blob/715c9b90f537a215f38da5a8ddad19977222860b/core/src/main/java/org/apache/iceberg/util/Tasks.java#L289). After this, the whole periodic refresh cycle is interrupted.
4. Even if we "fake" the `issued_token_type` field – see [here](https://github.com/projectnessie/nessie/blob/59f7217ba4b124c3d386fcdd7869554f92c2bd03/rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/auth/OAuthUtils.java#L62-L64) – to fix items 2 and 3 above, the token exchange request will at least be sent, but it is still
going to receive a 401 response from Keycloak, probably because of the authentication method used: bearer token instead
of basic auth or credentials in the form data. According to RFC7523, basic auth or form data auth should be supported, other forms of authentication are left to the backend to figure out.  Keycloak, in any case, doesn't seem to like it. Its logs show:
   ```
   type=TOKEN_EXCHANGE_ERROR, realmId=a1ae31fb-813f-4003-8de0-d6983b391aff, clientId=null, userId=null, ipAddress=172.23.0.1, error=client_not_found, grant_type=urn:ietf:params:oauth:grant-type:token-exchange
   ```
   Hopefully, with the "fake" `issued_token_type` field, the refresh cycle fails but is not interrupted, so after 3 failures, eventually a token exchange request with basic auth is fired, and that one succeeds.

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/auth/OAuthUtils.java
```diff
@@ -0,0 +1,119 @@
...
+    String error = json.getString(ERROR);
+    String errorDescription = json.getString(ERROR_DESCRIPTION);
+    if (error == null) {
+      error = "UnknownOAuthTokenEndpointError";
+      errorDescription = json.encode();
+    }
+    return new OAuthTokenEndpointException(statusCode, error, errorDescription);
+  }
+
+  public static void printOAuthTokenRequest(OAuthTokenRequest req) {
```
snazy, line 89: Isn't it unsafe to dump sensitive information?

snazy (commented):
Your comments on the Iceberg REST clients are all great! Can you add those to #7050? We might also want to provide a solution for that though.

The PR looks good to me - just one concern about logging sensitive information. I think with that addressed, we can go ahead and merge it to the feature branch.

rest-catalog/quarkus-rest-catalog/src/main/java/org/projectnessie/restcatalog/server/auth/OAuthUtils.java
adutra: Yes this is VERY unsafe :-D This was useful for me while developing this PR but I think we should simply get rid of it now.

snazy (approved):

PR #7045, opened 2023-06-14
New storage/references logic: prevent one database read

snazy:
`VersionStoreImpl.getNamedRef` indirectly calls `ReferenceLogicImpl.getReferences()` to resolve a reference.

Since references have a logic to keep references consistent, it has to consult the `int/refs` reference. Currently that `getReferences()` implementation performs a `Persist.fetchReferences()` for the user-provided references and then a `Persist.fetchReference()` for the `int/refs` reference.

This change combines `fetchReference("int/refs")` into the `fetchReferences()` invocation, which saves one DB round trip, leveraging database bulk-fetch operations.


snazy:
For posterity:

before this change:
image

with this change:
image

adutra (approved):

PR #7056, opened 2023-06-15
Make number of access checks configurable

snazy:
The hard coded default was `10`. That number is now configurable, with a new conservative default value of `100`. Operators are responsible to adjust this value to the capabilities and restrictions of the authz implementation.

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/config/ServerConfig.java
+++ b/servers/services/src/main/java/org/projectnessie/services/config/ServerConfig.java
@@ -18,6 +18,8 @@ package org.projectnessie.services.config;
 /** Nessie server configuration to be injected into the JAX-RS application. */
 public interface ServerConfig {
 
+  int DEFAULT_ACCESS_CHECK_BATCH_SIZE = 100;
+
   /**
    * Gets the branch to use if not provided by the user.
    *
@@ -31,4 +33,8 @@ public interface ServerConfig {
    * @return {@code true} if the server should send the stack trace to the client.
    */
   boolean sendStacktraceToClient();
+
+  default int accessChecksBatchSize() {
+    return DEFAULT_ACCESS_CHECK_BATCH_SIZE;
+  }
 }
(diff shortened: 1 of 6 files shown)


servers/services/src/main/java/org/projectnessie/services/config/ServerConfig.java
```diff
@@ -31,4 +33,8 @@ public interface ServerConfig {
    * @return {@code true} if the server should send the stack trace to the client.
    */
   boolean sendStacktraceToClient();
+
+  default int accessChecksBatchSize() {
```
XN137, line 37: just wondering: why are the other defaults not part of the interface?

site/docs/try/configuration.md
```diff
@@ -19,10 +19,11 @@ docker run \
...
+| `nessie.server.send-stacktrace-to-client` | `false`        | `boolean` | Sets if server stack trace should be sent to the client in case of error.                                                                                                                                                                                                                                                                         |
+| `nessie.server.access-checks-batch-size`  | `100`          | `int`     | The number of entity-checks that are grouped into a call to `BatchAccessChecker`. The default value is quite conservative, it is the responsibility of the operator to adjust this value according to the capabilities of the actual authz implementation. Note that the number of checks can be exceeded depending on the context of the checks. |
```
XN137, line 26: just to confirm: the last sentence means there might be more elements in a batch than configured?

XN137 (approved):

servers/services/src/main/java/org/projectnessie/services/config/ServerConfig.java
snazy: Good question - IDK - those were actually there before me

site/docs/try/configuration.md
snazy: Yes

snazy: A few, not "tons"

PR #7058, opened 2023-06-15
Fix Newer-Java + Mac/Win CI

snazy:
This PR does not fix all workflow at the moment, but at least brings some back into a good/better state.


XN137 (approved):
CI for newer java seems to fail on 20 but maybe thats expected as per PR description

PR #7059, opened 2023-06-15
Micro-optimization of `LazyIndexImpl.get()`

snazy:
Eliminates an `instanceof` in the usually executed code path.

### DIFF
--- a/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
+++ b/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
@@ -30,38 +30,40 @@ public final class SupplyOnce {
   }
 
   private static final class NonLockingSupplyOnce<T> implements Supplier<T> {
-    private final Supplier<T> loader;
+    private int loaded;
     private Object result;
-    private boolean loaded;
+    private final Supplier<T> loader;
 
     private NonLockingSupplyOnce(Supplier<T> loader) {
       this.loader = loader;
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public T get() {
-      if (loaded) {
-        return eval(result);
+      switch (loaded) {
+        case 1:
+          return (T) result;
+        case 2:
+          throw (RuntimeException) result;
+        case 0:
+          return load();
+        default:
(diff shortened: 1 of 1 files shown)


adutra (dismissed):

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
```diff
@@ -41,27 +42,28 @@ private NonLockingSupplyOnce(Supplier<T> loader) {
...
-    private T eval(Object i) {
-      if (i instanceof RuntimeException) {
-        throw (RuntimeException) i;
+    private T eval() {
+      Object r = result;
+      if (r != null) {
+        return (T) r;
       }
-      return (T) i;
+      throw ex;
```
dimas-b, line 66: I'm not sure the order of assignments between `results`, `ex` and `loaded` is actually guaranteed to be the the same in another thread :thinking:

snazy: This thing is not thread safe and not used in such a case

adutra: There wasn't any synchronization in this class, so I am assuming that it is not meant to be thread-safe. Even the previous version would have had issues if `get()` is accessed concurrently.

snazy:
Ugh - forgot that it's meant to lazy-load even `null` values.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/SupplyOnce.java
```diff
@@ -30,38 +30,40 @@ public static <T> Supplier<T> memoize(Supplier<T> loader) {
   }
 
   private static final class NonLockingSupplyOnce<T> implements Supplier<T> {
-    private final Supplier<T> loader;
+    private int loaded;
     private Object result;
```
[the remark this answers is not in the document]
snazy: Let's keep the object small. Exceptions are an exception ;)

adutra (approved):

PR #7067, opened 2023-06-16
Fix reference-handling concurrency annoyances

snazy:
In a scenario when references are created/deleted and concurrently listed, so that reference-creation/deletion-recovery kicks in, there are a few scenarios which can cause unnecessary but rather annoying user-facing errors (exceptions).

This has been exposed (and later verified) using the new test `AbstractReferenceLogicTests.refCreationDeletionWithConcurrentRefsListing()` (ran with up to 2000 repetitions locally, committed in this PR with 2 repetitions).

### DIFF
--- a/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractReferenceLogicTests.java
+++ b/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractReferenceLogicTests.java
@@ -16,8 +16,11 @@
 package org.projectnessie.versioned.storage.commontests;
 
 import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.Thread.currentThread;
+import static java.lang.Thread.sleep;
 import static java.util.Arrays.asList;
 import static java.util.Collections.singletonList;
+import static java.util.concurrent.Executors.newSingleThreadExecutor;
 import static java.util.stream.IntStream.rangeClosed;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
 import static org.mockito.ArgumentMatchers.any;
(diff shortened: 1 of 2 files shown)


versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractReferenceLogicTests.java
```diff
@@ -76,6 +85,59 @@ public class AbstractReferenceLogicTests {
...
+                    break;
+                  }
+                }
+              });
+      try {
+        for (int i = 0; i < num; i++) {
+          refLogic.createReference(refName.apply(i), EMPTY_OBJ_ID);
+          sleep(1);
+        }
+        Collections.shuffle(indexes);
```
dimas-b, line 124: Do we have to have a long list of references for this test? Why not operate on one reference? I suppose the test aims to cause a read/delete collision on one particular reference, so having a smaller list should actually increase the likelihood of collisions, I guess :thinking:

dimas-b (approved):
The code fix LGTM. One comment about the test.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractReferenceLogicTests.java
snazy: Yea, that's actually correct. I used this test originally to reproduce the issue reported on the Nessie chat.

snazy:
Hm - found another annoying behavior. Will try to fix it first and then put this one back into "Ready for review".

snazy:
Will open another PR with a different/bigger scope to fix the underlying issues.

PR #7088, opened 2023-06-20
Enhance references with created-timestamp + extended-info-obj-pointer

snazy:
The intent of this change is to provide the base to fix an annoying race
where users can failures for reference creations or deletions due to a
potential race with concurrent reference queries, which trigger
reference recovery, which can legitly finish a
half-reference-creation/deletion, in these cases the original reference
creation/deletion.

This race is legit and intended, the errors should not happen though,
and can be prevented by having some form of "token" - in this case the
reference creation timestamp, which this change duplicates in the `refs`
table.

Since this change already changes all the references persistence, it
also adds a new `extendedInfoObj` attribute to references. The intent
here is to remove the (not yet) ambiguous pointer to "commt pointers"
(extended tags via `TagObj`, not exposed as a functionality yet). The
goal here is to only store the HEAD commit ID in `Reference.pointer()`
and if a reference has more information (for example that `TagObj`),
store the pointer to that one in `extendedInfoObj`. This also saves one
additional load for `CommitObjReference` implementations (like
`TagObj`) and to remove that interface.


versioned/storage/common-proto/src/main/proto/org/projectnessie/versioned/storage/common/proto/storage.proto
```diff
@@ -90,7 +91,6 @@ enum CompressionProto {
 }
 
 message TagProto {
-  bytes commit_id = 1;
```
adutra, line 93: Isn't this a breaking change? How can we deserialize an existing tag that has this field?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/Hashes.java
```diff
@@ -80,9 +80,8 @@ static ObjId refHash(String name, ObjId initialPointer, long createdAtMicros) {
     return hashAsObjId(hasher);
   }
 
-  static ObjId tagHash(
-      ObjId commitId, String message, CommitHeaders headers, ByteString signature) {
-    Hasher hasher = newHasher().putString(TAG.name(), UTF_8).putBytes(commitId.asByteArray());
+  static ObjId tagHash(String message, CommitHeaders headers, ByteString signature) {
+    Hasher hasher = newHasher().putString(TAG.name(), UTF_8);
```
adutra, line 84: Just curious: is there any consequence to the fact that the hash of tag objects has changed? E.g. if we recompute the hash of an existing tag that was hashed with the previous code, we won't find the same hash.

versioned/storage/common-proto/src/main/proto/org/projectnessie/versioned/storage/common/proto/storage.proto
adutra: Also you might want to add `reserved 1;` and document why the field is gone.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraConstants.java
```diff
@@ -33,11 +33,11 @@ final class CassandraConstants {
 
   static final String COLS_COMMIT =
       "c_created, c_seq, c_message, c_headers, c_reference_index, c_reference_index_stripes, c_tail, c_secondary_parents, c_incremental_index, c_incomplete_index, c_commit_type";
-  static final String COLS_REF = "r_name, r_initial_pointer, r_created_at";
+  static final String COLS_REF = "r_name, r_initial_pointer, r_created_at, r_extended_info";
```
adutra, line 36: Just curious: what's our global stance on DB schema evolution? Here, all pre-existing rows won't have any info in this new column. Also, `t_commit_id` column has been virtually deleted, but would still be physically present in the database. Is the suggestion to do a full export-upgrade-import?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraConstants.java
```diff
@@ -101,6 +103,10 @@ final class CassandraConstants {
           + COL_REFS_POINTER
           + "=? AND "
           + COL_REFS_DELETED
+          + "=? AND "
+          + COL_REFS_CREATED_AT
```
adutra, line 107: Have you considered the (admittedly contrived) case where 2 distinct updates could succeed if they have the same timestamp, because they were generated by 2 distinct Nessie instances? Wouldn't it be safer to use some random thing, maybe a UUID?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Reference.java
```diff
@@ -43,14 +44,23 @@ public interface Reference {
   @Value.Parameter(order = 4)
   boolean deleted();
 
+  @Value.Parameter(order = 5)
+  long createdAtMicros();
+
+  @Value.Parameter(order = 6)
+  @Nullable
+  @jakarta.annotation.Nullable
+  ObjId extendedInfoObj();
```
adutra, line 53: Have you considered that future reference types could have more than one extended object id?

adutra (commented):
Changes look mostly good to me, although the schema and serialization changes make me a little nervous :-)

versioned/storage/common-proto/src/main/proto/org/projectnessie/versioned/storage/common/proto/storage.proto
snazy: The tag-objects have never been used. So this breaking change is fine.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/Hashes.java
snazy: The tag-objects have never been used. So this breaking change is fine.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraConstants.java
snazy: The backends tell users about the difference on startup, with the schema to apply.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Reference.java
snazy: No - why would you have that?

versioned/storage/common-proto/src/main/proto/org/projectnessie/versioned/storage/common/proto/storage.proto
adutra: OK. In this case you could even renumber the fields starting from 1.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Reference.java
adutra: I don't have concrete examples in mind. I was just wondering if it wouldn't be more future-proof to plan for that right now.

adutra (approved):

PR #7093, opened 2023-06-20
Mongo: unify condition handling

snazy:
no functional change

### DIFF
--- a/versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
+++ b/versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
@@ -110,6 +110,7 @@ import javax.annotation.Nonnull;
 import org.agrona.collections.Hashing;
 import org.agrona.collections.Object2IntHashMap;
 import org.bson.Document;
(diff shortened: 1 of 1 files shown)


versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -271,6 +265,15 @@ public Reference markReferenceAsDeleted(@Nonnull @jakarta.annotation.Nonnull Ref
     return reference.withDeleted(true);
   }
 
+  private Bson referenceCondition(Reference reference) {
+    List<Bson> filters = new ArrayList<>(5);
```
adutra, line 269:
```suggestion
    List<Bson> filters = new ArrayList<>(3);
```

adutra (approved):

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
snazy: That number will be correct in the future ;)

PR #7097, opened 2023-06-20
Enable authentication in REST Catalog server ITs

adutra:
Note for reviewers:

* This PR introduces tests in `:nessie-rest-catalog-server` where authentication is enabled in two different ways: by propagating the Iceberg client's token to Nessie, and by obtaining different tokens from those used by Iceberg.
* The tests involve two test resources: Keycloak and Nessie. The tricky part is to make talk to the same Keycloak and obtain identical tokens all the moving parts: the iceberg client, the test client used to initialize the repo, the REST Catalog server, and Nessie.
* This PR contains a significant refactoring of `KeycloakTestResourceLifecycleManager`, but it remains backwards-compatible. The class javadocs explain the biggest changes.
* As discussed in Zulip, it wasn't possible to use Nessie's AppRunner plugin. For now, I'm proposing a new test resource called `NessieDockerTestResourceLifecycleManager`.
  *  As the name implies, it runs Nessie on Docker and will not run the exact same version that is in tree. I think that this is somewhat irrelevant here, as in this module we are developing the REST Catalog server, and Nessie could arguably be considered an external dependency. 
  * I *think* it should be possible to turn this into a test resource that runs the in-tree quarkus jar instead, but for now I think what we have here is enough. Could be done later though.
  * For now, this new test resource resides in `:nessie-rest-catalog-server` and is not available anywhere else. But we could maybe move it to `:nessie-quarkus-tests` later, since that is where other test resources are declared.


rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/resources/NessieDockerTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,241 @@
...
+      NessieContainer nessie = NessieDockerTestResourceLifecycleManager.nessie;
+      if (nessie != null) {
+        nessie.stop();
+      }
+    } finally {
+      NessieDockerTestResourceLifecycleManager.nessie = null;
+    }
+  }
+
+  public class NessieContainer extends GenericContainer<NessieContainer> {
```
snazy, line 157: Should this provide a `StartupCheckStrategy`?
E.g. trying a dummy HTTP request against the `/` URI path (aka index.html).

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/BaseNessieRestCatalogTests.java
```diff
@@ -0,0 +1,179 @@
...
+              requireNonNull(controlClient.getConfig().getDefaultBranch())));
+      return catalog;
+    }
+  }
+
+  private String restCatalogServerUri() {
+    return format(
+        "http://127.0.0.1:%d/iceberg/",
+        Objects.requireNonNull(
+            Integer.getInteger("quarkus.http.test-port"),
```
snazy, line 141: Hm. Where does this property come from, now that the apprunner-plugin's not used.

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/resources/NessieDockerTestResourceLifecycleManager.java
adutra: We already have a WaitStrategy:

```java
waitingFor(Wait.forLogMessage(".*Nessie.*started.*Listening on.*", 1));
```

In which cases would this not be sufficient? I could see one case: if the OIDC backend is unreachable, then the Nessie container would start normally, but tests would fail. But OTOH it's non trivial to create a `StartupCheckStrategy` for that, we would need to ping a protected endpoint like `/api/v2/config` and check that we get a 200 back. Do you think it's worth the effort?

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/BaseNessieRestCatalogTests.java
adutra: It's a random port assigned by Quarkus. The property is declared in the `application.properties` file in this project:

```
quarkus.http.test-port=0
```

My understanding is that when the port is set to zero, a random port is assigned, then Quarkus sets the effective port as a system property for other components to consume it. See here:

https://github.com/quarkusio/quarkus/blob/main/test-framework/common/src/main/java/io/quarkus/test/common/LauncherUtil.java#L243

adutra: I think we can simplify this. It's totally fine imho to use a fixed port, e.g. 19131. I will try.

adutra: Nope. Fixed ports seem to pose conflict problems. Let's keep the random port.

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/BaseNessieRestCatalogTests.java
```diff
@@ -0,0 +1,179 @@
...
+  protected RESTCatalog catalog() {
+    try (NessieApiV2 controlClient = buildControlClient()) {
+      RESTCatalog catalog = new RESTCatalog();
+      catalog.setConf(new Configuration());
+      catalog.initialize(
+          "nessie-iceberg-rest",
+          ImmutableMap.of(
+              CatalogProperties.URI,
+              restCatalogServerUri(),
+              CatalogProperties.FILE_IO_IMPL,
```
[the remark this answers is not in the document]
adutra: In fact, a plain old HashMap seems even more readable, and saves lines. I will use that.

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/KeycloakTestResourceLifecycleManager.java
```diff
@@ -64,106 +81,209 @@
 @SuppressWarnings("SameParameterValue")
 public class KeycloakTestResourceLifecycleManager
```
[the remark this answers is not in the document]
adutra: Sure: https://github.com/projectnessie/nessie/pull/7101

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/resources/NessieDockerTestResourceLifecycleManager.java
snazy: Sorry - missed that one.

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/BaseNessieRestCatalogTests.java
snazy: Ah, okay. That was something the apprunner-plugin did in the past. All good.

snazy: Random port are better for testing IMO.

rest-catalog/quarkus-rest-catalog/src/intTest/java/org/projectnessie/restcatalog/server/resources/NessieDockerTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,241 @@
...
+  private Map<String, String> extraEnvVars;
+
+  @Override
+  public void setIntegrationTestContext(DevServicesContext context) {
+    this.context = context;
+  }
+
+  @Override
+  public void init(Map<String, String> initArgs) {
+    dockerImage = initArgs.get(NESSIE_DOCKER_IMAGE);
```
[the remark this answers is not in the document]
snazy: ping ;)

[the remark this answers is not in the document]
adutra: Whoops

PR #7100, opened 2023-06-21
Fix annoying ref race

snazy:


versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractReferenceLogicTests.java
```diff
@@ -83,6 +93,106 @@ protected AbstractReferenceLogicTests(Class<?> surroundingTestClass) {
...
   }
 
+  @SuppressWarnings("BusyWait")
+  @RepeatedTest(2)
+  public void refCreationDeletionWithConcurrentRefsListing() throws Exception {
+    assumeThat(surroundingTestClass.getSimpleName())
+        .isNotIn(
+            // Fails due to "write too old" errors, need to implement logic "everywhere" to handle
+            // `DatabaseSpecific.isRetryTransaction()`.
+            "ITCockroachDBPersist",
```
adutra, line 103: Do we have issues to track these?

snazy: No, but those are the reasons why we have not declared those as "production" on the web site.
For CR, it's rather a concurrency issue on a single row. For C*, I'm not sure at all - LWTs are quite fundamental for us - so I am not sure whether we shall declare C* as production-ready-for-Nessie at all.

snazy: BTW: C* bugs me every day with its driver-timeout-exceptions - no other DB does that

PR #7141, opened 2023-06-27
Fix inability to delete some references

snazy:
The new storage model's commit logic, which provides support for name indexes, is also used to provide the index of references. When a reference is about to be deleted, the logic checks whether the reference to be deleted is in that index. Unfortunately, `ReferenceLogicImpl.commitDeleteReference()` only considered the _incremental_ index and not the "spilled out" reference stripes, so that a reference that is not in the incremental index cannot be deleted. The fix is a simple one-line change to load the whole index and not just the incremental one.

### DIFF
--- a/versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestReferenceLogicImpl.java
+++ b/versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestReferenceLogicImpl.java
@@ -17,21 +17,29 @@ package org.projectnessie.versioned.storage.common.logic;
 
 import static com.google.common.collect.Lists.newArrayList;
 import static java.util.Collections.singletonList;
+import static java.util.Objects.requireNonNull;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 import static org.mockito.Mockito.spy;
 import static org.projectnessie.versioned.storage.common.config.StoreConfig.CONFIG_COMMIT_TIMEOUT_MILLIS;
+import static org.projectnessie.versioned.storage.common.config.StoreConfig.CONFIG_MAX_INCREMENTAL_INDEX_SIZE;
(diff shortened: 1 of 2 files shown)


versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestReferenceLogicImpl.java
```diff
@@ -77,6 +86,51 @@ public TestReferenceLogicImpl() {
...
+
+    for (int i = 0; i < num; i++) {
+      refLogic.createReference(refName.apply(i), EMPTY_OBJ_ID, null);
+    }
+
+    Reference refRefs = requireNonNull(persist.fetchReference(REF_REFS.name()));
+    CommitObj refRefsCommit = requireNonNull(commitLogic(persist).fetchCommit(refRefs.pointer()));
+    soft.assertThat(refRefsCommit.referenceIndexStripes())
+        .describedAs(
+            "This test requires must exercise against a striped reference-index, adjust the 'num' parameter.")
```
adutra, line 110: Nit: change to "This test must exercise..."

versioned/storage/common/src/test/java/org/projectnessie/versioned/storage/common/logic/TestReferenceLogicImpl.java
```diff
@@ -77,6 +86,51 @@ public TestReferenceLogicImpl() {
...
+        .isNotEmpty();
+
+    ArrayList<Reference> created = newArrayList(refLogic.queryReferences(referencesQuery()));
+    soft.assertThat(created)
+        .containsAll(before)
+        .map(Reference::name)
+        .containsAll(IntStream.range(0, num).mapToObj(refName).collect(Collectors.toList()))
+        .hasSize(num + 1);
+
+    refLogic.queryReferences(referencesQuery()).forEachRemaining(ref -> {});
```
adutra, line 120: Why this line?

adutra (dismissed):

adutra (approved):

PR #7150, opened 2023-06-28
Use WorkerThread if parallelism is one.

nk1506:
For a single parallelism, it doesn't require spawning new threads.

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
@@ -67,10 +67,15 @@ public abstract class DefaultLocalExpire implements Expire {
     Instant started = clock().instant();
     expireParameters().liveContentSet().startExpireContents(started);
 
-    ForkJoinPool forkJoinPool = new ForkJoinPool(parallelism());
+    ForkJoinPool forkJoinPool = null;
     try {
-      DeleteSummary deleteSummary =
-          forkJoinPool.invoke(ForkJoinTask.adapt(this::expireInForkJoinPool));
+      DeleteSummary deleteSummary;
+      if (parallelism() == 1) {
+        deleteSummary = this.expireContents();
+      } else {
+        forkJoinPool = new ForkJoinPool(parallelism());
+        deleteSummary = forkJoinPool.invoke(ForkJoinTask.adapt(this::expireContents));
+      }
       LOGGER.info(
           "live-set#{}: Expiry finished, took {}, deletion summary: {}.",
           expireParameters().liveContentSet().id(),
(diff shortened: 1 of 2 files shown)


gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
```diff
@@ -163,6 +163,11 @@ public UUID identifyLiveContents() {
       throw new IllegalStateException("identifyLiveContents() has already been called.");
     }
 
+    // If parallelism is one, use the existing thread.
+    if (parallelism() == 1) {
```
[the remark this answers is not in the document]
nk1506: Done

gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
```diff
@@ -67,6 +67,12 @@ public DeleteSummary expire() {
     Instant started = clock().instant();
     expireParameters().liveContentSet().startExpireContents(started);
 
+    if (parallelism() == 1) {
```
snazy, line 70: Looks like this can be unified with the try-finally below, primarily to not omit the logging.

nk1506: Done

gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
```diff
@@ -67,10 +67,15 @@ public DeleteSummary expire() {
...
     expireParameters().liveContentSet().startExpireContents(started);
 
-    ForkJoinPool forkJoinPool = new ForkJoinPool(parallelism());
+    ForkJoinPool forkJoinPool = null;
     try {
-      DeleteSummary deleteSummary =
-          forkJoinPool.invoke(ForkJoinTask.adapt(this::expireInForkJoinPool));
+      DeleteSummary deleteSummary;
+      if (parallelism() == 1) {
+        deleteSummary = this.expireInForkJoinPool();
```
ajantha-bhat, line 74: nit: the method name is weird now as it is not using the `ForkJoinPool` always. 
Can be called as `expireContents`

snazy (approved):

snazy:
Going to revert this change, because the code will then use the common FJP, which does not follow the parallelism configuration value, due to .parallel() usage here: [https://github.com/projectnessie/nessie/blob/2ef699bf997cc98d8e0c1b018d5eddc45a6a8[…]/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java](https://github.com/projectnessie/nessie/blob/2ef699bf997cc98d8e0c1b018d5eddc45a6a8e29/gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java#L96)

PR #7152, opened 2023-06-28
Revert "Use WorkerThread if parallelism is one."

snazy:
Reverts projectnessie/nessie#7150


XN137 (approved):
reason for revert:

we still call `Stream.parallel()` so using the common FJP (which does not use parallelism=1)

PR #7155, opened 2023-06-29
Avoid parallel streams If parallelism is one.

nk1506:
If parallelism is explicitly one, It should use sequential streams.

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/expire/local/DefaultLocalExpire.java
@@ -67,10 +67,16 @@ public abstract class DefaultLocalExpire implements Expire {
     Instant started = clock().instant();
     expireParameters().liveContentSet().startExpireContents(started);
 
-    ForkJoinPool forkJoinPool = new ForkJoinPool(parallelism());
+    ForkJoinPool forkJoinPool = null;
     try {
-      DeleteSummary deleteSummary =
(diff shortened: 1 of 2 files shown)


snazy (commented):
Both changed code blocks are rather long running tasks, usually tasks that have rather long (I/O-)wait times, so parallelism==1 would significantly slow those down. The overhead of parallelism==1 is negligible considering the time spent in these tasks.

Given that #7150 had to be reverted due to ending in using the default `ForkJoinPool`, because of "unguarded" `Stream.parallel()`, I'm not sure that a future change (introducing another `Stream.parallel()`) might not run into the same situation.

Thinking a bit more about this, can you maybe elaborate on the overall goals?

harshm-dev:
Trying to use it in Dremio. Overhead is not really the reason.
RequestContext (in the thread local) is lost once the control moves to ForkJoinPool, which is used during gRPC communication.

In general, the preference is to perform within the executor's thread as there building blocks there with the assumption; aforementioned could only be one of the reasons.

@nk1506 Is it possible to extend the checks wrt `Stream.parallel()` ?

nk1506:
Thanks @snazy  for reviewing. I am totally on the same page with [comment](https://github.com/projectnessie/nessie/pull/7155#pullrequestreview-1504942451. ) where adding another `Stream.parallel()` can end up with same problem. 

Although this changes was required for one of the engine where it is using threadLocal to set some context. Aftre Using `ForkJoinPool` , Engine was loosing the context for nessie rest calls. 

I am abandoning this change as we found a workaround.

snazy:
Note that there's a potential to enhance the GC code to use a custom ForkJoinThreadFactory, requires a PR though.

PR #7166, opened 2023-06-30
Spark-SQL-tests: Nessie client configuration + ignore committer

snazy:


integrations/spark-extensions-basetests/src/main/java/org/projectnessie/spark/extensions/SparkCommitLogEntry.java
```diff
@@ -64,7 +64,7 @@ String getSignedOffBy() {
   static SparkCommitLogEntry fromShowLog(Object[] row) {
     return ImmutableSparkCommitLogEntry.builder()
         .author((String) row[0])
-        .committer((String) row[1])
+        // omit the committer! .committer((String) row[1])
```
adutra, line 67: Why must we omit the committer? I see that `fromCommitMeta` also omits the committer... but not the commit time. This is a bit puzzling.

snazy: The tests are a bit ... strange. They compare the "locally" constructed `CommitMeta` (without the right committer) against the committer from the server, which has a committer when auth is enabled. However, that's not really the scope of the tests - that's why I've removed it here.

snazy: TBH, the tests need quite some love.

adutra (approved):

PR #7174, opened 2023-07-01
Make gc-base, gc-iceberg classes Java 8 compatible

ajantha-bhat:
GC modules can be used by engines to build its own distributed GC logic. Hence, it has to be java8 compatible.

### DIFF
--- a/build-logic/src/main/kotlin/nessie-conventions-iceberg8.gradle.kts
+++ b/build-logic/src/main/kotlin/nessie-conventions-iceberg8.gradle.kts
@@ -1,0 +1,31 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


snazy (changes requested):
TL;DR this is way too much, doing way more than necessary

ajantha-bhat:
> TL;DR this is way too much, doing way more than necessary

Goal was to make gc-base and gc-iceberg compatible with java 8. 
So, I changed `nessie-conventions-iceberg` to use java 8 and then started fixing its dependencies compatibility. 

May be you meant introduce a new profile `nessie-conventions-iceberg8` and use for only `gc-base` and `gc-iceberg` and its dependencies. I will give it a try.

ajantha-bhat:
> TL;DR this is way too much, doing way more than necessary

Updated.

snazy (approved):

PR #7184, opened 2023-07-05
Remove dead code

adutra:


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
```diff
@@ -101,19 +101,6 @@ private List<ShallowCommit> findMergeBases(ShallowCommit commitA, ShallowCommit
     // might be necessary to re-incarnate `removeRedundant`, if there is a situation in which it
```
snazy, line 101: Need to update this comment as well, at best linking to this PR (and the original PR maybe).

snazy (changes requested):

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
snazy: Oh, and update the PR title

adutra: My bad, I didn't notice this explanation. I'm fine leaving the commented-out code.

PR #7193, opened 2023-07-06
Forbid relative hashes in commits, merges and transplants

### LINKED ISSUE #7119: Inconsistent parsing of relative hash references in different API v2 endpoints
This works:
```
curl 'http://localhost:19120/api/v2/trees/main@1f7ec3eb98eef851c5e4348144ec13ce658673fabd155aefbd862be5cf07f88e~1/entries'
```
... but this:
```
curl -X 'POST' "http://localhost:19120/api/v2/trees/main@1f7ec3eb98eef851c5e4348144ec13ce658673fabd155aefbd862be5cf07f88e~1/history/commit" -d [...]
```
reports:
```

  "status" : 400,
  "reason" : "Bad Request",
  "message" : "Illegal hex character '~'",
  "errorCode" : "BAD_REQUEST",
  "serverStackTrace" : null
}
```

adutra:
Fixes #7119


api/model/src/test/java/org/projectnessie/model/TestValidation.java
```diff
@@ -250,4 +250,42 @@ void pathRefName(String rest, String ref, String hashOnRef, String relativeSpec)
     soft.assertThat(matcher.group(2)).isEqualTo(hashOnRef);
     soft.assertThat(matcher.group(3)).isEqualTo(relativeSpec);
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    ",false",
+    "'',false",
```
snazy, line 257: Huh? IIRC quotes should not be allowed here.

snazy (commented):
Generally LGTM - one comment about quotes passing _is-hash_ tests

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
```diff
@@ -855,8 +855,8 @@ MergeResponse mergeRefIntoBranch(
   @Override
   @POST
   @jakarta.ws.rs.POST
-  @Path("{branch}/history/commit")
-  @jakarta.ws.rs.Path("{branch}/history/commit")
+  @Path("{branch:" + REF_NAME_PATH_ELEMENT_REGEX + "}/history/commit")
+  @jakarta.ws.rs.Path("{branch:" + REF_NAME_PATH_ELEMENT_REGEX + "}/history/commit")
```
dimas-b, line 859: Would it make sense to do similar changes for merge and transplant endpoints?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieRest.java
```diff
@@ -634,4 +643,25 @@ public void paramConverterInvalidValue() {
...
+    "-*2021-04-07T14:42:25.534748Z",
+    "main*2021-04-07T14:42:25.534748Z",
+    "main@cafebabe*2021-04-07T14:42:25.534748Z",
+  })
+  public void commitWithRelativeHashesNotAllowed(String ref) {
+    ContentKey key1 = ContentKey.of("test", "Key");
+    IcebergTable table1 = IcebergTable.of("loc1", 1, 2, 3, 4);
+    NessieError error =
+        prepareCommitV2(ref, key1, table1, 2, true).statusCode(400).extract().as(NessieError.class);
+    assertThat(error.getMessage()).startsWith("Relative hash not allowed here");
```
dimas-b, line 665: from `curl` it looks like this:
```
$ curl -X 'POST' "http://localhost:19120/api/v2/trees/main@1f7ec3eb98eef851c5e4348144ec13ce658673fabd155aefbd862be5cf07f88e~1/history/commit" [...]

{
  "status" : 400,
  "reason" : "Bad Request",
  "message" : "Relative hash not allowed here: 1f7ec3eb98eef851c5e4348144ec13ce658673fabd155aefbd862be5cf07f88e~1",
  "errorCode" : "BAD_REQUEST",
  "serverStackTrace" : null
}
```
This is much better than before, but still "here" is a bit obscure, IMHO... Could we say something like `Relative hash not allowed in commit operations: 1f7ec3eb98eef851c5e4348144ec13ce658673fabd155aefbd862be5cf07f88e~1`?

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
snazy: For merge & transplant a relative hash doesn't really make any sense,

The old data model was relatively strict wrt `expectedHash` - unnecessarily actually.

It feels, that the merge-operation, as it exists today using a squashed commit based on the diff between the merge-base and the merge-source, does not need any expected-hash. The "expected state" and the "applied change" are implicit (present in the Nessie repo), not explicit (only the new value provided by the caller). Older implementations, that supported "merges w/ individual commits", the expected-hash was important, because it was effectively a transplant.

The implementation for the new data model doesn't use the expected-hash for merge nor for transplant. For the latter (transplant) that seems wrong though. @adutra do you have bandwidth to check whether the expected-hash is required for transplant (I think it is, but I might be wrong).

In any case, every committing operation that takes an expected-hash, should not allow relative commit IDs. Good catch!

api/model/src/test/java/org/projectnessie/model/TestValidation.java
adutra: This is testing an empty string. The line above is testing `null`.

snazy: Oh, right - CSV :facepalm:

snazy (approved):

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
adutra: From what I see, in the new model:

* **For all committing operations (commit, merge, transplant)**: if an expected-hash is provided, an expected `CommitObj` is loaded, if the object does not exist, an exception is thrown:

https://github.com/projectnessie/nessie/blob/795ba1da4cf630d40dca8b967c79c6ae2daa40eb/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java#L136-L142

* Other than that, the expected-hash itself is only used for reporting purposes (when creating error messages, and `MergeResult` instances).

* **For commits only**: the expected `CommitObj` is further used to load the so-called expected-index:

https://github.com/projectnessie/nessie/blob/c6388c3fd3120a23c8db1516369d1ebb478599da/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java#L117-L124

This index is then used to compute key conflicts.

* **For merges and transplants**: I do not see any further usage of expected-hash or the corresponding expected `CommitObj`.

adutra: > The implementation for the new data model doesn't use the expected-hash for merge nor for transplant. For the latter (transplant) that seems wrong though.

Why is it not wrong for merges? If you are merging into some branch, expecting its head to be a certain hash, shouldn't the system raise an error if the head of the branch has moved away?

snazy: With "not wrong" I meant that a diff-based squash-merge does not need it to work correctly.
Users _may_ want to only apply a squash, if the HEAD is exactly at that hash.

snazy:
> should I go ahead and forbid relative hashes in merges and transplants as well?

Hm, yea, if that can go into this PR, that would be great.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -42,38 +42,18 @@ public interface ApiDoc {
           + "at different times depending on the state of the system. Using the full 'name@hash' form is recommended "
           + "to avoid ambiguity.\n";
 
-  String BRANCH_DESCRIPTION =
+  String COMMIT_BRANCH_DESCRIPTION =
```
adutra, line 45: Renamed these constants for clarity.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -42,38 +42,18 @@ public interface ApiDoc {
...
       "A reference to a particular version of the contents tree (a point in history) on a branch.\n"
-          + "This reference can be specified in these forms:\n"
-          + "- \\- (literal minus character) - Identifies the HEAD of the default branch \n"
-          + "- name - Identifies the HEAD commit on the named branch\n"
+          + "This reference is specified in this form:\n"
```
adutra, line 47: I think I opened a can of works :-) In fact, even if the REST API accepts references without hashes, the Service layer does not. IOW the only form of reference that works is the full `name@hash` form.

adutra: Note that the client API allows the commit operation to be sent without hash:

```java
api()
    .commitMultipleOperations()
    .branch(Branch.of("main", null)) // null hash allowed here
    .operation(Put.of(a, ta))
    .commit();
```

This results in the following error:

```
Bad Request (HTTP/400): commitMultipleOperations.expectedHash: must not be null
org.projectnessie.error.NessieBadRequestException: Bad Request (HTTP/400): commitMultipleOperations.expectedHash: must not be null
	at org.projectnessie.error.ErrorCode.lambda$asException$1(ErrorCode.java:66)
	at java.base@11.0.18/java.util.Optional.map(Optional.java:265)
	at org.projectnessie.error.ErrorCode.asException(ErrorCode.java:66)
```

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -42,38 +42,18 @@ public interface ApiDoc {
...
-          + "If both 'name' and 'hash' are given, 'hash' must be reachable from the current HEAD of the branch.\n"
-          + "In this case 'hash' indicates the state of contents that should be used for validating incoming changes\n"
-          + "(commits / merges / transplants).\n"
-          + "\n"
-          + "Note that using the simple 'name' form will effectively disable content conflict checks and is "
-          + "generally discouraged.\n"
           + "\n"
-          + FULL_REF_INFO;
+          + "The 'hash' commit must be reachable from the current HEAD of the branch.\n"
+          + "In this case 'hash' indicates the state of contents that should be used for validating incoming changes.\n";
```
adutra, line 51: I kept the original sentence: `In this case 'hash' indicates the state of contents that should be used for validating incoming changes` – since this is indeed roughly what happens in `CommitImpl`.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -42,38 +42,18 @@ public interface ApiDoc {
...
           + "- name@hash - Identifies the 'hash' commit on the named branch.\n"
-          + "- -~3 - The 3rd predecessor commit from the HEAD of the default branch.\n"
-          + "- name~3 - The 3rd predecessor commit from the HEAD of a branch or tag.\n"
-          + "- name@hash^2 - The merge parent of the 'hash' commit of a branch or tag.\n"
-          + "- -*2021-04-07T14:42:25.534748Z - The predecessor commit closest to the HEAD of the default branch for the given ISO-8601 timestamp.\n"
-          + "- name*2021-04-07T14:42:25.534748Z - The predecessor commit closest to the HEAD of a branch or tag valid for the given ISO-8601 timestamp.\n"
-          + "- name*1685185847230 - The predecessor commit closest to the HEAD of a branch or tag valid for the given timestamp in milliseconds since epoch.\n"
           + "\n"
           + "The 'hash' commit must be reachable from the current HEAD of the branch.\n"
           + "In this case 'hash' indicates the state of contents known to the client and serves to ensure that the "
```
adutra, line 59: I also kept the original sentence here but to be fair, the new model does not appear to "ensure that the operation is performed on the contents that the client expects" – if the hashes differ, as long as the expected hash exists and is reachable, the operation will be carried out.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -575,6 +576,7 @@ public MergeResponse transplantCommitsIntoBranch(
     try {
       checkArgument(!hashesToTransplant.isEmpty(), "No hashes given to transplant.");
       validateCommitMeta(commitMeta);
+      validateNoRelativeSpec(expectedHash);
```
adutra, line 579: When we get here, hashes were already validated in `ImmutableMerge` and `ImmutableTransplant` objects (see their check methods) – but I figured it doesn't hurt to re-validate.

servers/jax-rs-tests/build.gradle.kts
```diff
@@ -53,6 +53,7 @@ dependencies {
   testImplementation(project(":nessie-versioned-storage-jdbc"))
 
   testImplementation(project(":nessie-jaxrs-testextension"))
+  testImplementation(libs.hibernate.validator.cdi)
```
adutra, line 56: Required, otherwise none of the `@NotNull` annotations and alike are actually enforced in this module.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
snazy: IIRC older versions allowed to commit without an expected hash. Probably a relict from that time.

snazy (commented):
LGTM now - waiting for @dimas-b now ;)

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
adutra: https://github.com/projectnessie/nessie/issues/7202

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
dimas-b: Can of worms... perhaps :) I believe `-` (default branch) should also be supported... the original idea was to make it easier for hand-written `curl` commands.

dimas-b (approved):
LGTM. Thanks!

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
adutra: In fact, you can use the dash... if you also provide the hash, e.g. `-@cafebabe` :-) Which, I suppose, defeats the initial purpose of the dash. But unfortunately any references without explicit hashes are currently refused (and were already before this PR).

adutra: For future reference, here is the status of `expectedHash` parameter across different layers and for different operations.

| Operation | REST layer (TreeResource) | Service layer (TreeService) | Version Store layer |
|--------|--------|--------|--------|
| Commit | Not required (path param), defaults to null | Required | Not required, defaults to HEAD |
| Merge | Required (request body) | Required | Not required, defaults to HEAD |
| Transplant | Required (request body) | Required | Not required, defaults to HEAD |

PR #7200, opened 2023-07-07
Separate Keycloak and Nessie test containers

snazy:
The Nessie testcontainer does not necessarily need the Keycloak container, but has a "hard" compile+runtime dependency to it. This change removes the implicit runtime dependency and also allows using a different Keycloak container, if needed.

### DIFF
--- a/servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/NessieDockerTestResourceLifecycleManager.java
+++ b/servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/NessieDockerTestResourceLifecycleManager.java
@@ -23,6 +23,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.net.URI;
 import java.util.Map;
+import org.projectnessie.testing.keycloak.CustomKeycloakContainer;
 import org.projectnessie.testing.nessie.NessieContainer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -57,7 +58,7 @@ public class NessieDockerTestResourceLifecycleManager
 
   private DevServicesContext context;
 
-  private NessieContainer.NessieConfig containerConfig;
+  private NessieContainer.NessieConfig.Builder containerConfig;
 
   @Override
   public void setIntegrationTestContext(DevServicesContext context) {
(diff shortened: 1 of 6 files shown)


testing/nessie-container/src/main/java/org/projectnessie/testing/nessie/NessieContainer.java
```diff
@@ -94,9 +102,24 @@ public interface Builder {
       @CanIgnoreReturnValue
       Builder authEnabled(boolean authEnabled);
 
+      default Builder oidcFromCustomKeycloakContainer(
+          CustomKeycloakContainer customKeycloakContainer) {
```
snazy, line 106: If I'm not mistaken, this should work even if callers do not have the `:nessie-keycloak-testcontainer` dependency.

adutra: I tried to create a project using the Nessie container without the Keycloak container, but the test failed with:

```
Caused by: java.lang.NoClassDefFoundError: org/projectnessie/testing/keycloak/CustomKeycloakContainer
	at org.projectnessie.testing.nessie.NessieContainer$NessieConfig$Builder.oidcFromCustomKeycloakContainer(NessieContainer.java:107)
	at org.projectnessie.quarkus.tests.profiles.NessieDockerTestResourceLifecycleManager.start(NessieDockerTestResourceLifecycleManager.java:77)
	at io.quarkus.test.common.TestResourceManager$TestResourceEntryRunnable.run(TestResourceManager.java:481)
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1736)
	... 3 more
Caused by: java.lang.ClassNotFoundException: org.projectnessie.testing.keycloak.CustomKeycloakContainer
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 7 more
```

That said, the test was using the Nessie Quarkus test resource, which may have a stronger dependency to Keycloak than the bare container. Although the error is being thrown in `org.projectnessie.testing.nessie.NessieContainer$NessieConfig$Builder.oidcFromCustomKeycloakContainer` so 🤷‍♂️ 

How about inverting the dependency graph? I.e., let Keycloak container depend on Nessie container and move `oidcFromCustomKeycloakContainer` to the Keycloak module? (I did not test it).

snazy: Damn - I thought, it would require the class once the method's used. But yea, my bad - that only works if a class is "first used" in a method body.

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/NessieDockerTestResourceLifecycleManager.java
```diff
@@ -66,17 +67,20 @@ public void setIntegrationTestContext(DevServicesContext context) {
...
-    context.containerNetworkId().ifPresent(configBuilder::dockerNetworkId);
-    containerConfig = configBuilder.build();
+    containerConfig = NessieContainer.builder().fromProperties(initArgs);
+    context.containerNetworkId().ifPresent(containerConfig::dockerNetworkId);
   }
 
   @Override
   public Map<String, String> start() {
-    nessie = containerConfig.createContainer();
+    CustomKeycloakContainer keycloak = KeycloakTestResourceLifecycleManager.getKeycloak();
```
adutra, line 76: If wonder if we could break the dependency here too, maybe have the 2 resources communicate through system properties... we can look into that later though.

testing/nessie-container/build.gradle.kts
```diff
@@ -20,7 +20,9 @@ extra["maven.name"] = "Nessie - Nessie testcontainer"
 
 dependencies {
   implementation(libs.slf4j.api)
-  api(project(":nessie-keycloak-testcontainer"))
+  implementation(libs.testcontainers.testcontainers)
+  // compileOnly, for non-custom-keycloak use cases
+  compileOnly(project(":nessie-keycloak-testcontainer"))
```
adutra, line 25: We can probably remove this now.

snazy: agree

adutra (approved):

PR #7223, opened 2023-07-12
IDE: Cleanup Idea excludes

snazy:


build-logic/src/main/kotlin/nessie-common-src.gradle.kts
```diff
@@ -141,32 +141,3 @@ plugins.withType<JavaPlugin>().configureEach {
     }
   }
 }
-
-// ////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-// IDE
-
-if (!System.getProperty("idea.sync.active").toBoolean()) {
```
adutra, line 148: Not required anymore?

snazy: Moved around

adutra (approved):

PR #7225, opened 2023-07-12
Allow re-adds in single commit

### LINKED ISSUE #7191: Allow re-add to happen in a single Nessie commit
We can distinguish three somehow similar but different scenarios using a pair of "delete"+"put" operations:
1. Rename - delete-op for an existing key, put-op (update) for another key using the content referenced by the delete-op
2. Re-add - delete-op for an existing key, put-op (new) for the same key but a new content
3. "unrelated drop + create" - delete-op for one key, put-op (new) for another key

Scenarios 1 + 3 are allowed to happen in a single Nessie commit.
Scenario 2 is actually prevented by the version-store implementations (both old + new model), so this must happen via two Nessie commits.

Technically, there's nothing that prevents scenario 2 to work inside a single commit - it's (by a brief look) just the validation that prevents scenario 2 to happen in a single Nessie commit.

Merge-squash operations are not affected.

adutra:
Fixes #7191 

For new storage only.


snazy (approved):
Looks really good! Thanks!
Feel free to address the nit or not.

snazy (approved):

PR #7227, opened 2023-07-12
Fix retrieval of default branch

snazy:


servers/services/src/main/java/org/projectnessie/services/impl/ConfigApiImpl.java
```diff
@@ -43,7 +43,7 @@ public NessieConfiguration getConfig() {
     }
     return ImmutableNessieConfiguration.builder()
         .from(NessieConfiguration.getBuiltInConfig())
-        .defaultBranch(this.config.getDefaultBranch())
+        .defaultBranch(defaultBranch)
```
snazy, line 46: Otherwise the retrieval above has no effect

dimas-b (approved):

PR #7232, opened 2023-07-13
Add `--author` option to `content-generator` commands

dimas-b:
* Applies to all commands that perform commits.

* Defaults to the `USER` env. var. (which was already the case in the `generate` command).

### DIFF
--- a/tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
+++ b/tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
@@ -20,7 +20,6 @@ import static org.projectnessie.model.ContentKey.fromPathString;
 import static org.projectnessie.tools.contentgenerator.keygen.KeyGenerator.newKeyGenerator;
 
 import java.time.Duration;
-import java.time.Instant;
 import java.time.LocalDateTime;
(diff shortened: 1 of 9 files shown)


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/GenerateContent.java
```diff
@@ -213,13 +213,13 @@ public void execute() throws BaseNessieClientServerException {
...
-                                "Commit #%d of %d on %s", commitNum, numCommits, branchName))
-                        .author(System.getProperty("user.name"))
-                        .authorTime(Instant.now())
-                        .build());
+                    withStandardCommitMetaFields(
+                        CommitMeta.builder()
+                            .message(
+                                String.format(
+                                    "Commit #%d of %d on %s", commitNum, numCommits, branchName))
+                            .authorTime(Instant.now())
```
snazy, line 221: Think it's fine to remove authorTime here, because it'll be set on the server side anyways.
This would make the `withStandardCommitMetaFields` take just a `String` for the message, saving one builder-from-build round trip - and makes the code easier.

dimas-b: updated

snazy (approved):

PR #7233, opened 2023-07-13
Add repository configuration objects

snazy:
Allows to retrieve and persist configuration objects via its API, via the REST v2 API using the `config/repository` endpoint. Read and/or write access to all or some types of repository configuration types can be restricted.

New storage is required for this functionality to work.

The Nessie server side must know the schema of each repository configuration type via an interface that extends `org.projectnessie.model.RepositoryConfig` and its type must be registered. This ensures that all repository configuration objects comply with the implicitly defined schema.

Creating and/or updating repository configurations is supposed to be a rare operation.

Nessie clients and servers need to know about the used repository config types via instances of `org.projectnessie.model.types.RepositoryConfigTypeBundle`. Instances of this interface are loaded via the standard Java services mechanism.

### DIFF
--- a/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/RepositoryConfigBackend.java
+++ b/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/RepositoryConfigBackend.java
@@ -1,0 +1,245 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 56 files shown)


snazy:
Left changes in separate commits to make reviewing easier.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/RepositoryConfigBackend.java
```diff
@@ -0,0 +1,245 @@
...
+import org.projectnessie.versioned.storage.common.objtypes.StringObj;
+import org.projectnessie.versioned.storage.common.persist.Obj;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.ObjType;
+import org.projectnessie.versioned.storage.common.persist.Persist;
+import org.projectnessie.versioned.storage.common.persist.Reference;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class RepositoryConfigBackend {
```
dimas-b, line 73: nit: Would `RepositoryConfigLogic` align with `CommitLogic` better? (just wondering)

dimas-b (approved):
LGTM :+1:

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/RepositoryConfigBackend.java
snazy: `*Logic` classes are in storage-common

PR #7265, opened 2023-07-19
Validate default-cut-off-policy with gc config/repository APIs

nk1506:
config/repository is accepting string as defaultCutOffPolicy. which is very vague. 
Validating this using regex with Duration/ISO Instant/Positive Numeric values.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Validation.java
+++ b/api/model/src/main/java/org/projectnessie/model/Validation.java
@@ -95,6 +95,24 @@ public final class Validation {
   public static final Pattern REF_NAME_PATH_ELEMENT_PATTERN =
       Pattern.compile(REF_NAME_PATH_ELEMENT_REGEX);
 
+  /*
+   * Default Cut-Off Policy can be a number or Duration or ISO instant.
+   * Following rules are to validate the same.
+   * */
+  private static final String DURATION_REGEX =
+      "([-+]?)P(?:([-+]?[0-9]+)D)?(T(?:([-+]?[0-9]+)H)?(?:([-+]?[0-9]+)M)?(?:([-+]?[0-9]+)(?:[.,]([0-9]{0,9}))?S)?)?";
+
+  private static final String ISO_TIME_REGEX =
+      "^(?:[1-9]\\d{3}-(?:(?:0[1-9]|1[0-2])-(?:0[1-9]|1\\d|2[0-8])|(?:0[13-9]|1[0-2])-(?:29|30)|(?:0[13578]|1[02])-31)|(?:[1-9]\\d(?:0[48]|[2468][048]|[13579][26])|(?:[2468][048]|[13579][26])00)-02-29)T(?:[01]\\d|2[0-3]):[0-5]\\d:[0-5]\\d(?:\\.\\d{1,9})?(?:Z|[+-][01]\\d:[0-5]\\d)$";
+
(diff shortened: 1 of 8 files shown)


api/model/src/main/java/org/projectnessie/model/Validation.java
```diff
@@ -95,6 +95,24 @@ public final class Validation {
   public static final Pattern REF_NAME_PATH_ELEMENT_PATTERN =
       Pattern.compile(REF_NAME_PATH_ELEMENT_REGEX);
 
+  /*
+   * Default Cut-Off Policy can be a number or Duration or ISO instant.
+   * Following rules are to validate the same.
+   * */
+  private static final String DURATION_REGEX =
```
ajantha-bhat, line 102: question: what about year and month?

snazy (approved):
Thanks!
LGTM!

api/model/src/main/java/org/projectnessie/model/Validation.java
```diff
@@ -95,6 +95,24 @@ public final class Validation {
   public static final Pattern REF_NAME_PATH_ELEMENT_PATTERN =
       Pattern.compile(REF_NAME_PATH_ELEMENT_REGEX);
 
+  /*
+   * Default Cut-Off Policy can be a number or Duration or ISO instant.
```
ajantha-bhat, line 99: this comment may be applicable for `DEFAULT_CUT_OFF_POLICY_REGEX` not `DURATION_REGEX`

api/model/src/main/java/org/projectnessie/model/Validation.java
```diff
@@ -196,6 +214,26 @@ public static String validateHash(String referenceName) throws IllegalArgumentEx
...
+    return matcher.matches();
+  }
+
+  /**
+   * Validate default cutoff policy.Policies can be one of: - number of commits as an integer value,
+   * - a duration (see java.time.Duration), - an ISO instant, - 'NONE,' means everything's
+   * considered as live
+   */
+  public static void validateDefaultCutOffPolicy(String value) {
+    if (isValidDefaultCutoffPolicy(value)) {
```
ajantha-bhat, line 230: I think we can inline `isValidDefaultCutoffPolicy` in this method and no need of private method.

ajantha-bhat:
oh, I didn't refresh the tab while reviewing. It was a race condition :D

api/model/src/main/java/org/projectnessie/model/Validation.java
ajantha-bhat: I might have confused with seeing example of string representation of `Period`. 
Duration doesn't have years and month
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#toString--

nk1506: We are using same format for nessie-gc . Also it seems Java Duration doesn't support others. 
The same pattern I took from Java class. 
`
 private static class Lazy {
        static final Pattern PATTERN =
            Pattern.compile("([-+]?)P(?:([-+]?[0-9]+)D)?" +
                    "(T(?:([-+]?[0-9]+)H)?(?:([-+]?[0-9]+)M)?(?:([-+]?[0-9]+)(?:[.,]([0-9]{0,9}))?S)?)?",
                    Pattern.CASE_INSENSITIVE);
    }
`

ajantha-bhat:
Question is answered and other comments are nit. So, No need a follow up. 

LGTM

Issue #7286, opened 2023-07-25
Helm Chart: Add support for custom PodLabels


c-thiel:
**Description**:
Currently the Helm chart does not offer a way to specify labels for the deployment.
https://github.com/projectnessie/nessie/blob/30d2db5916e5fb56c9c220dbeeef6b9f43e61b28/helm/nessie/templates/deployment.yaml#L21

For us this is a problem right now as many projects are moving from annotations to labels to control pod modification - among them istio.

Thus it would be great to be able to specify labels of the main deployment.

**Requested Changes in public API**:
Additional optional config option in the values.yaml

**Expected Use Cases**:
istio injection label

PR #7287, opened 2023-07-25
Add podLabels for Nessie Pods

c-thiel:
Fixes https://github.com/projectnessie/nessie/issues/7286

### DIFF
--- a/helm/nessie/templates/deployment.yaml
+++ b/helm/nessie/templates/deployment.yaml
@@ -19,6 +19,9 @@ spec:
       {{- end }}
       labels:
         {{- include "nessie.selectorLabels" . | nindent 8 }}
+        {{- if .Values.podLabels }}
+        {{- toYaml .Values.podLabels | nindent 8 }}
+        {{- end }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/helm/nessie/templates/deployment.yaml
+++ b/helm/nessie/templates/deployment.yaml
@@ -4,6 +4,9 @@ metadata:
   name: {{ include "nessie.fullname" . }}
   labels:
     {{- include "nessie.labels" . | nindent 4 }}
+    {{- if .Values.podLabels }}
+    {{- toYaml .Values.podLabels | nindent 4 }}
+    {{- end }}
 spec:
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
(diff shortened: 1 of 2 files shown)


dimas-b (approved):

helm/nessie/templates/deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
       {{- end }}
       labels:
         {{- include "nessie.selectorLabels" . | nindent 8 }}
+        {{- if .Values.podLabels }}
+        {{- toYaml .Values.podLabels | nindent 8 }}
```
adutra, line 23: Don't we want to add these labels also to the `selector.matchLabels` section above?

c-thiel: No, typically not. MatchLabels are there to find a pod in Kubernetes, thus they should include the app name, component, release and so on. They are there to find pods that are managed by the deployment.
Take my istio for example: We still want to manage pods with the deployment even if the istio label does not match. In that case, that label should be changed and not a new pod deployed.

Match Labels are typically not editable in the values.yaml.

However, what I did miss: I think it would be good to add those labels to the deployment and the pods. I will add that.

adutra: Yes, `spec.selector.matchLabels` has to be a subset of `spec.template.metadata.labels`.

> We still want to manage pods with the deployment even if the istio label does not match.

But if you add the custom labels to both places, the labels _will_ match. In fact I cannot think of a case where adding the labels to both places would result in an incorrect deployment. Are you planning to change or delete the labels after the pod is created?

My fear is, if we don't add the custom labels to `spec.selector.matchLabels`, I think it's only a matter of time until somebody opens an issue and asks to add the custom labels there :-)

c-thiel: I guess those are edge cases - but for example if you change manually this functional label of a pod, the pod would no longer be managed by the deployment and a new pod would be spun up.

In my opinion matchLabels should not include any functional labels and really only informative labels that are there to identify the pod.
Another option would be OPA / Gatekeeper that might enforce policies and thus change labels. It might change any functional label of a pod.

I am not sure if that counts, but one of my favorite helm charts, the community one from Airflow, does it the same way:
https://github.com/airflow-helm/charts/blob/main/charts/airflow/templates/scheduler/scheduler-deployment.yaml

Its battleproven and trusted by many of companies.

adutra (approved):

PR #7291, opened 2023-07-25
Fix examples in openapi.yaml

adutra:
The current operations example causes 3 errors when used as is:

1. "Cannot set the commit hash on the client side. It is set by the server."
2. "New value for key 'example.key' must not have a content ID."
3. "Namespace 'example' must exist."


adutra:
The current `transplant` example also has a wrong indentation issue.

snazy (approved):

dimas-b (approved):

Issue #7268, opened 2023-07-19
Support relative hashes across the entire API v2 surface


colan-dremio:
**Description**:

[PR 6932](https://github.com/projectnessie/nessie/pull/6932) added support for relative hashes (e.g. `main*10000000000`) to many APIs. It would be good to expose this concept across the entire API surface.

For example, the commit log API already supports relative hashes, e.g. `GET http://localhost:19120/api/v2/trees/main*2023-07-17T17:44:36.463564257Z/history`. Since the `GET api/v2/{ref}/history` endpoint is logically a child of the getReference endpoint `GET api/v2/{ref}`, it makes sense to support the same semantics in both. Today, `GET http://localhost:19120/api/v2/trees/main*2023-07-17T17:44:36.463564257Z` does not work, but it should.

Create reference `POST /api/v2/trees` is another example of a place where this would be useful. This would support operations like "Create branch dev at the state of main as it existed on 2023-07-17".

Since all relative hashes evaluate to a real hash, semantically it is similar / the same as specifying a raw commit hash directly (Detached). In some cases, they will evaluate to the commit at the tip of a branch. IMO committing to a relative hash should never be allowed even if it resolves to the tip of a branch. Treating them as resolving to Detached will give us this behavior.

Another edge case is timestamps in the future (barring clock drift). It would be good to throw an error if a timestamp is provided too far into the future. Perhaps this is better as a separate issue.

dimas-b:
> Create reference POST /api/v2/trees is another example of a place where this would be useful. This would support operations like "Create branch dev at the state of main as it existed on 2023-07-17".

IIRC, the new reference hash is in the `POST` payload, which is a JSON representation of a reference object. Relative hashes are not supported in the `hash` property of a reference JSON object in any API v2 endpoints.

Adding relative hash resolution to JSON payloads has a much bigger scope than make the `{ref}` path parameter work uniformly. I'd suggest to fork this into a separate issue.

adutra:
I think we need to agree on what we mean by "the entire API surface":

1. Do we want relative hashes in commits, merges, transplants? This was disallowed recently by #7193 and @colan-dremio seems to be onboard with that ("committing to a relative hash should never be allowed").
2. Do we want / need relative hashes in path parameters, where the hash is meant to be the expected hash on the branch? I don't see a lot of interest in that. Currently, path params accept relative hashes only for endpoints reading the contents / commitlog / diffs.
[...]

PR #7305, opened 2023-07-29
Add a generic resolveHash method to VersionStore

adutra:
Required for #7268.

For context: the existing `hashOnReference` method has the down side of throwing `ReferenceNotFound` if the hash is not reachable from the given `NamedRef` – even if the hash exists. This makes supporting relative hashes across the board harder to implement and could introduce unwanted behavioral changes in existing endpoints.

### DIFF
--- a/versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
+++ b/versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
@@ -140,6 +140,13 @@ public class PersistVersionStore implements VersionStore {
     return databaseAdapter.hashOnReference(namedReference, hashOnReference);
   }
 
+  @Override
+  public Hash resolveHash(Hash hash, List<RelativeCommitSpec> relativeLookups) {
+    checkArgument(
+        relativeLookups.isEmpty(), "Relative lookups not supported for old database model");
+    return hash;
+  }
+
   @Nonnull
   @jakarta.annotation.Nonnull
   @Override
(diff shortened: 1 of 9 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/ParsedHash.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+
+@Value.Immutable
+public interface ParsedHash {
```
adutra, line 32: Not required for now, but will be for #7268. Same for `ResolvedHash`.

snazy (commented):
Not yet sure I understand the reasoning here. But let's wait for the PR proposed for #7268. Other than that, we'd have two very similar methods then, does it make sense to keep the `relativeLookups` parameter for `hashOnReference` after this PR and and the one for #7268?

adutra:
> Not yet sure I understand the reasoning here. But let's wait for the PR proposed for #7268. Other than that, we'd have two very similar methods then, does it make sense to keep the `relativeLookups` parameter for `hashOnReference` after this PR and and the one for #7268?

Indeed the intent will become clear with #7268 I hope. And yes, we'd need both, since sometimes we need to resolve the hash AND check that it belongs to the given named ref, but some other times (DETACHED cases mostly), we only need to resolve the hash.

adutra:
Closing this, you were right: it turns out `hashOnReference` is enough for all cases. Sorry for the noise.

PR #7308, opened 2023-07-31
Extend relative hash support to whole API v2

### LINKED ISSUE #7268: Support relative hashes across the entire API v2 surface

adutra:
Fixes #7268.

Changes are dispatched in different commits for easier reviews.

Highlights:

* A new component, `org.projectnessie.services.hash.HashResolver` is responsible for resolving all hashes. It replaces the `namedRefWithHashOrThrow` method previously in `BaseApiImpl`.
* Most changes are in `TreeApiImpl`. The most difficult endpoints to adapt were `assignReference` and `deleteReference` because of their strict checking of the expected hash.
  * Note that because `TreeApiImpl` is shared between v1 and v2, some changes may affect v1's behavior (but not its API). Since v1 is deprecated at this point, I think this is acceptable.
* The REST layer is designed in such a way that relative hashes are accepted even if they are not particularly useful, or even desired, in certain cases. This is by design to avoid returning cryptic validation errors to clients.
* As a general rule, ambiguous hases are only accepted in reading operations (see below for examples).
* A (hopefully) thorough set of tests can be found in `org.projectnessie.jaxrs.tests.AbstractRelativeReferences`.

Note on hash ambiguity: it is my understanding that, in order for a hash to be unambiguous (i.e. it will be resolved deterministically), it is enough for it to have an absolute part (aka start commit ID). This is valid also for hashes with timestamps, including those in the future, because no relative spec can be resolved to a child of the start commit, only its parents. Thus the following are all unambiguous:

* `1234`
* `1234~1`
* `1234^1~2`
* `1234*1970-01-01T00:00:00.000Z`
* `1234*2050-01-01T00:00:00.000Z` => will always resolve to 1234, assuming 1234 was committed before 2050.

The following however are considered ambiguous and thus rejected for all writing operations:

* `~1`
* `^1~2`
* `*1970-01-01T00:00:00.000Z`
* `*2050-01-01T00:00:00.000Z`

Given the above conclusions, this PR makes no special treatment for clock drifts.


servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -373,10 +374,36 @@ public void references() throws Exception {
...
 
+    // We cannot delete a branch if the expected hash is not reachable from its HEAD. Here,
+    // the expected hash is not reachable anymore, because the branch was reassigned to main
+    // previously.
+    // In such cases we expect a not-found error.
     AbstractThrowableAssert<?, ? extends Throwable> deleteConflict =
         isV2()
             ? soft.assertThatThrownBy(() -> api().deleteBranch().branch(branch).getAndDelete())
             : soft.assertThatThrownBy(() -> api().deleteBranch().branch(branch).delete());
+    deleteConflict
```
adutra, line 385: This is the only test that failed after the changes. It is testing an edge case: the branch is at c1, but the provided expected hash is c2.

Previously there was no hash resolution, so the endpoint would throw a CONFLICT error since c2 is not the expected hash.

With the new hash resolution in place, the endpoint now throws NOT_FOUND because c2 is not reachable from c1 at all.

I added another test case below that generates a CONFLICT as before.

dimas-b: I'm not sure about the new behavour here. In this case, the hash is valid, but it is not in the current commit chain of the reference. This is exactly the case when one client assigns the reference, while another client attempts to delete it using the old HEAD. I tend to think that a "conflict" is more appropriate from the high-level perspective.

We seems to have two general approaches. Read operations throw "not found" when the hash is not found on the branch. Write operations throw "conflict" when the "expected" hash does not match server-side expectations. The relative hash case kind of mixes both cases.

Since write operations now require a concrete base hash, could we make the resolution more lenient and defer the (final) exception throwing to the highest-level method that knows the operation's end-user semantics? 

On the other hand, all non-trivial unambiguous relative hash references resolve to something other than HEAD. Perhaps, we could simplify the parsing of the expected hash parameter for assign/deleteReference requests and respond with a "bad request" if a relative spec is used. If the hash is absolute we could avoid checking it against the history, but do "conflict" by comparing it to current HEAD right away (also requires less backend work).

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
```diff
@@ -82,7 +82,7 @@ public Namespace createNamespace(String refName, Namespace namespace)
       throws NessieReferenceNotFoundException {
     Preconditions.checkArgument(!namespace.isEmpty(), "Namespace name must not be empty");
 
-    WithHash<NamedRef> refWithHash = namedRefWithHashOrThrow(refName, null);
+    ResolvedHash refWithHash = getHashResolver().resolveToHead(refName);
```
dimas-b, line 85: The Namespaces endpoint is not supported in API v2, so relative hashes normally cannot make their way here. It's probably a moot point, but if there is an easier way to get HEAD, it might be worth using it in this class.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
adutra: Your comments are 100% spot on. I hesitated a long time myself before picking one solution.

To summarize, we'd have 3 ways to move forward with this issue:

1. Accept the behavior change: if c2 is not in main, then `main@c2` should legitimately throw not-found.
2. Use `DETACHED` to resolve relative hashes.
  a.  Using `DETACHED`, `VersionStore.hashOnReference()` will simply lookup the commit but won't validate that it belongs to any branch.
3. respond with a "bad request" if a relative spec is used in `assignReference` or `deleteReference`
  a. This would work too, but it might not solve the issue completely because instead of conflict, we would throw a bad-request error.

Regarding option 2: this was actually my first idea. It solves the issue at hand, and doesn't seem to have any concerning downsides. I gave up on it, thinking that conceptually, option 1 was better than 2... but now I am not sure anymore :-)

Let's try option 2 first and see how it goes. If it doesn't look good, then I will go with your suggestion (option 3).

servers/services/src/main/java/org/projectnessie/services/impl/NamespaceApiImpl.java
adutra: The change here is purely cosmetic. `getHashResolver().resolveToHead(refName)` does exactly the same thing as `namedRefWithHashOrThrow(refName, null);`. But the new name is more meaningful imho. Do you see any other easier way to resolve the HEAD of a ref?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
adutra: @dimas-b Seems like CI is happy with option 2!

servers/services/src/main/java/org/projectnessie/services/hash/HashResolver.java
```diff
@@ -0,0 +1,164 @@
...
+import org.projectnessie.versioned.DetachedRef;
+import org.projectnessie.versioned.GetNamedRefsParams;
+import org.projectnessie.versioned.Hash;
+import org.projectnessie.versioned.NamedRef;
+import org.projectnessie.versioned.ReferenceInfo;
+import org.projectnessie.versioned.ReferenceNotFoundException;
+import org.projectnessie.versioned.RelativeCommitSpec;
+import org.projectnessie.versioned.VersionStore;
+
+public class HashResolver {
```
snazy, line 40: Nit:
```suggestion
public final class HashResolver {
```

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -369,42 +387,61 @@ public <R> R getCommitLog(
       String pageToken,
       PagedResponseHandler<R, LogEntry> pagedResponseHandler)
       throws NessieNotFoundException {
-    // we should only allow named references when no paging is defined
-    WithHash<NamedRef> endRef =
-        namedRefWithHashOrThrow(namedRef, null == pageToken ? youngestHash : pageToken);
+    // FIXME we should only allow named references when no paging is defined
```
snazy, line 390: TBH; I'm not sure that comment is still valid (aka: I think, it can just go away)

servers/services/src/main/java/org/projectnessie/services/hash/HashValidators.java
```diff
@@ -0,0 +1,51 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.Objects;
+import java.util.function.BiConsumer;
+
+public final class HashValidators {
+
+  private HashValidators() {}
+
+  /** Validates any parseable hash. */
```
snazy, line 27:
```suggestion
  /** No-op, for any parseable hash. */
```

servers/services/src/main/java/org/projectnessie/services/hash/HashValidators.java
```diff
@@ -0,0 +1,51 @@
...
+  public static final BiConsumer<String, ParsedHash> ANY_HASH = (name, parsed) -> {};
+
+  /** Validates that a hash has been provided (absolute or relative). */
+  public static final BiConsumer<String, ParsedHash> REQUIRED_HASH =
+      (name, parsed) -> Objects.requireNonNull(parsed, String.format("%s must be provided.", name));
+
+  /**
+   * Validates that, if a hash was provided, it is unambiguous. A hash is unambiguous if it has an
+   * absolute part, because it will always resolve to the same hash, even if it also has relative
```
snazy, line 36:
```suggestion
   * Validates that, if a hash was provided, it is unambiguous. A hash is unambiguous if it starts with
   * an absolute part, because it will always resolve to the same hash, even if it also has relative
```

servers/services/src/main/java/org/projectnessie/services/hash/HashValidators.java
```diff
@@ -0,0 +1,51 @@
...
+   * parts.
+   */
+  public static final BiConsumer<String, ParsedHash> UNAMBIGUOUS_HASH =
+      (name, parsed) ->
+          checkArgument(
+              parsed == null || parsed.getAbsolutePart().isPresent(),
+              String.format("%s must contain a starting commit ID.", name));
+
+  /**
+   * Validates that a hash has been provided, and that it is unambiguous (that is, it contains an
```
snazy, line 46:
```suggestion
   * Validates that a hash has been provided, and that it is unambiguous (that is, it starts with an
```

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: I thought about that too and came to the conclusion that it could still be relevant. If the named reference is reassigned between two pages to a completely different hash, paging could break I guess. Wdyt?

api/model/src/main/java/org/projectnessie/api/v2/params/Transplant.java
```diff
@@ -71,23 +72,35 @@ public interface Transplant extends BaseMergeTransplant {
...
+   *
+   * <p>Since Nessie spec 2.1.1, hashes can be absolute or relative.
+   */
   @NotNull
   @jakarta.validation.constraints.NotNull
   @Size
   @jakarta.validation.constraints.Size(min = 1)
-  List<String> getHashesToTransplant();
+  List<
+          @Pattern(
```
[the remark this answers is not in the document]
adutra: I manually validated that this annotation works both with Jaxrs/Weld and Quarkus. 

(For Jaxrs, I added Hibernate Validator recently in #7193, for Quarkus it's already present in quarkus-common.)

The problem is, the bean is being validated by the `@Check` method too, and that method is executed first.

If you remove the `@Check` method, then you would observe how Hibernate Validator validates that:

```
transplantCommitsIntoBranch.transplant.hashesToTransplant[0].<list element>: Hash with optional relative part must consist of either ... [etc.]
```

To be fair, I think we don't need the `@Check` method anymore, and besides, the HV error message is more explicit because it mentions where the problem was found. But maybe keeping it is safer, because we own the validation code. Wdyt?

servers/services/src/main/java/org/projectnessie/services/hash/ResolvedHash.java
```diff
@@ -0,0 +1,83 @@
...
+  Optional<Hash> getHead();
+
+  /**
+   * The effective resolved hash, never {@code null}. Will either be {@link #getProvidedHash()} if
+   * present, or {@link #getHead()} otherwise.
+   */
+  @Override
+  Hash getHash();
+
+  @Value.Derived
```
snazy, line 65:
```suggestion
  @Value.NonAttribute
```
otherwise the value would be stored twice (in two fields): in `WithHash` and in this one.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
snazy: Paging for commit-log uses a commit-hash passed around as the offset for the next page. It could be an issue when the reference is deleted, but then it's deleted, so fine to fail IMO.

adutra: OK let's remove the comment then.

api/model/src/main/java/org/projectnessie/api/v2/params/Transplant.java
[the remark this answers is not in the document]
adutra: OK but let's do it in a follow-up PR. I fear that some tests could fail, e.g. if they expect an exact error message.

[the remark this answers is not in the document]
adutra: #7316

snazy (approved):
LGTM! Thanks!

servers/services/src/testFixtures/java/org/projectnessie/services/impl/AbstractTestInvalidRefs.java
```diff
@@ -37,29 +37,21 @@ public void testUnknownHashesOnValidNamedRefs() throws BaseNessieClientServerExc
     createCommits(branch, 1, commits, currentHash);
     assertThatThrownBy(() -> commitLog(branch.getName(), MINIMAL, null, invalidHash, null))
         .isInstanceOf(NessieNotFoundException.class)
-        .hasMessageContaining(
-            String.format(
-                "Could not find commit '%s' in reference '%s'.", invalidHash, branch.getName()));
+        .hasMessageContaining(String.format("Commit '%s' not found", invalidHash));
```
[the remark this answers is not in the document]
dimas-b: LGTM :+1:

PR #7313, opened 2023-08-01
Add .python-version to .gitignore

adutra:
Useful when pynessie is installed in a virtualenv with pyenv.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,9 +61,10 @@ out/
 build/
 !gradle-wrapper.jar
 
-# Python
+# Python & Pyenv
 venv/
 __pycache__/
+.python-version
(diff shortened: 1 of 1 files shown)


snazy (approved):

snazy:
Does this make sense in https://github.com/projectnessie/pynessie as well?

adutra:
> Does this make sense in https://github.com/projectnessie/pynessie as well?

Yes, I will open a PR there as well.

adutra:
> > Does this make sense in https://github.com/projectnessie/pynessie as well?
> 
> Yes, I will open a PR there as well.

nvm it's there already.

PR #7314, opened 2023-08-01
DRAFT - Fix broken assertion when Nessie lib is upgraded to 0.61+ in Iceberg

adutra:
Will not compile until Nessie is upgraded to 0.61+ in Iceberg.


snazy:
Included in #7539, closing

PR #7320, opened 2023-08-02
Fix NesQuEIT IntelliJ import

snazy:
The NesQuEIT project includes the Nessie sources as two Gradle builds - one for everything except Iceberg and one for the rest that has dependencies to Iceberg.

The logic in `nessie-conventions-root.gradle.kts` refers to just `projectDir`, which points to `nessie-iceberg/` instead of the Nessie source root dir.

This change fixes the above issue.


build-logic/src/main/kotlin/nessie-conventions-root.gradle.kts
```diff
@@ -38,26 +38,31 @@ val projectName = rootProject.file("ide-name.txt").readText().trim()
...
+
   idea {
     module {
       name = ideName
       isDownloadSources = true // this is the default BTW
       inheritOutputDirs = true
 
-      val sparkScalaProps = Properties()
-      val integrationsDir = projectDir.resolve("integrations")
+      val nessieRootProjectDir =
```
adutra, line 48: Could you maybe add a comment here explaining the logic? Otherwise it's a bit intriguing.

adutra (approved):

PR #7348, opened 2023-08-08
Support assign/delete reference without type in java client

dimas-b:
#7282 allowed skipping type parameters in REST API requests to 
delete / assign references (tags or branches). However, the java 
client API still required the user to know the exact reference 
type and use type-specific API methods for those operations.

This change introduces generic assign / delete reference operations
to the java client to keep it functionally aligned to the REST API.

* Add new `AssignReferenceBuilder` and `DeleteReferenceBuilder`
  interfaces with corresponding methods in `NessieApiV2`.

* Introduce common base classes to share builder code across API
  methods. Refactor old requests builders to use the shared base
  classes.

* Fix validation exception in HashValidator.REQUIRED_HASH to make
  it produce "bad request" (400) errors at the API level as opposed
  to "internal server error" (500).

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
@@ -28,4 +28,10 @@ public interface NessieApiV2 extends NessieApiV1 {
   GetRepositoryConfigBuilder getRepositoryConfig();
 
   UpdateRepositoryConfigBuilder updateRepositoryConfig();
+
+  /** Delete a branch or a tag. */
+  DeleteReferenceBuilder deleteReference();
+
+  /** Update a branch or a tag (make it point to an arbitrary commit). */
+  AssignReferenceBuilder assignReference();
 }
--- a/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
@@ -15,6 +15,8 @@
  */
 package org.projectnessie.client.api;
 
+import org.projectnessie.model.Reference;
+
 /**
  * Interface for the Nessie V2 API implementation.
  *
@@ -30,8 +32,36 @@ public interface NessieApiV2 extends NessieApiV1 {
   UpdateRepositoryConfigBuilder updateRepositoryConfig();
 
   /** Delete a branch or a tag. */
-  DeleteReferenceBuilder deleteReference();
+  DeleteReferenceBuilder<Reference> deleteReference();
 
   /** Update a branch or a tag (make it point to an arbitrary commit). */
-  AssignReferenceBuilder assignReference();
+  AssignReferenceBuilder<Reference> assignReference();
+
+  /**
+   * @deprecated Use {@code assignReference().asTag()} instead.
+   */
+  @Override
+  @Deprecated
+  AssignTagBuilder assignTag();
+
+  /**
+   * @deprecated Use {@code assignReference().asBranch()} instead.
+   */
+  @Override
+  @Deprecated
+  AssignBranchBuilder assignBranch();
+
+  /**
+   * @deprecated Use {@code deleteReference().asTag()} instead.
+   */
+  @Override
+  @Deprecated
+  DeleteTagBuilder deleteTag();
+
+  /**
+   * @deprecated Use {@code deleteReference().asBranch()} instead.
+   */
+  @Override
+  @Deprecated
+  DeleteBranchBuilder deleteBranch();
 }
--- a/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
@@ -38,6 +38,8 @@ public interface NessieApiV2 extends NessieApiV1 {
   AssignReferenceBuilder<Reference> assignReference();
 
   /**
+   * {@inheritDoc}
+   *
    * @deprecated Use {@code assignReference().asTag()} instead.
    */
   @Override
@@ -45,6 +47,8 @@ public interface NessieApiV2 extends NessieApiV1 {
   AssignTagBuilder assignTag();
 
   /**
+   * {@inheritDoc}
+   *
    * @deprecated Use {@code assignReference().asBranch()} instead.
    */
   @Override
@@ -52,6 +56,8 @@ public interface NessieApiV2 extends NessieApiV1 {
   AssignBranchBuilder assignBranch();
 
   /**
+   * {@inheritDoc}
+   *
    * @deprecated Use {@code deleteReference().asTag()} instead.
    */
   @Override
@@ -59,6 +65,8 @@ public interface NessieApiV2 extends NessieApiV1 {
   DeleteTagBuilder deleteTag();
 
   /**
+   * {@inheritDoc}
+   *
    * @deprecated Use {@code deleteReference().asBranch()} instead.
    */
   @Override
--- a/api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
+++ b/api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
@@ -13,17 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.projectnessie.client.http.v2api;
+package org.projectnessie.client.builder;
 
-import org.projectnessie.client.api.AssignBranchBuilder;
-import org.projectnessie.client.http.HttpClient;
-import org.projectnessie.model.Branch;
 import org.projectnessie.model.Reference;
 
-final class HttpAssignBranch extends BaseHttpAssignReference<AssignBranchBuilder, Branch>
-    implements AssignBranchBuilder {
+public abstract class BaseAssignReferenceBuilder<R> extends BaseOnReferenceBuilder<R> {
 
-  HttpAssignBranch(HttpClient client) {
-    super(client, Reference.ReferenceType.BRANCH);
+  protected Reference assignTo;
+
+  protected BaseAssignReferenceBuilder(Reference.ReferenceType assumedType) {
+    super(assumedType);
+  }
+
+  @SuppressWarnings("unchecked")
+  public R assignTo(Reference assignTo) {
+    this.assignTo = assignTo;
+    return (R) this;
   }
 }
--- a/api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
+++ b/api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
@@ -17,13 +17,9 @@ package org.projectnessie.client.builder;
 
 import org.projectnessie.model.Reference;
 
-public abstract class BaseAssignReferenceBuilder<R> extends BaseOnReferenceBuilder<R> {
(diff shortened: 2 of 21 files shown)


api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
```diff
@@ -28,4 +28,10 @@ public interface NessieApiV2 extends NessieApiV1 {
   GetRepositoryConfigBuilder getRepositoryConfig();
 
   UpdateRepositoryConfigBuilder updateRepositoryConfig();
+
+  /** Delete a branch or a tag. */
+  DeleteReferenceBuilder deleteReference();
```
adutra, line 33: The old `deleteBranch` and `assignBranch` methods are still available, right? Wouldn't this be confusing to end users? Wasn't it possible to override the existing methods with covariant return types?

dimas-b: Not sure about covariant return types. I guess we still have to maintain method signatures to avoid breaking older clients compiled with java 8... Also, old `getAndDelete()` methods have more specific return types.

servers/services/src/main/java/org/projectnessie/services/hash/HashValidator.java
```diff
@@ -28,7 +27,7 @@ public interface HashValidator {
 
   /** Validates that a hash has been provided (absolute or relative). */
   HashValidator REQUIRED_HASH =
-      (name, parsed) -> Objects.requireNonNull(parsed, String.format("%s must be provided.", name));
+      (name, parsed) -> checkArgument(parsed != null, String.format("%s must be provided.", name));
```
adutra, line 30: Is there a difference on the client side between throwing NPE or IAE here?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -308,11 +308,17 @@ public Reference assignReference(
...
       throws NessieNotFoundException, NessieConflictException {
     try {
+      ReferenceInfo<CommitMeta> resolved =
+          getStore().getNamedRef(referenceName, GetNamedRefsParams.DEFAULT);
 
       ResolvedHash oldRef =
           getHashResolver()
               .resolveHashOnRef(
-                  referenceName, expectedHash, "Expected hash", REQUIRED_UNAMBIGUOUS_HASH);
+                  resolved.getNamedRef(),
```
adutra, line 317: Isn't this change functionally equivalent to the old version? Except for the detached use case... I suspect you are trying to special-case detached refs here, but why?

(BTW this change is probably why we are getting the strange 404 error in one of the test cases above.)

api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
adutra: That's said but I guess we don't have the choice then. 👍

servers/services/src/main/java/org/projectnessie/services/hash/HashValidator.java
dimas-b: Yes, NPE -> 500, IAE -> 400. I think the latter is more correct from the client's perspective.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: The problem is that assigning without an "expected hash" used to work, which is incorrect IMHO. This is case I was trying to fix. I'll see if I can find a better solution.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
```diff
@@ -204,8 +205,24 @@ public void assignReferenceFromAmbiguous() {
   @Test
   @NessieApiVersions(versions = NessieApiVersion.V2)
   public void assignReferenceFromAmbiguousDetached() {
```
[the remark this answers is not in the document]
dimas-b: I was confused by the "from" word (also missed the other test) :man_facepalming:

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
```diff
@@ -204,8 +205,24 @@ public void assignReferenceFromAmbiguous() {
...
+
+  /**
+   * The {@code @~1} ref spec means a "detached" reference, which is not a named reference, hence it
+   * cannot be assigned.
+   */
+  @Test
+  @NessieApiVersions(versions = NessieApiVersion.V2)
+  public void assignDetachedReferenceDeclined() {
+    NessieError error = expectError(rest().body(etl1).put("trees/@~1"), 404);
+    checkError(error, REFERENCE_NOT_FOUND, "Named reference 'DETACHED' not found");
```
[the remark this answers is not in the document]
dimas-b: Let's approach this from a different angle. A delete reference call with the same URL path currently produces `Named reference 'DETACHED' not found`.

I think both assign and delete reference end points should use the same reference lookup logic.

That said, `DETACHED` as a name cannot be use to create a reference. Even though `DetachedRef extends NamedRef`, as the API level `ReferenceType` only has `BRANCH` and `TAG`. This is asymmetrical, but essentially at the API level `DETACHED` references are not "named references". For example, they cannot be obtained from `getReferenceByName`.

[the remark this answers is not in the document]
dimas-b: Also, a request to `put("trees/@1122334455667788")` without this PR leads to `Argument must be a branch or tag, but was DetachedRef{}`. I think a `Named reference 'DETACHED' not found` error message is clearer.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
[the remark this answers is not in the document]
dimas-b: reverted

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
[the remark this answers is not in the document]
dimas-b: Let's sort this out in a follow-up PR.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: I've added a simple `checkArgument` for that use case for now on top of old reference resolution logic.

api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
snazy: WDYT about overriding the functions in `NessieApiV2` and mark them as deprecated there?

snazy: Changing functions would break existing code (ABI)

api/client/src/main/java/org/projectnessie/client/builder/BaseOnReferenceBuilder.java
```diff
@@ -15,24 +15,54 @@
  */
 package org.projectnessie.client.builder;
 
-import org.projectnessie.client.api.OnReferenceBuilder;
+import org.projectnessie.model.Reference;
 
-abstract class BaseOnReferenceBuilder<R extends OnReferenceBuilder<R>>
-    implements OnReferenceBuilder<R> {
+public abstract class BaseOnReferenceBuilder<R> {
```
snazy, line 20: I'd prefer to keep this class package-private, because users could simply up-cast to this one.
I suspect, this should still work with the added `public` functions.

api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
```diff
@@ -15,17 +15,19 @@
  */
 package org.projectnessie.client.builder;
 
-import org.projectnessie.client.api.AssignBranchBuilder;
 import org.projectnessie.model.Reference;
 
-public abstract class BaseAssignBranchBuilder extends BaseOnBranchBuilder<AssignBranchBuilder>
-    implements AssignBranchBuilder {
+public abstract class BaseAssignReferenceBuilder<R> extends BaseOnReferenceBuilder<R> {
```
snazy, line 20: Let's keep this class package-private as well

snazy (commented):
LGTM, just not a fan of having/making the `Base*` classes public

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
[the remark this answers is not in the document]
adutra: OK, I am going to approve since the main intent of the PR is valid. 

But indeed I would like us to come up, if possible, with a better solution for resolving detached refs and presenting errors to the client. The whole point of #7268 was to give a consistent behavior to all endpoints regarding resolution of named refs and relative hashes. I find it a bit disappointing to special-case one or two endpoints again.

I think we should instead revisit `resolveHashOnRef`. Maybe there are other endpoints where we should disallow detached refs completely? For example, I suspect that commit, merge and transplant endpoints could benefit from the same common handling of non-branch refs: for now, they all have special checks that throw HTTP 400 with message "Can only commit into branches."

The ideal error message in such cases would imho be HTTP 400 and a message like `Reference (to assign|to delete|to commit into|etc.) must be a branch, but was: (detached|tag)` or stg in the style.

With this refactoring, the `checkArgument` call could certainly be removed, since it does more or less the same of what `HashValidator.EXPECTED_HASH` is already supposed to do.

I can look into this follow-up work.

adutra (approved):
Approving but please don't forget @snazy 's comments :-)

api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
adutra: +1, very good suggestion.

snazy:
Don't forget to update `CHANGELOG.md`

api/client/src/main/java/org/projectnessie/client/builder/BaseOnReferenceBuilder.java
dimas-b: These builders are actually meant for reuse. Still you comment is relevant. I've reworked the interaces and base builders in the latest push. Now builders should be aligned to the java API, so casting to the will not expose extra / irrelevant functionality.... Is that ok from you POV?

api/client/src/main/java/org/projectnessie/client/builder/BaseAssignReferenceBuilder.java
dimas-b: discussed in https://github.com/projectnessie/nessie/pull/7348#discussion_r1291601444

api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
dimas-b: Deprecated V1 java client assign/delete methods in the latest push.

snazy (approved):

PR #7352, opened 2023-08-09
Retry erase repo with non-admin path if DropRowRange fails

adutra:
Fixes #7351 
Fixes #7347

PR #7353, opened 2023-08-09
Add Quarkus config option to disable BigTable admin client

adutra:


dimas-b (commented):
The change LGTM. Still I wonder why would we want to disable the admin client?

dimas-b (approved):
I guess this aligns with #7352

adutra:
Yes, the idea is that many (most?) operations available from the admin client are subject to quotas. In some cases, it might be just simpler (and cheaper) to not use it. We would be trading time for money :-)

PR #7356, opened 2023-08-09
Export secondary parents

adutra:
This commit fixes an issue when a secondary parent is not exported, because it is not reachable through any live reference.

In such cases, an `ObjNotFoundException` was being thrown on import.

### DIFF
--- a/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
+++ b/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
@@ -104,20 +104,24 @@ final class ExportDatabaseAdapter extends ExportCommon {
                 commitsToProcess.push(head);
                 while (!commitsToProcess.isEmpty()) {
                   Hash hash = commitsToProcess.pop();
-                  try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(hash)) {
-                    for (Iterator<CommitLogEntry> commitIter = commits.iterator();
-                      commitIter.hasNext(); ) {
-                      CommitLogEntry commit = commitIter.next();
-                      if (!identify.handleCommit(commit)) {
-                        break;
+                  if (identify.isCommitNew(hash)) {
+                    try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(hash)) {
+                      for (Iterator<CommitLogEntry> commitIter = commits.iterator();
+                          commitIter.hasNext(); ) {
+                        CommitLogEntry commit = commitIter.next();
+                        if (!identify.handleCommit(commit)) {
+                          break;
+                        }
+                        commitHandler.accept(commit);
+                        for (Hash parent : commit.getAdditionalParents()) {
+                          if (identify.isCommitNew(parent)) {
+                            commitsToProcess.push(parent);
+                          }
+                        }
                       }
-                      commitHandler.accept(commit);
-                      for (Hash h : commit.getAdditionalParents()) {
-                        commitsToProcess.push(h);
-                      }
+                    } catch (ReferenceNotFoundException e) {
+                      throw new RuntimeException(e);
                     }
-                  } catch (ReferenceNotFoundException e) {
-                    throw new RuntimeException(e);
                   }
                 }
               });
--- a/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
+++ b/versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
@@ -95,22 +95,29 @@ final class ExportDatabaseAdapter extends ExportCommon {
 
     try (Stream<ReferenceInfo<ByteString>> namedRefs =
         databaseAdapter.namedRefs(GetNamedRefsParams.DEFAULT)) {
-      Deque<Hash> commitsToProcess = new ArrayDeque<>();
-      namedRefs.map(ReferenceInfo::getHash).forEach(commitsToProcess::push);
-      while (!commitsToProcess.isEmpty()) {
-        Hash hash = commitsToProcess.pop();
-        try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(hash)) {
-          commits.forEach(
-              commit -> {
-                if (identify.handleCommit(commit)) {
-                  commitHandler.accept(commit);
-                  for (Hash h : commit.getAdditionalParents()) {
-                    commitsToProcess.push(h);
+      namedRefs
+          .map(ReferenceInfo::getHash)
+          .forEach(
+              head -> {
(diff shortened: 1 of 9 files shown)


adutra:
Hmm forgot to fix the old model, doing it now.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportPersist.java
```diff
@@ -105,21 +105,34 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitObj> commitHandler)
...
+      Consumer<CommitObj> commitHandler,
+      CommitLogic commitLogic) {
+    while (commitIter.hasNext()) {
+      CommitObj commit = commitIter.next();
+      if (!identify.handleCommit(commit)) {
+        break;
+      }
+      commitHandler.accept(commit);
+      for (ObjId objId : commit.secondaryParents()) {
+        scanCommitLogChain(
```
dimas-b, line 130: Can we avoid recursion here? I suppose the commit DAG can have a lot of nested branch points.

adutra: Hmm how? We need to scan the whole lineage of each secondary parent. (Note that the scan stops as soon as a commit has been seen before, so we are not re-scanning commits more than once.)

dimas-b: We could still have a cascade of merge parents that is longer than the java stack, I believe... we'd have to recurse into them even if we do not really traverse any father :thinking:

adutra: You are right, let's remove recursion.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
```diff
@@ -97,20 +99,25 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitLogEntry> commitHand
...
-                      commitIter.hasNext(); ) {
-                    CommitLogEntry commit = commitIter.next();
-                    if (!identify.handleCommit(commit)) {
-                      break;
-                    }
-                    commitHandler.accept(commit);
+                Deque<Hash> commitsToProcess = new ArrayDeque<>();
+                commitsToProcess.push(head);
+                while (!commitsToProcess.isEmpty()) {
+                  Hash hash = commitsToProcess.pop();
```
dimas-b, line 105: I guess we may want to check `hash` against the flags in `identify` at this point to avoid accessing the commit log if case the hash has already been processed.

I guess this can be a common situation when two branches have multiple merge points.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
```diff
@@ -97,20 +99,25 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitLogEntry> commitHand
...
+                  try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(hash)) {
+                    commits.forEach(
+                        commit -> {
+                          if (identify.handleCommit(commit)) {
+                            commitHandler.accept(commit);
+                            for (Hash h : commit.getAdditionalParents()) {
+                              commitsToProcess.push(h);
```
dimas-b, line 112: Why not append? If we have two parallel branches with cross-merges processing newer commits first will reduce the number of commit log calls, I guess. If branches are not related, then there is perhaps no difference.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportPersist.java
```diff
@@ -106,14 +108,20 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitObj> commitHandler)
...
+              commitsToProcess.push(ref.pointer());
+              while (!commitsToProcess.isEmpty()) {
+                ObjId id = commitsToProcess.pop();
+                Iterator<CommitObj> commitIter = commitLogic.commitLog(commitLogQuery(id));
+                while (commitIter.hasNext()) {
+                  CommitObj commit = commitIter.next();
+                  if (identify.handleCommit(commit)) {
+                    commitHandler.accept(commit);
+                    for (ObjId objId : commit.secondaryParents()) {
+                      commitsToProcess.push(objId);
```
snazy, line 121: Let's add a check in `IdentifyHeadsAndForkPoints` that just tests whether a commit-_ID_ has been seen (basically what `identify.handleCommit()` does at the beginning) - and only push the ID if the commit has not been seen before.
And the same check after the `.pop` above.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
```diff
@@ -97,20 +99,25 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitLogEntry> commitHand
...
-                    }
-                    commitHandler.accept(commit);
+                Deque<Hash> commitsToProcess = new ArrayDeque<>();
+                commitsToProcess.push(head);
+                while (!commitsToProcess.isEmpty()) {
+                  Hash hash = commitsToProcess.pop();
+                  try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(hash)) {
+                    commits.forEach(
+                        commit -> {
+                          if (identify.handleCommit(commit)) {
```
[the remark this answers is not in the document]
adutra: Whoops yes, I also need to fix `ExportPersist`. Sorry.

versioned/transfer/src/main/java/org/projectnessie/versioned/transfer/ExportDatabaseAdapter.java
```diff
@@ -97,17 +100,29 @@ private HeadsAndForkPoints scanAllReferences(Consumer<CommitLogEntry> commitHand
...
           .forEach(
               head -> {
-                try (Stream<CommitLogEntry> commits = databaseAdapter.commitLog(head)) {
-                  for (Iterator<CommitLogEntry> commitIter = commits.iterator();
-                      commitIter.hasNext(); ) {
-                    CommitLogEntry commit = commitIter.next();
-                    if (!identify.handleCommit(commit)) {
-                      break;
+                Deque<Hash> commitsToProcess = new ArrayDeque<>();
+                commitsToProcess.offerFirst(head);
```
dimas-b, line 104: `Queue.add(head)` would be more intuitive (same on line 118), WDYT?

"offer" methods return a value, which we do not check, so it feels awkward in this case (even though they should not fail in practice).

dimas-b (commented):
LGTM overall :+1: just one minor comment about queue operations.

PR #7376, opened 2023-08-16
Ability to update the repository description

adutra:
Note: the update repo functionality is implemented in the old model at adapter level, although, it seems unused. This PR brings the same functionality to the new model.

### DIFF
--- a/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
+++ b/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
@@ -248,8 +248,8 @@ public class AbstractRepositoryLogicTests {
     RepositoryDescription updated =
         RepositoryDescription.builder()
             .putProperties("updated", "true")
-            // read-only properties, should be ignored
             .defaultBranchName("main2")
+            // the following attributes are read-only, should not be updated
             .oldestPossibleCommitTime(Instant.ofEpochSecond(12345))
             .repositoryCreatedTime(Instant.ofEpochSecond(456789))
             .build();
@@ -262,6 +262,7 @@ public class AbstractRepositoryLogicTests {
             ImmutableRepositoryDescription.builder()
                 .from(initial)
                 .putProperties("updated", "true")
+                .defaultBranchName("main2")
                 .build());
   }
 }
--- a/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
+++ b/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
@@ -45,6 +45,7 @@ import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 import org.projectnessie.versioned.storage.common.exceptions.RefNotFoundException;
+import org.projectnessie.versioned.storage.common.exceptions.RetryTimeoutException;
 import org.projectnessie.versioned.storage.common.logic.CommitLogic;
 import org.projectnessie.versioned.storage.common.logic.CreateCommit;
 import org.projectnessie.versioned.storage.common.logic.InternalRef;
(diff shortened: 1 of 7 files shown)


dimas-b (commented):
Changed LGTM, but it does not look like the new functionality can be accessed. Do you mean to have another PR to expose it somewhere?

dimas-b:
Would it make sense to add a description update command to `quarkus-cli`?

adutra:
> Changes LGTM, but it does not look like the new functionality can be accessed. Do you mean to have another PR to expose it somewhere?

Exactly, the idea is to use this functionality in the importer, in order to "mark" the repository as fully imported. I have a draft already: #7380 .

adutra:
> Would it make sense to add a description update command to `quarkus-cli`?

At this point, that is not required as the only update to the repo description will happen internally inside the importer. We can always add that later though.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/RepositoryLogicImpl.java
```diff
@@ -169,14 +184,51 @@ private void addRepositoryDescription(
...
 
+  @Override
+  public RepositoryDescription updateRepositoryDescription(
+      RepositoryDescription repositoryDescription) throws RetryTimeoutException {
+    try {
+      return commitRetry(
+          persist,
+          (p, retryState) -> {
+            try {
+              Reference reference = Objects.requireNonNull(persist.fetchReference(REF_REPO.name()));
```
dimas-b, line 204: We should probably fix the javadoc of `REF_REPO`, which currently says `Internal reference with always exactly one commit ...`

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/RepositoryLogicImpl.java
```diff
@@ -169,14 +184,51 @@ private void addRepositoryDescription(
                   null,
                   "application/json",
                   SHARED_OBJECT_MAPPER.writeValueAsBytes(repoDesc.build()));
-      // can safely ignore the ID returned from storeObj() - it's fine, if the obj already exists
+      // can safely ignore the response from storeObj() - it's fine, if the obj already exists
       persist.storeObj(string);
-      b.addAdds(commitAdd(KEY_REPO_DESCRIPTION, 0, requireNonNull(string.id()), null, null));
+      b.addAdds(
+          commitAdd(KEY_REPO_DESCRIPTION, 0, requireNonNull(string.id()), null, UUID.randomUUID()));
```
snazy, line 190: This change will only work for new repositories, but existing repositories do not have a content-ID for the repo-description. The update-logic does not work without a content-ID.

I'd prefer to leave the content-ID as `null`. (There's no need for a CID here)

I know that the repo-config stuff uses a CID, see comment in SLImpl

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/StringLogicImpl.java
```diff
@@ -68,6 +86,53 @@ public StringObj updateString(StringValue previousValue, String contentType, Str
...
+    StringValue existing = null;
+    if (existingElement != null) {
+      CommitOp op = existingElement.content();
+      if (op.action().exists()) {
+        existingValueId = op.value();
+        existingContentId = op.contentId();
+        existing = existingValueId != null ? fetchString(requireNonNull(existingValueId)) : null;
+      }
+    }
+    if (existingContentId == null) {
```
snazy, line 116: This makes updating repo-descriptions of already existing repos impossible (those have no CID for repo-description).

Mabe replacing the `if` with a simple `else` helps here.

Should also have some comment to explain the logic wrt repo-desc + repo-config

snazy (changes requested):

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/RepositoryLogicImpl.java
```diff
@@ -169,14 +182,60 @@ private void addRepositoryDescription(
...
       b.addAdds(commitAdd(KEY_REPO_DESCRIPTION, 0, requireNonNull(string.id()), null, null));
     } catch (ObjTooLargeException | ObjNotFoundException | IOException e) {
       throw new RuntimeException(e);
     }
   }
 
+  @Override
+  public RepositoryDescription updateRepositoryDescription(RepositoryDescription newDescription)
+      throws RetryTimeoutException {
+    // prevent modification of read-only attributes
```
adutra, line 196: This came to my mind after @dimas-b suggestion to introduce an option in quarkus-cli to allow end users to modify the repo description. If we ever do that later, I think we'd need a safety net like this one.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
```diff
@@ -246,15 +247,21 @@ public void updateRepositoryDescription() throws RetryTimeoutException {
 
     RepositoryDescription updated =
         RepositoryDescription.builder()
-            .defaultBranchName("main")
             .putProperties("updated", "true")
+            // read-only properties, should be ignored
+            .defaultBranchName("main2")
```
snazy, line 252: Heh, exactly this one might be updated. Let's remove "defaultBranchName" from that check.
I'm not sure whether there'll ever be a command to update arbitrary information - rather per attribute.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractRepositoryLogicTests.java
```diff
@@ -246,15 +247,21 @@ public void updateRepositoryDescription() throws RetryTimeoutException {
...
             .oldestPossibleCommitTime(Instant.ofEpochSecond(12345))
```
snazy, line 253: This and the below setting are actually read-only attributes.

snazy (approved):

PR #7382, opened 2023-08-16
Support reading arguments from a file in the `delete` CLI command

dimas-b:
* Extract file processing code from RefreshContent into BulkCommittingCommand.

* Move the `--message` option to CommittingCommand

* Reuse BulkCommittingCommand in DeleteContent.

* Add support for CSV and path-encoded keys in input files.


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/BulkCommittingCommand.java
```diff
@@ -0,0 +1,251 @@
...
+          "The fields separator for CVS input files (if not set, each line is interpreted as a URL path "
+              + "encoded key string).")
+  private String separator;
+
+  @Option(
+      names = {"--storage-model"},
+      description =
+          "If set, only those entries that have the specified storage model will be processed. "
+              + "This option is applicable only to JSON input files. (ignored if --input is not set).")
+  private String storageModel;
```
snazy, line 83: It is not clear what this does and what it's meant for.
(Hard to figure out - maybe add a reference to the model object and some docs to it)

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/BulkCommittingCommand.java
```diff
@@ -0,0 +1,251 @@
...
+
+  @Option(
+      names = {"--skip-tags"},
+      description =
+          "If set, input references that are not branches will be ignored (as opposed to reported as errors).")
+  private boolean skipTags;
```
snazy, line 89: Not sure whether this flag's actually useful. If not set, the commands will probably fail on all repos, potentially after processing a bunch of branches.
I think, the commands should only fail, if a reference was explicitly specified and that's a tag.

snazy (changes requested):

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/BulkCommittingCommand.java
dimas-b: I've removed this option. With `DatabaseAdapter` being effectively deprecated, it's probably not that useful. JSON can be filtered by other tools too.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/BulkCommittingCommand.java
dimas-b: removed this option

snazy (approved):

PR #7395, opened 2023-08-21
Include Hibernate Validator in nessie-jaxrs-testextension

adutra:
This commit adds HV as a dependency of the JAX-RS test extension, so that it is always present when running tests.

This uncovered a few places where tests were doing something wrong, but it wasn't being detected properly due to HV not being present at runtime.

### DIFF
--- a/integrations/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
+++ b/integrations/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
@@ -83,7 +83,7 @@ public class TestNessieIcebergViews extends BaseIcebergTest {
   public void beforeEach() throws Exception {
     try {
       api.commitMultipleOperations()
-          .branchName(branch)
+          .branch(reference)
           .commitMeta(fromMessage(CREATE_REQUIRED_NAMESPACES))
           .operation(Put.of(ContentKey.of("db"), Namespace.of("db")))
           .operation(Put.of(ContentKey.of("nessie"), Namespace.of("nessie")))
--- a/integrations/iceberg-views/src/test/java/org/apache/iceberg/nessie/BaseIcebergTest.java
+++ b/integrations/iceberg-views/src/test/java/org/apache/iceberg/nessie/BaseIcebergTest.java
@@ -56,9 +56,9 @@ import org.projectnessie.versioned.storage.testextension.PersistExtension;
 @NessieBackend(InmemoryBackendTestFactory.class)
 public class BaseIcebergTest {
 
-  @NessiePersist static Persist perssit;
+  @NessiePersist static Persist persist;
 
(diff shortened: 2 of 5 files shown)


integrations/iceberg-views/src/test/java/org/apache/iceberg/nessie/TestNessieIcebergViews.java
```diff
@@ -83,7 +83,7 @@ public TestNessieIcebergViews() {
   public void beforeEach() throws Exception {
     try {
       api.commitMultipleOperations()
-          .branchName(branch)
+          .branch(reference)
```
adutra, line 86: Without the fix, this now (rightfully) throws:

```
org.projectnessie.error.NessieBadRequestException: Bad Request (HTTP/400): commitMultipleOperations.expectedHash: must not be null

	at org.projectnessie.error.ErrorCode.lambda$asException$1(ErrorCode.java:66)
	at java.base/java.util.Optional.map(Optional.java:265)
	at org.projectnessie.error.ErrorCode.asException(ErrorCode.java:66)```

dimas-b: IIRC we have a few (Spark) SQL tests that rely on being able to commit without specifying the expected hash.

adutra: Where are those tests? I would like to take a closer look. In any case, we'll need to decide what to do, we cannot let them as they are now since they are probably relying on requests not being properly validated.

adutra: Are you referring to the tests in `:nessie-spark-extensions`? They seem to be passing.

adutra: Hmm I see you are probably referring to NesQuEIT tests. Some of them are failing indeed with the error above. I will take a look.

adutra: I am not sure I agree: actual production servers (Quarkus-based) have been validating this since the beginning. Only JAX-RS tests, to my best knowledge, were missing this validation and being a little too permissive.

dimas-b: Ah, sorry, I missed that part :facepalm:

adutra: So, I double-checked NesQuEIT tests and it appears they are passing fine with this PR. They will become problematic for my other PR #7396 , but here I think we are good.

PR #7396, opened 2023-08-21
Enforce expected hash presence when creating and assigning references

adutra:
This commit also refactors the HashValidator and HashResolver APIs
to achieve better error messages.

This commit also enhances the set of tests in `AbstractRelativeReferences` and adds:

* Tests for references without hashes (resolving to HEAD when allowed)
* Tests for all kinds of references (branches, tags and detached)
* Tests for expected hashes in various situations: 
  * Resolving to HEAD
  * Resolving to an ancestor of HEAD
  * Resolving to an unreachable commit from HEAD
  * Non-existent hash

Some validation constraints in `TreeService` and `Merge` related to hashes were
lifted, in order to let HashValidator validate them and produce
standardized error messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ as necessary. Empty sections will not end in the release notes.
 ### Changes
 - Java client API to assign/delete reference operations without specifying a concrete reference type
   (no change to REST API).
+- Creating and assigning references now requires a target hash to be specified.
 
 ### Deprecations
 
--- a/api/model/src/main/java/org/projectnessie/api/v2/params/Merge.java
+++ b/api/model/src/main/java/org/projectnessie/api/v2/params/Merge.java
@@ -28,7 +28,6 @@ import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import javax.annotation.Nullable;
-import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
@@ -93,14 +92,14 @@ public interface Merge extends BaseMergeTransplant {
    *
    * <p>Since Nessie spec 2.1.1, the hash can be absolute or relative.
    */
-  @NotBlank
-  @jakarta.validation.constraints.NotBlank
   @Pattern(
       regexp = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_REGEX,
       message = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_MESSAGE)
   @jakarta.validation.constraints.Pattern(
       regexp = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_REGEX,
       message = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_MESSAGE)
+  @Nullable
+  @jakarta.annotation.Nullable
   String getFromHash();
 
   /**
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
@@ -394,7 +394,7 @@ public abstract class AbstractRelativeReferences {
   @NessieApiVersions(versions = NessieApiVersion.V2)
   public void assignReferenceFromDetachedUnambiguous() {
     NessieError error = expectError(rest().body(etl1).put("trees/@{hash}", c2 + "~1"), 400);
-    checkError(error, BAD_REQUEST, "Reference to assign from must be a branch or a tag.");
+    checkError(error, BAD_REQUEST, "Assignment target must be a branch or a tag.");
   }
 
   /**
@@ -405,7 +405,7 @@ public abstract class AbstractRelativeReferences {
   @NessieApiVersions(versions = NessieApiVersion.V2)
   public void assignReferenceFromDetachedAmbiguous() {
     NessieError error = expectError(rest().body(etl1).put("trees/@{hash}", "~1"), 400);
-    checkError(error, BAD_REQUEST, "Reference to assign from must be a branch or a tag.");
+    checkError(error, BAD_REQUEST, "Assignment target must be a branch or a tag.");
   }
 
   /**
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
@@ -1501,7 +1501,7 @@ public abstract class AbstractRelativeReferences {
   @NessieApiVersions(versions = NessieApiVersion.V2)
   void getDiffFromRefAmbiguousDetached() {
     NessieError error = expectError(rest().get("trees/@{from}/diff/base@{to}", "~1", "~2"), 400);
-    checkError(error, BAD_REQUEST, "From hash must contain a starting commit ID.");
+    checkError(error, BAD_REQUEST, "\"From\" hash must contain a starting commit ID.");
   }
 
   /**
@@ -1512,7 +1512,7 @@ public abstract class AbstractRelativeReferences {
   @NessieApiVersions(versions = NessieApiVersion.V2)
   void getDiffToRefAmbiguousDetached() {
     NessieError error = expectError(rest().get("trees/base@{from}/diff/@{to}", "~1", "~2"), 400);
-    checkError(error, BAD_REQUEST, "To hash must contain a starting commit ID.");
+    checkError(error, BAD_REQUEST, "\"To\" hash must contain a starting commit ID.");
   }
 
   private static Branch expectBranch(Response base) {
--- a/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
+++ b/servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
@@ -24,6 +24,7 @@ import io.restassured.response.Response;
 import io.restassured.response.ValidatableResponse;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
@@ -48,6 +49,7 @@ import org.projectnessie.model.LogResponse;
 import org.projectnessie.model.MergeResponse;
 import org.projectnessie.model.Reference;
 import org.projectnessie.model.SingleReferenceResponse;
+import org.projectnessie.model.Tag;
 
 /** A set of tests specifically aimed at testing relative references in API v2. */
 public abstract class AbstractRelativeReferences {
@@ -87,6 +89,8 @@ public abstract class AbstractRelativeReferences {
          etl1          c3
                          \
          etl2             c4 -- c5
+                           |
+                          tag1
     */
     assumeTrue(outer.isNewModel());
     Branch base = outer.createBranchV2("base");
@@ -100,19 +104,20 @@ public abstract class AbstractRelativeReferences {
     Branch etl2 = outer.createBranchV2("etl2", etl1);
     etl2 = outer.commitV2(etl2, key4, table4);
     c4 = etl2.getHash();
+    outer.createTagV2("tag1", etl2);
     etl2 = outer.commitV2(etl2, key5, table5);
     c5 = etl2.getHash();
   }
 
   /**
-   * Can create a reference from a hash that is unambiguous. Expect a 200 response with the created
-   * reference.
+   * Can create a reference from a branch + unambiguous target hash. Expect a 200 response with the
+   * created reference.
    */
   @Test
   @NessieApiVersions(versions = NessieApiVersion.V2)
-  public void createReferenceUnambiguous() {
+  public void createReferenceFromBranchUnambiguous() {
     Reference reference =
-        expectReference(
+        expectBranch(
             rest()
                 .queryParam("name", "branch1")
                 .queryParam("type", "branch")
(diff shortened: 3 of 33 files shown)


servers/services/src/main/java/org/projectnessie/services/hash/HashResolver.java
```diff
@@ -94,12 +97,8 @@ public ResolvedHash resolveHashOnRef(
         ReferenceInfo<CommitMeta> refInfo = store.getNamedRef(namedRef, GetNamedRefsParams.DEFAULT);
         ref = refInfo.getNamedRef();
         currentHead = refInfo.getHash();
-        // Shortcut: use HEAD, if hashOnRef is null or empty
-        if (hashOnRef == null || hashOnRef.isEmpty()) {
```
adutra, line 98: This shortcut was bypassing the validation rules so I had to get rid of it.

servers/services/src/main/java/org/projectnessie/services/hash/HashValidator.java
```diff
@@ -17,50 +17,122 @@
...
 
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import javax.annotation.Nullable;
+import org.projectnessie.versioned.BranchName;
+import org.projectnessie.versioned.NamedRef;
+import org.projectnessie.versioned.TagName;
 
-@FunctionalInterface
-public interface HashValidator {
+public final class HashValidator {
```
adutra, line 26: The refactoring is mainly motivated by the need to preserve the field names when generating error messages. Otherwise it's functionally equivalent.

servers/services/src/main/java/org/projectnessie/services/hash/ResolvedHash.java
```diff
@@ -36,29 +34,13 @@ public interface ResolvedHash extends WithHash<NamedRef> {
...
-  /**
-   * The user-provided hash. Only present if a hash was actually provided and successfully parsed;
-   * will be empty if the hash was implicitly resolved to the ref's HEAD.
-   *
-   * <p>If this is empty, then {@link #getHead()} is guaranteed to be non-empty.
-   *
-   * <p>If the provided hash had relative specs, this hash will be the result of resolving those
-   * specs against its absolute part, or the ref's HEAD if it doesn't have an absolute part.
-   */
-  Optional<Hash> getProvidedHash();
```
adutra, line 48: Not needed anymore.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -308,16 +305,22 @@ public Reference assignReference(
...
+                      .refMustBeBranchOrTag()
+                      .hashMustBePresent()
+                      .hashMustNotBeAmbiguous());
       ResolvedHash newRef =
           getHashResolver()
               .resolveHashOnRef(
-                  assignTo.getName(), assignTo.getHash(), "Target hash", UNAMBIGUOUS_HASH);
+                  assignTo.getName(),
+                  assignTo.getHash(),
+                  new HashValidator("Target hash").hashMustBePresent().hashMustNotBeAmbiguous());
```
adutra, line 323: And this is the core fix for `assignReference`.

servers/services/src/main/java/org/projectnessie/services/spi/TreeService.java
```diff
@@ -198,8 +198,6 @@ MergeResponse transplantCommitsIntoBranch(
           String branchName,
       @Valid
           @jakarta.validation.Valid
-          @NotNull
-          @jakarta.validation.constraints.NotNull
```
adutra, line 202: It's better to let the hash validator generate the error messages, they will be more homogenous and more informational.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ as necessary. Empty sections will not end in the release notes.
 ### Changes
 - Java client API to assign/delete reference operations without specifying a concrete reference type
   (no change to REST API).
+- Creating and assigning references now require a target hash to be specified.
```
snazy, line 23:
```suggestion
- Creating and assigning references now requires a target hash to be specified.
```

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -330,7 +333,7 @@ public Reference assignReference(
           .canAssignRefToHash(oldRef.getNamedRef())
           .checkAndThrow();
 
-      getStore().assign(oldRef.getNamedRef(), oldRef.getProvidedHash(), newRef.getHash());
+      getStore().assign(oldRef.getNamedRef(), Optional.of(oldRef.getHash()), newRef.getHash());
```
snazy, line 336: I suspect, with this change, the `Optional` in the signatures of `VersionStore.assign` + `VersionStore.create` can go away.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
```diff
@@ -87,13 +89,16 @@ void createFixtures() {
...
+                                 ^
+                               tag1
     */
     assumeTrue(outer.isNewModel());
     Branch base = outer.createBranchV2("base");
     base = outer.commitV2(base, key1, table1);
     c1 = base.getHash();
     base = outer.commitV2(base, key2, table2);
     c2 = base.getHash();
+    outer.createTagV2("tag0", base);
```
snazy, line 101: Looks like `tag0` is missing in the diagram above (`tag1`'s added).

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
```diff
@@ -123,143 +129,331 @@ public void createReferenceUnambiguous() {
...
   /**
-   * Can create a reference from a hash that is unambiguous and DETACHED. Expect a 200 response with
-   * the created reference.
+   * Cannot create a reference from a branch + hash that is ambiguous (writing operation). The hash
+   * must contain a starting commit ID. Expect a BAD_REQUEST error.
    */
   @Test
   @NessieApiVersions(versions = NessieApiVersion.V2)
-  public void createReferenceUnambiguousDetached() {
+  public void createReferenceFromBranchAmbiguous() {
```
snazy, line 137: Wonder if we need these tests as "pure REST tests" instead of using the Nessie API. I think, it's better to use the Nessie API and only have "pure REST tests" for test cases that do not work via the API.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
adutra: Actually we won't need `tag0` anymore, I forgot to remove it.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
adutra: I actually went for REST tests precisely because we can test any test case by forging the HTTP request by hand. Some of the test cases would indeed be untestable through the API.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: Yes. Do you prefer that I make the change now, or in a follow-up PR?

api/model/src/main/java/org/projectnessie/api/v2/params/Merge.java
```diff
@@ -93,14 +92,14 @@ public interface Merge extends BaseMergeTransplant {
...
-  @NotBlank
-  @jakarta.validation.constraints.NotBlank
   @Pattern(
       regexp = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_REGEX,
       message = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_MESSAGE)
   @jakarta.validation.constraints.Pattern(
       regexp = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_REGEX,
       message = Validation.HASH_OR_RELATIVE_COMMIT_SPEC_MESSAGE)
+  @Nullable
+  @jakarta.annotation.Nullable
```
adutra, line 102: Same here: goal is to let `HashValidator` produce the error message.

adutra:
NesQuEIT test are passing now.

servers/services/src/main/java/org/projectnessie/services/hash/ResolvedHash.java
```diff
@@ -36,29 +34,13 @@ public interface ResolvedHash extends WithHash<NamedRef> {
...
-   *
-   * <p>If this is empty, then {@link #getProvidedHash()} is guaranteed to be non-empty.
    */
   Optional<Hash> getHead();
 
-  /**
-   * The effective resolved hash, never {@code null}. Will either be {@link #getProvidedHash()} if
-   * present, or {@link #getHead()} otherwise.
-   */
+  /** The effective resolved hash, never {@code null}. */
```
dimas-b, line 43: nit: why not annotate as `@Nonnull`?

servers/services/src/main/java/org/projectnessie/services/hash/HashValidator.java
```diff
@@ -17,50 +17,94 @@
...
    */
-  default HashValidator and(HashValidator other) {
-    return (name, parsed) -> {
-      validate(name, parsed);
-      other.validate(name, parsed);
-    };
+  @CanIgnoreReturnValue
+  public HashValidator hashMustNotBeAmbiguous() {
+    validator =
+        validator.andThen(
```
dimas-b, line 102: It looks like top-level services use this method and `HashResolver` can also call this, which will cause duplicate conditions. Operation will still be correct, but it feels awkward... Would it be simpler to just keep several validation flags?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -308,16 +305,22 @@ public Reference assignReference(
...
     try {
-      checkArgument(expectedHash != null, "Expected hash must be provided");
 
       ResolvedHash oldRef =
           getHashResolver()
               .resolveHashOnRef(
-                  referenceName, expectedHash, "Expected hash", REQUIRED_UNAMBIGUOUS_HASH);
+                  referenceName,
+                  expectedHash,
+                  new HashValidator("Reference to assign from", "Expected hash")
```
dimas-b, line 314: maybe just `Reference to assign`?.. not sure I get the meaning of `from` in the message :thinking:

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: Let's remove `Optional` now, because this PR make the hash required and `Optional` unnecessary.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -710,11 +718,19 @@ public MergeResponse mergeRefIntoBranch(
 
       ResolvedHash fromRef =
           getHashResolver()
-              .resolveHashOnRef(fromRefName, fromHash, "Source hash", REQUIRED_UNAMBIGUOUS_HASH);
+              .resolveHashOnRef(
+                  fromRefName,
+                  fromHash,
+                  new HashValidator("Source hash").hashMustBePresent().hashMustNotBeAmbiguous());
```
dimas-b, line 724: nit: maybe add a helper method like `nonAmbiguousHash(String fieldname)`?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
dimas-b: I agree. The java client API seems weaker than the REST API in terms of expressing relative references. It might be worth adding specific methods for "merge parent #N", "ancestor #Y", etc., where hashes are manipulated.

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
```diff
@@ -803,11 +1515,16 @@ void getDiffToRefAmbiguousDetached() {
     checkError(error, BAD_REQUEST, "To hash must contain a starting commit ID.");
```
dimas-b, line 1515: Would it make sense to add quotation marks around `To`, i.e. `"To" hash ...`? In current form the whole message is not very readable, IMHO.

This is not really related to the purpose of this PR, but we're changing how validation messages are constructed :thinking:

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
snazy: > Some of the test cases would indeed be untestable through the API.

fair

servers/services/src/main/java/org/projectnessie/services/hash/ResolvedHash.java
adutra: The Immutables library has no support for `@Nonnull`: it encourages instead to consider everything as non-null and annotate the exceptions with `@Nullable` or use `Optional`. See http://immutables.github.io/immutable.html#nullable-attributes.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: Now that `hashMustBePresent` has been merged with `hashMustNotBeAmbiguous` maybe that's not required anymore?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: Done. We can also make a case for removing the `Optional` in:

* `VersionStore.commit.referenceHash`
* `MergeOp.expectedHash`
* `TransplantOp.expectedHash`

But if everybody agrees, I will look into that in another PR :-)

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/AbstractRelativeReferences.java
adutra: I used `"To" hash` as suggested... if that's not enough, we could maybe use `Hash to start the diff from ...` wdyt?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: `Assignment target` SGTM

dimas-b (approved):
LGTM :+1: thx! :)

PR #7399, opened 2023-08-22
Apply filters early in TreeApiImpl.getEntries

XN137:
`filterPredicate` will discard non-matching key entries in the final while loop however at that stage the entries (potentially including their content) have already been loaded and authz checks have been performed.

by applying the filters as early as possible we can avoid work that we will end up throwing away.

also fix a bug where `PersistVersionStore.getKeys` was ignoring `KeyRestrictions.contentKeyPredicate`.

also unify `ExportImport` tests where the payload had a different `contentId` than the key objects (and would now fail the new assertions in `KeyEntry.of`).

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -893,11 +893,13 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
         }
       }
 
-      int filterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
-      Predicate<KeyEntry> filterPredicate = filterEntries(filter);
+      Predicate<KeyEntry> paramFilterPredicate = filterEntries(filter);
+      final int filterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
       if (filterDepth > 0) {
-        filterPredicate = filterPredicate.and(e -> e.getKey().elements().size() >= filterDepth);
+        paramFilterPredicate =
+            paramFilterPredicate.and(e -> e.getKey().elements().size() >= filterDepth);
       }
+      final Predicate<KeyEntry> filterPredicate = paramFilterPredicate;
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
@@ -917,6 +919,9 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
                 entries, super::startAccessCheck, getServerConfig().accessChecksBatchSize()) {
               @Override
               protected Set<Check> checksForEntry(KeyEntry entry) {
+                if (!filterPredicate.test(entry)) {
+                  return Collections.emptySet();
+                }
                 return singleton(canReadContentKey(refWithHash.getValue(), entry.getKey()));
               }
             }.initialCheck(canReadEntries(refWithHash.getValue()));
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -893,6 +893,8 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
         }
       }
 
+      // apply filter as early as possible to avoid work (i.e. content loading, authz checks)
+      // for entries that we will eventually throw away
       final int namespaceFilterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
       if (namespaceFilterDepth > 0) {
         Predicate<ContentKey> depthFilter = e -> e.getElementCount() >= namespaceFilterDepth;
@@ -900,14 +902,14 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
             contentKeyPredicate == null ? depthFilter : contentKeyPredicate.and(depthFilter);
       }
       final Predicate<KeyEntry> filterPredicate = filterEntries(filter);
-      final Predicate<KeyEntry> withContentPredicate = withContent ? filterPredicate : null;
+      final Predicate<KeyEntry> loadContentPredicate = withContent ? filterPredicate : null;
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
               .getKeys(
                   refWithHash.getHash(),
                   pagingToken,
-                  withContentPredicate,
+                  loadContentPredicate,
                   VersionStore.KeyRestrictions.builder()
                       .minKey(minKey)
                       .maxKey(maxKey)
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -893,14 +893,18 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
         }
       }
 
+      int filterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
       Predicate<KeyEntry> filterPredicate = filterEntries(filter);
+      if (filterDepth > 0) {
+        filterPredicate = filterPredicate.and(e -> e.getKey().elements().size() >= filterDepth);
+      }
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
               .getKeys(
                   refWithHash.getHash(),
                   pagingToken,
-                  withContent,
+                  withContent ? filterPredicate : null,
                   VersionStore.KeyRestrictions.builder()
                       .minKey(minKey)
                       .maxKey(maxKey)
@@ -917,9 +921,7 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
               }
             }.initialCheck(canReadEntries(refWithHash.getValue()));
 
-        if (namespaceDepth != null && namespaceDepth > 0) {
-          int depth = namespaceDepth;
-          filterPredicate = filterPredicate.and(e -> e.getKey().elements().size() >= depth);
+        if (filterDepth > 0) {
           Set<ContentKey> seen = new HashSet<>();
           while (authz.hasNext()) {
             KeyEntry key = authz.next();
@@ -937,7 +939,7 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
                         key.getKey().type(),
                         key.getKey().lastElement().contentId());
 
-            entry = maybeTruncateToDepth(entry, depth);
+            entry = maybeTruncateToDepth(entry, filterDepth);
 
             // add implicit namespace entries only once (single parent of multiple real entries)
             if (seen.add(entry.getName())) {
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -893,20 +893,21 @@ public class TreeApiImpl extends BaseApiImpl implements TreeService {
         }
       }
 
-      Predicate<KeyEntry> paramFilterPredicate = filterEntries(filter);
-      final int filterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
-      if (filterDepth > 0) {
-        paramFilterPredicate =
-            paramFilterPredicate.and(e -> e.getKey().elements().size() >= filterDepth);
+      final int namespaceFilterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
+      if (namespaceFilterDepth > 0) {
+        Predicate<ContentKey> depthFilter = e -> e.getElementCount() >= namespaceFilterDepth;
+        contentKeyPredicate =
+            contentKeyPredicate == null ? depthFilter : contentKeyPredicate.and(depthFilter);
       }
-      final Predicate<KeyEntry> filterPredicate = paramFilterPredicate;
+      final Predicate<KeyEntry> filterPredicate = filterEntries(filter);
+      final Predicate<KeyEntry> withContentPredicate = withContent ? filterPredicate : null;
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
               .getKeys(
                   refWithHash.getHash(),
                   pagingToken,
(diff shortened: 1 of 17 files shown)


versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -591,20 +594,32 @@ public PaginationIterator<KeyEntry> getKeys(
...
             }
-
-            UUID contentId = commitOp.contentId();
-            String contentIdString =
-                contentId != null ? contentId.toString() : contentIdFromContent(commitOp);
-            return KeyEntry.of(
-                buildIdentifiedKey(key, index, contentType, contentIdString, x -> null));
+            KeyEntry keyEntry =
+                KeyEntry.of(
+                    buildIdentifiedKey(key, index, contentType, contentIdString, x -> null));
```
XN137, line 612: assumption is that building this "simple" key is cheap/in-memory

versioned/persist/store/src/main/java/org/projectnessie/versioned/persist/store/PersistVersionStore.java
```diff
@@ -562,7 +566,14 @@ public PaginationIterator<KeyEntry> getKeys(
     return new FilteringPaginationIterator<KeyListEntry, KeyEntry>(
         source.iterator(),
         entry -> {
-          if (withContent) {
+          KeyEntry keyEntry =
+              KeyEntry.of(
```
XN137, line 570: assumption is that building this "simple" key is cheap/in-memory

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -893,14 +893,18 @@ public <R> R getEntries(
...
+        filterPredicate = filterPredicate.and(e -> e.getKey().elements().size() >= filterDepth);
+      }
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
               .getKeys(
                   refWithHash.getHash(),
                   pagingToken,
-                  withContent,
+                  withContent ? filterPredicate : null,
```
adutra, line 907: I wonder if we shouldn't go one step further and merge `filterPredicate` with `KeyRestrictions.contentKeyPredicate`? The former operates on `KeyEntry`, the latter on `ContentKey`. But `KeyEntry` exposes `ContentKey` though the chain `KeyEntry.getKey().contentKey()`. So it should be possible to use just one predicate operating on `KeyEntry` and push it down to the Version Store.

XN137: yes but it means you have to "materialize" a `KeyEntry` before you can apply the predicate... thus you'd need to do what i do in the PR i.e. build the simple `KeyEntry` first, and if the predicate matches, fetch the content and build a new special `KeyEntry` that also includes the content.

by merging those predicates we wouldnt have any immediate benefits in terms of preventing backend lookups + in-memory computations i think?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -893,14 +893,20 @@ public <R> R getEntries(
         }
       }
 
-      Predicate<KeyEntry> filterPredicate = filterEntries(filter);
+      Predicate<KeyEntry> paramFilterPredicate = filterEntries(filter);
```
XN137, line 896: note that `filterEntries` allows filtering on `KeyEntryForCel` only, which does not have any real `Content` properties, so the predicate can be applied early and doesnt need to wait for the `KeyEntry` with the `Content`

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: I meant rather the opposite, i.e. the predicate you are adding is based on the namespace depth, and that can be expressed as a `Predicate<ContentKey>`. 

Therefore it should be possible to merge that predicate into `contentKeyPredicate` directly without even modifying `VersionStore.getKeys()`. Something like:

```java
      int filterDepth = namespaceDepth == null ? 0 : namespaceDepth.intValue();
      if (filterDepth > 0) {
        Predicate<ContentKey> depthFilter = e -> e.getElementCount() >= filterDepth;
        contentKeyPredicate = contentKeyPredicate == null ? depthFilter : contentKeyPredicate.and(depthFilter);
      }
```

That would prevent any key outside of the minimal depth to be even processed.

However, an open question remains: should we also push down the entire `filterPredicate` to avoid fetching unwanted contents? I think this is a good idea as the benefits of pushing the predicate down should vastly outweigh the cost of creating a `KeyEntry` instance. Except that this is a `Predicate<KeyEntry>` and as you pointed out, we'd need to construct the `KeyEntry` instance before applying the filter, which could potentially involve fetching the content. This is a chicken-and-egg problem.

For that part, I think we could get away by modifying `KeyEntry.getContent()` and making it a `@Lazy` attribute. We did something similar for `org.projectnessie.versioned.Put`, see https://github.com/projectnessie/nessie/blob/f8b9dceda1d12ba819f8fe8f7e9c831cae765ed7/versioned/spi/src/main/java/org/projectnessie/versioned/Put.java#L38-L41.

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
adutra: Ah good point. Maybe we should group the attributes that can be filtered on in a separate entity that does not expose `getContent()` (maybe a superinterface is enough).

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: I think there's still benefit in adding the `filter` to `KeyRestrictions.contentKeyPredicate`. This way, the `authz` iterator does not have to have extra checks and runs everything through the checkers (which is nicer from the security perspective).

We have to make a `IdentifiedContentKey` for each entry anyway, and we have to make a `KeyEntryForCel` on top of that if a filter is present. If we can refactor the code such that the extra work is not performed when the filter is not present, overall pushing the filter down seem beneficial to me.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -591,20 +594,32 @@ public PaginationIterator<KeyEntry> getKeys(
...
+
+            if (withContentPredicate != null && withContentPredicate.test(keyEntry)) {
+              if (content == null) {
+                content =
+                    contentMapping.fetchContent(
+                        requireNonNull(commitOp.value(), "Required value pointer is null"));
+              }
+              return KeyEntry.of(buildIdentifiedKey(key, index, content, x -> null), content);
```
dimas-b, line 620: `keyEntry.getKey()` could be reused instead of rebuilding it.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -591,20 +594,32 @@ public PaginationIterator<KeyEntry> getKeys(
...
-            if (withContent) {
-              Content c =
+            UUID contentId = commitOp.contentId();
+            String contentIdString;
+            if (contentId != null) {
+              contentIdString = contentId.toString();
+            } else {
+              // this should only be hit by imported legacy nessie repos
+              content =
                   contentMapping.fetchContent(
```
snazy, line 606: This one can probably be optimized by using bulk-fetching.
The current implementation is quite naive and fetches each `Content` individually - it would save quite some time and DB load when multiple `Content` objects are fetched at once using a bulk-operation. Requires some refactoring to have a 2-stage `PaginationIterator`.

XN137: yeah i had discussed the same with @dimas-b yesterday but due to the larger refactoring we first fix the unnecessary work and can look at doing bulk loading in a follow-up

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
XN137: ah ok sorry for misunderstand, i've added the filter to `KeyRestrictions.contentKeyPredicate` but turns out `PersistVersionStore.getKeys` was not using that value at all (so tests started failing), so added a fix for that as well

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
XN137: done

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -591,20 +595,32 @@ public PaginationIterator<KeyEntry> getKeys(
...
+            Content content = null;
 
-            if (withContent) {
-              Content c =
+            UUID contentId = commitOp.contentId();
+            String contentIdString;
+            if (contentId != null) {
+              contentIdString = contentId.toString();
+            } else {
+              // this should only be hit by imported legacy nessie repos
```
XN137, line 605: this comment is my understanding of why it might be null from [here](https://github.com/projectnessie/nessie/blob/aafa1868db660b397d6eda43c715ca1c0d122692/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/CommitOp.java#L50-L55)

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: IIRC, CEL does not allow reflection/introspection. It can only access what is declared, and that is in `KeyEntryForCel` (no getter for content).

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -913,57 +922,38 @@ public <R> R getEntries(
                 entries, super::startAccessCheck, getServerConfig().accessChecksBatchSize()) {
               @Override
               protected Set<Check> checksForEntry(KeyEntry entry) {
+                if (!filterPredicate.test(entry)) {
```
dimas-b, line 925: WDYT about making `AuthzPaginationIterator` extend `FilteringPaginationIterator`? Then, the filter can only be applied once (now twice - on lines 925 and 936).

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -893,14 +893,23 @@ public <R> R getEntries(
...
+      final Predicate<KeyEntry> filterPredicate = filterEntries(filter);
+      final Predicate<KeyEntry> loadContentPredicate = withContent ? filterPredicate : null;
 
       try (PaginationIterator<KeyEntry> entries =
           getStore()
               .getKeys(
                   refWithHash.getHash(),
                   pagingToken,
-                  withContent,
+                  loadContentPredicate,
```
dimas-b, line 912: WDYT about adding `Predicate<KeyEntry>` to `KeyRestrictions` as `keyEntryPredicate`? Then, "with content" can remain `boolean` and the authZ iterator does not have to change.

Granted, `VersionStore` will have two layers of filtering (one based on `ContentKey`, the other on `KeyEntry`) but it seems to yield simpler code at the service impl. layer :thinking:

adutra: I had a similar idea that I shared with @XN137 . That said I don't think the new predicate should be declared inside `KeyRestrictions` because it only applies to `getKeys` (not to `getDiffs`).

Also `KeyEntry` is probably not ideal for that because it needs to materialize the content in order to be instantiated. We need another type that basically exposes only `ContentKey` and `Content.Type`.

dimas-b: I agree, `KeyRestrictions` is not the right place for the new filter.

dimas-b: `KeyEntry` could be instantiated without content, but I'd be fine with introducing a more specific type for the sake of clarity.

versioned/spi/src/main/java/org/projectnessie/versioned/KeyEntry.java
```diff
@@ -40,6 +41,16 @@ static KeyEntry of(IdentifiedContentKey key) {
 
   static KeyEntry of(
       IdentifiedContentKey key, @NotNull @jakarta.validation.constraints.NotNull Content content) {
+    Preconditions.checkArgument(
```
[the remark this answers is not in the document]
XN137: https://github.com/projectnessie/nessie/pull/7420

versioned/spi/src/main/java/org/projectnessie/versioned/KeyEntry.java
```diff
@@ -40,6 +41,16 @@ static KeyEntry of(IdentifiedContentKey key) {
...
+        key.type().equals(content.getType()),
+        "Content type from key '%s' does not match actual type of content: %s",
+        key.type(),
+        content);
+    Preconditions.checkArgument(
+        key.lastElement().contentId().equals(content.getId()),
+        "Content id from key '%s' does not match actual id of content: %s",
+        key,
+        content);
     return builder().key(key).content(content).build();
```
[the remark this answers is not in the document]
XN137: https://github.com/projectnessie/nessie/pull/7420

XN137:
closing in favor of https://github.com/projectnessie/nessie/pull/7415

which i need rebase after https://github.com/projectnessie/nessie/pull/7420

PR #7402, opened 2023-08-23
CI: Free disk space for NesQuEIT

snazy:
See also https://github.com/projectnessie/query-engine-integration-tests/pull/373

This change also adds the "free disk space" action added to NesQuEIT, although it is currently unused by Nessie.


XN137 (approved):
+1 assuming its a 1-1 copy from the other PR

maybe PR title needs to be updated though

PR #7403, opened 2023-08-23
CI: Disable tests in Windows + macOS jobs

snazy:
Running tests on GH Windows+Mac runners was always brittle and recently never succeeded. The value of these tests is quite low, so CI for these environments is now limited to building Nessie and running only unit but no integration tests on macOS and no tests on Windows GH runners.

### DIFF
--- a/.github/workflows/ci-mac.yml
+++ b/.github/workflows/ci-mac.yml
@@ -35,55 +35,55 @@ jobs:
       CI_MAC: "true"
 
     steps:
-    - name: Brew update & upgrade
-      run: |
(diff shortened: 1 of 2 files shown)


XN137 (approved):
ok i guess but wondering whether we really need to comment it out instead of deleting it 
(if we ever bring it back we can look into git history)

dimas-b:
> wondering whether we really need to comment it out instead of deleting it

I think deleting is cleaner. We could leave a comment with the git hash / tag or the version that last had those tests enabled.

dimas-b (approved):

PR #7409, opened 2023-08-23
GC: Fix duplicate-key error w/ PostgreSQL

### LINKED ISSUE #7408: nessie-gc: Running mark phase against Postgres database results in duplicate rows
When running the `nessie-gc mark` command with a Postgres database as backing storage, the operation can fail due to duplicate rows in the `gc_live_set_contents` table.

**What happened**:
After attempting to run the mark phase, an error like the following occurs (only the last exception listed as that is the root cause):
```
Caused by: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "gc_live_set_contents_pkey"
  Detail: Key (live_set_id, content_id, commit_id)=(f8248937-1066-46c4-af57-c507df37398d, 201c5e9c-a664-43c3-8f42-e192969e6d80, 8c0bd90ff3f4cbb9cdafbcd368715e566bdfb33004f3d1fe0b4481b44fb0bc13) already exists.
```

The mark phase should run through without issue.

With a Postgres database running at `localhost:5432`, run the following command and observe the error message after a lot of debug messages for checking individual commits:
```
NESSIE_GC_JDBC_URL="jdbc:postgresql://localhost:5432/nessie_gc" NESSIE_GC_JDBC_USER=postgres NESSIE_GC_JDBC_PASSWORD=pass nessie-gc mark --jdbc-user=postgres
```

**Details**:
- Nessie server type (docker/built from source) and version: Docker (on Kubernetes) v0.65.0
- Client type (Ex: UI/Spark/pynessie ...) and version: nessie-gc CLI v0.67.0

I already looked into this a little myself with the current source of nessie-gc. It looks like the issue is in the `collectAllKeys` method:
https://github.com/projectnessie/nessie/blob/b3b28033285c210bdbb2879c0944630ae8e99ab1/gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java#L337-L346

So the iteration of all live references goes through smoothly, but once the first non-live commit is reached, this causes several (all?) commits to be added to the live set again.

In order to ensure there weren't any other issues, I added `ON CONFLICT DO NOTHING` to the query responsible for this:
https://github.com/projectnessie/nessie/blob/95bb6ccd59a80266d87183d930b4ee92815ec998/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java#L158-L161
The mark phase ran through in its entirety, but I am not at all confident in the results this generated. I'm not familiar enough with the GC to judge whether this "fix" is somehow right or should even be necessary.

snazy:
As reported in #7408, Nessie GC does not work correctly against PostgreSQL.

Although there is at least one test case that exercises the relevant code paths with the "failing" data scenarios, the tests did not fail due to a (now) obvious difference between H2 (used in the tests) and PostgreSQL. This change keeps the H2 in-memory tests and introduces ITs for PostgreSQL. The "difference" was actually H2 _not_ operating in PostgreSQL-mode.

`JdbcPersisxtenceSpi` did check for duplicate keys and should have ignored those via the `isIntegrityConstraintViolation(SQLException)` check, but that approach was wrong, because a failure marks the database transaction as "aborted", so having an `ON CONFLICT DO NOTHING` is the right approach. That change simplifies the code a little.

Fixes #7408

### DIFF
--- a/gc/gc-base-tests/src/main/java/org/projectnessie/gc/contents/tests/AbstractPersistenceSpi.java
+++ b/gc/gc-base-tests/src/main/java/org/projectnessie/gc/contents/tests/AbstractPersistenceSpi.java
@@ -391,6 +391,20 @@ public abstract class AbstractPersistenceSpi {
   }
 
   @Test
+  void identifyDuplicateContentReferences() throws Exception {
+    LiveSetVals vals1 = new LiveSetVals();
+    vals1.startIdentify();
+    soft.assertThat(persistenceSpi.addIdentifiedLiveContent(vals1.id, vals1.refs.stream()))
+        .isEqualTo(vals1.refs.size());
+    for (ContentReference ref : vals1.refs) {
+      soft.assertThat(persistenceSpi.addIdentifiedLiveContent(vals1.id, Stream.of(ref)))
+          .isEqualTo(0L);
+    }
+    soft.assertThat(persistenceSpi.addIdentifiedLiveContent(vals1.id, vals1.refs.stream()))
+        .isEqualTo(0L);
+  }
+
+  @Test
   void identifiedContents() throws Exception {
     LiveSetVals vals1 = new LiveSetVals();
     vals1.startIdentify();
--- a/gc/gc-repository-jdbc/build.gradle.kts
+++ b/gc/gc-repository-jdbc/build.gradle.kts
@@ -49,17 +49,23 @@ dependencies {
   compileOnly(platform(libs.jackson.bom))
   compileOnly("com.fasterxml.jackson.core:jackson-annotations")
 
-  testImplementation(project(":nessie-gc-base-tests"))
+  testFixturesApi(project(":nessie-gc-base"))
+  testFixturesApi(project(":nessie-gc-base-tests"))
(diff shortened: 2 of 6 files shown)


gc/gc-tool/src/test/java/org/projectnessie/gc/tool/TestCLI.java
```diff
@@ -63,7 +63,7 @@
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
 public class TestCLI {
 
-  public static final String JDBC_URL = "jdbc:h2:mem:nessie_gc;DB_CLOSE_DELAY=-1";
+  public static final String JDBC_URL = "jdbc:h2:mem:nessie_gc;MODE=PostgreSQL;DB_CLOSE_DELAY=-1";
```
dimas-b, line 66: Do we have any test using H2 in its "native" mode? Do we want such tests?

dimas-b (approved):

dimas-b (approved):

PR #7410, opened 2023-08-23
GC: Allow configuring gc-tool JDBC via env only

### LINKED ISSUE #7405: nessie-gc: Some JDBC command line argument is still required even if it's fully configured via environment variables
When using the JDBC capability of nessie-gc (e.g. with Postgres), the CLI appears to not be "clever enough" to get fully configured by present environment variables and instead requires one of the CLI arguments to still be specified.

**What happened**:
Running any command that requires a database connection with `NESSIE_GC_JDBC_*` options in the current environment but without any of the `--jdbc-*` command line arguments, I get the following error message:

```
Error: Missing required argument (specify one of these): ([--inmemory] | [[--jdbc-url=<url>] [--jdbc-properties[=<String=String>[,<String=String>...]...]]... [[--jdbc-user=<user>] [--jdbc-password=<password>]]])
```

I expected nessie-gc to get fully configured through the environment variables such that the command line arguments are not necessary anymore.

With a Postgres database running at `localhost:5432`, run the following command:
```bash
NESSIE_GC_JDBC_URL="jdbc:postgresql://localhost:5432/nessie_gc" NESSIE_GC_JDBC_USER=postgres NESSIE_GC_JDBC_PASSWORD=pass nessie-gc list
```
Observe the error message shown above.

**Details**:
- Nessie server type (docker/built from source) and version: Docker (on Kubernetes) v0.65.0
- Client type (Ex: UI/Spark/pynessie ...) and version: nessie-gc CLI v0.67.0

snazy:
When running the `nessie-gc` tool for example using `NESSIE_GC_JDBC_URL="jdbc:postgresql://localhost:5432/nessie_gc" NESSIE_GC_JDBC_USER=postgres NESSIE_GC_JDBC_PASSWORD=pass nessie-gc list`, it fails with `Error: Missing required argument (specify one of these): ([--inmemory] | [[--jdbc-url=<url>] [--jdbc-properties[=<String=String>[,<String=String>...]...]]... [[--jdbc-user=<user>] [--jdbc-password=<password>]]])`.

The `EnvironmentDefaultProvider` gets correctly called, but that does not seem to mark the JDBC option group in `ContentsStorageGroup` as "present". As a workaround, specify the newly introduced `--jdbc` option.

Fixes #7405

### DIFF
--- a/gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/JdbcOptions.java
+++ b/gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/JdbcOptions.java
@@ -23,6 +23,10 @@ import org.projectnessie.gc.contents.jdbc.AgroalJdbcDataSourceProvider;
 import picocli.CommandLine;
 
 public class JdbcOptions {
+  @CommandLine.Option(
+      names = "--jdbc",
+      description = "Flag whether to use the JDBC contents storage.")
+  boolean jdbc;
 
   @CommandLine.Option(
       names = "--jdbc-url",
--- a/gc/gc-tool/src/test/java/org/projectnessie/gc/tool/TestCLI.java
+++ b/gc/gc-tool/src/test/java/org/projectnessie/gc/tool/TestCLI.java
@@ -82,19 +82,19 @@ public class TestCLI {
         // missing contents-storage option
         arguments(
             singletonList("gc"),
-            "Error: Missing required argument (specify one of these): ([--inmemory] | [--jdbc-url="),
+            "Error: Missing required argument (specify one of these): ([--inmemory] | [[--jdbc]"),
         arguments(
             singletonList("mark-live"),
-            "Error: Missing required argument (specify one of these): ([--inmemory] | [--jdbc-url="),
+            "Error: Missing required argument (specify one of these): ([--inmemory] | [[--jdbc]"),
         arguments(
             asList("sweep", "--live-set-id=00000000-0000-0000-0000-000000000000"),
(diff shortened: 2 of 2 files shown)


dimas-b (approved):

gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/options/JdbcOptions.java
```diff
@@ -23,6 +23,10 @@
 import picocli.CommandLine;
 
 public class JdbcOptions {
+  @CommandLine.Option(
+      names = "--jdbc",
+      description = "Flag whether to use the JDBC contents storage.")
+  boolean jdbc;
 
   @CommandLine.Option(
       names = "--jdbc-url",
```
dimas-b, line 32: I wonder if adding am explicit default value attribute of `${NESSIE_GC_JDBC_URL}` is going to make this option "set" :thinking:

dimas-b: cf. https://github.com/projectnessie/nessie/blob/main/tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/CommittingCommand.java#L26

snazy: The culprit is that it actually has the values via the defaults from the environment.

marvin-roesch:
Thanks for the quick action on the issue! Specifying `--jdbc` seems like a good compromise.

PR #7415, opened 2023-08-24
Pushdown filters in TreeApiImpl.getEntries

XN137:
`filterPredicate` will discard non-matching key entries in the final while loop however at that stage the entries (potentially including their content) have already been loaded and authz checks have been performed.

by applying the filters through a new `BiPredicate<ContentKey, Content.Type>` as early as possible we can avoid work that we will end up throwing away.

also fix `PersistVersionStore.getKeys` and `getDiffs` ignoring `KeyRestrictions.contentKeyPredicate`.

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
+++ b/servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
@@ -171,13 +171,14 @@ public final class CELUtil {
     if (model instanceof Operation) {
       return new OperationForCelImpl((Operation) model);
     }
-    if (model instanceof KeyEntry) {
-      return new KeyEntryForCelImpl((KeyEntry) model);
-    }
     if (model instanceof ContentKey) {
       return new KeyForCelImpl((ContentKey) model);
     }
     return model;
+  }
+
+  public static Object forCel(ContentKey key, Content.Type type) {
+    return new KeyEntryForCelImpl(key, type);
   }
 
   private static class KeyForCelImpl extends AbstractKeyedEntity {
@@ -198,26 +199,33 @@ public final class CELUtil {
     }
   }
 
+  /**
+   * the class does not wrap a {@link KeyEntry} because we need to be able to evaluate {@link
+   * java.util.function.BiPredicate}&lt;ContentKey, Content.Type&gt; as early as possible to avoid
+   * redundant work.
+   */
   private static class KeyEntryForCelImpl extends AbstractKeyedEntity implements KeyEntryForCel {
-    private final KeyEntry entry;
+    private final ContentKey key;
+    private final Content.Type type;
 
-    private KeyEntryForCelImpl(KeyEntry entry) {
-      this.entry = entry;
+    private KeyEntryForCelImpl(ContentKey key, Content.Type type) {
+      this.key = key;
+      this.type = type;
     }
 
     @Override
     protected ContentKey key() {
-      return entry.getKey().contentKey();
+      return key;
     }
 
     @Override
     public String getContentType() {
-      return entry.getKey().type().name();
+      return type.name();
     }
 
     @Override
     public String toString() {
-      return entry.toString();
+      return "KeyEntryForCelImpl{" + "key=" + key + ", type=" + type + "}";
     }
   }
 
--- a/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
+++ b/servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
@@ -57,6 +57,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
+import java.util.function.BiPredicate;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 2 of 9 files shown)


versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -859,12 +862,13 @@ public PaginationIterator<Diff> getDiffs(
...
 
-    Predicate<ContentKey> contentKeyPredicate = keyRestrictions.contentKeyPredicate();
+    ContentKeyPredicate contentKeyPredicate = keyRestrictions.contentKeyPredicate();
     Predicate<DiffEntry> keyPred =
         contentKeyPredicate != null
             ? d -> {
               ContentKey key = storeKeyToKey(d.key());
-              return key != null && contentKeyPredicate.test(key);
+              return key != null
+                  && contentKeyPredicate.test(key, () -> contentTypeForPayload(d.toPayload()));
```
adutra, line 871: I am not sure this is correct. We are applying the filter to the "to" payload, why not apply it to the "from" payload as well? Payloads can change, if the same key is reused to create two objects of different content types. That's why I said elsewhere that this new filter should only ne used in `getKeys`, not in `getDiffs`.

api/model/src/main/java/org/projectnessie/model/EntriesResponse.java
```diff
@@ -87,8 +87,8 @@ static Entry entry(ContentKey name, Content.Type type, String contentId) {
       return ImmutableEntry.of(name, type, contentId, null);
     }
 
-    static Entry entry(ContentKey name, Content.Type type, Content content) {
-      return ImmutableEntry.of(name, type, content.getId(), content);
+    static Entry entry(ContentKey name, Content content) {
```
adutra, line 90: You can remove the redundant cas on line 83 now.

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
```diff
@@ -171,9 +171,6 @@ public static Object forCel(Object model) {
     if (model instanceof Operation) {
       return new OperationForCelImpl((Operation) model);
     }
-    if (model instanceof KeyEntry) {
```
adutra, line 174: In order to keep the current way of doing things, how about we introduce a new class holding the attributes we need?

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
```diff
@@ -996,9 +987,9 @@ private static Entry maybeTruncateToDepth(Entry entry, int depth) {
    *
    * @param filter The filter to filter by
    */
-  protected Predicate<KeyEntry> filterEntries(String filter) {
+  protected BiPredicate<ContentKey, Content.Type> filterEntries(String filter) {
     if (Strings.isNullOrEmpty(filter)) {
-      return x -> true;
+      return null;
```
adutra, line 992: I suggested this change initially but I'm not sure anymore it buys us any good. Reverting this change will simplify lines 910-916.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -859,12 +864,15 @@ public PaginationIterator<Diff> getDiffs(
...
 
-    Predicate<ContentKey> contentKeyPredicate = keyRestrictions.contentKeyPredicate();
+    BiPredicate<ContentKey, Content.Type> contentKeyPredicate =
+        keyRestrictions.contentKeyPredicate();
     Predicate<DiffEntry> keyPred =
         contentKeyPredicate != null
             ? d -> {
               ContentKey key = storeKeyToKey(d.key());
-              return key != null && contentKeyPredicate.test(key);
+              return key != null
```
adutra, line 873: This looks better imho, and corresponds to what the old model does with `KeyFilterPredicate` (it's applied to both "from" and "to").

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
dimas-b: Maybe add a utility method for combining nullable filters?

dimas-b (commented):
LGTM overall :+1:

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
XN137: i think its helpful because it can avoid determining the contenttype if no predicate is present (instead of an always true predicate)

servers/services/src/main/java/org/projectnessie/services/cel/CELUtil.java
XN137: i am missing the advantages of the current way i.e. whats the benefit of having an Object -> Object function?

if i introduce a new class it will have identical fields as `KeyEntryForCelImpl`
`forCel(ContentKey, Type)` sounds OK so i'll add that

api/model/src/main/java/org/projectnessie/model/EntriesResponse.java
```diff
@@ -87,8 +87,8 @@ static Entry entry(ContentKey name, Content.Type type, String contentId) {
       return ImmutableEntry.of(name, type, contentId, null);
     }
 
-    static Entry entry(ContentKey name, Content.Type type, Content content) {
```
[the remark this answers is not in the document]
XN137: reverted this change then, since we now have a validation method

api/model/src/main/java/org/projectnessie/model/EntriesResponse.java
XN137: undid the change

servers/services/src/main/java/org/projectnessie/services/impl/TreeApiImpl.java
XN137: > Maybe add a utility method for combining nullable filters?

added... not sure if its much nicer :sweat_smile:

dimas-b: thx!

dimas-b (approved):

adutra (approved):

snazy (approved):

XN137:
thanks for the reviews and especially @adutra ideas for pushing things further (down :sweat_smile:)

PR #7416, opened 2023-08-24
BigTable: use Batcher consistently for bulk mutations

adutra:
Switching to `Batcher` for bulk mutations has shown significant performance improvements in `eraseRepositoriesTable` method.

This method has also been given 2 other minor improvements:

1. Use a "value strip" filter to remove all cells from returned rows (as we don't need them);
2. Avoid creating an interleave filter if only one repository prefix is being erased.

### DIFF
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackend.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackend.java
@@ -21,24 +21,25 @@ import static com.google.protobuf.ByteString.copyFromUtf8;
 import static java.util.Objects.requireNonNull;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.FAMILY_OBJS;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.FAMILY_REFS;
-import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.MAX_BULK_MUTATIONS;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.TABLE_OBJS;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.TABLE_REFS;
+import static org.projectnessie.versioned.storage.bigtable.BigTablePersist.apiException;
 
+import com.google.api.gax.batching.Batcher;
 import com.google.api.gax.rpc.ApiException;
 import com.google.api.gax.rpc.NotFoundException;
+import com.google.api.gax.rpc.ServerStream;
 import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
 import com.google.cloud.bigtable.admin.v2.models.CreateTableRequest;
 import com.google.cloud.bigtable.data.v2.BigtableDataClient;
-import com.google.cloud.bigtable.data.v2.models.BulkMutation;
 import com.google.cloud.bigtable.data.v2.models.Filters;
-import com.google.cloud.bigtable.data.v2.models.Mutation;
+import com.google.cloud.bigtable.data.v2.models.Filters.Filter;
 import com.google.cloud.bigtable.data.v2.models.Query;
 import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 import com.google.protobuf.ByteString;
 import java.util.ArrayDeque;
 import java.util.Collection;
-import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackend.java
```diff
@@ -192,43 +193,35 @@ private void eraseRepositoriesNoAdminClient(Collection<String> repositoryIds) {
...
+      for (Row row : rows) {
+        batcher.add(RowMutationEntry.create(row.getKey()).deleteRow());
       }
-      paginator.advance(lastKey);
-      rows = dataClient.readRows(nextQuery(paginator)).iterator();
-    }
-
-    if (bulkDelete.getEntryCount() > 0) {
-      dataClient.bulkMutateRows(bulkDelete);
+      batcher.sendOutstanding();
```
snazy, line 206: Why is `sendOutstanding` needed? Doesn't `.close()` do that implicitly? Similar for the other changes.

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackend.java
```diff
@@ -192,43 +193,35 @@ private void eraseRepositoriesNoAdminClient(Collection<String> repositoryIds) {
...
-    }
-
-    Query.QueryPaginator paginator = Query.create(tableId).filter(repoFilter).createPaginator(100);
-    Iterator<Row> rows = dataClient.readRows(nextQuery(paginator)).iterator();
-    while (true) {
-      ByteString lastKey = null;
-      while (rows.hasNext()) {
-        Row row = rows.next();
-        lastKey = row.getKey();
-        if (prefixes.stream().anyMatch(prefix -> row.getKey().startsWith(prefix))) {
```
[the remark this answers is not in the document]
snazy: IIRC there was some issue, but I don't recall what exactly.

Can you add a new test similar to `AbstractBackendRepositoryTests.createEraseManyRepos` that erases repos iteratively (like 2 of 5 or so) and checks that the filer's always working (aka exactly deleting what it's supposed to delete).

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackendFactory.java
```diff
@@ -65,6 +65,13 @@ public static void configureDataClient(BigtableDataSettings.Builder settings) {
       retrySettings.setMaxRetryDelay(maxRetryDelay);
     }
 
+    stubSettings
+        .bulkMutateRowsSettings()
+        .setBatchingSettings(
+            stubSettings.bulkMutateRowsSettings().getBatchingSettings().toBuilder()
+                .setElementCountThreshold((long) BigTableConstants.MAX_BULK_MUTATIONS)
```
dimas-b, line 72: Do you think it might be worth making this configurable?

dimas-b (commented):
Changes LGTM. Looking forward to approval from @snazy :)

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackend.java
adutra: Indeed `close()` seems enough, but I was trying to stick as close as possible to the examples [here](https://cloud.google.com/bigtable/docs/delete-data#delete-stream-batch) and they use `.flush()` at the end of the iteration (which is the blocking version of `sendOutstanding()`.

I think we can remove all of them.

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackendFactory.java
adutra: I thought about that too, but on the other hand, we know our schema, average row sizes etc., so is it really useful to let users tweak these internal things? The retry delay just above is also not configurable currently.

snazy (approved):

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTableBackendFactory.java
dimas-b: Actually, `MAX_BULK_MUTATIONS` is only 1000, so I think we're fine (For some reason I thought it was very large)

dimas-b (approved):

PR #7444, opened 2023-08-29
Process outer test instances in clientFactoryForContext

dimas-b:
Make NessieClientResolver.clientFactoryForContext() invoke configuration callbacks on all test instances (inner as well as outer).


snazy:
LGTM

PR #7449, opened 2023-08-30
Add `--full-hashes` option to `commits` command

adutra:
...because abbreviated hashes cannot be used as valid `--hash` parameters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,9 @@ as necessary. Empty sections will not end in the release notes.
 ### New Features
 - Content Generator tool: added new `--hash` parameter to `commits`, `content` and `entries` 
   commands.
-- Content Generator tool: commit hashes are now printed in full when running the `commits` command.
 
 ### Changes
+- Content Generator tool: commit hashes are now printed in full when running the `commits` command.
 
 ### Deprecations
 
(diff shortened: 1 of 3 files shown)


snazy (commented):
IMO we should always print the full hash and not an abbreviated one, because that cannot be used elsewhere.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ as necessary. Empty sections will not end in the release notes.
 ### New Features
 - Content Generator tool: added new `--hash` parameter to `commits`, `content` and `entries` 
   commands.
+- Content Generator tool: commit hashes are now printed in full when running the `commits` command.
```
snazy, line 17: Let's move this to `Changes` ;)

snazy (approved):

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/ReadCommits.java
```diff
@@ -53,7 +53,7 @@ public void execute() throws NessieNotFoundException {
                     .getOut()
                     .printf(
                         "%s\t%s\t%s [%s]\n",
-                        Objects.requireNonNull(commitMeta.getHash()).substring(0, 8),
+                        Objects.requireNonNull(commitMeta.getHash()),
```
snazy, line 56: Actually, you could remove the requireNonNull here now

dimas-b (approved):

PR #7452, opened 2023-08-30
Propagate index stripes from parent to child commits

adutra:
Not propagating index stripes in the commit chain can potentially make some entries unreachable.

I don't have a proper unit test yet, will try to add one soonish, but I manually verified that this PR fixes some broken imports containing a huge number of entries per commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ as necessary. Empty sections will not end in the release notes.
 ### Deprecations
 
 ### Fixes
-- Fixed potential index corruption when importing repositories using the new storage model that 
+- Fixed potential index corruption when importing repositories into the new storage model that 
   could cause some contents to become inaccessible.
 
 ### Commits
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ as necessary. Empty sections will not end in the release notes.
 ### Deprecations
 
 ### Fixes
+- Fixed potential index corruption when importing repositories using the new storage model that 
+  could cause some contents to become inaccessible.
 
 ### Commits
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,8 @@ as necessary. Empty sections will not end in the release notes.
 ### Deprecations
 
 ### Fixes
-- Fixed potential index corruption when importing repositories into the new storage model that 
-  could cause some contents to become inaccessible.
+- Fixed potential index corruption when importing repositories with many keys into the new storage 
+  model that could cause some contents to become inaccessible.
 
 ### Commits
 
--- a/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
+++ b/versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
@@ -18,6 +18,7 @@ package org.projectnessie.versioned.storage.commontests;
 import static com.google.common.collect.Lists.newArrayList;
 import static java.util.Collections.emptyList;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
 import static org.assertj.core.api.InstanceOfAssertFactories.list;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
 import static org.projectnessie.versioned.storage.common.indexes.StoreIndexElement.indexElement;
@@ -41,14 +42,16 @@ import static org.projectnessie.versioned.storage.common.persist.ObjId.randomObj
 import static org.projectnessie.versioned.storage.commontests.KeyIndexTestSet.basicIndexTestSet;
 
 import java.util.Optional;
-import java.util.Random;
 import java.util.UUID;
 import java.util.function.Supplier;
+import java.util.stream.IntStream;
 import org.assertj.core.api.SoftAssertions;
 import org.assertj.core.api.junit.jupiter.InjectSoftAssertions;
 import org.assertj.core.api.junit.jupiter.SoftAssertionsExtension;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.projectnessie.versioned.storage.common.config.StoreConfig.Adjustable;
+import org.projectnessie.versioned.storage.common.exceptions.ObjTooLargeException;
 import org.projectnessie.versioned.storage.common.indexes.StoreIndex;
 import org.projectnessie.versioned.storage.common.indexes.StoreIndexElement;
 import org.projectnessie.versioned.storage.common.indexes.StoreKey;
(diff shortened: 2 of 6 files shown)


snazy (commented):
Nice catch! That's a really serious issue.

The change looks good to me, however we should definitely have a test case for this one.
You can change the store config to set tiny site limits for the incremental and reference indexes (example values in TestReferenceIndexes.referenceIndexLifecycle). Tests should cover incremental-index spill-out and reference-index-splitting, so it ends with at least two reference-index-splits.

snazy:
Update: probably sufficient to extend the existing `TestIndexesLogicImpl#completeIndexesInCommitChain`

snazy:
Hm - maybe also worth to add a round-trip test using export/import as well in `:nessie-versioned-transfer`.

One thing I just noticed is that the `CommitObj.tail` list only contains the direct parent, not up to `parents-per-commit`, which can lead to inefficient commit-log fetching. If I'm not mistaken, this is also the case for the old model, but IMO we can ignore the old model.

Feels relatively easy to add the "too minimalistic `CommitObj.tail` list ion this PR, would need to update the same code piece.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
```diff
@@ -338,4 +341,68 @@ public void perCommitOps() {
...
+      Random random = new Random();
+
+      for (int j = 0; j < 100; j++) {
+
+        // Use large keys to increase chances of ObjTooLargeException which
+        // in turn will trigger the striped index logic.
+        String key =
+            random
+                .ints('0', 'z' + 1)
+                .limit(100)
```
snazy, line 375: The `Random` approach can, although unlikely, produce duplicate keys.
Why not go with something like `String key = "x"+i+SOME_LONG_CONSTANT`, which effectively eliminates prefix-encoding in the serialized index?

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
```diff
@@ -338,4 +341,68 @@ public void perCommitOps() {
 
     soft.assertThat(indexesLogic.commitOperations(commit)).containsExactly(add1, add2, remove1);
   }
+
+  @Test
+  void completeIndexesInCommitChainWithStripes() throws Exception {
```
snazy, line 346: You can get a `Persist` instance with tiny limits, see `TestIndexesLogicImpl.completeIndexesInCommitChain`. A long constant (see below) that fits maybe 3 or 4 times into the incremental index and maybe 5 commits, each adding 2 or 3 keys, should then be enough to trigger the spill-out and reference-index splitting.

Would be good to add an assertion that there are reference index splits.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
```diff
@@ -338,4 +341,68 @@ public void perCommitOps() {
...
+              .headers(EMPTY_COMMIT_HEADERS)
+              .incompleteIndex(true);
+      StoreIndex<CommitOp> index = newStoreIndex(CommitOp.COMMIT_OP_SERIALIZER);
+
+      Random random = new Random();
+
+      for (int j = 0; j < 100; j++) {
+
+        // Use large keys to increase chances of ObjTooLargeException which
+        // in turn will trigger the striped index logic.
```
dimas-b, line 371: Could we assert that stripes do exist?

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
```diff
@@ -338,4 +342,79 @@ public void perCommitOps() {
...
+    CommitObj current = null;
+    for (int i = 0; i < 5; i++) {
+
+      ObjId parentId = current == null ? EMPTY_OBJ_ID : current.id();
+
+      CommitObj.Builder c =
+          commitBuilder()
+              .id(randomObjId())
+              .addTail(parentId)
+              .created(System.nanoTime())
```
snazy, line 397: Nit: its microseconds since epoch, better fix it here, before some copies it elsewhere.

Or just add some constant timestamps like `TimeUnit.SECONDS.toMicros(i + 1)` or so.

versioned/storage/common-tests/src/main/java/org/projectnessie/versioned/storage/commontests/AbstractIndexesLogicTests.java
```diff
@@ -338,4 +342,79 @@ public void perCommitOps() {
...
+      StoreIndex<CommitOp> index = newStoreIndex(CommitOp.COMMIT_OP_SERIALIZER);
+
+      for (int j = 0; j < 3; j++) {
+        StoreKey k = key(keys[i][j]);
+        UUID cid = new UUID(123L, j);
+        ObjId v = randomObjId();
+        index.add(indexElement(k, commitOp(ADD, 42, v, cid)));
+      }
+
+      // incremental index has size 175, max configured is 200
```
snazy, line 412: Can you make that `200` a constant and add an assertion here, like this (I know, a bit paranoid):
```java
soft.assertThat(serializedSize * 4 / 3 > THAT_CONSTANT);
```

snazy (commented):
LGTM - a few nits though.

Let's not forget https://github.com/projectnessie/nessie/pull/7452#issuecomment-1699813472

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
```diff
@@ -537,26 +540,37 @@ void completeIndexesInCommitChain(
...
               .incompleteIndex(false)
               .referenceIndex(referenceIndex)
-              .incrementalIndex(newIndex.serialize())
-              .build();
-      newCommit = commitLogic.updateCommit(newCommit);
+              .referenceIndexStripes(indexStripes)
+              .incrementalIndex(newIndex.serialize());
+
+      if (parent != null) {
+        c.tail(Collections.singleton(parent.id()));
```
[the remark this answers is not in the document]
snazy: Yes, that's logically correct.

I'd just propose to call `c.tail()` only once with a pre-populated list (otherwise immutables is a bit inefficient - think: list resizing).

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
```diff
@@ -550,17 +553,24 @@ void completeIndexesInCommitChain(
 
       commitOperations(current).forEach(newIndex::add);
 
-      CommitObj newCommit =
+      Builder c =
```
snazy, line 556: Let's qualify the `Builder` as `CommitObj.Builder` (otherwise checkstyle should also complain).

snazy (commented):
One last comment - and we're good

snazy (approved):

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ as necessary. Empty sections will not end in the release notes.
 ### Deprecations
 
 ### Fixes
+- Fixed potential index corruption when importing repositories into the new storage model that
```
dimas-b, line 26: nit: maybe "... when importing repositories with many keys ..." (just to be more precise)?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/IndexesLogicImpl.java
```diff
@@ -537,26 +539,38 @@ void completeIndexesInCommitChain(
...
-      newCommit = commitLogic.updateCommit(newCommit);
+              .referenceIndexStripes(indexStripes)
+              .incrementalIndex(newIndex.serialize());
+
+      if (parent != null) {
+        int parents = Math.min(parentsPerCommit - 1, parent.tail().size());
+        List<ObjId> tail = new ArrayList<>(parents + 1);
+        tail.add(parent.id());
+        for (int j = 0; j < parents; j++) {
+          tail.add(parent.tail().get(j));
```
dimas-b, line 568: nit: I wonder if `.addAll()` with a sub list might be more readable (if not more performant)?

dimas-b (approved):

dimas-b (approved):

PR #7478, opened 2023-09-06
MultiEnvTestEngine: append the environment name to test names

adutra:


testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
```diff
@@ -88,6 +81,12 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
             }
           });
 
+      // JupiterEngineDescriptor must be the root, that's what the JUnit Jupiter engine
+      // implementation expects.
+      JupiterEngineDescriptor multiEnvDescriptor =
+          new JupiterEngineDescriptor(
+              uniqueId, new DefaultJupiterConfiguration(new EmptyConfigurationParameters()));
```
dimas-b, line 88: Why not use `discoveryRequest.getConfigurationParameters()`?

dimas-b (commented):
Nice improvement overall. :+1:  Just one minor comment.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
adutra: This umbrella descriptor doesn't need configuration at all. The only thing we do with it is call `.addChild()`. We don't actually use it to collect any discovered tests using JUnit's discovery mechanism.

adutra:
Heads up: this PR doesn't seem to work unfortunately for XML reports. It does change the way IntelliJ and other tools show the names, but XML reports like the ones generated by Maven seem to use the result of `TestDescriptor.getLegacyReportingName()`, which is unfortunately not customizable :-(

We may want to still merge this, but it will be merely a nice-to-have. I would need to rephrase the changelog entry as well.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
dimas-b: What about engine selectors? (if we're running from JUnit 5 Runner CLI... hypothetically)

adutra: I think it wouldn't make any difference, but I don't mind putting back `discoveryRequest.getConfigurationParameters()` 👍

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ as necessary. Empty sections will not end in the release notes.
   `commits`, `references` and `entries` commands.
 - Helm charts: OpenTelemetry SDK is now completely disabled when tracing is disabled.
 - Helm charts: when auth is disabled, Quarkus OIDC doesn't print warnings anymore during startup.
+- `MultiEnvTestEngine`: this JUnit5 test engine used in multi-environment tests now reports the
```
snazy, line 24: Let's remove the changelog entry - this PR changes only test code, nothing user-facing.

dimas-b: I agree. The PR will be in the detailed list of changes (at least in GH) anyway.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
dimas-b: thx!

PR #7482, opened 2023-09-07
GC: Bump spark version for integration tests

ajantha-bhat:
I am trying to add row level delete test case for integration test. But it couldn't seem to work with existing spark-3.3 dependency. Hence, bumping to spark-3.4 (where it seems to work)

### DIFF
--- a/gc/gc-iceberg-inttest/src/intTest/java/org/projectnessie/gc/iceberg/inttest/ITSparkIcebergNessieS3.java
+++ b/gc/gc-iceberg-inttest/src/intTest/java/org/projectnessie/gc/iceberg/inttest/ITSparkIcebergNessieS3.java
@@ -91,6 +91,16 @@ public class ITSparkIcebergNessieS3 extends SparkSqlTestBase {
   }
 
   @Override
+  protected Map<String, String> sparkHadoop() {
+    Map<String, String> r = new HashMap<>();
+    r.put("fs.s3.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
+    r.put("fs.s3a.access.key", minio.accessKey());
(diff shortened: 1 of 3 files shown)


gc/gc-iceberg-inttest/src/intTest/java/org/projectnessie/gc/iceberg/inttest/ITSparkIcebergNessieS3.java
```diff
@@ -90,6 +90,16 @@ protected String warehouseURI() {
...
 
+  @Override
+  protected Map<String, String> sparkHadoop() {
+    Map<String, String> r = new HashMap<>();
+    r.put("fs.s3.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem");
+    r.put("fs.s3a.access.key", minio.accessKey());
+    r.put("fs.s3a.secret.key", minio.secretKey());
+    r.put("fs.s3a.endpoint", minio.s3endpoint());
+    return r;
+  }
```
ajantha-bhat, line 101: The AWS environment variables set from `nessieParams()` in line 109-112 is not working with spark-3.4. 
Iceberg can understand the env variables. But the error callstack looks like spark is directly trying to access the s3 bucket. Hence, the error. 

Setting the AWS credentials in a good old hadoop configuration for spark. 

Also observed that we have the similar code to set hadoop conf in `ITSparkIcebergNessieCLI` already.

ajantha-bhat: BTW the error was that `ITSparkIcebergNessieS3` was failing when spark was trying to access the minIO bucket while creating the namespace.

snazy (approved):

PR #7498, opened 2023-09-12
Fix behavior when swapping tables

snazy:
A DDL sequence like
```sql
CREATE TABLE t ...;
CREATE TABLE t_tmp ...;
ALTER TABLE t RENAME TO t_bak;
ALTER TABLE t_tmp RENAME TO t;
```
yields to the wrong error `New value for key 't' must not have a content ID`.

The root-cause is an incomplete check in `CommitImpl.commitAddPut()` that does not take the `CommitOp.Action.exists()` flag into account, letting the "new content must not have a content ID"-assertion firing.


dimas-b (approved):
Nice catch! :+1:

PR #7504, opened 2023-09-13
IntellJ workaround when using Iceberg snapshot builds

snazy:
This change adds a workaround for [this IntelliJ](https://youtrack.jetbrains.com/issue/IDEA-327421/IJ-fails-to-import-Gradle-project-with-dependency-with-classifier) issue.


settings.gradle.kts
```diff
@@ -47,6 +47,12 @@ dependencyResolutionManagement {
       name = "Apache Snapshots"
       url = URI("https://repository.apache.org/content/repositories/snapshots/")
       mavenContent { snapshotsOnly() }
+      metadataSources {
+        // Workaround for WORKAROUND for
```
dimas-b, line 51: nit: nested workaround? ;)

dimas-b (approved):

settings.gradle.kts
snazy: yup

PR #7524, opened 2023-09-19
Avoid re-running previously validated access checks on commit retries.

dimas-b:
This is mainly to speed up the retry cycle so as to improve throughput and reduce individual commit times.

Authorization is not dependent on commit history.

Races between authorization data changes and commits were possible even before this change, because there are no guarantees as to the relative timing of AuthZ checks (even without retries) and finilizing the commit.

### DIFF
--- a/servers/services/src/main/java/org/projectnessie/services/authz/RetriableAccessChecker.java
+++ b/servers/services/src/main/java/org/projectnessie/services/authz/RetriableAccessChecker.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2023 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.services.authz;
+
+import com.google.common.base.Preconditions;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Map;
+import java.util.function.Supplier;
+
+/**
+ * A utility class for performing access check in contexts where operations may have to be re-tried
+ * due to optimistic locking or similar mechanisms. Each {@link #newAttempt() attempt} forms a new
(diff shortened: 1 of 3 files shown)


servers/services/src/main/java/org/projectnessie/services/authz/RetriableAccessChecker.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public BatchAccessChecker newAttempt() {
+    return new Attempt();
+  }
+
+  private class Attempt extends AbstractBatchAccessChecker {
+    @Override
+    public Map<Check, String> check() {
+      Collection<Check> currentChecks =
+          new ArrayList<>(getChecks()); // shallow copy, elements are immutable
```
snazy, line 48: Why this new array-list? Isn't it sufficient to keep the instance returned by `getChecks()`?

servers/services/src/test/java/org/projectnessie/services/authz/TestRetriableAccessChecker.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.function.Supplier;
+import org.junit.jupiter.api.Test;
+import org.projectnessie.model.Content;
+import org.projectnessie.model.ContentKey;
+import org.projectnessie.model.IdentifiedContentKey;
+import org.projectnessie.versioned.BranchName;
+
+class TestRetriableAccessChecker {
+
+  private int checkCount;
```
snazy, line 35: Should this one be reset before each test?

snazy (commented):
Code looks good, a few comments

servers/services/src/main/java/org/projectnessie/services/authz/RetriableAccessChecker.java
dimas-b: Good point. The copy is an aftifact of the previous (unpublished) version.

servers/services/src/test/java/org/projectnessie/services/authz/TestRetriableAccessChecker.java
dimas-b: Each test gets a fresh instance of `TestRetriableAccessChecker` normally. If we ever switch to sharing the test instance, we could add a `@BeforeEach` method to reset the value.

servers/services/src/main/java/org/projectnessie/services/authz/RetriableAccessChecker.java
dimas-b: Actually, on a second though, I think it's worth keeping the shallow copy because the underlying collection is still exposed to modifications after the `check` method returns. Comments updated. WDYT?

snazy (approved):

PR #7525, opened 2023-09-20
CI: Use Java 21 in "newer java versions" workflow

snazy:


snazy:
But we should probably wait until https://github.com/mockito/mockito/issues/3121 is fixed.

adutra (approved):

PR #7529, opened 2023-09-20
GH actions - allow fork of Nessie in `projectnessie`

snazy:
Updates a few workflows to only run in `projectnessie/nessie` but not in another Nessie repo in the `projectnessie` org.


XN137 (approved):
lgtm but remind me, what other projects are we reusing the workflows in?

PR #7540, opened 2023-09-22
Store-cache: allow cache sizing based on max-heap-size

snazy:
This changes the way Nessie's object cache is configured. The previous default was to specify a fixed size in MB. When dealing with different JVM configurations / pods, it's easier to have a reasonable default of, as in this change, 70% of the Java max heap setting. This allows to have a single configuration that works in pretty much all situations/configurations.

If a fixed size is already configured, consider removing that explicit value and use fractional cache size setting based on the JVM's max heap size.


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/PersistProvider.java
```diff
@@ -137,12 +138,30 @@ public Persist producePersist() {
...
+    }
+    double cacheCapacityFraction = storeConfig.cacheCapacityFractionOfHeap();
+    if (cacheCapacityFraction > 0d) {
+      cacheSizing.fractionOfMaxHeapSize(cacheCapacityFraction);
+    }
+    int cacheCapacityFractionAdjustMB = storeConfig.cacheCapacityFractionAdjustMB();
+    if (cacheCapacityFractionAdjustMB > 0) {
+      cacheSizing.heapSizeAdjustmentMB(cacheCapacityFractionAdjustMB);
+    }
+    int effectiveCacheSizeMB = cacheSizing.build().calculateEffectiveSizeInMB();
```
dimas-b, line 158: nit: how about `CacheSizing.from(StoreConfig config)`?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CacheSizing.java
```diff
@@ -0,0 +1,109 @@
...
+    Builder heapSizeAdjustmentMB(int heapSizeAdjustmentMB);
+
+    CacheSizing build();
+  }
+
+  @Value.Check
+  default void check() {
+    checkState(
+        !(fixedSizeInMB().isPresent() && fractionOfMaxHeapSize().isPresent()),
+        "Cache sizing: must not specify both fractionOfMaxHeapSize and sizeInBytes, use either.");
```
dimas-b, line 91: This looks a bit too restrictive to me. Could we make one of the configuration sub-sets take precedence instead of making them mutually exclusive? That would simplify upgrades with Helm, I hope. Currently the old fix size settings are often present in `application.properties`, which ties them to binary artifact deployment.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/PersistProvider.java
snazy: Nope, it's a quarkus-config

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CacheSizing.java
snazy: Updated

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CacheSizing.java
```diff
@@ -0,0 +1,109 @@
...
+
+      long freeHeap = maxHeapInBytes - fractionAsBytes;
+      long minFree = heapSizeAdjustmentMB() * 1024L * 1024L;
+
+      long capacityInBytes = (minFree > freeHeap) ? maxHeapInBytes - minFree : fractionAsBytes;
+
+      if (fractionMinSizeMb().isPresent()) {
+        long fractionMin = (long) fractionMinSizeMb().getAsInt() * 1024 * 1024;
+        if (capacityInBytes < fractionMin) {
+          capacityInBytes = fractionMin;
```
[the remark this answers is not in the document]
snazy: MinFree defaults to 256M, minSize to 64M. MinFree would not work for tiny heap sizes (128M/256M).

dimas-b (approved):

PR #7547, opened 2023-09-25
BigTable: Use parallel reads for some objects to fetch

snazy:
Using parallel async reads can be better than using a single bulk-fetch.

This change updates `BigTablePersis.bulkFetch` to use parallel reads when up to 5 IDs are requested.

### DIFF
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
@@ -26,6 +26,7 @@ import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.CEL
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.FAMILY_OBJS;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.FAMILY_REFS;
 import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.MAX_BULK_READS;
+import static org.projectnessie.versioned.storage.bigtable.BigTableConstants.MAX_PARALLEL_READS;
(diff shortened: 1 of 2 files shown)


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -668,6 +669,44 @@ private <ID, R> void bulkFetch(
...
+      ID[] ids,
+      R[] r,
+      Function<ID, ByteString> keyGen,
+      Function<Row, R> resultGen,
+      Consumer<ID> notFound)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    int num = ids.length;
     @SuppressWarnings("unchecked")
     ApiFuture<Row>[] handles = new ApiFuture[num];
     int idx = 0;
```
adutra, line 712: Not related to this change, but I'm failing to understand why we have two nested for-loops below? But maybe I hadn't enough coffee this morning :-)

Imho `MAX_BULK_READS` should be used like `MAX_BULK_MUTATIONS`, i.e., when creating the client:

```java
    stubSettings.bulkReadRowsSettings().setBatchingSettings(
        stubSettings.bulkReadRowsSettings().getBatchingSettings().toBuilder()
            .setElementCountThreshold((long) BigTableConstants.MAX_BULK_READS)
            .build());
```

adutra (approved):

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
adutra: @snazy I can open a small PR for that, do you think it's worth it?

snazy: Sure, go ahead

PR #7563, opened 2023-09-28
APIv2: Properly throw an exception when using `GetEntriesBuilder.namespaceDepth`

snazy:


api/client/src/main/java/org/projectnessie/client/builder/BaseGetEntriesBuilder.java
```diff
@@ -95,8 +94,7 @@ public GetEntriesBuilder filter(String filter) {
 
   @Override
   public GetEntriesBuilder namespaceDepth(Integer namespaceDepth) {
-    this.namespaceDepth = namespaceDepth;
-    return this;
+    throw new UnsupportedOperationException("namespaceDepth is not supported for Nessie API v2");
```
dimas-b, line 97: WDYT about moving this exception to leaf sub-classes to minimize risk of affecting other builder implementations?

dimas-b (approved):

PR #7577, opened 2023-10-02
Tag all Nessie multi-environment tests

adutra:
Tagging multi-env tests can be useful e.g. to include or exclude multi-env tests.

With Maven:

```xml
  <plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
      <groups>nessie-multi-env</groups>
    </configuration>
  </plugin>
```

Or Gradle (using Groovy syntax):

```kotlin
test {
  useJUnitPlatform {
    includeTags 'nessie-multi-env'
  }
}
```

### DIFF
--- a/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
+++ b/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
@@ -29,6 +29,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
 import org.assertj.core.groups.Tuple;
+import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.api.NessieApiV2;
@@ -56,6 +57,7 @@ import org.projectnessie.tools.compatibility.api.VersionCondition;
 import org.projectnessie.versioned.storage.common.persist.ObjId;
 
 @VersionCondition(maxVersion = Version.NOT_CURRENT_STRING)
+@Tag("nessie-multi-env")
 public abstract class AbstractCompatibilityTests {
 
   @NessieAPI protected NessieApiV1 api;
--- a/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
+++ b/compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
@@ -29,7 +29,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Stream;
 import org.assertj.core.groups.Tuple;
-import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import org.projectnessie.client.api.NessieApiV1;
 import org.projectnessie.client.api.NessieApiV2;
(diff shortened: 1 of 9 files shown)


compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
```diff
@@ -56,6 +57,7 @@
 import org.projectnessie.versioned.storage.common.persist.ObjId;
 
 @VersionCondition(maxVersion = Version.NOT_CURRENT_STRING)
+@Tag("nessie-multi-env")
```
dimas-b, line 60: Are tags inherited from outer classes? I wish we could only put them where we have `
@ExtendWith(NessieUpgradesExtension.class)` or similar extensions.

dimas-b: Can we add tags from the engine?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
```diff
@@ -125,6 +125,7 @@
 import org.projectnessie.model.types.ImmutableGenericRepositoryConfig;
 
 /** Nessie-API tests. */
+@org.junit.jupiter.api.Tag("nessie-multi-env")
 @NessieApiVersions // all versions
```
dimas-b, line 129: Are tags inherited from `@NessieApiVersions`?

adutra: We _could_ enhance `@NessieApiVersions` as below:

```java
@Tag("nessie-multi-env")
@Target({ElementType.TYPE, ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
@ExtendWith(MultiVersionApiTest.class)
@Inherited
public @interface NessieApiVersions {
```

However this would work only if the test class is annotated with `@NessieApiVersions`. This is not mandatory: the annotation can also be placed on individual methods. If a test class contains only method-level occurrences of `@NessieApiVersions`, then ths test class would not be properly tagged.

compatibility/compatibility-tests/src/intTest/java/org/projectnessie/tools/compatibility/tests/AbstractCompatibilityTests.java
adutra: > Are tags inherited from outer classes?

Did you mean "from parent classes"? (This particular class is not inner/nested.)

If so, yes, which is why I placed the tag here, instead of in the 2 child classes: `ITOlderClients` and `ITOlderServers`.

Do you prefer to have the tag on each concrete test class?

> Can we add tags from the engine?

Apparently not, see https://github.com/junit-team/junit5/issues/2134.

dimas-b: > Do you prefer to have the tag on each concrete test class?
IMHO, it would be nice to add `@Tag("nessie-multi-env")` where we have `@ExtendWith(NessieUpgradesExtension.class)`. I suppose the intent will be clear to the reader.

It looks like `AbstractCompatibilityTests` is the only case where that is not so, so yes, I think it would be nice to move the tag from this class to its sub-classes: `ITOlderClients` and `ITOlderServers`. WDYT?

servers/jax-rs-tests/src/main/java/org/projectnessie/jaxrs/tests/BaseTestNessieApi.java
dimas-b: I see. I think it's fine to put `@Tag` on individual test classes.

dimas-b (approved):

PR #7597, opened 2023-10-05
BigTable backend: ability to disable telemetry

adutra:


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusBigTableConfig.java
```diff
@@ -65,4 +65,7 @@ public interface QuarkusBigTableConfig {
 
   @WithDefault("8086")
   int emulatorPort();
+
+  @WithDefault("true")
+  boolean enableTelemetry();
```
snazy, line 70: Can you add this to the docs on the website as well?

snazy (approved):

PR #7610, opened 2023-10-11
BigTable: reimplement scanAllObjects with server-side filtering

adutra:
This commit changes the BigTable implementation of scanAllObjects to use server-side instead of client-side filtering on object types.

This is implemented using a condition filter. Condition filters are not recommended generally because they are slower, but scanAllObjects is not meant to be particularly efficient. The fact that we are also limiting the query to a row prefix should alleviate the performance impact.

A quick experiment where I exported 1500 repos containing roughly 350K commits total showed a slight improvement in export time (export uses `scanAllObjects` with a filter on commit object types).

Without this fix: 2 minutes 46 seconds
With this fix: 2 minutes 24 seconds

### DIFF
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
@@ -52,7 +52,6 @@ import com.google.protobuf.ByteString;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
@@ -597,13 +596,10 @@ public class BigTablePersist implements Persist {
           all = false;
         }
       }
+      if (typeFilter == null) {
+        throw new IllegalArgumentException("No object types matched the provided predicate");
+      }
       if (!all) {
-        // Edge case: the predicate did not match any object type
-        if (typeFilter == null) {
-          this.paginator = null;
-          this.iter = Collections.emptyIterator();
-          return;
-        }
         // Condition filters are generally not recommended because they are slower, but
         // scanAllObjects is not meant to be particularly efficient. The fact that we are also
         // limiting the query to a row prefix should alleviate the performance impact.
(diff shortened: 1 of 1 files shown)


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -568,40 +568,50 @@ public CloseableIterator<Obj> scanAllObjects(
...
+                  .filter(FILTERS.qualifier().exactMatch(QUALIFIER_OBJ_TYPE))
+                  .filter(FILTERS.value().exactMatch(OBJ_TYPE_VALUES[type.ordinal()])));
+        } else {
+          all = false;
+        }
+      }
+      if (!all) {
+        // Edge case: the predicate did not match any object type
+        if (typeFilter == null) {
+          this.paginator = null;
```
snazy, line 603: Isn't this rather an error than an edge case?

snazy:
Were the times collected via the BT emulator or a real BT cluster?

adutra:
> Were the times collected via the BT emulator or a real BT cluster?

Real BT cluster, but with one single node. Do you want me to collect numbers using 3 or more nodes?

snazy:
> Real BT cluster, but with one single node. Do you want me to collect numbers using 3 or more nodes?

Nah - just wanted to know.

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
adutra: It's up for debate. Do you prefer to throw here?

snazy: It's either "all types" or an error IMO.

adutra: OK for error then.

snazy: It's rather an internal thingy - but an undefined situation ATM. Let's defer to another PR (I'd be in favor of falling back to "all types" actually).

PR #7626, opened 2023-10-13
 Set OAuth2Client idle when Nessie client is idle

adutra:
This PR introduces 2 mechanisms to mitigate the situation where the Nessie client is idle but the OAuth2 client keeps consuming resources:

1. An idle interval after which the tokens are not refreshed anymore. When the token is needed again, if it's expired, it will be renewed from step 1.
2. A keep alive interval after which the background thread is stopped. If a refresh is needed again, a new thread will be spawned.

It was hard to create tests for these new features. I will see if I can improve that more.

### DIFF
--- a/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
+++ b/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
@@ -15,7 +15,9 @@
  */
 package org.projectnessie.client.auth.oauth2;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
+import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -159,6 +161,33 @@ public class ITOAuth2Client {
       soft.assertThat(refreshedTokens.getRefreshToken()).isNotNull();
       tryUseAccessToken(validatingClient, refreshedTokens.getAccessToken());
       compareTokens(firstTokens, refreshedTokens, "Client2");
+    }
+  }
+
+  @Test
+  void testOAuth2ClientIdle() throws InterruptedException {
+    OAuth2ClientParams params =
+        clientParams("Client2")
+            .idleInterval(Duration.ofSeconds(5))
+            .keepAliveInterval(Duration.ofSeconds(5))
+            .build();
+    try (OAuth2Client client = new OAuth2Client(params);
+        HttpClient validatingClient = validatingHttpClient("Client2").build()) {
+      // will fetch initial tokens and schedule refresh in 5 seconds
+      // (10 secs lifespan - 5 secs safety window)
+      client.start();
+      assertThat(client.sleeping).isFalse();
+      Thread.sleep(5100); // will refresh tokens then enter sleep mode (idle interval 5 secs)
+      await().until(client.sleeping::get);
+      Thread.sleep(5100); // thread will exit (keep alive interval)
+      assertThat(((OAuth2TokenRefreshExecutor) client.executor).getPoolSize()).isZero();
+      Thread.sleep(5100); // access token will expire (10 secs lifespan)
+      // will wake up client, fetch new tokens immediately (on main thread),
+      // then schedule refresh, which will spawn a new thread
+      AccessToken accessToken = client.authenticate();
+      assertThat(client.sleeping).isFalse();
+      assertThat(((OAuth2TokenRefreshExecutor) client.executor).getPoolSize()).isOne();
+      tryUseAccessToken(validatingClient, accessToken);
     }
   }
 
--- a/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
+++ b/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
@@ -15,9 +15,7 @@
  */
 package org.projectnessie.client.auth.oauth2;
 
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
-import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -161,33 +159,6 @@ public class ITOAuth2Client {
       soft.assertThat(refreshedTokens.getRefreshToken()).isNotNull();
       tryUseAccessToken(validatingClient, refreshedTokens.getAccessToken());
       compareTokens(firstTokens, refreshedTokens, "Client2");
-    }
-  }
-
-  @Test
-  void testOAuth2ClientIdle() {
-    OAuth2ClientParams params =
-        clientParams("Client2")
-            .idleInterval(Duration.ofSeconds(1))
-            .keepAliveInterval(Duration.ofSeconds(1))
-            .refreshSafetyWindow(Duration.ofSeconds(1))
-            .build();
-    try (OAuth2Client client = new OAuth2Client(params);
-        HttpClient validatingClient = validatingHttpClient("Client2").build()) {
-      // will fetch initial tokens and schedule refresh in 4 seconds
-      // (5 secs lifespan - 1 secs safety window)
-      client.start();
-      assertThat(client.sleeping).isFalse();
-      // after refresh tokens, will enter sleep mode (idle interval was 1 sec)
-      await().until(client.sleeping::get);
-      // thread will exit after 1 sec (keep alive interval 1 sec)
-      await().until(() -> ((OAuth2TokenRefreshExecutor) client.executor).getPoolSize() == 0);
-      // access token is now expired; calling authenticate() will wake up client, fetch new tokens
-      // immediately (on main thread), then schedule refresh, which will in turn spawn a new thread
-      AccessToken accessToken = client.authenticate();
-      assertThat(client.sleeping).isFalse();
-      assertThat(((OAuth2TokenRefreshExecutor) client.executor).getPoolSize()).isOne();
-      tryUseAccessToken(validatingClient, accessToken);
     }
   }
 
--- a/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
+++ b/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
@@ -75,7 +75,7 @@ public class ITOAuth2Client {
         URI.create(KEYCLOAK.getAuthServerUrl() + "/realms/master/protocol/openid-connect/token");
     Keycloak keycloakAdmin = KEYCLOAK.getKeycloakAdminClient();
     master = keycloakAdmin.realms().realm("master");
-    updateMasterRealm(10, 15);
+    updateMasterRealm(5, 10);
     // Create 2 clients, one sending refresh tokens for client_credentials, the other one not
     createClient("Client1", false);
     createClient("Client2", true);
@@ -89,7 +89,7 @@ public class ITOAuth2Client {
    * This test exercises the OAuth2 client "in real life", that is, with background token refresh
    * running.
    *
-   * <p>For 20 seconds, 2 OAuth2 clients will strive to keep the access tokens valid; in the
+   * <p>For 15 seconds, 2 OAuth2 clients will strive to keep the access tokens valid; in the
    * meantime, another HTTP client will attempt to validate the obtained tokens.
    *
    * <p>This should be enough to exercise the OAuth2 client's background refresh logic with the 4
@@ -119,10 +119,10 @@ public class ITOAuth2Client {
                 tryUseAccessToken(validatingClient, client2.getCurrentTokens().getAccessToken());
               },
               0,
-              1,
-              TimeUnit.SECONDS);
+              500,
+              TimeUnit.MILLISECONDS);
       try {
-        future.get(20, TimeUnit.SECONDS);
+        future.get(15, TimeUnit.SECONDS);
       } catch (TimeoutException e) {
         // ok, expected for a ScheduledFuture
       } catch (ExecutionException e) {
@@ -165,25 +165,25 @@ public class ITOAuth2Client {
   }
 
   @Test
-  void testOAuth2ClientIdle() throws InterruptedException {
+  void testOAuth2ClientIdle() {
     OAuth2ClientParams params =
         clientParams("Client2")
-            .idleInterval(Duration.ofSeconds(5))
-            .keepAliveInterval(Duration.ofSeconds(5))
+            .idleInterval(Duration.ofSeconds(1))
+            .keepAliveInterval(Duration.ofSeconds(1))
+            .refreshSafetyWindow(Duration.ofSeconds(1))
             .build();
     try (OAuth2Client client = new OAuth2Client(params);
         HttpClient validatingClient = validatingHttpClient("Client2").build()) {
-      // will fetch initial tokens and schedule refresh in 5 seconds
-      // (10 secs lifespan - 5 secs safety window)
+      // will fetch initial tokens and schedule refresh in 4 seconds
+      // (5 secs lifespan - 1 secs safety window)
       client.start();
       assertThat(client.sleeping).isFalse();
-      Thread.sleep(5000); // will refresh tokens then enter sleep mode (idle interval 5 secs)
+      // after refresh tokens, will enter sleep mode (idle interval was 1 sec)
       await().until(client.sleeping::get);
-      Thread.sleep(5000); // thread will exit (keep alive interval)
+      // thread will exit after 1 sec (keep alive interval 1 sec)
       await().until(() -> ((OAuth2TokenRefreshExecutor) client.executor).getPoolSize() == 0);
-      Thread.sleep(5000); // access token will expire (10 secs lifespan)
-      // will wake up client, fetch new tokens immediately (on main thread),
-      // then schedule refresh, which will in turn spawn a new thread
+      // access token is now expired; calling authenticate() will wake up client, fetch new tokens
+      // immediately (on main thread), then schedule refresh, which will in turn spawn a new thread
       AccessToken accessToken = client.authenticate();
       assertThat(client.sleeping).isFalse();
       assertThat(((OAuth2TokenRefreshExecutor) client.executor).getPoolSize()).isOne();
@@ -385,9 +385,9 @@ public class ITOAuth2Client {
         .clientSecret("s3cr3t")
         .username("Alice")
         .password("s3cr3t")
-        .defaultAccessTokenLifespan(Duration.ofSeconds(10))
-        .defaultRefreshTokenLifespan(Duration.ofSeconds(15))
-        .refreshSafetyWindow(Duration.ofSeconds(5));
+        .defaultAccessTokenLifespan(Duration.ofSeconds(5))
+        .defaultRefreshTokenLifespan(Duration.ofSeconds(10))
+        .refreshSafetyWindow(Duration.ofSeconds(2));
   }
 
   @SuppressWarnings("SameParameterValue")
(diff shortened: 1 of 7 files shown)


api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -162,6 +164,33 @@ void testOAuth2ClientInitialRefreshToken() {
     }
   }
 
+  @Test
+  void testOAuth2ClientIdle() throws InterruptedException {
```
adutra, line 168: The test is admittedly a little lame, and takes 15 seconds. I wonder if it's necessary to keep it around. We could disable it maybe and only run when necessary.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -65,15 +65,18 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
   private final Duration defaultAccessTokenLifespan;
   private final Duration defaultRefreshTokenLifespan;
   private final Duration refreshSafetyWindow;
+  private final Duration idleInterval;
   private final boolean tokenExchangeEnabled;
   private final HttpClient httpClient;
-  private final ScheduledExecutorService executor;
+  /* Visible for testing. */ final ScheduledExecutorService executor;
```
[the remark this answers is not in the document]
adutra: That would bring Guava into compile scope, are we OK with that?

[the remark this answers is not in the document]
dimas-b: I initially thought a compile-only dependency would be fine, but now I'm thinking that it is likely to cause compilation warnings in downstream projects... I'm fine keeping the comment.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: The idea of injecting a Clock is nice, I will look into that. However, it will imho help mostly with unit tests. Here, just the clock won't be enough because we also need to control when the executor is going to run the tasks, and even worse, we can't really control the token expiration times that Keycloak creates.

adutra: I will try however to "shrink" the test execution time by making the overall delays more aggressive.

adutra: The unit test `TestOAuth2Client#testBackgroundRefresh`, which mocks both the executor + keycloak, now checks the sleeping flag.

The only extra thing we test here is the executor shrinking its core thread pool to zero.

I managed to reduce this test to ~5 secs btw.

Otherwise yes we could mock Keycloak but that would lead to something like Quarkus `OidcWiremockResource` which seems a lot of work for little gain, wdyt?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -65,15 +66,19 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
...
   private final boolean shouldCloseExecutor;
   private final ObjectMapper objectMapper;
   private final CompletableFuture<Void> started = new CompletableFuture<>();
+  /* Visible for testing. */ final AtomicBoolean sleeping = new AtomicBoolean();
+  private final Supplier<Instant> clock;
```
snazy, line 77: Why's this a supplier instead of a `Clock`? (We do have that threeten-extra package which provides a mutable clock impl).

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -162,6 +164,33 @@ void testOAuth2ClientInitialRefreshToken() {
...
+      client.start();
+      assertThat(client.sleeping).isFalse();
+      // after refresh tokens, will enter sleep mode (idle interval was 1 sec)
+      await().until(client.sleeping::get);
+      // thread will exit after 1 sec (keep alive interval 1 sec)
+      await().until(() -> ((OAuth2TokenRefreshExecutor) client.executor).getPoolSize() == 0);
+      // access token is now expired; calling authenticate() will wake up client, fetch new tokens
+      // immediately (on main thread), then schedule refresh, which will in turn spawn a new thread
+      AccessToken accessToken = client.authenticate();
+      assertThat(client.sleeping).isFalse();
```
dimas-b, line 188: This is potentially flaky as there's no barrier that prevents the client from sleeping again and the test execution could be delayed.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
dimas-b: Could we mock at the OAuth2 HTTP/REST level?... e.g. delegating `fetchNewTokens()` and similar methods to an impl. controlled by the test?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -165,35 +169,57 @@ public void close() {
...
 
-  private CompletionStage<Tokens> renewTokens() {
-    return currentTokensStage
-        // try refreshing the current tokens
-        .thenApply(this::refreshTokens)
-        // if that fails, try fetching brand-new tokens
-        .exceptionally(error -> fetchNewTokens())
-        .whenComplete(this::log);
+  /** Visible for testing. */
+  boolean idle(Instant now) {
```
snazy, line 214: Can you rename this to `shouldSleep`?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -165,35 +169,57 @@ public void close() {
...
+    return Duration.between(lastAccess, now).compareTo(idleInterval) > 0;
+  }
+
+  private void wakeUp(Instant now) {
+    scheduleOrExecuteTokensRenewal(getCurrentTokens(), now, Duration.ZERO);
   }
 
-  private void log(Tokens tokens, Throwable error) {
+  private void log(Tokens redacted, Throwable error) {
```
snazy, line 222: Should the `redacted` parameter go away? it's unused.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: That's what `TestOAuth2Client` does.

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -156,6 +156,30 @@ public final class NessieConfigConstants {
...
+   * href="https://en.wikipedia.org/wiki/ISO_8601#Durations">ISO-8601 duration</a>.
+   */
+  public static final String CONF_NESSIE_OAUTH2_IDLE_INTERVAL =
+      "nessie.authentication.oauth2.idle-interval";
+
+  public static final String DEFAULT_KEEP_ALIVE_INTERVAL = "PT30S";
+
+  /**
+   * Config property name ({@value #CONF_NESSIE_OAUTH2_KEEP_ALIVE_INTERVAL}) for the OAuth2
+   * authentication provider. The keep alive interval to use; if the OAuth2 provider background
```
snazy, line 175: I think the description should be different - as it's from a user's perspective.

Form a user's perspective it's the time the background thread stays around to speed up token-refreshes.
It's rather the time the Nessie client will proactively refresh tokens.

Actually, the name of the parameter should also be something around that meaning and not "keep alive" - maybe "idle client token refresh timeout" or so (also not a brilliant idea...).

Should mention that setting this to a "too high" value _could_ is not a good way (thread leak, memory leak - if clients are re-created).

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: Yes, at least in theory. I am leaning in favor of deleting this test. We don't need to test the keep alive stuff, it's not our code.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2TokenRefreshExecutor.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.time.Duration;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class OAuth2TokenRefreshExecutor extends ScheduledThreadPoolExecutor {
```
snazy, line 26: this class can be `final` (it's inner classes as well)

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
dimas-b: So can we do the sleep test there?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -165,35 +169,57 @@ public void close() {
...
+      renewTokens();
+    } else {
+      LOGGER.debug("Scheduling token refresh in {}", delay);
+      tokenRefreshFuture =
+          executor.schedule(this::renewTokens, delay.toMillis(), TimeUnit.MILLISECONDS);
+    }
+  }
+
+  private void renewTokens() {
+    CompletionStage<Tokens> oldTokensStage = currentTokensStage;
```
snazy, line 202: Nit: Should `oldTokansStage` be inlined?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -165,35 +169,57 @@ public void close() {
     }
   }
 
-  private void scheduleTokensRenewal(Tokens currentTokens) {
-    Instant now = Instant.now();
+  private void maybeScheduleTokensRenewal(Tokens currentTokens) {
+    Instant now = clock.get();
+    if (idle(now)) {
+      sleeping.set(true);
```
snazy, line 175: Probably unlikely, but I suspect this update can race with the conditional update in `authenticate()`. A user (indirectly) calls `authenticate()` - that thread gets stuck after the CAS - this function runs and sets sleeping back to true - and then the user thread resumes. It feels, that it can wake-up twice.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -165,35 +169,57 @@ public void close() {
...
-        // if that fails, try fetching brand-new tokens
-        .exceptionally(error -> fetchNewTokens())
-        .whenComplete(this::log);
+  /** Visible for testing. */
+  boolean idle(Instant now) {
+    return Duration.between(lastAccess, now).compareTo(idleInterval) > 0;
+  }
+
+  private void wakeUp(Instant now) {
+    scheduleOrExecuteTokensRenewal(getCurrentTokens(), now, Duration.ZERO);
```
snazy, line 219: Can you inline this to the (only) call site?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: Because it's more flexible than `Clock` and a `Clock` can be turned into a `Supplier<Instant>` with `clock::instant`.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: It makes it possible to use a method reference in `whenComplete(this::log)`, but why not.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: IJ gives a warning: "Non-atomic operation on volatile field 'currentTokensStage'".

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: I thought having a method would better self-document the intent, but sure. We can also remove `shouldSleep` in fact.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
snazy: Ah, NVM then

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
snazy: Oh, interesting.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: > this function runs

In which thread? I don't think that's possible: if `sleeping` was set to true, this means that the background thread pool is idle, and there is no refresh task scheduled for later execution.

It could happen though, that a user thread enters `authenticate()`, resets `sleeping` to false, then renew the tokens (which involves network IO), then gets stuck for several seconds somewhere, then enters `maybeScheduleTokensRenewal`, then resets `sleeping` to true again. But that's not a pathological code path.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: But... we already do: 

https://github.com/projectnessie/nessie/blob/a77c21ed9ef70d607834401885d738b59939af4d/api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java#L122-L130

Is there anything else you would like to test in that test class?

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: Alright, no objections so far: I'm removing this integration test.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -73,7 +73,7 @@ static void setUpKeycloak() {
         URI.create(KEYCLOAK.getAuthServerUrl() + "/realms/master/protocol/openid-connect/token");
     Keycloak keycloakAdmin = KEYCLOAK.getKeycloakAdminClient();
     master = keycloakAdmin.realms().realm("master");
-    updateMasterRealm(10, 15);
+    updateMasterRealm(5, 10);
```
adutra, line 76: I thought we could keep these changes though, it's nice to have a shorter test execution time.

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
dimas-b: `CONF_NESSIE_OAUTH2_IDLE_INTERVAL` -> `CONF_NESSIE_OAUTH2_PREEMPTIVE_TOKEN_REFRESH_IDLE_TIMEOUT` ? The "idle" part is a hint that the preemptive refresh will not happen when the client is idle.

`CONF_NESSIE_OAUTH2_BACKGROUND_THREAD_IDLE_TIMEOUT` - LGTM.

dimas-b: Also, I think we can keep the constant names shorter as they are not user-visible, but have descriptive property names (values of those constants).

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
dimas-b: Hmmm.... I must have missed it or got confused :facepalm:

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -83,28 +88,29 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
...
+    currentTokensStage.thenAccept(this::maybeScheduleTokensRenewal);
   }
 
   @Override
   public AccessToken authenticate() {
+    Instant now = clock.get();
+    lastAccess = now;
+    if (sleeping.compareAndSet(true, false)) {
+      LOGGER.debug("Waking up...");
+      scheduleOrExecuteTokensRenewal(getCurrentTokens(), now, Duration.ZERO);
```
dimas-b, line 112: if this throws, should we set `sleeping` back to `true` to force a renewal on the next `authenticate()` call?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2TokenRefreshExecutor.java
```diff
@@ -0,0 +1,62 @@
...
+      Thread thread = new TokenRefreshThread(r);
+      thread.setDaemon(true);
+      return thread;
+    }
+  }
+
+  private static final class TokenRefreshThread extends Thread {
+
+    TokenRefreshThread(Runnable r) {
+      super(r, "oauth2-token-refresh");
```
dimas-b, line 49: nit: `nessie-client-oauth2-token-refresh` maybe?

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -115,6 +118,16 @@ void testBackgroundRefresh() throws Exception {
...
         assertThat(client.getCurrentTokens()).isInstanceOf(RefreshTokensResponse.class);
+
+        // emulate executor running the scheduled renewal task and detecting that the client is idle
+        // after 30+ seconds of inactivity
+        i.set(i.get().plusSeconds(31));
+        currentRenewalTask.get().run();
+        assertThat(client.sleeping).isTrue();
+
+        // should exit sleeping mode on next authenticate() call
+        client.authenticate();
```
dimas-b, line 129: Do you think we could add a test case for task scheduling failures? Those may happen with user-supplied executors (e.g. short queue), I guess... WDYT?

dimas-b (commented):
LGTM overall. A few polishing comments :)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: The only way I can image where this would throw is if `executor.schedule` throws `RejectedExecutionException`. I prefer for this exception to propagate in fact, as the OAuth2 client is probably left in an usable state.

But you are remark is excellent; and I agree that we might want to revisit this and strengthen things a bit: e.g. if `executor.schedule` throws `RejectedExecutionException` inside the bakground thread, the error will be swallowed here:

https://github.com/projectnessie/nessie/blob/98448b542fd8c9f2f69a6dbeb265c3d94c6bf21c/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java#L212

That's indeed bad and the client would be left in an unrecoverable state for sure. I think we should fail the `currentTokensStage` in that case, so that the exception gets propagated to client code when the user calls `authenticate()`.

Do you mind if I do this in another PR? As it's only partially related to sleeping/waking up.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
adutra: Yes, good idea. Do you mind if I tackle the scheduling failures stuff in a follow-up PR?

dimas-b (approved):

PR #7651, opened 2023-10-20
Enhance released jars

snazy:
This change adds a couple of manifest attributes to jars, if the `release` Gradle project property is present.

The attributes are:
* `Nessie-Build-Git-Describe`: result of `git describe` - example: `nessie-0.72.2-4-g0ce4208995`
* `Nessie-Build-Git-Head`: Git HEAD commit ID
* `Nessie-Build-Java-Version`: Java version used for the build - example: `21.0.1`
* `Nessie-Build-System`: result of `uname -a`
* `Nessie-Build-Timestamp`: result of `date +%Y-%m-%d-%H:%M:%S%:z` - example: `2023-10-20-12:01:33+02:00`
* `Nessie-Version`: version as in `version.txt`, example: `0.72.3-SNAPSHOT`

Those attributes are _not_ added for non-release builds, because manifest attribtues count towards Gradle inputs, potentially rendering cached Gradle outputs unusable - i.e. every change could cause a full rebuild.


build-logic/src/main/kotlin/nessie-common-src.gradle.kts
```diff
@@ -141,3 +143,58 @@ plugins.withType<JavaPlugin>().configureEach {
...
+    }
+
+    fun gitInfo(rootProject: Project): Map<String, String> {
+      if (!rootProject.hasProperty("release")) {
+        return emptyMap()
+      }
+
+      return if (rootProject.extra.has("gitReleaseInfo")) {
+        @Suppress("UNCHECKED_CAST")
+        rootProject.extra["gitReleaseInfo"] as Map<String, String>
```
adutra, line 180: Where is this property being set?

adutra (approved):

build-logic/src/main/kotlin/nessie-common-src.gradle.kts
snazy: oops

adutra (approved):

PR #7659, opened 2023-10-23
Docker: re-enable JAVA_OPTS_APPEND env var

adutra:
This commit fixes an issue where many supported environment variables, among which, `JAVA_OPTS_APPEND`, weren't being honored.

This happens because the `JAVA_OPTS` environment variable was being overwritten in Dockerfile-jvm, while it shouldn't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ as necessary. Empty sections will not end in the release notes.
 
 ### Fixes
 
+- Docker images again honor environment variables such as `JAVA_OPTS_APPEND` that are used to pass
+  additional JVM options to the Nessie server. See the 
+  [ubi8/openjdk-17](https://catalog.redhat.com/software/containers/ubi8/openjdk-17/618bdbf34ae3739687568813)
+  base image documentation for the list of all supported environment variables.
+
 ### Commits
 
 ## [0.72.2] Release (2023-10-19)


adutra:
Here is a comparison of before/after this change:

Before (note how env vars are ignored):

```
docker run -e JAVA_MAX_MEM_RATIO=99.0 -e JAVA_OPTS_APPEND="-Dfoo=bar" ghcr.io/projectnessie/nessie  
Starting the Java application using /opt/jboss/container/java/run/run-java.sh ...
INFO exec -a "java" java -Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -cp "." -jar /deployments/quarkus-run.jar 
```

After (note how both vars were honored, and many more were automatically added):

```
docker run -e JAVA_MAX_MEM_RATIO=99.0 -e JAVA_OPTS_APPEND="-Dfoo=bar" library/nessie-unstable:0.72.3
Starting the Java application using /opt/jboss/container/java/run/run-java.sh ...
INFO exec -a "java" java -XX:MaxRAMPercentage=99.0 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:+ExitOnOutOfMemoryError -Dfoo=bar -cp "." -jar /deployments/quarkus-run.jar
```

tools/dockerbuild/docker/Dockerfile-jvm
```diff
@@ -34,5 +34,4 @@ COPY --chown=185 build/quarkus-app/quarkus/ /deployments/quarkus/
 EXPOSE 19120
 USER 185
 ENV AB_JOLOKIA_OFF=""
-ENV JAVA_OPTS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
```
dimas-b, line 37: Do we not need the `-Djava.util.logging.manager` option or is it set somewhere else?

adutra: I don't think so? I don't see any difference in logging with or without it.

adutra: Apparently setting this property is only required for tests:

https://quarkus.io/guides/logging#configure-logging-for-quarkustest

dimas-b (approved):
Nice catch :+1:

dimas-b (approved):

PR #7660, opened 2023-10-23
Relax CacheSizing validation rules.

dimas-b:
Accept wider range of property values.

* Any non-negative value for `fixedSizeInMB` This may be useful in test / POC environments.

* `fixedSizeInMB` and `fractionOfMaxHeapSize` may be specified together (`fixedSizeInMB` wins). Allowing both values to be set can be useful in helm-based deployments that have custom Quarkus `application.properties`.

* No change to actual cache size computation logic.

### DIFF
--- a/versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheSizing.java
+++ b/versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheSizing.java
@@ -68,7 +68,7 @@ public class TestCacheSizing {
 
   @Test
   void tinyHeapNoCache() {
-    // Assuming a 256MB max heap, requesting 70% (358MB), calc yields fractionMinSizeMb, i.e. zero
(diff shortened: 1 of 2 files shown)


versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheSizing.java
```diff
@@ -52,6 +66,18 @@ void tinyHeap() {
         .isEqualTo(64);
   }
 
+  @Test
+  void tinyHeapNoCache() {
+    // Assuming a 256MB max heap, requesting 70% (358MB), calc yields fractionMinSizeMb, i.e. zero
```
adutra, line 71:
```suggestion
    // Assuming a 256MB max heap, requesting 70% (179MB), calc yields fractionMinSizeMb, i.e. zero
```

adutra (approved):
LGTM, one comment was slightly off

PR #7663, opened 2023-10-24
OAuth2Client: forbid recursive calls to renewTokens()

adutra:
This commit is a purely technical refactoring.

It fixes a possible, although very unlikely, situation where the thread executing renewTokens could potentially re-enter that method, because the next refresh delay is too short. That would be incorrect.

This situation is only theoretical at the moment, given the current minimum delays, but making it completely impossible by design will prevent future changes from accidentally introducing a regression.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
@@ -108,8 +108,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
     Instant now = clock.get();
     lastAccess = now;
     if (sleeping.compareAndSet(true, false)) {
-      LOGGER.debug("Waking up...");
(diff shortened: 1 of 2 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -171,34 +170,32 @@ public void close() {
     }
   }
 
+  private void wakeUp(Instant now) {
+    LOGGER.debug("Waking up...");
+    Duration delay = nextTokenRefresh(getCurrentTokens(), now, Duration.ZERO);
+    if (delay.compareTo(MIN_REFRESH_DELAY) < 0) {
+      LOGGER.debug("Refreshing tokens immediately");
+      renewTokens();
```
adutra, line 178: The goal is to isolate this (synchronous) call to `renewTokens` in a method that is only executed in a wake-up situation.

dimas-b (approved):

PR #7667, opened 2023-10-25
Make Conflict.conflictType non nullable

adutra:
`Conflict.conflictType` is marked nullable, but in practice it isn't.

The reason why it was marked nullable seems related to JSON serialization. But after checking the history of the `Conclict` type, it appears that the field `conflictType()` was present since the beginning, and always serialized. So there cannot be an older Nessie version that would serialize this type without that field. (And besides, this type is only serialized in server-to-client direction.)

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Conflict.java
+++ b/api/model/src/main/java/org/projectnessie/model/Conflict.java
@@ -34,8 +34,6 @@ import org.immutables.value.Value;
     ignoreUnknown = true)
 public interface Conflict {
   @Value.Parameter(order = 1)
-  @Nullable
-  @jakarta.annotation.Nullable
   @JsonDeserialize(using = ConflictType.Deserializer.class)
   ConflictType conflictType();
 
@@ -48,7 +46,7 @@ public interface Conflict {
   String message();
 
   static Conflict conflict(
-      @Nullable @jakarta.annotation.Nullable ConflictType conflictType,
+      ConflictType conflictType,
       @Nullable @jakarta.annotation.Nullable ContentKey key,
       String message) {
     return ImmutableConflict.of(conflictType, key, message);
@@ -110,7 +108,7 @@ public interface Conflict {
       public ConflictType deserialize(JsonParser p, DeserializationContext ctxt)
           throws IOException {
         String name = p.readValueAs(String.class);
-        return name != null ? ConflictType.parse(name) : null;
+        return name != null ? ConflictType.parse(name) : UNKNOWN;
       }
     }
   }
--- a/api/model/src/main/java/org/projectnessie/model/Conflict.java
+++ b/api/model/src/main/java/org/projectnessie/model/Conflict.java
@@ -35,7 +35,10 @@ import org.immutables.value.Value;
 public interface Conflict {
   @Value.Parameter(order = 1)
   @JsonDeserialize(using = ConflictType.Deserializer.class)
-  ConflictType conflictType();
+  @Value.Default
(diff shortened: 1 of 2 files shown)


api/model/src/test/java/org/projectnessie/error/TestNessieError.java
```diff
@@ -67,8 +67,7 @@ static Stream<Arguments> conflictDeserialization() {
                 .put("message", "msg")
                 .put("someNewField", "blah")
                 .set("key", ckJson),
-            conflict(UNKNOWN, ContentKey.of("foo"), "msg")),
-        arguments(objectNode().put("message", "msg"), conflict(null, null, "msg")));
```
dimas-b, line 71: Would it be worth keeping this test case to validate the logic on line 111 in `[Conflict.java](https://github.com/projectnessie/nessie/pull/7667/files#diff-ef0143fc6d8dab3c14a453ada3e4ae98263f7211578ec223921369753da19260)`? Perhaps we could change the expected value to use `UNKNOWN`? WDYT?

adutra: I tried, but it doesn't work:

```
com.fasterxml.jackson.databind.exc.ValueInstantiationException: Cannot construct instance of `org.projectnessie.model.ImmutableConflict`, problem: Cannot build Conflict, some of required attributes are not set [conflictType]
 at [Source: UNKNOWN; byte offset: #UNKNOWN]
```

adutra: Hmm wait. This is suspicious. It should work. I'll investigate.

api/model/src/main/java/org/projectnessie/model/Conflict.java
```diff
@@ -106,11 +107,17 @@ public static ConflictType parse(String conflictType) {
...
+      public ConflictType getNullValue(DeserializationContext ctxt) {
+        return UNKNOWN;
+      }
+
       @Override
       public ConflictType deserialize(JsonParser p, DeserializationContext ctxt)
           throws IOException {
         String name = p.readValueAs(String.class);
-        return name != null ? ConflictType.parse(name) : null;
+        return ConflictType.parse(name);
```
adutra, line 120: The javadocs make things clear about null literals:

> Note that this method is never called for JSON null literal, and thus deserializers need (and should) not check for it.

api/model/src/main/java/org/projectnessie/model/Conflict.java
```diff
@@ -34,10 +34,11 @@
     ignoreUnknown = true)
 public interface Conflict {
   @Value.Parameter(order = 1)
-  @Nullable
-  @jakarta.annotation.Nullable
   @JsonDeserialize(using = ConflictType.Deserializer.class)
-  ConflictType conflictType();
+  @Value.Default
+  default ConflictType conflictType() {
+    return ConflictType.UNKNOWN;
```
adutra, line 40: This helps with the case where the field is not even present in the JSON payload. Not sure how realistic that is though.

dimas-b: Since we're removing `null` handling from `deserialize()`, I guess we do not have to provide a default here. All server-generated values will have a type, as you noted previously.

adutra: I was about to suggest the same thing :-)

dimas-b (approved):

PR #7669, opened 2023-10-26
OAuth2Client: implement fault tolerance

adutra:
See [this Zulip topic](https://project-nessie.zulipchat.com/#narrow/stream/371193-dev/topic/OAuth2.20client.20fault.20tolerance/near/398272049) for context.

Here is what I'm proposing:

1. Failures encountered when fetching tokens do not stop anymore the background refresh cycle; this allows the tokens to be eventually rotated, if the error was transient. Calls to `authenticate()` will eventually recover and succeed. If the error is permanent though (bad credentials for example), then the behavior is the same: all calls to `authenticate()` will fail.
2. Failures encountered when scheduling tasks are handled in 2 different ways:
  a. If the rejected task is the very first token fetch, then `authenticate()` will throw `RejectedExecutionException` as before;
  b. If, however, the rejected task is a background refresh task, then instead of failing, we keep the current tokens and enter sleep mode. In sleep mode, each call to `authenticate()` will, if necessary, rotate the tokens synchronously and attempt to schedule a refresh. If that schedule is rejected again and again, we stay in sleep mode forever, effectively making `authenticate()` fetch new tokens synchronously whenever they expire.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
@@ -28,9 +28,12 @@ import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Arrays;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -172,42 +182,50 @@ public void close() {
...
+    }
   }
 
   private void renewTokens() {
     CompletionStage<Tokens> oldTokensStage = currentTokensStage;
     currentTokensStage =
         oldTokensStage
-            // try refreshing the current tokens
+            // try refreshing the current tokens (if they exist)
             .thenApply(this::refreshTokens)
```
dimas-b, line 223: Not really related to this change, but I did not notice this before... Will this create a chain of stages that grows indefinitely with every `renewTokens` call?

dimas-b (approved):
Specific changes LGTM. It's a nice improvement.

WDYT about my comment regarding the chain of stages? Is that a concern?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
dimas-b: I guess it's fine, because `renewTokens()` will be called only after `currentTokensStage` completes.

adutra: No: each (incomplete) future holds references to its direct children, but not to its parent. Furthermore, when a future completes, the completion signal is passed downstream to all its children, then all child references are cleared.

PR #7671, opened 2023-10-26
Report all missing namespaces at once

adutra:
Opening this to trigger a discussion around this idea. We would need to increment Nessie's spec to 2.1.3 I guess.

### DIFF
--- a/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
+++ b/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
@@ -21,6 +21,7 @@ import static java.util.Collections.singletonList;
 import static org.assertj.core.api.Assertions.tuple;
 import static org.assertj.core.api.InstanceOfAssertFactories.list;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.projectnessie.model.CommitMeta.fromMessage;
 import static org.projectnessie.model.Conflict.ConflictType.NAMESPACE_ABSENT;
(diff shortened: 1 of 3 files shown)


versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
```diff
@@ -62,16 +63,17 @@ protected AbstractNamespaceValidation(VersionStore store) {
...
-        ContentKey.of("ns2", "ns", "table"));
+    return Stream.of(ContentKey.of("ns", "table"), ContentKey.of("ns2", "ns", "table"));
   }
 
   @ParameterizedTest
   @MethodSource("contentKeys")
   void commitWithNonExistingNamespace(ContentKey key) throws Exception {
+
+    assumeTrue(
+        key.getElementCount() == 2 || !isNewStorageModel(),
```
dimas-b, line 74: this leaves only one parameter for this method, why bother with `@ParameterizedTest`?

dimas-b (commented):
LGTM overall :)

dimas-b:
+1 to bumping the spec version

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
adutra: We are still testing the multiple missing namespaces case for the old model here.

dimas-b: `!isNewStorageModel()` makes sense, but do we have more one parameter?

dimas-b: sorry, misinterpreted :facepalm:

dimas-b (approved):

PR #7678, opened 2023-10-27
OAuth2Client: improve stack traces and reduce log frequency

adutra:
One more PR inspired by my musings with `OAuthClient` in a Spark session:

1. When `authenticate()` throws, the thrown error now has a more meaningful stack trace.
2. When attempting to recover from a failure, failed attempts would all be logged, one per second (the minimum
refresh interval), until success or sleep mode is reached. From now on, only one failed attempt is logged every 10 seconds, and at WARN level instead of ERROR.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
@@ -61,6 +61,7 @@ import org.slf4j.LoggerFactory;
 class OAuth2Client implements OAuth2Authenticator, Closeable {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(OAuth2Client.class);
+  private static final Duration MIN_WARN_INTERVAL = Duration.ofSeconds(10);
 
   private final String grantType;
   private final String username;
@@ -83,6 +84,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
   private volatile CompletionStage<Tokens> currentTokensStage;
   private volatile ScheduledFuture<?> tokenRefreshFuture;
   private volatile Instant lastAccess;
+  private volatile Instant lastWarn;
 
   OAuth2Client(OAuth2ClientParams params) {
     grantType = params.getGrantType();
@@ -126,11 +128,12 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
     } catch (ExecutionException e) {
       Throwable cause = e.getCause();
       if (cause instanceof RuntimeException) {
+        cause.fillInStackTrace();
         throw (RuntimeException) cause;
       } else if (cause instanceof Error) {
         throw (Error) cause;
       } else {
-        throw new RuntimeException(cause);
+        throw new RuntimeException(cause.getMessage());
       }
     }
   }
@@ -222,7 +225,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
         // We raced with close(), ignore
         return;
       }
-      LOGGER.warn("Failed to schedule next token renewal, forcibly sleeping", e);
+      maybeWarn("Failed to schedule next token renewal, forcibly sleeping");
       sleeping.set(true);
     }
   }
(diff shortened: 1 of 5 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -126,11 +128,12 @@ Tokens getCurrentTokens() {
     } catch (ExecutionException e) {
       Throwable cause = e.getCause();
       if (cause instanceof RuntimeException) {
+        cause.fillInStackTrace();
```
dimas-b, line 131: How about throwing a new `RuntimeException` but adding `cause` as a "suppressed" exception? This way we preserve both stack traces for analysis.

adutra: The original stack trace is really useless. It can only originate from `checkErrorResponse` below.

dimas-b: Perhaps... but the call to re-fill the cause's stack trace unnerves me :) What do we gain by reusing the exception with a different stack trace vs. deriving a fresh exception from it?

adutra: Just less clutter in the console. I'm fine removing this change.

dimas-b: Thanks! I happen to appreciate the finer stack details when I have to dig into errors :)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -125,12 +127,10 @@ Tokens getCurrentTokens() {
       throw new RuntimeException(e);
     } catch (ExecutionException e) {
       Throwable cause = e.getCause();
-      if (cause instanceof RuntimeException) {
-        throw (RuntimeException) cause;
-      } else if (cause instanceof Error) {
+      if (cause instanceof Error) {
         throw (Error) cause;
```
dimas-b, line 131: We could do the same for `RuntimeException`. That should hopefully simplify (undo) recent changes in test code, WDYT?

adutra: But I would like the stack trace to be populated here, so to inform the user where the error was thrown. If we do `throw (RuntimeException) e`, the stack trace will be that of the thread that instantiated the exception – which is pretty much meaningless for the user since they don't called that code directly.

Here is the result of throwing a new exception; note how the first printed stack clearly shows which API V2 method was being called:

```
java.lang.RuntimeException: Failed to acquire a valid OAuth2 access token
	at org.projectnessie.client.auth.oauth2.OAuth2Client.getCurrentTokens(OAuth2Client.java:133)
	at org.projectnessie.client.auth.oauth2.OAuth2Client.authenticate(OAuth2Client.java:118)
	at org.projectnessie.client.auth.oauth2.OAuth2AuthenticationProvider$OAuth2Authentication.addAuthHeader(OAuth2AuthenticationProvider.java:64)
	at org.projectnessie.client.http.impl.BaseHttpRequest.lambda$prepareRequest$0(BaseHttpRequest.java:68)
	...
	at org.projectnessie.client.http.HttpRequest.get(HttpRequest.java:80)
	...
	at org.projectnessie.client.http.impl.BaseHttpRequest.prepareRequest(BaseHttpRequest.java:68)
	at org.projectnessie.client.http.impl.jdk11.JavaRequest.executeRequest(JavaRequest.java:95)
	at org.projectnessie.client.http.HttpRequest.get(HttpRequest.java:80)
	at org.projectnessie.client.rest.v2.HttpGetAllReferences.get(HttpGetAllReferences.java:51)
  ...

Caused by: org.projectnessie.client.auth.oauth2.OAuth2Exception: OAuth2 server replied with HTTP status code 401 and error code "invalid_credentials": Try Again
	at org.projectnessie.client.auth.oauth2.OAuth2Client.checkErrorResponse(OAuth2Client.java:384)
	at org.projectnessie.client.http.impl.jdk11.JavaRequest.lambda$executeRequest$1(JavaRequest.java:140)
	at java.base/java.util.Collections$SingletonList.forEach(Collections.java:4856)
	at org.projectnessie.client.http.impl.jdk11.JavaRequest.executeRequest(JavaRequest.java:140)
	at org.projectnessie.client.http.HttpRequest.post(HttpRequest.java:90)
	at org.projectnessie.client.http.HttpRequest.postForm(HttpRequest.java:105)
	at org.projectnessie.client.auth.oauth2.OAuth2Client.fetchNewTokens(OAuth2Client.java:272)
  ...
	at org.projectnessie.client.auth.oauth2.OAuth2TokenRefreshExecutor$TokenRefreshThread.run(OAuth2TokenRefreshExecutor.java:56)

```

If we rethrow, we would get only the second stack above.

dimas-b: I got it now, sorry about the confusion. Current PR LGTM :+1:

dimas-b (approved):

PR #7704, opened 2023-11-09
Report a single conflict when deleting a namespace with many children

adutra:
Previously, in this situation, the namespace validation logic would emit as many conflicts as there are children, for example if there are 2 children:

    There are multiple conflicts that prevent committing the
    provided operations: The namespace 'ns' would be deleted,
    but cannot, because it has children., The namespace 'ns'
    would be deleted, but cannot, because it has children.."

Also note that capitalization and punctuation were wrong.

### DIFF
--- a/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
+++ b/versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
@@ -24,6 +24,7 @@ import static org.assertj.core.api.InstanceOfAssertFactories.type;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import static org.projectnessie.model.CommitMeta.fromMessage;
 import static org.projectnessie.model.Conflict.ConflictType.NAMESPACE_ABSENT;
+import static org.projectnessie.model.Conflict.ConflictType.NAMESPACE_NOT_EMPTY;
 import static org.projectnessie.model.Conflict.ConflictType.NOT_A_NAMESPACE;
 import static org.projectnessie.versioned.testworker.OnRefOnly.newOnRef;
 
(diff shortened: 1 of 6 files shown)


versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
```diff
@@ -303,7 +313,16 @@ void preventNamespaceDeletionWithChildren(boolean childNamespace) throws Excepti
                     Optional.empty(),
                     fromMessage("try delete ns"),
                     singletonList(Delete.of(ns.toContentKey()))))
-        .isInstanceOf(ReferenceConflictException.class);
+        .hasMessage("The namespace 'ns' would be deleted, but cannot, because it has children.")
```
adutra, line 316: I wonder if we couldn't simplify the message to just `namespace 'ns' is not empty`?

dimas-b: I guess the deletion / check is recursive?

dimas-b (approved):

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
```diff
@@ -303,7 +313,16 @@ void preventNamespaceDeletionWithChildren(boolean childNamespace) throws Excepti
                     Optional.empty(),
                     fromMessage("try delete ns"),
                     singletonList(Delete.of(ns.toContentKey()))))
-        .isInstanceOf(ReferenceConflictException.class);
```
snazy, line 306: mind keeping this check?

versioned/tests/src/main/java/org/projectnessie/versioned/tests/AbstractNamespaceValidation.java
snazy: > I wonder if we couldn't simplify the message to just `namespace 'ns' is not empty`?

wfm

snazy (approved):

PR #7709, opened 2023-11-13
Nessie-GC: Support Google Cloud Storage (GCS) (experimental)

snazy:
Adding support for Google Cloud Storage as experimental. Unfortunately there's no way to test the integration in CI, because there's no emulator available.

PR #7715, opened 2023-11-14
Nessie-GC: Support Azure Blob Storage (experimental)

snazy:
Adding support for Azure Blob Storage as experimental. Unfortunately there's no way to test the integration in CI, although an emulator (Azurite) is available, because Iceberg does not allow using plain username/password for that purpose.


snazy:
This one is similar to #7709, not testable yet :(

snazy:
Opened this [iceberg PR](https://github.com/apache/iceberg/pull/9068).

dimas-b (dismissed):

PR #7727, opened 2023-11-16
Reduce usage of conditional writes in Nessie

adutra:
Following an internal discussion on the usage of conditional writes in Nessie new model, I am proposing this change that significantly reduces conditional writes, effectively limiting those to updates to the `refs` table only. For comparison, today almost _every_ write performed by Nessie is a conditional write.

The PR should be reviewable as a whole, but to facilitate the review for those who prefer smaller changes, the changes are split in different commits:

* 4914b446b0fafafe3dae541740a37a85caab45ed : Replaces conditional writes with regular ones in places where the condition wasn't being checked. This is a straightforward change.
* 648642e60dd4fd0926eb093d32f12afdf1413a66: Replaces conditional writes with regular ones for commits, adapts `CommitLogic` accordingly, adapts affected classes and tests. This commit does not modify `Persist`, it only switches the entire codebase to use `upsertObj*` exclusively.
* e6aece9f2964d7ae05034f7e30257d79c85cebce: Removes `Persist.storeObj*` completely, and fixes a few faulty `upsertObj*` implementations (details will be provided in comments).
* 44faa572a1325424f2c6b308069c923b5150ecd2: Simply renames `upsertObj*` back to `storeObj*`, since that name is more natural, and was more widely used before. This commit is not strictly necessary, it can be removed if we prefer to use the name `upsertObj*` instead.
* a0db44f7e8a95dc81fe7ab23c2594469613c7026: adds a changelog entry.

### DIFF
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
@@ -407,7 +407,7 @@ public class BigTablePersist implements Persist {
   }
 
   @Override
-  public void upsertObj(
+  public void storeObj(
       @Nonnull @jakarta.annotation.Nonnull Obj obj, boolean ignoreSoftSizeRestrictions)
       throws ObjTooLargeException {
     ObjId id = obj.id();
@@ -440,7 +440,7 @@ public class BigTablePersist implements Persist {
   }
 
   @Override
-  public void upsertObjs(@Nonnull @jakarta.annotation.Nonnull Obj[] objs)
+  public void storeObjs(@Nonnull @jakarta.annotation.Nonnull Obj[] objs)
       throws ObjTooLargeException {
     if (objs.length == 0) {
       return;
@@ -448,13 +448,17 @@ public class BigTablePersist implements Persist {
 
     if (objs.length == 1) {
       if (objs[0] != null) {
-        upsertObj(objs[0]);
+        storeObj(objs[0]);
       }
       return;
     }
 
     try (Batcher<RowMutationEntry, Void> batcher =
         backend.client().newBulkMutationBatcher(backend.tableObjs)) {
+
+      int incrementalIndexSizeLimit = effectiveIncrementalIndexSizeLimit();
+      int indexSizeLimit = effectiveIndexSegmentSizeLimit();
+
       for (Obj obj : objs) {
         if (obj == null) {
           continue;
@@ -465,9 +469,7 @@ public class BigTablePersist implements Persist {
         ByteString key = dbKey(id);
 
         ByteString serialized =
-            unsafeWrap(
-                serializeObj(
-                    obj, effectiveIncrementalIndexSizeLimit(), effectiveIndexSegmentSizeLimit()));
+            unsafeWrap(serializeObj(obj, incrementalIndexSizeLimit, indexSizeLimit));
 
         batcher.add(
             RowMutationEntry.create(key)
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
@@ -369,101 +369,11 @@ public class BigTablePersist implements Persist {
     }
   }
 
-  @Override
-  public boolean storeObj(
-      @Nonnull @jakarta.annotation.Nonnull Obj obj, boolean ignoreSoftSizeRestrictions)
-      throws ObjTooLargeException {
-
-    try {
-      ConditionalRowMutation conditionalRowMutation =
-          mutationForStoreObj(obj, ignoreSoftSizeRestrictions);
-
-      boolean success = backend.client().checkAndMutateRow(conditionalRowMutation);
-      return !success;
-    } catch (ApiException e) {
-      throw apiException(e);
-    }
-  }
-
   private static final ByteString[] OBJ_TYPE_VALUES =
       Arrays.stream(ObjType.values())
           .map(Enum::name)
           .map(ByteString::copyFromUtf8)
           .toArray(ByteString[]::new);
-
-  @NotNull
-  private ConditionalRowMutation mutationForStoreObj(
-      @NotNull Obj obj, boolean ignoreSoftSizeRestrictions) throws ObjTooLargeException {
-    checkArgument(obj.id() != null, "Obj to store must have a non-null ID");
-    ByteString key = dbKey(obj.id());
-
-    int incrementalIndexSizeLimit =
-        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIncrementalIndexSizeLimit();
-    int indexSizeLimit =
-        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIndexSegmentSizeLimit();
-    byte[] serialized = serializeObj(obj, incrementalIndexSizeLimit, indexSizeLimit);
-
-    ByteString ref = unsafeWrap(serialized);
-
-    Mutation mutation =
-        Mutation.create()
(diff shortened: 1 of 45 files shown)


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -497,51 +407,78 @@ public void deleteObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids) {
...
-                  obj, effectiveIncrementalIndexSizeLimit(), effectiveIndexSegmentSizeLimit()));
+          unsafeWrap(serializeObj(obj, incrementalIndexSizeLimit, indexSizeLimit));
 
       backend
           .client()
           .mutateRow(
               RowMutation.create(backend.tableObjs, key)
-                  .setCell(FAMILY_OBJS, QUALIFIER_OBJS, CELL_TIMESTAMP, serialized));
+                  .setCell(FAMILY_OBJS, QUALIFIER_OBJS, CELL_TIMESTAMP, serialized)
+                  .setCell(
```
adutra, line 432: The upsert methods weren't setting the `QUALIFIER_OBJ_TYPE` cell, which is not correct if the row does not exist yet.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
```diff
@@ -132,49 +132,20 @@ public Obj[] fetchObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids)
...
+    persist.storeObj(obj, ignoreSoftSizeRestrictions);
+    cache.put(obj);
   }
 
   @Override
-  public void upsertObj(@jakarta.annotation.Nonnull @Nonnull Obj obj) throws ObjTooLargeException {
-    try {
-      persist.upsertObj(obj);
-    } finally {
-      cache.remove(obj.id());
```
adutra, line 164: I don't understand why the upsert methods were removing objects from the cache. I think it's an oversight.

versioned/storage/dynamodb/src/main/java/org/projectnessie/versioned/storage/dynamodb/DynamoDBPersist.java
```diff
@@ -547,38 +499,47 @@ public void deleteObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids) {
...
+      checkAndRethrow(e);
     }
   }
 
   @Override
-  public void upsertObjs(@Nonnull @jakarta.annotation.Nonnull Obj[] objs)
+  public void storeObjs(@Nonnull @jakarta.annotation.Nonnull Obj[] objs)
       throws ObjTooLargeException {
-    // DynamoDB does not support "PUT IF NOT EXISTS" in a BatchWriteItemRequest/PutItem
     try (BatchWrite batchWrite = new BatchWrite(backend, backend.tableObjs)) {
```
adutra, line 520: We should observe some improvement here since we can now use batches writes.

versioned/storage/rocksdb/src/main/java/org/projectnessie/versioned/storage/rocksdb/RocksDBPersist.java
```diff
@@ -429,20 +380,24 @@ public void deleteObjs(@Nonnull @jakarta.annotation.Nonnull ObjId[] ids) {
...
-      byte[] key = dbKey(id);
+    byte[] key = dbKey(id);
 
-      byte[] serialized =
-          serializeObj(obj, effectiveIncrementalIndexSizeLimit(), effectiveIndexSegmentSizeLimit());
+    int incrementalIndexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIncrementalIndexSizeLimit();
+    int indexSizeLimit =
+        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIndexSegmentSizeLimit();
+    byte[] serialized = serializeObj(obj, incrementalIndexSizeLimit, indexSizeLimit);
```
adutra, line 395: Moved this call outside the locked section to reduce lock time.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseMergeTransplantSquash.java
```diff
@@ -119,10 +119,8 @@ MergeResult<Commit> squash(
     } else {
       CommitLogic commitLogic = commitLogic(persist);
       newHead = mergeCommit.id();
-      boolean committed = commitLogic.storeCommit(mergeCommit, objsToStore);
-      if (committed) {
```
adutra, line 123: Always true, because merges now always create a squash commit.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/TransplantIndividualImpl.java
```diff
@@ -114,8 +114,9 @@ public MergeResult<Commit> transplant(Optional<?> retryState, TransplantOp trans
       empty = false;
       if (!transplantOp.dryRun()) {
         newHead = newCommit.id();
-        boolean committed = commitLogic.storeCommit(newCommit, objsToStore);
-        if (committed) {
+        boolean fastForward = sourceCommit.id().equals(newHead);
+        if (!fastForward) {
```
adutra, line 118: In case of fast forward, the new commit is identical to the source commit and doesn't need to be persisted again. 
Unfortunately we cannot call `CommitObj.equals` because the two commits actually differ by one property: `CommitObj.created()`.
This begs the question: should `CommitObj.created()` be annotated with `@Value.Auxiliary`?

adutra:
Note: if the idea is deemed worth adopting, I would like to run benchmarks at least with BigTable, DynamoDB and Cassandra, since these backends are the ones where I would intuitively expect some performance improvements.

snazy (changes requested):
This change completely removes the conditional writes, which ensure that there are no duplicate keys. This can, although unlikely, lead to silent data corruption. I'm not a fan of this change TBH.

adutra:
> This change completely removes the conditional writes, which ensure that there are no duplicate keys. This can, although unlikely, lead to silent data corruption. 

I hear your concerns, and that's why my initial proposal was to use random UUIDs in lieu of commit hashes.

That idea, as you recall, was discarded because random commit IDs make it impossible to establish the proof of correctness of the commit chain.

I think we can still solve this problem and get the best of both worlds, _by introducing more randomness in the commit hashes themselves_. 

A rather non-intrusive way of achieving this would be to enforce a new commit meta property containing a random number or UUID.

This can be easily done in `DefaultMetadataRewriter.buildCommitMeta()`:

```java
return metaBuilder
    .commitTime(now)
    .putProperties("COMMIT_UUID", UUID.randomUUID().toString()) // could also be time-based
    .build();
```

The new commit meta property would automatically participate in the commit hash computation, thus enhancing the entropy to a level where collisions cannot happen anymore (note that commit time is already part of the hash computation and is already introducing _some_ randomness in the process – only not enough). OTOH, since the new property is persisted as part of the commit object, establishing the proof of correctness of the commit chain is still feasible.

Also, this solution does not require any updates to the database schema nor to the REST API. We _could_ instead introduce a new property in `CommitMeta` itself, e.g. `public UUID getCommitTimeUuid()` – but that would have consequences for older clients I believe.

This solution would have one drawback: colliding writes could not happen anymore, but garbage writes (orphaned objects) would happen more often, especially in contented servers. We would need to perform some benchmarks to evaluate how much more garbage it generates compared to the amount of garbage being currently generated.

Should I go ahead with this idea?

snazy:
I don't see a need to make commits having a random ID. Commit-ID (aka commit _hashes_) are meant to be deterministic. The source for the hash is already quite diverse - but reproducible - and that's the guarantee here.

Let's just make sure that the result of `storeObj(s)` is evaluated for all objects - and fail if the existing object is different from the one to be persisted.

adutra:
Fair enough. Closing.

PR #7729, opened 2023-11-17
Adding metadataLocation in ViewVersionMetadata

leewjae:
In OSS Iceberg's view, we have toJson method and fromJson method that takes metadataLocation in place, and that can be used in various places (The code snippet is from OSS Iceberg). Nessie can take advantage using this new code.


snazy:
@leewjae the "Iceberg Views" code in Nessie is based on a very very early draft of the Iceberg View Spec, which is largely different from the released Apache Iceberg View Spec. This code here has always been considered as "experimental" and the whole `nessie-iceberg-views` module will eventually be removed, in favor of the code that is now present in Apache Iceberg.

snazy:
Closed as per the discussion above and off-line.

PR #7742, opened 2023-11-21
Model: Make sqlText optional for IcebergView

ajantha-bhat:
Iceberg client creates a new view metadata file for every commit. Hence, Just the view metadata location is enough to be tracked from Nessie side. No need to track other fields (or make them mandatory). 

Reference: https://github.com/apache/iceberg/pull/8909#discussion_r1397190437

This PR makes `sqlText` as optional for builder.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/IcebergView.java
+++ b/api/model/src/main/java/org/projectnessie/model/IcebergView.java
@@ -49,11 +49,10 @@ public abstract class IcebergView extends IcebergContent {
 
   public abstract int getSchemaId();
 
-  @NotBlank
-  @jakarta.validation.constraints.NotBlank
-  @NotNull
-  @jakarta.validation.constraints.NotNull
-  public abstract String getSqlText();
+  @Value.Default
+  public String getSqlText() {
+    return "default sql text";
+  }
 
   @Nullable
   @jakarta.annotation.Nullable // TODO this is currently undefined in Iceberg
--- a/api/model/src/main/java/org/projectnessie/model/IcebergView.java
+++ b/api/model/src/main/java/org/projectnessie/model/IcebergView.java
@@ -49,6 +49,8 @@ public abstract class IcebergView extends IcebergContent {
 
   public abstract int getSchemaId();
 
+  @NotBlank
+  @jakarta.validation.constraints.NotBlank
   @Value.Default
   public String getSqlText() {
     return "default sql text";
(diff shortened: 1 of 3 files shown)


snazy:
Strong -1 on this one - it _will_ break compatibility with older clients.

snazy (changes requested):
This breaks compatibility.

ajantha-bhat:
>  This breaks compatibility.

I see. If the new client writes the view model without the sql text and the old client tries to read it, It will fail as it expects non-null?

Shall I fill the dummy sql text from the new client instead of making it Nullable then?

Or do you have any other suggestions?

ajantha-bhat:
> Or do you have any other suggestions?

Maybe I can add a default value

api/model/src/main/java/org/projectnessie/model/IcebergView.java
```diff
@@ -49,11 +49,10 @@ public abstract class IcebergView extends IcebergContent {
...
   public abstract int getSchemaId();
 
-  @NotBlank
-  @jakarta.validation.constraints.NotBlank
-  @NotNull
-  @jakarta.validation.constraints.NotNull
-  public abstract String getSqlText();
+  @Value.Default
+  String getSqlText() {
+    return "dummy sql text";
```
dimas-b, line 54: `dummy sql text` does not look nice in an API :) why not `null` now that this field is nullable?

ajantha-bhat: renamed to "default sql text". Suggestions are welcome.

ajantha-bhat:
>  Old client failing hard when sqlText is null is not acceptable.

`sqlText` is not `Nullable` even with the present changes. So, how it can be null for old client?

With the present changes, I tried manually setting `null` for `sqlText` which throws an error as expected.

```
java.lang.NullPointerException: sqlText
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at org.projectnessie.model.ImmutableIcebergView$Builder.sqlText(ImmutableIcebergView.java:603)
```

ajantha-bhat:
> Old client failing hard when sqlText is null is not acceptable.

I think you meant blank instead of null. I will handle it.

snazy:
Still - nothing to change on the Nessie side, no?

ajantha-bhat:
> Still - nothing to change on the Nessie side, no?

But it still feels so weird to set the dummy SQL text from Iceberg side when we have the one or more SQL text representation.

ajantha-bhat:
I also found that dialect is optional in IcebergView but not optional in the proto serialization here
https://github.com/projectnessie/nessie/blob/0e3191a82d7f48b3cafa62f76ab8506de89ba7e4/servers/store-proto/src/main/proto/org/projectnessie/server/store/proto/table.proto#L104

PR #7745, opened 2023-11-21
CachePersistImpl: fix upsert methods, add specific tests

adutra:
This PR fixes a few methods in `CachePersistImpl`:

* `fetchObj` was unnecessarily complex;
* `upsertObj` and `upsertObjs` were not putting the upserted objects in the cache.

I added specific tests to make sure the expectations around cache behavior are all met.

### DIFF
--- a/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
+++ b/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
@@ -48,14 +48,9 @@ class CachingPersistImpl implements Persist {
     if (o != null) {
       return o;
     }
-    try {
-      o = persist.fetchObj(id);
-      cache.put(o);
-      return o;
-    } catch (ObjNotFoundException e) {
-      cache.remove(id);
-      throw e;
-    }
+    o = persist.fetchObj(id);
+    cache.put(o);
+    return o;
   }
 
   @Override
@@ -158,23 +153,17 @@ class CachingPersistImpl implements Persist {
 
   @Override
   public void upsertObj(@jakarta.annotation.Nonnull @Nonnull Obj obj) throws ObjTooLargeException {
-    try {
-      persist.upsertObj(obj);
-    } finally {
-      cache.remove(obj.id());
-    }
+    persist.upsertObj(obj);
+    cache.put(obj);
   }
 
   @Override
   public void upsertObjs(@jakarta.annotation.Nonnull @Nonnull Obj[] objs)
       throws ObjTooLargeException {
-    try {
-      persist.upsertObjs(objs);
-    } finally {
-      for (Obj obj : objs) {
-        if (obj != null) {
-          cache.remove(obj.id());
-        }
+    persist.upsertObjs(objs);
+    for (Obj obj : objs) {
+      if (obj != null) {
+        cache.put(obj);
       }
     }
   }
@@ -210,6 +199,8 @@ class CachingPersistImpl implements Persist {
     }
   }
 
+  // plain delegates...
+
   @Override
   @Nonnull
   @jakarta.annotation.Nonnull
@@ -217,8 +208,6 @@ class CachingPersistImpl implements Persist {
       @Nonnull @jakarta.annotation.Nonnull Set<ObjType> returnedObjTypes) {
     return persist.scanAllObjects(returnedObjTypes);
   }
-
-  // plain delegates...
 
   @Override
   public int hardObjectSizeLimit() {
--- a/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
+++ b/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
@@ -130,11 +130,16 @@ class CachingPersistImpl implements Persist {
   public boolean storeObj(
       @jakarta.annotation.Nonnull @Nonnull Obj obj, boolean ignoreSoftSizeRestrictions)
       throws ObjTooLargeException {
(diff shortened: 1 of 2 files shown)


versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
```diff
@@ -158,23 +153,17 @@ public boolean[] storeObjs(@jakarta.annotation.Nonnull @Nonnull Obj[] objs)
 
   @Override
   public void upsertObj(@jakarta.annotation.Nonnull @Nonnull Obj obj) throws ObjTooLargeException {
-    try {
-      persist.upsertObj(obj);
-    } finally {
-      cache.remove(obj.id());
-    }
+    persist.upsertObj(obj);
+    cache.put(obj);
```
dimas-b, line 157: What if `upsertObj` fails? Would we not want to remove the possible old entry from the cache in that case (same for `upsertObjs`)? We cannot be certain what the next `get` will return because we do not know the nature of failure.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
```diff
@@ -210,6 +199,8 @@ public void erase() {
     }
   }
 
+  // plain delegates...
```
dimas-b, line 202: nit: is this comment that much helpful?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
snazy: Similar for `upsertObjs`

adutra: My initial assumption was that If `upsertObj` fails, then we can be certain that the object has _not_ been persisted. In which case, the most correct thing to do is nothing, i.e. the cache and the underlying store both should keep the previous version of the object, if any, so that they remain in sync.

Thus, I suppose that you are instead considering that a hard failure in `upsertObj` might still have been able to persist the object, in which case, indeed we don't know for sure which version the underlying store has: the previous one (if any), or the current one.

That's a rather pessimistic view, but I am not against it as we never know how future backends could implement `upsertObj`. 

However, if we remove the cache entry on a hard failure for `upsertObj(s)`, I don't see why we wouldn't do the same for `storeObj(s)` as well. Wdyt?

snazy: `upsertObj` may or may not have done something. But since `upsertObj` is only used during repo-import, it's okay to pessimistically remove it from the cache. `upsertObj` is not meant to be used in production code, because we never update existing objects.

dimas-b: Do we benefit from the cache during import _after_ upserts are done?

snazy: My point is: `upsertObj` is irrelevant for production performance.

> How about ... remove on exception?.. however this will probably make the code more complex 🤔

This is mostly what it's been doing before?

Not sure whether it's really worth to look into a function that's rarely used, if ever used.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
```diff
@@ -48,14 +48,9 @@ public Obj fetchObj(@Nonnull @jakarta.annotation.Nonnull ObjId id) throws ObjNot
...
-    try {
-      o = persist.fetchObj(id);
-      cache.put(o);
-      return o;
-    } catch (ObjNotFoundException e) {
-      cache.remove(id);
-      throw e;
-    }
+    o = persist.fetchObj(id);
+    cache.put(o);
```
dimas-b, line 52: Thinking more about upserts and concurrent access, I'm not sure the impl. is perfectly correct. I believe with concurrent get/upsert we can end up with inconsistent data in the cache even without failures. WDYT?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
dimas-b: If upsert is only for imports, shall we refactor the `Persist` interface to make it explicit (maybe have a sub-interface)?

dimas-b: > I don't see how upsert's broken ("fix upsert")

I agree that upsert is not broken in current code, while it may be considered not efficient... nonetheless, figuring it all out requires knowledge from code areas far from the cache :) (such as import code).

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CachingPersistImpl.java
snazy: Again - upsert is _not_ meant for production use cases - only for repository import - see [javadoc](https://github.com/projectnessie/nessie/blob/f8d41bc8bcd0434de88111da1552d95966a52843/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/Persist.java#L293-L294).

snazy:
I think we should not aim for a 110% perfect API that’s only used internally.

adutra:
Alright, I was wrong all the way. Closing.

PR #7643, opened 2023-10-19
Expose object cache metrics

snazy:
Currently there's no way to inspect the runtime behavior of Nessie's object cache, nor a way to inspect its configuration.

This change exposes the runtime metrics and the cache's configuration.

Example output from Quarkus' metrics endpoint:
```
cache_gets_total{application="Nessie",cache="nessie-objects",result="hit"} 1006.0 # {span_id="237f43a65f9b9b55",trace_id="2d010ee3253ad98c2075ed4dbf72d6e5"} 1.0 1697710156.971
cache_gets_total{application="Nessie",cache="nessie-objects",result="miss"} 0.0
cache_loads_seconds_max{application="Nessie",cache="nessie-objects",result="success"} 0.0
cache_loads_seconds_max{application="Nessie",cache="nessie-objects",result="failure"} 0.0
cache_loads_seconds_count{application="Nessie",cache="nessie-objects",result="success"} 0.0
cache_loads_seconds_sum{application="Nessie",cache="nessie-objects",result="success"} 0.0
cache_loads_seconds_count{application="Nessie",cache="nessie-objects",result="failure"} 0.0
cache_loads_seconds_sum{application="Nessie",cache="nessie-objects",result="failure"} 0.0
cache_evictions_count{application="Nessie",cache="nessie-objects",cause="REPLACED"} 0.0
cache_evictions_sum{application="Nessie",cache="nessie-objects",cause="REPLACED"} 0.0
cache_evictions_count{application="Nessie",cache="nessie-objects",cause="COLLECTED"} 0.0
cache_evictions_sum{application="Nessie",cache="nessie-objects",cause="COLLECTED"} 0.0
cache_evictions_count{application="Nessie",cache="nessie-objects",cause="EXPIRED"} 0.0
cache_evictions_sum{application="Nessie",cache="nessie-objects",cause="EXPIRED"} 0.0
cache_evictions_count{application="Nessie",cache="nessie-objects",cause="SIZE"} 0.0
cache_evictions_sum{application="Nessie",cache="nessie-objects",cause="SIZE"} 0.0
cache_evictions_count{application="Nessie",cache="nessie-objects",cause="EXPLICIT"} 0.0
cache_evictions_sum{application="Nessie",cache="nessie-objects",cause="EXPLICIT"} 0.0
cache_evictions_max{application="Nessie",cache="nessie-objects",cause="REPLACED"} 0.0
cache_evictions_max{application="Nessie",cache="nessie-objects",cause="COLLECTED"} 0.0
[...]

PR #7746, opened 2023-11-21
Fix wrong cache capacity unit in `NessiePersistCache`

adutra:


dimas-b (approved):
Nice catch! Could you link the PR that changed cache size units?

adutra:
> Could you link the PR that changed cache size units?

Sure it's https://github.com/projectnessie/nessie/pull/7643. I will mention that in the commit message.

PR #7757, opened 2023-11-24
Ensure that content IDs are unique in a Nessie repository

snazy:
Nessie content IDs are random IDs, but we do not guarantee that those are actually really unique.

This change adds a new object type to ensure that a generated ID is unique by leveraging existing functionality of the `Persist` framework that already provides "`INSERT IF NOT EXIST`" guarantees.

New content IDs from this change on are now verified. This change does not include functionality to automatically add already existing content-IDs. IMHO it is probably okay for now given the practically non-existing probability of content-ID conflicts.


adutra (commented):
LGTM although there are multiple conflicts because of #7771.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/Hashes.java
```diff
@@ -122,4 +123,10 @@ static ObjId stringDataHash(
     hasher.putBytes(text.asReadOnlyByteBuffer());
     return hashAsObjId(hasher);
   }
+
+  static ObjId uniqueIdHash(String space, String value) {
```
adutra, line 127: One general remark: wouldn't it be simpler to model the value as an opaque byte array? It seems we could save some conversions to and from string in the common case where the value is an UUID.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/UniqueIdObj.java
```diff
@@ -0,0 +1,62 @@
...
+import javax.annotation.Nullable;
+import org.immutables.value.Value;
+import org.projectnessie.versioned.storage.common.logic.ReferenceLogic;
+import org.projectnessie.versioned.storage.common.persist.Obj;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.ObjType;
+import org.projectnessie.versioned.storage.common.persist.Persist;
+
+/**
+ * Describes the <em>internal</em> state of a reference when it has been created, managed by {@link
```
adutra, line 29: The javadoc seems to refer to `RefObj`.

adutra (approved):

PR #7771, opened 2023-11-28
Extensible object types

adutra:
Can be reviewed one commit at a time. Opening it now to gather some early feedback.

TODOs:

- [ ] Tests for custom object types
- [ ] Import / export changes ?


versioned/storage/rocksdb-proto/src/main/proto/org/projectnessie/versioned/storage/rocksdb/proto/rocksdb.proto
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+syntax = "proto3";
+package nessie.storage.rocksdb;
+
+option java_package = "org.projectnessie.versioned.storage.rocksdb.proto";
+option java_outer_classname = "RocksDBStorageTypes";
+option java_generate_equals_and_hash = true;
+
+message RocksDBObjProto {
```
snazy, line 23: I think we don't need a separate project for this single type.
Just rename it to sth like `GenericObjProto` or so and put it into the existing `-proto` project.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+- Nessie's persistence layer now supports extensible object types. This allows to store objects that
```
snazy, line 17: It's an internal change - nothing user facing - no need for a changelog entry IMO

snazy (commented):
Looks promising!

However I suspect that we should have a generic way to register custom types w/o the need to implement custom, database-specific serializers. I think that it's good enough to add custom serialization (for example using Jackson Smile) in this change and also the columns that are needed. Database specific serialization would then check whether the type's a "built in" one or a custom one - and serialize the custom ones always as a blob using e.g. Jackson Smile.

I don't mind to keep the new separate classes for object type (de)serialization, because those make the `*Persist` implementation classes smaller.

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -658,13 +659,25 @@ protected Obj computeNext() {
...
       return null;
     }
     return objIdFromByteBuffer(bytes.asReadOnlyByteBuffer());
   }
 
+  private static ObjType deserializeObjType(Row row) {
+    ObjType type = null;
+    List<RowCell> objectTypeCells = row.getCells(FAMILY_OBJS, QUALIFIER_OBJ_TYPE);
+    if (!objectTypeCells.isEmpty()) {
+      // Object type cell wasn't always present before Nessie 0.74.0,
```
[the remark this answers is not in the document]
adutra: Yes, it was fixed recently. See https://github.com/projectnessie/nessie/pull/7731.

gradle/libs.versions.toml
```diff
@@ -100,6 +100,7 @@ immutables-builder = { module = "org.immutables:builder", version.ref = "immutab
 immutables-value-annotations = { module = "org.immutables:value-annotations", version.ref = "immutables" }
 immutables-value-processor = { module = "org.immutables:value-processor", version.ref = "immutables" }
 jackson-bom = { module = "com.fasterxml.jackson:jackson-bom", version = "2.16.0" }
+jackson-smile = { module = "com.fasterxml.jackson.dataformat:jackson-dataformat-smile", version = "2.16.0" }
```
snazy, line 103: Why not import the bom and just mention the dependency?

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -385,11 +387,10 @@ public boolean storeObj(
     }
   }
 
-  private static final ByteString[] OBJ_TYPE_VALUES =
-      Arrays.stream(ObjType.values())
-          .map(Enum::name)
-          .map(ByteString::copyFromUtf8)
-          .toArray(ByteString[]::new);
+  private static final Map<String, ByteString> OBJ_TYPE_VALUES =
```
snazy, line 390: Can't we have `Map<ObjType, ...` here?

versioned/storage/common-serialize/src/main/java/org/projectnessie/versioned/storage/serialize/SmileSerialization.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.dataformat.smile.databind.SmileMapper;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.ByteBuffer;
+import org.projectnessie.versioned.storage.common.persist.Obj;
+
+public final class SmileSerialization {
+
+  private static final SmileMapper SMILE_MAPPER =
```
snazy, line 27: Nit: use `ObjectMapper` here (no cast needed)

gradle/libs.versions.toml
adutra: It's not part of the bom :-(

adutra: Hmm actually it is there, but when I tried it didn't work. I'll dig further.

adutra: It's working now.

snazy:
The change LGTM - just that CI issue needs to be fixed. We can nit-pick-optimize later, if needed.

snazy:
Force-pushed your branch to resolve the conflict

snazy (approved):

PR #7779, opened 2023-11-29
Increase access level for MultiEnvTestEngine helpers

colan-dremio:
Several existing helpers are currently package private and would be helpful to expose for re-use in other packages.


colan-dremio:
At present, this represents a set of classes/methods that I am explicitly planning to use. It leaves a few classes in a state that is partially public and partially package private. If there is interest, I'm happy to make more items public for consistency and reuse elsewhere too.

dimas-b (approved):

snazy (changes requested):
The changed classes are all in the **`internal`** package of the test extension, those were not intended to be used externally.

Can you elaborate which functionality is missing and maybe have a PR that add the potentially missing functionality to the JUnit extension rather than relying on internal behavior of the test extension?

colan-dremio:
The primary use case was to support more than one `MultiEnvTestExtension`. I've created #7816 as a more robust solution.

PR #7783, opened 2023-11-30
Require Java 17 for the build, prepare for Quarkus 3.7

snazy:
Quarkus 3.7 will require Java 17, so this PR prepares for this change, mostly changing existing GH workflows.


snazy:
Note: we cannot move to Java 21 yet, because Gradle still has issues w/ Java 21 in some situations :facepalm:

.github/actions/dev-tool-java/action.yml
```diff
@@ -16,7 +16,9 @@ runs:
       uses: actions/setup-java@v4
       with:
         distribution: 'temurin'
-        java-version: ${{ inputs.java-version }}
+        java-version: |
+          11
```
dimas-b, line 20: Why do we need `11` here?

snazy: Spark tests

dimas-b (dismissed):

PR #7795, opened 2023-12-04
Persist/custom objects: allow compression

snazy:
Adds compression capability to custom object types. Uses snappy for serialized data larger than 8192 bytes.

Snappy has a relatively good compression ratio, outperforms other compression algorithms during compression and is comparably quick during decompression.

Also shortens the field names for DynamoDB + MongoDB.


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/Compressions.java
```diff
@@ -0,0 +1,121 @@
...
+      throw new RuntimeException(e);
+    }
+    return out.toByteArray();
+  }
+
+  private static byte[] snappyCompress(byte[] uncompressed) {
+    return SnappyCompression.compress(uncompressed);
+  }
+
+  private static byte[] snappyUnompress(byte[] uncompressed) {
```
adutra, line 96:
```suggestion
  private static byte[] snappyUncompress(byte[] uncompressed) {
```

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/util/Compressions.java
```diff
@@ -0,0 +1,121 @@
...
+      ByteArrayOutputStream out = new ByteArrayOutputStream(uncompressed.length);
+      try (OutputStream def = new SnappyOutputStream(out)) {
+        def.write(uncompressed);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
+      }
+      return out.toByteArray();
+    }
+
+    static byte[] uncompress(byte[] uncompressed) {
```
adutra, line 111:
```suggestion
    static byte[] uncompress(byte[] compressed) {
```

PR #7799, opened 2023-12-04
Nit: Remove unsed Quarkus config options

snazy:
... fixing following warnings when running Quarkus tests:
```
Unrecognized configuration key "*" was provided; it will be ignored; verify that the dependency extension for this configuration is set or that you did not make a typo
```

### DIFF
--- a/build-logic/src/main/kotlin/nessie-jacoco.gradle.kts
+++ b/build-logic/src/main/kotlin/nessie-jacoco.gradle.kts
@@ -56,8 +56,6 @@ if (plugins.hasPlugin("io.quarkus")) {
       excludeClassLoaders =
         listOf("*QuarkusClassLoader") + (if (excluded != null) excluded else emptyList())
     }
-    systemProperty("quarkus.jacoco.report", "false")
-    systemProperty("quarkus.jacoco.reuse-data-file", "true")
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/nessie-jacoco.gradle.kts
```diff
@@ -56,8 +56,6 @@ if (plugins.hasPlugin("io.quarkus")) {
       excludeClassLoaders =
         listOf("*QuarkusClassLoader") + (if (excluded != null) excluded else emptyList())
     }
-    systemProperty("quarkus.jacoco.report", "false")
-    systemProperty("quarkus.jacoco.reuse-data-file", "true")
```
dimas-b, line 60: Was it fixed in Quarkus?

snazy: There's nothing to be fixed - it warns about non-consumed configs

dimas-b: Right, but we added this config a while ago to fix javaco reports. Do we not have the same problem now?

dimas-b (approved):

PR #7805, opened 2023-12-05
Update Persist API to detect and handle object mismatches

adutra:
This revision introduces exceptions for object mismatches in the `Persist.storeObj` API. These exceptions allow for better error handling when there are discrepancies between existing and expected objects. A key part of this change is the `ObjMismatchException` which provides detailed information about the mismatched objects.

### DIFF
--- a/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
+++ b/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
@@ -60,6 +60,7 @@ import javax.annotation.Nullable;
 import org.projectnessie.nessie.relocated.protobuf.ByteString;
 import org.projectnessie.versioned.storage.common.exceptions.CommitConflictException;
 import org.projectnessie.versioned.storage.common.exceptions.CommitWrappedException;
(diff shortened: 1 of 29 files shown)


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/ReferenceLogicImpl.java
```diff
@@ -485,7 +486,7 @@ CommitReferenceResult commitCreateReference(
             RefObj ref = ref(name, pointer, refCreatedTimestamp, extendedInfoObj);
             try {
               p.storeObj(ref);
```
adutra, line 488: I wonder: shouldn't we also check the return value here?

snazy (commented):
Hm...
originally I had in mind to change only the call-sites like this:
```java
if (!persist.storeObj(obj)) {
  if (!obj.equals(persist.fetchObj(obj.id())) {
    throw ...
  }
}
```
Now it looks that we have a multi-way return status - via the `boolean` return value and a new exception, that is in a couple of places "mapped back" to return `false`.

WRT to the new exception returning the all details to the caller. I suspect that's can expose internal information all the way up to the end user, potentially unveiling internal details, which should not be exposed.

Let me think about this a bit more.

adutra:
Closing because it became obsolete now.

PR #7806, opened 2023-12-05
Nessie HTTP: allow `..` in path elements

snazy:


api/client/src/main/java/org/projectnessie/client/http/impl/UriBuilder.java
```diff
@@ -99,9 +99,15 @@ public URI build() throws HttpClientException {
       for (int i = 0; i < l; i++) {
         char c = uri.charAt(i);
         if (c == '/') {
-          uriBuilder.append(encode(pathElement.toString()));
+          String element = pathElement.toString();
+          if ("..".equals(element)) {
```
dimas-b, line 103: Should this apply only to the base URI? Why would clients use `..` in the path?

snazy: It's actually intended for the base URI, but this way it's easier.
Otherwise you'd have to do this in `path()`, handle all (non-)trailing `/` and duplicate a lot of stuff there.

dimas-b: API inputs are already resolved into `element` at this point. Those can be `..` (at least from the UriBuilder's perspective)... This makes me uneasy :thinking: WDYT?

PR #7808, opened 2023-12-05
Expose `HttpClient` from `NessieApi` when available

snazy:


api/client/src/main/java/org/projectnessie/client/api/NessieApi.java
```diff
@@ -15,10 +15,17 @@
...
+import org.projectnessie.client.http.HttpClient;
+
 /** Base interface for all Nessie-API versions. */
 public interface NessieApi extends AutoCloseable {
 
   // Overridden to "remove 'throws Exception'"
   @Override
   void close();
+
+  default Optional<HttpClient> httpClient() {
```
dimas-b, line 28: How about `<T> T unwrap(Class<T> lowLevelClient)`?

dimas-b (approved):

PR #7830, opened 2023-12-12
Helm chart: remove mentions of legacy storage types

adutra:


ajantha-bhat (dismissed):
LGTM. It is as per the new model.

helm/nessie/values.yaml
```diff
@@ -1,5 +1,5 @@
 # -- The number of replicas to deploy (horizontal scaling).
-# Beware that replicas are stateless; don't set this number > 1 when using INMEMORY or ROCKS version store types.
+# Beware that replicas are stateless; don't set this number > 1 when using IN_MEMORY or ROCKSDB version store types.
```
ajantha-bhat, line 2: Line 31 and many places in this file still uses the old model names. Do we need to replace them too?

dimas-b (approved):

helm/nessie/values.yaml
adutra: Good catch, I will do a more thorough pass on the file.

dimas-b (approved):
I guess we need to make the PR title more generic now :)

PR #7834, opened 2023-12-12
[WIP] Persist: configurable custom object serialization

adutra:


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -391,19 +394,16 @@ public boolean storeObj(
...
+      @Nonnull Obj obj, @Nonnull PersistOptions options) throws ObjTooLargeException {
     checkArgument(obj.id() != null, "Obj to store must have a non-null ID");
     ByteString key = dbKey(obj.id());
 
-    int incrementalIndexSizeLimit =
-        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIncrementalIndexSizeLimit();
-    int indexSizeLimit =
-        ignoreSoftSizeRestrictions ? Integer.MAX_VALUE : effectiveIndexSegmentSizeLimit();
-    byte[] serialized = serializeObj(obj, incrementalIndexSizeLimit, indexSizeLimit);
+    SizeLimits limits = options.ignoreSoftSizeRestrictions() ? SizeLimits.NO_LIMITS : defaultLimits;
```
snazy, line 405: This is the same computation for all calls to `serializeObj`. Better move into `serializeObj` and make it non-static

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -401,7 +403,8 @@ private StoreIndex<CommitOp> createReferenceIndexForCommit(
 
   private List<StoreIndex<CommitOp>> updateExistingReferenceIndex(
       CommitObj commitObj, StoreIndex<CommitOp> newIncremental) {
-    int maxSize = persist.effectiveIndexSegmentSizeLimit();
+    int maxSize =
```
snazy, line 406: I'd prefer to leave the (now duplicated `Math.min(soft-limit, hard-limit)`) computation a concern of the actual `Persist` implementation.

snazy (commented):
It looks like this PR mixes two concerns:
* adding an additional constraint for custom objects (not sure we need that?)
* the serialization logic customization

I'd prefer to focus only on the serialization code in the PR.

PR #7837, opened 2023-12-13
Make server-side components use only Jakarta EE

snazy:
In the past we allowed using either Java 8 or 11+ to test Nessie services. Some time ago, Nessie started to require Java 11 for all services. This change is a consequence of the former two changes and removes dependencies to `javax.*` (Java EE), favoring `jakarta.*` (Jakarta EE).

Most code changes are pretty boring:
* Remove `javax.*` annotations
* Remove dependencies from the build scripts
* `JerseyServer` in the compatibility tests has the only "real" code change


compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
```diff
@@ -172,20 +174,24 @@ private static void withClass(
...
       String javaxClassName, Consumer<Class<?>> whenClassExists, boolean mandatory) {
-    try {
-      Class<?> clazz = Thread.currentThread().getContextClassLoader().loadClass(javaxClassName);
-      whenClassExists.accept(clazz);
-    } catch (NoClassDefFoundError | ClassNotFoundException e) {
+    Iterator<String> classNames =
+        List.of(
+                javaxClassName,
+                javaxClassName.replace("restjavax.", "restjakarta."),
+                javaxClassName.replace("restjavax.", "rest."))
```
dimas-b, line 181: How does it work with the old `rest` stuff when extensions are linked to the `jakarta` SPI.

snazy: This is only class-loading in a different class-loader using the code from the old versions.
Seems to work, since the compatibility-tests still pass.

compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
```diff
@@ -172,20 +174,24 @@ private static void withClass(
 
   private static void withJavaxClass(
```
adutra, line 175: Since we are moving towards jakarta and eventually javax will go away, it would be easier to understand if we inverted the logic:

```java
withJakartaClass(
    "org.projectnessie.services.restjakarta.ContentKeyParamConverterProvider",
    config::register,
    true);

private static void withJakartaClass(
    String jakartaClassName, Consumer<Class<?>> whenClassExists, boolean mandatory) {
  Iterator<String> classNames =
      List.of(jakartaClassName, jakartaClassName.replace("restjakarta.", "restjavax."))
          .iterator();
// ...
```

adutra: Are we sure we need this replacement: `javaxClassName.replace("restjavax.", "rest.")` ? I removed it, and the build passed.

snazy: Some older Nessie versions required it (a while ago). Maybe no longer needed.

compatibility/jersey/src/main/java/org/projectnessie/tools/compatibility/jersey/JerseyServer.java
dimas-b: Does Jersey support both sets of annotations?

snazy: Not sure I understand the background of the question? It works.

dimas-b: fair enough

dimas-b (approved):

PR #7846, opened 2023-12-14
Make `DiffParams` work with resteasy-reactive

snazy:
Fixes `jakarta.enterprise.inject.spi.DeploymentException: Method 'setFromRefWithHash' of class 'org.projectnessie.api.v1.params.DiffParams' is annotated with @PathParam annotation which is prohibited. Classes uses as @BeanParam parameters must have a JAX-RS parameter annotation on fields only.`

Note that this deployment exception does _not_ occur with classic resteasy.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/api/v1/params/DiffParams.java
+++ b/api/model/src/main/java/org/projectnessie/api/v1/params/DiffParams.java
@@ -30,8 +30,6 @@ public class DiffParams {
 
   private static final char HASH_SEPARATOR = '*';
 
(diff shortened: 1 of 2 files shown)


ajantha-bhat:
Please retrigger the build, @XN137 has fixed the `NesQuEIT` failure due to spark 3.2 removal from Iceberg master branch.

dimas-b (approved):
Supporting Resteasy reactive looks reasonable to me even if the main build does not use it ATM

PR #7854, opened 2023-12-14
Use `/` as the resteasy base path in Quarkus

snazy:
Updates the resource implementation classes to use the `api/...` prefix, but leave the client side interfaces with `@Path` annotations without that prefix.

### DIFF
--- a/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestConfigResource.java
+++ b/servers/rest-services/src/main/java/org/projectnessie/services/rest/RestConfigResource.java
@@ -18,6 +18,7 @@ package org.projectnessie.services.rest;
 import com.fasterxml.jackson.annotation.JsonView;
 import jakarta.enterprise.context.RequestScoped;
 import jakarta.inject.Inject;
+import jakarta.ws.rs.Path;
 import org.projectnessie.api.v1.http.HttpConfigApi;
 import org.projectnessie.model.NessieConfiguration;
 import org.projectnessie.model.ser.Views;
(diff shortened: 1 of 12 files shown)


servers/rest-services/src/main/java/org/projectnessie/services/rest/RestConfigResource.java
```diff
@@ -18,13 +18,15 @@
...
 import jakarta.inject.Inject;
+import jakarta.ws.rs.Path;
 import org.projectnessie.api.v1.http.HttpConfigApi;
 import org.projectnessie.model.NessieConfiguration;
 import org.projectnessie.model.ser.Views;
 import org.projectnessie.services.spi.ConfigService;
 
 /** REST endpoint to retrieve server settings. */
 @RequestScoped
+@Path("api/v1/config")
```
dimas-b, line 29: I wonder if we could have a parent `@RequestScoped` resource with class-based sub-resource locator methods to hold sub-path annotations. This way, we do not have to overwrite the whole API path. WDYT?

snazy: Sounds complicated.
I tried to "override" the existing resource class like this:
```
@Path("api/v1/config")
public class RelocatedResource extends RestConfigResource {...}
```
But that leads to both resource classes being registered - with the "original" one uninitialized - so much fun there :(

dimas-b: I see

dimas-b (approved):

dimas-b (approved):

PR #7884, opened 2023-12-20
OAuth2 client: support endpoint discovery

adutra:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Utils.java
```diff
@@ -30,4 +37,18 @@ static String randomAlphaNumString(int length) {
...
+
+  public static JsonNode fetchDiscoveryData(URI discoveryEndpoint, HttpClient httpClient) {
+    URI uri = new UriBuilder(discoveryEndpoint).path(".well-known/openid-configuration").build();
+    HttpResponse response = httpClient.newRequest().path(uri.getPath()).get();
+    Status status = response.getStatus();
+    if (status != Status.OK) {
+      throw new HttpClientException("Discovery endpoint returned status code " + status.getCode());
+    }
+    JsonNode data = response.readEntity(JsonNode.class);
+    if (!data.has("issuer") || !data.has("authorization_endpoint")) {
```
dimas-b, line 49: It looks like we expect the payload to follow the [OpenID Connect Discovery spec](https://openid.net/specs/openid-connect-discovery-1_0.html)... shall we make it explicit in docs?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Utils.java
```diff
@@ -30,4 +37,18 @@ static String randomAlphaNumString(int length) {
         .collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append)
         .toString();
   }
+
+  public static JsonNode fetchDiscoveryData(URI discoveryEndpoint, HttpClient httpClient) {
+    URI uri = new UriBuilder(discoveryEndpoint).path(".well-known/openid-configuration").build();
```
dimas-b, line 42: Since the OIDC spec indicates that `.well-known/openid-configuration` is appended to the `issuer` URL. In that context using "discovery" in config names seems a bit misleading... WDYT about renaming the `authentication.oauth2.token-endpoint` config to `authentication.oauth2.issuer`? We can still talk about "discovery" in the javadoc / web docs, but I think using `issuer` as the config name is more precise.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -201,7 +201,7 @@ private Tokens fetchNewTokens(String code) {
             .clientId(params.getClientId())
             .scope(params.getScope().orElse(null))
             .build();
-    HttpRequest request = httpClient.newRequest().path(params.getTokenEndpoint().getPath());
+    HttpRequest request = httpClient.newRequest().path(params.getResolvedTokenEndpoint().getPath());
```
dimas-b, line 204: I'm a bit uneasy about reusing URI paths on top of the base URI, especially with discovery supported. OpenID specs do not require endpoints to have the same host and they explicitly [allow](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) using different ports for different endpoints.

Would it be too much trouble to use full endpoint URIs for requests? It might require a parent PR for refactoring, I guess... but ultimately it looks like the "base URI" concept exists only in the Nessie Client, HTTP client implementations can send different requests to different servers. WDYT?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Utils.java
adutra: I agree that "discovery" is misleading, but the token endpoint is _not_ the issuer URL. I suggest instead:

* Use the term "Issuer URL" for what is currently named the "discovery endpoint";
* Rename `authentication.oauth2.discovery-endpoint` to `authentication.oauth2.issuer-url`;
* Keep using the terms "token endpoint" and "authorization endpoint" since these names adhere to standards.
* Also keep the existing options `authentication.oauth2.token-endpoint` and `authentication.oauth2.auth-endpoint`; they override whatever the discovery process finds out.

WDYT?

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -53,24 +53,34 @@ public final class NessieConfigConstants {
    */
   public static final String CONF_NESSIE_AUTH_TOKEN = "nessie.authentication.token";
 
+  /**
+   * Config property name ({@value #CONF_NESSIE_OAUTH2_DISCOVERY_ENDPOINT}) for the OAuth2
+   * authentication provider. The root URL of the OAuth2 provider, which will be used for
+   * discovering supported endpoints and their locations. For Keycloak, this is typically the realm
+   * URL: {@code https://<keycloak-server>/realms/<realm-name>}.
```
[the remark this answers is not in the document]
adutra: Whoops indeed in example is wrong. I will amend.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
adutra: Yes I agree, but the consequence of doing as you suggested is that we'd need one `HttpClient` per endpoint. But let's try, it's probably more future-proof anyway.

dimas-b: I suggest refactoring `HttpClient` to allow independent URIs in each request (as the underlying HTTP client allow)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Utils.java
dimas-b: That's exactly what I mean :)

dimas-b: I mistyped in my original message :facepalm: I meant to rename the discovery config, of course.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
adutra: Then in this case I suggest we do it in 2 steps: in this PR I will make it so that there is one `HttpClient` per endpoint, then in a follow-up PR we refactor `HttpClient`. I think this is acceptable because so far there is only one endpoint constantly used, the token one.

dimas-b: SGTM :+1:

dimas-b (approved):

PR #7888, opened 2023-12-22
WIP: OAuth2 client: support manual authorization code input

adutra:
Requires #7872.


adutra:
Unfortunately this idea is not going to be feasible IRL: reading from System.in in a Spark shell session for example makes the OAuth2 client compete with the Spark shell interpreter for input coming from System.in. This results in some key strokes being "swallowed" and makes the whole input method unreliable. I think we'd need to explore more appropriate flows, e.g. Device Authorization Flow.

PR #7893, opened 2023-12-26
Fix Helm CI tests

adutra:
... and fix a few leftovers from #6890


ajantha-bhat (approved):
LGTM. 
Thanks for fixing it.

PR #7921, opened 2024-01-08
Fix resource leak in DeviceCodeFlow.java

adutra:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/DeviceCodeFlow.java
```diff
@@ -127,10 +120,17 @@ private DeviceCodeResponse requestDeviceCode() {
...
             .scope(config.getScope().orElse(null))
             .build();
-    return deviceAuthEndpointClient
-        .newRequest()
-        .postForm(body)
-        .readEntity(DeviceCodeResponse.class);
+    try (HttpClient deviceAuthEndpointClient =
+        config
+            .newHttpClientBuilder()
+            .setBaseUri(config.getResolvedDeviceAuthEndpoint())
```
snazy, line 126: Do we really need a _new_ HttpClient here? Mean, can't we reuse the `tokenEndpointClient`, assuming we can lift the artificial base-URI restriction (somehow related: #7806)?

Note: `this.executor` is never shutdown

adutra: See https://github.com/projectnessie/nessie/pull/7884#discussion_r1437165699 for context. In short, we cannot guarantee that both endpoints share a common root URL.

> Note: this.executor is never shutdown

🤦 fixing.

adutra: WDYT about https://github.com/projectnessie/nessie/pull/7922 ? That would allow to share the same client, regardless of base URIs.

snazy: I think it's okay to wait for #7922 for this

adutra:
Superseded by #7925

PR #7928, opened 2024-01-09
HttpClient: ability to define per-request authentication

adutra:


dimas-b (dismissed):

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaRequest.java
```diff
@@ -162,6 +163,10 @@ public org.projectnessie.client.http.HttpResponse executeRequest(Method method,
           // ignore
         }
       }
+      HttpAuthentication auth = this.auth;
```
snazy, line 166: Looks like this piece should live in `BaseHttpRequest` (shared w/ other implementations)

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2ClientConfig.java
```diff
@@ -257,7 +257,7 @@ void testFromConfig(
               "resolvedTokenEndpoint",
               "resolvedAuthEndpoint")
           .isEqualTo(expected);
-      assertThat(actual.newHttpClientBuilder()).isNotNull();
+      assertThat(actual.getHttpClient()).isNotNull();
```
snazy, line 260: try-with-resource on `actual` missing

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -143,18 +141,44 @@ URI getResolvedDeviceAuthEndpoint() {
...
-  HttpClient.Builder newHttpClientBuilder() {
+  /**
+   * Returns the {@link HttpClient} that will be used to communicate with the OAuth2 server.
+   *
+   * <p>Note that it does not have any authentication configured, so each request must be
+   * authenticated explicitly. The appropriate authentication object can be obtained from {@link
+   * #getBasicAuthentication()}.
+   */
+  @Value.Lazy
+  HttpClient getHttpClient() {
```
snazy, line 165: Can't we share the "main" HttpClient for OAuth2 instead of creating a new one?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -143,18 +141,44 @@ URI getResolvedDeviceAuthEndpoint() {
...
         .setSslContext(getSslContext().orElse(null))
         .setDisableCompression(true)
-        .addResponseFilter(this::checkErrorResponse);
+        .addResponseFilter(this::checkErrorResponse)
+        .build();
+  }
+
+  @Override
+  public void close() {
+    getHttpClient().close();
```
snazy, line 176: This would close a "shared" `HttpClient`, if the caller provides one.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -49,7 +49,7 @@
  */
 @Value.Immutable
 @SuppressWarnings("immutables:subtype")
-abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig {
+abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig, AutoCloseable {
```
snazy, line 52: Seems that making this config holding resources (adding `AutoCloseable`) adds a bunch of overhead (need to close the config in addition to the client).

I wonder whether it's worth to have a (lazily started) executor for each `HttpClient`, which then has the same lifecycle as the `HttpClient`. We can make that one have the same "thread-keep-alive" setting as the `OAuth2TokenRefreshExecutor`. WDYT?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
adutra: That would create a chicken-and-egg problem unfortunately.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
adutra: Currently the user cannot provide its own `HttpClient`, its defined in `OAuth2ClientConfig` (which is package-private) and not in `OAuth2AuthenticatorConfig` (which is public).

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
snazy: Wonder whether the configuration needs to resolve the "missing parts" eagerly or whether it could do that lazily, latter would allow re-using the "current" client (in theory).
I'm a bit concerned that we have more than one "closeable" object, that callers need to take care of.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
adutra: 1. I agree that making `OAuth2ClientConfig` auto-closeable wasn't a great idea. Let's revert that and have `OAuth2Client` close relevant resources as before.
2. I like your idea of having a (lazily started) executor for each `HttpClient`. Can we tackle that separately though?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
adutra: Thinking about reusing the "main" `HttpClient`: I think this would be doable by changing the `start()` method into something like `onClientReady(HttpClient ignored)`

```java
public interface NessieAuthentication extends AutoCloseable {

  default void onClientReady(HttpClient client) {
    // no-op
  }
...
}
```

That is very easy to do imho, but I have a few concerns: 

1. Reusing the main `HttpClient` means that users could set custom request and response filters and these filters could potentially interfere with the `OAuth2Client` internal error handling.
2. The OAuth server and the Nessie server could require different SSL configurations. Reusing the same `HttpClient` would make this impossible to do.

snazy: I wonder whether we need more abstraction from the underlying implementation in general. So one "backend" shared by multiple `HttpClient`s. I suspect, we'll need that in the future.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
snazy: > Can we tackle that separately though?
sure

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaRequest.java
```diff
@@ -162,6 +162,7 @@ public org.projectnessie.client.http.HttpResponse executeRequest(Method method,
           // ignore
         }
       }
+      cleanUp();
```
snazy, line 165: Can you put this in a `finally`?

api/client/src/main/java/org/projectnessie/client/auth/AwsAuthenticationProvider.java
```diff
@@ -85,6 +84,8 @@ private static Region awsRegion(String regionName) {
   private static class AwsAuthentication implements HttpAuthentication {
     private final Region region;
     private final AwsCredentialsProvider awsCredentialsProvider;
+    private final ObjectMapper objectMapper;
+    private final Aws4Signer signer;
```
snazy, line 88: I think we do not need to memoize the `signer` instance, the implementations don't keep any state.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -239,7 +232,10 @@ Tokens fetchNewTokens() {
           ClientCredentialsTokensRequest body =
               ImmutableClientCredentialsTokensRequest.builder().scope(scope).build();
           HttpResponse httpResponse =
-              httpClient.newRequest(config.getResolvedTokenEndpoint()).postForm(body);
+              httpClient
```
snazy, line 235: Nit: add a helper method for the `httpClient.newRequest ... postForm()` (it's duplicated often)

snazy (commented):
LGTM - one nit left

snazy (approved):

PR #7931, opened 2024-01-10
Caching behavior per object type

snazy:


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/CustomObjType.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    long calculateExpiresAt(T obj, long currentTimeMicros);
+  }
+
+  public static <T extends Obj> ObjType dynamicCaching(
+      String name,
+      String shortName,
+      Class<T> targetClass,
+      CacheExpireCalculation<T> cacheExpireCalculation) {
+    return new CustomObjType<>(name, shortName, targetClass) {
```
adutra, line 78: We might want to implement `equals` and `hashCode` for `CustomObjType`, and only use `name` or `shortName` since they are unique. Wdyt?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -100,34 +115,58 @@ public void clear(@Nonnull String repositoryId) {
...
-  static final class CacheKey {
+  /**
+   * Class used for both the cache key and cache value including the expiration timestamp. This is
+   * (should be) more efficient (think: mono-morphic vs bi-morphic call sizes) and more GC/heap
+   * friendly (less object instances) than having different object types.
+   */
+  static final class CacheKeyValue {
 
-    static final int HEAP_OVERHEAD = 3 * JAVA_OBJ_HEADER;
+    static final int HEAP_OVERHEAD = 16 + 3 * JAVA_OBJ_HEADER;
```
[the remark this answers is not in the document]
snazy: Good point about my math being wrong here.
We need some way to estimate the object heap size, overshooting is fine though.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -100,34 +115,58 @@ public void clear(@Nonnull String repositoryId) {
...
+
+  private static CacheKeyValue cacheKeyValue(
+      String repositoryId, ObjId id, byte[] value, long expiresAt) {
+    return new CacheKeyValue(repositoryId, id, value, expiresAt);
   }
 
-  static final class CacheKey {
+  /**
+   * Class used for both the cache key and cache value including the expiration timestamp. This is
+   * (should be) more efficient (think: mono-morphic vs bi-morphic call sizes) and more GC/heap
```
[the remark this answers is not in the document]
snazy: Why is it "premature"? You'd need one object to have a composite key of repo-ID + obj-ID plus a holder for the value having the expiration and the actual value.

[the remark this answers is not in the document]
adutra: Because in terms of space, having 2 separate objects, `CacheKey` and `CacheValue`, would take only 16 bytes more (2 object headers instead of one) when storing an entry, so it doesn't seem it would make a huge difference. OTOH, cache lookups only need the key, so having smaller keys would save some space.

As for GC... I don't know enough about GC internals to tell if it's more efficient to GC a big object, or 2 smaller ones. I've read that it's generally better to have fewer but larger objects indeed, but I've also read that G1, for instance, is sensitive to very large objects.

Anyways, it's OK to leave as is.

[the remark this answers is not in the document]
snazy: Yes, G1 is very allergic to very large object, but "very large" means really many megabytes (humongous objects).

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -49,8 +51,8 @@ static ImmutableCaffeineCacheBackend.Builder builder() {
   abstract MeterRegistry meterRegistry();
 
   @Value.Derived
-  Cache<CacheKey, byte[]> cache() {
-    Caffeine<CacheKey, byte[]> cacheBuilder =
+  Cache<CacheKeyValue, CacheKeyValue> cache() {
+    Caffeine<CacheKeyValue, CacheKeyValue> cacheBuilder =
         Caffeine.newBuilder().maximumWeight(capacity() * 1024L * 1024L).weigher(this::weigher);
```
adutra, line 56: Have you considered using Caffeine's `expireAfter()` utility here? Stg like below seems doable:

```java
.expireAfter(
                new Expiry<>() {
                  @Override
                  public long expireAfterCreate(
                      CacheKeyValue key, CacheKeyValue value, long currentTime) {
                    return value.expiresAt == CACHE_UNLIMITED ? Long.MAX_VALUE : value.expiresAt;
                  }

                  @Override
                  public long expireAfterUpdate(
                      CacheKeyValue key,
                      CacheKeyValue value,
                      long currentTime,
                      long currentDuration) {
                    return expireAfterCreate(key, value, currentTime);
                  }

                  @Override
                  public long expireAfterRead(
                      CacheKeyValue key,
                      CacheKeyValue value,
                      long currentTime,
                      long currentDuration) {
                    return currentDuration;
                  }
                });
```

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjType.java
```diff
@@ -25,4 +27,21 @@ public interface ObjType {
...
+   * Allows an object type to define how long a particular object instance can be cached.
+   *
+   * <p>{@value #CACHE_UNLIMITED}, which is the default implementation, defines that an object
+   * instance can be cached forever.
+   *
+   * <p>{@value #NOT_CACHED} defines that an object instance must never be cached.
+   *
+   * <p>A positive value defines the maximum age of an object
+   */
+  default long cachedObjectExpiresAtMicros(Obj obj, LongSupplier clock) {
```
adutra, line 41: Would it make sense to declare a duration here, instead of the wall time? This would leave the actual time computation to the caching backend and wouldn't require a clock at all.

snazy: Wouldn't work for all cases, especially when the object's been read from the database.

adutra: >  when the object's been read from the database.

Sorry, I'm confused, can you elaborate? On a normal cache setup, the object would be read from the DB, put in the cache, and its expiry time would be = load time + object's cache duration.

Are you implying that objects would have "hard" expiration times, past which they should be evicted _from both the cache and the database_? But if that's the case, shouldn't all `Persist` implementations be modified to check if the fetched object is expired?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -48,10 +54,50 @@ static ImmutableCaffeineCacheBackend.Builder builder() {
...
+                  public long expireAfterCreate(
+                      CacheKeyValue key, byte[] value, long currentTimeNanos) {
+                    long expire = key.expiresAt;
+                    if (expire == CACHE_UNLIMITED) {
+                      return Long.MAX_VALUE;
+                    }
+                    if (expire == NOT_CACHED) {
+                      return 0L;
+                    }
+                    long currentMicros = NANOSECONDS.toMicros(currentTimeNanos);
```
adutra, line 77: Nit: I think ti would be slightly easier to store nanos directly in `CacheKeyValue.expiresAt`.

snazy: done

adutra (approved):

PR #7937, opened 2024-01-10
fix free-disk-space action

XN137:
the buildpack-deps images have been removed from the runners: https://github.com/actions/runner-images/pull/9093

so we no longer need to clear them up
(the command was failing any workflow that uses the action)

```
Error response from daemon: No such image: buildpack-deps:buster
Error response from daemon: No such image: buildpack-deps:bullseye
```

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -28,8 +28,6 @@ runs:
           node:16-alpine \
           node:18 \
           node:18-alpine \
-          buildpack-deps:buster \
-          buildpack-deps:bullseye \
           moby/buildkit:latest \
           alpine:3.16 \
           alpine:3.17 \


adutra:
We still should remove all existing docker images. I suggest instead:

```yaml
    - name: Remove unused Docker images
      shell: bash
      # See https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#cached-docker-images
      run: |
        docker image rm node:16 || true
        docker image rm node:16-alpine || true
        docker image rm node:18 || true
        docker image rm node:18-alpine || true
        docker image rm buildpack-deps:buster || true
        docker image rm buildpack-deps:bullseye || true
        docker image rm moby/buildkit:latest || true
        docker image rm alpine:3.16 || true
        docker image rm alpine:3.17 || true
        docker image rm alpine:3.18 || true
        docker image rm debian:10 || true
        docker image rm debian:11 || true
        docker image rm ubuntu:22.04 || true
        docker image rm ubuntu:20.04 || true
```

wdyt?

dimas-b (approved):

dimas-b:
> `|| true`

adutra (approved):

PR #7946, opened 2024-01-12
remove unused jersey-test-framework-provider dependencies

XN137:
one is supposed to select exactly one of the providers: https://eclipse-ee4j.github.io/jersey.github.io/documentation/latest/test-framework.html

our usage of JerseyTest uses the default (=grizzly2), so we can remove the hard dependency on the other providers.


adutra (approved):
I also tested your changes against the Iceberg codebase to make sure we aren't breaking anything there.

snazy (approved):

PR #7947, opened 2024-01-12
Coordinated Tasks

snazy:
Adds an API and SPI to coordinate possibly long-running work to build and persist an `Obj`.

The API is pretty much just this:
```java
Persist persist; // provided
TasksService service; // provided
Tasks tasks = service.forPersist(persist);

SomeTaskRequest taskRequest; // specific for each TaskType/ObjType

CompletionStage<SomeTaskObj> finalResult = tasks.submit(taskRequest);
```

The returned `CompletionStage` can be chained in other `CompletionStage`s/`CompletableFuture`s and other implementations like Mutiny's `Uni` or Vert.X's `Future`.

The implementation ensures that only one "worker job" runs for each task, even across multiple Nessie instances.

Each task is identified using an implementation of a `TaskRequest` instance. Each persisted `TaskObj` has a state, which can be a final state (no more changes happening) like SUCCESS and FAILURE, or a transient state like RUNNING or ERROR_RETRY. WRT final states have no cache expiration, but tranient states have a task-specific cache expiration.

Recent changes for conditional-updates and "dynamic cache expiration" were prerequisites for this change.


dimas-b (commented):
LGTM overall :+1:

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskObj.java
```diff
@@ -0,0 +1,72 @@
...
+    @CanIgnoreReturnValue
+    Builder type(ObjType type);
+
+    @CanIgnoreReturnValue
+    Builder taskState(TaskState taskState);
+
+    TaskObj build();
+  }
+
+  CacheExpireCalculation<? extends TaskObj> _TASK_DEFAULT_CACHE_EXPIRE =
```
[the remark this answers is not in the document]
snazy: done (but `BasicTaskObj` is a test class)

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,509 @@
...
+      TaskObj taskObj = (TaskObj) obj;
+      switch (taskObj.taskState().status()) {
+        case FAILURE:
+          metrics.taskHasFinalFailure();
+          return failedStage(taskDefinition.stateAsException(taskObj));
+        case SUCCESS:
+          metrics.taskHasFinalSuccess();
+          return completedStage(taskObj);
+        default:
+          // task object exists but has a non-final state, handle it asynchronously
```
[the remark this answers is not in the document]
snazy: done

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,509 @@
...
+
+      if (params.persist.updateConditional(obj, retryState)) {
+        metrics.taskLostReassigned();
+        issueLocalTaskExecution(params, retryState);
+      } else {
+        metrics.taskLostReassignRace();
+        reattemptAfterRace(params);
+      }
+    } else {
+      async.schedule(() -> tryLocal(params), state.retryNotBefore());
```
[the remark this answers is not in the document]
snazy: Why? All outcomes are already handled

tasks/service/impl/src/test/java/org/projectnessie/nessie/tasks/service/tasktypes/BasicTaskRequest.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Value.Parameter(order = 1)
+  String taskParameter();
+
+  @Value.Parameter(order = 2)
+  ObjId objId();
+
+  @Value.Parameter(order = 3)
+  Supplier<CompletionStage<TaskObj.Builder>> taskCompletionStageSupplier();
```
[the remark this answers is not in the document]
snazy: yea - but it's just a test object

tasks/service/spi/src/main/java/org/projectnessie/nessie/tasks/service/spi/TaskTypeController.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  TaskState retryableErrorTaskState(Clock clock, TaskObj base, Throwable t);
+
+  /**
+   * Check whether the given exception, which has been thrown from a {@linkplain
+   * #submitExecution(TasksAsync, TaskRequest) task execution}, can be retried. Retryable exception
+   * result in a {@linkplain TaskStatus#ERROR_RETRY error-retry} state, non-retryable in a final
+   * {@linkplain TaskStatus#FAILURE failure} state
+   */
+  default boolean isRetryableError(Throwable t) {
```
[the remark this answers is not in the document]
snazy: isRetryableError takes a Throwable and then calls retryable/failureTaskState

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
```diff
@@ -0,0 +1,114 @@
...
+                completableFuture.completeExceptionally(new CompletionException(t));
+              }
+            });
+
+    return new VertxScheduledHandle(timerId, completableFuture);
+  }
+
+  @Override
+  public void cancel(ScheduledHandle handle) {
+    vertx.cancelTimer(((VertxScheduledHandle) handle).timerId);
```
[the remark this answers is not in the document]
snazy: done

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/testing/TestingTasksAsync.java
```diff
@@ -0,0 +1,136 @@
...
+import org.projectnessie.nessie.tasks.async.TasksAsync;
+
+/**
+ * {@link TasksAsync} implementation for testing purposes. Collects tasks to run in an ordered
+ * collection, tasks are run by calling {@link #doWork()}. There are no threads nor is there any
+ * other concurrent code involved. Maybe (not) surprisingly, this class is <em>not</em> thread safe.
+ */
+public class TestingTasksAsync implements TasksAsync {
+
+  private final TreeSet<ScheduledTask> scheduledTasks = new TreeSet<>();
```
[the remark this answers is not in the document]
snazy: done

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
```diff
@@ -0,0 +1,40 @@
...
+
+  <R> CompletionStage<R> supply(Supplier<R> runnable);
+
+  ScheduledHandle schedule(Runnable runnable, Instant scheduleNotBefore);
+
+  void cancel(ScheduledHandle handle);
+
+  default long calculateDelay(Clock clock, long minimumDelayMillis, Instant scheduleNotBefore) {
+    long retryEarliestEpochMillis = scheduleNotBefore.toEpochMilli();
+    long delayMillis = retryEarliestEpochMillis - clock.millis();
```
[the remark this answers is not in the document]
snazy: How? `scheduleNotBefore` can be legibly very "old" and also far in the future, so `delayMillis` can become pretty high or negative, latter bounded by the `minimumDelayMillis`.

tasks/service/spi/src/main/java/org/projectnessie/nessie/tasks/service/spi/TaskTypeController.java
```diff
@@ -0,0 +1,89 @@
...
+import org.projectnessie.nessie.tasks.api.TaskType;
+import org.projectnessie.nessie.tasks.async.TasksAsync;
+
+/**
+ * Implements the logic required for each {@linkplain TaskType task type}.
+ *
+ * <p>This included the asynchronous execution, calculation or retry timestamps and lost-task
+ * timeouts and mapping between exceptions and task state.
+ */
+public interface TaskTypeController {
```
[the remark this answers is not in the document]
snazy: Updated the code quite a bit.

(However, it was possible to pass runtime services around via `TaskRequest`.)

gradle/projects.main.properties
```diff
@@ -38,6 +38,10 @@ nessie-server-store=servers/store
 nessie-server-store-proto=servers/store-proto
 nessie-content-generator=tools/content-generator
 nessie-protobuf-relocated=tools/protobuf-relocated
+nessie-tasks-api=tasks/api
+nessie-tasks-service=tasks/service
```
adutra, line 42: This project's path is a parent directory of the two following ones, is that on purpose?

Also the spi module seems undeclared (see my other comment).

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,514 @@
...
+    currentTasks.remove(taskKey);
+  }
+
+  private void tryLocal(ExecParams params) {
+    try {
+      metrics.taskAttempt();
+      LOGGER.trace("{}: Task evaluation attempt for {}", name, params);
+
+      TaskObj obj = (TaskObj) params.persist.fetchObj(params.objId());
+      // keep in mind: `obj` might be a locally cached instance that is not in sync w/ the
```
adutra, line 167: In this case shouldn't we "refresh" the task object and force a fetch from the DB?

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/testing/TestingTasksAsync.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.function.Supplier;
+import org.projectnessie.nessie.tasks.async.ScheduledHandle;
+import org.projectnessie.nessie.tasks.async.TasksAsync;
+
+/**
+ * {@link TasksAsync} implementation for testing purposes. Collects tasks to run in an ordered
+ * collection, tasks are run by calling {@link #doWork()}. There are no threads nor is there any
+ * other concurrent code involved. Maybe (not) surprisingly, this class is <em>not</em> thread safe.
+ */
+public class TestingTasksAsync implements TasksAsync {
```
adutra, line 34: Could this be moved to a `testFixtures` folder?

tasks/service/impl/src/test/java/org/projectnessie/nessie/tasks/service/tasktypes/BasicTaskObj.java
```diff
@@ -0,0 +1,69 @@
...
+import org.projectnessie.nessie.tasks.api.TaskState;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.ObjType;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableBasicTaskObj.class)
+@JsonDeserialize(as = ImmutableBasicTaskObj.class)
+public interface BasicTaskObj extends TaskObj {
+  ObjType TYPE =
+      dynamicCaching("basic", "basic", BasicTaskObj.class, TaskObj.taskDefaultCacheExpire());
```
adutra, line 35: Nit: "basic" seems vague for a task object name. Maybe "basic-task" / "bt" ?

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,489 @@
...
+  CompletionStage<TaskObj> submit(Persist persist, TaskRequest taskRequest) {
+    ObjId objId = taskRequest.objId();
+
+    // Try to get the object and immediately return if it has a final state. We expect to hit final
+    // states way more often, so preventing the concurrent-hash-map interactions and especially the
+    // asynchronous task handling improves the implementation.
+    // TODO using `fetchObj()` would be wrong here, because it is *synchronous* and can block.
+    //  Options:
+    //  a) remove this "optimization"
+    //  b) add a `getObjIfCached(ObjId)` --> chosen for now
```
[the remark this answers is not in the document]
adutra: Or simply `getNow` to mimic the `CompletableFuture` API.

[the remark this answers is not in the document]
snazy: Renamed to `getImmediate`

tasks/service/spi/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+plugins {
+  id("nessie-conventions-server")
+  id("nessie-jacoco")
+  alias(libs.plugins.jmh)
+}
+
+extra["maven.name"] = "Nessie - Tasks - SPI"
```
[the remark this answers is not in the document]
snazy: ah - left-over, removed

gradle/projects.main.properties
snazy: oops - removed

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
snazy: It might be stale. Cache-expiration however should make it pretty safe though.

snazy: And it would be refreshed implicitly in the "next round"

tasks/service/impl/src/test/java/org/projectnessie/nessie/tasks/service/tasktypes/BasicTaskObj.java
snazy: hm - it's testing code only

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
```diff
@@ -0,0 +1,116 @@
...
+  private final long minimumDelayMillis;
+
+  public VertxTasksAsync(Vertx vertx, Clock clock, long minimumDelayMillis) {
+    this.vertx = vertx;
+    this.clock = clock;
+    this.minimumDelayMillis = minimumDelayMillis;
+  }
+
+  @Override
+  public CompletionStage<Void> call(Runnable runnable) {
```
adutra, line 40: This impl could be promoted to default impl at interface level.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/testing/TestingTasksAsync.java
snazy: moved to test/src/main

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public CompletionStage<Void> toCompletionStage() {
+      return stage;
+    }
+
+    @Override
+    public void cancel() {
+      vertx.cancelTimer(timerId);
```
adutra, line 113:
```suggestion
      vertx.cancelTimer(timerId);
      stage.toCompletableFuture().cancel(true);
```

tasks/service/impl/build.gradle.kts
```diff
@@ -0,0 +1,79 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+plugins {
+  id("nessie-conventions-server")
+  id("nessie-jacoco")
+  alias(libs.plugins.jmh)
```
adutra, line 20: I don't see any benchmarks in this module.

tasks/service/impl/src/test/java/org/projectnessie/nessie/tasks/service/tasktypes/BasicTaskObj.java
adutra: Sorry.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/ScheduledHandle.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.nessie.tasks.async;
+
+import java.util.concurrent.CompletionStage;
+
+public interface ScheduledHandle {
```
adutra, line 20: This interface seems a bit overkill, since it's just a wrapped completion stage. I think we can get rid of it and adapt the 2 impls:

```java
// TasksAsync
CompletionStage<Void> schedule(Runnable runnable, Instant scheduleNotBefore);

// JavaPoolTasksAsync impl:
completable.whenComplete(
        (v, t) -> {
          if (t instanceof CancellationException) {
            future.cancel(true); // allow interruption of blocking tasks
          }
        });

// VertxTasksAsync impl:
completableFuture.whenComplete(
        (v, t) -> {
          if (t instanceof CancellationException) {
            vertx.cancelTimer(timerId); // won't interrupt blocking tasks!
          }
        });
```

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
snazy: Why? those are different for every backend.

tasks/service/async/src/test/java/org/projectnessie/nessie/tasks/async/BaseTasksAsync.java
```diff
@@ -0,0 +1,236 @@
...
+    TasksAsync async = tasksAsync();
+    ScheduledHandle handle =
+        async.schedule(() -> {}, async.clock().instant().plus(1500, ChronoUnit.MILLIS));
+    soft.assertThatCode(handle::cancel).doesNotThrowAnyException();
+  }
+
+  @Test
+  @Disabled(
+      "Testing scheduled/timer cancellation requires either tweaking the inner workings of Vert.X / ScheduledExecutorService or results in extremely long test durations.")
+  public void cancelReallyWorks() throws InterruptedException {
```
adutra, line 207: I think it's important to test cancellation, and besides, Vertx has weird cancellation effects.

For the Java pool impl, you could use the below test:

```java
  public void cancelReallyWorks() throws InterruptedException {
    TasksAsync async = tasksAsync();

    CountDownLatch started = new CountDownLatch(1);
    CountDownLatch finished = new CountDownLatch(1);

    CompletionStage<Void> handle =
        async.schedule(
            () -> {
              try {
                started.countDown();
                new Semaphore(0).acquire();
              } catch (InterruptedException ignored) {
                finished.countDown();
              }
            },
            async.clock().instant());
    soft.assertThat(handle).isNotNull();

    assertThat(started.await(10, TimeUnit.SECONDS)).isTrue();
    handle.toCompletableFuture().cancel(true);
    assertThat(finished.await(10, TimeUnit.SECONDS)).isTrue();
  }
```

(The example above includes my suggestion of getting rid of `ScheduleHandle`.)

For the Vrtx impl, it's tricky because Vertx, being event-loop based, does _not_ interrupt threads when the timer is cancelled. So indeed it's very hard to test whether cancellation works because, in practice, timer tasks _cannot_ be cancelled if they were started already. Cancelling the timer only has an effect if the timer wasn't triggered yet.

As a side note: it's important to note that for the Vertx impl, runnables scheduled with `TasksAsync.schedule()` _must not block_.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
adutra: Because it seems like the natural way to implement `call` as a special case of `supply`.

tasks/service/async/src/test/java/org/projectnessie/nessie/tasks/async/BaseTasksAsync.java
```diff
@@ -0,0 +1,258 @@
...
+    AtomicReference<Throwable> error = new AtomicReference<>();
+    handle.whenComplete(
+        (r, t) -> {
+          cancelled.countDown();
+          result.set(r);
+          error.set(t);
+        });
+
+    handle.toCompletableFuture().cancel(true);
+    soft.assertThat(cancelled.await(10, TimeUnit.SECONDS)).isTrue();
```
adutra, line 237: Hmm this is only testing that `CompletionStage.whenComplete` works as designed. It's not testing that the runnable submitted to `async.schedule` was unscheduled / cancelled / interrupted. I think that, short of better options for Vertx, we could delete this test altogether and keep the Java-specific one in its own test class. Wdyt?

tasks/service/async/src/test/java/org/projectnessie/nessie/tasks/async/BaseTasksAsync.java
```diff
@@ -0,0 +1,258 @@
...
+        async.schedule(() -> mark.set(true), async.clock().instant().plus(1500, ChronoUnit.MILLIS));
+    handle.toCompletableFuture().cancel(true);
+    soft.assertThat(mark).isFalse();
+  }
+
+  @Test
+  public void cancelReallyWorks() throws InterruptedException {
+    TasksAsync async = tasksAsync();
+
+    CountDownLatch started = new CountDownLatch(1);
```
adutra, line 211: Looks like `started` is not needed anymore.

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCachingInmemoryPersist.java
```diff
@@ -15,8 +15,45 @@
...
+
+  @Nested
+  @ExtendWith({PersistExtension.class, SoftAssertionsExtension.class})
+  public class CacheSpecific {
+    @InjectSoftAssertions protected SoftAssertions soft;
+
+    @NessiePersist protected Persist persist;
+
+    @Test
+    public void getIfCached() throws Exception {
```
adutra, line 47:
```suggestion
    public void getImmediate() throws Exception {
```

tasks/service/async/src/test/java/org/projectnessie/nessie/tasks/async/pool/TestJavaPoolTasksAsync.java
```diff
@@ -0,0 +1,76 @@
...
+    assertTrue(executorService.awaitTermination(30, TimeUnit.SECONDS));
+  }
+
+  @Override
+  protected TasksAsync tasksAsync() {
+    return new JavaPoolTasksAsync(executorService, Clock.systemUTC(), 1L);
+  }
+
+  @Test
+  public void cancelRunningWithInterrupt() throws InterruptedException {
```
adutra, line 53: Thanks for considering my suggestion. How about doing similar tests for `call` and `supply` ? The tests should be very similar to this one imho.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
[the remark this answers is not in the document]
dimas-b: Maybe compare `retryEarliestEpochMillis < clock.millis()` and use `minimumDelayMillis` in that case?

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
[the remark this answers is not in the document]
dimas-b: It was just a thought :) Totally optional.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/ScheduledHandle.java
dimas-b: I'm not actually sure about using `CompletionStage` as a scheduled task's handle. `CompletionStage` itself does not have a `cancel()` method.

If we do `CompletionStage.toCompletableFuture().cancel()`, IMHO, the relationship of the `cancel()` call to the original `CompletableFuture` is very indirect and implementation specific to rely upon.

I'd prefer to stick with a custom `ScheduledHandle` and a dedicated `cancel()` method in it.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
```diff
@@ -0,0 +1,44 @@
...
+    return supply(
+        () -> {
+          runnable.run();
+          return null;
+        });
+  }
+
+  <R> CompletionStage<R> supply(Supplier<R> runnable);
+
+  CompletionStage<Void> schedule(Runnable runnable, Instant scheduleNotBefore);
```
dimas-b, line 36: See my other comment about cancellation.

tasks/service/spi/src/main/java/org/projectnessie/nessie/tasks/service/spi/TaskTypeController.java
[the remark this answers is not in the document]
dimas-b: We currently have this:
```
boolean retryable = behavior.isRetryableError(failure);
TaskState newState =
    retryable
        ? behavior.retryableErrorTaskState(async.clock(), expected, failure)
        : behavior.failureTaskState(failure);
```
What I'm suggesting is something like:
```
TaskState newState = behavior.asState(async.clock(), expected, failure);
if (newState.isRetryable()) { ... }
```

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskObj.java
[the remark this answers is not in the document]
dimas-b: Looks so real :)

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskRequest.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.nessie.tasks.api;
+
+import java.util.concurrent.CompletionStage;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.ObjType;
+
+/** Base interface for value objects that identify a task's input parameters. */
+public interface TaskRequest<T extends TaskObj, B extends TaskObj.Builder> {
```
dimas-b, line 23: `T` is not used. Also, `B` should be a builder for `T`, but it is not apparent. 

How about this?
`TaskRequest<T extends TaskObj, B extends TaskObj.Builder<T>>` and
`interface Builder<T extends TaskObj>` and
```
Builder<T> from(TaskObj base);
...
T build();
```

tasks/service/impl/src/test/java/org/projectnessie/nessie/tasks/service/impl/TestTasksServiceImpl.java
```diff
@@ -0,0 +1,734 @@
...
+    CompletableFuture<BasicTaskObj.Builder> taskCompletionStage2 = new CompletableFuture<>();
+
+    BasicTaskRequest taskRequest1 = basicTaskRequest("hello", () -> taskCompletionStage1);
+    BasicTaskRequest taskRequest2 = basicTaskRequest("hello", () -> taskCompletionStage2);
+
+    // Want the same ObjId to verify that multiple repositories work fine
+    soft.assertThat(taskRequest1.objId()).isEqualTo(taskRequest2.objId());
+
+    // Submit task request for repo 1
+    CompletableFuture<BasicTaskObj> taskFuture1 = tasks1.submit(taskRequest1).toCompletableFuture();
```
dimas-b, line 84: why `.toCompletableFuture()`?

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,515 @@
...
+    //  c) add a `fetchObjAsync()`, but adding async variants to all database implementations will
+    //     be tricky
+    Obj obj = persist.getImmediate(taskRequest.objId());
+    if (obj != null) {
+      TaskObj taskObj = (TaskObj) obj;
```
dimas-b, line 101: How about `taskRequest.behavior().cast(obj)` (assuming my comment about generics works)?

adutra:
> I'm not actually sure about using CompletionStage as a scheduled task's handle. CompletionStage itself does not have a cancel() method.
> If we do CompletionStage.toCompletableFuture().cancel(), IMHO, the relationship of the cancel() call to the original CompletableFuture is very indirect and implementation specific to rely upon.
> I'd prefer to stick with a custom ScheduledHandle and a dedicated cancel() method in it.

I rather disagree, `.toCompletableFuture().cancel()` is common idiom when cancelling, all `CompletionStage` impls must comply with that.

Also, if we consider that `.toCompletableFuture().cancel()` is not reliable and decide to keep `ScheduledHandle` around, then we'd need to also return some handle from `call` and `supply`.

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/vertx/VertxTasksAsync.java
```diff
@@ -0,0 +1,105 @@
...
+  public CompletionStage<Void> call(Runnable runnable) {
+    return supply(
+        () -> {
+          runnable.run();
+          return null;
+        });
+  }
+
+  @Override
+  public <R> CompletionStage<R> supply(Supplier<R> supplier) {
```
adutra, line 49: As pointed offline, the current impl will not interrupt running tasks on cancellation. It's quite tricky to do that.

Taking inspiration from [this conversation](https://groups.google.com/g/vertx/c/byh_qr6a6_s/m/18sYs3E7BQAJ), I came up with the (rather convoluted) solution below:

```java
  @Override
  public <R> CompletionStage<R> supply(Supplier<R> supplier) {
    AtomicBoolean completed = new AtomicBoolean(false);
    AtomicReference<Thread> thread = new AtomicReference<>(null);
    CountDownLatch interrupted = new CountDownLatch(1);
    CompletionStage<R> stage =
        vertx
            .executeBlocking(
                () -> {
                  if (!thread.compareAndSet(null, Thread.currentThread())) {
                    return null;
                  }
                  try {
                    return supplier.get();
                  } catch (Throwable t) {
                    throw new CompletionException(t);
                  } finally {
                    try {
                      if (!completed.compareAndSet(false, true)) {
                        // this task has been cancelled, wait until the interruption attempt is done
                        try {
                          interrupted.await();
                        } catch (InterruptedException ignored) {
                        }
                      }
                    } finally {
                      thread.set(null);
                    }
                  }
                },
                false)
            .toCompletionStage();
    stage.whenComplete(
        (r, t) -> {
          if (t instanceof CancellationException) {
            if (completed.compareAndSet(false, true)) {
              Thread worker = thread.getAndSet(Thread.currentThread());
              if (worker != null) {
                try {
                  worker.interrupt();
                } catch (SecurityException ignored) {
                } finally {
                  interrupted.countDown();
                }
              }
            }
          }
        });
    return stage;
  }

  @Override
  public CompletionStage<Void> schedule(Runnable runnable, Instant scheduleNotBefore) {
    long realDelay = calculateDelay(clock, minimumDelayMillis, scheduleNotBefore);

    // Cannot use Vertx.timer(), because current Quarkus 3.6.5 has a Vertx version that does not
    // have Vertx.timer(). We can use this once Quarkus uses a Vertx version >= 4.5.1:
    //    Timer timer = vertx.timer(realDelay, TimeUnit.MILLISECONDS);
    //    CompletionStage<Void> stage = timer.toCompletionStage().thenRun(runnable);
    //    return new VertxScheduledHandle(timer, stage);

    CompletableFuture<Void> completableFuture = new CompletableFuture<>();
    long timerId =
        vertx.setTimer(
            realDelay,
            id ->
            // scheduling the runnable as a blocking task, a timer handler must not block
            {
              CompletionStage<Void> stage = call(runnable);
              completableFuture.whenComplete(
                  (v, t) -> {
                    if (t instanceof CancellationException) {
                      stage.toCompletableFuture().cancel(true);
                    }
                  });
            });

    completableFuture.whenComplete(
        (v, t) -> {
          if (t instanceof CancellationException) {
            vertx.cancelTimer(timerId);
          }
        });

    return completableFuture;
  }
```

With the above change, the `cancelRunningWithInterrupt` test also works for Vertx. 

But at this point, I'm wondering if we need the Vertx impl at all? Id doesn't seem suitable for long, blocking tasks.

snazy: Vert.X is still needed for the `TaskServiceImpl` when running in Quarkus.

snazy: How actual implementations run tasks via `TaskRequest.submitExecution()` is up to them - added notes to the javadoc.

snazy: I think that it's okay that tasks cannot be interrupted. `TasksServiceImpl` is fine with not interrupting tasks. Other implementations would need to choose the right implementation.

dimas-b:
> `.toCompletableFuture().cancel()` is common idiom when cancelling, all `CompletionStage` impls must comply with that.

That is true, but my point is not about `cancel()` itself, but about the propagation of the cancellation requests upstream. 

The contact (javadoc) of `cancel()` requires the cancellation of only _dependent_ `CompletableFuture`s... Upstream tasks do not have to be cancelled, as far as I understand. Furthermore, `toCompletableFuture()` does not have to return `this`, it _may_ return a downstream `CompletableFuture`, so in the end I do not think it is required that `cancel()` be applied to the `first` task in the `first.toCompletableFuture().cancel()` example. WDYT?

snazy:
> That is true, but my point is not about `cancel()` itself, but about the propagation of the cancellation requests upstream.
> 
> The contract (javadoc) of `cancel()` requires the cancellation of only _dependent_ `CompletableFuture`s... Upstream tasks do not have to be cancelled, as far as I understand. Furthermore, `toCompletableFuture()` does not have to return `this`, it _may_ return a downstream `CompletableFuture`, so in the end I do not think it is required that `cancel()` be applied to the `first` task in the `first.toCompletableFuture().cancel()` example. WDYT?

Yes, seems that upstream tasks don't get cancelled. But I think that is the correct behavior overall and everywhere. You can legibly cancel _any_ step, but that doesn't necessarily mean that _all_ steps need to be cancelled. Some chains react on cancellation of some steps as part of their "normal" control flow.

I think, doing a `.toCompletableFuture().cancel(false)` for the scheduled runnable is fine - also the behavior that only that step is cancelled.

tasks/service/async/src/test/java/org/projectnessie/nessie/tasks/async/pool/TestJavaPoolTasksAsync.java
snazy: Hehe - so `.toCompletableFuture().cancel(true)` doesn't work for `ScheduledThreadPoolExecutor` (other types probably as well). The CF however is properly cancelled.

And yea, the parameter is named **`may`**`InterruptIfRunning`

dimas-b:
> doing a .toCompletableFuture().cancel(false) for the scheduled runnable is fine - also the behavior that only that step is cancelled.

TBH, I do not see any guarantees in the javadoc that even "that" step will be cancelled. I understand that current JRE's return `this` from `.toCompletableFuture()`, but nothing requires that `this` be returned... it could be a downstream future :shrug:

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
dimas-b: If we want to call `.cancel()` on the returned object, let's return a `CompletableFuture` to avoid ambiguity.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,540 @@
...
+    try {
+      @SuppressWarnings("unchecked")
+      T taskObj = (T) clazz.cast(obj);
+      return taskObj;
+    } catch (ClassCastException e) {
+      throw new ClassCastException(
+          "Failed to cast obj of type "
+              + obj.type().name()
+              + " to the task request's expected type "
+              + clazz.getName());
```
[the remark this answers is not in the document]
snazy: Nope. There is none taking a cause?

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,540 @@
...
+    TaskState state = current.taskState();
+    if (state.status() == TaskStatus.RUNNING) {
+      TaskBehavior<TaskObj, TaskObj.Builder> behavior = params.taskRequest.behavior();
+      TaskObj updated =
+          withNewVersionToken(
+              behavior
+                  .newObjBuilder()
+                  .from(current)
+                  .taskState(behavior.runningTaskState(async.clock(), null)));
+      if (params.runningObj.compareAndSet(current, updated)) {
```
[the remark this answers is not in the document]
dimas-b: Would it be safer to linearise all local operations on `Persist` by chaining their respective tasks?

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,540 @@
...
+              // Return something (we're done with the execution / completion-stage)
+              return null;
+            });
+  }
+
+  private void scheduleTaskRunningUpdate(ExecParams params, TaskObj current) {
+    Instant scheduleNotBefore =
+        params.taskRequest.behavior().performRunningStateUpdateAt(async.clock(), current);
+    params.runningUpdateScheduled.set(
+        async.schedule(() -> updateRunningState(params), scheduleNotBefore));
```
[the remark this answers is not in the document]
snazy: Local races are actually unexpected - no need to prepare more than currently for that case.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,540 @@
...
+      if (params.runningObj.compareAndSet(current, updated)) {
+        try {
+          if (params.persist.updateConditional(current, updated)) {
+            metrics.taskRunningStateUpdated();
+            // Current state successfully updated in database, reschedule running task update
+            LOGGER.trace(
+                "{}: Successfully updated state for locally running task for {}", name, params);
+          } else {
+            metrics.taskRunningStateUpdateRace();
+            // Ran into a (remote) race, retry running-update
```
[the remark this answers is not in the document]
snazy: Refactored TaskServiceImpl quite a bit

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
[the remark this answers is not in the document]
snazy: Refactored TaskServiceImpl quite a bit

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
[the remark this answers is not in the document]
snazy: Refactored TaskServiceImpl quite a bit

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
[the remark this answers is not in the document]
snazy: That's what the Math.max() below does, right?

dimas-b (commented):
LGTM overall, with some internal API concerns in `TasksAsync`.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,606 @@
...
+    scheduleTaskRunningUpdate(params, obj);
+
+    params
+        .taskRequest
+        .submitExecution()
+        .handle(
+            (resultBuilder, failure) -> {
+              localTaskFinished(params, resultBuilder, failure);
+              // Return something (we're done with the execution / completion-stage)
+              return null;
```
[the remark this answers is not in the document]
snazy: > Maybe use .whenComplete() (on line 319) since we do not produce a new value?

done

tasks/service/async/src/main/java/org/projectnessie/nessie/tasks/async/TasksAsync.java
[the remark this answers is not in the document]
snazy: > `Instant notBefore = Instant.ofEpochMilli(Long.MIN_VALUE);`

come on - that timestamp doesn't make sense.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,606 @@
...
+  private void reattemptAfterRetryableError(ExecParams params, Instant retryNotBefore) {
+    async.schedule(() -> tryLocal(params), retryNotBefore);
+  }
+
+  private static final class ExecParams {
+    final Persist persist;
+    final CompletableFuture<TaskObj> resultFuture;
+    final TaskRequest<TaskObj, TaskObj.Builder> taskRequest;
+
+    final Lock lock = new ReentrantLock();
```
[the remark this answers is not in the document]
snazy: clean up for later

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -0,0 +1,606 @@
...
+    LOGGER.debug("{}: Starting local task execution for {}", name, params);
+    metrics.taskExecution();
+
+    params.runningObj = obj;
+    scheduleTaskRunningUpdate(params, obj);
+
+    params
+        .taskRequest
+        .submitExecution()
+        .handle(
```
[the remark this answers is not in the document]
snazy: I don't understand you comment. There is _nothing_ synchronous wrt to a handler.

snazy:
Added abstract `WrappingTasksAsync` + concrete `ThreadContextTasksAsync` for Eclipse microprofile context-propagation. This allows using a "plain" Java thread-pool in Quarkus w/ proper thread context propagation. "Just using" Vert.x does not propagate the thread context properly (runs into `WARN [io.qua.ope.run.QuarkusContextStorage] Context in storage not the expected context, Scope.close was not called correctly. Details: OTel context before:`) - this does not happen when using MP's `ThreadContext.contextualRunnable/Supplier`.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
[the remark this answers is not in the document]
dimas-b: By synchronous, I mean that is it called on the same thread that completes the future. I suggest to schedule it on another thread.

dimas-b (commented):
All changes LGTM, except that I still think that using `CompletionStage` as a handle for cancelling tasks scheduled by `TasksAsync` is not a sound approach, sorry.

snazy:
> All changes LGTM, except that I still think that using CompletionStage as a handle for cancelling tasks scheduled by TasksAsync is not a sound approach, sorry.

It's an internal API - we can change things if we need to. It's IMHO subjective, nothing technically founded.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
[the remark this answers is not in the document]
dimas-b: True, but the thread that actually produces the result (completes the future) may not expect to hit locks inside the handler code. AFAIK, the handler will be called on the thread that completes the future (by default).

This is a minor point.

dimas-b:
> It's an internal API - we can change things if we need to. It's IMHO subjective, nothing technically founded.

Technically, `CompletionStage.toCompletableFuture()` does not have to return `this`. If it returns a downstream future, calling `.cancel()` on it will not be related to the original scheduled task. 

It works now because  `CompletionStage.toCompletableFuture()` _happens_ to return `this`. So far this is factual, not subjective (unless I misread javadoc and JRE code).

I think this behaviour is very obscure to rely on. That is my opinion, indeed.

snazy:
> Technically, `CompletionStage.toCompletableFuture()` does not have to return `this`. If it returns a downstream future, calling `.cancel()` on it will not be related to the original scheduled task.
> 
> It works now because `CompletionStage.toCompletableFuture()` _happens_ to return `this`. So far this is factual, not subjective (unless I misread javadoc and JRE code).
> 
> I think this behaviour is very obscure to rely on. That is my opinion, indeed.

Still, the function say `toCF` (emphasizing `to`) - it is _not_ a _new_ stage - so everything done to the CF is done to the CS. It has nothing to do with "`this`". If things and assumptions turn out to be not true, it can be changed later - this is not a public library but a vehicle for coordinated tasks.

dimas-b:
From javadoc of `toCompletableFuture`: `If this stage is already a CompletableFuture, this method may return this stage itself.` So my reading that even if the stage itself is a CF, the method does not have to return `this`.

My point that there is not guarantee that `x.toCompletableFuture().cancel()` actually cancels `x`.

It is internal code, for sure, but why use something obscure even internally?

snazy:
> From javadoc of `toCompletableFuture`: `If this stage is already a CompletableFuture, this method may return this stage itself.` So my reading that even if the stage itself is a CF, the method does not have to return `this`.
> 
> My point that there is not guarantee that `x.toCompletableFuture().cancel()` actually cancels `x`.
> 
> It is internal code, for sure, but why use something obscure even internally?

Common idiom, as mentioned by @adutra [here](https://github.com/projectnessie/nessie/pull/7947#discussion_r1455966729) and [here](https://github.com/projectnessie/nessie/pull/7947#issuecomment-1898124153).

dimas-b (approved):

PR #8011, opened 2024-02-03
Use TLS in ITOAuth2Client

dimas-b:
Apparently Keycloak no longer allows administration (involving OIDC) over plain HTTP.


dimas-b:
Apparently Quarkus tests also need TLS now

snazy:
You sure this is not a problem on your machine, because I can all all the tests touched in this PR w/o issues against the `latest` + `23.0` keycloak images.

adutra:
I am also able to run the tests locally, both with JUnit and Gradle.

dimas-b:
Environment issues, indeed... sorry about the noise :facepalm:

snazy:
> Environment issues, indeed

What was it?

dimas-b:
> What was it?

I do not have a definitive root case yet... something related to `docker` bridges... works file with `podman`.

snazy:
>  works file with podman.

podman's nice - it's also rootless by default :)

PR #8043, opened 2024-02-12
Mark ContentResponse.getEffectiveReference as not null

adutra:
And also: document nullability of other similar methods.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/ContentResponse.java
+++ b/api/model/src/main/java/org/projectnessie/model/ContentResponse.java
@@ -40,10 +40,10 @@ public interface ContentResponse {
 
   /**
    * The effective reference (for example a branch or tag) including the commit ID from which the
-   * entries were fetched.
+   * entries were fetched. Never null.
    */
-  @Nullable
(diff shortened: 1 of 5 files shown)


snazy (approved):

api/model/src/main/java/org/projectnessie/model/ContentResponse.java
```diff
@@ -40,10 +40,10 @@ static ImmutableContentResponse.Builder builder() {
 
   /**
    * The effective reference (for example a branch or tag) including the commit ID from which the
-   * entries were fetched.
+   * entries were fetched. Never null.
    */
-  @Nullable
-  @jakarta.annotation.Nullable
+  @NotNull
```
dimas-b, line 45: I believe it is `null` in API v1 responses... is it not?

dimas-b: Ah, my mistake... this class is not used in API v1 :facepalm:

dimas-b (approved):

PR #8046, opened 2024-02-12
Separate production from test code in storage modules

adutra:


gradle/projects.main.properties
```diff
@@ -44,17 +44,23 @@ nessie-tasks-service-impl=tasks/service/impl
 nessie-versioned-spi=versioned/spi
 nessie-versioned-storage-batching=versioned/storage/batching
 nessie-versioned-storage-bigtable=versioned/storage/bigtable
+nessie-versioned-storage-bigtable-tests=versioned/storage/bigtable-tests
```
snazy, line 47: Those new modules would need to be added to the bom as well

snazy (commented):
The test classes don't need to be moved - feels a bit awkward to have the tests in another module. It's sufficient to add the new `*-tests` artifacts as a dependency in the main module for test-fixtures or intTest.

adutra:
> The test classes don't need to be moved - feels a bit awkward to have the tests in another module. It's sufficient to add the new `*-tests` artifacts as a dependency in the main module for test-fixtures or intTest.

I recall a similar discussion a while ago, where we concluded that it would be awkward as well to have:

```
main module[intTest] 
  \--> test module[main]
    \--> main module[main]
``` 

Don't you think it's cleaner to have test infra code + tests together in the test module?

snazy:
> > The test classes don't need to be moved - feels a bit awkward to have the tests in another module. It's sufficient to add the new `*-tests` artifacts as a dependency in the main module for test-fixtures or intTest.
> 
> I recall a similar discussion a while ago, where we concluded that it would be awkward as well to have:
> 
> ```
> main module[intTest] 
>   \--> test module[main]
>     \--> main module[main]
> ```
> 
> Don't you think it's cleaner to have test infra code + tests together in the test module?

The thing is that the _test classes_ get distributed to other modules as well, which isn't nice IMO. Supporting code however is okay.

BTW: we do exactly what you quoted (`module_a(intTest) -> other_module -> module_a(main)`) in other places.

versioned/storage/jdbc/src/main/resources/META-INF/services/org.projectnessie.versioned.storage.testextension.BackendTestFactory
```diff
@@ -1,18 +0,0 @@
-#
```
snazy, line 1: Why is this file deleted?

adutra: In fact it was moved to `versioned/storage/jdbc-tests/src/main/resources/META-INF/services/org.projectnessie.versioned.storage.testextension.BackendTestFactory` – but since the contents changed significantly (because of package renamings), git considers this as a removal + addition.

snazy: Ah, thanks for the clarification.

snazy (dismissed):
Big PR, but rather boring, so +1

snazy:
BTW: there's a conflicting change

PR #8047, opened 2024-02-12
Fix `invalid @Startup method` again

snazy:
Fixes the `Ignored an invalid @Startup method declared on ...` warning, especially visible during tests.

`@Startup` mustn't be used with `@Produces`, instead an event-observer for `StartupEvent` shall be used, taking the produced bean as an argument.


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/PersistProvider.java
```diff
@@ -98,10 +99,14 @@ public Persist produceWithInitializedRepository(@Default Persist persist) {
     return persist;
   }
 
+  public void eagerPersistInitialization(
+      @Observes StartupEvent event, @NotObserved Persist persist) {
+    LOGGER.debug("Eager initialization of persist implementation '{}'", persist.name());
```
dimas-b, line 104: Could you add a comment about using this to force `Persist` initialization?

snazy: done

adutra (approved):

PR #8055, opened 2024-02-13
Publish Docker images for the GC tool

### LINKED ISSUE #8054: [Feature]: Provide Docker image for GC tool
### Description

In Kubernetes clusters, the Nessie GC tool should typically be executed as a Kubernetes Job.

But that is hard to achieve for now, because there is no ready-to-use Docker image containing the GC tool.

The GC tool is already packaged as an executable, see [here](https://github.com/projectnessie/nessie/releases/download/nessie-0.76.6/nessie-gc-0.76.6) for example. Providing a Docker image that simply adds a layer with that executable should be easy to do.

### Expected Use Cases

_No response_

_No response_

adutra:
Fixes #8054


tools/dockerbuild/docker/Dockerfile-gctool
```diff
@@ -0,0 +1,14 @@
+FROM registry.access.redhat.com/ubi9/openjdk-21:1.17-2.1705482269
+
+LABEL org.opencontainers.image.source=https://github.com/projectnessie/nessie
+LABEL org.opencontainers.image.description="Projectnessie Nessie GC Tool"
```
dimas-b, line 4: nit: Maybe `ProjectNessie GC Tool`?

tools/dockerbuild/docker/Dockerfile-gctool
```diff
@@ -0,0 +1,14 @@
...
+LABEL org.opencontainers.image.licenses=Apache-2.0
+
+ENV LANGUAGE='en_US:en'
+
+# We make four distinct layers so if there are application changes the library layers can be re-used
```
dimas-b, line 9: Does this comment still apply here?

.github/workflows/release-publish.yml
```diff
@@ -134,12 +134,19 @@ jobs:
...
         # Add version to the openapi file name
         cp api/model/build/generated/openapi/META-INF/openapi/openapi.yaml api/model/build/nessie-openapi-${RELEASE_VERSION}.yaml
 
         cp gc/gc-tool/build/executable/nessie-gc gc/gc-tool/build/executable/nessie-gc-${RELEASE_VERSION}
+        tools/dockerbuild/build-push-images.sh \
+          -a "${ARTIFACTS}" \
+          -g ":nessie-gc-tool" \
+          -p "gc/gc-tool" \
+          -d "Dockerfile-gctool" \
+          ghcr.io/projectnessie/nessie-gc
```
adutra, line 149: @snazy This would end up as a new repo in Quay and DockerHub as well. Do we need to create those repos upfront? Do we need to change regsync settings?

dimas-b (dismissed):
LGTM :+1:

.github/workflows/release-publish.yml
snazy: The `regsync.yml` file needs to be changed, yes.

I _think_ we don't need to create repos upfront. Let's see whether the snapshot-publish job has something to complain - I guess not.

tools/dockerbuild/docker/Dockerfile-gctool
```diff
@@ -0,0 +1,13 @@
+FROM registry.access.redhat.com/ubi9/openjdk-21:1.17-2.1705482269
+
+LABEL org.opencontainers.image.source=https://github.com/projectnessie/nessie
+LABEL org.opencontainers.image.description="ProjectNessie GC Tool"
```
snazy, line 4:
```suggestion
LABEL org.opencontainers.image.description="Projectnessie GC Tool"
```

tools/dockerbuild/docker/Dockerfile-gctool
snazy: I think, lower-case `n` is correct here - don't recall a mention of the word _Projectnessie_ anywhere else - so hard to tell. But there was a two-word projectnessie used by a big online retailer - maybe better to avoid the already unlucky coincidence ;)

.github/docker-sync/regsync.yml
```diff
@@ -35,3 +36,24 @@ sync:
   - source: ghcr.io/projectnessie/nessie
     target: docker.io/projectnessie/nessie
     type: repository
+  # GC Tool
+  - source: ghcr.io/projectnessie/nessie-gc-unstable
+    target: quay.io/projectnessie/nessie-gc-unstable
+    type: repository
+    tags:
```
snazy, line 43: You can omit the tags here. We needed those for Nessie to not mirror super-old versions.
Since you're already touching this file - can you change the regex for Nessie server above to something like `0[.][789]\\d[.].*|[123][.].*` - to include everything like `0.7[0-9]....` and newer - and `1.xxx` as well.

.github/workflows/release-publish.yml
```diff
@@ -219,6 +226,8 @@ jobs:
         (\`chmod 744 nessie-gc-${RELEASE_VERSION}\` after download.)
         Can also be run using \`java -jar nessie-gc-${RELEASE_VERSION}\`, because it is actually a Java archive.
         Shell completion can be generated from the \`nessie-gc\` tool.
+        Nessie GC tool is also available as Docker image:
```
snazy, line 229:
```suggestion
        Nessie GC tool is also available as a Docker image: 
```

snazy (approved):
Just a nit left

snazy (approved):

PR #8058, opened 2024-02-14
Remove `iceberg-views` module for Iceberg 1.5+

snazy:


ajantha-bhat:
I was also about to raise a PR for this. I prepared by branch and saw that it is already there 🤦 . 
https://github.com/projectnessie/nessie/compare/main...ajantha-bhat:nessie:view

Only thing missing in this PR is to update the NOTICE file (to remove mention about Netflix code)

PR #8062, opened 2024-02-14
Version catalog: split versions for Quarkus/Platform and Quakus Gradle Plugin

snazy:
... to not get "premature" update-PRs for a yet not published Quarkus platform version.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -25,7 +25,8 @@ opentelemetryAlpha = "1.31.0-alpha"
 opentracing = "0.33.0"
 picocli = "4.7.5"
 protobuf = "3.25.2"
-quarkus = "3.7.3"
+quarkusPlatform = "3.7.3"
+quarkusPlugin = "3.7.3"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -26,6 +26,7 @@ opentracing = "0.33.0"
 picocli = "4.7.5"
 protobuf = "3.25.2"
 quarkus = "3.7.2"
+quarkusPlugin = "3.7.3"
```
adutra, line 29: Don't you think we should rather distinguish `quarkus` and `quarkusPlatform`? That's where we usually see discrepancies.

snazy: Those are all platform-deps. But I've renamed `quarkus` to `quarkusPlatform` - so it's `quarkusPlatform` and `quarkusPlugin` and clear that the one refers to the platform and the other only to the plugin.

adutra (dismissed):

adutra (approved):

PR #8081, opened 2024-02-18
GC Tool: ability to skip creating existing tables (IF NOT EXISTS)

adutra:
This commit introduces the ability to create the schema only if it does not exist. It also introduces the ability to optionally create the schema before executing the main command `gc`.

The rationale is that `gc` should typically be executed on Kubernetes as a `CronJob` but we don't want to special-case the first execution and do something different than the other executions, just because the schema doesn't exist yet.

With the new option `--jdbc-schema CREATE_IF_NOT_EXISTS`, the first job execution would find an empty schema and create the tables, while the subsequent executions would simply reuse the existing schema. Note that we don't handle schema evolution for now (could be done later).

There is also the option `--jdbc-schema DROP_AND_CREATE` if people prefer to upgrade an old schema by dropping it and recresating it afresh.

The option `--jdbc-schema` is present in all commands that embed `JdbcOptions`.

### DIFF
--- a/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
+++ b/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
@@ -18,6 +18,7 @@ package org.projectnessie.gc.contents.jdbc;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import org.intellij.lang.annotations.Language;
 
 final class SqlDmlDdl {
@@ -169,13 +170,12 @@ final class SqlDmlDdl {
           + "    FROM gc_live_set_contents \n"
           + "    WHERE live_set_id = ? AND content_id = ?";
 
-  static final List<String> ALL_CREATES =
-      Collections.unmodifiableList(
-          Arrays.asList(
-              CREATE_LIVE_SETS,
-              CREATE_LIVE_SET_CONTENTS,
-              CREATE_LIVE_SET_LOCATIONS,
-              CREATE_FILE_DELETIONS));
+  static final Map<String, String> ALL_CREATES =
+      Map.of(
+          "gc_live_sets", CREATE_LIVE_SETS,
+          "gc_live_set_contents", CREATE_LIVE_SET_CONTENTS,
+          "gc_live_set_content_locations", CREATE_LIVE_SET_LOCATIONS,
+          "gc_file_deletions", CREATE_FILE_DELETIONS);
 
   static final List<String> ALL_TABLE_NAMES =
       Collections.unmodifiableList(
--- a/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
+++ b/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
@@ -15,9 +15,7 @@
  */
 package org.projectnessie.gc.contents.jdbc;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
+import com.google.common.collect.ImmutableMap;
 import java.util.Map;
 import org.intellij.lang.annotations.Language;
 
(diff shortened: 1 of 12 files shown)


gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
```diff
@@ -169,13 +170,12 @@ private SqlDmlDdl() {}
...
-              CREATE_LIVE_SET_LOCATIONS,
-              CREATE_FILE_DELETIONS));
+  static final Map<String, String> ALL_CREATES =
+      Map.of(
+          "gc_live_sets", CREATE_LIVE_SETS,
+          "gc_live_set_contents", CREATE_LIVE_SET_CONTENTS,
+          "gc_live_set_content_locations", CREATE_LIVE_SET_LOCATIONS,
+          "gc_file_deletions", CREATE_FILE_DELETIONS);
 
   static final List<String> ALL_TABLE_NAMES =
```
snazy, line 180: Should this be initialized from `ALL_CREATES.keySet()`? Or removed at all (it's only used in tests IIRC)

gc/gc-repository-jdbc/src/testFixtures/java/org/projectnessie/gc/contents/jdbc/AbstractJdbcPersistenceSpi.java
```diff
@@ -58,18 +57,18 @@ protected PersistenceSpi createPersistenceSpi() {
   @BeforeEach
   void setup() throws Exception {
     try (Connection conn = dataSource.getConnection()) {
-      JdbcHelper.createTables(conn);
+      JdbcHelper.createTables(conn, false);
+      // Test table creation with existing tables
```
snazy, line 61: Should this actually be a test? Not sure whether we want to call `createTables` two times before each test case

gc/gc-repository-jdbc/src/testFixtures/java/org/projectnessie/gc/contents/jdbc/AbstractJdbcPersistenceSpi.java
```diff
@@ -58,18 +57,18 @@ protected PersistenceSpi createPersistenceSpi() {
...
   @AfterEach
   void tearDown() throws Exception {
     try (Connection conn = dataSource.getConnection()) {
-      try (Statement st = conn.createStatement()) {
-        for (String tableName : SqlDmlDdl.ALL_TABLE_NAMES) {
-          st.execute(String.format("DROP TABLE %s", tableName));
-        }
-      }
+      JdbcHelper.dropTables(conn);
+      // Test table deletion with non-existing tables
```
snazy, line 70: Same here (above)

gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/JdbcCreateSchema.java
```diff
@@ -39,9 +39,12 @@ public class JdbcCreateSchema extends BaseCommand {
   @Override
   protected Integer call(Closeables closeables) throws Exception {
     DataSource dataSource = closeables.maybeAdd(jdbc.createDataSource());
+    SchemaCreateStrategy schemaCreateStrategy = jdbc.getSchemaCreateStrategy();
+    if (schemaCreateStrategy == null) {
```
snazy, line 43: You can omit this null-check by specifying a default-value for the option. ANd tell picocli to always show the default value

gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/SqlDmlDdl.java
```diff
@@ -169,13 +170,12 @@ private SqlDmlDdl() {}
...
 
-  static final List<String> ALL_CREATES =
-      Collections.unmodifiableList(
-          Arrays.asList(
-              CREATE_LIVE_SETS,
-              CREATE_LIVE_SET_CONTENTS,
-              CREATE_LIVE_SET_LOCATIONS,
-              CREATE_FILE_DELETIONS));
+  static final Map<String, String> ALL_CREATES =
+      Map.of(
```
snazy, line 174: Should this maybe be a sorted map, so the output is rather deterministic for humans?

gc/gc-tool/src/test/java/org/projectnessie/gc/tool/TestCLI.java
```diff
@@ -77,6 +81,26 @@ static void setNessieUri(@NessieClientUri URI uri) {
     nessieUri = uri;
   }
 
+  private static DataSource dataSource;
```
snazy, line 84: Please add an `@AfterAll` and close the `dataSource` - the Agroal DSs are closeable.

gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/JdbcCreateSchema.java
adutra: I wanted the default value to be null in order to not introduce a behavioral change. Currently:

1. `create-schema` assumes `CREATE` if this parameter is null
2. `gc` assumes nothing if this parameter is null (no schema creation).

This is equivalent to the current behavior. Are we OK to change this?

snazy: Ah, okay - so that option is used from two commands.

snazy (approved):

PR #8089, opened 2024-02-20
Fix logging class in `ConfigChecks`

snazy:


dimas-b (approved):

XN137:
didnt we use to have `errorprone-slf4j` to catch this? :thinking:

PR #8113, opened 2024-02-22
Add test case for re-create table and merge

snazy:


dimas-b (approved):
I thought we already had this test, but I could not find it :sweat_smile:

PR #8131, opened 2024-02-28
JDBC backend: infer catalog and schema if not specified

adutra:


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusJdbcConfig.java
```diff
@@ -17,24 +17,19 @@
...
 @StaticInitSafe
 @ConfigMapping(prefix = "nessie.version.store.persist.jdbc")
 public interface QuarkusJdbcConfig extends JdbcBackendBaseConfig {
 
   @WithName("catalog")
-  @WithDefault("")
-  @WithConverter(RepoIdConverter.class)
   @Override
-  String catalog();
+  Optional<String> catalog();
```
adutra, line 30: I believe defaulting to the empty string is wrong. According to the javadocs of `java.sql.DatabaseMetaData#getTables`: 

> catalog – a catalog name; must match the catalog name as it is stored in the database; "" retrieves those without a catalog; null means that the catalog name should not be used to narrow the search

So, using the empty string means that we are going to look for tabels that do not belogn to a catalog, which seems wrong. Same for schema.

Still according to the javadocs, it's imo better to pass `null` if the catalog is not known, the risk being that a table in a different catalog could exist with the same name as a Nessie table. But in this case, the schema check would fail anyways.

snazy (approved):

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusJdbcConfig.java
adutra: Update: the PG driver does not seem to comply with the javadocs: looking at the implementation of `getTables`, `catalog` is not used at all, and `schema` is ignored when it's null _or empty_.

So the previous code was probably correct for PG, but maybe not correct for other drivers.

PR #8141, opened 2024-02-29
Docker image tool: implement local image builds

snazy:


dimas-b (dismissed):

dimas-b (approved):

tools/dockerbuild/build-push-images.sh
```diff
@@ -136,47 +152,58 @@ echo "Placing binaries in: ${ARTIFACTS}"
...
+  PLATFORMS="linux/amd64,linux/arm64/v8,linux/ppc64le,linux/s390x"
+  ${TOOL} buildx build \
+    --file "${BASE_DIR}/tools/dockerbuild/docker/${DOCKERFILE}" \
+    --platform "${PLATFORMS}" \
+    --tag "${IMAGE_NAME}:latest" \
+    --tag "${IMAGE_NAME}:latest-java" \
+    --tag "${IMAGE_NAME}:${IMAGE_TAG_BASE}" \
+    --tag "${IMAGE_NAME}:${IMAGE_TAG_BASE}-java" \
+    "${BASE_DIR}/${PROJECT_DIR}" \
+    --push \
```
adutra, line 198: Not a full review yet, but usually when I need to build a local image for use in minikube, I don't bother pushing to insecure registries.

Instead I remove this `--push` option and execute `eval $(minikube docker-env)` prior to building the image. This makes the build happen inside minikube, and the image is immediately available in the minikube node.

Have you tried this approach? Is it worth providing support for it as well?

adutra: Too late :-)

snazy: The difficulties begin when you use podman :(

Loads of issues with podman in minikube - that was the only approach that worked for me. (Yes, they document that you can use `podman-remote` and let that talk to minikube - but no - a huge API version mismatch... - just to name one annoyance)

PR #8154, opened 2024-03-04
HTTP client: add request and response filters programmatically

adutra:


api/client/src/main/java/org/projectnessie/client/http/NessieHttpClientBuilder.java
```diff
@@ -56,6 +56,20 @@ public interface NessieHttpClientBuilder extends NessieClientBuilder {
   @CanIgnoreReturnValue
   NessieHttpClientBuilder withResponseFactory(HttpResponseFactory responseFactory);
 
+  /**
+   * Add a request filter to the client. Enables low-level access to the request/response
+   * processing.
+   */
+  @CanIgnoreReturnValue
+  NessieHttpClientBuilder addRequestFilter(RequestFilter filter);
```
dimas-b, line 64: nit: `withRequestFilter` maybe?... just to follow existing naming patterns.

dimas-b (commented):
LGTM overall :+1:

api/client/src/main/java/org/projectnessie/client/http/NessieHttpClientBuilder.java
adutra: But you can add more than one filter.

dimas-b: hmmm.... good point :)

PR #8224, opened 2024-03-19
Add Apache HTTP client, choose HTTP client by name

snazy:
This change adds the Apache HTTP client (synchronous, HTTP/1.1 only for now).

The Apache HTTP client is not being used by default, and only works if it is available on the classpath:

HTTP clients become somewhat pluggable with this change as well.
A specific HTTP client can be chosen via the `nessie.client-builder-name` configuration option:
* `HTTP` (default) chooses the default HTTP client
* `JavaHttp` forces the new Java HTTP client (default when running on Java 11 or newer)
* `URLConnection` forces the URLConnection based client (default when running on Java 8)
* `ApacheHttp` forces the Apache HTTP client

The configuration option `nessie.force-url-connection-client` is deprecated now, in favor of `nessie.client-builder-name=URLConnection`.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
@@ -15,20 +15,24 @@
  */
 package org.projectnessie.client.http;
 
+import static java.util.Locale.ROOT;
+
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import java.net.URI;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
-import java.util.function.Function;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.Set;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLParameters;
 import org.projectnessie.client.NessieConfigConstants;
+import org.projectnessie.client.http.impl.HttpClientFactory;
 import org.projectnessie.client.http.impl.HttpRuntimeConfig;
 import org.projectnessie.client.http.impl.HttpUtils;
-import org.projectnessie.client.http.impl.jdk11.JavaHttpClient;
-import org.projectnessie.client.http.impl.jdk8.UrlConnectionClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,6 +59,7 @@ final class HttpClientBuilderImpl implements HttpClient.Builder {
   private boolean http2Upgrade;
   private String followRedirects;
   private boolean forceUrlConnectionClient;
+  private String httpClientName;
   private int clientSpec = 2;
 
   HttpClientBuilderImpl() {}
@@ -75,6 +80,7 @@ final class HttpClientBuilderImpl implements HttpClient.Builder {
     this.http2Upgrade = other.http2Upgrade;
     this.followRedirects = other.followRedirects;
     this.forceUrlConnectionClient = other.forceUrlConnectionClient;
+    this.httpClientName = other.httpClientName;
     this.clientSpec = other.clientSpec;
   }
 
@@ -153,6 +159,12 @@ final class HttpClientBuilderImpl implements HttpClient.Builder {
   @Override
   public HttpClient.Builder setForceUrlConnectionClient(boolean forceUrlConnectionClient) {
     this.forceUrlConnectionClient = forceUrlConnectionClient;
+    return this;
+  }
+
+  @Override
+  public HttpClient.Builder setHttpClientName(String httpClientName) {
+    this.httpClientName = httpClientName;
     return this;
   }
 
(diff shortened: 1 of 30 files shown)


adutra:
1. Why don't we [force the Java HTTP client to use HTTP/1](https://docs.oracle.com/en/java/javase/17/docs/api/java.net.http/java/net/http/HttpRequest.Builder.html#version(java.net.http.HttpClient.Version))?
2. If we really need to ditch the Java client, why not simply keep one single impl, URLConnection?

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -240,32 +252,56 @@ public HttpClient build() {
...
+      for (HttpClientFactory s : ServiceLoader.load(HttpClientFactory.class)) {
+        if (s.available()) {
+          if (preferred == null || preferred.priority() > s.priority()) {
+            preferred = s;
+          }
+          IMPLEMENTATIONS.put(s.name().toLowerCase(ROOT), s);
+        }
       }
-      FACTORY = factory;
+      PREFERRED = preferred;
```
adutra, line 304: Should we log a warning or throw if `preferred` is null?

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -240,32 +252,56 @@ public HttpClient build() {
             .addAllResponseFilters(responseFilters)
             .isHttp11Only(!http2Upgrade)
             .followRedirects(followRedirects)
-            .forceUrlConnectionClient(forceUrlConnectionClient)
             .authentication(authentication)
             .build();
 
-    return ImplSwitch.FACTORY.apply(config);
+    String clientName = httpClientName;
+    if ("HTTP".equalsIgnoreCase(clientName)) {
```
adutra, line 259: Is this special name really worth it? What's the added value compared to simply not providing any client name?

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
adutra: OK let's assume it will never happen since at least `URLConnection` will be available. Nevermind.

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
dimas-b: Why not just put the preferred impl into the `IMPLEMENTATIONS` map under the `HTTP` key?

snazy: `HTTP` refers to `org.projectnessie.client.http.NessieHttpClientBuilderImpl.name()`

dimas-b: true, but we resolve it to the "preferred" client, right? My point is: why have complex logic involving `null` for that when we can put the preferred impl. under `http` in the map?

snazy: Need the `null` & `forceUrlConnectionClient` handling anyway

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -240,32 +252,57 @@ public HttpClient build() {
...
+      clientName = null;
+    }
+    if (clientName == null && forceUrlConnectionClient) {
+      // forced use of URLConnectionClient
+      clientName = "URLConnection";
+    }
+    if (clientName != null
+        && forceUrlConnectionClient
+        && !"URLConnection".equalsIgnoreCase(clientName)) {
+      throw new IllegalArgumentException(
```
dimas-b, line 271: Is this validation worth the trouble? Why not let `forceUrlConnectionClient` with a WARN log message? The intent of that property is clear. On the other hand, breaking on misalignment with client name may complicate rollouts in case config is shared across multiple versions.

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -240,32 +252,57 @@ public HttpClient build() {
...
+    if ("HTTP".equalsIgnoreCase(clientName)) {
+      // fall back to default
+      clientName = null;
+    }
+    if (clientName == null && forceUrlConnectionClient) {
+      // forced use of URLConnectionClient
+      clientName = "URLConnection";
+    }
+    if (clientName != null
+        && forceUrlConnectionClient
```
dimas-b, line 269: nit: at this point `forceUrlConnectionClient` implies `clientName != null`

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
snazy: I don't understand. `HTTP` is the name of the `NessieHttpClientBuilderImpl` - it is NOT an explicitly documented discriminator - it has to be handled as if no client name has been set.

dimas-b: but, I think `HTTP` is a client discriminator that we supported before, right?

In any case, it can be provided by a user. Also `null` can be provided by the user. I'm suggesting to convert `null` to `HTTP` and hopefully simplify the following look up logic (basically avoid special cases and always get the impl. from the map). WDYT?

dimas-b (approved):

PR #8226, opened 2024-03-19
HTTP-Client: capture beginning of unparseable error response

snazy:


api/client/src/main/java/org/projectnessie/client/rest/ResponseCheckFilter.java
```diff
@@ -132,4 +139,44 @@ private static NessieError decodeErrorObject(
     }
     return error;
   }
+
+  /**
+   * Captures the first 2kB of the input in case the response is not parsable, to provide a better
```
adutra, line 144: Should we skip this if the media type is binary, e.g. `application/octet-stream`?

api/client/src/main/java/org/projectnessie/client/rest/ResponseCheckFilter.java
```diff
@@ -132,4 +139,44 @@ private static NessieError decodeErrorObject(
...
+   * error message to the user.
+   */
+  static final class CapturingInputStream extends InputStream {
```
adutra, line 147: Maybe extend `FilterInputStream`.

api/client/src/main/java/org/projectnessie/client/rest/ResponseCheckFilter.java
dimas-b: We can get garbage even if the content type is set to `text`... I guess the concern is about presenting that data to users... How about we exclude non-printable chars?

adutra: What about printing a hex dump? This person has a nice solution: https://gist.github.com/jen20/906db194bd97c14d91df

snazy: I doubt that is necessary. No HTTP server or gateway sends non-printable data for the kinds of errors we want to catch here.

adutra: I think @dimas-b is rather worried by giant HTML spaghettis in the response... I've seen those once or twice. Presenting the data as a hex dump or stg in the style would avoid messing up with the logs.

snazy: ... and become unreadable for humanes

snazy: b/c users report these kinds of errors when using tools

dimas-b: I agree that the risk of getting binary output here is very low. As for HTML, hex dumping will make it harder to read indeed... so I'm ok with including the text as a simple string.

api/client/src/main/java/org/projectnessie/client/rest/ResponseCheckFilter.java
```diff
@@ -132,4 +140,43 @@ private static NessieError decodeErrorObject(
...
+    public int read() throws IOException {
+      int rd = super.read();
+      if (rd >= 0 && captured < capture.length) {
+        capture[captured++] = (byte) rd;
+      }
+      return rd;
+    }
+
+    public String captured() {
+      return new String(capture, 0, captured, UTF_8);
```
dimas-b, line 179: Can we get encoding from the HTTP response?

snazy: Why? Unparseable characters are represented as `?` - so IMO that would be overkill (mapping character-set names) and become error-prone.

dimas-b: fair enough.

api/client/src/test/java/org/projectnessie/client/rest/TestResponseFilter.java
```diff
@@ -329,6 +334,37 @@ void testGood() {
...
+  void capturingInputStream(int blockSize, byte[] source, String expected) throws Exception {
+    CapturingInputStream capturing = new CapturingInputStream(new ByteArrayInputStream(source));
+    if (blockSize == 0) {
+      while (capturing.read() >= 0) {
+        // noop
+      }
+    } else {
+      byte[] tmp = new byte[blockSize * 2];
+      while (capturing.read(tmp, 8, blockSize) >= 0) {
+        // noop
```
dimas-b, line 348: Shouldn't we validate that we read the expected bytes too?

snazy: We do that in line 354, no?

dimas-b: No, on line 354 we only validate the we capture the truncated chunk correctly, but we do not validate that the normal read path returns all data without modifications.

snazy: Isn't that testing `FilterInputStream`?

snazy: Also, I suspect that it's working fine, since no other tests fail

PR #8268, opened 2024-03-28
Testing: Move `Bucket.createHeapStorageBucket()` to separate class, llow clearing it

snazy:


testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/HeapStorageBucket.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.ConcurrentModificationException;
+import java.util.TreeMap;
+import java.util.stream.Stream;
+
+public class HeapStorageBucket {
+  private final TreeMap<String, MockObject> objects = new TreeMap<>();
+
+  private HeapStorageBucket() {}
+
+  public static HeapStorageBucket newHeapStorageBucket() {
```
adutra, line 34: Not sure we need this static method + the `bucket()` method. I was rather expecting stg like:

```java
Bucket bucket = HeapStorageBucketFactory.createNew();
```

snazy: Need `HeapStorageBucket` here, because `Bucket` is a rather algorithmic approach, while `HeapStorageBucket` has state.

adutra (approved):

PR #8271, opened 2024-03-29
Add --recursive to BulkCommittingCommand

dimas-b:
A convenience to recursively delete a Namespace.


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
```diff
@@ -0,0 +1,111 @@
...
+          Entry entry = it.next();
+          if (ns.equals(entry.getName())) {
+            Preconditions.checkArgument(
+                NAMESPACE.equals(entry.getType()),
+                "Key %s does not point to a Namespace, it points to %s",
+                entry.getName(),
+                entry.getType());
+          }
+
+          commit.operation(Delete.of(entry.getName()));
```
adutra, line 87: Could it be that we have thousands of child namespaces? Shouldn't we proceed by batches?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
```diff
@@ -0,0 +1,111 @@
...
+import org.projectnessie.model.ContentKey;
+import org.projectnessie.model.EntriesResponse.Entry;
+import org.projectnessie.model.Operation.Delete;
+import org.projectnessie.model.Reference;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** Deletes a namespace recursively. */
+@Command(
+    name = "delete-recursive",
```
adutra, line 38: How about `delete-namespace`? Whether it's recursive or not could be made an option. In that case this command would resemble the `Delete` command but specialized for namespaces, and thus safer since it cannot delete tables or views by accident, and also cannot delete namespaces that contain tables or views.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
```diff
@@ -0,0 +1,111 @@
...
+    mixinStandardHelpOptions = true,
+    description = "Delete selected namespace and its nested objects (recursively)")
```
adutra, line 40:
```suggestion
    description = "Delete selected namespace and its child namespaces (recursively)")
```

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
```diff
@@ -0,0 +1,111 @@
...
+      int count = 0;
+      CommitMultipleOperationsBuilder commit = api.commitMultipleOperations();
+      try (Stream<Entry> entries = api.getEntries().reference(branch).prefixKey(ns).stream()) {
+        Iterator<Entry> it = entries.iterator();
+        while (it.hasNext()) {
+          Entry entry = it.next();
+          if (ns.equals(entry.getName())) {
+            Preconditions.checkArgument(
+                NAMESPACE.equals(entry.getType()),
+                "Key %s does not point to a Namespace, it points to %s",
```
adutra, line 82: A better message would be "Namespace X is not empty: found child object Y".

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
dimas-b: It is a possibility, but doing batches is kind of complicated because we'd have to sort deleted in the "deep first" manner.

I figured it might be an overkill for this "exploratory" tool. Let me see if it is too hard to implement, though.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
dimas-b: Hmmm, "nested objects" also includes tables :thinking:

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
dimas-b: Well, the check is for the CLI argument only. It is expected to be a namespace. Child entries are deleted automatically.

adutra: Oooh I completely misread this check, I thought it was preventing the accidental deletion of anything else than child namespaces inside the target namespace. Let me think again then; the tool is way more "dangerous" than I initially thought :-)

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
adutra: Nevermind, I thought that tables would never be deleted by this tool, which is why I thought that "nested objects" was rather deceiving.

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
dimas-b: I'm re-implementing this as a new option for `DeleteContent`... so far looks nicer overall :)

dimas-b: I've reworked the impl.... however, the amount of heap need is probably still the same.

At least now the new option is beneficial to other commands that process entries in bulk.

We can certainly limit the amount of in-memory keys at the expense of re-fetching them from the server. Do you think it's worth the effort?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/DeleteRecursive.java
dimas-b: re-implemented in a completely different way :)

adutra:
So this is looking A LOT better, thanks for looking into it.

I am just a little concerned by the possible permutations of options and which ones make sense and which ones don't. Also slightly worried by the ambiguous meaning of `--all`: it means all branches, but sometimes also all keys.

I think the decision tree should be stg like:

```
1. If --input is set, then process the branches and keys in the file;
2. If not, then 
    a. Select the branches to process:
        i. If --all is set, process all branches; [-all now means "all branches" only]
        ii. Else if --ref is set, process the given branch
        iii. Else, process default branch
    b. For each selected branch, select the keys to process:
        i. If --key is set, process only that key
            - If --nested is also set, also process child keys of --key
        ii. Else, process all keys
```

I gave a try this morning using arg groups, let me know what you think:

https://github.com/dimas-b/nessie/compare/recursive-delete-tool...adutra:nessie:recursive-delete-tool-adutra?expand=1

dimas-b:
> the decision tree should be stg like [...]

The arg groups ideal LGTM, but it's a rather big change set. Would you mind applying it after we merge this PR?

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/BulkCommittingCommand.java
```diff
@@ -50,6 +54,13 @@ public abstract class BulkCommittingCommand extends CommittingCommand {
           "Process all active keys in all branches (supersedes --input and all other selector options).")
   private boolean all;
 
+  @Option(
+      names = {"--nested"},
```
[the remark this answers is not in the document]
dimas-b: updated

tools/content-generator/src/intTest/java/org/projectnessie/tools/contentgenerator/ITDeleteContent.java
```diff
@@ -178,4 +181,75 @@ void deleteContentWithMessage() throws NessieNotFoundException {
       assertThat(commitMeta.getAuthor()).isEqualTo("Test Author 123 <test@example.com>");
     }
   }
+
+  @Test
+  void deleteNested() throws NessieNotFoundException, NessieConflictException {
```
adutra, line 186:
```suggestion
  void deleteRecursive() throws NessieNotFoundException, NessieConflictException {
```

adutra (approved):

PR #8288, opened 2024-04-03
Testing/obj-storage-mock: add dummy assume-role endpoint

snazy:


testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/StsResource.java
```diff
@@ -0,0 +1,67 @@
...
+  @Produces(MediaType.APPLICATION_XML)
+  public Object assumeRole(
+      @FormParam("Action") String action,
+      @FormParam("Version") String version,
+      @FormParam("RoleArn") String roleArn,
+      @FormParam("RoleSessionName") String roleSessionName,
+      @FormParam("Policy") String policy,
+      @FormParam("DurationSeconds") Integer durationSeconds,
+      @FormParam("ExternalId") String externalId,
+      @FormParam("SerialNumber") String serialNumber,
```
adutra, line 48: It seems `SerialNumber` is only for MFA, and also requires `TokenCode`. Not sure this is needed for testing STS.

testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/StsResource.java
```diff
@@ -0,0 +1,67 @@
...
+@Produces(MediaType.APPLICATION_XML)
+@Consumes(MediaType.APPLICATION_XML)
+public class StsResource {
+  @Inject ObjectStorageMock mockServer;
+
+  @Path("assumeRole")
+  @POST
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+  @Produces(MediaType.APPLICATION_XML)
+  public Object assumeRole(
```
adutra, line 40: I noticed we don't support `SourceIdentity`, is it worth adding it?

testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/StsResource.java
```diff
@@ -0,0 +1,67 @@
...
+import org.projectnessie.objectstoragemock.sts.ImmutableResponseMetadata;
+
+@Path("/sts/")
+@Produces(MediaType.APPLICATION_XML)
+@Consumes(MediaType.APPLICATION_XML)
+public class StsResource {
+  @Inject ObjectStorageMock mockServer;
+
+  @Path("assumeRole")
+  @POST
```
adutra, line 37: According to [this doc](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) the original STS endpoint uses GET + query parameters.

snazy: > According to [this doc](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) the original STS endpoint uses GET + query parameters.

The awssdk actually uses POST + form

testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/StsResource.java
snazy: This is testing code - we can change and adopt later at any time.

PR #8282, opened 2024-04-02
[NOOP] Update README.md to point to correct url

Highstead:
Add the file extension so the link in the readme works.

PR #8290, opened 2024-04-03
Commit after schema creation for GC tool

Highstead:
By default the connections from the JDBC datasource do not have autocommit on, so when we create schema the results of that creation are then rolled back.  This change is intentional about committing the schema changes.

### DIFF
--- a/gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/JdbcCreateSchema.java
+++ b/gc/gc-tool/src/main/java/org/projectnessie/gc/tool/cli/commands/JdbcCreateSchema.java
@@ -45,6 +45,8 @@ public class JdbcCreateSchema extends BaseCommand {
     }
     try (Connection conn = dataSource.getConnection()) {
       schemaCreateStrategy.apply(conn);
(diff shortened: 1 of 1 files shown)


Highstead:
I dont know if this is desirable or by design?  But to me it seems like if i want to create a schema explicitly in the command prompt it should be committed to the database.

adutra:
Hi @Highstead thanks for reporting this. What database are you using? I am a bit surprised because DDL statements are usually committed right away (regardless of JDBC's  auto-commit flag), and cannot be rolled back.

Highstead:
I'm running Google's cloudsql in postgresql 15.

dimas-b:
IIRC, PostreSQL, indeed, may put DDL in a normal Tx and not commit automatically.

adutra (approved):
TIL Postgres has [transactional DDL](https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis) :-)

The changes make totally sense, thanks for fixing this @Highstead!

adutra:
Per #8282 the CLA has been signed.

PR #8322, opened 2024-04-12
OAuth2 client config: "bark" with all check-errors

snazy:
The current validation of oauth2 options fails immediately for the first validation error. This can be more user friendly by throwing one exception with all violations, which is what this change does. It also replaces the "NPE-validation" for client-ID+secret from `OAuth2AuthenticatorConfig` to "IAE-validation" in `OAuth2ClientConfig`.

The change results in better validation error messages, which also include the config parameter key.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
@@ -246,7 +246,7 @@ abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig {
         "client secret must not be empty");
     check(
         violations,
-        CONF_NESSIE_OAUTH2_ISSUER_URL + "/" + CONF_NESSIE_OAUTH2_TOKEN_ENDPOINT,
+        CONF_NESSIE_OAUTH2_ISSUER_URL + " / " + CONF_NESSIE_OAUTH2_TOKEN_ENDPOINT,
         getIssuerUrl().isPresent() || getTokenEndpoint().isPresent(),
         "either issuer URL or token endpoint must be set");
     if (getTokenEndpoint().isPresent()) {
@@ -303,7 +303,7 @@ abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig {
     if (grantType == GrantType.AUTHORIZATION_CODE) {
       check(
           violations,
-          CONF_NESSIE_OAUTH2_ISSUER_URL + "/" + CONF_NESSIE_OAUTH2_AUTH_ENDPOINT,
+          CONF_NESSIE_OAUTH2_ISSUER_URL + " / " + CONF_NESSIE_OAUTH2_AUTH_ENDPOINT,
           getIssuerUrl().isPresent() || getAuthEndpoint().isPresent(),
           "either issuer URL or authorization endpoint must be set if grant type is '%s'",
           CONF_NESSIE_OAUTH2_GRANT_TYPE_AUTHORIZATION_CODE);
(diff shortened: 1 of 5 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -200,30 +219,64 @@ private static HttpClientException genericError(Status status) {
         "OAuth2 server replied with HTTP status code: " + status.getCode());
   }
 
+  private static void check(
+      List<String> violations, String paramKey, boolean cond, String msg, Object... args) {
+    if (!cond) {
+      if (args.length == 0) {
+        violations.add(msg + " (" + paramKey + ")");
```
dimas-b, line 226: nit: we could probably call `format()` with an empty arg list and avoid this `if`

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -200,30 +219,64 @@ private static HttpClientException genericError(Status status) {
...
+
+    check(
+        violations,
+        CONF_NESSIE_OAUTH2_CLIENT_ID,
+        !getClientId().isEmpty(),
+        "client ID must not be empty");
+    check(
+        violations,
+        CONF_NESSIE_OAUTH2_CLIENT_SECRET,
+        getClientSecret().length() > 0,
```
dimas-b, line 245: nit: why do we do `!getClientId().isEmpty()` above but check length here?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -234,61 +287,100 @@ void check() {
...
-    checkArgument(
+    check(
+        violations,
+        CONF_NESSIE_OAUTH2_PREEMPTIVE_TOKEN_REFRESH_IDLE_TIMEOUT,
         getPreemptiveTokenRefreshIdleTimeout().compareTo(getMinPreemptiveTokenRefreshIdleTimeout())
             >= 0,
         "preemptive token refresh idle timeout must be greater than or equal to %s",
         getMinPreemptiveTokenRefreshIdleTimeout());
+
+    // This check is rather useless, because the executor is created
```
dimas-b, line 372: not sure I understand this comment, sorry :thinking:

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2ClientConfig.java
```diff
@@ -52,85 +57,99 @@ class TestOAuth2ClientConfig {
...
+                + join(", ", expected));
   }
 
   static Stream<Arguments> testCheck() {
     return Stream.of(
         Arguments.of(
             OAuth2ClientConfig.builder().clientId("Alice").clientSecret("s3cr3t"),
-            new IllegalArgumentException("either issuer URL or token endpoint must be set")),
+            singletonList(
+                "either issuer URL or token endpoint must be set (nessie.authentication.oauth2.issuer-url/nessie.authentication.oauth2.token-endpoint)")),
```
dimas-b, line 73: nit: maybe add spaces around `/` for readability?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
snazy: it's `Secret`

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
snazy: When I started to write it, I realized that it's not good to have the executor in this type, because that's created before the config's validated.

dimas-b (approved):

PR #8323, opened 2024-04-12
Make `nessie.authentication.oauth2.client-secret` optional

snazy:
Not all configured clients in e.g. `Keycloak` require a client secret, especially those that are "publicly" accessible. This change makes the `nessie.authentication.oauth2.client-secret` optional.


dimas-b (commented):
:+1:

dimas-b (approved):

Issue #8316, opened 2024-04-10
Improve errors on attempts to reuse closed clients


dimas-b:
### Issue description

To follow up on https://github.com/apache/iceberg/issues/10110 , it would be nice for Nessie to provide nicer error messages if/when tables/views associated with a `NessieCatalog` are used after the catalog is closed.

dimas-b:
@pseudomuto: feel free to take a stab at this issue. Contributions are always welcome :)

PR #8325, opened 2024-04-12
OAuth2 client: make cancellation of device/auth-code flows possible

snazy:
Given the situation that a command line tool wants to authenticate using OAuth2 w/ a device or auth-code flow, the client waits until it gets the "authenticated" response/callback. If the user wants to cancel the authentication (think: hit ctrl-C), it's performed via a `j.l.InterruptedException`/`j.l.Thread.interrupt()`. However, interrupting the thread that performs authentication results in a bunch of exceptions being loggged uncondictionally to the console.

This change makes interrupting the authentication flow using `Thread.interrupt()` work w/o user "confusion".

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
@@ -32,7 +32,9 @@ import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 import org.projectnessie.client.http.HttpClient;
 import org.projectnessie.client.http.HttpClientException;
 import org.projectnessie.client.http.HttpRequest;
@@ -85,7 +87,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
     lastAccess = config.getClock().get();
     currentTokensStage = started.thenApplyAsync((v) -> fetchNewTokens(), this.executor);
     currentTokensStage
-        .whenComplete((tokens, error) -> log(error))
+        .whenComplete((tokens, error) -> maybeHandleError(error))
         .whenComplete((tokens, error) -> maybeScheduleTokensRenewal(tokens));
   }
 
@@ -103,7 +105,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
   Tokens getCurrentTokens() {
     try {
       return currentTokensStage.toCompletableFuture().get();
-    } catch (InterruptedException e) {
+    } catch (CancellationException | InterruptedException e) {
       Thread.currentThread().interrupt();
       throw new RuntimeException(e);
     } catch (ExecutionException e) {
@@ -209,12 +211,19 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
             // if that fails, of if tokens weren't available, try fetching brand-new tokens
             .exceptionally(error -> fetchNewTokens());
     currentTokensStage
-        .whenComplete((tokens, error) -> log(error))
+        .whenComplete((tokens, error) -> maybeHandleError(error))
         .whenComplete((tokens, error) -> maybeScheduleTokensRenewal(tokens));
   }
 
-  private void log(Throwable error) {
+  private void maybeHandleError(Throwable error) {
     if (error != null) {
+      if (hasCauseMatching(error, t -> t instanceof TimeoutException)) {
+        LOGGER.debug("Closing due to {}", error.toString());
+        // The auth-request already timed out, close this client to prevent future requests, which
+        // are unnecessary and also cause confusing logs (and in a user-facing client confusion).
+        close();
+        return;
+      }
       boolean tokensStageCancelled = error instanceof CancellationException && closing.get();
       if (tokensStageCancelled) {
         return;
@@ -226,6 +235,15 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
     } else {
       LOGGER.debug("Successfully renewed tokens");
     }
+  }
+
+  static boolean hasCauseMatching(Throwable t, Predicate<Throwable> test) {
+    for (; t != null; t = t.getCause()) {
+      if (test.test(t)) {
+        return true;
+      }
+    }
+    return false;
   }
 
   Tokens fetchNewTokens() {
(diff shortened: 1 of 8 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -209,12 +211,19 @@ private void renewTokens() {
...
-        .whenComplete((tokens, error) -> log(error))
+        .whenComplete((tokens, error) -> maybeHandleError(error))
         .whenComplete((tokens, error) -> maybeScheduleTokensRenewal(tokens));
   }
 
-  private void log(Throwable error) {
+  private void maybeHandleError(Throwable error) {
     if (error != null) {
+      if (hasCauseMatching(error, t -> t instanceof TimeoutException)) {
+        LOGGER.debug("Closing due to {}", error.toString());
```
dimas-b, line 221: Include `error` in the log message?

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
```diff
@@ -84,4 +91,13 @@ static void check(int clientApiVersion, HttpClient httpClient)
           clientApiVersion, minServerApiVersion, maxServerApiVersion, actualServerApiVersion);
     }
   }
+
+  static boolean hasCauseMatching(Throwable t, Predicate<Throwable> test) {
```
dimas-b, line 95: reuse <-> `OAuth2Client`?

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
```diff
@@ -65,9 +67,14 @@ static void check(int clientApiVersion, HttpClient httpClient)
           e.toString());
       return;
     } catch (Exception e) {
-      LOGGER.warn(
-          "API compatibility check: failed to contact config endpoint, proceeding without check",
-          e);
+      if (hasCauseMatching(
+          e, t -> t instanceof InterruptedException || t instanceof TimeoutException)) {
+        Thread.currentThread().interrupt();
```
dimas-b, line 72: We still failed to validate API compatibility in this case... Why skip the WARN log?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -65,6 +65,9 @@ final class JavaHttpClient implements org.projectnessie.client.http.HttpClient {
 
   @Override
   public HttpRequest newRequest(URI baseUri) {
+    if (client == null) {
+      throw new IllegalStateException("Client already closed");
```
dimas-b, line 69: Could you reference #8316 in the description?

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
snazy: yea

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
snazy: The compatibility check _cannot_ succeed in these cases - it's aborted. No need to confuse users.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
snazy: Not sure whether this and 8316 are the same?

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
dimas-b: How do we know the interrupt actually kills the client? It may be used it customized environments?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
snazy: meh - it'll already show you the exception class and message - guess that's enough

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
snazy: That's hypothetical, no?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
dimas-b: ok

api/client/src/main/java/org/projectnessie/client/http/NessieApiCompatibilityFilter.java
dimas-b: No. Interrupting a thread does not mean the client will not be reused.

dimas-b: I'd say, if we silently ignore the interrupt here, we ought to redo the API check on the next request (which currently will not happen).

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2TokenRefreshExecutor.java
```diff
@@ -28,8 +28,14 @@ final class OAuth2TokenRefreshExecutor extends ScheduledThreadPoolExecutor
 
   private static final Logger LOGGER = LoggerFactory.getLogger(OAuth2TokenRefreshExecutor.class);
 
+  // Need multiple threads:
+  // - One to run org.projectnessie.client.auth.oauth2.OAuth2Client.renewTokens
+  // - One to run org.projectnessie.client.auth.oauth2.DeviceCodeFlow.pollForNewTokens
+  // - (a spare one - just in case) ;)
+  private static final int CORE_POOL_SIZE = 3;
```
adutra, line 35: Heh, this resonates with my comment above. But I am not convinced we need that many threads, 1 is probably enough because the core size is 1, but the max size is unbounded. And as I said before, the thread that runs the device code flow is a short-lived one.

adutra: Also, not entirely related to this change, but I wonder if we need to keep `TokenRefreshThread`. The only 2 things it does is: give a name to the thread, and some logging before/after execution. 

First off, the thread name could be managed by the factory:

```java
Thread thread = new Thread(r, "nessie-client-oauth2-token-refresh");
```

Secondly, the thread name is fixed, which is not great.

And lastly, the logging is questionable imho (I think I did this for debugging purposes when developing this.)

snazy:
Closing in favor of #8347

PR #8333, opened 2024-04-15
Cleanup ResourceOwnerEmulator and AuthorizationCodeFlow

adutra:
Not yet a fix for #8331 but a few things that will hopefully make it easier to diagnose it:

* `ResourceOwnerEmulator`: added logs; enabled auto-flush in `consoleOut`; reworked `readConsole()` for improved readability.
* `AuthorizationCodeFlow`: cleanup + added a few logs.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
@@ -80,15 +80,15 @@ class AuthorizationCodeFlow implements AutoCloseable {
   private final CompletableFuture<Tokens> tokensFuture;
 
   /**
-   * A future that will complete when the close method is called. It is used only to avoid closing
-   * resources multiple times. Its completion stops the internal HTTP server.
+   * A future that will complete when the close() method is called. It is used merely to avoid
+   * closing resources multiple times. Its completion stops the internal HTTP server.
    */
   private final CompletableFuture<Void> closeFuture = new CompletableFuture<>();
 
   /**
-   * A phaser that will prevent the close() method from returning until all inflight requests have
-   * been processed. It is used to avoid closing the server prematurely and leaving the user's
-   * browser with an aborted HTTP request.
+   * A phaser that will delay closing the internal HTTP server until all inflight requests have been
+   * processed. It is used to avoid closing the server prematurely and leaving the user's browser
+   * with an aborted HTTP request.
    */
   private final Phaser inflightRequestsPhaser = new Phaser(1);
 
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
@@ -64,10 +64,32 @@ class AuthorizationCodeFlow implements AutoCloseable {
   private final URI authorizationUri;
   private final Duration flowTimeout;
 
+  /**
+   * A future that will complete when the redirect URI is called for the first time. It will then
+   * trigger the code extraction then the token fetching. Subsequent calls to the redirect URI will
+   * not trigger any action. Note that the response to the redirect URI will be delayed until the
+   * tokens are received.
+   */
   private final CompletableFuture<HttpExchange> redirectUriFuture = new CompletableFuture<>();
-  private final CompletableFuture<Void> closeFuture = new CompletableFuture<>();
+
+  /**
+   * A future that will complete when the tokens are received in exchange for the authorization
+   * code. Its completion will release all pending responses to the redirect URI. If the redirect
+   * URI is called again after the tokens are received, the response will be immediate.
+   */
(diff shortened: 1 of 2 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -64,9 +64,8 @@ class AuthorizationCodeFlow implements AutoCloseable {
   private final URI authorizationUri;
   private final Duration flowTimeout;
 
-  private final CompletableFuture<HttpExchange> requestFuture = new CompletableFuture<>();
+  private final CompletableFuture<HttpExchange> redirectUriFuture = new CompletableFuture<>();
   private final CompletableFuture<Void> closeFuture = new CompletableFuture<>();
-  private final CompletableFuture<String> authCodeFuture;
```
adutra, line 69: We don't need this intermediary future, it is probably a leftover from previous designs.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -138,7 +161,7 @@ public Tokens fetchNewTokens() {
     } catch (ExecutionException e) {
       abort();
       Throwable cause = e.getCause();
-      LOGGER.error("Authentication failed: " + cause.getMessage());
+      LOGGER.error("Authentication failed: {}", cause.getMessage());
```
dimas-b, line 164: nit: How about just `cause` for the log argument (`Exception.toString()`)... Since this is a generic handler we cannot be sure that the message alone is meaningful (e.g. it'll be `null` for `NullPointerException` in older JVMs).

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -149,15 +172,19 @@ public Tokens fetchNewTokens() {
...
         .whenComplete((v, error) -> exchange.close())
         .whenComplete((v, error) -> inflightRequestsPhaser.arriveAndDeregister());
   }
 
   private Void doResponse(HttpExchange exchange, Throwable error) {
-    LOGGER.debug("Authorization Code Flow: sending response");
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(
+          "Authorization Code Flow: sending response, error: {}",
+          error == null ? "none" : error.toString());
```
dimas-b, line 186: nit: why not just pass `error` to the logger? It should handle nulls.. Is it so much worse to have "null" in a DEBUG message instead of "none"?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -199,6 +227,16 @@ private Tokens fetchNewTokens(String code) {
     return tokens;
   }
 
+  private void log(Tokens tokens, Throwable error) {
+    if (LOGGER.isDebugEnabled()) {
+      if (error == null) {
+        LOGGER.debug("Authorization Code Flow: tokens received");
+      } else {
+        LOGGER.debug("Authorization Code Flow: error fetching tokens: {}", error.toString());
```
dimas-b, line 235: nit: a simple `error` arg would do.

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
adutra: Same, IntelliJ won't like it if it is used as a message argument. If it's not a message argument, then the whole stack trace will be printed, which is overkill.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
adutra: Same problem :-(

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
dimas-b: Ah, yes, I forgot about this IJ nuisance... nevermind.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
dimas-b: noted, nevermind

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
dimas-b: yep

dimas-b (approved):

PR #8337, opened 2024-04-16
OAuth2Client: improve sleep/close detection

adutra:
A few improvements on the sleep/close detection mechanisms:

1. Move the first `lastAccess` computation to `start()`, in case the client is instantiated long before being used for the first time;
2. Update `lastAccess` whenever `fetchNewTokens()` require user interaction (authorization and device code flows): prevents incorrect switch to sleep mode;
3. Check whether the client is closed in `wakeUp()` : no need to wake up; can happen if the client was sleeping, then is closed, then `authenticate()` is called.
4. Check whether the client is closed in `maybeScheduleTokensRenewal()`: no need to check and switch to sleep; avoids the case where the client is closed after some inactivity, then `maybeScheduleTokensRenewal` is called and incorrectly puts the client to sleep.
5. In `log()` simplify the check if log is needed: the token future can only be cancelled iif `closing` is true.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
@@ -82,7 +82,6 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
             .getExecutor()
             .orElseGet(
                 () -> new OAuth2TokenRefreshExecutor(config.getBackgroundThreadIdleTimeout()));
-    lastAccess = config.getClock().get();
     currentTokensStage = started.thenApplyAsync((v) -> fetchNewTokens(), this.executor);
     currentTokensStage
         .whenComplete((tokens, error) -> log(error))
@@ -128,6 +127,7 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
 
   @Override
   public void start() {
+    lastAccess = config.getClock().get();
     started.complete(null);
   }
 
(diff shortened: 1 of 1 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -158,6 +158,9 @@ public void close() {
   }
 
   private void wakeUp(Instant now) {
+    if (closing.get()) {
+      return;
```
dimas-b, line 162: nit: maybe log a debug message in this case?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -250,10 +255,14 @@ Tokens fetchNewTokens() {
       case AUTHORIZATION_CODE:
         try (AuthorizationCodeFlow flow = new AuthorizationCodeFlow(config)) {
           return flow.fetchNewTokens();
+        } finally {
+          lastAccess = config.getClock().get();
```
dimas-b, line 259: nit: Does it make sense to _also_ set `lastAccess` before we start `AuthorizationCodeFlow` in case the flow runs for a long time?

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: Let me merge this as is and I'll add the debug message in #8338.

PR #8338, opened 2024-04-16
OAuth2Client: refactor HTTP exchanges into separate flows

adutra:
Extracts all interactions with the authorization server out of `OAuth2Client` into separate classes implementing the new `Flow` interface. 

Simplifies `OAuth2Client` which now becomes merely a token refresh engine. Also makes it easier to implement more flows in the future.


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -191,6 +191,11 @@ HttpClient getHttpClient() {
         .build();
   }
 
+  void clearSecrets() {
```
adutra, line 194: A corollary of this change is that secrets will be kept in the config object until the client is closed. I think we can live with that.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/Secret.java
```diff
@@ -38,27 +31,11 @@ public int length() {
     return value.length;
   }
 
-  public char[] getCharsAndClear() {
```
adutra, line 41: This is *in theory* part of the public API, but I don't think anybody out there was calling this method or `getStringAndClear`. But I'm happy to put them back and mark them as deprecated instead.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/MustFetchNewTokensException.java
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client.auth.oauth2;
+
+class MustFetchNewTokensException extends RuntimeException {
+  public MustFetchNewTokensException(String message) {
```
dimas-b, line 19: Since the class is not `public` we should probably keep the constructor non-public too.

dimas-b (commented):
LGTM :+1: just some minor comments.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -141,7 +144,7 @@ private void abort() {
     redirectUriFuture.cancel(true);
   }
 
-  public Tokens fetchNewTokens() {
+  public Tokens fetchNewTokens(@Nullable Tokens ignored) {
```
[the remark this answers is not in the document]
adutra: > nit: `@Override`

👍 

> Should we assert that the `Tokens` parameter is in fact `null` for the sake of correctness?

Not sure, the `OAuth2Client` *could* invoke these flows with whatever tokens are currently available. That wouldn't be incorrect per se imho.

> Alternatively, maybe we could have two methods: `fetchNewTokens()` and `refreshTokens(Tokens)` with the base class throwing `UnsupportedOperationException` by default?

Again not sure, because only one of the two methods would work, the other would necessarily throw.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
snazy: However, that changes the semantics of an _immutable_ object.
WDYT of removing the ability to clear entirely, but make the accessors in `Secret` package-private?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/Secret.java
```diff
@@ -38,27 +31,11 @@ public int length() {
     return value.length;
```
snazy, line 31: Nit: refactor `length()` to `isNotEmpty()` - and make the function package-private

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
adutra: Actually `Secret` was introduced precisely to isolate all the mutable stuff in a class that is not `@Immutable`. 

Let me try your suggestion.

snazy (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
dimas-b: I tend to think that clearing these secrets is not very useful because the client is long-lived. I believe JRE crypto method generally clear passwords immediately after a single use, but in our case we use them many times.

+1 to reducing visibility of those methods.

PR #8347, opened 2024-04-17
Ability to cancel interactive client authentication

snazy:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -114,7 +115,10 @@ private Tokens getCurrentTokensIfAvailable() {
   }
 
   @Override
-  public void start() {
+  public void start(Consumer<Runnable> cancellationCallbackConsumer) {
+    if (cancellationCallbackConsumer != null) {
+      cancellationCallbackConsumer.accept(this::close);
```
snazy, line 120: This exposes the `close()` function to the call site building the client.

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -279,7 +289,22 @@ public HttpClient build() {
       throw new IllegalArgumentException(
           "No HTTP client factory for name '" + clientName + "' found");
     }
-    return httpClientFactory.buildClient(config);
+    HttpClient client = httpClientFactory.buildClient(config);
+
+    if (authentication != null) {
+      try {
+        authentication.start(cancellationCallbackConsumer);
```
snazy, line 296: Also moved the call to `.start()` here

adutra: Not that I'm against changing the `start()` method, but isn't this much simpler:

```
        authentication.start();
        cancellationCallbackConsumer.accept(authentication::close);
```

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -279,7 +289,25 @@ public HttpClient build() {
...
+        try {
+          client.close();
+        } catch (RuntimeException e2) {
+          e.addSuppressed(e2);
+        }
+        throw e;
+      }
+    }
+
+    return client;
```
adutra, line 310: Thinking a bit more about cancellation, I think you should close the `HttpClient` itself, not only the authentication:

```suggestion
    if (cancellationCallbackConsumer != null) {
      cancellationCallbackConsumer.accept(client::close);
    }
    return client;
```

Closing the HTTP client will close the authenticator.

Which actually begs the question: why not put the cancellation stuff at the `NessieApiV2` level? Closing the `NessieApiV2` instance will cascade-close everything underneath.

api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
```diff
@@ -171,6 +172,16 @@ default Set<String> names() {
   @CanIgnoreReturnValue
   NessieClientBuilder withSSLParameters(SSLParameters sslParameters);
 
+  /**
+   * Registers a callback to cancel an ongoing, blocking client setup.
+   *
+   * <p>When using "blocking" authentication, for example OAuth2 device or code flows, is being
+   * used, users may want to cancel an ongoing authentication. An application can register a
```
adutra, line 179:
```suggestion
   * <p>When using "blocking" authentication, for example OAuth2 device or code flows, users
   * may want to cancel an ongoing authentication. An application can register a
```

Also the description is slightly off: there is no "blocking client setup": the OAuth2 authenticator does not block when started (but it may block the first HTTP request, until the first token fetch succeeds).

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -279,7 +288,25 @@ public HttpClient build() {
...
-    return httpClientFactory.buildClient(config);
+    HttpClient client = httpClientFactory.buildClient(config);
+    if (cancellationCallbackConsumer != null) {
+      cancellationCallbackConsumer.accept(client::close);
+    }
+
+    if (authentication != null) {
+      try {
+        authentication.start();
+      } catch (RuntimeException e) {
```
adutra, line 299: Is this catch a hypothetical situation, or are you seeing actual errors here with your experiments?

Asking because `start()` currently cannot throw (and cannot block either): even with OAuth2, the only thing it does is that it completes the `started` future. The actual token fetching happens asynchronously. If an error occurs during token fetching, it will surface to the caller only when calling `OAuth2Authenticator.authenticate()` at the first HTTP request – not here.

api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
snazy: It blocks the first requests - so it is blocking

api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
snazy: It is only a safety net

api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
adutra: Sure, but client setup is never blocking.

adutra:
The more I think about this the less I'm convinced we need a cancellation callback. Closing the `NessieApiV2` should close all underlying resources, and interrupt all running threads if necessary. I don't see anything being "left" behind. Do you have a concrete example of situation where the Nessie client hangs and don't honor the SIGINT signal?

Looking at #8348 why don't you try something like this instead:

```java
try (NessieApiV2 client =
    NessieClientBuilder.createClientBuilderFromSystemSettings(
            NessieClientConfigSources.mapConfigSource(commandSpec.getParameters()))
        .withUri(commandSpec.getUri())
        .build(NessieApiV2.class)) {

  Terminal terminal = nessieCli.terminal();
  Terminal.SignalHandler sigIntHandler =
      terminal.handle(Terminal.Signal.INT, sig -> client.close());

  try {
    doSomethingWithTheClient(client);
  } catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    client.close();
  }
}
```

snazy:
> The more I think about this the less I'm convinced we need a cancellation callback. Closing the `NessieApiV2` should close all underlying resources, and interrupt all running threads if necessary. I don't see anything being "left" behind. Do you have a concrete example of situation where the Nessie client hangs and don't honor the SIGINT signal?

The OAuth authenticator is started in `HttpClientBuilderImpl.build()` - so the caller has no client to close, correct?
`OAuth2Client.start()` triggers `.fetchNewTokens()`, triggers the interaction with the user.

adutra:
OK, what about:

```java
List<AutoCloseable> closeables = new ArrayList<>();

Terminal terminal = nessieCli.terminal();
Terminal.SignalHandler sigIntHandler =
    terminal.handle(
        Terminal.Signal.INT,
        sig ->
            closeables.forEach(
                c -> {
                  try {
                    c.close();
                  } catch (Exception ignored) {
                  }
                }));

try (NessieAuthentication authentication =
    NessieAuthenticationProvider.fromConfig(commandSpec.getParameters())) {
  closeables.add(authentication);
  authentication.start();

  try (NessieApiV2 client =
      NessieClientBuilder.createClientBuilderFromSystemSettings(
              NessieClientConfigSources.mapConfigSource(commandSpec.getParameters()))
          .withUri(commandSpec.getUri())
          .withAuthentication(authentication)
          .build(NessieApiV2.class)) {
    closeables.add(client);

    try {
      doSomethingWithTheClient(client);
    } catch (InterruptedException e) {
      Thread.currentThread().interrupt();
      client.close();
    }
  }
}
```

adutra:
OK, I think I'm convinced now that we need a cancellation mechanism.

However I would like to propose an alternate design: the `Consumer` approach is a bit convoluted. A more common approach is to use a cancellation future that completes when the user cancels / interrupts his action:

```java
@CanIgnoreReturnValue
Builder setCancellationFuture(CompletionStage<?> cancellationFuture);
```

The builder can then "attach" any action(s) to the future:

```java
if (cancellationFuture != null) {
  cancellationFuture.thenRun(client::close);
}
```

And you experiment with jline would look like:

```java
CompletableFuture<Terminal.Signal> cancel = new CompletableFuture<>();
Terminal terminal = nessieCli.terminal();
terminal.handle(Terminal.Signal.INT, cancel::complete);

try (NessieApiV2 client =
    NessieClientBuilder.createClientBuilderFromSystemSettings(
            NessieClientConfigSources.mapConfigSource(commandSpec.getParameters()))
        .withUri(commandSpec.getUri())
        .withCancellationFuture(cancel)
        .build(NessieApiV2.class)) {
  closeables.add(client);

  try {
    doSomethingWithTheClient(client);
  } catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    client.close();
  }
}
```

snazy:
> ```java
> cancel::complete
> ```

Yea, that's fine. It avoids the need for a separate thread as well.

adutra (approved):

adutra (approved):

PR #8354, opened 2024-04-18
OAUth2Client: defer initial token fetch when user interaction is required

adutra:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -270,10 +278,12 @@ private void maybeWarn(String message, Throwable error) {
     boolean shouldWarn =
         lastWarn == null || Duration.between(lastWarn, now).compareTo(MIN_WARN_INTERVAL) > 0;
     if (shouldWarn) {
+      // defer logging until the client is used to avoid confusing log messages appearing
+      // before the client is actually used
       if (error instanceof HttpClientException) {
-        LOGGER.warn("{}: {}", message, error.toString());
```
dimas-b, line 274: Is it even (how?) possible to gent this log message before `used` is complete?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -270,10 +278,12 @@ private void maybeWarn(String message, Throwable error) {
...
+        used.thenRun(() -> LOGGER.warn("{}: {}", message, error.toString()));
```
dimas-b, line 284: This could accumulate futures... May be not much, but should we perhaps cancel `used` on close?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -135,6 +143,8 @@ public void close() {
         }
         // Always close the HTTP client (can't be shared).
         httpClient.close();
+        // Cancel this future to invalidate any pending log messages
+        used.cancel(true);
```
dimas-b, line 147: nit: Why not do it earlier (or in `finally`)... in case closing the HTTP client or executor throws?

dimas-b (approved):

PR #8361, opened 2024-04-19
Make bulk read timeout configurable in BigTablePersist

dimas-b:
Use the same timeout as for the Bigtable Data Client in Quarkus.

Tests use the old default of 5 sec.


versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -689,10 +689,13 @@ private <ID, R> void bulkFetch(
       }
     }
 
+    long timeoutMillis =
```
adutra, line 692: Maybe save to a field?

adutra (approved):

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
dimas-b: updated

PR #8368, opened 2024-04-21
Site: exlude old releases from search index, fix released-version title and page-dir

snazy:
The site search index currently returns results for _all_ Nessie versions, but should better only return results for the latest released version.

### DIFF
--- a/site/bin/_lib.sh
+++ b/site/bin/_lib.sh
@@ -131,9 +131,13 @@ create_latest () {
   rm -rf build/versions/latest/
   mkdir build/versions/latest/
 
-  # Create symbolic links and copy configuration files for the 'latest' documentation
-  ln -s "../${version}/docs" build/versions/latest/docs
+  # Copy docs + configuration files for the 'latest' documentation
+  cp -rL "build/versions/${version}/docs" build/versions/latest/
   cp "build/versions/${version}/mkdocs.yml" build/versions/latest/
+
+  # Exclude the latest-version-NUMBER directory from the search index
+  # (keeps the 'nessie-latest' in the search index)
+  search_exclude_versioned_docs "build/versions/${version}"
 
   cd build/versions/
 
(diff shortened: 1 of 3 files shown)


site/bin/_lib.sh
```diff
@@ -131,10 +131,14 @@ create_latest () {
   rm -rf build/versions/latest/
   mkdir build/versions/latest/
 
-  # Create symbolic links and copy configuration files for the 'latest' documentation
-  ln -s "../${version}/docs" build/versions/latest/docs
+  # Copy docs + configuration files for the 'latest' documentation
+  cp -rL "build/versions/${version}/docs" build/versions/latest/
```
snazy, line 135: Necessary to have only `nessie-latest` in the search index.

site/bin/_lib.sh
```diff
@@ -165,21 +169,33 @@ update_version () {
...
+  echo "     ... in ${version_docs_dir}"
 
-  cd "${version}/docs/"
+  cd "${version_docs_dir}"
 
   # Modify .md files to exclude versioned documentation from search indexing
-  python3 -c "import os
-for f in filter(lambda x: x.endswith('.md'), os.listdir()): lines = open(f).readlines(); open(f, 'w').writelines(lines[:2] + ['search:\n', '  exclude: true\n'] + lines[2:]);"
+  python3 -c "import glob
+for f in glob.glob('./**/*.md', recursive=True):
```
snazy, line 189: Prepraration, if we have in-dev-docs in directories.

site/bin/_lib.sh
```diff
@@ -165,21 +169,33 @@ update_version () {
...
+  # Generated docs are only included and become part of another page.
+  if not \"/generated-docs\" in f:
+    lines = open(f).readlines()
+    # Add an empty front-matter, if not present.
+    if lines[0] != '---\n':
```
snazy, line 194: Files that don't have an existing "front-matter"; would look odd otherwise.

site/bin/_lib.sh
```diff
@@ -258,14 +281,17 @@ release() {
   mv "${target}/docs/index-release.md" "${target}/docs/index.md"
 
   echo "     ... replace title in versioned mkdocs.yml"
-  sed -i'' -E "s/(^site_name:[[:space:]]+).*$/\1\"Nessie ${version}\"/" "${target}/mkdocs.yml"
+  sed -i'' -E "s/(^site_name:[[:space:]]+).*$/\1\"Nessie ${RELEASE_VERSION}\"/" "${target}/mkdocs.yml"
```
snazy, line 284: This is actually a fix

site/in-dev/index-release.md
```diff
@@ -1,3 +1,7 @@
+---
```
snazy, line 1: Technically not necessary, but probably good practice

site/bin/_lib.sh
adutra: This is the same command as the one in line 166. Don't we need to distinguish Darwin vx Linux here too?

adutra: Hmm I see; you are assuming this is always executed by Github CI as part of the release process.

PR #8372, opened 2024-04-21
OAuth2Client: fully support public clients

adutra:
In #8323 client secrets were made optional, but support for public clients wasn't fully implemented.

### DIFF
--- a/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
+++ b/api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
@@ -19,6 +19,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.InstanceOfAssertFactories.type;
 import static org.projectnessie.client.auth.oauth2.GrantType.AUTHORIZATION_CODE;
+import static org.projectnessie.client.auth.oauth2.GrantType.CLIENT_CREDENTIALS;
 import static org.projectnessie.client.auth.oauth2.GrantType.DEVICE_CODE;
 import static org.projectnessie.client.auth.oauth2.GrantType.PASSWORD;
 
@@ -134,7 +135,7 @@ public class ITOAuth2Client {
    * This test exercises the OAuth2 client "in real life", that is, with background token refresh
    * running.
    *
-   * <p>For 20 seconds, 2 OAuth2 clients will strive to keep the access tokens valid; in the
+   * <p>For 20 seconds, 3 OAuth2 clients will strive to keep the access tokens valid; in the
    * meantime, another HTTP client will attempt to validate the obtained tokens.
    *
    * <p>This should be enough to exercise the OAuth2 client's background refresh logic with all
@@ -150,7 +151,8 @@ public class ITOAuth2Client {
    */
   @Test
   void testOAuth2ClientWithBackgroundRefresh() throws Exception {
-    OAuth2ClientConfig config1 = clientConfig("Private1", true, false).build();
+    OAuth2ClientConfig config1 =
+        clientConfig("Private1", true, false).grantType(CLIENT_CREDENTIALS).build();
     OAuth2ClientConfig config2 = clientConfig("Private2", true, false).grantType(PASSWORD).build();
     OAuth2ClientConfig config3 =
         clientConfig("Public1", false, false).grantType(AUTHORIZATION_CODE).build();
@@ -190,23 +192,14 @@ public class ITOAuth2Client {
   }
 
   /**
-   * Dummy sentence to make Checkstyle happy.
-   *
-   * <p>This test exercises the OAuth2 client with the following setup:
-   *
-   * <ul>
-   *   <li>endpoint discovery on;
-   *   <li>client_credentials, password, authorization_code or device_code grant type for obtaining
-   *       the initial access token;
-   *   <li>refresh token sent on the initial response;
-   *   <li>refresh_token grant type for refreshing the access token.
-   * </ul>
+   * This test exercises the OAuth2 client with the following setup: endpoint discovery on; refresh
+   * token sent on the initial response; refresh_token grant type for refreshing the access token.
    */
   @ParameterizedTest
   @EnumSource(
       value = GrantType.class,
       names = {"CLIENT_CREDENTIALS", "PASSWORD", "AUTHORIZATION_CODE", "DEVICE_CODE"})
-  void testOAuth2RefreshTokenOn(GrantType initialGrantType) throws Exception {
+  void testOAuth2ConfidentialClientRefreshTokenOn(GrantType initialGrantType) throws Exception {
     OAuth2ClientConfig config =
         clientConfig("Private2", true, true).grantType(initialGrantType).build();
     try (OAuth2Client client = new OAuth2Client(config);
@@ -227,22 +220,14 @@ public class ITOAuth2Client {
   }
 
   /**
-   * Dummy sentence to make Checkstyle happy.
-   *
-   * <p>This test exercises the OAuth2 client with the following setup:
-   *
-   * <ul>
-   *   <li>client_credentials, password, authorization_code or device_code grant type for obtaining
-   *       the access token;
-   *   <li>no refresh token sent on the initial response;
-   *   <li>token exchange for obtaining the refresh token.
-   * </ul>
+   * This test exercises the OAuth2 client with the following setup: endpoint discovery off; no
+   * refresh token sent on the initial response; token exchange for obtaining the refresh token.
    */
   @ParameterizedTest
   @EnumSource(
       value = GrantType.class,
       names = {"CLIENT_CREDENTIALS", "PASSWORD", "AUTHORIZATION_CODE", "DEVICE_CODE"})
-  void testOAuth2RefreshTokenOff(GrantType initialGrantType) throws Exception {
+  void testOAuth2ConfidentialClientRefreshTokenOff(GrantType initialGrantType) throws Exception {
     OAuth2ClientConfig config =
         clientConfig("Private1", true, false).grantType(initialGrantType).build();
     try (OAuth2Client client = new OAuth2Client(config);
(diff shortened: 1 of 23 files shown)


api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -168,6 +172,7 @@ void testOAuth2ClientWithBackgroundRefresh() throws Exception {
               () -> {
                 tryUseAccessToken(validatingClient, client1.getCurrentTokens().getAccessToken());
                 tryUseAccessToken(validatingClient, client2.getCurrentTokens().getAccessToken());
+                tryUseAccessToken(validatingClient, client3.getCurrentTokens().getAccessToken());
```
adutra, line 175: This was an embarrassing "whoops!" moment :-)

api/client/src/main/java/org/projectnessie/client/auth/oauth2/PasswordFlow.java
```diff
@@ -22,9 +22,9 @@
  * href="https://datatracker.ietf.org/doc/html/rfc6749#section-4.3">Resource Owner Password
  * Credentials Grant</a> flow.
  */
-class ResourceOwnerPasswordFlow extends AbstractFlow {
+class PasswordFlow extends AbstractFlow {
 
-  ResourceOwnerPasswordFlow(OAuth2ClientConfig config) {
+  PasswordFlow(OAuth2ClientConfig config) {
```
adutra, line 27: Renamed, merely for consistency with other existing classes belonging to the `password` grant type.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/DeviceCodeFlow.java
```diff
@@ -130,12 +130,7 @@ public Tokens fetchNewTokens(@Nullable Tokens currentTokens) {
   }
 
   private DeviceCodeResponse requestDeviceCode() {
```
snazy, line 132: Why not just let all call sites invoke `invokeDeviceAuthEndpoint()` directly?

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -581,7 +581,7 @@ private void handleTokenEndpoint(HttpServletRequest req, HttpServletResponse res
         throws IOException {
       soft.assertThat(req.getMethod()).isEqualTo("POST");
       soft.assertThat(req.getContentType()).isEqualTo("application/x-www-form-urlencoded");
-      soft.assertThat(req.getHeader("Authorization")).isEqualTo("Basic QWxpY2U6czNjcjN0");
+      soft.assertThat(req.getHeader("Authorization")).isEqualTo("Basic Q2xpZW50MTpzM2NyM3Q=");
```
snazy, line 584: Is this change caused by the change to `configBuilder` below? (Just reviewing in a browser, not an IDE)

snazy (commented):
Just some initial comments, will review more tomorrow

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
adutra: `client3` was introduced a while ago, but it wasn't being tested because this line was missing.

api/client/src/intTest/java/org/projectnessie/client/auth/oauth2/ITOAuth2Client.java
```diff
@@ -146,17 +150,17 @@ static void setUpKeycloak() {
    */
   @Test
   void testOAuth2ClientWithBackgroundRefresh() throws Exception {
-    OAuth2ClientConfig config1 = clientConfig("Client1", false).build();
-    OAuth2ClientConfig config2 = clientConfig("Client2", false).grantType(PASSWORD).build();
+    OAuth2ClientConfig config1 = clientConfig("Private1", true, false).build();
```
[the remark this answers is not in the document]
adutra: It's client credentials by default, but I can make it explicit.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
adutra: See line 843 below. I changed `.clientId("Alice")` to `.clientId("Client1")` because the old client name was misleading. But that changed the Base 64 hash.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -596,6 +596,7 @@ private void handleTokenEndpoint(HttpServletRequest req, HttpServletResponse res
       Object response;
       int statusCode = 200;
       String grantType = data.get("grant_type");
+      data.remove("grant_type"); // determined by the target type
```
adutra, line 599: This is because I changed `@Value.Default` to `@Value.Derived` for the method `TokensRequestBase.getGrantType()`. Now the property is still serialized, but deserialization from Json cannot contain the field `grant_type` anymore.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeTokensRequest.java
```diff
@@ -39,10 +39,10 @@
 @Value.Immutable
 @JsonSerialize(as = ImmutableAuthorizationCodeTokensRequest.class)
 @JsonDeserialize(as = ImmutableAuthorizationCodeTokensRequest.class)
-interface AuthorizationCodeTokensRequest extends TokensRequestBase {
+interface AuthorizationCodeTokensRequest extends PublicTokensRequestBase {
 
   /** REQUIRED. Value MUST be set to "authorization_code". */
-  @Value.Default
+  @Value.Derived
```
adutra, line 45: This change is because I realized that this property was available in builders, but it shouldn't because the value is fixed. AFAICT there is no annotation to say "exclude from builder, but keep serializing/deserializing it".

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
snazy: You probably want something like this on the base class

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "grantType")
@JsonSubTypes({
  @JsonSubTypes.Type(value = AuthorizationCodeTokensRequest.class, name = GrantType.AUTHORIZATION_CODE),
  ...
})```

and something like this on the sub types:
```java
@JsonTypeName(GrantType.AUTHORIZATION_CODE)
```

and remove the `getGrantType()` functions.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AbstractFlow.java
```diff
@@ -27,15 +27,31 @@ abstract class AbstractFlow implements Flow {
     this.config = config;
   }
 
-  <T> T invokeEndpoint(URI endpoint, Object request, Class<? extends T> responseClass) {
+  <REQ extends TokensRequestBase, RESP extends TokensResponseBase> RESP invokeTokenEndpoint(
+      TokensRequestBase.Builder<REQ> request, Class<? extends RESP> responseClass) {
+    request.scope(config.getScope().orElse(null));
+    if (request instanceof PublicTokensRequestBase.Builder
+        && !config.getClientSecret().isPresent()) {
```
snazy, line 34: Why does it only include the client-ID, if the secret is not present? The comment in AuthorizationCodeFlow says it must always be present.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AbstractFlow.java
```diff
@@ -27,15 +27,31 @@ abstract class AbstractFlow implements Flow {
...
+      ((PublicTokensRequestBase.Builder<?>) request).clientId(config.getClientId());
+    }
+    return invokeEndpoint(config.getResolvedTokenEndpoint(), request.build(), responseClass);
+  }
+
+  DeviceCodeResponse invokeDeviceAuthEndpoint() {
+    DeviceCodeRequest.Builder request =
+        DeviceCodeRequest.builder().scope(config.getScope().orElse(null));
+    if (!config.getClientSecret().isPresent()) {
```
snazy, line 43: Same question here.

snazy (commented):
LGTM overall, but a question regarding the client-ID/secret requirements in this change.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
adutra: Oh of course! That worked, thanks!

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AbstractFlow.java
adutra: The general recommendation for all grants is:

* For confidential clients:
  -  Authenticate the client with a Basic authentication header
  - do NOT include `client_id` nor `client_secret` in POST request body
    - While it is also possible to authenticate the client using `client_id` nor `client_secret` in POST request body, this method is considered less secure. The `OAuth2Client` does not use it.
 * For public clients:
   - Do not include Basic authentication header (since there is no secret)
   - But include `client_id` in POST request body (but no `client_secret` of course), in order to identify the client.

In `AuthorizationCodeFlow`, when constructing the authorization URI, the OAuth2 spec mandates that the client ID be included. My comment there is however a bit superfluous, I will remove it.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AbstractFlow.java
adutra: Basically: include the client ID for public clients. Let me refactor this a bit to clarify.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AbstractFlow.java
snazy: Mind adding this to the javadoc? I'll likely forget soon ;)

snazy (commented):
Just an ask left to add the requirements for client-id as a comment in the code.

snazy (commented):
Thanks!

snazy (approved):

PR #8382, opened 2024-04-23
Helm charts: add support for AWS profiles

adutra:


helm/nessie/README.md
```diff
@@ -76,10 +76,11 @@ $ helm uninstall --namespace nessie-ns nessie
 | cassandra.contactPoints | string | `nil` |  |
 | cassandra.keyspace | string | `"nessie"` |  |
 | cassandra.localDatacenter | string | `nil` |  |
+| dynamodb.profile | string | `"default"` | The name of the profile that should be used, when loading AWS credentials from a profile file. Required only if no secret is provided below. |
```
snazy, line 79: Unrelated to this PR, but should we put this stuff on projectnessie.org?

helm/nessie/values.yaml
```diff
@@ -46,8 +46,12 @@ rocksdb:
 dynamodb:
   # -- The AWS region to use.
   region: us-west-2
+  # -- The name of the profile that should be used, when loading AWS credentials from a profile
+  # file. Required only if no secret is provided below.
+  profile: default
```
snazy, line 51: Shouldn't the default be `nil`, since it's optional?

adutra: I think that's fine, the profile won't be used anyways if the secret is present.

helm/nessie/README.md
adutra: Yes that would be an excellent idea. Currently the docs point users to the codebase:

* https://github.com/projectnessie/nessie/blob/3923cffd1103349457102f5ff458bb1190ed03d5/site/docs/guides/kubernetes.md#L32
* https://github.com/projectnessie/nessie/blob/8157b0b450094fd59492c8b29f2626e7706969f5/grafana/README.md#L5 (3 times in this file, 1 with a wrong anchor)

helm/nessie/values.yaml
snazy: Fair enough

snazy (approved):

dimas-b (approved):

helm/nessie/README.md
adutra: https://github.com/projectnessie/nessie/issues/8388

PR #8415, opened 2024-04-29
OAuthClient: consider zero as null for refresh_expires_in field

adutra:
This PR is an outcome of #8331 investigation: Keycloak sends "0" for `refresh_expires_in` when there is no refresh token.

See:

https://github.com/keycloak/keycloak/blob/ad32896725b05bc4c6d274395268c8eda6459037/services/src/main/java/org/keycloak/broker/oidc/AbstractOAuth2IdentityProvider.java#L231-L232

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/JacksonSerializers.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/JacksonSerializers.java
@@ -56,7 +56,9 @@ class JacksonSerializers {
     public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
       if (p.currentToken().isNumeric()) {
         int seconds = p.getValueAsInt();
-        return Instant.now().plusSeconds(seconds);
+        if (seconds != 0) {
+          return Instant.now().plusSeconds(seconds);
+        }
       }
       return null;
     }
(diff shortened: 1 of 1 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/JacksonSerializers.java
```diff
@@ -56,7 +56,9 @@ public SecondsToInstantDeserializer() {
     public Instant deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
       if (p.currentToken().isNumeric()) {
         int seconds = p.getValueAsInt();
-        return Instant.now().plusSeconds(seconds);
+        if (seconds != 0) {
+          return Instant.now().plusSeconds(seconds);
```
dimas-b, line 60: This change LGTM as it stands, but in general, would it be preferable to avoid binding to the current clock (`Instant.now()`) in JSON deserializers? Could we delegate that to the code that uses this value?

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/JacksonSerializers.java
adutra: Agreed. I will tackle that in a follow-up PR.

PR #8419, opened 2024-04-29
OAuthClient: don't call Instant.now() in serde layer

adutra:
Context: https://github.com/projectnessie/nessie/pull/8415#discussion_r1583212080


api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -891,4 +899,10 @@ private static AtomicReference<Runnable> mockTokensRefreshSchedule(
             });
     return task;
   }
+
+  private static TokensResponseBase asResponse(Tokens tokens) throws Exception {
+    Field enclosingInstance = tokens.getClass().getDeclaredField("this$0");
```
adutra, line 904: Admittedly a little brittle, but works.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/TokensResponseBase.java
```diff
@@ -33,31 +33,43 @@
...
-          .expirationTime(getRefreshTokenExpirationTime())
-          .build();
-    }
-    return null;
+    Duration accessExpiresIn = getAccessTokenExpiresIn();
+    AccessToken accessToken =
+        ImmutableAccessToken.builder()
+            .tokenType(getTokenType())
+            .payload(getAccessTokenPayload())
+            .expirationTime(accessExpiresIn == null ? null : clock.get().plus(accessExpiresIn))
```
snazy, line 46: Nit: Add an external variable `Instant now = clock.get()` at the beginning of `asTokens()` and use it here and below

api/client/src/main/java/org/projectnessie/client/auth/oauth2/TokensResponseBase.java
```diff
@@ -83,9 +95,9 @@ default RefreshToken getRefreshToken() {
   @Nullable
   @JsonProperty("expires_in")
   @JsonUnwrapped
-  @JsonSerialize(using = JacksonSerializers.InstantToSecondsSerializer.class)
-  @JsonDeserialize(using = JacksonSerializers.SecondsToInstantDeserializer.class)
-  Instant getAccessTokenExpirationTime();
+  @JsonSerialize(using = JacksonSerializers.DurationToSecondsSerializer.class)
+  @JsonDeserialize(using = JacksonSerializers.SecondsToDurationDeserializer.class)
+  Duration getAccessTokenExpiresIn();
```
snazy, line 100: I wonder whether the (de)serialization to a `Duration` type is really necessary. The OAuth & Keycloak specs say that this is an integer value representing 0 or or more seconds.
I suspect, just using `Integer` instead of `Duration` here and below is probably easier overall.

The function names could be extended with `...Seconds()` tho.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
snazy: Hacky - but since it's a test, that works for me.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/TokensResponseBase.java
snazy: Just using an `Integer` also saves a bunch of code.

adutra: Note that we lose the ability to use sub-second durations in unit tests – but all in all, I think that's fine.

snazy (approved):

PR #8414, opened 2024-04-29
OAuthClient: minor tweaks to the token exchange flow

adutra:
This PR comes from investigation of #8331: in some cases, Keycloak stops updating the expiration time of returned access tokens (maybe a bug). As a consequence, the token eventually expires, and the client recovers by re-fetching using the initial grant. However, if in that tiny window the expired token is used, that would result in an HTTP 401. This happens sometimes in `ITOAuthClient`. Here are some log from the tests that show this behavior:

```
// initial grant
13:33:55.434 - Fetching new tokens using CLIENT_CREDENTIALS
13:33:55.558 - Scheduling token refresh in PT4.995602S

// token exchanges
13:34:00.559 - Refreshing tokens using TOKEN_EXCHANGE
13:34:00.619 - Successfully renewed tokens, access exp: 2024-04-29T11:34:10.618557Z, refresh exp: null
13:34:00.619 - Scheduling token refresh in PT4.999249S
13:34:05.623 - Refreshing tokens using TOKEN_EXCHANGE
13:34:05.650 - Successfully renewed tokens, access exp: 2024-04-29T11:34:15.649901Z, refresh exp: null
13:34:05.650 - Scheduling token refresh in PT4.999807S
13:34:10.650 - Refreshing tokens using TOKEN_EXCHANGE

// returned access tokens have same expiration time as before => client starts "panicking"
13:34:10.673 - Successfully renewed tokens, access exp: 2024-04-29T11:34:15.673563Z, refresh exp: null
13:34:10.673 - Scheduling token refresh in PT1S
13:34:11.674 - Refreshing tokens using TOKEN_EXCHANGE
13:34:11.706 - Successfully renewed tokens, access exp: 2024-04-29T11:34:15.706808Z, refresh exp: null
13:34:11.707 - Scheduling token refresh in PT1S
13:34:12.711 - Refreshing tokens using TOKEN_EXCHANGE
13:34:12.733 - Successfully renewed tokens, access exp: 2024-04-29T11:34:15.733798Z, refresh exp: null
13:34:12.734 - Scheduling token refresh in PT1S
13:34:13.739 - Refreshing tokens using TOKEN_EXCHANGE
13:34:13.762 - Successfully renewed tokens, access exp: 2024-04-29T11:34:15.762344Z, refresh exp: null
13:34:13.763 - Scheduling token refresh in PT1S
13:34:14.764 - Refreshing tokens using TOKEN_EXCHANGE

// Returned token is near expiration
[...]

PR #8423, opened 2024-04-30
OAuthClient: fix AbstractFlow.isAboutToExpire

adutra:
Sorry, my bad: the fix for #8414 introduced a small regression. This PR addresses it.


snazy (commented):
What's the regression? Looks like only the config-get was moved?

adutra:
The previous version was always considering `config.getDefaultRefreshTokenLifespan()` as the default lifespan, but that depends on the token (access or refresh).

snazy (approved):

PR #8425, opened 2024-04-30
OAuth2Client: refactor ResourceOwnerEmulator

adutra:
This PR refactors `ResourceOwnerEmulator` into
separate classes per grant type, which not
only reduces the class sizes, but also allows
to use more than one instance simultaneously,
in the same test.


api/client/src/testFixtures/java/org/projectnessie/client/auth/oauth2/InteractiveResourceOwnerEmulator.java
```diff
@@ -0,0 +1,306 @@
...
+      Pattern.compile("<form.*action=\"([^\"]+)\".*>");
+
+  private static final AtomicInteger COUNTER = new AtomicInteger(1);
+
+  protected final String username;
+  protected final String password;
+
+  private final ExecutorService executor;
+
+  private final PipedOutputStream pipeOut = new PipedOutputStream();
```
snazy, line 69: Hint: these PipedIn/OutputStreams only "try" once per second. Not sure if that's a "problem".

snazy (approved):

api/client/src/testFixtures/java/org/projectnessie/client/auth/oauth2/InteractiveResourceOwnerEmulator.java
adutra: It's unfortunate, but not a problem imho. That's how they implemented those pipes :-(

That said, the "writer side" (the auth or device flow) always calls `flush()` after writing something, which in theory should `notifyAll()` and wake up the reader.

snazy: Yea - if they `flush()`, it's probably fine. I recall that I ran into these "1 second delays" in other situations in the past.

PR #8446, opened 2024-05-02
Doc-gen: traverse really all supertypes

snazy:
Under some not fully understood circumstances, `ConfigMappingInterface.getSuperTypes()` does not return all extended interfaces. This change additionaly traverse the super types retrieved via reflection as well, skipping all the types that have been visited via `getSuperTypes()`.


tools/doc-generator/doclet/src/main/java/org/projectnessie/nessie/docgen/SmallryeConfigs.java
```diff
@@ -112,12 +115,22 @@ public Void visitType(TypeElement e, Void ignore) {
 
             mappingInfo.processType(env, configMappingInterface, e);
 
-            Set<ConfigMappingInterface> seen = new HashSet<>();
+            if (className.contains("Adls")) {
```
adutra, line 118: Debug code?

snazy: oops

adutra (approved):

PR #8497, opened 2024-05-10
Add license reports and checks

snazy:
Dependencies, runtime dependencies including those that are distributed as part of a fat/uber jar, must use a "compatible" license. Checking those licenses manually is good, but some automatic checks are better.

This change adds a Gradle plugin to check licenses in the runtime configurations of the Nessie distributables (see below), and to produce license reports as HTML.

Nessie distributable are:
* Nessie Quarkus Server
* Nessie Quarkus CLI
* Nessie GC tool
* Nessie Content Generator
* Nessie Spark SQL extensions
* Nessie CLI/REPL

Checks and reports are run against the `runtimeClasspath` configurations of the above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,6 +103,11 @@ jobs:
       - name: Gradle / Assemble
         run: ./gradlew assemble --scan
 
+      - name: Gradle / License Check
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: checkLicense
+
       - name: Gradle / Publish to Maven local
         run: ./gradlew publishToMavenLocal --scan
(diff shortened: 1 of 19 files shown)


dimas-b (commented):
Code LGTM. Would you like in-depth reviews on `gradle/license/allowed-licenses.json` or is it somehow pre-validated?

snazy:
> Would you like in-depth reviews on gradle/license/allowed-licenses.json or is it somehow pre-validated?

A second pair of eyes looking into that would be good. It's been a while since (June last year) I originally built this change (forgot to open a PR).

dimas-b (approved):

PR #8498, opened 2024-05-10
Verify BSD+MIT+Go+UPL+ISC license mentions in `NOTICE` file + expose in Nessie server, GC-tool, Admin-Tool and CLI/REPL

snazy:
Dependencies that are not Apache licensed but MIT/BSD/Go/UPL/ISC licensed must include an entry in the `NOTICE` file. This change verifies that an "entry" for those dependencies is present.

This change also allows "retrieving" the `NOTICE` via:
* `/NOTICE.txt` from Nessie/Quarkus server
* via `HELP LICENSE` in the CLI/REPL
* via the new `show-licenses` command in the GC tool

Also:
* Adds a new "global" convenience `codeChecks` task, that includes license-checks, spotless-check and checkstyle.
* Adds MIT/BSD/ISC licenses from the Nessie web UI to the `NOTICE` file


build-logic/build.gradle.kts
```diff
@@ -33,3 +33,15 @@ dependencies {
 }
 
 tasks.withType<Test>().configureEach { useJUnitPlatform() }
+
+tasks.register("compileAll").configure {
```
snazy, line 37: Adding these two tasks here as well, as a convenience

build-logic/src/main/kotlin/nessie-common-src.gradle.kts
```diff
@@ -219,3 +222,17 @@ class MemoizedGitInfo {
     }
   }
 }
+
+afterEvaluate {
+  tasks.register("codeChecks").configure {
```
snazy, line 227: Convenience task to run "all the code checks"

servers/quarkus-server/build.gradle.kts
```diff
@@ -51,6 +51,7 @@ dependencies {
   implementation(project(":nessie-rest-common"))
   implementation(project(":nessie-rest-services"))
   implementation(project(":nessie-versioned-spi"))
+  implementation(project(":nessie-notice"))
```
snazy, line 54: That's all what's needed to make `curl http://127.0.0.1:19120/NOTICE.txt` work

build-logic/src/main/kotlin/NoticeReportValidation.kt
```diff
@@ -0,0 +1,84 @@
...
+    val missing = mutableMapOf<String, String>()
+
+    data.allDependencies.forEach { mod ->
+      val licenses =
+        (mod.manifests.map { it.license } +
+            mod.licenseFiles.flatMap { it.fileDetails }.map { it.license } +
+            mod.poms.flatMap { it.licenses }.map { it.name })
+          .distinct()
+
+      if (!licenses.any { needsNoNotice(it) } && licenses.any { needsNotice(it) }) {
```
adutra, line 54: I suppose this is meant for dual licensing, maybe add a comment explaining the rationale?

adutra (approved):

build-logic/src/main/kotlin/NoticeReportValidation.kt
snazy: Correct - it's for (most) dependencies that use more than one license.

PR #8503, opened 2024-05-11
Use `ubi9/openjdk-21-runtime` instead of `ubi9/openjdk-21` as the base image

snazy:
Supersedes #8502

### DIFF
--- a/tools/dockerbuild/docker/Dockerfile-server
+++ b/tools/dockerbuild/docker/Dockerfile-server
@@ -20,7 +20,7 @@
 # (See site/docs/try/configuration.md)
 #
 ###
-FROM registry.access.redhat.com/ubi9/openjdk-21:1.18-4
+FROM registry.access.redhat.com/ubi9/openjdk-21-runtime:1.18-4
 
 LABEL org.opencontainers.image.source=https://github.com/projectnessie/nessie
 LABEL org.opencontainers.image.description="Projectnessie Nessie Core Server"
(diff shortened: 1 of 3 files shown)


tools/dockerbuild/docker/Dockerfile-server
```diff
@@ -20,7 +20,7 @@
 # (See site/docs/try/configuration.md)
 #
 ###
-FROM registry.access.redhat.com/ubi9/openjdk-21:1.18-4
+FROM registry.access.redhat.com/ubi9/openjdk-21-runtime:1.18-4
```
adutra, line 23: After doing a `docker build & docker save` I observed that the total uncompressed image size went down from 682 MiB to 584 MiB. Nice!

I also validated that the `JAVA_OPTS_APPEND`,`JAVA_TOOL_OPTIONS` and `JAVA_DEBUG` env vars are still working.

adutra (approved):
LGTM – is this worth a mention in the changelog?

tools/dockerbuild/docker/Dockerfile-server
snazy: >  I observed that the total uncompressed image size went down from 682 MiB to 584 MiB. Nice!

Not so much on for amd64 - "just" down by 40MB - but still.

snazy:
> is this worth a mention in the changelog?

Don't think so - it's the same base image, just without the Java tools.

PR #8507, opened 2024-05-13
Publish docker images of the server admin tool

adutra:
...and fix a few things during release publish.


snazy (commented):
Also, the `regsync.yml` file needs to be updated

snazy (approved):

adutra:
BTW @snazy if I recall correctly you'd need to create the repository `ghcr.io/projectnessie/nessie-server-admin` before we can push to it, right?

PR #8512, opened 2024-05-13
Use StorageUri in ObjectIO

dimas-b:
Note: S3 locations recorded in Iceberg metadata files may not match the RFC URI syntax.

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsLocation.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsLocation.java
@@ -20,28 +20,29 @@ import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
 import java.util.Optional;
+import org.projectnessie.storage.uri.StorageUri;
 
 public final class AdlsLocation {
 
-  private final URI uri;
+  private final StorageUri uri;
   private final String storageAccount;
   private final String container;
   private final String path;
 
-  private AdlsLocation(URI uri, String storageAccount, String container, String path) {
+  private AdlsLocation(StorageUri uri, String storageAccount, String container, String path) {
     this.uri = uri;
     this.storageAccount = requireNonNull(storageAccount);
     this.container = container;
     this.path = requireNonNull(path);
   }
 
-  public static AdlsLocation adlsLocation(URI location) {
+  public static AdlsLocation adlsLocation(StorageUri location) {
     checkArgument(location != null, "Invalid location: null");
-    String scheme = location.getScheme();
+    String scheme = location.scheme();
     checkArgument(
         "abfs".equals(scheme) || "abfss".equals(scheme), "Invalid ADLS scheme: %s", location);
 
-    String authority = location.getAuthority();
+    String authority = location.requiredAuthority();
     String[] parts = authority.split("@", -1);
     String container;
     String storageAccount;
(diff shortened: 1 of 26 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
```diff
@@ -73,17 +73,17 @@ public InputStream readObject(URI uri) throws IOException {
...
+    checkArgument("s3".equals(uri.scheme()), "Invalid S3 scheme: %s", uri);
 
     return new ByteArrayOutputStream() {
       @Override
       public void close() throws IOException {
         super.close();
 
         S3Client s3client = s3clientSupplier.getClient(uri);
-        S3Uri s3uri = s3client.utilities().parseUri(uri);
+        S3Uri s3uri = s3client.utilities().parseUri(uri.asURI());
```
snazy, line 86: Should this be refactored to prevent using `java.net.URI` and run into the issue that this change intents to avoid?

tools/storage/uri/src/main/java/org/projectnessie/storage/uri/StorageUri.java
```diff
@@ -70,6 +71,14 @@ private StorageUri(String scheme, String authority, String path) {
     this.path = path;
   }
 
+  public URI asURI() {
```
snazy, line 74: I think, this could be removed with the other comments addressed?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
dimas-b: `uri.asURI()` produces the correct `URI` from a custom-parsed set of components (scheme, host, path, etc.).

The problem with S3 URI in Iceberg is that metadata contains location data that does not match the `URI` RFC. `StorageUri` handles that by parsing the location string directly.

dimas-b: However, we cannot pass that `URI.toString()` back to Iceberg (`S3FileIO`). That `URI` works only with `S3Client`.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsLocation.java
```diff
@@ -53,11 +54,11 @@ public static AdlsLocation adlsLocation(URI location) {
       storageAccount = authority;
     }
 
-    return new AdlsLocation(location, storageAccount, container, location.getPath());
+    return new AdlsLocation(location, storageAccount, container, location.path());
   }
 
   public URI getUri() {
```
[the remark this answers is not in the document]
dimas-b: Somehow I thought ADLS required `URI` in its tools, but I was mistaken... will fix.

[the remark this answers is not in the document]
dimas-b: fixed

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
dimas-b: Switched to using `StorageUri` components directly.... not sure now why I added the `asURI()` method initially :facepalm:

tools/storage/uri/src/main/java/org/projectnessie/storage/uri/StorageUri.java
dimas-b: removed

adutra (approved):

PR #8516, opened 2024-05-14
Force table / view location

adutra:


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -138,14 +145,34 @@ public Uni<Void> commitTransaction(
...
+                  IcebergCatalogOperation.builder()
+                      .updates(updates)
+                      .requirements(tableChange.requirements())
+                      .key(key)
+                      .type(ICEBERG_TABLE);
+
+              if (tableChange.hasAssertCreate()) {
+                List<IcebergUpdateRequirement> invalidRequirements =
+                    tableChange.requirements().stream()
+                        .filter(req -> !(req instanceof IcebergUpdateRequirement.AssertCreate))
```
dimas-b, line 165: nit: `SetLocation` (above) has a dedicated `import` statement.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -138,14 +145,34 @@ public Uni<Void> commitTransaction(
...
+                        .collect(Collectors.toList());
+                checkArgument(
+                    invalidRequirements.isEmpty(),
+                    "Invalid create requirements: %s",
+                    invalidRequirements);
+                String location = defaultTableLocation(decoded.warehouse(), key);
+                builder.addUpdate(setLocation(location));
```
dimas-b, line 172: What if `AssertCreate` is missing? Aren't we losing the `SetLocation` update in that case?

adutra: Yes we are. That's the idea: ignore any locations sent by the client, and use our own.

dimas-b: I mean should we not do `builder.addUpdate(setLocation(defaultLocation))` even if/when `AssertCreate` is not present?

adutra: I think not, because in this case it's an update, so the table already exists at its default location. Am I missing something?

dimas-b: To clarify: `AssertCreate` is a requirement that the client chooses to submit. Can we be sure that a table cannot be created without submitting an `AssertCreate`?

dimas-b: While it seems that it is not possible to create a table without an `AssertCreate` now, it seems to me that it's a co-incidence rather than a deliberate check. From the REST API perspective an `AssertCreate` does not have to be present. 

Should we set the location in `CatalogServiceImpl` perhaps rather than at the REST level?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1ResourceBase.java
```diff
@@ -259,53 +255,18 @@ static Branch checkBranch(Reference reference) {
...
-      } catch (NessieNamespaceNotFoundException e) {
-        // ignore
-      } catch (NessieNotFoundException e) {
-        // do nothing we want the same behavior that if the location is not defined
-      }
-
-      location = concatLocation(baseLocation, key.getName());
-    } else {
-      location = concatLocation(warehouse.location(), key.getName());
+      baseLocation = concatLocation(warehouse.location(), ns.toString());
```
dimas-b, line 267: `ns.toString()` will contain special chars (`\u0000`) if it has multiple elements.

dimas-b: Given https://github.com/apache/iceberg/issues/10279 I'd be very cautious with constructing storage paths from user-provided data. Nessie keys are pretty liberal and permit all sorts of special chars. I think we ought to weed them out either by replacing with a placeholder or encoding (unidirectional is ok, IMHO) with non-special chars.

dimas-b: Putting -1 on this just in case :)

dimas-b (changes requested):

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1ResourceBase.java
adutra: I noticed. I didn't change that though, the `ns.toString()` call was there already. Will look into an alternative.

dimas-b: Yeah, but we only recently discovered the Iceberg-side processing issues with special chars :facepalm: I think it would be wise to avoid any special chars in the locations we control at least.

dimas-b: I'd be fine with doing this in a follow-up PR too.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
adutra: Yes I was leaning towards that already, and you convinced me. I will revisit this PR entirely.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1ResourceBase.java
adutra: Indeed I think this will have to go in a separate PR because it will likely modify `org.projectnessie.model.Namespace` and add a `toSafePathString` method or stg in the style.

dimas-b: SGTM

adutra:
I tried to push the location computation down to `CatalogServiceImpl`, however many tests are now failing, and I still have no clue why. Will investigate more tomorrow.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -138,14 +145,33 @@ public Uni<Void> commitTransaction(
...
+                      .key(key)
+                      .warehouse(decoded.warehouse())
+                      .type(ICEBERG_TABLE);
+
+              if (tableChange.hasAssertCreate()) {
+                List<IcebergUpdateRequirement> invalidRequirements =
+                    tableChange.requirements().stream()
+                        .filter(req -> !(req instanceof AssertCreate))
+                        .collect(Collectors.toList());
+                checkArgument(
```
snazy, line 168: Don't we already do this implicitly via the `tableExists`/`viewExists` parameter in `IcebergUpdateRequirement`?

catalog/model/src/main/java/org/projectnessie/catalog/model/ops/CatalogOperation.java
```diff
@@ -31,4 +31,12 @@ public interface CatalogOperation {
...
 
   Content.Type getType();
+
+  /**
+   * The logical warehouse name where this operation will occur. Must correspond to a warehouse
+   * configured under {@code nessie.catalog.warehouses.<name>}.
+   *
+   * <p>If not set, the default warehouse will be used.
+   */
+  String warehouse();
```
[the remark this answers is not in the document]
adutra: That was my first impl, but in fact there is no need of that because of how `CatalogConfig.getWarehouse` is implemented: it already handles nulls gracefully. I should have added `@Nullable` here though, I will fix that.

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -755,7 +757,33 @@ public static long safeUnbox(Long value, long defaultValue) {
...
+    }
+    baseLocation = concatLocation(baseLocation, key.getName());
+    // Different tables with same table name can exist across references in Nessie.
+    // To avoid sharing same table path between two tables with same name, use uuid in the table
+    // path.
+    // TODO: support GZIP ??
+    // TODO: support TableProperties.WRITE_METADATA_LOCATION
+    String location =
+        String.format(
+            "%s_%s/metadata/00000-%s.metadata.json", baseLocation, randomUUID(), randomUUID());
```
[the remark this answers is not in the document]
adutra: Yes, I was mixing base location and metadata json location. It's fixed now.

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -755,7 +757,33 @@ public static long safeUnbox(Long value, long defaultValue) {
     return value != null ? value : defaultValue;
   }
 
-  public static NessieTableSnapshot newIcebergTableSnapshot(List<IcebergMetadataUpdate> updates) {
+  public static String defaultIcebergLocation(String warehouseLocation, ContentKey key) {
+    String baseLocation = warehouseLocation;
```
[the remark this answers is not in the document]
adutra: If you prefer to do this here, fine with me. You tell me :-)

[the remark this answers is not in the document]
dimas-b: Maybe flat structure (no namespace) in this PR, plus more advanced strategies as a follow-up?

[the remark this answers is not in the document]
adutra: What do you mean by "object store naming strategy", is it whether to use path-style or virtual-host style?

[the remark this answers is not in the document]
snazy: Nah - with "naming strategy" I meant the "randomized prefix" from `o.a.iLocationProviders.ObjectStoreLocationProvider` - but it seems my brain was somewhere else, because this change is about the base location - the the actual object location. Please disregard my comment and ignore everything I said or recommended above.

catalog/model/src/main/java/org/projectnessie/catalog/model/ops/CatalogOperation.java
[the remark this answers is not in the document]
snazy: WFM

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1ResourceBase.java
dimas-b: #8524

dimas-b (approved):

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -755,7 +756,36 @@ public static long safeUnbox(Long value, long defaultValue) {
...
 
-  public static NessieTableSnapshot newIcebergTableSnapshot(List<IcebergMetadataUpdate> updates) {
+  /** Returns the default table or view base location. */
+  public static String icebergBaseLocation(String warehouseLocation, ContentKey key) {
+    String baseLocation = warehouseLocation;
+    Namespace ns = key.getNamespace();
+    if (!ns.isEmpty()) {
+      baseLocation = concatLocation(warehouseLocation, ns.toString());
+    }
+    baseLocation = concatLocation(baseLocation, key.getName());
```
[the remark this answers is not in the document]
adutra: Sorry... what is a flat namespace?

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -755,7 +756,36 @@ public static long safeUnbox(Long value, long defaultValue) {
...
+    // Different tables with same table name can exist across references in Nessie.
+    // To avoid sharing same table path between two tables with same name, use uuid in the table
+    // path.
+    // Also: we deliberately ignore the TableProperties.WRITE_METADATA_LOCATION property here.
+    // TODO: support GZIP ??
```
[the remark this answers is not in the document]
adutra: I just moved the TODO form its old location to here. I can remove it.

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -755,7 +756,36 @@ public static long safeUnbox(Long value, long defaultValue) {
     return value != null ? value : defaultValue;
   }
 
-  public static NessieTableSnapshot newIcebergTableSnapshot(List<IcebergMetadataUpdate> updates) {
+  /** Returns the default table or view base location. */
+  public static String icebergBaseLocation(String warehouseLocation, ContentKey key) {
```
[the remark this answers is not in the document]
adutra: Would other table formats be able to use the same base location? It's probably safer for now to consider this Iceberg-specific, don't you think?

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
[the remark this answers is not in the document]
dimas-b: I meant no `ns` in the path, but I guess `key.getName()` can also contain special chars, so I'm ok leaving this method "as is" for now and addressing later, under #8524

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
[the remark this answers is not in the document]
dimas-b: Agreed. Let's keep it iceberg-specific until we support other formats.

dimas-b (approved):

dimas-b (approved):

PR #8523, opened 2024-05-16
API: Handle resource cleanup of HttpClient

ajantha-bhat:
Reported from Trino during a thread leak investigation: https://github.com/trinodb/trino/pull/21996

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
@@ -21,6 +21,8 @@ import java.net.http.HttpClient.Redirect;
 import java.net.http.HttpClient.Version;
 import java.time.Duration;
 import java.util.Locale;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
 import org.projectnessie.client.http.HttpRequest;
 import org.projectnessie.client.http.impl.HttpRuntimeConfig;
 
@@ -75,6 +77,26 @@ final class JavaHttpClient implements org.projectnessie.client.http.HttpClient {
 
   @Override
   public void close() {
+    if (client != null) {
+      client
+          .executor()
+          .ifPresent(
+              executor -> {
+                if (executor instanceof ExecutorService) {
+                  ExecutorService executorService = (ExecutorService) executor;
+                  executorService.shutdown();
+                  try {
+                    if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
+                      executorService.shutdownNow();
+                    }
+                  } catch (InterruptedException e) {
+                    executorService.shutdownNow();
+                    Thread.currentThread().interrupt();
+                  }
+                }
+              });
+    }
+
     client = null;
     config.close();
   }


api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -75,6 +76,15 @@ public URI getBaseUri() {
 
   @Override
   public void close() {
+    client
+        .executor()
+        .ifPresent(
+            executor -> {
+              if (executor instanceof ExecutorService) {
+                ((ExecutorService) executor).shutdown();
```
adutra, line 84: Shouldn't we do the traditional dance of `shutdown` + `awaitTermination` + `shutdownNow` ?

ajantha-bhat: added.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -75,6 +77,26 @@ public URI getBaseUri() {
...
+              executor -> {
+                if (executor instanceof ExecutorService) {
+                  ExecutorService executorService = (ExecutorService) executor;
+                  executorService.shutdown();
+                  try {
+                    if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
+                      executorService.shutdownNow();
+                    }
+                  } catch (InterruptedException e) {
+                    executorService.shutdownNow();
```
XN137, line 93: just FYI if this is supposed to mimic the "default impl" of guava's `MoreExecutors.shutdownAndAwaitTermination` (which we cant use in the client code i guess)
we are getting some details wrong...

also wondering if the 10 sec timeout should instead be related to our http client default timeout (if we have that, didnt check :sweat_smile:)

ajantha-bhat: a. Yes, I am aware of `MoreExecutors.shutdownAndAwaitTermination`, but this module doesn't depend on those lib. Hence, I didn't use it.

b. do you mean the order of executorService.shutdownNow(); and Thread.currentThread().interrupt();?

c. Used 10 seconds because of existing usages.
https://github.com/projectnessie/nessie/blob/627e9d73144b346c8ce5e8c16497a9a61bc0b8b5/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2TokenRefreshExecutor.java#L40-L50

ajantha-bhat: My implementation was based on oracle docs `shutdownAndAwaitTermination`

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ExecutorService.html

XN137: afaik both your link and guava do some more waiting after calling `shutdownNow` but the code in this PR does not

XN137: if we have multiple places in the client code we might want to have a single re-usable helper method potentially.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -75,6 +77,26 @@ public URI getBaseUri() {
 
   @Override
   public void close() {
+    if (client != null) {
+      client
+          .executor()
+          .ifPresent(
+              executor -> {
+                if (executor instanceof ExecutorService) {
```
dimas-b, line 85: This looks like we're digging into implementation details... How is `HttpClient` supposed to be closed normally? I'd guess its threads should eventually finish when not in use. WDYT?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -75,6 +77,26 @@ public URI getBaseUri() {
 
   @Override
   public void close() {
+    if (client != null) {
+      client
```
findepi, line 81: Why not just call `java.net.http.HttpClient.close()` or `java.net.http.HttpClient.shutdownNow()`?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
findepi: @dimas-b there seems to be couple options: `java.net.http.HttpClient.close()`, `java.net.http.HttpClient.shutdown()`, `java.net.http.HttpClient.shutdownNow()` (apparently copied from ExecutorService)

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
dimas-b: These methods do not exist on `java.net.http.HttpClient` as far as I can tell :)

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
dimas-b: Perhaps we could check if it implements `AutoClosable` (java 21), then cast and call `.close()`... WDYT?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
dimas-b: @findepi : What JREs do you use with Trino?

findepi: My fault. I am looking at Java 22.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
findepi: Yes, except that you probably want shutdownNow-like behavior over just shutdown. 
Java 21's `java.net.http.HttpClient.close()` will issue shutdown, then wait and then shutdownNow only if interrupted (same as ExecutorService).

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
```diff
@@ -75,6 +77,26 @@ public URI getBaseUri() {
...
+          .executor()
+          .ifPresent(
```
snazy, line 83: I wonder: is the executor actually present?

adutra: According to the javadocs, it's always present, a default one is created if none is provided.

snazy: Then the javadoc's wrong :P

adutra: Indeed it's confusing, the builder method states:

> Sets the executor to be used for asynchronous and dependent tasks. If this method is not invoked prior to building, a default executor is created for each newly built HttpClient.

And the getter method states:

> Even though this method may return an empty optional, the HttpClient may still have an non-exposed default executor that is used for executing asynchronous and dependent tasks.

🤷‍♂️

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
ajantha-bhat: Because the implementation of `HttpClient` doesn't extend the `Closable` or have `shutdownNow` interface.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
ajantha-bhat: I will check. 

@findepi: Could you please also share me the thread dump which proved the thread leak?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaHttpClient.java
ajantha-bhat: I have asked for the thread dump which caused Trino folks to report the issue. I will also take a thread dump and analyze tomorrow.

dimas-b: If we waited a bit, would the executor release its thread automatically? Could this be a matter of configuring the thread leak detection timeout?

snazy:
Also note that we have to have `org.projectnessie.client.http.impl.jdk11.JavaRequest#writerPool`, which uses a keep-alive time of 60 seconds - so these threads disappear after 60 seconds.

But anyway - we need the thread-dump and an analysis where these threads are created / from which pool.

If it's that FJP, then there's nothing we can do about that - it is due to the design of Java's new HTTP client.
Or they switch to another HTTP client for Nessie (Apache for example). All documented on our web site.

ajantha-bhat:
@findepi: I am closing this as I didn't find any thread leak in local test from `HttpClient` as `client.executor().isPresent()` is false for me. 

Let me know if you have a thread dump.

If it is FJP as mentioned in https://github.com/projectnessie/nessie/pull/8523#issuecomment-2115859062, I can change Trino side Nessie client initialization to use `ApacheHttpClient` which might help in avoiding confusion?

PR #8527, opened 2024-05-17
[Catalog] Improve warehouse lookup a little

snazy:
and add some unit tests

also remove unused `CatalogServerConfig`

### DIFF
--- a/servers/quarkus-common/src/test/java/org/projectnessie/quarkus/config/TestQuarkusCatalogConfig.java
+++ b/servers/quarkus-common/src/test/java/org/projectnessie/quarkus/config/TestQuarkusCatalogConfig.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


dimas-b (approved):

servers/quarkus-common/src/test/java/org/projectnessie/quarkus/config/TestQuarkusCatalogConfig.java
```diff
@@ -0,0 +1,137 @@
...
+        //
+        arguments(cfgWithDefault1, "w1", "w1"),
+        arguments(cfgWithDefault1, loc12, "w1"),
+        arguments(cfgWithDefault1, loc12 + "/", "w1"),
+        arguments(cfgWithDefault1, loc12, "w1"),
+        arguments(cfgWithDefault1, loc12 + "/", "w1"),
+        arguments(cfgWithDefault1, "w3", "w3"),
+        arguments(cfgWithDefault1, loc3, "w3"),
+        arguments(cfgWithDefault1, loc3 + "/", "w3"),
+        arguments(cfgWithDefault1, "fooooo", "w2"),
```
adutra, line 123: Shouldn't we throw instead? It feels weird to return the default warehouse when the user explicitly asked for some other warehouse (that doesn't exist).

PR #8533, opened 2024-05-18
Persistence: properly handle timeout-ish exceptions

snazy:
Databases can run into timeout situations, which should not happen, but do happen.

This change introduces a new `UnknownOperationResultException`, which signals that the database operation did not return a meaningful result and the outcome of the operation is unknown. Within the Nessie commit-retry loop this exception is handled via a "regular" commit-retry.

Included are also related changes to the database specific implementations.

### DIFF
--- a/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
+++ b/versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
@@ -39,7 +39,11 @@ import static org.projectnessie.versioned.storage.serialize.ProtoSerialization.s
 
 import com.google.api.core.ApiFuture;
 import com.google.api.gax.batching.Batcher;
+import com.google.api.gax.rpc.AbortedException;
 import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.DeadlineExceededException;
+import com.google.api.gax.rpc.UnknownException;
+import com.google.api.gax.rpc.WatchdogTimeoutException;
 import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
 import com.google.cloud.bigtable.data.v2.models.Filters;
 import com.google.cloud.bigtable.data.v2.models.Filters.Filter;
@@ -72,6 +76,7 @@ import org.projectnessie.versioned.storage.common.exceptions.ObjTooLargeExceptio
 import org.projectnessie.versioned.storage.common.exceptions.RefAlreadyExistsException;
 import org.projectnessie.versioned.storage.common.exceptions.RefConditionFailedException;
 import org.projectnessie.versioned.storage.common.exceptions.RefNotFoundException;
+import org.projectnessie.versioned.storage.common.exceptions.UnknownOperationResultException;
 import org.projectnessie.versioned.storage.common.persist.CloseableIterator;
 import org.projectnessie.versioned.storage.common.persist.Obj;
 import org.projectnessie.versioned.storage.common.persist.ObjId;
@@ -97,6 +102,12 @@ public class BigTablePersist implements Persist {
   }
 
   static RuntimeException apiException(ApiException e) {
+    if (e instanceof DeadlineExceededException
+        || e instanceof WatchdogTimeoutException
+        || e instanceof UnknownException
+        || e instanceof AbortedException) {
+      throw new UnknownOperationResultException("Unhandled BigTable exception", e);
+    }
     throw new RuntimeException("Unhandled BigTable exception", e);
   }
 
--- a/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
+++ b/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
@@ -49,7 +49,6 @@ import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.DriverException;
 import com.datastax.oss.driver.api.core.DriverTimeoutException;
-import com.datastax.oss.driver.api.core.NodeUnavailableException;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.BoundStatementBuilder;
@@ -61,7 +60,7 @@ import com.datastax.oss.driver.api.core.metadata.Metadata;
 import com.datastax.oss.driver.api.core.metadata.schema.ColumnMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.KeyspaceMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
-import com.datastax.oss.driver.api.core.servererrors.QueryExecutionException;
+import com.datastax.oss.driver.api.core.servererrors.QueryConsistencyException;
 import jakarta.annotation.Nonnull;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
@@ -345,15 +344,23 @@ public final class CassandraBackend implements Backend {
   }
 
   static RuntimeException unhandledException(DriverException e) {
-    if (e instanceof QueryExecutionException) {
+    if (isUnknownOperationResult(e)) {
       return new UnknownOperationResultException(e);
-    } else if (e instanceof DriverTimeoutException
-        || e instanceof NodeUnavailableException
-        || e instanceof AllNodesFailedException) {
-      return new UnknownOperationResultException(e);
-    } else {
-      return e;
+    } else if (e instanceof AllNodesFailedException) {
+      AllNodesFailedException all = (AllNodesFailedException) e;
+      if (all.getAllErrors().values().stream()
+          .flatMap(List::stream)
+          .filter(DriverException.class::isInstance)
+          .map(DriverException.class::cast)
+          .anyMatch(CassandraBackend::isUnknownOperationResult)) {
+        return new UnknownOperationResultException(e);
+      }
     }
+    return e;
+  }
+
+  private static boolean isUnknownOperationResult(DriverException e) {
+    return e instanceof QueryConsistencyException || e instanceof DriverTimeoutException;
   }
 
   @Override
--- a/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
+++ b/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
@@ -328,8 +328,7 @@ public final class CassandraBackend implements Backend {
       ResultSet rs = session.execute(stmt);
       return rs.wasApplied();
     } catch (DriverException e) {
-      handleDriverException(e);
-      return false;
+      throw unhandledException(e);
     }
   }
 
@@ -354,18 +353,6 @@ public final class CassandraBackend implements Backend {
       return new UnknownOperationResultException(e);
     } else {
       return e;
-    }
-  }
-
-  static void handleDriverException(DriverException e) {
-    if (e instanceof QueryExecutionException) {
-      throw new UnknownOperationResultException(e);
-    } else if (e instanceof DriverTimeoutException
-        || e instanceof NodeUnavailableException
-        || e instanceof AllNodesFailedException) {
-      throw new UnknownOperationResultException(e);
-    } else {
-      throw e;
     }
   }
 
--- a/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
+++ b/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
@@ -45,8 +45,11 @@ import static org.projectnessie.versioned.storage.cassandra.CassandraConstants.S
 import static org.projectnessie.versioned.storage.cassandra.CassandraConstants.TABLE_OBJS;
 import static org.projectnessie.versioned.storage.cassandra.CassandraConstants.TABLE_REFS;
 
+import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.DriverException;
+import com.datastax.oss.driver.api.core.DriverTimeoutException;
+import com.datastax.oss.driver.api.core.NodeUnavailableException;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.BoundStatementBuilder;
@@ -59,6 +62,7 @@ import com.datastax.oss.driver.api.core.metadata.schema.ColumnMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.KeyspaceMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
 import com.datastax.oss.driver.api.core.servererrors.CASWriteUnknownException;
+import com.datastax.oss.driver.api.core.servererrors.QueryExecutionException;
 import jakarta.annotation.Nonnull;
 import java.lang.reflect.Array;
 import java.util.ArrayList;
@@ -80,6 +84,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.agrona.collections.Hashing;
 import org.agrona.collections.Object2IntHashMap;
+import org.projectnessie.versioned.storage.common.exceptions.UnknownOperationResultException;
 import org.projectnessie.versioned.storage.common.persist.Backend;
 import org.projectnessie.versioned.storage.common.persist.PersistFactory;
 import org.slf4j.Logger;
@@ -321,7 +326,7 @@ public final class CassandraBackend implements Backend {
 
   boolean executeCas(BoundStatement stmt) {
     try {
-      ResultSet rs = execute(stmt);
+      ResultSet rs = session.execute(stmt);
       return rs.wasApplied();
     } catch (DriverException e) {
       handleDriverException(e);
@@ -330,16 +335,40 @@ public final class CassandraBackend implements Backend {
   }
 
   ResultSet execute(BoundStatement stmt) {
-    return session.execute(stmt);
+    try {
+      return session.execute(stmt);
+    } catch (DriverException e) {
+      throw unhandledException(e);
+    }
   }
 
   CompletionStage<AsyncResultSet> executeAsync(BoundStatement stmt) {
     return session.executeAsync(stmt);
   }
 
-  void handleDriverException(DriverException e) {
-    if (e instanceof CASWriteUnknownException) {
-      logCASWriteUnknown((CASWriteUnknownException) e);
+  static RuntimeException unhandledException(DriverException e) {
+    if (e instanceof QueryExecutionException) {
+      return new UnknownOperationResultException(e);
+    } else if (e instanceof DriverTimeoutException
+        || e instanceof NodeUnavailableException
+        || e instanceof AllNodesFailedException) {
+      return new UnknownOperationResultException(e);
+    } else {
+      return e;
+    }
+  }
+
+  static void handleDriverException(DriverException e) {
+    if (e instanceof QueryExecutionException) {
+      if (e instanceof CASWriteUnknownException) {
+        logCASWriteUnknown((CASWriteUnknownException) e);
+      } else {
+        throw new UnknownOperationResultException(e);
+      }
+    } else if (e instanceof DriverTimeoutException
+        || e instanceof NodeUnavailableException
+        || e instanceof AllNodesFailedException) {
+      throw new UnknownOperationResultException(e);
     } else {
       throw e;
     }
--- a/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
+++ b/versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
@@ -61,7 +61,6 @@ import com.datastax.oss.driver.api.core.metadata.Metadata;
 import com.datastax.oss.driver.api.core.metadata.schema.ColumnMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.KeyspaceMetadata;
 import com.datastax.oss.driver.api.core.metadata.schema.TableMetadata;
-import com.datastax.oss.driver.api.core.servererrors.CASWriteUnknownException;
 import com.datastax.oss.driver.api.core.servererrors.QueryExecutionException;
 import jakarta.annotation.Nonnull;
 import java.lang.reflect.Array;
@@ -360,29 +359,13 @@ public final class CassandraBackend implements Backend {
 
   static void handleDriverException(DriverException e) {
     if (e instanceof QueryExecutionException) {
-      if (e instanceof CASWriteUnknownException) {
-        logCASWriteUnknown((CASWriteUnknownException) e);
-      } else {
-        throw new UnknownOperationResultException(e);
-      }
+      throw new UnknownOperationResultException(e);
     } else if (e instanceof DriverTimeoutException
         || e instanceof NodeUnavailableException
         || e instanceof AllNodesFailedException) {
       throw new UnknownOperationResultException(e);
     } else {
       throw e;
-    }
-  }
-
-  @SuppressWarnings("Slf4jDoNotLogMessageOfExceptionExplicitly")
-  private static void logCASWriteUnknown(CASWriteUnknownException e) {
-    if (LOGGER.isDebugEnabled()) {
-      LOGGER.debug(
-          "Assuming CAS failed due to CASWriteUnknownException with message'{}', coordinator: {}, errors: {}, warnings: {}",
-          e.getMessage(),
-          e.getExecutionInfo().getCoordinator(),
-          e.getExecutionInfo().getErrors(),
-          e.getExecutionInfo().getWarnings());
     }
   }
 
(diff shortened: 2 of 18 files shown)


versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -329,16 +334,40 @@ boolean executeCas(BoundStatement stmt) {
...
 
   CompletionStage<AsyncResultSet> executeAsync(BoundStatement stmt) {
     return session.executeAsync(stmt);
   }
 
-  void handleDriverException(DriverException e) {
-    if (e instanceof CASWriteUnknownException) {
-      logCASWriteUnknown((CASWriteUnknownException) e);
+  static RuntimeException unhandledException(DriverException e) {
+    if (e instanceof QueryExecutionException) {
```
adutra, line 349: I think you can narrow down to `QueryConsistencyException`.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -329,16 +334,40 @@ boolean executeCas(BoundStatement stmt) {
...
+      return new UnknownOperationResultException(e);
+    } else if (e instanceof DriverTimeoutException
+        || e instanceof NodeUnavailableException
```
adutra, line 352: `NodeUnavailable` is safe imho, this means that we couldn't even connect it, let alone send a query to it.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -329,16 +334,40 @@ boolean executeCas(BoundStatement stmt) {
...
+        || e instanceof AllNodesFailedException) {
```
adutra, line 353: In theory you should iterate over `getAllErrors()` and recursively call this method for each one of them.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -329,16 +334,40 @@ boolean executeCas(BoundStatement stmt) {
...
+      return new UnknownOperationResultException(e);
+    } else {
+      return e;
+    }
+  }
+
+  static void handleDriverException(DriverException e) {
+    if (e instanceof QueryExecutionException) {
+      if (e instanceof CASWriteUnknownException) {
+        logCASWriteUnknown((CASWriteUnknownException) e);
```
adutra, line 363: Why don't we throw in this case?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
```diff
@@ -60,6 +61,10 @@ static <T> T commitRetry(Persist persist, CommitAttempt<T> attempt, TryLoopState
           throw new RetryTimeoutException(i, tls.currentNanos() - t0);
         }
         retryState = e.retryState();
+      } catch (UnknownOperationResultException e) {
+        if (!tls.retry(t1)) {
+          throw new RetryTimeoutException(i, tls.currentNanos() - t0);
```
adutra, line 66: Does this retry with a different commit timestamp? (I guess so, just wanted to confirm.)

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -649,13 +747,63 @@ protected Obj computeNext() {
...
+    switch (e.getError().getCategory()) {
+      case EXECUTION_TIMEOUT:
+      case UNCATEGORIZED:
+        return new UnknownOperationResultException(e);
+      default:
+        return e;
+    }
+  }
+
+  static RuntimeException handleMongoWriteError(MongoException e, WriteError error) {
```
adutra, line 800: These two method could maybe be merged together.

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
```diff
@@ -649,13 +747,63 @@ protected Obj computeNext() {
...
+        || e instanceof MongoExecutionTimeoutException) {
+      return new UnknownOperationResultException(e);
+    }
+    if (e instanceof MongoWriteException) {
+      return handleMongoWriteException((MongoWriteException) e);
+    }
+    if (e instanceof MongoBulkWriteException) {
+      MongoBulkWriteException specific = (MongoBulkWriteException) e;
+      for (BulkWriteError error : specific.getWriteErrors()) {
+        switch (error.getCategory()) {
```
adutra, line 778: Maybe call `handleMongoWriteError`?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: But that would ignore OverloadedException + UnavailableException

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: However the javadoc says: "A common reason to encounter this error is when the configured number of connections per node and requests per connection is not high enough to absorb the overall request rate."

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
```diff
@@ -60,6 +61,10 @@ static <T> T commitRetry(Persist persist, CommitAttempt<T> attempt, TryLoopState
           throw new RetryTimeoutException(i, tls.currentNanos() - t0);
         }
         retryState = e.retryState();
+      } catch (UnknownOperationResultException e) {
+        if (!tls.retry(t1)) {
```
dimas-b, line 65: I'm not sure about retrying the whole commit in this case. Suppose the timeout occurs, but the change is still persisted by the database... Would it make sense to try reading the reference HEAD to check whether it's advanced?

In case of a persisted write with a timeout, if we retry building the commit, it will conflict with its own previous (persisted and readable) state and result in a cryptic conflict error at the user level.. WDYT?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: Yea - with this change we can remove this special handling.

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestVersionStoreImpl.java
```diff
@@ -156,4 +165,80 @@ public Reference updateReferencePointer(
...
+            (BiFunction<AtomicBoolean, Persist, Persist>)
+                (called, p) ->
+                    new PersistDelegate(p) {
+                      @Nonnull
+                      @Override
+                      public Reference updateReferencePointer(
+                          @Nonnull Reference reference, @Nonnull ObjId newPointer)
+                          throws RefNotFoundException, RefConditionFailedException {
+                        if (called.compareAndSet(false, true)) {
+                          throw new UnknownOperationResultException("test", new RuntimeException());
```
dimas-b, line 206: Could we also simulate persisted writes (`updateReferencePointer`) with a timeout (`UnknownOperationResultException`)?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
snazy: Logically no (`CommitMetadata` is supplied externally) - but the underlying `CommitObj` is created for every attempt.

versioned/storage/mongodb/src/main/java/org/projectnessie/versioned/storage/mongodb/MongoDBPersist.java
snazy: Slightly different here

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: Is it worth it?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
```diff
@@ -329,30 +333,38 @@ boolean executeCas(BoundStatement stmt) {
...
-          e.getExecutionInfo().getCoordinator(),
-          e.getExecutionInfo().getErrors(),
-          e.getExecutionInfo().getWarnings());
+  static void handleDriverException(DriverException e) {
+    if (e instanceof QueryExecutionException) {
+      throw new UnknownOperationResultException(e);
+    } else if (e instanceof DriverTimeoutException
+        || e instanceof NodeUnavailableException
+        || e instanceof AllNodesFailedException) {
+      throw new UnknownOperationResultException(e);
```
dimas-b, line 365: Isn't it the same as the first `if`? Why have a separate `if` block?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
adutra: But that's not OK, is it? If the first commit attempt actually succeeded, but `UnknownOperationResultException` was thrown, then attempting to store the same object again would necessarily result in `Persist.storeObj()` returning false.

dimas-b: Could we have a test for this (I also commented on that in another thread)?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
```diff
@@ -186,11 +199,29 @@ CommitResult<Commit> commit(
...
+      // If 'commitRetryState.storedContents' already contains the commit-ID, __we__ already
+      // successfully persisted that commit. This can happen, if the `Persist` implementation raised
+      // an `UnknownOperationResultException` in one of the following operations (reference bump for
+      // example).
+      boolean stored = commitLogic.storeCommit(newHead, objectsToStore);
+      if (stored) {
+        commitRetryState.commitPersisted = newHead.id();
+        commitRetryState.storedContents.addAll(toStore);
+      }
+      boolean commitPersisted = stored || newHead.id().equals(commitRetryState.commitPersisted);
```
dimas-b, line 222: When can `newHead.id()` equal `commitRetryState.commitPersisted` if `stored == false`?

I suppose `newHead` is unique in every call to `buildCommitObj` if the microsecond clock ticks (which is mostly a given in our retry loop, I guess).

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
```diff
@@ -186,11 +199,29 @@ CommitResult<Commit> commit(
...
+              NO_VALUE_REPLACEMENT);
+
+      // If 'commitRetryState.storedContents' already contains the commit-ID, __we__ already
+      // successfully persisted that commit. This can happen, if the `Persist` implementation raised
+      // an `UnknownOperationResultException` in one of the following operations (reference bump for
+      // example).
+      boolean stored = commitLogic.storeCommit(newHead, objectsToStore);
+      if (stored) {
+        commitRetryState.commitPersisted = newHead.id();
+        commitRetryState.storedContents.addAll(toStore);
```
dimas-b, line 220: Is `commitRetryState.storedContents` still needed? I could not find where its value is read :thinking:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
dimas-b: I think we should rebuild the commit object only if the reference HEAD changes between retries. WDYT?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
snazy: > But that's not OK, is it? If the first commit attempt actually succeeded, but UnknownOperationResultException was thrown, then attempting to store the same object again would necessarily result in Persist.storeObj() returning false.

These cases are all handled in `CommitImpl.commit()` and tested via `TestVersionStoreImpl.commitWithDatabaseTimeout()`.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
snazy: > Would it make sense to try reading the reference HEAD to check whether it's advanced?

That's what `BaseCommitHelper.committingOperation()` does, no?

> In case of a persisted write with a timeout, if we retry building the commit, it will conflict with its own previous (persisted and readable) state and result in a cryptic conflict error at the user level.

How should it conflict, if the ref HEAD didn't change?

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestVersionStoreImpl.java
snazy: Sorry - but, have you read the body of `public static Stream<Arguments> commitWithDatabaseTimeout()`?

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: Sorry, I do not see any similarity.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
snazy: It's not unique, think how commit-obj-IDs are generated.

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestVersionStoreImpl.java
dimas-b: I have, but I do not see where it simulates a timeout after a persisted write... I might have missed it, though.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
adutra: Yes but for these two exceptions the result is not "unknown": it's actually 100% sure the write did not happen.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
adutra: Right; the only place where this error is raised is when there is no Netty channel available in the pool:

https://github.com/apache/cassandra-java-driver/blob/16260261d3df50fcf24fac1fc2d37896c4a111bf/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java#L255-L259

So yes, I think we can safely assume the coordinator never got the write.

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
adutra: I think so? This is the error that is surfaced to clients when retries happened behind the scenes. For example, `CASWriteUnknownException` is usually retried, so you normally don't encounter it directly, but rather, nested inside `AllNodesFailedException`.

(that said, I think we don't retry a ton of things... the default idempotence is false – I will look into that.)

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
dimas-b: Like 141 is a comment in this file :thinking:

versioned/storage/cassandra/src/main/java/org/projectnessie/versioned/storage/cassandra/CassandraBackend.java
snazy: > In theory you should iterate over getAllErrors() and recursively call this method for each one of them.

(Side rant: this C* driver is overly complicated)

versioned/storage/bigtable/src/main/java/org/projectnessie/versioned/storage/bigtable/BigTablePersist.java
```diff
@@ -97,6 +102,12 @@ public class BigTablePersist implements Persist {
   }
 
   static RuntimeException apiException(ApiException e) {
+    if (e instanceof DeadlineExceededException
+        || e instanceof WatchdogTimeoutException
+        || e instanceof UnknownException
+        || e instanceof AbortedException) {
+      throw new UnknownOperationResultException("Unhandled BigTable exception", e);
```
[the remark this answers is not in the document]
snazy: This exception message is never exposed to users.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
snazy: Sorry if I sounded overly harsh - Just didn't understand that you meant something different.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitRetry.java
snazy: Updated the PR (added some more test cases) - basically the existing 3 existing cases, which "timeout" without changing the persisted state, but with a "timeout" after changing the persisted state.

* For `storeObjs` (content objects) - that's a "no brainer"
* For storing the commit itself it result(ed) in a "hash collision" - but that could be worked around - with a related change for transplant operations ("fast forward transplants") - not breaking tho - so should be fine.
* For bumping the reference it's trickier - the workaround in this change works often - but won't work for concurrent commits to the same same reference that run into timeouts. I don't see a way to guarantee the "right" outcome without adding a ton of complexity - which might not even work. I think this falls into the "_Adventures in distributed computing_" category... - probably okay to leave it as it is.

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestVersionStoreImpl.java
snazy: (see above - and apologies that I sounded too harsh)

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
snazy: There's a `if (storedContents.add(...))`

dimas-b: Sorry, I really cannot find that `if/add` :sweat_smile:

versioned/storage/store/src/test/java/org/projectnessie/versioned/storage/versionstore/TestVersionStoreImpl.java
dimas-b: no worries :+1:

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
snazy: I meant line 181 - but that's somehow not exactly how I always thought it is

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
dimas-b: My mistake... I thought the microsecond commit time was part of the commit hash (but it is not) :facepalm:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -302,14 +303,38 @@ public boolean storeCommit(@Nonnull CommitObj commit, @Nonnull List<Obj> additio
...
+      return commit;
+    }
+
+    // Check whether the existing object is the same commit (w/o considering the internal "created"
+    // timestamp of the commit-obj).
+    try {
+      CommitObj existing = persist.fetchTypedObj(commit.id(), COMMIT, CommitObj.class);
+      CommitObj commitWithNewCreatedTimestamp =
+          CommitObj.commitBuilder().from(commit).created(existing.created()).build();
+      return commitWithNewCreatedTimestamp.equals(existing) ? existing : null;
```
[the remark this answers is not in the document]
snazy: Because this one shall return the _persisted_ object, not necessarily the one that's been passed in.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
```diff
@@ -618,6 +619,14 @@ void validateMergeTransplantCommit(
   void bumpReferencePointer(ObjId newHead, Optional<?> retryState) throws RetryException {
     try {
       persist.updateReferencePointer(reference, newHead);
+    } catch (UnknownOperationResultException e) {
+      // If the above pointer-bump returned an "unknown result", we check once (and only once!)
+      // whether the reference-pointer-change succeeded. This mitigation may not always work,
+      // especially not in highly concurrent update situations.
+      Reference r = persist.fetchReference(reference.name());
```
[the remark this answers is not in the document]
snazy: Don't wanna wait here. This is really only a "best effort", but timeout handling here makes things overly complex IMO.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
```diff
@@ -186,11 +199,31 @@ CommitResult<Commit> commit(
...
+      // an `UnknownOperationResultException` in one of the following operations (reference bump for
+      // example).
+      CommitObj stored = commitLogic.storeCommit(newHead, objectsToStore);
+      if (stored != null) {
+        newHead = stored;
+        commitRetryState.commitPersisted = newHead.id();
+        commitRetryState.storedContents.addAll(toStore);
+      }
+      boolean commitPersisted =
+          stored != null || newHead.id().equals(commitRetryState.commitPersisted);
```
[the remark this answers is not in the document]
snazy: > why even attempt a duplicate storeCommit call?

We have to do something here, if we get here again - either the commit, related objects or the reference weren't properly persisted. But we do not exactly now what and why and whether sth was persisted in case of an `UnknownOperationResultException`.

I suspect these timeouts to be rare - so would like to not add to much code to handle these cases (but complicate the usual good cases).

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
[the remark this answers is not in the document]
dimas-b: I meant we'd just let the normal retry loop wait in this case. I'm ok with a best-effort local re-read too.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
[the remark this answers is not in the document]
dimas-b: yes, sorry, I misread.

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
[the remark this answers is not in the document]
snazy: > normal retry loop wait in this case

That (sadly) causes a key-conflict.

> ... but re-reading still need a call to fetchReferenceForUpdate, right?

FML - I updated it yesterday - pushed now

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -302,14 +300,38 @@ public boolean storeCommit(@Nonnull CommitObj commit, @Nonnull List<Obj> additio
...
+    if (storeResult) {
+      return commit;
+    }
+
+    // Check whether the existing object is the same commit (w/o considering the internal "created"
+    // timestamp of the commit-obj).
+    try {
+      CommitObj existing = persist.fetchTypedObj(commit.id(), COMMIT, CommitObj.class);
+      CommitObj commitWithNewCreatedTimestamp =
+          CommitObj.commitBuilder().from(commit).created(existing.created()).build();
```
dimas-b, line 328: Would it make sense to mark `create()` as `@Value.Auxiliary`? It should hopefully also make it clear why we do not hash it. WDYT?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/CommitImpl.java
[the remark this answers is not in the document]
dimas-b: good point :+1:

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
snazy: Why? That's not a marker annotation.

dimas-b (approved):

PR #8539, opened 2024-05-20
[Catalog] Use production config+tree resources from Nessie Catalog

snazy:
The Nessie client used from Nessie Catalog is currently a "simple" one. This change lets it use a client that's configured against the actual config+tree resources, which also perform authZ.

### DIFF
--- a/servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
+++ b/servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
@@ -29,6 +29,7 @@ import io.smallrye.context.SmallRyeThreadContext;
 import io.vertx.ext.web.client.WebClientOptions;
 import io.vertx.mutiny.core.Vertx;
 import io.vertx.mutiny.ext.web.client.WebClient;
+import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.inject.Disposes;
 import jakarta.enterprise.inject.Produces;
 import jakarta.inject.Named;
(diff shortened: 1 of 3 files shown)


testing/combined-cs/build.gradle.kts
```diff
@@ -32,6 +32,7 @@ dependencies {
   implementation(project(":nessie-versioned-storage-common"))
   implementation(project(":nessie-versioned-storage-inmemory"))
   implementation(project(":nessie-versioned-storage-store"))
+  implementation(libs.guava)
```
adutra, line 35: Shall we move the combined client out of the `testing` folder?

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
```diff
@@ -284,9 +286,13 @@ public Executor importJobExecutor(ThreadContext threadContext, CatalogServiceCon
...
-  @Singleton
-  public NessieApiV2 nessieApiV2(Persist persist) {
-    return new CombinedClientBuilder().withPersist(persist).build(NessieApiV2.class);
+  @RequestScoped
+  public NessieApiV2 nessieApiV2(
+      RestV2ConfigResource configResource, RestV2TreeResource treeResource) {
+    return new CombinedClientBuilder()
+        .withConfigResource(configResource)
+        .withTreeResource(treeResource)
+        .buildForProduction(NessieApiV2.class);
```
adutra, line 295: Do we really need the new method? Calling the old `build` method would yield the same result.

testing/combined-cs/build.gradle.kts
snazy: Want to do that, but in a separate PR.

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
snazy: Wanna give that another look separately. Potentially removing the code path in `build()` if it's not used anywhere.

adutra (approved):

PR #8544, opened 2024-05-21
Nessie server: add support for MariaDB and MySQL backends

adutra:


versioned/storage/jdbc-tests/src/main/java/org/projectnessie/versioned/storage/jdbctests/MariaDBBackendTestFactory.java
```diff
@@ -16,21 +16,34 @@
 package org.projectnessie.versioned.storage.jdbctests;
 
 import jakarta.annotation.Nonnull;
+import java.util.Map;
 import org.projectnessie.versioned.storage.jdbc.JdbcBackendFactory;
 import org.testcontainers.containers.JdbcDatabaseContainer;
 import org.testcontainers.containers.MariaDBContainer;
 
-public class MariaDbBackendTestFactory extends ContainerBackendTestFactory {
+public class MariaDBBackendTestFactory extends ContainerBackendTestFactory {
```
adutra, line 24: Renaming, for consistency with how the MariaDB project names itself.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
```diff
@@ -508,6 +508,13 @@ private boolean[] upsertObjs(
               if (updated[i] == 1) {
                 r[batchIndexToObjIndex.get(i)] = true;
               }
+              if (updated[i] == PreparedStatement.SUCCESS_NO_INFO) {
```
adutra, line 511: This is a consequence of how MariaDB sometimes reports a successful batch: it returns `SUCCESS_NO_INFO` by default, unless `useBulkStmtsForInserts` is set to false. I think it's wrong to ignore this here. For now I'm throwing ISE, but if #8533 is merged we could instead throw `UnknownOperationResultException`.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/DatabaseSpecific.java
```diff
@@ -19,29 +19,12 @@
 import java.util.Map;
 
 public interface DatabaseSpecific {
-  /** H2 integrity constraint violation. */
-  int CONSTRAINT_VIOLATION_SQL_CODE = 23505;
```
adutra, line 23: Unused constants and methods.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/DatabaseSpecifics.java
```diff
@@ -52,8 +52,7 @@ private DatabaseSpecifics() {}
   public static final DatabaseSpecific H2_DATABASE_SPECIFIC =
       new BasePostgresDatabaseSpecific("VARCHAR");
 
-  public static final DatabaseSpecific MYSQL_MARIADB_DATABASE_SPECIFIC =
```
adutra, line 55: Renaming to remove the word `MYSQL`.

servers/quarkus-server/src/main/resources/application.properties
```diff
@@ -85,15 +85,41 @@ nessie.version.store.type=IN_MEMORY
...
 ## Transactional database configuration
-# Note: Nessie Quarkus Server is built with io.quarkus:quarkus-jdbc-postgresql only. If you need
-# another database, the corresponding Quarkus extensions + driver need to be included in the build
-# and the following setting updated.
+
+# Note: Nessie Quarkus Server comes with built-in support for Postgres and MariaDB, or any database
+# compatible with these. If you need another database, the corresponding Quarkus extension + driver
+# needs to be included in the build.
+# Select the datasource to use with the `nessie.version.store.persist.jdbc.datasource` property;
+# The possible built-in values are: "default" (deprecated), "postgresql" and "mariadb".
```
snazy, line 93: Same Q as in `QuarkusJdbcConfig` ("Is it deprecated in Quarkus?").
Nessie doesn't support multiple datasources to be used in parallel/concurrently. Why not just always use the default one?

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/BigTableTestResourceLifecycleManager.java
```diff
@@ -37,13 +36,9 @@ public void setIntegrationTestContext(DevServicesContext context) {
   @Override
   public Map<String, String> start() {
     bigTable = new BigTableBackendContainerTestFactory();
-    bigTable.startBigtable(containerNetworkId);
+    bigTable.start(containerNetworkId);
```
snazy, line 39: Confused - how are these refactorings touching all databases part of the Maria/MySQL change?

snazy (commented):
Looks like a bunch of unrelated changes made it into this PR. Can you move these to a separate one?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/QuarkusJdbcConfig.java
```diff
@@ -30,6 +30,13 @@
 @ConfigMapping(prefix = "nessie.version.store.persist.jdbc")
 public interface QuarkusJdbcConfig extends JdbcBackendBaseConfig {
 
+  /**
+   * The name of the datasource to use. Must correspond to a configured datasource under {@code
```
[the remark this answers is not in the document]
adutra: No, it's deprecated only in Nessie.

servers/quarkus-server/src/main/resources/application.properties
adutra: > Why not just always use the default one?

Because `db-kind` is a build-time property. So you need to declare each datasource you intend to _maybe_ use, one by one, using different names.

adutra:
> Looks like a bunch of unrelated changes made it into this PR. Can you move these to a separate one?

Which changes are unrelated? All the changes in the persist layer were done in order to optimize the configuration of Quarkus tests, by introducing the new `BackendTestFactory.getQuarkusConfig()`. (This not only reduces some code duplication, but also allows the server admin tool to reuse the same test setups.)

The only change that is a bit unrelated indeed is the one I mentioned here:

https://github.com/projectnessie/nessie/pull/8544#discussion_r1607861446

I will move that one to another PR.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
adutra: This snippet has been moved to #8551.

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/BigTableTestResourceLifecycleManager.java
snazy: Makes sense, but I'd prefer to have that as a separate change.

adutra: Okay moved all that stuff to #8553 – this PR is back to draft state.

adutra:
Ready for review again!

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,74 @@
...
+      return VersionStoreType.IN_MEMORY;
+    }
+    return VersionStoreType.valueOf(versionStoreType.getValue());
+  }
+
+  private static String dataSourceName(String name) {
+    if (name.equals("quarkus.datasource.active")) {
+      return "default";
+    }
+    return name.substring("quarkus.datasource.".length(), name.length() - ".active".length());
```
snazy, line 57: Is this safe for quoted datasource names like `quarkus.datasource."my.foo.bar-source".active`?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,74 @@
...
+      if (versionStoreType(context) == VersionStoreType.JDBC) {
+        active = dataSourceName(name).equals(activeDataSourceName(context));
+      }
+      return newConfigValue(active);
+    }
+    return context.proceed(name);
+  }
+
+  private static VersionStoreType versionStoreType(ConfigSourceInterceptorContext context) {
+    ConfigValue versionStoreType = context.proceed("nessie.version.store.type");
```
snazy, line 46: Can we memoize this in a field of `DataSourceActivator`?

servers/quarkus-server/src/main/resources/application.properties
adutra: To add a bit more context about how this works:

* Quarkus can support more than one datasource, but since `db-kind` is a build-time property, all datasources and their respective extensions must be present at build time. 
* However, the `active` property is a runtime one, and can be used to selectively disable one or more datasources at runtime. This is the mechanism implemented here.
* 2 new datasources are declared with names: `postgresql` and `mariadb`. (Note that adding more datasources in the future becomes easier: "just" add the extension + one new named datasource.)
* Instead of requiring users to manually activate the datasource they want to use, I came up instead with the new option `nessie.version.store.persist.jdbc.datasource`. The`DataSourceActivator` class will lookup the datasource name and activate only that specific datasource.
* The big pain point is what to do with the "default" datasource. For now it still works as before and maps to PostgresQL, which makes upgrading easier. But my suggestion is to gradually deprecate then remove this datasource and encourage users to migrate to the new named `postgresql` datasource.
* For posterity, I also tried to implement something similar to Quarkus' `OTelFallbackConfigSourceInterceptor` (which transparently migrates legacy properties to their new names) – but datasources are really a build-time thing, and this approach did not work because the properties would be migrated at runtime – too late.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private static String dataSourceName(String name) {
+    if (name.equals("quarkus.datasource.active")) {
+      return "default";
+    }
+    return name.substring("quarkus.datasource.".length(), name.length() - ".active".length());
+  }
+
+  private static String activeDataSourceName(ConfigSourceInterceptorContext context) {
+    ConfigValue nessieDataSource = context.proceed("nessie.version.store.persist.jdbc.datasource");
```
snazy, line 61: Can we memoize this in a field of `DataSourceActivator`?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final int APPLICATION_PROPERTIES_CLASSPATH_ORDINAL = 250;
+
+  @Override
+  public ConfigValue getValue(ConfigSourceInterceptorContext context, String name) {
+    if (name.startsWith("quarkus.datasource.") && name.endsWith(".active")) {
+      boolean active = false;
+      if (versionStoreType(context) == VersionStoreType.JDBC) {
+        active = dataSourceName(name).equals(activeDataSourceName(context));
+      }
+      return newConfigValue(active);
```
snazy, line 40: Wonder whether we always want a new `ConfigValue` - even for the case when the value isn't changed.

WDYT about this?
```java
  @Override
  public ConfigValue getValue(ConfigSourceInterceptorContext context, String name) {
    ConfigValue configValue = context.proceed(name);
    if (name.startsWith("quarkus.datasource.") && name.endsWith(".active")) {
      boolean active = false;
      if (versionStoreType(context) == VersionStoreType.JDBC) {
        active = dataSourceName(name).equals(activeDataSourceName(context));
      }
      if (active != Boolean.parseBoolean(configValue.getValue())) {
        configValue = ConfigValue.builder()
          .withValue(active ? "true" : "false")
          .withConfigSourceOrdinal(configValue.getSourceOrdinal() + 1).build();
      }
    }
    return configValue;
  }
```

servers/quarkus-tests/src/main/java/org/projectnessie/quarkus/tests/profiles/QuarkusTestProfilePersistMariaDB.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class QuarkusTestProfilePersistMariaDB extends BaseConfigProfile {
+
+  @Override
+  public Map<String, String> getConfigOverrides() {
+    return ImmutableMap.<String, String>builder()
+        .putAll(super.getConfigOverrides())
+        .put("nessie.version.store.type", JDBC.name())
+        .put("nessie.version.store.persist.jdbc.datasource", "mariadb")
+        .put("quarkus.datasource.mariadb.jdbc.extended-leak-report", "true")
```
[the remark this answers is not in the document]
adutra: I copied exactly what was present before in similar profile classes. And yes I think that's enough: `detect-statement-leaks` is true by default and I recall seeing one leak report when I did something stupid a while ago (opened a connection for debugging purposes and forgot to close it.)

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,74 @@
...
+    if (nessieDataSource == null || nessieDataSource.getValue() == null) {
+      return "default";
+    }
+    return nessieDataSource.getValue();
+  }
+
+  private static ConfigValue newConfigValue(boolean value) {
+    return ConfigValue.builder()
+        .withValue(value ? "true" : "false")
+        .withConfigSourceOrdinal(APPLICATION_PROPERTIES_CLASSPATH_ORDINAL + 1)
```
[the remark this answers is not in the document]
adutra: I wanted the value to be less than 260 to allow users to override it, if they know what they are doing (could be an "escape hatch" if `DataSourceActivator` does not behave as intended for some people.) But it also must be more than 250 to make sure it will override the default value `false` in `application.properties`.

259 maybe?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
adutra: The situation doesn't happen right now since we don't have quoted names. But I tested with a `"my.foo.bar-source"` and the code worked, with the caveat that this datasource can only be activated if `nessie.version.store.persist.jdbc.datasource` is quoted as well.

But just to be future proof, let's unquote the names before comparing them.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
adutra: I hesitated because I think this class has to be thread-safe; let me give a try.

site/in-dev/configuration.md
```diff
@@ -68,6 +68,7 @@ Related Quarkus settings:
...
 | Amazon DynamoDB  | beta, only tested against the simulator          | `DYNAMODB`                                              |                                                                                                                                                                                                                                 |
 | PostgreSQL       | production                                       | `JDBC`                                                  |                                                                                                                                                                                                                                 |
+| MariaDB          | experimental, feedback welcome                   | `JDBC`                                                  | This backend can also be used with a MySQL server.                                                                                                                                                                              |
```
snazy, line 71: Should we mention MySQL here as well. Mean - mentioning is not distributing. WDYT?

servers/quarkus-server/src/main/resources/application.properties
snazy: Not great - but since that db-kind is a build-time property I don't see a better way etiher.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/JdbcBackendBuilder.java
```diff
@@ -17,43 +17,63 @@
...
-  String databaseKind;
+  @All
+  @SuppressWarnings("CdiInjectionPointsInspection")
+  List<InstanceHandle<DataSource>> dataSources;
 
   @Inject QuarkusJdbcConfig config;
 
   @Override
   public Backend buildBackend() {
-    if (!DatabaseKind.isPostgreSQL(databaseKind) && !DatabaseKind.isH2(databaseKind)) {
```
snazy, line 47: Oversight or intenional removal of this check?

site/in-dev/configuration.md
adutra: Yep you are right. I was being overly cautious.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
```diff
@@ -0,0 +1,100 @@
...
+public class DataSourceActivator implements ConfigSourceInterceptor {
+
+  /**
+   * The default ordinal to use for modified {@code quarkus.datasource.*.active} properties. It must
+   * be higher than the ordinal of the application.properties classpath config source (250) but
+   * lower than the ordinal of the user-supplied application.properties (260).
+   */
+  private static final int DEFAULT_ORDINAL = 251;
+
+  private static String activeDataSourceName;
```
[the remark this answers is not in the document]
adutra: Hmm the 2 methods that read/write these fields were made synchronized, isn't that enough to guarantee the happens-before relationship between a write and a subsequent read?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
snazy: However, if users configure their instances via system-properties / environment, 300/400 would be the "natural" ordinals.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
snazy: OOps - that's definitely good enough

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/JdbcBackendBuilder.java
snazy: How would it interfere with additional DSs as long as the db-kind is supported?

adutra: Think of someone adding support for, say, Oracle. They'd need to add `&& !DatabaseKind.isOracle(databaseKind)` here, otherwise their application wouldn't start.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
adutra: So, if I use 300 by default, is that OK for you?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/JdbcBackendBuilder.java
snazy: They'd need to adopt the SQL statements as well.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
snazy: Nah - let's use some value between 251 and 259 - that leaves room in the "custom overrides".

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/JdbcBackendBuilder.java
adutra: Fair point, I will put the check back then.

adutra: Hmm sorry but do we need this though? `db-kind` is a build-time property, so for the built-in datasources, we _know_ that the kind is supported. This check will never throw. It would only throw if 1) the server is rebuilt and 2) it is built with a missing extension or driver, or with a wrong `db-kind`.

I mean I don't mind adding it back, but I don't see how it's useful.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
adutra: Right now it's 251 if the value doesn't exist, or the value's ordinal, if it exists. Is that OK?

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/JdbcBackendBuilder.java
snazy: Ugh - yea - it's build time. But maybe leave it in - it would fail quickly in such a case, so devs do know that something needs to be done.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/datasource/DataSourceActivator.java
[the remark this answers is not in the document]
snazy: Sure

snazy (approved):
LGTM!
But let's wait for "all the MariaDB" PRs to be ready and merge those "together".

PR #8545, opened 2024-05-21
Nessie GC: add support for MariaDB and MySQL backends

adutra:


site/in-dev/gc.md
```diff
@@ -228,8 +231,8 @@ Modules that supplement the `gc-base` module:
 * `gc-iceberg-files` implements file listing + deletion using Iceberg's `FileIO`.
 * `gc-iceberg-mock` is a testing-only module to generate mock metadata, manifest-lists, manifests
   and (empty) data files.
-* `gc-repository-jdbc` implements the live-content-sets-store using JDBC (PostgreSQL and compatible)
-  .
+* `gc-repository-jdbc` implements the live-content-sets-store using JDBC (PostgreSQL, MariaDB, MySQL 
+  and any other compatible database).
```
snazy, line 235:
```suggestion
  and compatible databases).
```

gc/gc-repository-jdbc/src/intTest/java/org/projectnessie/gc/contents/jdbc/ITPostgresPersistenceSpi.java
```diff
@@ -44,7 +44,7 @@ static void stopContainer() {
     }
   }
 
-  protected static String dockerImage(String dbName) {
+  public static String dockerImage(String dbName) {
```
snazy, line 47: I think this is a conflict (or fails)

gc/gc-repository-jdbc/src/intTest/resources/org/projectnessie/gc/contents/jdbc/Dockerfile-mariadb-version
```diff
@@ -0,0 +1,3 @@
+# Dockerfile to provide the image name and tag to a test.
+# Version is managed by Renovate - do not edit.
+FROM mariadb:11.3.2
```
snazy, line 3: Can you adopt this and the mysql-version to use pinned dependencies?

gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
```diff
@@ -87,6 +88,15 @@ public interface Builder {
     JdbcPersistenceSpi build();
   }
 
+  @Value.Lazy
+  protected String productName() {
+    try (Connection conn = dataSource().getConnection()) {
```
snazy, line 93: Not strongly opposed to get the "product name" this way - although using a JDBC connection during an "immutables.build()" isn't "nice".
Isn't there a way to get that information from the data-source directly?

gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
```diff
@@ -419,6 +429,16 @@ LiveContentSet liveContentSet(ResultSet rs) throws SQLException {
         .build();
   }
 
+  protected String decorateInsertStatement(String statement) {
+    switch (productName()) {
+      case "mysql":
+      case "mariadb":
+        return statement.replace("INSERT", "INSERT IGNORE");
+      default:
```
snazy, line 437: Can you replace this with `case "postgres":` and let the `default:` case just throw?

gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
adutra: It's marked lazy exactly because of that – the JDBC connection will be created only when the method is called from `decorateInsertStatement`.

adutra: Unfortunately there is nothing useful under `((AgroalDataSource) dataSource()).getConfiguration()` that could hint at which database is being used.

We _could_ parse the user-provided JDBC URL; but then we'd need to pass it around. Not sure that's better.

snazy (dismissed):
LGTM!
But let's wait for "all the MariaDB" PRs to be ready and merge those "together".

snazy (dismissed):

snazy (approved):

PR #8546, opened 2024-05-21
Testing: Centralize image resolution

snazy:
* Add `ContainerSpecHelper` to help resolving a Docker image
* Add support for "pinned" image references (including "sha256")

### DIFF
--- a/testing/container-spec-helper/src/main/java/org/projectnessie/nessie/testing/containerspec/ContainerSpecHelper.java
+++ b/testing/container-spec-helper/src/main/java/org/projectnessie/nessie/testing/containerspec/ContainerSpecHelper.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.nessie.testing.containerspec;
+
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.immutables.value.Value.Style.ImplementationVisibility.PACKAGE;
+
+import java.io.BufferedReader;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+import java.util.Locale;
+import java.util.Objects;
+import org.immutables.value.Value;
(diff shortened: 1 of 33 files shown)


snazy:
Required for #8541

testing/container-spec-helper/src/main/java/org/projectnessie/nessie/testing/containerspec/ContainerSpecHelper.java
```diff
@@ -0,0 +1,115 @@
...
+
+  public abstract Class<?> containerClass();
+
+  public DockerImageName dockerImageName(String explicitImageName) {
+    if (explicitImageName != null) {
+      return DockerImageName.parse(explicitImageName);
+    }
+
+    String dockerfile = format("Dockerfile-%s-version", name());
+    URL resource = containerClass().getResource(dockerfile);
```
adutra, line 59: `containerClass()` is only used to serve as the class loader to use here. Isn't possible to just use `this.getClass()` instead?

versioned/storage/jdbc-tests/src/main/java/org/projectnessie/versioned/storage/jdbctests/CockroachBackendTestFactory.java
```diff
@@ -30,6 +30,7 @@ public String getName() {
   @Nonnull
   @Override
   protected JdbcDatabaseContainer<?> createContainer() {
-    return new CockroachContainer(dockerImage("cockroach"));
+    return new CockroachContainer(
+        dockerImage("cockroach").asCompatibleSubstituteFor("cockroachdb/cockroach"));
```
adutra, line 34: What motivates the addition of `.asCompatibleSubstituteFor()`? I've seen it added in a few places, but not everywhere.

testing/container-spec-helper/src/main/java/org/projectnessie/nessie/testing/containerspec/ContainerSpecHelper.java
snazy: That would result in a re-build and re-test of all modules that use a container for all unrelated image version bumps - not a fan of that.

versioned/storage/jdbc-tests/src/main/java/org/projectnessie/versioned/storage/jdbctests/CockroachBackendTestFactory.java
snazy: Required by the testcontainer classes.

testing/container-spec-helper/src/main/java/org/projectnessie/nessie/testing/containerspec/ContainerSpecHelper.java
snazy: Yea - that comment was for the other thread.
We have to do one thing - either use an "absolute path" to the Dockerfile or a `Class` +name tuple (note: we need the name anyways).

snazy: Note: having all Dockerfiles in the "same directory" would yield duplicate resources - causing warnings and potentially conflicts later.

adutra (dismissed):

adutra (approved):

PR #8551, opened 2024-05-21
JDBC persist: properly handle SUCCESS_NO_INFO and EXECUTE_FAILED

adutra:
`PreparedStatement.executeBatch()` returns an array of int results. Each element of the array denotes how many rows were affected, but two special values exist: `SUCCESS_NO_INFO` and `EXECUTE_FAILED`. 

So far, no supported Jdbc backend returns such values, but this will change soon (spoiler alert: MariaDB does return `SUCCESS_NO_INFO`).

Ignoring these special values can be very bad as the code would assume no rows were affected, while in fact a row has been affected. It's safer to throw an exception when that happens (in the case of MariaDB, there is a connection option to disallow `SUCCESS_NO_INFO`).

Note: when #8533 is merged we could change the exception thrown. Currently, it's ISE.

### DIFF
--- a/versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
+++ b/versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
@@ -508,6 +508,13 @@ abstract class AbstractJdbcPersist implements Persist {
               if (updated[i] == 1) {
                 r[batchIndexToObjIndex.get(i)] = true;
               }
+              if (updated[i] == PreparedStatement.SUCCESS_NO_INFO) {
+                throw new IllegalStateException(
+                    "driver returned SUCCESS_NO_INFO for a batch update");
+              } else if (updated[i] == PreparedStatement.EXECUTE_FAILED) {
+                throw new IllegalStateException(
+                    "driver returned EXECUTE_FAILED for a batch update");
+              }
             }
           };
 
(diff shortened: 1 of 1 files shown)


versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
```diff
@@ -508,6 +508,13 @@ private boolean[] upsertObjs(
               if (updated[i] == 1) {
                 r[batchIndexToObjIndex.get(i)] = true;
               }
+              if (updated[i] == PreparedStatement.SUCCESS_NO_INFO) {
```
dimas-b, line 511: Since all special values are an error to Nessie, why not check `< 0`?

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
```diff
@@ -508,6 +508,13 @@ private boolean[] upsertObjs(
               if (updated[i] == 1) {
```
dimas-b, line 508: tangential: what about `updated[i] > 1`?.. Should we treat it as an unexpected error too?

dimas-b (approved):

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
adutra: Since each statement we do is a simple row insert, I guess that could never happen. But OK to add a check.

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/AbstractJdbcPersist.java
adutra: To get a better error message.

adutra: OK to change, better to cover all cases outside of 0 and 1.

dimas-b (approved):

PR #8558, opened 2024-05-21
Add exception mappers to convert storage failures to Iceberg REST client exceptions

### LINKED ISSUE #8738: [Catalog] Verify that the Iceberg exception mapping returns the correct status codes and error object contents

dimas-b:
* Note: before this change a storage 403 error would manifest as 500 on the REST client side.

* Remove `ObjectIOException`. Use `BackendErrorStatus` instead.

* Add exception mappers to convert storage failures to `BackendErrorStatus`
  and to Iceberg REST client exceptions.

* Store error codes in TaskStatus and convert them to Iceberg REST
  status code when failed tasks are accessed again.

* Add AccessCheckHandler to ObjectStorageMock for simulating access 
  failures in tests.

Closes #8738


catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/StorageFailureException.java
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.files.api;
+
+import org.projectnessie.storage.uri.StorageUri;
+
+public class StorageFailureException extends NonRetryableException {
```
snazy, line 20: I think the "hard"/unconditional association with "non-retryable" is not good.
We need a way to distinguish "hard/final failures" from "retryable" ones, so that `org.projectnessie.catalog.service.impl.Util#throwableAsErrorTaskState` (via e.g. `org.projectnessie.catalog.service.impl.EntitySnapshotTaskBehavior#asErrorTaskState`) can return the right state (retryable vs final).

Retryable examples:
* Throttled requests
* Technical network errors ("no route to host", "read timeout")
* Other unknown (IO)Exceptions
* Forbidden (users may "fix" credentials or ACLs)

Non-retryable ones:
* Not found

We might eventually need a third TaskState - to handle "forbidden"/"access denied" better, to handle the case when the object store's ACL's just wrong: a state that's immediately reported as an error, but retried later.

snazy:
Oh - seems the deletion of the "merge base" closed this PR :(

dimas-b:
No worries. I'll resume on main.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOOutputStream.java
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.files.api;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import javax.annotation.Nonnull;
+
+public class ObjectIOOutputStream extends OutputStream {
```
snazy, line 22: Need `flush()` as well

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOInputStream.java
```diff
@@ -0,0 +1,57 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.files.api;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+public class ObjectIOInputStream extends InputStream {
```
snazy, line 21: Need these methods as well:
`available()`
`readAllBytes()`
`readNBytes`
`skip`
`skipNBytes`
`transferTo`

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
```diff
@@ -89,6 +93,33 @@ public Response toResponse(Throwable ex, IcebergEntityKind kind) {
     return Response.status(code == null ? 500 : code).entity(body).build();
   }
 
+  private static String message(ObjectIOException e) {
+    return String.format(
+        "Unable to %s %s due to: %s",
+        (e.ioMode().name().toLowerCase(Locale.ROOT)),
```
snazy, line 99: Maybe add the lower-case representation via a function on the enum to avoid the `.toLowerCase`

servers/quarkus-server/src/test/java/org/projectnessie/server/catalog/AbstractIcebergCatalogUnitTests.java
```diff
@@ -205,4 +215,52 @@ public void namespacesPaging() throws Exception {
                 .mapToObj(i -> TableIdentifier.of("namespace_0", "view_" + i))
                 .toList());
   }
+
+  @Test
```
snazy, line 219: You can possibly tweak the object-storage-mock to return failures to test.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/StorageFailureException.java
snazy: > We might eventually need a third TaskState - to handle "forbidden"/"access denied" better, to handle the case when the object store's ACL's just wrong: a state that's immediately reported as an error, but retried later.

Isn't that what `TaskStatus.ERROR_RETRY` already does?

snazy: To mitigate potentially unmapped/unhandled error states, WDYT about making all current `FAILURE` states `ERROR_RETRY` ones but with a somewhat higher retry interval?

snazy: > Non-retryable ones:    Not found

Actually, this feels also retryable. Sure, in a perfect world, "not found" should be a "final" state.
But it could also be an ephemeral case, for example when users messed up the object store locations - in theory though, but possible. Or if someone manually migrates files.

dimas-b: Yes, that's pretty much where I'm moving (in my local ATM). Each caller should timeout at some point and concurrent caller would get the same result within some time window. If another caller, touches the same task after a certain time period, a re-try will happen regardless of the root cause. We can tune with the first and second timeout separately for each "cause".

dimas-b: Re: "not found" : It may be in reality represent a permission error (e.g. in GH), so we cannot really rely on the object being really missing.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOOutputStream.java
dimas-b: streams are no longer wrapped

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: not used anymore

servers/quarkus-server/src/test/java/org/projectnessie/server/catalog/AbstractIcebergCatalogUnitTests.java
dimas-b: What do you mean? The object storage mock is already producing failures when the test callbacks instruct it to.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +70,52 @@ public interface TaskState {
...
+   * the deadline may or may not forcibly cancel a running task.
+   */
+  @Value.Parameter(order = 5)
+  @Nullable
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  @JsonDeserialize(using = InstantAsLongDeserializer.class)
+  @JsonSerialize(using = InstantAsLongSerializer.class)
+  Instant deadline();
+
+  TaskState SUCCESS = ImmutableTaskState.builder().status(TaskStatus.SUCCESS).build();
```
snazy, line 84: Interesting that this is not marked as a potential class-loader deadlock

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
```diff
@@ -114,11 +115,16 @@ <T extends TaskObj, B extends TaskObj.Builder> CompletionStage<T> submit(
...
     if (obj != null) {
       T taskObj = castObj(taskRequest, obj);
-      TaskStatus status = taskObj.taskState().status();
+      TaskState taskState = taskObj.taskState();
+      TaskStatus status = taskState.status();
       switch (status) {
         case FAILURE:
-          metrics.taskHasFinalFailure();
-          return failedStage(taskRequest.behavior().stateAsException(taskObj));
+          if (canRetryNow(taskState)) {
```
snazy, line 122: This change deserves a change in `org.projectnessie.nessie.tasks.api.TaskObjUtil#TASK_DEFAULT_CACHE_EXPIRE` as well. And likely a test case w/ the persist-cache enabled.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -20,6 +20,7 @@
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.base.Supplier;
```
snazy, line 23: Java `Supplier`?

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +70,52 @@ public interface TaskState {
...
 
   static TaskState successState() {
     return SUCCESS;
   }
 
-  static TaskState runningState(@Nonnull Instant retryNotBefore, @Nonnull Instant lostNotBefore) {
-    return ImmutableTaskState.of(TaskStatus.RUNNING, retryNotBefore, lostNotBefore, null);
+  private static Instant deadlineFor(@Nullable TaskState old, Supplier<Instant> deadline) {
+    if (old == null || old.deadline() == null) {
+      return deadline.get();
```
snazy, line 92: Do we really need a Supplier for this one?

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -98,19 +124,22 @@ default void check() {
       case SUCCESS:
         checkState(retryNotBefore() == null, "retryNotBefore must be null for SUCCESS");
         checkState(lostNotBefore() == null, "retryNotBefore must be null for SUCCESS");
+        checkState(deadline() == null, "deadline must be null for SUCCESS");
         break;
       case FAILURE:
-        checkState(retryNotBefore() == null, "retryNotBefore must be null for FAILURE");
```
snazy, line 103: Maybe just `!=` instead of removing?

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
```diff
@@ -107,22 +93,22 @@ private EntitySnapshotObj.Builder importIcebergTable(
...
       try {
         InputStream input = taskRequest.objectIO().readObject(metadataLocation);
         if (metadataLocation.requiredPath().endsWith(".gz")
             || metadataLocation.requiredPath().endsWith(".gz.metadata.json")) {
           input = new GZIPInputStream(input);
         }
         tableMetadata = IcebergJson.objectMapper().readValue(input, IcebergTableMetadata.class);
-      } catch (IOException e) {
-        throw new IOException(
+        table = entityObjForContent(content, tableMetadata, entityObjId);
```
snazy, line 106: Any specific reason why you moved this inside the try-block? (same for view below)

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/EntitySnapshotTaskBehavior.java
```diff
@@ -43,12 +51,17 @@ public Instant performRunningStateUpdateAt(Clock clock, EntitySnapshotObj runnin
     return clock.instant().plus(2, SECONDS);
   }
 
+  private TaskState taskState(@Nullable EntitySnapshotObj obj) {
+    return obj == null ? null : obj.taskState();
```
snazy, line 55: Nit: maybe just inline this one

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+      case 429:
+        return ObjectIOStatus.of(httpStatus, true, retryAfter(retryAfterThrottled()), th);
+
+      case 502:
+      case 503:
+      case 504:
+        return ObjectIOStatus.of(httpStatus, true, retryAfter(retryAfterNetworkError()), th);
+
+      default:
+        return ObjectIOStatus.of(httpStatus, false, retryAfter(reattemptAfterFetchError()), th);
```
snazy, line 73: Hm - should this also be retryable?

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+  protected abstract Duration retryAfterThrottled();
+
+  protected abstract Duration retryAfterNetworkError();
+
+  protected abstract Duration reattemptAfterFetchError();
+
+  protected abstract Clock clock();
+
+  public static ImmutableObjectIOExceptionMapper.Builder builder() {
```
snazy, line 39: Let's use a separate builder interface for this

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+    return Optional.empty();
+  }
+
+  private Instant retryAfter(Duration delay) {
+    return clock().instant().plus(delay);
+  }
+
+  private ObjectIOStatus toStatus(int httpStatus, Throwable th) {
+    switch (httpStatus) {
```
snazy, line 61: Isn't this mapping possibly specific for a particular object store?
BTW: some exception might not even include a http status code.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public static ImmutableObjectIOExceptionMapper.Builder builder() {
+    return ImmutableObjectIOExceptionMapper.builder();
+  }
+
+  public Optional<ObjectIOStatus> analyze(Throwable ex) {
+    for (Throwable th = ex; th != null; th = th.getCause()) {
+      for (Analyzer analyzer : analyzers()) {
+        OptionalInt statusCode = analyzer.httpStatusCode(th);
+        if (statusCode.isPresent()) {
```
snazy, line 47: A missing status code is probably some misconfiguration - like invalid credentials, wrong endpoint, http read timeout, so anything that can throw before a response is received.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
```diff
@@ -0,0 +1,80 @@
...
+    return Optional.empty();
+  }
+
+  private Instant retryAfter(Duration delay) {
+    return clock().instant().plus(delay);
+  }
+
+  private ObjectIOStatus toStatus(int httpStatus, Throwable th) {
+    switch (httpStatus) {
+      case 408:
```
snazy, line 62: Note: maybe `com.google.cloud.storage.StorageException#RETRYABLE_ERRORS` helps

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ExceptionMapper.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class S3ExceptionMapper implements ObjectIOExceptionMapper.Analyzer {
+
+  public static final S3ExceptionMapper INSTANCE = new S3ExceptionMapper();
+
+  private S3ExceptionMapper() {}
+
+  @Override
+  public OptionalInt httpStatusCode(Throwable th) {
+    if (th instanceof SdkServiceException) {
```
snazy, line 30: Wanna handle `SdkClientException` as well?

snazy (commented):
Not a full review. It's a lot of changes in this PR. It seems, your new approach is around handling the exception when running in the "import worker", feels appropriate to me. Just need to keep in mind that we have to eventually do the same when writing files.

I think, the approach can also be implemented w/ using the `ERROR_RETRY` and leave the `FAILURE` state as it is (the changes for that are incomplete BTW).

dimas-b:
> leave the FAILURE state as it is

I'm not sure about that. I think we have to be able to re-attempt after any failure because the conditions for the failure might change over time. In a way, no failure is a "final" failure. The different between `ERROR_RETRY` and `FAILURE` is only in failing the _currently_ waiting tasks (in the latter case).

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: Old tasks loaded from storage can have `null` values here.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: This is an `interface`, I suppose the JVM does not need to resolve all its static fields before it can use it.... In a way static fields on an `interfaces` are just global constants.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -662,24 +665,23 @@ static final class SingleTableUpdate {
...
     IcebergTableMetadata tableMetadata =
         nessieTableSnapshotToIceberg(snapshot, Optional.empty(), p -> {});
-    try (OutputStream out = objectIO.writeObject(StorageUri.of(metadataJsonLocation))) {
+    StorageUri uri = StorageUri.of(metadataJsonLocation);
+    try (OutputStream out = objectIO.writeObject(uri)) {
       IcebergJson.objectMapper().writeValue(out, tableMetadata);
-    } catch (IOException ex) {
-      throw new RuntimeException(ex);
+    } catch (Exception ex) {
+      throw new RuntimeException("Failed to write table snapshot to: " + uri, ex);
```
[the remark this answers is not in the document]
dimas-b: Sorry, I'm confused... Is `objectIO.writeObject` (line 681) not writing to the object store?

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: I'm thinking we should probably let the storage client (provider-specific) handle those retries.

So, on the Catalog side we can only do meaningful retries for "throttled" scenarios (under the same request `Future`) and for auth and "unknown" failures (under a new request).

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ExceptionMapper.java
dimas-b: `SdkClientException` can happen only before the storage service is contacted... Can we really do better than a "Server Failure" in this case?

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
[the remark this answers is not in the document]
snazy: Meh - I meant another place

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
snazy: I'm surprised, because there should actually no real difference.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ExceptionMapper.java
snazy: Sure, it can happen before. But IIUC that can also happen for misconfigured clients, and user configuration mistakes.

snazy:
> > leave the FAILURE state as it is
> 
> I'm not sure about that. I think we have to be able to re-attempt after any failure because the conditions for the failure might change over time. In a way, no failure is a "final" failure. The different between `ERROR_RETRY` and `FAILURE` is only in failing the _currently_ waiting tasks (in the latter case).

Well, `FAILURE` is well defined now as a final state. Then I'd rather add another type. For that it's probably easier to do it in a separate PR and only focus on the tasks-stuff.

dimas-b:
I've just rebased the previous code. I'll probably trim this PR to just the parts that analyse storage exceptions, but will move the task retry/re-attempt changes into a follow-up PR for ease of reviewing.

tasks/service/impl/src/main/java/org/projectnessie/nessie/tasks/service/impl/TasksServiceImpl.java
dimas-b: removed this from current PR... will follow-up separately.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: removed this from current PR... will follow-up separately.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: removed this from current PR... will follow-up separately.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: removed this from current PR... will follow-up separately.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/EntitySnapshotTaskBehavior.java
dimas-b: this method got removed

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: re-implemented

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: re-implemented

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: I can still move it to a separate class if you prefer... WDYT?

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
dimas-b: Just to wrap it with a `RuntimeException` mentioning the view/table name for `IcebergErrorMapper.message()` - makes user-level errors easier to interpret (and the new tests check for that).

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: done

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: Yes. Currently it will not get a status code and will be handled separately at the REST layer (HTTP 500) and at the Tasks layer ("unknown" status).

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIOExceptionMapper.java
dimas-b: refactored

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ExceptionMapper.java
dimas-b: In the latest code those exceptions will get the `UNKNOWN` status code and be represented as Internal Server Errors (500) at the REST layer.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendExceptionMapper.java
```diff
@@ -0,0 +1,75 @@
...
+  public Optional<BackendErrorStatus> analyze(Throwable ex) {
+    for (Throwable th = ex; th != null; th = th.getCause()) {
+      for (Analyzer analyzer : analyzers()) {
+        BackendErrorStatus status = analyzer.analyze(th);
+        if (status != null) {
+          if (status.statusCode() == BackendErrorCode.THROTTLED) {
+            return Optional.of(status.withReattemptAfter(retryAfter(retryAfterThrottled())));
+          }
+          return Optional.of(status);
+        }
```
adutra, line 48: Wouldn't it be better if the analyzer could construct the final `BackendErrorStatus`?

Something like:

```suggestion
        BackendErrorStatus status = analyzer.analyze(th, retryAfter(retryAfterThrottled()));
        if (status != null) {
          return Optional.of(status);
        }
```

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
```diff
@@ -174,6 +180,7 @@ public HttpClient adlsHttpClient(CatalogAdlsConfig adlsConfig) {
   @Produces
   @Singleton
   public ObjectIO objectIO(
+      BackendExceptionMapper exceptionMapper,
```
snazy, line 183: Looks unused

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +69,37 @@ public interface TaskState {
   @JsonInclude(JsonInclude.Include.NON_NULL)
   String message();
 
-  TaskState SUCCESS = ImmutableTaskState.of(TaskStatus.SUCCESS, null, null, null);
+  @Value.Parameter(order = 5)
+  @Nullable
+  @JsonInclude(JsonInclude.Include.NON_NULL)
+  String errorCode();
```
snazy, line 75: Mind adding a javadoc?

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorStatus.java
```diff
@@ -0,0 +1,57 @@
...
+  default BackendErrorStatus withReattemptAfter(Instant reattemptAfter) {
+    return ImmutableBackendErrorStatus.builder().from(this).reattemptAfter(reattemptAfter).build();
+  }
+
+  static BackendErrorStatus of(BackendErrorCode statusCode, boolean retryable, Throwable cause) {
+    // Use Instant.EPOCH for `reattemptAfter` here. The reattempt delay may be re-assigned later.
+    return ImmutableBackendErrorStatus.of(statusCode, retryable, Instant.EPOCH, cause);
+  }
+
+  static BackendErrorStatus of(int httpStatusCode, Throwable cause) {
```
snazy, line 45: Mind renaming this to 
```suggestion
  static BackendErrorStatus fromHttpStatusCode(int httpStatusCode, Throwable cause) {
```

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -146,7 +150,7 @@ public Stream<Supplier<CompletionStage<SnapshotResponse>>> retrieveSnapshots(
             .keys(keys)
             .getWithResponse();
 
-    IcebergStuff icebergStuff = new IcebergStuff(objectIO, persist, tasksService, executor);
+    IcebergStuff icebergStuff = icebergStuff();
```
snazy, line 153: Unrelated: Seems you start to like this weird name ;)

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
```diff
@@ -103,22 +89,22 @@ private EntitySnapshotObj.Builder importIcebergTable(
...
+      } catch (Exception e) {
```
snazy, line 103: Not sure about this (would exclude the location from the exception message), but as a suggestion.
```suggestion
      } catch (RuntimeException e) {
        throw e;
      } catch (Exception e) {
```

snazy (commented):
A few initial comments

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
dimas-b: `Iceberg` and `Stuff` those two word are enough to binary-encode anything ;)

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
dimas-b: If we do not report the file path to the client, the error on the client side looks kind of cryptic... something like `Forbidden: java.io.IOException: software.amazon.awssdk.services.s3.model.S3Exception: Access Denied. (Service: S3, Status Code: 403, Request ID: )`

With this change we can report it as `Forbidden: Failed to read table metadata from s3://bucket1/warehouse/ns/table1_06a7ebe4-7dc6-4526-a85e-578993e09a12_test_access_denied_file (due to: software.amazon.awssdk.services.s3.model.S3Exception: Access Denied. (Service: S3, Status Code: 403, Request ID: ))`.

I know S3 servers do not report paths in errors, however, in this case my rationale is that the user is in control of both  the credentials and the S3 objects. Nessie validates the user's credentials to access the catalog before we produce this error... so I reasoned that it should be fine to expose the extra info for ease of trouble-shooting. WDYT?

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
dimas-b: removed

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: added

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorStatus.java
dimas-b: renamed

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendExceptionMapper.java
dimas-b: Actually, I moved the delay setting to `EntitySnapshotTaskBehavior`, where it is in fact relevant.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
```diff
@@ -55,29 +57,34 @@ public class IcebergErrorMapper {
...
-      BaseNessieClientServerException e = (BaseNessieClientServerException) ex;
-      body = mapNessieError(e, e.getErrorCode(), e.getErrorDetails(), kind);
-    } else if (ex instanceof IllegalArgumentException) {
-      body = errorResponse(400, "IllegalArgumentException", ex.getMessage(), ex);
-    } else if (ex instanceof IcebergException) {
-      body = ((IcebergException) ex).toErrorResponse();
+    Optional<BackendErrorStatus> status = ioExceptionMapper.analyze(ex);
+    if (status.isPresent()) {
+      body = mapStorageFailure(status.get(), ex);
     }
```
adutra, line 88:
```suggestion
    if (body == null) {
      Optional<BackendErrorStatus> status = ioExceptionMapper.analyze(ex);
      if (status.isPresent()) {
        body = mapStorageFailure(status.get(), ex);
      }
    }
```

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
```diff
@@ -89,6 +96,30 @@ public Response toResponse(Throwable ex, IcebergEntityKind kind) {
...
+    return String.format("%s (due to: %s)", ex.getMessage(), status.cause().toString());
+  }
+
+  private IcebergErrorResponse mapStorageFailure(BackendErrorStatus status, Throwable ex) {
+    // Log full stack trace on the server side for troubleshooting
+    LOGGER.info("Propagating storage failure to client: {}", status, ex);
+
+    switch (status.statusCode()) {
+      case THROTTLED:
+        return errorResponse(429, "RuntimeException", message(status, ex), ex);
```
adutra, line 109: Does `type` need to be the name of an existing exception? I think `"TooManyRequests"` would be more explicit.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
```diff
@@ -89,6 +96,30 @@ public Response toResponse(Throwable ex, IcebergEntityKind kind) {
...
+      case UNAUTHORIZED:
+        return errorResponse(401, "NotAuthorizedException", message(status, ex), ex);
+      case FORBIDDEN:
+        return errorResponse(403, "ForbiddenException", message(status, ex), ex);
+      case NOT_FOUND:
+        // Convert storage-side "not found" into `IllegalArgumentException`.
+        // In most cases this results from bad locations in Iceberg metadata.
+        return errorResponse(400, "IllegalArgumentException", message(status, ex), ex);
+      default:
+        return null;
```
adutra, line 119:
```suggestion
      case UNKNOWN:
        return null;
      default:
        throw new IllegalStateException("Unexpected status: " + status);
```

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: It's inconsistent on the Iceberg client side, IIRC, but many call paths use java exception names, indeed.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: Throwing from an exception mapper will probably result in a cryptic client-side error :thinking: 

I'll change the `default` to 500 mentioning the unknown enum value in the message.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: good point - fixed.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
snazy: Seems to be fine

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ExceptionMapper.java
snazy: We need HTTP 600 error codes - server failed because of user configuration mistake :troll:

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
snazy: SGTM

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
snazy: To add some complications: a bunch of Iceberg tests assert on the names

snazy: (Not in this case though)

snazy: Maybe put in `TooManyRequests` there in this case.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +69,47 @@ public interface TaskState {
   @JsonInclude(JsonInclude.Include.NON_NULL)
   String message();
 
-  TaskState SUCCESS = ImmutableTaskState.of(TaskStatus.SUCCESS, null, null, null);
+  /**
+   * An error code associated with this task state. Error code are meant to be interpreted by the
```
snazy, line 73:
```suggestion
   * An error code associated with this task state. Error codes are meant to be interpreted by the
```

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +69,47 @@ public interface TaskState {
...
+   * {@link TaskBehavior} implementation that manages the related task.
+   *
+   * <p>In general, error codes are intended to allow reporting task errors / exceptions to the
```
snazy, line 76:
```suggestion
   * <p>Error codes are machine readable representations, intended to allow reporting task errors / exceptions to the
```

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
```diff
@@ -69,27 +69,47 @@ public interface TaskState {
...
+   * caller the same way whether the task fails for the first time in the server that executed it,
+   * or it is reloaded from storage in another server.
```
snazy, line 78:
```suggestion
   * or it is reloaded from storage in another server. The value of {@link #message()} is meant to be interpreted by humans.
```

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorStatus.java
```diff
@@ -0,0 +1,47 @@
...
+  BackendErrorCode statusCode();
+
+  Throwable cause();
+
+  static BackendErrorStatus of(BackendErrorCode statusCode, Throwable cause) {
+    return ImmutableBackendErrorStatus.of(statusCode, cause);
+  }
+
+  static BackendErrorStatus fromHttpStatusCode(int httpStatusCode, Throwable cause) {
+    switch (httpStatusCode) {
```
snazy, line 36: Do we need 429 here as well?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
```diff
@@ -79,16 +68,8 @@ public InputStream readObject(StorageUri uri) throws IOException {
...
-        throw new BackendThrottledException(
-            clock
-                .instant()
-                .plus(s3clientSupplier.s3config().retryAfter().orElse(Duration.of(10, SECONDS))),
-            "S3 throttled",
-            e);
-      }
-      throw new NonRetryableException(e);
+    } catch (Exception e) {
+      throw new IOException(e);
```
snazy, line 72:
```suggestion
      // All AWS `SdkException`s are `RuntimeException`s
      throw new IOException(e);
```

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
```diff
@@ -201,6 +207,24 @@ public RequestSigner signer(
...
+  @Produces
+  @Singleton
+  public EntitySnapshotTaskBehavior entitySnapshotTaskBehavior(
+      CatalogServiceConfig config, BackendExceptionMapper mapper) {
+    return new EntitySnapshotTaskBehavior(mapper, config.retryAfterThrottled());
+  }
+
+  @Produces
+  @Singleton
+  public BackendExceptionMapper objectIOExceptionMapper() {
```
snazy, line 219:
```suggestion
  public BackendExceptionMapper backendExceptionMapper() {
```

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogServiceConfig.java
```diff
@@ -22,6 +22,14 @@
 
 @ConfigMapping(prefix = "nessie.catalog.service")
 public interface CatalogServiceConfig {
+  /**
+   * Interval after which a request is retried when storage I/O responds with some "retry later"
+   * response.
+   */
+  @WithName("throttled-retry-after")
+  @WithDefault("PT10S")
+  Duration retryAfterThrottled();
```
snazy, line 31: If this is moved to `CatalogConfig` (and `QuarkusCatalogConfig`), `EntitySnapshotTaskBehavior` doesn't need to be produced in `CatalogProducers` can can stay package-private.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogServiceConfig.java
```diff
@@ -22,6 +22,14 @@
 
 @ConfigMapping(prefix = "nessie.catalog.service")
 public interface CatalogServiceConfig {
+  /**
+   * Interval after which a request is retried when storage I/O responds with some "retry later"
+   * response.
+   */
+  @WithName("throttled-retry-after")
```
snazy, line 29:
```suggestion
  @WithName("error-handling.throttled-retry-after")
```
So we've got more flexibility to add e.g. exponential backoff/jitter in the future.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
```diff
@@ -55,29 +57,36 @@ public class IcebergErrorMapper {
...
 
   public Response toResponse(Throwable ex, IcebergEntityKind kind) {
-    if (ex instanceof CompletionException) {
-      ex = ex.getCause();
-    }
-
-    if (ex.getClass() == RuntimeException.class && ex.getCause() != null) {
-      ex = ex.getCause();
+    IcebergErrorResponse body = null;
+    for (Throwable th = ex; th != null; th = th.getCause()) {
```
snazy, line 71: In case of a backend exception we traverse the cause-chain twice: here and in `BackendExceptionMapper.analyze()`.

WDYT of having this in `BackendExceptionMapper`:
```java

  public Optional<BackendErrorStatus> analyze(Throwable ex) {
    for (Throwable th = ex; th != null; th = th.getCause()) {
      Optional<BackendErrorStatus> status = analyzeNoChain(th);
      if (status.isPresent()) {
        return status;
      }
    }

    return Optional.empty();
  }

  public Optional<BackendErrorStatus> analyzeNoChain(Throwable ex) {
    for (Analyzer analyzer : analyzers()) {
      BackendErrorStatus status = analyzer.analyze(ex);
      if (status != null) {
        return Optional.of(status);
      }
    }

    return Optional.empty();
  }
```
and changing this function to
```java
  public Response toResponse(Throwable ex, IcebergEntityKind kind) {
    IcebergErrorResponse body = null;
    for (Throwable th = ex; th != null; th = th.getCause()) {
      if (th instanceof BaseNessieClientServerException) {
        BaseNessieClientServerException e = (BaseNessieClientServerException) th;
        body = mapNessieError(e, e.getErrorCode(), e.getErrorDetails(), kind);
        break;
      } else if (th instanceof IllegalArgumentException) {
        body = errorResponse(400, "IllegalArgumentException", th.getMessage(), th);
        break;
      } else if (th instanceof IcebergException) {
        body = ((IcebergException) th).toErrorResponse();
        break;
      }
      Optional<BackendErrorStatus> status = backendExceptionMapper.analyzeNoChain(ex);
      if (status.isPresent()) {
        body = mapStorageFailure(status.get(), ex);
        break;
      }
    }

    if (body == null) {
      LOGGER.warn("Unhandled exception returned as HTTP/500: {}", ex, ex);
      body = errorResponse(500, ex.getClass().getSimpleName(), ex.getMessage(), ex);
    }

    Integer code = body.error().code();
    return Response.status(code == null ? 500 : code).entity(body).build();
  }
```

snazy (commented):
Sorry for the late review, added a few comments. But overall it looks good.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorStatus.java
dimas-b: I can add it, but currently it'll probably be a dead branch. I'd like to use the storage client library's retry capabilities when we can (follow-up PR).

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: Yes, I was thinking about that too... will do.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogServiceConfig.java
dimas-b: renamed

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogS3Config.java
```diff
@@ -39,10 +38,6 @@
  */
 @ConfigMapping(prefix = "nessie.catalog.service.s3")
 public interface CatalogS3Config extends S3Config, S3Options<CatalogS3BucketConfig> {
-  @WithName("throttled-retry-after")
```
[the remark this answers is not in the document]
dimas-b: added to `breaking changes`

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: updated

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: I did not find any handlers for 429 in the REST Catalog client, so I guess it's fine to use any error `type` here. I changed to `TooManyRequestsException` to be in-line with the other cases.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorStatus.java
dimas-b: Actually, I added it for the sake of completeness.

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogServiceConfig.java
dimas-b: Would you like this config to be moved to `CatalogConfig`?

My impressions is that `CatalogConfig` contains user-level settings, but `throttled-retry-after` is more like an admin/service level setting (which is why I put it into `CatalogServiceConfig`). WDYT?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergErrorMapper.java
dimas-b: Implemented... slightly differently :)

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/CatalogProducers.java
dimas-b: renamed

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
dimas-b: I removed `IOException` wrappers completely.

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: updated

tasks/api/src/main/java/org/projectnessie/nessie/tasks/api/TaskState.java
dimas-b: updated

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogServiceConfig.java
snazy: `CatalogServiceConfig` is rather Quarkus specific, while `CatalogConfig` is "generic".

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/EntitySnapshotTaskBehavior.java
```diff
@@ -61,8 +71,27 @@ public ObjType objType() {
     return EntitySnapshotObj.OBJ_TYPE;
   }
 
+  private boolean isRetryable(BackendErrorStatus status) {
+    return THROTTLED == status.statusCode();
+  }
+
```
snazy, line 77:
```suggestion
```

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/EntitySnapshotTaskBehavior.java
```diff
@@ -61,8 +71,27 @@ public ObjType objType() {
...
   @Override
   public TaskState asErrorTaskState(Clock clock, EntitySnapshotObj base, Throwable t) {
-    return throwableAsErrorTaskState(t);
+    return exceptionMapper
+        .analyze(t)
+        .map(
+            status -> {
+              if (isRetryable(status)) {
```
snazy, line 88:
```suggestion
              if (status.statusCode().retryable()) {
```

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorCode.java
```diff
@@ -15,24 +15,13 @@
...
-  public abstract Optional<Instant> retryNotBefore();
+public enum BackendErrorCode {
+  UNKNOWN,
+  THROTTLED,
+  NOT_FOUND,
+  UNAUTHORIZED,
+  FORBIDDEN,
+  BAD_REQUEST,
+  NESSIE_ERROR,
+  ICEBERG_ERROR,
```
snazy, line 26:
```suggestion
public enum BackendErrorCode {
  /** Unknown errors are retryable, because we do not know the exact reason, only that we do not know the code. */
  UNKNOWN(true),
  THROTTLED(true),
  NOT_FOUND(false),
  UNAUTHORIZED(false),
  FORBIDDEN(false),
  BAD_REQUEST(false),
  NESSIE_ERROR(false),
  ICEBERG_ERROR(false),
  ;

  private final boolean retryable;

  BackendErrorCode(boolean retryable) {
    this.retryable = retryable;
  }

  public boolean retryable() {
    return retryable;
  }
```

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/PreviousTaskException.java
```diff
@@ -0,0 +1,51 @@
...
+
+    if (state.errorCode() == null) {
+      return new PreviousTaskException(UNKNOWN, state.message());
+    }
+
+    try {
+      BackendErrorCode errorCode = BackendErrorCode.valueOf(state.errorCode());
+      return new PreviousTaskException(errorCode, state.message());
+    } catch (IllegalArgumentException e) {
+      return new PreviousTaskException(UNKNOWN, state.message());
```
snazy, line 48: I wonder how far we should go here wrt to new `BackendErrorCode` values added in future releases, and behavior during a rolling-upgrade (new version writing `NEW_CODE` but old version reading it as `UNKNOWN`, writing it as `UNKNOWN` and new version re-reading it as `UNKNOWN` as well). Would it make sense to serialize the error code including the HTTP status code or to make `BackendErrorCode` a value type?

snazy (commented):
Basically two things left:
* Moving the config property (not a big one)
* Compatibility/behavior during rolling upgrades

Otherwise LGTM

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/PreviousTaskException.java
snazy: I see. All good then.

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorCode.java
dimas-b: I moved the `retryable` flag here, but I kept it `false` for `UNKNOWN` - rationale in code comments.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/PreviousTaskException.java
dimas-b: Approve? ;)

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorCode.java
```diff
@@ -15,29 +15,29 @@
...
 
-import java.time.Instant;
-import java.util.Optional;
+public enum BackendErrorCode {
+  /**
+   * Treat unknown errors as non-retryable to prevent indefinite retry cycles when the server is not
+   * able to determine a specific failure reason. This may be changed in the future when task
+   * deadlines are supported.
+   */
+  UNKNOWN(false),
```
snazy, line 24: Let's make `UNKNOWN` retryable - we don't know what happened, so it can be retryable.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
```diff
@@ -32,43 +32,28 @@
...
+  public void ping(StorageUri uri) {
     AdlsLocation location = adlsLocation(uri);
 
     DataLakeFileSystemClient fileSystem = clientSupplier.fileSystemClient(location);
-    try {
-      fileSystem.getAccessPolicy();
-    } catch (Exception e) {
-      throw new IOException(e);
-    }
+    fileSystem.getAccessPolicy();
```
snazy, line 48: Can you sneak in a change to `getProperties()` - see [here](https://github.com/projectnessie/nessie/pull/9086#discussion_r1679410901)

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/BackendErrorCode.java
dimas-b: I know real `UNKNOWN` will end up in infinite retry

dimas-b: I'm planning to add tast deadlines in a follow-up PR and also fix #8860

snazy (approved):

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
dimas-b: #9134

PR #8564, opened 2024-05-22
[Catalog] Remove OAuth proxy

adutra:


snazy (commented):
Just a comment about providing users some guidance with the error message.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -123,6 +124,19 @@ public IcebergS3SignResponse s3sign(
...
 
+  @POST
+  @Path("/v1/oauth/tokens")
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+  @PermitAll
+  public Response getToken() {
+    return Response.status(Response.Status.SERVICE_UNAVAILABLE)
+        .entity(
+            Map.of(
+                "error", "OAuthTokenEndpointUnavailable",
```
dimas-b, line 135: Why not give a response in `IcebergApiV1AuthResource`?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -123,6 +124,19 @@ public IcebergS3SignResponse s3sign(
     return icebergS3SignResponse(signed.uri().toString(), signed.headers());
   }
 
+  @POST
+  @Path("/v1/oauth/tokens")
+  @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
+  @PermitAll
+  public Response getToken() {
+    return Response.status(Response.Status.SERVICE_UNAVAILABLE)
```
dimas-b, line 132: 503 is supposed to be transient... how about 501 (Not Implemented)?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
adutra: That file is deleted. The endpoint is now so dumb that I figured it didn't make sense to put it in its own file.

snazy (dismissed):

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
dimas-b: but will this method be called for other (sub-) paths?

dimas-b: Should we allow this to match `:ref/v1/oauth/tokens`?

dimas-b (commented):
LGTM, but I wonder whether we can/should send 501 for `":ref/v1/oauth/tokens` too.

adutra:
> LGTM, but I wonder whether we can/should send 501 for `":ref/v1/oauth/tokens` too.

OK I think you have a point, let's add that endpoint.

dimas-b (approved):

PR #8577, opened 2024-05-23
Blog post: support for MariaDB and MySQL backends

adutra:


site/docs/blog/posts/2024-05-mariadb-mysql-support.md
```diff
@@ -0,0 +1,162 @@
...
+but here is a quick step-by-step guide:
+
+1. Select the `JDBC` version store type as usual:
+
+```properties
+nessie.version.store.type=JDBC
+```
+
+2. Declare the JDBC datasource to use as `mariadb` (_this should be done regardless of whether you
+   are using MariaDB or MySQL_):
```
dimas-b, line 44: Could we have `mysql` as an alias for user convenience? I imagine MySQL users will tend to think about their datasource as "MySQL" in daily activities... The JDBC URL will remain `jdbc:mariadb:`, of course, but config-wise it might be easier for users to manage ` quarkus.datasource.mysql.username=my_user`. WDYT?

dimas-b (approved):
Nice blog post :+1: Just one minor usability comment / question.

site/docs/blog/posts/2024-05-mariadb-mysql-support.md
adutra: Interesting idea! Yes, I think we can add a 3rd datasource `quarkus.datasource.mysql`, its `db-kind` must be `mariadb` of course, but it could make the user XP slightly better indeed. Besides, if there are specificities for MySQL, these could easily go under `quarkus.datasource.mysql.jdbc.additional-jdbc-properties` without affecting MariaDB connections.

Also, it _might_ be possible to create a `ConfigSourceInterceptor` that would transparently change `jdbc:mysql` to `jdbc:mariadb` in all JDBC urls.

snazy (approved):

PR #8592, opened 2024-05-23
Exclude `CommitObj.created()` from `.equals()` and `.hashCode()`

dimas-b:
This is to simplify reasoning about `CommitObj` since the value of `created()` was never part of its "commit hash", therefore it was never considered for equality comparisons at the storage layer.

* Remove the rebuilding of `CommitObj` in `mitigateHashCollision()`, which is redundant now.

* Fix the handling of fast-forward transplants by comparing source and transplanted commits vs. transplanted in-memory and transplanted in storage.

### DIFF
--- a/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/CommitObj.java
+++ b/versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/CommitObj.java
@@ -137,7 +137,12 @@ public interface CommitObj extends Obj {
     CommitObj build();
   }
 
-  /** Creation timestamp in microseconds since epoch. */
+  /**
+   * Creation timestamp in microseconds since epoch.
+   *
+   * <p>Note that this value is not considered during Nessie commit hash value computations.
+   */
+  @Value.Auxiliary
   long created();
 
   /**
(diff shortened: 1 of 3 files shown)


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/CommitObj.java
```diff
@@ -137,7 +137,12 @@ interface Builder {
     CommitObj build();
   }
 
-  /** Creation timestamp in microseconds since epoch. */
+  /**
+   * Creation timestamp in microseconds since epoch.
+   *
+   * <p>Note that this value is not considered during Nessie commit hash value computations.
```
adutra, line 143: Maybe worth mentioning here, because I find it sometimes confusing: this timestamp (supposedly in microseconds, but that depends on the underlying clock) is indeed not included in the hash. 

However, most commits (maybe all?) do include a header containing the so-called "commit time", coming from the commit meta: see `TypeMapping#fromCommitMeta`. That timestamp is a vanilla `Instant.now()` – so accuracy is more or less 1 ms depending on the OS.

The corollary is that for 2 commits to produce the same hash, they'd need to contain the same data incl. parents, and would need to be generated in the same "moment" – roughly, the same millisecond.

dimas-b: Yep, I'll update the javadoc(s) after I resolve the related CI failures.

snazy: Note that both timestamps can differ a lot. One is the commit-timestamp the other the timestamp when the commit was persisted (think: repo imports).

snazy:
Had to add this to #9401 anyway :(

snazy:
Superseded by #9401

PR #8596, opened 2024-05-24
Split Release workflow job into multiple, retryable jobs

### LINKED ISSUE #8581: Split release workflow into multiple jobs
If a release "target" (e.g. Sonatype, ghcr, Helm, etc) fails, the whole release fails.

Splitting the single `publish-release` in `.github/workflows/release-publish.yml` into one job per "target" should allow re-runs of the failed jobs.

This should eliminate the need to do a full re-release in case of a single "target" failure / half-released version.

snazy:
The Nessie release for 0.83 failed two times, we had to re-release as 0.83.1 and finally 0.83.2, beacuse the release-publish workflow is not retryable, mostly because multiple publications of Maven artifacts of the same versions are not possible.

This change splits the single workflow job into multiple:

```
Prepare -->--+
             |
             +-->-- Publish to Maven Central -->--+
             |                                    |
             +-->-- Publish OpenAPI ----------->--+
             |                                    |
             +-->-- Publish Docker images ----->--+
             |                                    |
             +-->-- Publish Helm Chart -------->--+
                                                  |
                                                  +-->-- GitHub Release notes
```

This change also moves the big script to create the GH release notes to a separate shell script.

Fixes #8581


adutra (approved):

XN137 (approved):
not easy to review the diff but nothing stood out as wrong/missed.

wondering though:
if any of the sub-jobs is eventually not retry-able for whatever reason, do we end up with a "worse" release state compared to previously (i.e. partially published artifacts) or is the result equally bad as the "single flow" we had before?
(and we dont advertise that version as usable)

snazy:
> if any of the sub-jobs is eventually not retry-able for whatever reason, do we end up with a "worse" release state compared to previously (i.e. partially published artifacts) or is the result equally bad as the "single flow" we had before? 

I don't think it's worse than before. The only non-retryable part is the Maven publication. We already had a few "failed releases", where only some targets succeeded.

PR #7043, opened 2023-06-14
[FEATURE] TBD

snazy:
PR for ongoing work for "Nessie Catalog Server"

PR #8600, opened 2024-05-24
[Catalog] Refactor secrets handling

snazy:
The handling of secrets is a bit awkward at the moment. We declare "IDs" on the objects, which are to be looked up via the `SecretsProvider`, which is currently just a `Map<String, String>`.

This change changes the behavior: The (default) behavior to lookup via another map is gone.

Secrets are now generally looked up in the following order:
1. If the value is defined on the "per bucket" config, it is used as is.
2. If a secret for the bucket can be resolved, it is used.
3. If the value is defined on the "default" config, it is used as is.
4. If a secret for the "default" config can be resolved, it is used.
5. The secret is assumed to not exist.

The above logic is implemented in the new `SecretsHelper` class. It calls the `SecretProvider` using a collection of keys to look up, which shall return a map with the existing, resolved secrets.

Follow-up, after #7043 is merged to `main`: refactor secrets handlings in "Nessie Core", where/if necessary.

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsOptions.java
@@ -15,8 +15,17 @@
  */
 package org.projectnessie.catalog.files.gcs;
 
+import static org.projectnessie.catalog.files.secrets.SecretsHelper.Specializeable.specializable;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import org.projectnessie.catalog.files.gcs.GcsProgrammaticOptions.GcsPerBucketOptions;
+import org.projectnessie.catalog.files.secrets.SecretsHelper;
+import org.projectnessie.catalog.files.secrets.SecretsProvider;
 import org.projectnessie.nessie.docgen.annotations.ConfigDocs.ConfigPropertyName;
 
 public interface GcsOptions<PER_BUCKET extends GcsBucketOptions> extends GcsBucketOptions {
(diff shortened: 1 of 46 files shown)


snazy:
The central piece of this PR is to remove the "artificial indirection" - how we later really resolve the secrets is a 2ndary concern IMO (aka: not used yet, so it can be changed ;) )

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/secrets/SecretsHelper.java
```diff
@@ -0,0 +1,117 @@
...
+   * <ol>
+   *   <li>The value of the attribute on the {@code specific} object,
+   *   <li>The resolved secret for the specific object (lookup via {@code baseName + '.' +
+   *       specificName + '.' + Specializeable.name()},
+   *   <li>The value of the attribute on the {@code base} object,
+   *   <li>The resolved secret for the base object (lookup via {@code baseName + '.' +
+   *       Specializeable.name()},
+   * </ol>
+   */
+  public static <R, B> B resolveSecrets(
```
adutra, line 49: Hard to follow, even with the javadocs above. Looks slightly over-engineered, and I wonder if this will play nicely with Quarkus own secrets management: will this work with the Quarkus Vault extension? Or with the Quarkus Kubernetes Config extension?

adutra: Also: the return value seems ignored everywhere.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsOptions.java
```diff
@@ -28,15 +37,52 @@ public interface GcsOptions<PER_BUCKET extends GcsBucketOptions> extends GcsBuck
...
+    GcsBucketOptions perBucket = buckets().get(name);
     if (perBucket == null) {
-      return this;
+      return resolveSecrets(name, null, secretsProvider);
+    }
+
+    return resolveSecrets(name, perBucket, secretsProvider);
+  }
+
+  Set<String> SECRET_NAMES =
```
adutra, line 54: Unused.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/secrets/SecretsProvider.java
```diff
@@ -15,6 +15,23 @@
...
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
 public interface SecretsProvider {
-  String getSecret(String secret);
+  Map<String, String> resolveSecrets(Set<String> names);
+
+  static SecretsProvider mapSecretsProvider(Map<String, String> secretsMap) {
```
adutra, line 25: Seems only used in tests.

adutra (approved):
Approving because I don't want to hold you any longer on this, but I find the overall design a bit convoluted.

adutra:
Heads up: the file `docker/catalog-full/docker-compose.yml` needs to be changed accordingly.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/secrets/SecretsHelper.java
```diff
@@ -0,0 +1,117 @@
...
+          secret.applicator().accept(builder, resolved);
+        }
+      }
+    }
+
+    return builder;
+  }
+
+  @NessieImmutable
+  public interface Specializeable<R, B> {
```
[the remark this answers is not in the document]
snazy: I'm not sure (yet).

The only secrets-extension that can be "neatly" used in config files is the one for GCS. The other ones (AWS + Vault) need to be accessed programmatically, which is sad.

I was looking for a way to declare the "secrets options" but have a central place for the resolve-secrets-logic. Not yet sure how to integrate these three different secrets sources - especially wrt planned rotation and unplanned rotation (aka how to cache those).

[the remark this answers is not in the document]
snazy: But that's likely for a different PR - I've started a WIP branch for that though.

adutra (approved):

PR #8604, opened 2024-05-26
[Catalog] Bearer auth in Spark extensions

snazy:


integrations/spark-extensions-base/src/main/java/org/apache/spark/sql/execution/datasources/v2/RestCatalogBridge.java
```diff
@@ -93,9 +100,17 @@ final class RestCatalogBridge implements CatalogBridge {
               return CatalogUtils.resolveOAuthScope(catalogProperties);
               // TODO need the "token" (initial bearer token for OAuth2 as in
```
adutra, line 101: We can remove this comment now.

adutra (approved):

adutra (approved):

PR #8612, opened 2024-05-27
Add `whoami` to Nessie REST API

snazy:
Basic functionality to retrieve information about the current user, helpful for development purposes.

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/NessieUserInfo.java
+++ b/api/model/src/main/java/org/projectnessie/model/NessieUserInfo.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import java.util.List;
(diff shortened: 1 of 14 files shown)


api/model/src/main/java/org/projectnessie/model/NessieUserInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  @Value.Parameter(order = 2)
+  @Nullable
+  @jakarta.annotation.Nullable
+  String name();
+
+  @Value.Parameter(order = 3)
+  List<String> roles();
+
+  static NessieUserInfo nessieUserInfo(boolean anonymous, String name, List<String> roles) {
+    return null; // ImmutableNessieUserInfo.of(anonymous, name, roles);
```
adutra, line 46: Is this a temporary workaround?

snazy: Oops, debug left-over

api/model/src/main/java/org/projectnessie/model/NessieUserInfo.java
```diff
@@ -0,0 +1,68 @@
...
+  @Value.Parameter(order = 1)
+  boolean anonymous();
+
+  @Value.Parameter(order = 2)
+  @Nullable
+  @jakarta.annotation.Nullable
+  String name();
+
+  @Value.Parameter(order = 3)
+  List<String> roles();
```
dimas-b, line 43: Do we have to expose roles? In general, I tend to think the less auth info we expose, the better.

snazy: How does this expose "more" than you could get yourself from an OAuth2 server anyway?

dimas-b: The Nessie server may have more privileges for introspecting users than the actual user.

dimas-b: In other words, this could (potentially) be a way for the (malicious) user to discover something about the authenticated identity that is not meant to be available by other means.... this is potential, of course, but if we do not have to expose this information, I think we ought not to expose it.

snazy: But again: how is that different from any other way to authenticate and ask the oauth2 server? Or how Linux' `id` command shows all the groups? Nothing says that the server _has to_ return all roles - it _can_.

PR #8619, opened 2024-05-27
Docker compose: add example with Authelia

adutra:


docker/authn-authelia/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# To request an access token, use the following command:
+# token=$(curl http://127.0.0.1:9091/api/oidc/token --user nessie-client:s3cr3t -d 'grant_type=client_credentials' -d 'scope=profile' | jq -r .access_token)
```
dimas-b, line 18: Should this be a token for "Alice"?

adutra: No, Alice is a user, here we need a client – don't get fooled by the name of the curl parameter ;-) In OAuth terms, it's the client that must authenticate.

adutra: In fact, the user Alice has to be created, otherwise Authelia won't start. But it's not used anywhere.

dimas-b (approved):

PR #8629, opened 2024-05-28
Remove usages of `*.nip.io` lookups

snazy:
The current code base uses host/domain names like `s3.127.0.0.1.nio.ip` for testing purposes. Current routers however have have DNS rebind protection enabled (aka: public domains cannot re-bind to local addresses) - and we do not want to force people to disable that.

For Linux, host/domain names like `s3.localhost` or `foo.bar.localhost.localdomain` work fine. Sadly, that does not work for macOS - only the "plain" `localhost` works on macOS.

### DIFF
--- a/api/client/src/intTest/resources/org/projectnessie/client/auth/oauth2/authelia-config.yaml
+++ b/api/client/src/intTest/resources/org/projectnessie/client/auth/oauth2/authelia-config.yaml
@@ -30,8 +30,8 @@ session:
   secret: 'insecure_session_secret'
   cookies:
     - name: 'authelia_session'
-      domain: 'authelia.127.0.0.1.nip.io'
-      authelia_url: 'https://authelia.127.0.0.1.nip.io'
+      domain: 'localhost'
+      authelia_url: 'https://localhost'
 
 storage:
   encryption_key: 'you_must_generate_a_random_string_of_more_than_twenty_chars_and_configure_this'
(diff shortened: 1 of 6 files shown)


api/client/src/intTest/resources/org/projectnessie/client/auth/oauth2/authelia-config.yaml
```diff
@@ -30,8 +30,8 @@ session:
   secret: 'insecure_session_secret'
   cookies:
     - name: 'authelia_session'
-      domain: 'authelia.127.0.0.1.nip.io'
-      authelia_url: 'https://authelia.127.0.0.1.nip.io'
+      domain: 'localhost'
```
adutra, line 33: Error:

```
Configuration: session: domain config #1 (domain 'localhost'): option 'domain' is not a valid cookie domain: must have at least a single period or be an ip address
```

docker/authn-authelia/config/configuration.yml
```diff
@@ -30,8 +30,8 @@ session:
   secret: 'insecure_session_secret'
   cookies:
     - name: 'authelia_session'
-      domain: 'authelia.127.0.0.1.nip.io'
-      authelia_url: 'https://authelia.127.0.0.1.nip.io'
+      domain: 'localhost'
```
adutra, line 33: This doesn't work, try `127.0.0.1` or `localhost.localdomain`

adutra (commented):
I'm a bit on the fence about nuking nip.io completely, can we disable them and add an option to enable them locally? Otherwise I won't be able to run any of the minio tests locally.

snazy:
I've disabled the Authelia integration test for now.

nip.io cannot be used due to the DNS rebinding protection, which is a good thing to have.

snazy:
> Otherwise I won't be able to run any of the minio tests locally.

Would need to find another solution then - or only run a subset of those on macOS - those that do not rely on S3 "compatible" host names.

PR #8630, opened 2024-05-28
[Catalog] Remove nip.io

snazy:
Related to #8629


servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/ObjectStorageMockTestResourceLifecycleManager.java
```diff
@@ -37,10 +37,7 @@ public class ObjectStorageMockTestResourceLifecycleManager
   public static final String INIT_ADDRESS =
       "ObjectStorageMockTestResourceLifecycleManager.initAddress";
 
-  public static final String S3_INIT_ADDRESS =
```
adutra, line 40: On my machine:

```
> Task :nessie-quarkus:test

1119 tests completed, 104 failed, 448 skipped
```

adutra: In fact, just `localhost` seems to work fine on macOS. I suggest to remove this field completely, remove the `init()` method and just use `localhost` all the time, if that works on linux as well.

PR #8633, opened 2024-05-28
[Catalog] Implement paging for Iceberg REST

snazy:


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -120,6 +120,8 @@ public Map<String, String> icebergConfigOverrides(String reference, String wareh
     Map<String, String> config = new HashMap<>();
     config.putAll(uriInfo.icebergConfigOverrides());
     config.putAll(storeConfigOverrides(StorageUri.of(warehouseConfig.location())));
+    // allow users to override the 'rest-page-size' in the Nessie configuration
+    config.put("rest-page-size", "200");
```
adutra, line 124: This is a config override, not a config default. Users won't be able to change this value. Is that the intent?

snazy: Yea - that should be in "defaults"

adutra (approved):

PR #8641, opened 2024-05-28
Honor set-location from a staged table create

adutra:


adutra:
Closing in favor of #8646

PR #8650, opened 2024-05-29
Deprecate support for Java 8

snazy:
... and remove testing overhead for Java 8.


adutra (commented):
Don't we need to change `nessie-conventions-client.gradle.kts` as well?

adutra (approved):

PR #8654, opened 2024-05-29
[Catalog] Restrict S3 signing to URIs within table base prefix

### LINKED ISSUE #8739: [Catalog] Security: Harden s3-signing endpoint

adutra:
S3 signing is now only possible within the table base location.

Fixes #8739

### DIFF
--- a/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
+++ b/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
@@ -15,11 +15,11 @@
  */
 package org.projectnessie.catalog.service.impl;
 
-import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.CompletableFuture.completedStage;
+import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.icebergBaseLocation;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.icebergMetadataJsonLocation;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.icebergMetadataToContent;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.nessieTableSnapshotToIceberg;
@@ -27,6 +27,8 @@ import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceber
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.newIcebergTableSnapshot;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.newIcebergViewSnapshot;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.typeToEntityName;
+import static org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate.SetProperties.setProperties;
+import static org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate.SetTrustedLocation.setTrustedLocation;
 import static org.projectnessie.catalog.service.api.NessieSnapshotResponse.nessieSnapshotResponse;
 import static org.projectnessie.catalog.service.impl.Util.objIdToNessieId;
 import static org.projectnessie.error.ReferenceConflicts.referenceConflicts;
@@ -34,12 +36,14 @@ import static org.projectnessie.model.Conflict.conflict;
 import static org.projectnessie.model.Content.Type.ICEBERG_TABLE;
 import static org.projectnessie.versioned.storage.common.persist.ObjIdHasher.objIdHasher;
 
+import jakarta.annotation.Nullable;
 import jakarta.enterprise.context.RequestScoped;
 import jakarta.inject.Inject;
 import jakarta.inject.Named;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -61,7 +65,10 @@ import org.projectnessie.catalog.formats.iceberg.nessie.IcebergTableMetadataUpda
 import org.projectnessie.catalog.formats.iceberg.nessie.IcebergViewMetadataUpdateState;
 import org.projectnessie.catalog.formats.iceberg.rest.IcebergCatalogOperation;
 import org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate;
+import org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate.AssignUUID;
 import org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate.SetLocation;
+import org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate.SetProperties;
+import org.projectnessie.catalog.formats.iceberg.rest.IcebergUpdateRequirement.AssertCreate;
 import org.projectnessie.catalog.model.id.NessieId;
 import org.projectnessie.catalog.model.ops.CatalogOperation;
 import org.projectnessie.catalog.model.snapshot.NessieEntitySnapshot;
@@ -175,7 +182,7 @@ public class CatalogServiceImpl implements CatalogService {
 
   @Override
   public CompletionStage<SnapshotResponse> retrieveSnapshot(
-      SnapshotReqParams reqParams, ContentKey key, Content.Type expectedType)
+      SnapshotReqParams reqParams, ContentKey key, @Nullable Content.Type expectedType)
       throws NessieNotFoundException {
 
     ParsedReference reference = reqParams.ref();
@@ -193,7 +200,7 @@ public class CatalogServiceImpl implements CatalogService {
             .hashOnRef(reference.hashWithRelativeSpec())
             .getSingle(key);
     Content content = contentResponse.getContent();
-    if (!content.getType().equals(expectedType)) {
+    if (expectedType != null && !content.getType().equals(expectedType)) {
       throw new NessieContentNotFoundException(key, reference.name());
     }
     Reference effectiveReference = contentResponse.getEffectiveReference();
@@ -371,11 +378,14 @@ public class CatalogServiceImpl implements CatalogService {
           .append(op.getType())
           .append(" ")
           .append(op.getKey());
+
       if (op.getType().equals(ICEBERG_TABLE)) {
+        verifyIcebergOperation(op, reference, content);
         commitBuilderStage =
             applyIcebergTableCommitOperation(
                 target, op, content, multiTableUpdate, commitBuilderStage);
       } else if (op.getType().equals(Content.Type.ICEBERG_VIEW)) {
+        verifyIcebergOperation(op, reference, content);
         commitBuilderStage =
             applyIcebergViewCommitOperation(
                 target, op, content, multiTableUpdate, commitBuilderStage);
@@ -434,43 +444,52 @@ public class CatalogServiceImpl implements CatalogService {
                             commitResponse.getTargetBranch())));
   }
 
+  private static void verifyIcebergOperation(
+      CatalogOperation op, ParsedReference reference, Content content)
+      throws NessieContentNotFoundException, NessieReferenceConflictException {
+    IcebergCatalogOperation icebergOp = (IcebergCatalogOperation) op;
+    boolean hasAssertCreate = icebergOp.hasRequirement(AssertCreate.class);
+    if (hasAssertCreate && content != null) {
+      throw new CatalogEntityAlreadyExistsException(
+          true, op.getType(), op.getKey(), content.getType());
+    }
+    if (!hasAssertCreate && content == null) {
+      throw new NessieContentNotFoundException(op.getKey(), reference.name());
+    }
+    if (content != null) {
+      if (!op.getType().equals(content.getType())) {
+        String msg =
+            format(
+                "Cannot update %s %s as a %s",
+                typeToEntityName(content.getType()).toLowerCase(Locale.ROOT),
+                op.getKey(),
+                typeToEntityName(op.getType()).toLowerCase(Locale.ROOT));
+        throw new NessieReferenceConflictException(
+            referenceConflicts(conflict(Conflict.ConflictType.PAYLOAD_DIFFERS, op.getKey(), msg)),
+            msg,
+            null);
+      }
+    }
+  }
+
   private CompletionStage<MultiTableUpdate> applyIcebergTableCommitOperation(
       Branch reference,
       CatalogOperation op,
       Content content,
       MultiTableUpdate multiTableUpdate,
       CompletionStage<MultiTableUpdate> commitBuilderStage)
-      throws NessieContentNotFoundException, NessieReferenceConflictException {
+      throws NessieContentNotFoundException {
     // TODO serialize the changes as well, so that we can retrieve those later for content-aware
     //  merges and automatic conflict resolution.
 
     IcebergCatalogOperation icebergOp = (IcebergCatalogOperation) op;
 
-    validateIcebergOperation(icebergOp);
-
-    if (icebergOp.hasAssertCreate()) {
-      if (content != null) {
-        throw new CatalogEntityAlreadyExistsException(
-            true, op.getType(), op.getKey(), content.getType());
-      }
-    } else if (!op.getType().equals(content.getType())) {
-      String msg =
-          format(
-              "Cannot update %s %s as a %s",
-              typeToEntityName(content.getType()).toLowerCase(Locale.ROOT),
-              op.getKey(),
-              typeToEntityName(op.getType()).toLowerCase(Locale.ROOT));
-      throw new NessieReferenceConflictException(
-          referenceConflicts(conflict(Conflict.ConflictType.PAYLOAD_DIFFERS, op.getKey(), msg)),
-          msg,
-          null);
-    }
-
     String contentId;
     CompletionStage<NessieTableSnapshot> snapshotStage;
     if (content == null) {
       contentId = null;
-      snapshotStage = completedStage(newIcebergTableSnapshot(icebergOp.updates()));
+      String icebergUuid = icebergOp.getSingleUpdateValue(AssignUUID.class, AssignUUID::uuid);
+      snapshotStage = completedStage(newIcebergTableSnapshot(icebergUuid));
     } else {
       contentId = content.getId();
       snapshotStage = loadExistingTableSnapshot(content);
@@ -492,7 +511,7 @@ public class CatalogServiceImpl implements CatalogService {
                   return new IcebergTableMetadataUpdateState(
                           nessieSnapshot, op.getKey(), content != null)
                       .checkRequirements(icebergOp.requirements())
-                      .applyUpdates(icebergOp.updates())
+                      .applyUpdates(pruneUpdates(icebergOp, content != null))
                       .snapshot();
                   // TODO handle the case when nothing changed -> do not update
                   //  e.g. when adding a schema/spec/order that already exists
@@ -539,13 +558,12 @@ public class CatalogServiceImpl implements CatalogService {
 
     IcebergCatalogOperation icebergOp = (IcebergCatalogOperation) op;
 
-    validateIcebergOperation(icebergOp);
-
     String contentId;
     CompletionStage<NessieViewSnapshot> snapshotStage;
     if (content == null) {
       contentId = null;
-      snapshotStage = completedStage(newIcebergViewSnapshot(icebergOp.updates()));
+      String icebergUuid = icebergOp.getSingleUpdateValue(AssignUUID.class, AssignUUID::uuid);
+      snapshotStage = completedStage(newIcebergViewSnapshot(icebergUuid));
     } else {
       contentId = content.getId();
       snapshotStage = loadExistingViewSnapshot(content);
@@ -567,7 +585,7 @@ public class CatalogServiceImpl implements CatalogService {
                   return new IcebergViewMetadataUpdateState(
                           nessieSnapshot, op.getKey(), content != null)
                       .checkRequirements(icebergOp.requirements())
-                      .applyUpdates(icebergOp.updates())
+                      .applyUpdates(pruneUpdates(icebergOp, content != null))
                       .snapshot();
                   // TODO handle the case when nothing changed -> do not update
                   //  e.g. when adding a schema/spec/order that already exists
@@ -601,16 +619,31 @@ public class CatalogServiceImpl implements CatalogService {
     return commitBuilderStage;
   }
 
-  private void validateIcebergOperation(IcebergCatalogOperation icebergOp) {
-    for (IcebergMetadataUpdate u : icebergOp.updates()) {
-      if (u instanceof SetLocation) {
-        StorageUri uri = StorageUri.of(((SetLocation) u).location());
-        checkArgument(objectIO.isValidUri(uri), "Unsupported table or view location: %s", uri);
-        WarehouseConfig warehouse = catalogConfig.getWarehouse(icebergOp.warehouse());
-        boolean inWarehouse = !StorageUri.of(warehouse.location()).relativize(uri).isAbsolute();
-        checkArgument(inWarehouse, "Invalid table or view location : %s", uri);
+  private List<IcebergMetadataUpdate> pruneUpdates(IcebergCatalogOperation op, boolean update) {
+    if (update) {
+      return op.updates();
+    }
+    List<IcebergMetadataUpdate> prunedUpdates = new ArrayList<>(op.updates());
+    String location = null;
+    if (op.hasUpdate(SetProperties.class)) {
+      Map<String, String> properties =
+          op.getSingleUpdateValue(SetProperties.class, SetProperties::updates);
+      if (properties.containsKey(IcebergTableMetadata.STAGED_PROPERTY)) {
+        String stagedLocation = op.getSingleUpdateValue(SetLocation.class, SetLocation::location);
+        // TODO verify integrity of staged location
+        prunedUpdates.removeIf(u -> u instanceof SetProperties);
+        properties = new HashMap<>(properties);
+        properties.remove(IcebergTableMetadata.STAGED_PROPERTY);
+        prunedUpdates.add(setProperties(properties));
+        location = stagedLocation;
       }
     }
+    if (location == null) {
+      WarehouseConfig w = catalogConfig.getWarehouse(op.warehouse());
+      location = icebergBaseLocation(w.location(), op.getKey());
+    }
+    prunedUpdates.add(setTrustedLocation(location));
+    return prunedUpdates;
   }
 
   static final class MultiTableUpdate {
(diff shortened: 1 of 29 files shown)


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -97,31 +110,95 @@ public IcebergConfigResponse getConfig(
...
     Optional<String> bucket = s3options.extractBucket(uri);
     Optional<String> body = Optional.ofNullable(request.body());
 
     SigningRequest signingRequest =
         SigningRequest.signingRequest(
             uri, request.method(), request.region(), bucket, body, request.headers());
 
-    SigningResponse signed = signer.sign(ref.name(), identifier, signingRequest);
+    SigningResponse signed =
+        signer.sign(decodedPrefix.parsedReference().name(), key.toPathString(), signingRequest);
```
adutra, line 187: I noticed that the current impl of this component ignores the reference and the key, FYI.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -97,31 +110,95 @@ public IcebergConfigResponse getConfig(
...
+                  Multi.createFrom()
+                      .emitter(
+                          e -> {
+                            e.emit(snapshot.icebergLocation());
+                            snapshot.additionalKnownLocations().forEach(e::emit);
+                            e.complete();
+                          }))
+          // allow the parent directory of each iceberg location,
+          // assuming the locations are of the form: <base-prefix>/<random-part>
+          .map(location -> location.substring(0, location.lastIndexOf('/') + 1));
```
snazy, line 166: The location(s) is/are always absolute base paths. It is safe to assume that everything for a table is underneath one of these paths.
I.e. the correct way here would be to append a `/` (if not already present), but truncating would expose all tables in the same namaspace.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
snazy: Yes, the original idea at that time (not fully thought through) was that the signer implementation checks reference and key. But that's no longer the case, so I guess the parameters can go away.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -97,31 +110,95 @@ public IcebergConfigResponse getConfig(
...
-    ParsedReference ref = decodePrefix(prefix).parsedReference();
+      @PathParam("identifier") String identifier)
+      throws NessieNotFoundException {
 
-    URI uri = URI.create(request.uri());
+    DecodedPrefix decodedPrefix = decodePrefix(prefix);
+    ContentKey key = ContentKey.fromPathString(identifier);
+    String requestedS3Uri = S3Utils.asS3Location(request.uri());
+
+    return collectBasePrefixes(decodedPrefix, key)
```
snazy, line 126: What happens here if multiple prefixes match? Does the `toUni` bark in that case?

snazy (commented):
Haven't fully reviewed the changes, only looked at some "relevant production parts" and left a few comments.

> Note that I changed the default base location from table1_<uuid> to table1/<uuid> 

I don't mind actually. Not sure if that really matters though, because we have to consider the _full_ base location (with a trailing `/`), not the "parent directory" of a base location, because that would "expose" the whole namespace for already existing tables.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
adutra: I've hardened the logic a bit, let me know if that's better. Unfortunately we cannot just take the locations as-is.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
adutra: The first one is converted to `Uni` the other ones, if any, are discarded. From the javadoc:

> any item or failure events received after the first event is dropped

adutra: Hmm I should invert `map().toUni()` it seems.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -175,4 +179,129 @@ public Uni<Void> commitTransaction(
...
+                e -> {
+                  e.emit(snapshot.icebergLocation());
+                  snapshot.additionalKnownLocations().forEach(e::emit);
+                  e.complete();
+                })
+            // if location is within the base prefix, use the base prefix;
+            // otherwise, use the location as-is. This allows for temporary
+            // locations to be used for signing, as long as they are not
+            // outside the base prefix.
+            .map(location -> location.startsWith(basePrefix) ? basePrefix : location);
```
snazy, line 277: Hm - not really a fan, because it basically exposes the whole basePrefix. Is this only for one or two tests?

If it's more than that, I'd be okay to leave it as it is but have a follow-up issue to tackle this now.

snazy (commented):
Overall the change LGTM and I don't see anything that's wrong or could go completely sideways.

What worries me a bit is the new `IcebergRequestSigner` class. Can you make it a package-private class and add some tests for it - I suspect using plain Mockito should help to simulate calls against the catalog-service.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
adutra: Yes it's basically to make tests pass. Would you be OK if we changed the tests, or disabled them?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
```diff
@@ -175,4 +179,129 @@ public Uni<Void> commitTransaction(
...
+          .filter(requestedS3Uri::startsWith)
+          .toUni()
+          .onItem()
+          .ifNull()
+          .failWith(this::unauthorized)
+          .replaceWith(request.uri())
+          .map(this::sign);
+    }
+
+    private Uni<Content> fetchContent() {
```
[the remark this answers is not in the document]
adutra: Hmm weird, I looked into it and it seemed to me that `entityObject()` wasn't returning the `Content`. I will re-check.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
snazy: I think, we can live without this fallback.

`testRegisterTableFromFileSystem` ... maybe let that write the metadata of the table to be registered somewhere underneath the original table's location - dirty, but at least a quick fix.
Side note: the javadoc of that one needs to be changed.

`testCompleteCreateOrReplaceTransactionReplace` + `testCompleteReplaceTransaction` could write the file directly into the table's location directory (somewhere in there at least).

WDYT?

adutra: That wouldn't work, the tests do not know the definitive table location upfront, because it contains a random uuid.

At this point, I think the only option is do disable the tests, or change them in a way that they use the default location, not a custom one.

snazy: The location should be there from the create-table before.

adutra: In certain tests there is no create-table before:

https://github.com/projectnessie/nessie/blob/170dd11d4c04dbccc042b7aa11cb76f9fa144c17/servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/CatalogTests.java#L630-L637

snazy: But that can be "guarded" with an `if (overridesRequestedLocation())` and omit the `withLocation`

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/CatalogTests.java
```diff
@@ -1972,15 +1971,20 @@ public void testCompleteCreateOrReplaceTransactionReplace() {
...
         catalog
             .buildTable(TABLE, SCHEMA)
-            .withLocation(location)
             .withPartitionSpec(SPEC)
             .withSortOrder(WRITE_ORDER)
-            .withProperties(properties)
-            .createOrReplaceTransaction();
+            .withProperties(properties);
+
+    String location = temporaryLocation();
```
adutra, line 1982: I changed 2 or 3 tests like this – I think the change makes sense: if the catalog overrides locations, the client should not call `withLocation` nor write files to a random location before the catalog gets a chance to properly create the table.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1GenericResource.java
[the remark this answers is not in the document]
adutra: Found it, although, I need the content type to call `retrieveSnapshot`, so I would need to change the method impl to allow passing null as a content type.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -601,16 +620,37 @@ private CompletionStage<MultiTableUpdate> applyIcebergViewCommitOperation(
...
+    if (update) {
+      return op.updates();
+    }
+    List<IcebergMetadataUpdate> prunedUpdates = new ArrayList<>(op.updates());
+    String location = null;
+    if (op.hasUpdate(SetProperties.class)) {
+      Map<String, String> properties =
+          op.getSingleUpdateValue(SetProperties.class, SetProperties::updates);
+      if (properties.containsKey(IcebergTableMetadata.STAGED_PROPERTY)) {
+        String stagedLocation = op.getSingleUpdateValue(SetLocation.class, SetLocation::location);
```
snazy, line 633: Let's just use the location from `SetLocation` for now. We can add actual verifications later.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
```diff
@@ -91,11 +91,13 @@ default Map<String, String> icebergConfigOverrides() {
         "uri", icebergBaseURI().toString());
   }
 
-  default URI icebergS3SignerUri(String prefix, ContentKey contentKey) {
+  default URI icebergS3SignerUri(String prefix, ContentKey contentKey, String encryptedLocation) {
```
snazy, line 94: Same here, let's remove the new parameter for now.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3Signer.java
```diff
@@ -0,0 +1,194 @@
...
+      return Uni.createFrom().failure(e);
+    }
+  }
+
+  private Uni<?> checkForbiddenLocations(SnapshotResponse snapshotResponse) {
+    if (snapshotResponse != null && write()) {
+      Content content = snapshotResponse.content();
+      String metadataLocation = null;
+      if (content instanceof IcebergTable) {
+        metadataLocation = ((IcebergTable) content).getMetadataLocation();
```
snazy, line 125: Actually, this can be simplified by checking for the object-name-pattern for metadata-files.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3Signer.java
```diff
@@ -0,0 +1,194 @@
...
+import org.projectnessie.error.NessieNotFoundException;
+import org.projectnessie.model.Content;
+import org.projectnessie.model.ContentKey;
+import org.projectnessie.model.IcebergTable;
+import org.projectnessie.model.IcebergView;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Value.Immutable
+abstract class IcebergS3Signer {
```
snazy, line 53: Maybe `IcebergS3SignParams`?
There's already `RequestSigner`.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3Signer.java
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  @Value.Lazy
+  String decryptedLocation() {
+    return encryptionService().decrypt(encryptedLocation());
+  }
+
+  @Value.Lazy
+  boolean write() {
+    return request().method().equalsIgnoreCase("PUT")
```
snazy, line 85: Maybe add `PATCH` as well.

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/CatalogTests.java
```diff
@@ -173,9 +174,7 @@ protected boolean supportsNamesWithDot() {
     return true;
   }
 
-  protected String temporaryLocation() {
-    return "file:/tmp/ns/table";
-  }
+  protected abstract String temporaryLocation();
```
snazy, line 177: Better leave the original one. The intent of `CatalogTests.java` + `ViewCatalogTests.java` being here is to adopt those for Iceberg REST and later contribute the changes back to Iceberg.

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/CatalogTests.java
snazy: Yup - that's the change I had in mind.
However, the call to `temporaryLocation()` can be moved inside the `if`-body.

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/ViewCatalogTests.java
```diff
@@ -71,11 +69,7 @@ public abstract class ViewCatalogTests<C extends ViewCatalog & SupportsNamespace
 
   protected abstract Catalog tableCatalog();
 
-  @TempDir private Path tempDir;
```
snazy, line 74: Let's leave this as is (for the same reason as mentioned in `CatalogTests.java`).

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -434,43 +446,51 @@ public CompletionStage<Stream<SnapshotResponse>> commit(
                             commitResponse.getTargetBranch())));
   }
 
+  private static void verifyIcebergOperation(
+      CatalogOperation op, ParsedReference reference, Content content)
+      throws NessieContentNotFoundException, NessieReferenceConflictException {
+    IcebergCatalogOperation icebergOp = (IcebergCatalogOperation) op;
+    if (icebergOp.hasRequirement(AssertCreate.class) && content != null) {
```
snazy, line 453: Memoize the result of `icebergOp.hasRequirement(AssertCreate.class)`?

catalog/service/rest/src/test/java/org/projectnessie/catalog/service/rest/TestIcebergS3Signer.java
```diff
@@ -0,0 +1,300 @@
...
+          key,
+          table,
+          nessieTableSnapshot);
+  CompletionStage<SnapshotResponse> successStage =
+      CompletableFuture.completedStage(snapshotResponse);
+  SigningResponse signingResponse =
+      ImmutableSigningResponse.builder().uri(URI.create(dataFileUri)).build();
+
+  @ParameterizedTest
+  @ValueSource(strings = {"GET", "HEAD"})
```
snazy, line 124: Not sure if those ever become valid method, but maybe add `OPTIONS` + `TRACE`?

catalog/service/rest/src/test/java/org/projectnessie/catalog/service/rest/TestIcebergS3Signer.java
```diff
@@ -0,0 +1,300 @@
...
+    when(encryptionService.decrypt(encryptedLocation)).thenReturn(baseLocation);
+    when(catalogService.retrieveSnapshot(any(), eq(key), isNull())).thenReturn(successStage);
+    when(signer.sign(any())).thenReturn(signingResponse);
+    IcebergS3Signer icebergSigner = newBuilder().request(readRequest).build();
+    Uni<IcebergS3SignResponse> response = icebergSigner.verifyAndSign();
+    expectSuccess(response);
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"PUT", "POST", "DELETE"})
```
snazy, line 136: Add `PATCH`?

servers/quarkus-server/src/intTest/java/org/projectnessie/server/catalog/auth/AbstractAuthEnabledTests.java
```diff
@@ -107,6 +107,11 @@ protected NessieClientBuilder nessieClientBuilder() {
                     .build()));
   }
 
+  @Override
```
snazy, line 110: Noop-change?

servers/quarkus-server/src/intTest/java/org/projectnessie/server/catalog/gcs/ITGcsIcebergCatalog.java
```diff
@@ -41,6 +40,6 @@ protected Map<String, String> catalogOptions() {
 
   @Override
   protected String temporaryLocation() {
-    return warehouseLocation.toString() + "/temp/" + randomUUID();
+    return warehouseLocation.toString() + "/temp/" + UUID.randomUUID();
```
snazy, line 43: Noop-change?

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/AbstractIcebergCatalogTests.java
```diff
@@ -128,11 +133,6 @@ protected boolean overridesRequestedLocation() {
     return true;
   }
 
-  @Override
```
snazy, line 131: Leave this one? (See `CatalogTests.java`)

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/AbstractIcebergCatalogTests.java
```diff
@@ -204,6 +207,16 @@ public void testRegisterTableFromFileSystem() throws Exception {
     Assertions.assertThat(catalog.tableExists(TABLE)).isFalse();
   }
 
+  protected FileIO temporaryFileIO(RESTCatalog catalog) {
```
snazy, line 210: Abstract to be reusable in `ITS3IcebergCatalog`?

catalog/service/common/src/main/java/org/projectnessie/catalog/service/api/EncryptionService.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.service.api;
+
+public interface EncryptionService {
```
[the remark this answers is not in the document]
adutra: I thought we agreed to introduce an empty shell for now, but OK.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
adutra: We can't, we need it.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3Signer.java
adutra: Is it carved in stone? Isn't it better to trust whatever was saved in the database?

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/AbstractIcebergCatalogTests.java
adutra: We would need to implement this method in: 

- TestAdlsIcebergCatalog (org.projectnessie.server.catalog)
- TestS3IcebergCatalog (org.projectnessie.server.catalog)
- TestGcsIcebergCatalog (org.projectnessie.server.catalog)
- ITGcsIcebergCatalog (org.projectnessie.server.catalog.gcs)
- ITS3IcebergCatalog (org.projectnessie.server.catalog.s3)
- ITOAuthIcebergCatalog (org.projectnessie.server.catalog.auth)
- ITBearerTokenIcebergCatalog (org.projectnessie.server.catalog.auth)

Not sure if the duplication is worth it.

catalog/service/common/src/main/java/org/projectnessie/catalog/service/api/EncryptionService.java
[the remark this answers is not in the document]
snazy: Not that I remember TBH.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3Signer.java
snazy: The pattern for the metadata-JSONs is pretty much defined, yes.
Checking the pattern prevents overwriting metadata.

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/AbstractIcebergCatalogTests.java
snazy: It is already overridden in `ITS3IcebergCatalog` - the only difference is the access-key-id and secret-access-key.
It's okay though as it is, if `ITS3IcebergCatalog` is the only overridden one.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
adutra: For staged tables.

snazy: Yes - `location` but not "encrypted".

snazy: (Meant "encrypted" with "remove" - my bad)

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -145,7 +146,9 @@ public Map<String, String> icebergConfigPerTable(
       config.put(
           S3_SIGNER_ENDPOINT,
           // TODO does it make sense to use a separate endpoint (service) just for signing?
```
snazy, line 148: Nit: this comment can be removed.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -0,0 +1,193 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(IcebergS3SignParams.class);
+
+  abstract IcebergS3SignRequest request();
+
+  abstract ParsedReference ref();
+
+  abstract ContentKey key();
+
+  abstract String baseLocation();
```
snazy, line 63: Does it make sense to add a check that this is a "pure" `s3://` location (and not s3a/s3n)?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -0,0 +1,193 @@
...
+      String metadataLocation = null;
+      if (content instanceof IcebergTable) {
+        metadataLocation = ((IcebergTable) content).getMetadataLocation();
+      } else if (content instanceof IcebergView) {
+        metadataLocation = ((IcebergView) content).getMetadataLocation();
+      }
+      if (metadataLocation != null
+          && requestedS3Uri().equals(S3Utils.normalizeS3Scheme(metadataLocation))) {
+        return Uni.createFrom().failure(unauthorized());
+      }
```
snazy, line 126: Mind adding a TODO to disallow all table and view metadata objects?
I know, if you add the code now, some tests fill fail (the ones that write to temporary locations).

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -0,0 +1,193 @@
...
+        .replaceWith(request().uri())
+        .map(this::sign);
+  }
+
+  private Uni<SnapshotResponse> fetchSnapshot() {
+    try {
+      CompletionStage<SnapshotResponse> stage =
+          catalogService()
+              .retrieveSnapshot(
+                  SnapshotReqParams.forSnapshotHttpReq(ref(), "iceberg", null), key(), null);
```
snazy, line 103: Mind adding a TODO here to let `CatalogServiceImpl.retrieveSnapshot()` get a `boolean forWrite` to perform access checks for writes early? #8768 would allow that.

snazy (commented):
Just a few comments that I think are important - otherwise LGTM!

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
adutra: OK for now – long term I was rather leaning the idea of "pruning" at the warehouse config level. Basically all the locations are derived from that, so if you prune that, all the rest is pruned too.

snazy: We can't do that there. The `a`+`n` could have a special meaning to clients in the way they read & write data.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -68,6 +70,12 @@ abstract class IcebergS3SignParams {
 
   abstract S3BucketOptions s3options();
 
+  @Check
+  public void check() {
+    checkArgument(
+        URI.create(baseLocation()).getScheme().equals("s3"), "baseLocation must be an S3 URI");
```
snazy, line 76: Nit: `.startsWith("s3:")` would have worked, too ;)

snazy (dismissed):

snazy (approved):

PR #8667, opened 2024-05-30
[Catalog] Handle `CompletionStage` from snapshot store

snazy:


adutra (approved):
And thanks for the added comments, that wasn't clear at all for me!

snazy:
> And thanks for the added comments, that wasn't clear at all for me!

Heh - I had to look into that as well ;)

PR #8672, opened 2024-05-30
Helm chart: add support for JDBC catalog and schema

adutra:


snazy (approved):
LGTM - But actually, I'd prefer to have the docs/references from the README on the web site :)

PR #8674, opened 2024-05-31
Make Catalog base URI configurable

dimas-b:
Helps when running with LB / TLS termination.

### DIFF
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
@@ -30,6 +30,10 @@ public interface CatalogConfig {
    *
    * <p>This parameter controls the URI visible to Catalog clients. It can be useful when the serer
    * is behind a proxy or load balancer that alters the port or adds/removes URI path elements.
+   *
+   * <p>Example:
+   * <li>External URI: {@code https://example.com:1234/nessie/iceberg/v1/config}
+   * <li>Internal URI: {@code http://nessie.svc.cluster.local/iceberg/v1/config}
    */
   Optional<URI> externalBaseUri();
 
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
@@ -17,11 +17,22 @@ package org.projectnessie.catalog.service.config;
 
 import static org.projectnessie.catalog.service.config.CatalogConfig.removeTrailingSlash;
 
+import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
 import org.projectnessie.nessie.docgen.annotations.ConfigDocs.ConfigPropertyName;
 
 public interface CatalogConfig {
+
+  /**
+   * Optional override for the Catalog Server's base URI. If not set, URI info from the application
+   * server will be used.
+   *
+   * <p>This parameter controls the URI visible to Catalog clients. It can be useful when the serer
+   * is behind a proxy or load balancer that alters the protocol (TLS termination) or adds/removes
+   * URI path elements.
+   */
+  Optional<URI> baseUri();
 
   /**
    * Name of the default warehouse. This one is used when a warehouse is not specified in a query.
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
@@ -29,10 +29,9 @@ public interface CatalogConfig {
    * server will be used.
    *
    * <p>This parameter controls the URI visible to Catalog clients. It can be useful when the serer
-   * is behind a proxy or load balancer that alters the protocol (TLS termination) or adds/removes
-   * URI path elements.
+   * is behind a proxy or load balancer that alters the port or adds/removes URI path elements.
    */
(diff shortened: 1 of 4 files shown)


catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
```diff
@@ -17,12 +17,23 @@
...
 
+  /**
+   * Optional override for the Catalog Server's base URI. If not set, URI info from the application
+   * server will be used.
+   *
+   * <p>This parameter controls the URI visible to Catalog clients. It can be useful when the serer
+   * is behind a proxy or load balancer that alters the protocol (TLS termination) or adds/removes
+   * URI path elements.
+   */
+  Optional<URI> baseUri();
```
snazy, line 35: We need this to be configurable via our Helm chart as well.

snazy, line 35: We already have a similar parameter for S3 endpoints, there it's called "external". Guess, we should have a common nomenclature for those things.
```suggestion
  Optional<URI> externalBaseURI();
```

catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
```diff
@@ -17,12 +17,23 @@
...
+import java.net.URI;
 import java.util.Map;
 import java.util.Optional;
 import org.projectnessie.nessie.docgen.annotations.ConfigDocs.ConfigPropertyName;
 
 public interface CatalogConfig {
 
+  /**
+   * Optional override for the Catalog Server's base URI. If not set, URI info from the application
+   * server will be used.
```
snazy, line 29:
```suggestion
   * The base URI used by Iceberg REST clients to contact Nessie, needs to be specified if Nessie runs behind a proxy and/or load-balancer.
```

I guess, an example would be nice here.

snazy (commented):
Wonder whether Quarkus' built-in support (`quarkus.http.proxy.allow-forwarded=true`) for reverse-proxies isn't enough?
https://quarkus.io/guides/http-reference#reverse-proxy

dimas-b:
Re: the `Forwarded` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded): it appears to allow Quarkus to obtain the original Host and URI scheme (proto) information, but in the (reverse) proxy rewrites the URL path, the original path appears to be lost.

For example: external URI: `https://example.com:1234/nessie/api/v2/config`, internal: `http://nessie.localdomain/api/v2/config`.

The `Forwarded` header does not appear to communicate the original port (1234) and the original path (`/nessie/...`).

snazy:
Have you taken a look at https://quarkus.io/guides/http-reference#reverse-proxy ??

catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
dimas-b: `env NESSIE_CATALOG_BASE_URI=https://nessie.example.com/`

catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
dimas-b: added

dimas-b: I did not rephrase the javadoc because the override is not strictly necessary. It is needed only if the proxy changes port or URI path.

catalog/service/common/src/main/java/org/projectnessie/catalog/service/config/CatalogConfig.java
dimas-b: renamed

dimas-b:
> Have you taken a look at https://quarkus.io/guides/http-reference#reverse-proxy ??

I even tried it in practice. If you have a specific config set suggestion that would work with URI path rewriting, please share.

dimas-b:
> Wonder whether Quarkus' built-in support (quarkus.http.proxy.allow-forwarded=true) for reverse-proxies isn't enough?

@snazy : if you mean `X-Forwarded-Prefix`, yes, that seems applicable, but it's not what you original comment suggested :) That property is for the `Forwarded` header.

Also, I could not find any documentation about `X-Forwarded-Prefix` outside of Quarkus.

dimas-b:
Cf. https://github.com/quarkusio/quarkus/issues/14537

dimas-b:
Currently we effectively get the base URI from [UriInfoImpl](https://github.com/quarkusio/quarkus/blob/d9b3f180f2ec2f25ca6e89f6e081cfe2143d3cc7/independent-projects/resteasy-reactive/server/runtime/src/main/java/org/jboss/resteasy/reactive/server/jaxrs/UriInfoImpl.java#L125), but it does not appear to respect prefix overrides. 

The [Quarkus changes](https://github.com/quarkusio/quarkus/pull/25117) for `X-Forwarded-Prefix` are effective in this case, but the prefix does not appear to make its way into `UriInfo` in this case... The behaviour is complicated and uncertain, IMHO.

snazy:
I'd prefer to have it fixed in Quarkus supporting rather "de facto" standards, over having a custom implementation that requires additional explicit configuration. Adding this to Nessie would tie Nessie to a single "external" URI, eliminating the ability to have multiple ingresses (e.g. internal and external).

PR #8717, opened 2024-06-04
Helm chart: minor improvement to JDBC section in docs

adutra:


site/docs/guides/kubernetes.md
```diff
@@ -175,13 +174,43 @@ database credentials from the secret that was created previously. We can do this
...
+```
+
+!!! note
+    The exact format of the JDBC URL may vary depending on the database you are using. Check the
+    documentation of your database and its JDBC driver for more information (for PostgreSQL, [check 
+    out this page](https://jdbc.postgresql.org/documentation/use/) and for MariaDB, [check out this 
+    one](https://mariadb.com/kb/en/about-mariadb-connector-j/#connection-strings)).
+
+The database and schema can be either passed in the JDBC URL (recommended) or as separate
+configuration options. The following configuration is equivalent to the one above:
```
dimas-b, line 191: I believe for PosgreSQL _both_ the JDBC URL and the `schema` config variable must be set (to the same value). If the `schema` config is set, but not the URL default schema, Nessie will try to re-create tables on the _second_ run. I'm pretty sure I saw that in practice.

dimas-b: Nessie will probably work fine with PosgreSQL if the schema is set in neither the JDBC URL, nor in the config.

site/docs/guides/kubernetes.md
```diff
@@ -175,13 +174,43 @@ database credentials from the secret that was created previously. We can do this
...
+  catalog: nessiedb
+  schema: nessie
   secret:
     name: postgres-creds
     username: postgres_username
     password: postgres_password
 ```
 
+!!! note 
+    The required tables will be created automatically by Nessie if they don't exist, in the
```
dimas-b, line 206: (side note) do we have any control over this in our helm charts? I imagine two pods running DDL at the same time is not ideal :thinking:

site/docs/guides/kubernetes.md
adutra: Hmm not sure, I just tried deployed a toy Nessie with this URL: `jdbc:postgresql://nessie-postgres-postgresql:5432/nessiedb?currentSchema=nessie` and no `schema`, and everything works as expected.

Then I tried `jdbc:postgresql://nessie-postgres-postgresql:5432/` with `catalog=nessiedb` and `schema=nessie2`, it works too.

As for the second run issue, I'm surprised given how the code proceeds:

https://github.com/projectnessie/nessie/blob/e9d7bebfd71deeec76ac9d0df661ca29c425a730/versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/JdbcBackend.java#L178-L183

If you saw this happening it would be good to have a reproducer.

adutra: Hmm. Actually I just saw something similar. I think the code above might be a bit off. I'm investigating.

site/docs/guides/kubernetes.md
adutra: I thought about that too, there is no formal control over that, and `JdbcBackendBuilder.setupSchema()` could race and throw if called concurrently.

adutra:
On hold until we sort out #8723 .

site/docs/guides/kubernetes.md
dimas-b: Will Nessie Server stop on DDL failure? If yes, that should be good enough for k8s (I did not test :) )

dimas-b (approved):

PR #8723, opened 2024-06-04
JDBC: remove options 'catalog' and 'schema' and revisit schema creation

adutra:
This PR:

* Removes the `catalog` and `schema` options, which did not work as expected;
* Makes `JdbcBackend` rely exclusively on the catalog and schema coming from the connection, and hence from the JDBC URL;
* Also protects against potential races when creating the JDBC schema, by checking if the error is an "already exists" error.

### DIFF
--- a/versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/JdbcBackend.java
+++ b/versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/JdbcBackend.java
@@ -253,10 +253,11 @@ public final class JdbcBackend implements Backend {
 
       try (ResultSet rs = conn.getMetaData().getTables(catalog, schema, tableName, null)) {
         if (rs.next()) {
+          // table already exists
+
           Map<String, Integer> primaryKey = new LinkedHashMap<>();
           Map<String, Integer> columns = new LinkedHashMap<>();
 
-          // table already exists
           try (ResultSet cols = conn.getMetaData().getColumns(catalog, schema, tableName, null)) {
             while (cols.next()) {
               String colName = cols.getString("COLUMN_NAME").toLowerCase(Locale.ROOT);
@@ -299,11 +300,11 @@ public final class JdbcBackend implements Backend {
       try {
         st.executeUpdate(createTable);
       } catch (SQLException e) {
-        if (databaseSpecific.isAlreadyExists(e)) {
-          // table was created by another process, proceed
-          return;
+        if (!databaseSpecific.isAlreadyExists(e)) {
+          throw e;
         }
-        throw e;
+        // table was created by another process, try again to check the schema
+        createTableIfNotExists(conn, tableName, createTable, expectedColumns, expectedPrimaryKey);
       }
     }
   }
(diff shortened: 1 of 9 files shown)


snazy (commented):
I don't mind also adding the functional change in this PR, but please update the PR title and description.

adutra:
Manually validated by deploying Nessie to our dev environment:

The following:

```bash
helm upgrade nessie-postgres-helm -n nessie-ns helm/nessie \
  --set versionStoreType=JDBC \
  --set 'postgres.jdbcUrl=jdbc:postgresql://nessie-postgres-postgresql:5432/' \
  --set jdbc.secret.name=postgres-creds \
  --set postgres.secret.password=postgres_password \
  --set postgres.secret.username=postgres_username
```

Creates the tables in the default catalog (`postgres`) and default schema (`public`).

The following:

```bash
 helm install nessie-postgres-helm -n nessie-ns helm/nessie \
  --set versionStoreType=JDBC \
  --set 'postgres.jdbcUrl=jdbc:postgresql://nessie-postgres-postgresql:5432/nessiedb?currentSchema=nessie' \
  --set jdbc.secret.name=postgres-creds \
  --set postgres.secret.password=postgres_password \
  --set postgres.secret.username=postgres_username
```

Creates the tables in the catalog `nessiedb` and schema `nessie`.

adutra:
One interesting thing to notice is that Postgres requires a slash in the URL; the following URL is incorrect:

`jdbc:postgresql://nessie-postgres-postgresql:5432`

But this one is correct:

`jdbc:postgresql://nessie-postgres-postgresql:5432/`

adutra:
Manually verified for MariaDB as well. It seems MariaDB /MySQL have no support for schemas inside catalogs, and specifying the catalog/database name is mandatory in the URL. The following:

```
helm upgrade nessie-mariadb-helm -n nessie-ns helm/nessie \
  --set versionStoreType=JDBC \
  --set 'postgres.jdbcUrl=jdbc:mariadb://mariadb-helm-release.nessie-ns.svc.cluster.local:3306/nessiedb' \
  --set jdbc.secret.name=mariadb-creds \
  --set postgres.secret.password=mariadb_password \
  --set postgres.secret.username=mariadb_username
```

Creates the tables in catalog `nessiedb` (schema is null).

A URL like `jdbc:mariadb://mariadb-helm-release.nessie-ns.svc.cluster.local:3306` is invalid (no database selected).

versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/JdbcBackend.java
```diff
@@ -169,18 +187,24 @@ public PersistFactory createFactory() {
   Connection borrowConnection() throws SQLException {
     Connection c = dataSource.getConnection();
     c.setAutoCommit(false);
+    if (first.compareAndSet(true, false)) {
+      catalog = c.getCatalog();
+      schema = c.getSchema();
```
dimas-b, line 192: Do we actually need these fields? I can only see them being used in `configInfo()` but there's no guarantee that `borrowConnection()` is called before `configInfo()`... WDYT?

dimas-b: Actually, `PersistProvider` calls both `setupSchema()` and `configInfo()`... I wonder if we could have the config info as a return value from `setupSchema()`... WDYT?

adutra: You are correct, and yes that could be a nice little improvement. Mind if I do this in a follow-up PR?

dimas-b: sure :+1:

dimas-b (approved):

PR #8725, opened 2024-06-04
Helm chart: reload pods when configmap changes

adutra:


dimas-b (approved):

adutra:
Manually validated that this is working as expected.

PR #8746, opened 2024-06-05
Add explicit authorization-type = CEL

snazy:


design/authorization.md
```diff
@@ -126,6 +126,7 @@ An implementation of the `BatchAccessChecker` interface could be written with th
 Some example rules are shown below:
 ```
 nessie.server.authorization.enabled=true
+nessie.server.authorization.type=CEL
```
adutra, line 129: Do we need to change the Helm chart?

snazy: CEL's the default

adutra (approved):

PR #8751, opened 2024-06-05
JDBC: add proper support for H2

adutra:
Knowing of course that H2 is not a choice suitable for production, we claim nevertheless that Nessie has support for H2, but it seems this has never been really possible, because the `quarkus-jdbc-h2` extension was missing.

This change introduces proper support for H2, and also logs a warning when H2 is used.


servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/AbstractQuarkusSmoke.java
```diff
@@ -41,7 +40,6 @@
 import org.projectnessie.model.Tag;
 
 @SuppressWarnings("resource")
-@QuarkusIntegrationTest
 @ExtendWith(QuarkusNessieClientResolver.class)
```
dimas-b, line 43: nit: `QuarkusNessieClientResolver` may rely on running under `QuarkusIntegrationTest`... but I do not remember that code well anymore :) so just a note, not critical.

dimas-b (approved):

PR #8754, opened 2024-06-06
Site: Add RSS+JSON feeds

snazy:


adutra (approved):
LGTM although I thought RSS was kind of dead 😄

PR #8769, opened 2024-06-07
API compatibility filter: don't share authentication with main client

adutra:
This PR:

* Introduces a new `NessieAuthentication.copy()` method, called when calling `HttpClientBuilder.copy()`.
* Introduces a new (private) OAuth2 config option: client name, set to a random name, to facilitate debugging when more than one client is active. The client name is printed with all debug messages.
* Introduces minor changes in `OAuth2TokenRefreshExecutor` to also print the client name.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
@@ -77,13 +77,13 @@ final class HttpClientBuilderImpl implements HttpClient.Builder {
 
   HttpClientBuilderImpl(HttpClientBuilderImpl other) {
     this.baseUri = other.baseUri;
-    this.mapper = other.mapper;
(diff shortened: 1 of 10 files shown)


api/client/src/main/java/org/projectnessie/client/http/HttpClientBuilderImpl.java
```diff
@@ -77,13 +77,13 @@ final class HttpClientBuilderImpl implements HttpClient.Builder {
 
   HttpClientBuilderImpl(HttpClientBuilderImpl other) {
     this.baseUri = other.baseUri;
-    this.mapper = other.mapper;
+    this.mapper = other.mapper == null ? null : other.mapper.copy();
```
adutra, line 80: Not related to the bug with the OAuth2 client, but I think `ObjectMapper` instances, while thread-safe, are not immutable. So it's probably safer to copy them as well.

dimas-b (approved):

PR #8787, opened 2024-06-11
OAuth2Client: fix race condition when scheduling refreshes

adutra:
Observed in catalog tests.


snazy (approved):
change makes sense to me

PR #8790, opened 2024-06-11
OAuth2Client: don't refresh tokens using token exchange

### LINKED ISSUE #8779: OAuth client: deprecate token exchange flow when refreshing tokens
### Issue description

I would like to suggest that we deprecate, for removal, the usage of the token exchange flow for refreshing access tokens, before we go GA.

The motivation is that token exchange is not really meant for that, even if some servers accept to deliver a refresh token in this situation.

Disabling token exchange should not cause any visible disruption: if the `refresh_token` grant type cannot be used, because there is no refresh token available, instead of attempting a token exchange, the client would naturally fall back to re-issuing an initial token request using a grant like `client_credentials`.

Later, token exchange should be re-implemented or rethought from scratch to support more advanced use cases like impersonation or delegation, possibly allowing actor tokens to be defined, as well as supporting many other token types than just access tokens (SAML, etc).

adutra:
Fixes #8779


adutra:
I'm not completely removing the token exchange machinery, because the idea is to bring token exchange back later, but in a better / different form.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -77,7 +77,13 @@ void testBackgroundRefresh() throws Exception {
 
     try (HttpTestServer server = new HttpTestServer(handler(), true)) {
 
-      OAuth2ClientConfig config = configBuilder(server, false).executor(executor).build();
+      OAuth2ClientConfig config =
+          configBuilder(server, false)
+              .executor(executor)
+              .grantType(GrantType.PASSWORD)
```
adutra, line 83: Need a different grant type now, in order to exercise token refresh.

CHANGELOG.md
```diff
@@ -41,6 +41,10 @@ as necessary. Empty sections will not end in the release notes.
   specified explicitly in the JDBC URL.
   - `nessie.version.store.persist.jdbc.catalog`
   - `nessie.version.store.persist.jdbc.schema`
+- Support for token exchange in the Nessie client is deprecated and will be removed in a future
```
snazy, line 44: The token-exchange has actually been removed/replaced w/ the "proper" flows. Mind rewording this?

dimas-b: I'd say _will be replaced_... we do not have a functional (alternative) token exchange flow ATM, do we?

dimas-b (commented):
Code changes LGTM :+1:

dimas-b (approved):

PR #8803, opened 2024-06-12
OAuth2Client: introduce proper support for token exchange

adutra:
This PR introduces a generic support for token exchanges. The idea is that if enabled, each access token obtained will be exchanged for a another one.

I tried to keep the new `TokenExchangeConfig` API as flexible as possible, in order to allow even complex usages of token exchange, such as impersonation and delegation.

It is possible to use an alternate IdP for exchanging tokens, as well as alternate client ID/client secret. The default is to use the global IdP and credentials.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,16 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+- Support for token exchange in the Nessie client has been completely redesigned. The new API and
+  configuration options are described in the [Nessie authentication settings]. If this feature is
+  enabled, each time a new access token is obtained, the client will exchange it for another one by
+  performing a token exchange with the authorization server. We hope that this new feature will
+  unlock many advanced use cases for Nessie users, such as impersonation and delegation. Please note
+  that token exchange is considered in beta state and both the API and configuration options are
+  subject to change at any time; we appreciate early feedback, comments and suggestions.
+
+[Nessie authentication settings]: https://projectnessie.org/tools/client_config/#authentication-settings
+
 ### Changes
 
 ### Deprecations
@@ -44,15 +54,6 @@ as necessary. Empty sections will not end in the release notes.
   on our web site projectnessie.org for more information.
 - CEL access check scripts now receive the variable `roles` that can be used to check whether the current
   principal has a role assigned using a CEL expression like `'rolename' in roles`.
-- Support for token exchange in the Nessie client has been completely redesigned. The new API and
-  configuration options are described in the [Nessie authentication settings]. If this feature is
-  enabled, each time a new access token is obtained, the client will exchange it for another one by
-  performing a token exchange with the authorization server. We hope that this new feature will 
-  unlock many advanced use cases for Nessie users, such as impersonation and delegation. Please note 
-  that token exchange is considered in beta state and both the API and configuration options are 
-  subject to change at any time; we appreciate early feedback, comments and suggestions.
-
-[Nessie authentication settings]: https://projectnessie.org/tools/client_config/#authentication-settings
 
 ### Deprecations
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -44,6 +44,15 @@ as necessary. Empty sections will not end in the release notes.
   on our web site projectnessie.org for more information.
 - CEL access check scripts now receive the variable `roles` that can be used to check whether the current
   principal has a role assigned using a CEL expression like `'rolename' in roles`.
+- Support for token exchange in the Nessie client has been completely redesigned. The new API and
```
snazy, line 47: It's in the wrong spot - need to move it "up" for the next release.

CHANGELOG.md
```diff
@@ -55,13 +64,14 @@ as necessary. Empty sections will not end in the release notes.
...
   - `nessie.version.store.persist.jdbc.catalog`
   - `nessie.version.store.persist.jdbc.schema`
-- Support for token exchange in the Nessie client, in its current form, is now deprecated. The token
-  exchange flow was being used to exchange a refresh token for an access token, but this is not its
-  recommended usage. From now on, if a refresh token is provided by the authorization server, the
-  Nessie client will use the `refresh_token` grant type to obtain a new access token; if a refresh
-  token is not provided, the client will use the configured initial grant type to fetch a new access
-  token. _This change should thus be transparent to all users._ The `token_exchange` flow will be
-  redesigned in a future release to support more advanced use cases.
+- As mentioned above, token exchange in the Nessie client has been redesigned. As part of this
```
snazy, line 67: Same - need to move it up

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -294,9 +294,136 @@ public final class NessieConfigConstants {
    */
   @Deprecated
```
snazy, line 295: BTW/FYI: There's a javadoc tag (or annotation?) to hide this, if you want to.

snazy: Or do we want to rewrite it automatically to the new name?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
```diff
@@ -73,7 +75,10 @@ class OAuth2Client implements OAuth2Authenticator, Closeable {
         config.getGrantType().requiresUserInteraction()
             ? CompletableFuture.allOf(started, used)
             : started;
-    currentTokensStage = ready.thenApplyAsync((v) -> fetchNewTokens(), executor);
+    currentTokensStage =
+        ready
+            .thenApplyAsync((v) -> fetchNewTokens(), executor)
+            .thenApply(this::maybeExchangeTokens);
```
snazy, line 81: Do we need `thenApplyAsync` here (and below)? I wonder, because the call to `fetchNewTokens` is async.

snazy (commented):
Not a full review. The changes look good from a brief look, left a few comments/questions.

CHANGELOG.md
adutra: Let's keep it here because it's not new, but I will rephrase slightly because otherwise it's mentioning a "future release" but we already know it's going to be the next one.

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
adutra: To hide what?

About automatic remapping: I thought about it but I prefer not to. The settings have similar names but completely different behaviors.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Client.java
adutra: Async is not needed: the code will be executed by whatever thread executed the previous step, so here: an executor thread, which is what we want.

adutra:
FYI I found this resource from Forgerock / PingIdentity useful, it presents many use cases for delegation and impersonation:

https://backstage.forgerock.com/docs/idcloud/latest/am-oauth2/token-exchange.html

CHANGELOG.md
snazy: Nope - this is the already released 0.90.2 version - cannot change release notes for a published version. Needs to go up.

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
snazy: The setting in the generated docs.

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -291,12 +291,140 @@ public final class NessieConfigConstants {
...
+      "nessie.authentication.oauth2.token-exchange.scopes";
+
+  /**
+   * For token exchanges only. The subject token to exchange.
+   *
+   * <p>By default, the client will use its current access token as the subject token. But if this
+   * property is set, the client will use the static token provided here instead.
+   */
+  @ConfigItem(section = "OAuth2 Authentication")
+  public static final String CONF_NESSIE_OAUTH2_TOKEN_EXCHANGE_SUBJECT_TOKEN =
```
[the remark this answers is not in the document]
adutra: Yes if you pass a token this way, it will be a static one. But if you don't pass anything, the client will use its current access token as the subject token – which is probably what most users will want to do.

PR #8831, opened 2024-06-14
Helm chart: support for REST Catalog

### LINKED ISSUE #8737: [Catalog] Update Helm chart
- [ ] Add configuration options for warehouse + object storage
- [ ] Update docs on web site

We're probably good with what we have in the Helm chart.

However, we'd need a way to "mount" an(other) `application.properties` file into the containers due to the potential amount of configuration options when using many object stores/warehouses.

adutra:
Finally ready for review.

Quite frankly, I think we are reaching the limits of what a Helm chart can do.

The only test is the smoke test done by chart-testing, so it's hard to have a high confidence that all catalog configurations are properly supported. I think we need to look into improving these tests, e.g. by creating a complete environment to test against, including secrets, databases, ingresses, etc.

Fixes #8737


helm/nessie/ci/inmemory-values.yaml
```diff
@@ -4,3 +4,64 @@ tracing:
...
+    s3:
+      region: us-west-2
+      endpoint: s3.amazonaws.com
+      transport:
+        readTimeout: PT1M
+      buckets:
+        - name: bucket1.prod-us
+          endpoint: prod-us.s3.amazonaws.com
+          region: us-east-1
+#          accessKeySecret:
```
adutra, line 36: Need to create the secrets beforehand, I suggest that we look into it in another PR.

helm/nessie/templates/_helpers.tpl
```diff
@@ -92,3 +92,133 @@ names should coincide with jdbc schemes in connection URIs.
...
+{{- $_ = set $map ( printf "nessie.catalog.warehouses.%s.iceberg-config-overrides.%s" ( quote $warehouse.name ) $k ) $v -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Apply S3 catalog options.
+*/}}
+{{- define "nessie.applyCatalogStorageS3Options" -}}
```
adutra, line 126: These templates apply to both the default config, and the per-bucket configs. But note that some options are only supported at the default level. If they are supplied at per-bucket level, the chart would install but the pod would fail to start with SmallRye complaining that the option is not mapped.

helm/nessie/templates/deployment.yaml
```diff
@@ -129,6 +129,116 @@ spec:
...
+                  key: {{ .Values.catalog.storage.s3.accessKeySecret.awsAccessKeyId }}
+            - name: NESSIE_CATALOG_SERVICE_S3_ACCESS_KEY_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.catalog.storage.s3.accessKeySecret.name }}
+                  key: {{ .Values.catalog.storage.s3.accessKeySecret.awsSecretAccessKey }}
+            {{- end }}
+            {{- range $bucket := .Values.catalog.storage.s3.buckets -}}
+            {{- if $bucket.accessKeySecret }}
+            - name: NESSIE_CATALOG_SERVICE_S3_BUCKETS__{{- $bucket.name -}}__ACCESS_KEY_NAME
```
adutra, line 146: I had to manually validate that this works. We can't put those names into upper-case letters because they are meant to be map keys, and a bucket name can be something like `bucket-1.prod-us`.

Note that such environment variable names are not Posix compliant (IEEE Std 1003.1-2001 says: uppercase letters, digits, and the '_' (underscore) sign only). But this doesn't seem to be a problem for the process running Nessie.

helm/nessie/values.yaml
```diff
@@ -119,6 +119,330 @@ bigtable:
...
     # key: sa_json
 
+# -- The Nessie catalog server configuration.
+catalog:
+
+  # -- Whether to enable the REST catalog service.
+  enabled: false
+
+  # -- Iceberg catalog settings.
+  iceberg:
```
adutra, line 129: The general grouping is tentative, if you feel options could be grouped otherwise, I'm happy to change.

helm/nessie/templates/deployment.yaml
adutra: At this point, I wonder if we shouldn't switch all our env vars to their "original" names. WDYT?

helm/nessie/templates/_helpers.tpl
```diff
@@ -92,3 +92,133 @@ names should coincide with jdbc schemes in connection URIs.
 {{- $v := . | split ":" -}}
 {{ $v._1 }}
 {{- end }}
+
+{{/*
+Apply Iceberg catalog options.
```
snazy, line 97:
```suggestion
Apply Nessie Catalog (Iceberg REST) options.
```

helm/nessie/templates/_helpers.tpl
adutra: The ones in these classes:

* `org.projectnessie.catalog.files.s3.S3Config`
* `org.projectnessie.catalog.files.adls.AdlsConfig`

helm/nessie/templates/deployment.yaml
adutra: Yes I've seen that. My fear was double-quotes in var names: we don't have any in the docker compose examples.
But I did a manual test and even that seems to work.
That said, OK I will change the new vars in this PR but let's maybe change the old ones in another PR.

helm/nessie/values.yaml
```diff
@@ -119,6 +119,330 @@ bigtable:
...
+
+    # -- The default warehouse name (must refer to a declared warehouse below).
+    defaultWarehouse: warehouse1
```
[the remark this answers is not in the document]
adutra: It can. I went for a mock value because this property is required. But I don't mind changing.

helm/nessie/templates/deployment.yaml
snazy: SGTM

helm/nessie/values.yaml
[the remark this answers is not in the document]
snazy: Yup - it's required. Just not sure whether `warehouse1` is what people want. But any default value we choose is probably not the "right" one ;) So, I don't actually know :shrug:

helm/nessie/templates/_helpers.tpl
```diff
@@ -92,3 +92,181 @@ names should coincide with jdbc schemes in connection URIs.
...
+{{- include "nessie.secretToEnv" (list $bucket.accessKeySecret "awsSecretAccessKey" (printf "nessie.catalog.service.s3.buckets.\"%s\".access-key.secret" $bucket.name)) }}
+{{- end -}}
+{{- include "nessie.secretToEnv" (list .gcs.authCredentialsJsonSecret "key" "nessie.catalog.service.gcs.auth-credentials-json.key") }}
+{{- include "nessie.secretToEnv" (list .gcs.oauth2TokenSecret "token" "nessie.catalog.service.gcs.oauth-token.token") }}
+{{- include "nessie.secretToEnv" (list .gcs.oauth2TokenSecret "expiresAt" "nessie.catalog.service.gcs.oauth-token.expiresAt") }}
+{{- range $bucket := .gcs.buckets -}}
+{{- include "nessie.secretToEnv" (list $bucket.authCredentialsJsonSecret "key" (printf "nessie.catalog.service.gcs.buckets.\"%s\".auth-credentials-json.key" $bucket.name)) }}
+{{- include "nessie.secretToEnv" (list $bucket.oauth2TokenSecret "token" (printf "nessie.catalog.service.gcs.buckets.\"%s\".oauth-token.token" $bucket.name)) }}
+{{- include "nessie.secretToEnv" (list $bucket.oauth2TokenSecret "expiresAt" (printf "nessie.catalog.service.gcs.buckets.\"%s\".oauth-token.expiresAt" $bucket.name)) }}
```
adutra, line 242: I have a doubt regarding this option: I need to check whether the key ends with `expiresAt` or whether it should be input as `expires-at`.

snazy: Default naming would yield 'expires-at`

dimas-b (commented):
The PR LGTM, although I cannot say I understand all the templating magic :sweat_smile: I'm fine merging as soon as @snazy 's comments are resolved.

helm/nessie/templates/_helpers.tpl
adutra: You are right it's rendered `expires-at`.

adutra:
>  I cannot say I understand all the templating magic

Quite frankly it's a nightmare to implement – hours chasing the missing break line that screws up your YAML :-D

snazy (approved):
I suspect we're at a point where maintaining a helm chart becomes a real nightmare, and more work than maintaining an operator.

Maybe we should "close" development on the helm chart and gradually move over to the operator? WDYT?

adutra:
> we should "close" development on the helm chart and gradually move over to the operator? WDYT?

I like the idea 👍

snazy:
> > we should "close" development on the helm chart and gradually move over to the operator? WDYT?
> 
> I like the idea 👍

Maybe start a discussion on our ML and refer to it on Zulip?

adutra:
> Maybe start a discussion on our ML and refer to it on Zulip?

Not sure about the ML – it's been quite idle recently. Would a GitHub discussion work better maybe? https://github.com/projectnessie/nessie/discussions

snazy:
WFM - but ping on Zulip as well.

PR #8847, opened 2024-06-17
 OAuth2Client: differentiate token exchange from impersonation

adutra:
Ready for review.

I am proposing that we distinguish properties specific to the token exchange grant type, from properties that are only relevant in the context of impersonation/delegation.

Docs were updated accordingly.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/ImpersonationConfig.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/ImpersonationConfig.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.client.auth.oauth2;
+
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_OAUTH2_IMPERSONATION_CLIENT_ID;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_OAUTH2_IMPERSONATION_CLIENT_SECRET;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_OAUTH2_IMPERSONATION_ENABLED;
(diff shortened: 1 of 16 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/ImpersonationFlow.java
```diff
@@ -0,0 +1,81 @@
...
+
+  ImpersonationFlow(OAuth2ClientConfig config) {
+    super(config);
+    impersonationConfig = config.getImpersonationConfig();
+  }
+
+  @Override
+  public Tokens fetchNewTokens(Tokens currentTokens) {
+    Tokens newTokens = super.fetchNewTokens(currentTokens);
+    return Tokens.of(newTokens.getAccessToken(), currentTokens.getRefreshToken());
```
adutra, line 38: Here we keep: the exchanged access token + the _original_ refresh token (discarding any refresh token sent by the impersonation IdP, if any). I believe this is the correct behavior in all cases.

As for expiration: for all purposes, the validity of the exchanged token is the authoritative expiration time. The client will schedule a refresh at `min ( exchanged_token.expires_at, refresh_token == null ? NEVER : refresh_token.expires_at )`.

site/in-dev/client_config.md
```diff
@@ -178,40 +210,116 @@ nessie.server.authentication.enabled=true
...
+
+1. Initial token exchange: enabled when `authentication.oauth2.grant-type` is `token_exchange`. In
+   this scenario, the client will use token exchange as the primary grant. A subject token must be
+   provided with `nessie.authentication.oauth2.token-exchange.subject-token`. Other properties under
+   `nessie.authentication.oauth2.token-exchange.*` may also be required.
+
+2. Impersonation or delegation: this is the most typical usage, enabled when
+   `nessie.authentication.oauth2.impersonation.enabled` is `true`. Here, the client will first
+   obtain an initial token using another grant type, then exchange the received access token for
+   another access token, possibly from a second OAuth2 server. If a second OAuth2 server must be
```
dimas-b, line 318: nit: It might be worth mentioning that the admin user is responsible for configuring the trust relationship between the two OAuth2 servers (for the sake of clarity).

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/ImpersonationConfig.java
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Configuration for OAuth2 impersonation.
+ *
+ * <p>STILL IN BETA. API MAY CHANGE.
+ */
+@Value.Immutable
+public interface ImpersonationConfig {
+
+  List<String> SCOPES_INHERIT = Collections.singletonList("\\inherit\\");
```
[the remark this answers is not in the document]
adutra: The idea was to define an _invalid_ scope name – backslashes are not valid. This way if this scope was accidentally sent to the server it would reply with an error.

site/in-dev/client_config.md
adutra: https://github.com/projectnessie/nessie/pull/8928

PR #8848, opened 2024-06-17
Support v2 iceberg tables with gzip

jgprogramming:
Iceberg saves metadata with "gz.metadata.json" suffix when compressed with gzip for v2 tables and with "metadata.json.gz" for v1 tables. Current implementation recognizes only v1 gz tables and fails for v2.

### DIFF
--- a/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
+++ b/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
@@ -109,7 +109,7 @@ final class ImportSnapshotWorker {
       StorageUri metadataLocation = StorageUri.of(content.getMetadataLocation());
       try {
         InputStream input = taskRequest.objectIO().readObject(metadataLocation);
-        if (metadataLocation.requiredPath().endsWith(".gz")) {
+        if (metadataLocation.requiredPath().endsWith("metadata.json.gz") || metadataLocation.requiredPath().endsWith(".gz.metadata.json")) {
           input = new GZIPInputStream(input);
         }
         tableMetadata = IcebergJson.objectMapper().readValue(input, IcebergTableMetadata.class);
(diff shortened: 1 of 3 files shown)


catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
```diff
@@ -109,7 +109,7 @@ private EntitySnapshotObj.Builder importIcebergTable(
       StorageUri metadataLocation = StorageUri.of(content.getMetadataLocation());
       try {
         InputStream input = taskRequest.objectIO().readObject(metadataLocation);
-        if (metadataLocation.requiredPath().endsWith(".gz")) {
+        if (metadataLocation.requiredPath().endsWith("metadata.json.gz") || metadataLocation.requiredPath().endsWith(".gz.metadata.json")) {
```
dimas-b, line 112: Would you mind keeping the first case more generic as `.endsWith(".gz")` to be consistent with old behaviour?

Could you add a test for the second case?

dimas-b (commented):
Thanks for the contribution @jgprogramming !

The fix LGTM in general, just a couple of minor comments.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/ImportSnapshotWorker.java
dimas-b: Maybe in `AbstractIcebergCatalogTests`?

jgprogramming:
instead of creating new test added two new cases to existing test:
- for v1 format
- for v2 format
- for generic format (previous case with just .gz at the end)

Hope now it's ok, let me know what you think

dimas-b (approved):
Thanks again @jgprogramming !

The PR LGTM. @snazy WDYT?

snazy (approved):

snazy:
LGTM! Thanks for the contribution!

Issue #8832, opened 2024-06-14
Test failure: testAuthorizedDeviceCode() FAILED :nessie-quarkus:test  org.projectnessie.server.TestOAuth2Authentication


snazy:
Sample [run](https://github.com/projectnessie/nessie/actions/runs/9510459678/job/26214987357) in the `Exercise Java version (21)` job:
```
org.projectnessie.client.http.HttpClientException: OAuth2 server replied with HTTP status code: 404	
at org.projectnessie.client.auth.oauth2.OAuth2ClientConfig.genericError(OAuth2ClientConfig.java:307)	
at org.projectnessie.client.auth.oauth2.OAuth2ClientConfig.checkErrorResponse(OAuth2ClientConfig.java:286)	
at org.projectnessie.client.http.impl.jdk11.JavaRequest.lambda$executeRequest$1(JavaRequest.java:145)	
at java.util.Collections$SingletonList.forEach(Collections.java:5186)	
at org.projectnessie.client.http.impl.jdk11.JavaRequest.executeRequest(JavaRequest.java:145)	
at org.projectnessie.client.http.HttpRequest.post(HttpRequest.java:116)	
at org.projectnessie.client.http.HttpRequest.postForm(HttpRequest.java:131)	
at org.projectnessie.client.auth.oauth2.AbstractFlow.invokeEndpoint(AbstractFlow.java:102)	
at org.projectnessie.client.auth.oauth2.AbstractFlow.invokeTokenEndpoint(AbstractFlow.java:60)	
at org.projectnessie.client.auth.oauth2.DeviceCodeFlow.pollForNewTokens(DeviceCodeFlow.java:156)	
at org.projectnessie.client.auth.oauth2.DeviceCodeFlow.lambda$fetchNewTokens$0(DeviceCodeFlow.java:106)	
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)	
at java.util.concurrent.FutureTask.run(FutureTask.java:317)	
at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)	
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)	
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)	
at java.lang.Thread.run(Thread.java:1583)
```

Output:
```

[...]

PR #8861, opened 2024-06-19
OAuth2 client: improve error reporting

adutra:
This change will help diagnose #8832.


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2Utils.java
```diff
@@ -58,11 +56,6 @@ public static JsonNode fetchOpenIdProviderMetadata(HttpClient httpClient, URI is
     for (String path : WELL_KNOWN_PATHS) {
       try {
         HttpResponse response = httpClient.newRequest(issuerUrl).path(path).get();
-        Status status = response.getStatus();
```
adutra, line 61: Already checked by `OAuth2ClientConfig.checkErrorResponse()`

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
```diff
@@ -16,17 +16,17 @@
...
 import java.net.URI;
+import org.projectnessie.client.rest.io.CapturingInputStream;
 
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
   Status getResponseCode() throws IOException;
 
-  InputStream getInputStream() throws IOException;
+  CapturingInputStream getInputStream() throws IOException;
```
adutra, line 27: This is technically a breaking change, although I seriously doubt anybody implemented this interface outside of Nessie.

api/client/src/main/java/org/projectnessie/client/http/HttpClientResponseException.java
```diff
@@ -15,16 +15,56 @@
...
+  }
+
+  public HttpClientResponseException(String message, URI uri, Status status, String body) {
     super(message);
+    this.uri = uri;
     this.status = status;
+    this.body = body;
+  }
+
+  public URI getUri() {
```
snazy, line 44: Don't see a production code usage of this getter.

api/client/src/main/java/org/projectnessie/client/http/HttpClientResponseException.java
```diff
@@ -15,16 +15,56 @@
...
+    return uri;
   }
 
   public Status getStatus() {
     return status;
   }
+
+  public Optional<String> getBody() {
```
snazy, line 52: Don't see a production code usage of this getter.

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
snazy, line 27: TBH, this change will affect _all_ requests, not just the ones for OAuth, so I'm really not a fan of this change. `OAuth2ClientConfig` already does a `new CapturingInputStream()` - so it would be a `CapturingInputStream` around a `CapturingInputStream`?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -281,19 +282,25 @@ HttpClient getHttpClient() {
...
-            ErrorResponse errorResponse = getObjectMapper().readValue(is, ErrorResponse.class);
-            throw new OAuth2Exception(status, errorResponse);
-          } catch (IOException ignored) {
-            throw genericError(status);
+      // The only normal responses are 200 OK and 302 Found (in the authorization code flow).
+      if (status != Status.OK && status != Status.FOUND) {
+        String body = null;
+        if (responseContext.getErrorStream() != null) {
+          try (CapturingInputStream capturing =
+              new CapturingInputStream(responseContext.getErrorStream())) {
```
snazy, line 290: `responseContext.getErrorStream()` should be memoized

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
adutra: Sorry I forgot to change `OAuth2ClientConfig` after changing `ResponseContext`.

That said, you have a point that the cost in space of this change is probably too high.

api/client/src/main/java/org/projectnessie/client/http/HttpClientResponseException.java
adutra: There isn't; it's exposed for callers outside of the nessie client.

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
adutra: I reverted that part of the change and reduced the scope of this change to just the oauth client.

PR #8864, opened 2024-06-19
Add docs and docker-compose illustrating the use of a reverse proxy

snazy:


site/docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,146 @@
...
+quarkus.http.proxy.proxy-address-forwarding=true
+quarkus.http.proxy.allow-x-forwarded=true
+quarkus.http.proxy.enable-forwarded-host=true
+quarkus.http.proxy.enable-forwarded-prefix=true
+```
+
+!!! warn
+    Do NOT enable the above options unless your reverse proxy (for example istio or nginx)
+    is properly setup to set these headers but also filter those from incoming requests.
+    Make sure to understand what the actual settings mean. Misconfigurations can lead to security issues!
```
adutra, line 26: Repeated warning.

helm/nessie/values.yaml
```diff
@@ -141,6 +141,27 @@ advancedConfig:
 #      console.format: "%d{HH:mm:ss} %s%e%n"
 #      category."org.projectnessie".level: DEBUG
 
+# Reverse Proxy Settings
```
adutra, line 144: It's not clear whether this is a new section, or the continuation of the `advancedConfig` section above.

adutra (commented):
Very nice and helpful!

helm/nessie/values.yaml
```diff
@@ -141,6 +141,27 @@ advancedConfig:
...
+# Do NOT enable these option unless your reverse proxy (for example istio or nginx)
+# is properly setup to set these headers but also filter those from incoming requests.
+#
+#  quarkus:
+#   http:
+#     proxy:
+#       proxy-address-forwarding=true
+#       allow-x-forwarded=true
+#       enable-forwarded-host=true
+#       enable-forwarded-prefix=true
```
dimas-b, line 162: The prefix does not work to the best of my knowledge. Did you manage to have a working example with a prefix in the reverse proxy? I'd prefer to note that explicitly if it is indeed not working well.

https://github.com/projectnessie/nessie/pull/8674#issuecomment-2145309722

dimas-b (commented):
LGTM overall, but I have a concern with URI prefixes.

helm/nessie/values.yaml
snazy: Let me check

snazy: Works with a simple fix now. Updated the example to use `/nessie/` as the prefix.

servers/quarkus-server/src/main/java/org/projectnessie/server/catalog/ExternalBaseUriImpl.java
```diff
@@ -17,17 +17,35 @@
...
+    URI base = uriInfo.getBaseUri();
+    if (enableForwardedPrefix) {
+      // Use the value of the `X-Forwarded-Prefix` as the "prefix", if present and only if
+      // `quarkus.http.proxy.enable-forwarded-prefix` is set to `true`.
+      String prefix = httpHeaders.getHeaderString("X-Forwarded-Prefix");
+      if (prefix != null && !prefix.isEmpty()) {
+        if (!prefix.endsWith("/")) {
+          prefix += "/";
+        }
+        base = base.resolve(prefix);
```
dimas-b, line 46: We assume that Quarkus did not take the prefix into account, but I guess that's the best we can do ATM.

Note: this may work differently when using non-reactive REST, but I did not dig to the bottom of this on the Quarkus side.

Would you mind adding a comment to point to the related Quarkus code?

site/docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,161 @@
...
+!!! info
+    Passing the `X-Forwarded-Proto`, `X-Forwarded-Host` and `X-Forwarded-Port` headers to Nessie works for
+    Iceberg REST.
+
+The follow config options are mentioned only for documentation purposes. Consult the
+[Quarkus documentation](https://quarkus.io/guides/http-reference#reverse-proxy) for "Running behind a reverse proxy"
+and configure those depending on your actual needs.
+```properties
+quarkus.http.proxy.proxy-address-forwarding=true
+quarkus.http.proxy.allow-x-forwarded=true
```
dimas-b, line 18: Quarkus proxy docs make the impression that the standardized `Forwarded` header may be preferred. However, it does not provide the same level of [functionality](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) as the `X-Forwarded-*` headers together (specifically port and prefix cannot be defined in `Forwarded`) . I think it might be worth adding a paragraph to clarify that. WDYT?

site/docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+!!! warn
+    Do NOT enable the above options unless your reverse proxy (for example istio or nginx)
+    is properly setup to set these headers but also filter those from incoming requests.
+
+!!! info
+    The newer `Forwarded` header is the newer version of the older `X-Forwarded-For`, `X-Forwarded-Host`,
+    `X-Forwarded-Proto` headers, but does _not_ provide a replacement for the `X-Forwarded-Prefix` header,
+    which is needed in some scenarios.
```
dimas-b, line 31: There's no replacement for `X-Forwarded-Port` too, AFAIK. In a way, the port is more important. With TLS termination, a difference between external and internal port numbers is unavoidable, I think.

dimas-b: The `by` directive of the `Forwarded` header seems like it could be used to propagate the external port information to the server, but it does not appear to be interpreted by Quarkus [code](https://github.com/quarkusio/quarkus/blob/cdc1637bebb289320c36c2c859691746751e65c8/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/ForwardedParser.java#L129) ATM.

I believe it would be worth mentioning that `X-Forwarded-Port` is preferred from the Nessie perspective to avoid surprises on the users' side.

I might be missing something but then again our examples use onfy `X-Forwarded-*` headers.

dimas-b (approved):

Issue #8860, opened 2024-06-18
[Catalog] Unable to recover from running with invalid credentials


dimas-b:
### Issue description

If the Catalog Server is started with invalid credentials and makes some attempts to fetch Iceberg metadata from storage, it remembers those failures and is unable to recover when it is restarted with _valid_ credentials.

----

Use case:

1) Run Nessie Server with permanent Nessie storage (e.g. PosgreSQL) and _invalid_ storage (e.g. S3) credentials

2) Run Spark using the basic Nessie Core API

```
$ bin/spark-sql \
 --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.5.0,org.apache.iceberg:iceberg-aws-bundle:1.5.0 \
 --conf spark.sql.catalog.nessie.uri=http://localhost:19120/api/v2 \
 --conf spark.sql.catalog.nessie.ref=main \
 --conf spark.sql.catalog.nessie.catalog-impl=org.apache.iceberg.nessie.NessieCatalog \
 --conf spark.sql.catalog.nessie.warehouse=s3://EXAMPLE \
 --conf spark.sql.catalog.nessie.io-impl=org.apache.iceberg.io.ResolvingFileIO \
 --conf spark.sql.catalog.nessie=org.apache.iceberg.spark.SparkCatalog \
  --master 'local[1]'
```

3) Create an Iceberg table
```
spark-sql (default)> create namespace nessie.ns
                   > ;
Time taken: 2.297 seconds
spark-sql (default)> create table nessie.ns.t3(x int);
Time taken: 1.68 seconds
spark-sql (default)> insert into nessie.ns.t3 values (123);
Time taken: 2.985 seconds
spark-sql (default)> select * from nessie.ns.t3;
123
Time taken: 1.6 seconds, Fetched 1 row(s)
```

4) Run Spark using the REST Catalog API
```
$ bin/spark-sql \
 --packages org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.5.0,org.apache.iceberg:iceberg-aws-bundle:1.5.0 \
 --conf spark.sql.catalog.nessie.uri=http://localhost:19120/iceberg \
 --conf spark.sql.catalog.nessie.type=rest \
 --conf spark.sql.catalog.nessie=org.apache.iceberg.spark.SparkCatalog
```

5) Try to select from the test table:
```
spark-sql (default)> select * from nessie.ns.t3;
[...]

PR #8869, opened 2024-06-20
Add admin command: delete-catalog-tasks

dimas-b:
This is to help admin users recover from #8860

* Move `CatalogObjTypeBundle` to catalog-service-common

* Add new admin command: `DeleteCatalogTasks`

* Move some helper methods in `BaseCommand`

### DIFF
--- a/tools/server-admin/build.gradle.kts
+++ b/tools/server-admin/build.gradle.kts
@@ -160,7 +160,7 @@ if (Os.isFamily(Os.FAMILY_MAC) && System.getenv("CI") != null) {
 tasks.named<Test>("intTest").configure {
   // Reduce likelihood of OOM due to too many Quarkus apps in memory;
   // Ideally, set this number to the number of IT classes to run for each backend.
-  forkEvery = 4
+  forkEvery = 5
   // Optional; comma-separated list of backend names to test against;
   // see NessieServerAdminTestBackends for valid values.
   systemProperty("backends", System.getProperty("backends"))
--- a/tools/server-admin/build.gradle.kts
+++ b/tools/server-admin/build.gradle.kts
@@ -55,6 +55,11 @@ dependencies {
   implementation(project(":nessie-versioned-storage-mongodb"))
   implementation(project(":nessie-versioned-storage-rocksdb"))
 
+  implementation(project(":nessie-catalog-service-common"))
+  implementation(project(":nessie-tasks-api"))
+
+  implementation(libs.guava)
+
   implementation(enforcedPlatform(libs.quarkus.bom))
   implementation("io.quarkus:quarkus-core-deployment")
   implementation(enforcedPlatform(libs.quarkus.amazon.services.bom))
--- a/tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/DeleteCatalogTasks.java
+++ b/tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/DeleteCatalogTasks.java
@@ -100,7 +100,10 @@ public class DeleteCatalogTasks extends BaseCommand {
 
     spec.commandLine()
         .getOut()
-        .printf("Deleted %d snapshot task object(s) in total.%n", idsProcessed.get());
+        .printf(
+            "Deleted %d snapshot task object(s) in total.%n%n"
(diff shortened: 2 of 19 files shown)


snazy (commented):
Instead of completely removing, I'd prefer to have options to delete only those that are in a particular state like `TaskStatus.FAILURE` or maybe `TaskStatus.ERROR_RETRY` or convert `TaskStatus.FAILURE` into `TaskStatus.ERROR_RETRY`.

It should also be clear to users that such an operation should really happen while all Nessie servers are down (caching).

tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/ExpireSnapshotTasks.java
```diff
@@ -0,0 +1,142 @@
...
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.versionstore.ContentMapping;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "expire-snapshot-tasks",
+    mixinStandardHelpOptions = true,
+    description =
+        "Delete all state of Iceberg snapshot loading tasks previously executed by the Nessie Catalog.")
+public class ExpireSnapshotTasks extends BaseCommand {
```
adutra, line 52: Imho the name is a bit misleading. Initially I thought this would run some GC expire-snapshot job. Maybe `DeleteIcebergTasks` or `DeleteCatalogTasks`?

adutra (commented):
Should we consider making failures due to invalid credentials retriable? Maybe with a large retry period?

dimas-b:
> Should we consider making failures due to invalid credentials retriable?

Yes. I'm planning to do that under #8558, but I thought that a recovery tool for older servers might be worth having too.

tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/ExpireSnapshotTasks.java
dimas-b: I'll rename to `DeleteCatalogTasks`

tools/server-admin/build.gradle.kts
```diff
@@ -55,6 +55,11 @@ dependencies {
   implementation(project(":nessie-versioned-storage-mongodb"))
   implementation(project(":nessie-versioned-storage-rocksdb"))
 
+  implementation(project(":nessie-catalog-service-common"))
+  implementation(project(":nessie-tasks-api"))
+
+  implementation(libs.guava)
+
```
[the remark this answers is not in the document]
dimas-b: updated

tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/ExpireSnapshotTasks.java
dimas-b: renamed

tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/DeleteCatalogTasks.java
```diff
@@ -0,0 +1,176 @@
...
+          "Elements or a specific content key to process (zero or more). "
+              + "If not set, all current keys will get their snapshot tasks expired.")
+  private List<String> keyElements;
+
+  @CommandLine.Option(
+      names = {"-s", "--task-status"},
+      description =
+          "Delete tasks having these statuses (zero or more). "
+              + "If not set, only failed tasks for matching content objects are deleted.")
+  private List<String> statuses;
```
snazy, line 76:
```suggestion
  private List<TaskStatus> statuses;
```

snazy:
Don't forget to update the commit message / PR description

tools/server-admin/src/main/java/org/projectnessie/tools/admin/cli/DeleteCatalogTasks.java
dimas-b: Good point - fixed

snazy (approved):

PR #8870, opened 2024-06-20
Disable Iceberg GC by default

snazy:


adutra (approved):

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1TableResource.java
```diff
@@ -158,6 +159,13 @@ private Uni<IcebergLoadTableResponse> loadTable(TableRef tableRef, String prefix
       R loadTableResultFromSnapshotResponse(
           SnapshotResponse snap, B builder, String prefix, ContentKey contentKey) {
     IcebergTableMetadata tableMetadata = (IcebergTableMetadata) snap.entityObject().orElseThrow();
+    if (!tableMetadata.properties().containsKey(GC_ENABLED)) {
```
ajantha-bhat, line 162: we should also disable `TableProperties.METADATA_DELETE_AFTER_COMMIT_ENABLED`

reference:
https://github.com/apache/iceberg/blob/601efaf309712eca21228f8949813133606a42b8/nessie/src/main/java/org/apache/iceberg/nessie/NessieCatalog.java#L74

snazy: Can you show me how that is important with REST?

ajantha-bhat: True. `RESTTableOperations` doesn't implement cleaning the metadata files as
table metadata file is optional for REST catalogs.

PR #8879, opened 2024-06-20
Update nessie-IDs in returned metadata properties, simplify some code

### LINKED ISSUE #8734: [Catalog] Ensure and verify that we always return the absolute Nessie reference with a snapshot
Nessie Catalog should always return the absolute Nessie reference (aka: reference name, if available, and commit ID) in the metadata and as a HTTP header for both Nessie Catalog and Iceberg REST endpoints.

snazy:
Fixes #8734 (was already fine, just simplified the properties)

### DIFF
--- a/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/IcebergConstants.java
+++ b/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/IcebergConstants.java
@@ -18,7 +18,7 @@ package org.projectnessie.catalog.formats.iceberg.nessie;
 import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 
-final class IcebergConstants {
+public final class IcebergConstants {
   private IcebergConstants() {}
 
   /**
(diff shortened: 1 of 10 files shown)


catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/IcebergConstants.java
```diff
@@ -76,5 +85,11 @@ private IcebergConstants() {}
...
           CURRENT_SCHEMA,
           DEFAULT_PARTITION_SPEC,
-          DEFAULT_SORT_ORDER);
+          DEFAULT_SORT_ORDER,
+          NESSIE_CONTENT_ID,
+          NESSIE_COMMIT_ID,
+          NESSIE_COMMIT_REF);
+
+  public static final Set<String> TRANSIENT_PROPERTIES =
+      ImmutableSet.of(NESSIE_CONTENT_ID, NESSIE_COMMIT_ID, NESSIE_COMMIT_REF);
```
dimas-b, line 94: Not sure about the "transient" term here... Content and Commit IDs never change or go away :thinking: How about `INTERNAL_PROPERTIES`?

snazy: Transient in the sense that these properties are removed/ignored when received but added (so not stored)

dimas-b: Still "transient" looks confusing to me in this context. How about "derived" properties?

dimas-b (approved):

PR #8889, opened 2024-06-21
Committing operations: cleanup metadata after failures

### LINKED ISSUE #8728: [Catalog] Committing operations: cleanup metadata after failures

snazy:
Ensures that objects/files written during a commit are cleaned up if either one of the object/file write operations or the Nessie commit operation fails.

Refactors `CatalogServiceImpl.MultiTableUpdate` to be a top-level type and adds a list of successfully written objects/files in a separate `exceptionally()` stage, that calls `ObjectIO.deleteObjects()` for the written objects/files.

Fixes #8728

### DIFF
--- a/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
+++ b/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
@@ -55,7 +55,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalInt;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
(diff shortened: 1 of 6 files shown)


catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -415,6 +417,31 @@ CompletionStage<MultiTableUpdate> commit(ParsedReference reference, CatalogCommi
...
+        // applyIcebergTable/ViewCommitOperation().
+        .exceptionally(
+            e -> {
+              if (e instanceof CompletionException) {
+                e = e.getCause();
+              }
+              if (e.getClass() == RuntimeException.class
+                  && e.getMessage().equals(e.getCause().toString())) {
+                e = e.getCause();
+              }
```
adutra, line 431: Why this? We don't do anything with the unwrapped exception. Also `exceptionally` is not designed to throw exceptions: I think `whenComplete` would be more appropriate here since you don't need to unwrap the exception.

snazy: The code is just more boilerplate w/ `whenComplete` (null-check), but the exceptions are exactly the same - same unwrapping etc.
The handling for exceptions in `java.util.concurrent.CompletableFuture#uniExceptionally` and `java.util.concurrent.CompletableFuture#uniWhenComplete` is also the same.

adutra: More boilerplate? I think it's less:

```java
        .whenComplete(
            (result, error) -> {
              if (error != null) {
                try {
                  objectIO.deleteObjects(
                      multiTableUpdate.storedLocations().stream()
                          .map(StorageUri::of)
                          .collect(toList()));
                } catch (Exception e) {
                  error.addSuppressed(e);
                }
              }
            })
```

snazy: You still get a CompletionException there. The code is the same - but with the added `if (e != null)`.

snazy: Try it - run TestCatalogServiceImpl w/ a breakpoint there - there's a CE

adutra: And what is the problem with having a CE here? BTW the tests pass with `whenComplete`.

snazy: The other way around: what's the problem with `exceptionally`, since the code's _less_ boilerplate because of the `if (e != null)`?

snazy: Your proposal would wrap a `CompletionException` in another `CompletionException`, right?

snazy: Oh - I see now. You're relyong on `addSuppressed` here

adutra (approved):

PR #8893, opened 2024-06-21
Expose object store(s) availability as a readiness check

### LINKED ISSUE #8833: [Catalog] Check object store connections
We should have a way to check all object store configurations.

Options:
* during startup - this may block startup for too long, OTOH ensures that object stores are already initialized
* using an administrative endpoint - as a health/readiness check

(Not fully thought through - but something like this should help / improve UX)

snazy:
Sample of a successful check:
```json
{ "status": "UP",
  "checks": [
    {
      "name": "Warehouses Object Stores",
      "status": "UP",
      "data": {
        "warehouse.warehouse.object-store.availability": true
      }
} ] }
```

Sample of a failed check (with a wrong S3 access key):
```json
{ "status": "DOWN",
  "checks": [ {
    "name": "Warehouses Object Stores",
    "status": "DOWN",
    "data": {
      "warehouse.warehouse.object-store.availability": false,
      "warehouse.warehouse.object-store.error": "java.io.IOException: software.amazon.awssdk.services.s3.model.S3Exception: The Access Key Id you provided does not exist in our records. (Service: S3, Status Code: 403, Request ID: 17DB0B619B94C358, Extended Request ID: dd9025bab4ad464b049177c95eb6ebf374d3b3fd1af9251148b658df7ac2e3e8)"
  } } ] }
```

Fixes #8833

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
@@ -46,6 +46,16 @@ public class S3ObjectIO implements ObjectIO {
   }
 
   @Override
+  public void ping(StorageUri uri) throws IOException {
+    S3Client s3client = s3clientSupplier.getClient(uri);
+    try {
+      s3client.headBucket(b -> b.bucket(uri.requiredAuthority()));
+    } catch (RuntimeException e) {
+      throw new IOException(e);
+    }
+  }
+
+  @Override
   public InputStream readObject(StorageUri uri) throws IOException {
     checkArgument(uri != null, "Invalid location: null");
     String scheme = uri.scheme();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ as necessary. Empty sections will not end in the release notes.
 - Nessie's metrics now support custom, user-defined tags (dimensional labels). To 
   define a custom tag, set the `nessie.metrics.tags.<tag-name>=<tag-value>` configuration property.
   Such tags are added to all metrics published by Nessie.
+- Readiness/health check testing the general availability of the object stores configured for the warehouses.
 
 ### Changes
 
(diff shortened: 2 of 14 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
```diff
@@ -45,6 +45,16 @@ public S3ObjectIO(S3ClientSupplier s3clientSupplier, Clock clock) {
     this.clock = clock;
   }
 
+  @Override
+  public void ping(StorageUri uri) throws IOException {
+    S3Client s3client = s3clientSupplier.getClient(uri);
+    try {
+      s3client.getBucketLocation(b -> b.bucket(uri.requiredAuthority()));
```
dimas-b, line 52: This is fine from my POV, but it probably requires extra permissions in AWS.

dimas-b (approved):

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ObjectIO.java
snazy: For the bucket-location? Uh

dimas-b: Also `S3Client` javadoc has: `This operation is not supported by directory buckets.`

snazy: Oh! there's a `headBucket` anyways

snazy: rebased + updated this one

dimas-b: +1 to `headBucket`. MinIO does not appear to support `GetBucketLocaltion`: https://min.io/docs/minio/linux/reference/s3-api-compatibility.html

dimas-b (approved):

PR #8923, opened 2024-06-24
Migrate off deprecated `AwsS3V4Signer` to `AwsV4HttpSigner`

snazy:


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Signer.java
```diff
@@ -66,43 +64,38 @@ public SigningResponse sign(SigningRequest clientRequest) {
...
+
+    SignRequest.Builder<AwsCredentialsIdentity> signRequest =
+        SignRequest.builder(credentials)
+            .request(request.build())
+            .putProperty(AwsV4HttpSigner.REGION_NAME, clientRequest.region())
+            .putProperty(AwsV4HttpSigner.SERVICE_SIGNING_NAME, S3Client.SERVICE_NAME)
+            .putProperty(AwsV4HttpSigner.DOUBLE_URL_ENCODE, false)
+            .putProperty(AwsV4HttpSigner.NORMALIZE_PATH, false)
+            .putProperty(AwsV4HttpSigner.CHUNK_ENCODING_ENABLED, false)
+            .putProperty(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false);
```
adutra, line 88: Should we enforce the algorithm?
```suggestion
            .putProperty(AwsV4HttpSigner.PAYLOAD_SIGNING_ENABLED, false)
            .putProperty(AwsV4HttpSigner.CHECKSUM_ALGORITHM, DefaultChecksumAlgorithm.SHA256);
```

snazy: That's been there before. 🤷

snazy: And the suggested change actually breaks the functionality :(

adutra (approved):

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Signer.java
snazy: BTW: It uses SHA256 (I looked into the headers)

PR #8937, opened 2024-06-26
Don't use reference-style links in CHANGELOG.md

adutra:
...as they don't appear as links in GitHub releases page.


snazy (changes requested):
Let keep it as it is.
We should really never change the release notes of a published release, even if there's a bug in the links.

PR #8953, opened 2024-06-27
Fix STDOUT in `ConnectCommand`

dimas-b:
Properly restore the saved output stream.

Follow-up to #8885

### DIFF
--- a/cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/ConnectCommand.java
+++ b/cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/ConnectCommand.java
@@ -150,7 +150,7 @@ public class ConnectCommand extends NessieCommand<ConnectCommandSpec> {
             icebergClient = new RESTCatalog();
             icebergClient.initialize("iceberg", icebergProperties);
           } finally {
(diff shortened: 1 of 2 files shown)


cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/ConnectCommand.java
```diff
@@ -150,7 +150,7 @@ public void execute(@Nonnull BaseNessieCli cli, ConnectCommandSpec spec) throws
             icebergClient = new RESTCatalog();
             icebergClient.initialize("iceberg", icebergProperties);
           } finally {
-            System.setErr(out);
+            System.setOut(out);
```
snazy, line 153: :facepalm:

snazy (approved):

cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/ConnectCommand.java
snazy: Who wrote this? And who reviewed it?

snazy (approved):

PR #8955, opened 2024-06-27
Fix a bunch of IntelliJ inspections, rather nits + cosmetic

snazy:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
```diff
@@ -103,7 +103,7 @@ class AuthorizationCodeFlow extends AbstractFlow {
         redirectUriFuture
             .thenApply(this::extractAuthorizationCode)
             .thenApply(this::fetchNewTokens)
-            .whenComplete(this::log);
+            .whenComplete((tokens, error) -> log(error));
```
dimas-b, line 106: What's the warning here? Old code LGTM :thinking:

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/meta/IcebergPartitionSpec.java
```diff
@@ -123,7 +123,7 @@ default boolean compatibleWith(IcebergPartitionSpec other) {
       IcebergPartitionField thisField = fields().get(i);
       IcebergPartitionField thatField = other.fields().get(i);
       if (thisField.sourceId() != thatField.sourceId()
-          || !thisField.transform().toString().equals(thatField.transform().toString())
+          || !thisField.transform().equals(thatField.transform())
```
dimas-b, line 126: This does not look cosmetic to me :)

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -156,7 +156,6 @@ public Stream<Supplier<CompletionStage<SnapshotResponse>>> retrieveSnapshots(
               try {
                 snapshotId = snapshotIdFromContent(c.getContent());
               } catch (Exception e) {
-                e.printStackTrace();
```
dimas-b, line 159: Should we re-throw `e` here? I know this is beyond the cosmetic cleanup, but since it caught our attention, we might as well fix it, perhaps? Also, I do not think an exception is actually expected in this case... WDYT?

api/client/src/main/java/org/projectnessie/client/auth/oauth2/AuthorizationCodeFlow.java
snazy: `tokens` parameter removed - so IJ refactored it to a lambda

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/meta/IcebergPartitionSpec.java
snazy: `String.toString()` is :)

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
snazy: But not in this PR

snazy: Meh - added a logging here

adutra (approved):

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/meta/IcebergPartitionSpec.java
dimas-b: Ah, ok... I only looked at the textual diff :)

PR #8963, opened 2024-06-28
Fix potential class-loader deadlock via `Namespace.EMPTY`

### LINKED ISSUE #8961: Possible class loader deadlock in `Namespace`
`org.projectnessie.model.Namespace#EMPTY` references subclass `ImmutableNamespace`, which can lead to a class loader deadlock.

snazy:
Fixes #8961

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Namespace.java
+++ b/api/model/src/main/java/org/projectnessie/model/Namespace.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.model;
 
+import static org.projectnessie.model.Namespace.Empty.EMPTY_NAMESPACE;
 import static org.projectnessie.model.Util.DOT_STRING;
 import static org.projectnessie.model.Util.FIRST_ALLOWED_KEY_CHAR;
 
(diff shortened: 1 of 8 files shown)


api/model/src/main/java/org/projectnessie/model/NessieConstants.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.model;
+
+/** Utility class holding some constants, mostly to prevent potential class loader deadlocks. */
+public final class NessieConstants {
+  public static final Namespace EMPTY_NAMESPACE = Namespace.builder().build();
```
dimas-b, line 20: This could probably be in a `static` inner class in `Namespace`... or do you prefer a separate class?

snazy: That would not solve the issue, right?

dimas-b: But it would. A `static` inner class is mostly equivalent to a top-level class (no extra loading dependencies), except it's namespaced under the outer class.

snazy: That's not much different then. I thought you wanted to propose class-loading-lazy initialization from a static method in Namespace.

api/model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -48,6 +49,18 @@ public abstract class Namespace extends Content {
...
+
+    public static final Namespace EMPTY_NAMESPACE = builder().build();
+  }
+
+  /**
+   * Refactor all code references to this constant to use {@link Empty#EMPTY_NAMESPACE}, there's a
+   * non-zero risk of causing a class loader deadlock.
+   */
+  @Deprecated
   public static final Namespace EMPTY = builder().elements(Collections.emptyList()).build();
```
dimas-b, line 64: One more nittish comment, but this is public API, so I hope it might be cleaner for outside callers:

We could probably have in `Namespace`:
```
public static Namespace emptyNamespace() {
   return Empty.EMPTY_NAMESPACE;
}
```
... then `Empty` can be `private`... WDYT?

snazy: Nah - let's keep it as is

dimas-b (approved):

PR #8966, opened 2024-06-28
Site: add docs on Kubernetes troubleshooting

adutra:


dimas-b (approved):
Nice doc addition :+1:

snazy (dismissed):

site/docs/guides/kubernetes.md
```diff
@@ -332,3 +332,153 @@ To uninstall the Helm chart and delete the `nessie` release from the `nessie-ns`
...
+cp /root/.bashrc /home/debug/ && chown 185:0 /home/debug/.bashrc 
+su - debug
+```
+
+After running the above commands, you should be able to use `jps`, `jcmd`, `jmap`, etc. as well as
+other tools like `async-profiler`, `jfr` (Java Flight Recorder), etc. For example, you can check the
+Nessie server's JVM threads with `jcmd`, or run `async-profiler` to profile the Nessie server:
+
+```bash
+async-profiler -d 30 -f profile.html 1
```
adutra, line 455: Sorry for dismissing your approvals, I'm trying to validate this example but async-profiler seems buggy on an arm64 machine. Switching to amd64 to see if it's better.

adutra: Well, async-profiler in this version seems broken. I will switch the example to JFR instead which seems to be working fine.

adutra: Ready for review again.

dimas-b (approved):

site/docs/guides/kubernetes.md
adutra: Ideally we could publish our own debug image with the right user (185), all the tools (CLI, admin) and all the utilities both for network and JVM diagnostics. But that's enough for now :-)

PR #8983, opened 2024-07-01
Add `returnNotFound` for `VersionStore.getValue(s)`

snazy:
This change is a prerequisite for #8768

### DIFF
--- a/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
+++ b/versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
@@ -382,20 +382,33 @@ public interface VersionStore {
    *
    * @param ref Any ref type allowed
    * @param key The key for the specific value
-   * @return The value.
+   * @param returnNotFound whether to return a non-{@code null} result object with a {@code null}
+   *     value in {@link ContentResult#content()} instead of a {@code null} return-value for
+   *     non-existing content
+   * @return The value or {@code null} if the key does not exist and {@code returnNotFound} is
+   *     {@code false}. If the content does not exist and {@code returnNotFound} is {@code true},
+   *     {@link ContentResult#content()} will be {@code null}.
    * @throws ReferenceNotFoundException if {@code ref} is not present in the store
    */
-  ContentResult getValue(Ref ref, ContentKey key) throws ReferenceNotFoundException;
+  ContentResult getValue(Ref ref, ContentKey key, boolean returnNotFound)
+      throws ReferenceNotFoundException;
 
   /**
    * Get the values for a list of keys.
    *
    * @param ref The ref to use.
    * @param keys An ordered list of keys to retrieve within the provided ref.
-   * @return A parallel list of values.
+   * @param returnNotFound whether to return non-{@code null} values with a {@code null} value in
+   *     {@link ContentResult#content()} instead of omitting non-existing content in the returned
+   *     map
+   * @return Map of keys to content results. For keys that do not exist: if {@code returnNotFound}
+   *     is {@code false}, the map will not contain an entry for non-existing keys.If {@code
+   *     returnNotFound} is {@code true}, the map will return entries for all requested keys, but
+   *     {@link ContentResult#content()} will be {@code null} for non-existing keys.
    * @throws ReferenceNotFoundException if {@code ref} is not present in the store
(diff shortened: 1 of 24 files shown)


versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
```diff
@@ -382,20 +382,33 @@ List<IdentifiedContentKey> getIdentifiedKeys(Ref ref, Collection<ContentKey> key
...
+   * @param returnNotFound whether to return a non-{@code null} result object with a {@code null}
+   *     value in {@link ContentResult#content()} instead of a {@code null} return-value for
+   *     non-existing content
+   * @return The value or {@code null} if the key does not exist and {@code returnNotFound} is
+   *     {@code false}. If the content does not exist and {@code returnNotFound} is {@code true},
+   *     {@link ContentResult#content()} will be {@code null}.
    * @throws ReferenceNotFoundException if {@code ref} is not present in the store
    */
-  ContentResult getValue(Ref ref, ContentKey key) throws ReferenceNotFoundException;
+  ContentResult getValue(Ref ref, ContentKey key, boolean returnNotFound)
```
adutra, line 393: Shouldn't we either:

1. keep the old methods as delegates to the new ones? Could also minimize the impact since existing callers wouldn't need to be aware of this change.
2. Or, conversely, generalize the new behavior and adapt existing callers? (Might be a bit dangerous though, since it would be a behavioral change without any binary change).

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -919,6 +921,15 @@ public ContentResult getValue(Ref ref, ContentKey key) throws ReferenceNotFoundE
     }
   }
 
+  private static ContentResult getValueNotFound(
+      ContentKey key, boolean returnNotFound, StoreIndex<CommitOp> index) {
+    if (!returnNotFound) {
+      return null;
+    }
+    IdentifiedContentKey identifiedKey = buildIdentifiedKey(key, index, null, null, x -> null);
+    return contentResult(identifiedKey, null, null);
```
adutra, line 930: Slightly more readable:

```suggestion
    if (returnNotFound) {
      IdentifiedContentKey identifiedKey = buildIdentifiedKey(key, index, null, null, x -> null);
      return contentResult(identifiedKey, null, null);
    }
    return null;
```

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
```diff
@@ -962,28 +973,41 @@ static IdentifiedContentKey buildIdentifiedKey(
...
+                              e.getValue(),
+                              null)));
+      if (returnNotFound) {
+        keys.stream()
+            .filter(k -> !result.containsKey(k))
+            .forEach(
+                key -> {
+                  IdentifiedContentKey identifiedKey =
+                      buildIdentifiedKey(key, index, null, null, x -> null);
+                  result.put(key, contentResult(identifiedKey, null, null));
```
adutra, line 1007: 1. This is relying on `ContentMapping.fetchContents` returning a mutable map.
2. Wouldn't it be more efficient to pass `returnNotFound` down to `fetchContents`?

versioned/spi/src/main/java/org/projectnessie/versioned/VersionStore.java
snazy: Nah - it's a "boring" change, and we'd have to carry the "older" one (1).

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/VersionStoreImpl.java
snazy: Not sure, but the code here can be refactored.

snazy: I think, it looks easier w/o the Java stream dance

adutra (approved):

PR #8986, opened 2024-07-01
Docs: reuse Nessie CLI history in docker examples

adutra:
...and use host network for easier connectivity.

Also: automatically remove docker containers for unstable images when the container exits.

### DIFF
--- a/site/docs/downloads/index.md
+++ b/site/docs/downloads/index.md
@@ -73,20 +73,23 @@ The main Nessie server serves the Nessie repository using the Iceberg REST API a
 === "Docker Image"
 
     Docker images are multiplatform images for amd64, arm64, ppc64le, s390x.
+
     They are available from the following repositories:
-     
+    
     * [GitHub Container Registry](https://github.com/projectnessie/nessie/pkgs/container/nessie-cli):
 
     ```bash
     docker pull ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
-    docker run -it ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
+    docker create -it --name nessie-cli --network=host -v $HOME/.nessie:/home/default/.nessie ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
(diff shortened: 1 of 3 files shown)


site/docs/downloads/index.md
```diff
@@ -73,20 +73,23 @@ The main Nessie server serves the Nessie repository using the Iceberg REST API a
...
+
     They are available from the following repositories:
-     
+    
     * [GitHub Container Registry](https://github.com/projectnessie/nessie/pkgs/container/nessie-cli):
 
     ```bash
     docker pull ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
-    docker run -it ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
+    docker create -it --name nessie-cli --network=host -v $HOME/.nessie:/home/default/.nessie ghcr.io/projectnessie/nessie-cli:{{ versions.nessie }}
```
snazy, line 83: Why not just `run` in one command?

snazy, line 83: Hm - the problem here is that the `$HOME/.nessie/' directory and/or the file `$HOME/.nessie:/home/default/.nessie` will be get the ID of the user running the process in the container with mode `644`, not the user's "natural" user ID / group ID.

If the user ID of the image changes or the user wants to use the history running Nessie CLI "directly" via the jar, the history file will never be updated again.

adutra: To reuse the same container. Since it's potentially a container that we want to reuse many times, there is no point in creating one container each time.

adutra: Are you observing this on linux? Because on mac everything is working fine, my history file has my own user id after running nessie-cli. Or is it a problem with podman?

snazy: That's for all Linux based containers.

adutra: I am not observing that at all.

adutra:
Closing as it seems it doesn't work as expected on Linux.

PR #8987, opened 2024-07-01
Catalog S3: use DefaultCredentialsProvider if no access key configured

adutra:
Using DefaultCredentialsProvider as a last resort allows leveraging instance profile credentials delivered through the Amazon EC2 metadata service.

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -103,15 +103,6 @@ public interface S3BucketOptions {
    * <ul>
    *   <li>{@code DEFAULT}: Default credentials provider chain.
    *   <li>{@code STATIC}: Static credentials provided through the {@code access-key} option.
-   *   <li>{@code ENV_VARIABLE}: Loads credentials from the {@code AWS_ACCESS_KEY_ID}, {@code
-   *       AWS_SECRET_ACCESS_KEY} and {@code AWS_SESSION_TOKEN} environment variables.
-   *   <li>{@code SYSTEM_PROPERTY}: Loads credentials from the system properties {@code
-   *       aws.accessKeyId} and {@code aws.secretAccessKey}.
-   *   <li>{@code PROFILE}: Loads credentials from AWS profile files.
-   *   <li>{@code INSTANCE_PROFILE}: Loads credentials from the Amazon EC2 metadata service.
-   *   <li>{@code CONTAINER}: Loads credentials from a local metadata service on ECS or AWS
-   *       Greengrass.
-   *   <li>{@code ANONYMOUS}: Anonymous access.
    * </ul>
    */
   Optional<S3ServerAuthenticationMode> serverAuthenticationMode();
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -101,7 +101,7 @@ public interface S3BucketOptions {
    * <p>Valid values are:
    *
    * <ul>
-   *   <li>{@code DEFAULT}: Default credentials provider chain.
+   *   <li>{@code APPLICATION_GLOBAL}: Default credentials provider chain.
    *   <li>{@code STATIC}: Static credentials provided through the {@code access-key} option.
    * </ul>
    */
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -123,16 +123,6 @@ public interface S3BucketOptions {
   Optional<BasicCredentials> accessKey();
 
   /**
-   * The profile name to use when loading AWS credentials from the standard AWS configuration files.
-   *
-   * <p>Required when {@code server-authentication-mode} is {@code PROFILE}.
-   *
-   * <p>For STS, this defines the profile name to be used as a basic credential provider for
-   * obtaining temporary session credentials.
-   */
-  Optional<String> profile();
-
-  /**
    * The <a href="https://docs.aws.amazon.com/STS/latest/APIReference/welcome.html">Security Token
    * Service</a> endpoint.
    *
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -101,7 +101,9 @@ public interface S3BucketOptions {
    * <p>Valid values are:
    *
    * <ul>
-   *   <li>{@code APPLICATION_GLOBAL}: Default credentials provider chain.
+   *   <li>{@code APPLICATION_GLOBAL}: Use the AWSSDK <a
+   *       href="https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html">default
+   *       credentials provider</a>.
    *   <li>{@code STATIC}: Static credentials provided through the {@code access-key} option.
    * </ul>
    */
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -101,14 +101,17 @@ public interface S3BucketOptions {
    * <p>Valid values are:
    *
    * <ul>
-   *   <li>{@code DEFAULT} - Default credentials provider chain.
-   *   <li>{@code STATIC} - Static credentials (access-key-id and secret-access-key).
-   *   <li>{@code ENV_VARIABLE} - Environment variables.
-   *   <li>{@code SYSTEM_PROPERTY} - System properties.
-   *   <li>{@code PROFILE} - AWS profile.
-   *   <li>{@code INSTANCE_PROFILE} - AWS Instance profile.
-   *   <li>{@code CONTAINER} - Container credentials.
-   *   <li>{@code ANONYMOUS} - Anonymous access.
+   *   <li>{@code DEFAULT}: Default credentials provider chain.
+   *   <li>{@code STATIC}: Static credentials provided through the {@code access-key} option.
+   *   <li>{@code ENV_VARIABLE}: Loads credentials from the {@code AWS_ACCESS_KEY_ID}, {@code
+   *       AWS_SECRET_ACCESS_KEY} and {@code AWS_SESSION_TOKEN} environment variables.
+   *   <li>{@code SYSTEM_PROPERTY}: Loads credentials from the system properties {@code
+   *       aws.accessKeyId} and {@code aws.secretAccessKey}.
+   *   <li>{@code PROFILE}: Loads credentials from AWS profile files.
+   *   <li>{@code INSTANCE_PROFILE}: Loads credentials from the Amazon EC2 metadata service.
+   *   <li>{@code CONTAINER}: Loads credentials from a local metadata service on ECS or AWS
+   *       Greengrass.
+   *   <li>{@code ANONYMOUS}: Anonymous access.
    * </ul>
    */
   Optional<S3ServerAuthenticationMode> serverAuthenticationMode();
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
@@ -33,10 +33,16 @@ public interface S3BucketOptions {
       Duration.ofHours(1); // 1 hour lifetime is common for session credentials in S3
 
   /**
+   * Default value for {@link #serverAuthenticationMode()}, being {@link
+   * S3ServerAuthenticationMode#STATIC}.
+   */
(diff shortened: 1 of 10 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Clients.java
```diff
@@ -19,6 +19,8 @@
 import org.projectnessie.catalog.secrets.BasicCredentials;
 import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
```
snazy, line 22: The reason why we do not have the `DefaultCredentialsProvider` here is that it can run for a really long time. This code here would run for every access to the credentials.

If we want to infer data from EC2 metadata, that should IMHO only happen for the default settings and only be done once.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Clients.java
```diff
@@ -38,29 +40,19 @@ public static SdkHttpClient apacheHttpClient(S3Config s3Config) {
...
-            () -> new IllegalStateException("Secret reference to S3 access key ID is not defined"));
-
-    String secret =
-        secretAccessKeyIdRef.orElseThrow(
-            () ->
-                new IllegalStateException(
-                    "Secret reference to S3 secret access key is not defined"));
-
-    return () -> AwsBasicCredentials.create(key, secret);
+      Optional<BasicCredentials> accessKey) {
```
snazy, line 43: The signature change for this function is totally fine!

dorsegal:
Will this work for eks IRSA/Pod Idendity?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Clients.java
adutra: Or maybe we could memoize the resolved credentials per bucket? That would also allow e.g. to print a warning message without spamming the logs.

snazy: No memoizing here, please.
We also need to consider how this change could conflict in case a secrets manager/vault is being use - e.g. when the "right" secret is added later.
Let's not rush this.

adutra:
> Will this work for eks IRSA/Pod Idendity?

@dorsegal I can't tell for sure I'm afraid, but it seems so: IRSA uses `WebIdentityTokenFileCredentialsProvider` according to this SO question:

https://stackoverflow.com/questions/70964807/aws-java-sdk-v2-not-using-irsa-iam-role-in-eks

And `DefaultCredentialsProvider` uses `WebIdentityTokenFileCredentialsProvider`. 

Do you think you could try this PR out in your environment?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Clients.java
adutra: Fair enough. That said I don't think each call to this method will pay the penalty of traversing the whole supplier chain because `reuseLastProviderEnabled` is true by default.

adutra:
Side note: GKE clusters have a similar way of binding roles to service accounts. Whatever solution we come up with here needs to be applied to GCS options as well. (and maybe to AKS / ADLS as well).

dorsegal:
> > Will this work for eks IRSA/Pod Idendity?
> 
> @dorsegal I can't tell for sure I'm afraid, but it seems so: IRSA uses `WebIdentityTokenFileCredentialsProvider` according to this SO question:
> 
> https://stackoverflow.com/questions/70964807/aws-java-sdk-v2-not-using-irsa-iam-role-in-eks
> 
> And `DefaultCredentialsProvider` uses `WebIdentityTokenFileCredentialsProvider`.
> 
> Do you think you could try this PR out in your environment?

@adutra I tested it with pod identity and it works. IRSA has the same mechanism so it will work with it as well.

olivier-derom:
Tested with IRSA and can confirm it works as well

snazy:
Actually, we could introduce a similar mechanism like auth-type (`GcsBucketOptions.GcsAuthType`)  to S3 and users can configure it explicitly to `WORKLOAD_IDENTITY`. WDYT?

olivier-derom:
So instead of making it a fallback option, making it an option in clientAuthenticationMode?
I like that idea as it will more clearly state the possibilities.

snazy:
> Actually, we could introduce a similar mechanism like auth-type (`GcsBucketOptions.GcsAuthType`) to S3 and users can configure it explicitly to `WORKLOAD_IDENTITY`. WDYT?

@adutra ^

adutra:
> > Actually, we could introduce a similar mechanism like auth-type (`GcsBucketOptions.GcsAuthType`) to S3 and users can configure it explicitly to `WORKLOAD_IDENTITY`. WDYT?
> 
> @adutra ^

As Olivier said, this idea has the nice benefit of making things explicit, so I'm +1. However it won't make `DefaultCredentialsProvider` any faster, will it? We'd still need to find a solution for that (cache / memoize / stg else).

snazy:
> As Olivier said, this idea has the nice benefit of making things explicit, so I'm +1. However it won't make `DefaultCredentialsProvider` any faster, will it? We'd still need to find a solution for that (cache / memoize / stg else).

The difference is that you explicitly tell Nessie to do it, so if you configure it explicitly, you have to live with the fact :)

adutra:
> The difference is that you explicitly tell Nessie to do it, so if you configure it explicitly, you have to live with the fact :)

Hm if people configure it that way, it may be because they don't have the choice. I don't think it's fair to tell them that in this specific setup, the server will become sluggish and there is nothing we can do about it.

snazy:
> the server will become sluggish and there is nothing we can do about it.

It's for credentials provided to the pod/VM, so those are there. If not, then it's bad. But then it's explicit.

olivier-derom:
What makes the process slow/messy/sluggish?
The DefaultCredentialsProvider Class in general, or the fact that you have to go through the other checks first and only use it as a last resort?
For the former, you could try having a look at how other applications handle it (such as Trino when its responsible for writing to S3 with nessie catalog).
For the latter, explicitly telling to use that, skipping to it and not using it as last resort can indeed speed it up.

I think that explicitly choosing this option implies that credentials are present, and could be fetched upon nessie-server startup and kept into memory and not be retrieved again for every transaction, so that would not hurt the overall performance, correct?

snazy:
> What makes the process slow/messy/sluggish? The DefaultCredentialsProvider Class in general, or the fact that you have to go through the other checks first and only use it as a last resort?

It _can_ be slow - e.g. when it tries to contact AWS internal endpoints. That would be surprising as a "fallback". But IMHO it's okay, if it's explicitly configured.

snazy:
Putting it behind a "memoizing supplier" is fine as well for it.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
```diff
@@ -224,24 +229,7 @@ Credentials sessionCredentialsForClient(String repositoryId, S3BucketOptions opt
   private static SessionKey buildSessionKey(String repositoryId, S3BucketOptions options) {
     // Client parameters are part of the credential's key because clients in different regions may
     // issue different credentials.
-    return ImmutableSessionKey.builder()
-        .repositoryId(repositoryId)
-        .region(
```
adutra, line 229: We need so many options that I figured it was easier to store the whole `S3BucketOptions` in the key. @dimas-b do you think that's ok in terms of cache key lookup efficiency?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Clients.java
```diff
@@ -118,31 +115,12 @@ public static SdkHttpClient apacheHttpClient(S3Config s3Config, SecretsProvider
...
-  }
-
   public static AwsCredentialsProvider awsCredentialsProvider(
       S3BucketOptions bucketOptions, S3Sessions sessions) {
-    Optional<String> role = bucketOptions.assumeRole();
-    if (role.isEmpty()) {
-      return basicCredentialsProvider(
-          bucketOptions.accessKey().map(BasicCredentials::name),
-          bucketOptions.accessKey().map(BasicCredentials::secret));
+    if (bucketOptions.assumeRole().isPresent()) {
```
adutra, line 120: @dimas-b is it possible to define a role to be assumed, but keep using `REQUEST_SIGNING` for clients? I suppose in this case credentials would be fetched and cached, and the server would use them to sign requests?

Asking because the Helm chart currently assumes that if a role is provided, then `ASSUME_ROLE` is used for clients. Maybe we need to change that?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3ServerAuthenticationMode.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+  },
+
+  STATIC {
+    @Override
+    public AwsCredentialsProvider newCredentialsProvider(S3BucketOptions bucketOptions) {
+      return bucketOptions
+          .accessKey()
+          // TODO support session token: AwsSessionCredentials.create(accessKeyId, secretAccessKey,
+          // sessionToken)
```
adutra, line 52: Should I provide support for this in this PR? This is supported in Quarkus. I guess in this case, the STS token is obtained prior to starting the server and assumed to be valid for long enough, "à la" bearer token auth.

helm/nessie/templates/_helpers.tpl
```diff
@@ -158,7 +158,9 @@ Apply S3 catalog options.
 {{- if .pathStyleAccess -}}{{- $_ := set $map ( print $prefix "path-style-access" ) .pathStyleAccess -}}{{- end -}}
 {{- if .accessPoint -}}{{- $_ := set $map ( print $prefix "access-point" ) .accessPoint -}}{{- end -}}
 {{- if .allowCrossRegionAccessPoint -}}{{- $_ := set $map ( print $prefix "allow-cross-region-access-point" ) .allowCrossRegionAccessPoint -}}{{- end -}}
-{{- if .clientAuthenticationMode -}}{{- $_ := set $map ( print $prefix "auth-mode" ) .clientAuthenticationMode -}}{{- end -}}
+{{- if .serverAuthenticationMode -}}{{- $_ := set $map ( print $prefix "server-auth-mode" ) .serverAuthenticationMode -}}{{- end -}}
+{{- if .clientAuthenticationMode -}}{{- $_ := set $map ( print $prefix "client-auth-mode" ) .clientAuthenticationMode -}}{{- end -}}
```
adutra, line 162: Not porting over the "legacy" `auth-mode` property since this is still more or less in beta. If someone disagrees I can add it.

snazy (commented):
Woot - wait...

All we need is an enum `S3CredentialsSource` analogous to `GcsAuthType` with two values:
* `PROVIDED`
* `DEFAULT`

`PROVIDED` (or whatever name) is what we have at the moment.
`APPLICATION_DEFAULT` would use `software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider`

`software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider` already tries all the sources.

`S3Clients.basicCredentialsProvider` (arguments should be `Optional<BasicCredentials> accessKey, S3CredentialsSource authType` or just `S3BucketOptions`) would then only need to evaluate the enum and either do what it did before or return the `DefaultCredentialsProvider.create()` (which actually returns a simple constant).

adutra:
Supporting other providers is not a problem imho, I basically copied everything from Quarkiverse. Other providers could be useful for users that know which type of authentication they want, as this spares the burden of going through the provider chain.

adutra:
As requested, I simplified the design to just 2 modes: default and static.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3BucketOptions.java
```diff
@@ -88,11 +94,31 @@ public interface S3BucketOptions {
...
   Optional<String> region();
 
+  /**
+   * The authentication mode to use by the Catalog server. If not set, the default is {@code
+   * STATIC}. Depending on the authentication mode, other properties may be required.
+   *
+   * <p>Valid values are:
+   *
+   * <ul>
+   *   <li>{@code DEFAULT}: Default credentials provider chain.
```
snazy, line 104: This feels a bit odd. Above it says, "the default is static" - but this value is named DEFAULT, which is not default.
Maybe rename `DEFAULT` to `APPLICATION_GLOBAL` or so.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
snazy: Just make sure that `hashCode` is not computed every time

servers/quarkus-common/src/main/java/org/projectnessie/quarkus/config/CatalogS3BucketConfig.java
```diff
@@ -61,7 +66,7 @@ public interface CatalogS3BucketConfig extends S3BucketOptions {
   @Override
   Optional<String> externalId();
 
-  @WithName("auth-mode")
+  @WithName("client-auth-mode")
```
snazy, line 69: This is a breaking change, deserves a note in CHANGELOG.md

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
adutra: `SessionKey` is annotated with `@NessieImmutable` so I think we are good.

snazy (approved):

PR #8984, opened 2024-07-01
Attempt to fix flaky `TestAddressResolver.resolveGoodName`

snazy:

PR #8992, opened 2024-07-02
Move `AddressResolve` to a separate module, so it's reusable

snazy:


snazy:
Note: this PR already includes the changes from #8984 - so it's in "draft" for now

tools/network/src/main/java/org/projectnessie/nessie/networktools/AddressResolver.java
```diff
@@ -38,13 +39,13 @@
 import org.slf4j.LoggerFactory;
 
 @VisibleForTesting
```
adutra, line 41: Could remove the annotation now.

adutra (approved):

PR #9001, opened 2024-07-03
Refactor `MultiTableUpdate`

snazy:
Refactors `CatalogServiceImpl.MultiTableUpdate` to be a top-level type and adds a list of successfully written objects/files in a separate `exceptionally()` stage, that calls `ObjectIO.deleteObjects()` for the written objects/files.

The change is unfortunately quite verbose.

Prerequisite for testing #8889 and #8768

### DIFF
--- a/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/MultiTableUpdate.java
+++ b/catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/MultiTableUpdate.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/MultiTableUpdate.java
```diff
@@ -0,0 +1,112 @@
...
+  private Branch targetBranch;
+  private boolean committed;
+
+  MultiTableUpdate(CommitMultipleOperationsBuilder nessieCommit, Branch target) {
+    this.nessieCommit = nessieCommit;
+    this.targetBranch = target;
+  }
+
+  MultiTableUpdate commit() throws NessieConflictException, NessieNotFoundException {
+    synchronized (this) {
```
adutra, line 48: The synchronized block was there already but I was wondering if we really need it?

adutra (approved):

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/MultiTableUpdate.java
snazy: Not really sure, but it might happen that different threads access the fiels (completion-stages)

PR #9004, opened 2024-07-04
Support using the default application credentials for google client

zymap:
---

### Motivation

In the kubernetes environment, we more like use workload identity to all a pod to access the google service. Using the default application credentials will scan all possible credentails to access the google service.

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java
@@ -23,6 +23,7 @@ import com.google.api.gax.retrying.RetrySettings;
 import com.google.auth.Credentials;
 import com.google.auth.http.HttpTransportFactory;
 import com.google.auth.oauth2.AccessToken;
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.OAuth2Credentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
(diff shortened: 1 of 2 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java
```diff
@@ -148,7 +149,12 @@ static Credentials buildCredentials(
                 oauth2token.expiresAt().map(i -> new Date(i.toEpochMilli())).orElse(null));
         return OAuth2Credentials.create(accessToken);
       default:
-        throw new IllegalArgumentException("Unsupported auth type " + authType);
+        try {
```
snazy, line 152: I'd rather leave the `default` case as it is but add a new enum value (e.g. `WORKLOAD_IDENTITY`) to `GcsBucketOptions.GcsAuthType`.

snazy (commented):
@zymap Thank you for your contribution!

I've left a comment on the PR.

adutra:
Related to #8987.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java
zymap: Sure

zymap: @snazy PTAL

snazy (approved):
LGTM

PR #9008, opened 2024-07-04
Catalog/Bug: Fix double-write for S3

snazy:
Writing (metadata) objects to S3 has a bug that makes it write the content twice, actually as often as `.close()` is called on the output stream.

It goes back to an somewhat unexpectable side effect of Jackson, which closes the output-stream when it writes an object (`ObjectMapper.writeValue(OutputStream,Object)` et al).

Thankfully, this "double metadata" doesn't have any negative effect, because OTOH a call like `ObjectMapper.readValue()` returns after the first deserialized object and ignores any following ones.


adutra (approved):
Good catch!

snazy:
I run into this so often :facepalm:

PR #9029, opened 2024-07-06
Fix older-client/server dependency resolution errors

snazy:
Maven-resolver now fails to resolve dependencies due to a profile activation error in `org.apache.maven.model.profile.activation.OperatingSystemProfileActivator#isActive` when it tries to resolve the system properties starting with `os.`. This change adds more "standard" system properties.

Fixes CI errors for #9024 and #9025


adutra (approved):
I don't fully understand the fix but I'll trust you :-)

PR #9030, opened 2024-07-06
Nit: move EntityObj ID calculation from impl class

snazy:
No functional change.


catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/EntitySnapshotObj.java
```diff
@@ -68,7 +68,7 @@ default ObjType type() {
           TaskObj.taskDefaultCacheExpire(),
           c -> ObjType.NOT_CACHED);
 
-  static ObjId snapshotIdFromContent(Content content) {
+  static ObjId snapshotObhIdFromContent(Content content) {
```
adutra, line 71: Small typo:

```suggestion
  static ObjId snapshotObjIdFromContent(Content content) {
```

Also, this method name uses "from" while the `entityObjIdForContent` method uses "for". Would be nice to use one or the other consistently.

adutra (approved):

PR #9032, opened 2024-07-06
WIP: Add ability to (de)serialize generic custom `ObjType`s

snazy:
Adds functionality for export/import to (de)serialize unregistered `ObjType`s, for example Nessie Catalog types.


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/GenericObjTypeBundle.java
```diff
@@ -0,0 +1,98 @@
...
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.storage.common.persist.ObjType;
+import org.projectnessie.versioned.storage.common.persist.ObjTypeBundle;
+import org.projectnessie.versioned.storage.common.persist.UpdateableObj;
+
+/** Registers a "fallback" handler to handle otherwise unhandled object types. */
+public class GenericObjTypeBundle implements ObjTypeBundle, ObjTypeBundle.ObjTypeMapper {
+  @Override
+  public void register(Consumer<ObjType> registrar) {
+    // no owh object types
```
adutra, line 48:
```suggestion
    // no own object types
```

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypes.java
```diff
@@ -55,11 +60,23 @@ public static ObjType forShortName(@Nonnull String shortName) {
   private static final class Registry {
     private static final Map<String, ObjType> BY_ANY_NAME;
     private static final Set<ObjType> OBJ_TYPES;
+    private static final List<ObjTypeBundle.ObjTypeMapper> OBJ_TYPE_MAPPERS;
+
+    static ObjType maybeMapped(String name) {
+      for (ObjTypeBundle.ObjTypeMapper mapper : OBJ_TYPE_MAPPERS) {
+        ObjType type = mapper.mapGenericObjType(name);
```
adutra, line 67: I'm not sure that this functionality should be part of the SPI. What if 2 or more bundles declare conflicting `ObjTypeMapper` instances? The result would be non-deterministic. There should imho be only one generic object type mapper, that would be invoked if all registered bundles failed to "understand" the object type.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypes.java
```diff
@@ -55,11 +60,23 @@ public static ObjType forShortName(@Nonnull String shortName) {
...
+        if (type != null) {
```
adutra, line 68: Is it even possible that `type` could be null here? Since `GenericObjTypeBundle` never returns null, and is always registered.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypes.java
snazy: Yea - we probably will only ever need just one. Updated.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypes.java
snazy: With the added commit, nope.

The original plan was that there can be multiple fallbacks, which could decide whether they support it or not. But processing order would likely make that non-deterministic.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypes.java
adutra: Looking better now, but is it really necessary to load the (only) generic object mapper via service-loader? Why not just instantiate it normally?

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/persist/ObjTypeMapper.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.storage.common.persist;
+
+import jakarta.annotation.Nonnull;
+
+public interface ObjTypeMapper {
```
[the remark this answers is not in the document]
snazy: Hm - let's leave it there. I find it more explicit than an "obscure" generic `java.util.functions.*` type.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/objtypes/GenericObjTypeMapper.java
```diff
@@ -0,0 +1,86 @@
...
+    @JsonIgnore
+    @JacksonInject(OBJ_VERS_KEY)
+    @Nullable
+    public abstract String versionToken();
+
+    @JsonAnyGetter
+    @JsonAnySetter
+    @AllowNulls
+    public abstract Map<String, Object> attributes();
+  }
```
adutra, line 81: I'm concerned that `GenericObj` implements `UpdateableObj`. I think that if it's generic, then it should implement only `Obj`. If the original object was updateable, I'd expect the version token to appear in the `attributes` property bag.

I understand that you might have done this because the version token is an injectable value, but there is a way to capture the injected value and include it in the property bag:

```suggestion
  @Value.Immutable
  @Value.Style(jdkOnly = true)
  abstract static class GenericObj implements Obj {
    @Override
    @JsonIgnore
    public abstract ObjType type();

    @Override
    @JsonIgnore
    public abstract ObjId id();

    @JsonAnyGetter
    @AllowNulls
    public abstract Map<String, Object> attributes();

    @JsonCreator
    static GenericObj create(
        @JacksonInject(OBJ_TYPE_KEY) ObjType objType,
        @JacksonInject(OBJ_ID_KEY) ObjId id,
        @JacksonInject(OBJ_VERS_KEY) String versionToken,
        @JsonAnySetter Map<String, Object> attributes) {
      ImmutableGenericObj.Builder builder =
          ImmutableGenericObj.builder().type(objType).id(id).attributes(attributes);
      if (versionToken != null) {
        builder.putAttributes("versionToken", versionToken);
      }
      return builder.build();
    }
  }
```

snazy: It must implmenent UpdateableObj so that the `Persist` implementations set the field in the database appropriately.

adutra: Conceptually, not every generic object is updateable. What is sad is that now `UpdateableObj.versionToken()` became nullable which is confusing.

Wouldn't it be better to transfer the burden of extracting the version token to `Persist` implementations? Something like:

```java
    if (obj instanceof UpdateableObj) {
      item.put(COL_OBJ_VERS, fromS(((UpdateableObj) obj).versionToken());
    }
    if (obj instanceof GenericObj && ((GenericObj)obj).hasAttribute("versionToken")) {
      item.put(COL_OBJ_VERS, fromS(((GenericObj)obj).getAttribute("versionToken"));
    }
```

(The above can be moved to a utility class of course.)

snazy: I know it's not great. Happy to open a follow-up issue for this.

adutra: OK, approving then. I can look into it.

adutra (approved):

PR #9033, opened 2024-07-06
Ability to identify related objects for Nessie export

snazy:
Nessie export needs to be able to identify which objects are related to the repository, references, commits and contents that are being exported.

This service-loader based functionality allows Nessie export to identify those objects. Each relevant module provides an implementation of the `TransferRelatedObjects` interface, ideally in an artifact that has no "superfluous" dependencies, so only the code that is required to compute the `ObjId`s. This change includes two implementations of `TransferRelatedObjects`, one to provide the `ObjId` of the `UniqueIdObj` for each `Content` and one for Nessie Catalog.

### DIFF
--- a/catalog/service/transfer/src/main/java/org/projectnessie/catalog/service/objtypes/transfer/CatalogTransferRelated.java
+++ b/catalog/service/transfer/src/main/java/org/projectnessie/catalog/service/objtypes/transfer/CatalogTransferRelated.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


catalog/service/transfer/src/main/java/org/projectnessie/catalog/service/objtypes/transfer/CatalogTransferRelated.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Set;
+import org.projectnessie.model.Content;
+import org.projectnessie.versioned.storage.common.persist.ObjId;
+import org.projectnessie.versioned.transfer.related.TransferRelatedObjects;
+
+public class CatalogTransferRelated implements TransferRelatedObjects {
+  @Override
+  public Set<ObjId> contentRelatedObjects(Content content) {
+    ObjId entityId = CatalogObjIds.entityIdForContent(content);
+    ObjId snapshotId = CatalogObjIds.entityIdForContent(content);
```
adutra, line 29:
```suggestion
    ObjId snapshotId = CatalogObjIds.snapshotIdForContent(content);
```

snazy: Argh - I fixed that. but only in #9034

PR #9045, opened 2024-07-09
Fix behavior of metadata-update/set-statistics + set-partition-statistics

### LINKED ISSUE #9044: [Bug]: REST Catalog creates duplicate statistics files references in table metadata (may be Trino-only/related)
### What happened

As far as I can tell, this happens when running successive commits against a table, each with a `set-statistics` metadata update operation, for example by running multiple `INSERT`s from Trino. When writing the Iceberg metadata files, Nessie will reuse (?) data from the previous Iceberg snapshot and append the new statistics file reference to that, resulting in two conflicting references in the metadata JSON.

### How to reproduce it

I have encountered this issue through Trino, so reproduction steps are for that, but this should hypothetically apply directly to multiple commits with `set-statistics` operations them.

In Trino, running this script (with an `iceberg` connector hooked up to the Nessie REST compatibility layer) will surface the issue. The third insert ought to fail:

```sql
CREATE SCHEMA iceberg.default;
CREATE TABLE IF NOT EXISTS iceberg.default.repro (a int);
INSERT INTO iceberg.default.repro (a) VALUES (1);
INSERT INTO iceberg.default.repro (a) VALUES (2);
INSERT INTO iceberg.default.repro (a) VALUES (3);
```

Docker, Nessie version 0.91.3

Trino 451

### Additional information

I have tried investigating the root cause of this myself and how to potentially fix it as well. It appears the culprit is this line:
https://github.com/projectnessie/nessie/blob/cfb20bf0c9654aac60f66a5372c19e307451d901/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java#L237

If I understand things correctly, that should indeed override the entire statistics file list rather than adding to it, similar to the `remove-statistics` implementation. For our use case, this would certainly work correctly, but I'm not sure if this applies generally, so I'm holding off on submitting a patch.

snazy:
`o.p.catalog.formats.iceberg.rest.IcebergMetadataUpdate.SetStatistics` wrongly _adds_ the given statistics file, but should instead _set_ it as the only one, which is the behavior of `org.apache.iceberg.MetadataUpdate.SetStatistics`. Similarly for `SetPartitionStatistics`.

Fixes #9044

### DIFF
--- a/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
+++ b/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
@@ -17,6 +17,7 @@ package org.projectnessie.catalog.formats.iceberg.rest;
 
 import static com.google.common.base.Preconditions.checkState;
 import static java.util.Collections.emptyList;
+import static java.util.Collections.singleton;
 import static org.projectnessie.catalog.formats.iceberg.nessie.NessieModelIceberg.icebergStatisticsFileToNessie;
 
 import com.fasterxml.jackson.annotation.JsonAlias;
@@ -234,7 +235,7 @@ public interface IcebergMetadataUpdate {
     default void applyToTable(IcebergTableMetadataUpdateState state) {
       long snapshotId = Objects.requireNonNull(state.snapshot().icebergSnapshotId());
       if (snapshotId == snapshotId()) {
-        state.builder().addStatisticsFile(icebergStatisticsFileToNessie(statistics()));
+        state.builder().statisticsFiles(singleton(icebergStatisticsFileToNessie(statistics())));
       }
     }
   }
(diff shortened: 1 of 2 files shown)


catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
```diff
@@ -234,7 +235,7 @@ interface SetStatistics extends IcebergMetadataUpdate {
     default void applyToTable(IcebergTableMetadataUpdateState state) {
       long snapshotId = Objects.requireNonNull(state.snapshot().icebergSnapshotId());
       if (snapshotId == snapshotId()) {
-        state.builder().addStatisticsFile(icebergStatisticsFileToNessie(statistics()));
+        state.builder().statisticsFiles(singleton(icebergStatisticsFileToNessie(statistics())));
```
ajantha-bhat, line 238: Similar problem exist for partition statistics file below at line 274. 

Also, how can we catch them with tests? Is the unit tests missing?

adutra (approved):

ajantha-bhat (approved):

PR #9050, opened 2024-07-10
Compatibility filter: bypass filters when contacting the config endpoint

adutra:


api/client/src/main/java/org/projectnessie/client/http/impl/apache/ApacheRequest.java
```diff
@@ -58,8 +58,6 @@ public HttpResponse executeRequest(Method method, Object body) throws HttpClient
 
     boolean doesOutput = prepareRequest(context);
 
-    config.getRequestFilters().forEach(a -> a.filter(context));
```
adutra, line 61: Request filters were being executed twice for `ApacheRequest`.

snazy (approved):

PR #9055, opened 2024-07-10
HTTP client: refactor BaseHttpRequest

adutra:
This change refactors `BaseHttpRequest` with the following improvements:

* Removed duplicated code, e.g. error mapping code
* Simplified subclasses
* Consistent invocation of response callbacks (these were not always invoked, and / or invoked at different steps)

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/RequestContext.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/RequestContext.java
@@ -40,10 +40,14 @@ public interface RequestContext {
 
   Optional<Object> getBody();
 
+  default boolean doesOutput() {
+    return (getMethod() == Method.PUT || getMethod() == Method.POST) && getBody().isPresent();
+  }
+
   /**
    * Adds a callback to be called when the request has finished. The {@code responseCallback} {@link
    * BiConsumer consumer} is called with a non-{@code null} {@link ResponseContext}, if the HTTP
-   * request technically succeeded. The The {@code responseCallback} {@link BiConsumer consumer} is
+   * request technically succeeded. The {@code responseCallback} {@link BiConsumer consumer} is
    * called with a non-{@code null} {@link Exception} object, if the HTTP request technically
    * failed.
    *
--- a/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
@@ -22,22 +22,8 @@ import java.net.URI;
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
-  /**
-   * Get the status of the response.
-   *
-   * @deprecated use {@link #getStatus()} instead
-   */
-  @Deprecated
-  Status getResponseCode() throws IOException;
-
   /** Get the status of the response. */
-  default Status getStatus() {
-    try {
-      return getResponseCode();
-    } catch (IOException e) {
-      throw new HttpClientException(e);
-    }
-  }
+  Status getStatus();
 
   InputStream getInputStream() throws IOException;
 
--- a/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
@@ -35,7 +35,7 @@ public interface ResponseContext {
     try {
       return getResponseCode();
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      throw new HttpClientException(e);
     }
   }
 
--- a/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
@@ -22,8 +22,22 @@ import java.net.URI;
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
+  /**
+   * Get the status of the response.
+   *
+   * @deprecated use {@link #getStatus()} instead
+   */
+  @Deprecated
+  Status getResponseCode() throws IOException;
+
   /** Get the status of the response. */
-  Status getStatus();
+  default Status getStatus() {
+    try {
+      return getResponseCode();
+    } catch (IOException e) {
+      throw new HttpClientException(e);
+    }
+  }
 
   InputStream getInputStream() throws IOException;
 
--- a/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
@@ -22,7 +22,22 @@ import java.net.URI;
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
+  /**
+   * Get the status of the response.
+   *
+   * @deprecated use {@link #getStatus()} instead
+   */
+  @Deprecated
   Status getResponseCode() throws IOException;
+
+  /** Get the status of the response. */
+  default Status getStatus() {
+    try {
+      return getResponseCode();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
 
   InputStream getInputStream() throws IOException;
 
@@ -43,4 +58,14 @@ public interface ResponseContext {
   String getContentType();
 
   URI getRequestedUri();
+
+  /**
+   * Close the response context.
+   *
+   * @param error if the request failed, this is the exception that caused the failure; will be null
+   *     otherwise
+   */
+  default void close(Exception error) {
+    // no-op
+  }
 }
--- a/api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
@@ -15,8 +15,8 @@
  */
 package org.projectnessie.client.http.impl;
 
-import static org.projectnessie.client.http.impl.HttpUtils.ACCEPT_ENCODING;
 import static org.projectnessie.client.http.impl.HttpUtils.GZIP;
+import static org.projectnessie.client.http.impl.HttpUtils.GZIP_DEFLATE;
 import static org.projectnessie.client.http.impl.HttpUtils.HEADER_ACCEPT;
 import static org.projectnessie.client.http.impl.HttpUtils.HEADER_ACCEPT_ENCODING;
 import static org.projectnessie.client.http.impl.HttpUtils.HEADER_CONTENT_ENCODING;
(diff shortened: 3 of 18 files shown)


api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
```diff
@@ -22,8 +22,23 @@
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
+  /**
+   * Get the status of the response.
+   *
+   * @deprecated use {@link #getStatus()} instead
+   */
+  @Deprecated
   Status getResponseCode() throws IOException;
```
dimas-b, line 31: Could we just remove the exception declaration? If I'm not mistaken, older clients will be fine without recompilation (will have to be update only when they recompile).

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaResponseContext.java
```diff
@@ -75,6 +84,21 @@ public URI getRequestedUri() {
...
+    if (error != null) {
+      try {
+        LOGGER.debug(
+            "Closing unprocessed input stream for {} request to {} delegating to {} ...",
+            response.request().method(),
+            response.uri(),
+            response.body());
+        response.body().close();
+      } catch (IOException e) {
+        // ignore
```
dimas-b, line 97: why not log it?

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaResponseContext.java
```diff
@@ -75,6 +84,21 @@ public URI getRequestedUri() {
     return response.uri();
   }
 
+  public void close(Exception error) {
+    if (error != null) {
+      try {
+        LOGGER.debug(
+            "Closing unprocessed input stream for {} request to {} delegating to {} ...",
```
dimas-b, line 91: How do we know the input stream is "unprocessed"?

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
```diff
@@ -27,65 +27,170 @@
...
+      RequestContext requestContext,
+      @Nullable ResponseContext responseContext,
+      @Nullable RuntimeException error) {
+    List<BiConsumer<ResponseContext, Exception>> callbacks = requestContext.getResponseCallbacks();
+    if (callbacks != null) {
+      for (BiConsumer<ResponseContext, Exception> callback : callbacks) {
+        try {
+          callback.accept(responseContext, error);
+        } catch (Exception e) {
+          LOGGER.warn("Callback invocation failed", e);
```
dimas-b, line 180: Would it make sense to collect all callback exceptions and re-throw as "suppressed" under a new `RuntimeException`? As far as I can see, callback exceptions used to bubble up to the client caller before this PR.

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
```diff
@@ -27,65 +27,170 @@
...
     }
+  }
 
-    return doesOutput;
+  protected ResponseContext processResponse(
+      URI uri, Method method, Object body, RequestContext requestContext) {
+    try {
+      ResponseContext responseContext = sendAndReceive(uri, method, body, requestContext);
+      processResponseFilters(responseContext);
+      mimicUrlConnectionBehavior(method, responseContext, uri);
```
dimas-b, line 117: An exception here will miss closing the `responseContext` it seems.

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
adutra: In theory, the HTTP client can be built by a pluggable `HttpClientFactory` impl loaded via service-loader. If a user has a custom implementation, we'd want it to keep working with the updated `ResponseContext` interface.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaResponseContext.java
adutra: I did not change this part of the code, only moved it.

api/client/src/main/java/org/projectnessie/client/http/impl/jdk11/JavaResponseContext.java
adutra: If an error has been thrown, it is absolutely certain that the response body won't be read because the caller has no access to it.

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
adutra: Good catch! These 2 statements need to be moved into `executeRequest`.

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
dimas-b: Yes, my point is that removing a checked exception will not break binary compatibility, IIRC.

adutra: I think it does, see [this document](https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs-2.md#Evolving-API-interfaces---API-methods) for details.

That being said, I'm happy to remove it if everybody is on board with that. I don't think anyone is actually using custom client factories in the wild.

adutra: Okay, since it's not a binary compatibility issue, I went ahead and removed the method.

adutra: OTOH I think the new name is better than the old one... which is a binary compatibility issue for sure. I might revert my last commit actually.

snazy: See JLS:
```
13.4.21. Method and Constructor Throws

Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time. 
```

snazy: Also indirectly via [method descriptors](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.3)

adutra: So what should I do?

1. Keep both methods
2. Keep only `getResponseCode`, remove the `throws` clause
3. Keep only `getStatus`

dimas-b: Do we actually expect custom `ResponseContext` implementations to exist?... My guess is they do not exist :)

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
```diff
@@ -22,7 +22,8 @@
 /** Interface for the important parts of a response. This is created after executing the request. */
 public interface ResponseContext {
 
-  Status getResponseCode() throws IOException;
+  /** Get the status of the response. */
+  Status getStatus();
```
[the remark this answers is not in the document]
snazy: Actually, this PR #9062 should do the trick.
I came up with [another one](https://github.com/snazy/nessie/tree/http-status-extensible-alt1), but the one in the PR feels better (no volatile reads).

api/client/src/main/java/org/projectnessie/client/http/RequestContext.java
```diff
@@ -40,10 +40,14 @@ public interface RequestContext {
...
   Optional<Object> getBody();
 
+  default boolean doesOutput() {
+    return (getMethod() == Method.PUT || getMethod() == Method.POST) && getBody().isPresent();
+  }
+
   /**
    * Adds a callback to be called when the request has finished. The {@code responseCallback} {@link
    * BiConsumer consumer} is called with a non-{@code null} {@link ResponseContext}, if the HTTP
-   * request technically succeeded. The The {@code responseCallback} {@link BiConsumer consumer} is
```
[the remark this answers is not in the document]
adutra: Eagle eye :-D – more seriously though, it's IntelliJ spell checks.

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
```diff
@@ -49,43 +60,138 @@ protected BaseHttpRequest(HttpRuntimeConfig config, URI baseUri) {
...
+      config.getResponseFilters().forEach(responseFilter -> responseFilter.filter(responseContext));
+    }
   }
 
-  protected void cleanUp() {
+  /**
+   * This mimics the (weird) behavior of java.net.HttpURLConnection.getResponseCode() that throws an
+   * IOException for some status codes.
+   */
+  protected void mimicUrlConnectionBehavior(
```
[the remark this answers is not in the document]
adutra: > "bother" the jdk11 + apache impls but only the url-connection impl

You mean, catch the error thrown and transform it into a "regular" `ResponseContext`? Mind if we do that in another PR? My fear is that some parts may be relying on this behavior (the OAuth2 client especially).

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
```diff
@@ -49,43 +60,138 @@ protected BaseHttpRequest(HttpRuntimeConfig config, URI baseUri) {
...
+      @Nullable ResponseContext responseContext,
+      @Nullable RuntimeException originalError) {
+    List<BiConsumer<ResponseContext, Exception>> callbacks = requestContext.getResponseCallbacks();
+    RuntimeException error = originalError;
+    if (callbacks != null) {
+      for (BiConsumer<ResponseContext, Exception> callback : callbacks) {
+        try {
+          callback.accept(responseContext, originalError);
+        } catch (RuntimeException e) {
+          if (error == null) {
```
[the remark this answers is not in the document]
adutra: I thought about that and decided not to do it: `NessieHttpResponseFilter`, for instance, throws exceptions and these are meant to be thrown right away.

api/client/src/main/java/org/projectnessie/client/http/impl/RequestContextImpl.java
```diff
@@ -90,16 +90,6 @@ && getContentType()
             .isPresent();
   }
 
-  /**
-   * Adds a callback to be called when the request has finished. The {@code responseCallback} {@link
-   * BiConsumer consumer} is called with a non-{@code null} {@link ResponseContext}, if the HTTP
-   * request technically succeeded. The {@code responseCallback} {@link BiConsumer consumer} is
```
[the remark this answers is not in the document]
adutra: The comment appeared almost identical in two places: interface and impl, with one typo in the interface. I kept only the interface comment and fixed the typo.

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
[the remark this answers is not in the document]
snazy: sure

api/client/src/main/java/org/projectnessie/client/http/impl/BaseHttpRequest.java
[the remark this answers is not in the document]
snazy: okay

snazy (approved):
Looks good from my POC, deferring to @dimas-b

dimas-b (approved):
LGTM :)

PR #9062, opened 2024-07-10
Make `o.p.client.http.Status` safe against unknown HTTP status codes

snazy:

PR #9071, opened 2024-07-11
HTTP client: remove hard-coded behavior for status >= 400

adutra:
As requested here:

https://github.com/projectnessie/nessie/pull/9055#discussion_r1672844009

This PR removes all the "oddities" due to the legacy `HttpURLConnection` request impl:

* No more hard-coded error when status >= 400
* `RequestContext` returns only a single `InputStream`

As a consequence, `HttpResponse` now handles such statuses gracefully and returns `null` if the response has no body. It was slightly refactored  to use `CapturingInputStream` for improved error messages.

#### Remarks on the "UrlConnection behavior" and impacts of removing it to end users

To my surprise, the "UrlConnection behavior" consisting of throwing a hard-coded exception for statuses >= 400, and which was being mimicked for Java and Apache request impls... does not manifest itself when using `UrlConnectionRequest` :-)

The odd-behaving method seemed to be `HttpURLConnection.getResponseCode()`, judging from comments in the code. Looking at the implementations, it calls `getInputStream()`, which indeed throws a hard-coded `IOException` when the status is >= 400. But what's really weird is that `getResponseCode()` catches the exception, and ignores it completely! I verified this for JDKs 17, 11 and 8.

The only way to surface that `IOException` is to call `HttpURLConnection.getInputStream()` directly. I suppose that in earlier versions of the HTTP client, this might have happened, but it doesn't anymore: we don't call this method anymore if the status is >= 400.

Apart from that, this PR introduces an obvious behavioral change since we are not throwing any error for status >= 400 anymore. Instead, an `HttpResponse` is returned, which may or may not contain an entity.

I had to adapt the code that reads the entity a bit because of that.

In overall I think this behavioral change is fine, because the HTTP client is never used in the wild without the `NessieHttpResponseFilter` which already intercepts statuses >= 400. So from a user perspective, there is no change because the removed behavior wasn't exercised.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
+++ b/api/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
@@ -15,13 +15,16 @@
  */
 package org.projectnessie.client.http;
 
+import com.fasterxml.jackson.databind.MappingIterator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import org.projectnessie.client.rest.NessieBadResponseException;
+import org.projectnessie.client.rest.io.CapturingInputStream;
 import org.projectnessie.error.ImmutableNessieError;
+import org.projectnessie.error.NessieError;
 
 /** Simple holder for http response object. */
 public class HttpResponse {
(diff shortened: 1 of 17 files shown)


api/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
```diff
@@ -34,49 +37,76 @@ public HttpResponse(ResponseContext context, ObjectMapper mapper) {
...
+            responseContext.getContentType(),
+            responseContext.getRequestedUri(),
+            captured.isEmpty()
+                ? "and no response body"
+                : "and a response body beginning with: \"" + captured + "\"");
+    NessieError error =
         ImmutableNessieError.builder()
+            .message(message)
             .status(status.getCode())
-            .message(status.getReason())
```
adutra, line 66: I fixed the inconsistency here where the reason was being input as message, and the message as reason.

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
```diff
@@ -25,10 +26,15 @@ public interface ResponseContext {
   /** Get the status of the response. */
   Status getStatus();
 
+  /**
+   * Get the input stream for the response. Returns null if the response did not contain any
+   * readable entity.
+   *
+   * @implSpec Calling this method multiple times should return the same input stream.
```
adutra, line 33: This is especially true for `GZIPInputStream`: it's illegal to create 2 streams from the same non-compressed stream.

api/client/src/main/java/org/projectnessie/client/http/impl/apache/ApacheResponseContext.java
```diff
@@ -43,25 +44,13 @@ public Status getStatus() {
...
-    return reader();
-  }
-
-  @Override
-  public InputStream getErrorStream() throws IOException {
-    return reader();
-  }
-
-  @Override
-  public boolean isJsonCompatibleResponse() {
```
adutra, line 55: Was duplicating the default method in `ResponseContext`.

api/client/src/main/java/org/projectnessie/client/http/HttpResponse.java
snazy: Eagle eye ;)

api/client/src/main/java/org/projectnessie/client/http/ResponseContext.java
```diff
@@ -25,10 +26,15 @@ public interface ResponseContext {
...
+   */
+  @Nullable
   InputStream getInputStream() throws IOException;
 
-  InputStream getErrorStream() throws IOException;
```
snazy, line 30: Oh yea - that's another nice side effect / cleanup!

api/client/src/test/java/org/projectnessie/client/http/BaseTestHttpClient.java
```diff
@@ -479,31 +480,107 @@ static Stream<Status> testHttpResponses() {
   }
 
   @ParameterizedTest
-  @MethodSource
+  @MethodSource("httpResponses")
   void testHttpResponses(Status status) throws Exception {
```
snazy, line 484: why not just this?
```suggestion
  void httpResponses(Status status) throws Exception {
```

adutra: Because the source is now being used for many test methods.

api/client/src/test/java/org/projectnessie/client/http/BaseTestHttpClient.java
```diff
@@ -479,31 +480,107 @@ static Stream<Status> testHttpResponses() {
...
+        (req, resp) -> {
+          resp.setStatus(status.getCode());
+          if (status != Status.NO_CONTENT && status != Status.NOT_MODIFIED) {
+            resp.setContentType("application/json");
+            try (OutputStream os = resp.getOutputStream()) {
+              MAPPER.writeValue(os, expected);
+            }
+          }
+        };
+    try (HttpTestServer server = new HttpTestServer("/a/b", handler)) {
```
snazy, line 496: Hm - the overall test duration increased quite a bit (I've actually made it worse than before with #9062).

I think, we can create the test server once per test class and reuse it by having a `static AtomicReference<HttpTestServer.RequestHandler>` and then re-delegate:
```java
    static final AtomicReference<HttpTestServer.RequestHandler> requestHandler = new AtomicReference<>();
    static HttpTestServer server;
    @BeforeAll
    static void setHttpServer() {
      server = new HttpTestServer("/a/b", (req, resp) -> {
        handler.get().handle(req, resp);
      });
    }
    @AfterAll
    static void shutdown() {
      server.stop();
    }
```

That should make the tests run faster.  (We run those for all clients and a bunch of Jackson versions.)

adutra: I improved things a bit but there is something weird with the Apache client: it's much slower than the other two. Not sure if this is a regression introduced in this PR though.

adutra: On main:
Screenshot 2024-07-12 at 12 27 02

On this branch, removing the new tests for an apples to apples comparison, but including the changes to share the same client and server between tests:

Screenshot 2024-07-12 at 12 34 04

It's roughly the same times, but indeed the Apache client is performing worse than the others. 

It seems to "hang" when processing some HTTP codes like 429 or 503. Looking a bit in the javadocs, I found the [`DefaultBackoffStrategy`](https://hc.apache.org/httpcomponents-client-5.1.x/current/httpclient5/apidocs/org/apache/hc/client5/http/impl/classic/DefaultBackoffStrategy.html) class:

> This [ConnectionBackoffStrategy](https://hc.apache.org/httpcomponents-client-5.1.x/current/httpclient5/apidocs/org/apache/hc/client5/http/classic/ConnectionBackoffStrategy.html) backs off either for a raw network socket or connection timeout or if the server explicitly sends a 429 (Too Many Requests) or a 503 (Service Unavailable) response.

So, this might explain the differences.

snazy: Oh! Backoff :+1:

snazy (approved):

PR #9074, opened 2024-07-11
GC: Support Iceberg Views

### LINKED ISSUE #9067: GC: Add support for Iceberg Views
* Identify live view versions (metadata locations).
* Expire metadata locations

snazy:
The code is similar to tables, just only expiring metadata objects.

Fixes #9067

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/IcebergContent.java
+++ b/api/model/src/main/java/org/projectnessie/model/IcebergContent.java
@@ -15,4 +15,8 @@
  */
 package org.projectnessie.model;
 
-public abstract class IcebergContent extends Content {}
+public abstract class IcebergContent extends Content {
+  public abstract String getMetadataLocation();
+
+  public abstract long getVersionId();
+}
--- a/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
+++ b/gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
@@ -16,6 +16,7 @@
 package org.projectnessie.gc.contents.jdbc;
 
 import static com.google.common.base.Throwables.getStackTraceAsString;
+import static org.projectnessie.gc.contents.ContentReference.icebergContent;
 import static org.projectnessie.gc.contents.jdbc.JdbcHelper.isIntegrityConstraintViolation;
 import static org.projectnessie.gc.contents.jdbc.SqlDmlDdl.ADD_CONTENT;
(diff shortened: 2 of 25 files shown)


api/model/src/main/java/org/projectnessie/model/IcebergContent.java
```diff
@@ -15,4 +15,8 @@
  */
 package org.projectnessie.model;
 
-public abstract class IcebergContent extends Content {}
+public abstract class IcebergContent extends Content {
+  public abstract String getMetadataLocation();
+
+  public abstract long getVersionId();
```
adutra, line 21: Isn't this method addition breaking binary backwards compatibility?

adutra: Also: don't you need to modify `nessie-model-quarkus` accordingly?

snazy: How would an addition break backwards compatibility in Java?

BTW: `nessie-model-quarkus` uses the same sources

gc/gc-repository-jdbc/src/main/java/org/projectnessie/gc/contents/jdbc/JdbcPersistenceSpi.java
```diff
@@ -230,17 +233,17 @@ public long addIdentifiedLiveContent(UUID liveSetId, Stream<ContentReference> co
...
             stmt.setString(4, ref.contentKey().toPathString());
             stmt.setString(5, ref.contentType().name());
-            if (ref.contentType().equals(Content.Type.ICEBERG_TABLE)) {
+            if (ref.contentType().equals(ICEBERG_TABLE) || ref.contentType().equals(ICEBERG_VIEW)) {
               stmt.setString(
                   6,
                   Objects.requireNonNull(
                       ref.metadataLocation(),
-                      "Illegal null metadataLocation in ContentReference for ICEBERG_TABLE"));
+                      "Illegal null metadataLocation in ContentReference for ICEBERG_TABLE/ICEBERG_VIEW"));
```
adutra, line 241:
```suggestion
                  Objects.requireNonNull(
                      ref.metadataLocation(),
                      String.format(
                          "Illegal null metadataLocation in ContentReference for %s",
                          ref.contentType())));
```

api/model/src/main/java/org/projectnessie/model/IcebergContent.java
adutra: Because it's an abstract method, thus requiring code extending this class to implement it.

snazy: Nothing except the types in `:nessie-model` extend `IcebergContent`.

adutra (approved):

PR #9085, opened 2024-07-13
Catalog: Fix assignment of nested fields in a table's initial schema

snazy:
The field IDs of nested structures/lists/maps are not properly (re)assigned during the creation of a table. This can lead to duplicate field IDs.

This change fixes the oversight and adds a test for this case.

Another bug fixed by this change it the calculation of the `last-column-id`, which needs to even respect IDs in collection types (maps + lists).

### DIFF
--- a/catalog/format/iceberg/src/test/java/org/projectnessie/catalog/formats/iceberg/nessie/TestNessieModelIceberg.java
+++ b/catalog/format/iceberg/src/test/java/org/projectnessie/catalog/formats/iceberg/nessie/TestNessieModelIceberg.java
@@ -16,20 +16,36 @@
 package org.projectnessie.catalog.formats.iceberg.nessie;
 
 import static java.util.Collections.singletonList;
+import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 import static org.projectnessie.catalog.formats.iceberg.fixtures.IcebergFixtures.icebergSchemaAllTypes;
 import static org.projectnessie.catalog.formats.iceberg.fixtures.IcebergFixtures.tableMetadataBare;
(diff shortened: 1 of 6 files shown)


catalog/format/iceberg/src/test/java/org/projectnessie/catalog/formats/iceberg/nessie/TestNessieModelIceberg.java
```diff
@@ -316,6 +326,65 @@ public void icebergNessieTypeConversion(IcebergType icebergType) throws Exceptio
         .isEqualTo(nessieIdHasher("NessieTypeSpec").hash(nessieAgain).generate());
   }
 
+  @Test
+  public void icebergNested() throws Exception {
```
adutra, line 330: Maybe add a field of type Map? It doesn't look like we are testing this specific type.

sdasdasd1:
@snazy 
@adutra 
I think we also encountered this problem when creating a table through the nessie REST interface

```
Caused by: org.apache.iceberg.shaded.com.fasterxml.jackson.databind.JsonMappingException: Multiple entries with same key: 288=day and 288=__metadata.key (through reference chain: org.apache.iceberg.rest.responses.LoadTableResponse["metadata"])

```
__metadata is our nested field

sdasdasd1:
so we are really waiting for the fix :)

snazy:
> @snazy @adutra I think we also encountered this problem when creating a table through the nessie REST interface
> 
> ```
> Caused by: org.apache.iceberg.shaded.com.fasterxml.jackson.databind.JsonMappingException: Multiple entries with same key: 288=day and 288=__metadata.key (through reference chain: org.apache.iceberg.rest.responses.LoadTableResponse["metadata"])
> ```
> 
> __metadata is our nested field

That _could_ be fixed by the last-column-id fix I added to this PR as well, because it is very related.

adutra (approved):

snazy:
This is a "serious enough" bug that justifies a patch release (coming soon).

PR #9086, opened 2024-07-13
feat: support using default azure credential builder for ADLS Gen2

cgpoh:
Use DefaultAzureCredentialBuilder as last option for ADLS Gen2. Use DataLakeFileSystemClient getProperties instead of getAccessPolicy for health check as some companies may disallow getAccessPolicy permission.


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsClientSupplier.java
```diff
@@ -89,10 +90,7 @@ DataLakeFileSystemClient fileSystemClient(AdlsLocation location) {
       String accountKey = fileSystemOptions.account().get().secret();
       clientBuilder.credential(new StorageSharedKeyCredential(accountName, accountKey));
     } else {
-      throw new IllegalStateException(
-          "Neither an SAS token nor account name and key are available for ADLS file system '"
-              + location.path()
-              + "'");
+      clientBuilder.credential(new DefaultAzureCredentialBuilder().build());
```
adutra, line 93: @cgpoh thank you for your PR!

Very recently, we adopted a new configuration strategy for S3 and GCS: instead of trying options one by one like here, we introduced a config option like this one for GCS:

https://github.com/projectnessie/nessie/blob/e58d579ef68ea907f3d3fa2e00ff7472ca3150c0/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsBucketOptions.java#L54

This option is then processed like this:

https://github.com/projectnessie/nessie/blob/b6187e2cbe7a92bdfc85a8ddfb65a26b30ab3a00/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java#L123-L171

I think it could be cool to do the same for ADLS, WDYT?

cgpoh: @adutra , thanks for the review. I see whether I can do the same for ADLS.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -305,17 +307,28 @@ private Map<String, String> adlsConfigOverrides(StorageUri warehouseLocation) {
...
-        .ifPresent(
-            key -> {
-              configOverrides.put(ADLS_SHARED_KEY_ACCOUNT_NAME, key.name());
-              configOverrides.put(ADLS_SHARED_KEY_ACCOUNT_KEY, key.secret());
-            });
-    fileSystemOptions
-        .sasToken()
-        .ifPresent(
-            s -> configOverrides.put(ADLS_SAS_TOKEN_PREFIX + location.storageAccount(), s.key()));
+    AdlsFileSystemOptions.AzureAuthType authType =
```
snazy, line 310: Heads up: this will conflict w/ #9100

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
```diff
@@ -50,7 +50,7 @@ public void ping(StorageUri uri) throws IOException {
 
     DataLakeFileSystemClient fileSystem = clientSupplier.fileSystemClient(location);
     try {
-      fileSystem.getAccessPolicy();
+      fileSystem.getProperties();
```
snazy, line 53: Why this change? This `ping` function is intended to hit ADLS.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsFileSystemOptions.java
```diff
@@ -64,4 +67,10 @@ enum AdlsRetryStrategy {
     EXPONENTIAL_BACKOFF,
     FIXED_DELAY,
   }
+
+  enum AzureAuthType {
+    STORAGE_SHARED_KEY,
```
snazy, line 72: Can you add a `NONE` option as well and let `AdlsClientSupplier.fileSystemClient()` default to `NONE`?

Although it's probably rare that file-systems are publicly accessible, and even unlikely that those are writeable, but we can't predict which auth method users want.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
cgpoh: I will have permission error with `getAccessPolicy` as my company doesn't allow getting access policy.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsFileSystemOptions.java
cgpoh: Should I throw `IllegalArgumentException("Unsupported auth type " + authType);` if it is `NONE` as `clientBuilder.credential` will throw `NPE` if credential is null

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
snazy: Fair enough.

snazy:
Looks like CI's not happy with the code formatting and tests are failing.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
dimas-b: Maybe move this change to a separate PR for clarity as it relates not to credentials but to server-side policies?

dimas-b: #9134

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
cgpoh: Thanks! I removed the credentials sending part.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsClientSupplier.java
```diff
@@ -92,16 +93,31 @@ DataLakeFileSystemClient fileSystemClient(AdlsLocation location) {
...
-      clientBuilder.credential(new StorageSharedKeyCredential(accountName, accountKey));
-    } else {
-      throw new IllegalStateException(
-          "Neither an SAS token nor account name and key are available for ADLS file system '"
-              + location.path()
-              + "'");
+    AdlsFileSystemOptions.AzureAuthType authType =
+        fileSystemOptions
+            .authType()
+            .orElse(AdlsFileSystemOptions.AzureAuthType.APPLICATION_DEFAULT);
```
snazy, line 99: Guess this should default to `NONE` and let the user explicitly define the type.

cgpoh: @snazy, just tested the behavior when I don't call `clientBuilder.credential`, I will get the following error: 
```
"com.azure.storage.file.datalake.models.DataLakeStorageException: Status code 401, \"﻿<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>NoAuthenticationInformation</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nRequestId:22377f67-e01e-0060-54g2-d9e145123456\nTime:2024-07-19T15:35:55.3953452Z</Message></Error>\""
```
which to me is correct behavior since no authentication information is provided. I will proceed to add in the `NONE` enum and default to it

snazy: I feared that it behaves like that, but it looks like you can use `com.microsoft.azure.storage.StorageCredentialsAnonymous#ANONYMOUS` in the `NONE` case - it's probably not that useful in practice, but who knows.

cgpoh: I think I can set Anonymous access in the `NONE` case by `clientBuilder.setAnonymousAccess();`. I will push the changes, let me know if it is ok.

snazy:
LGTM
@adutra ?

adutra (approved):
LGTM, I will adapt the Helm chart in a follow-up PR.

snazy (approved):
Thanks @cgpoh for the contribution!

cgpoh:
Thanks @snazy @adutra for your help to review!

PR #9099, opened 2024-07-15
Add `message` argument to `Objects.requireNonNull()`

snazy:


api/client/src/main/java/org/projectnessie/client/auth/oauth2/RefreshTokensFlow.java
```diff
@@ -30,7 +30,7 @@ class RefreshTokensFlow extends AbstractFlow {
 
   @Override
   public Tokens fetchNewTokens(@Nullable Tokens currentTokens) {
-    Objects.requireNonNull(currentTokens);
+    Objects.requireNonNull(currentTokens, "currentTokens is null");
     Objects.requireNonNull(currentTokens.getRefreshToken());
```
adutra, line 34: Why not this one?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
```diff
@@ -32,7 +32,7 @@ public final class S3Utils {
   private S3Utils() {}
 
   public static Optional<String> extractBucketName(URI uri) {
-    String scheme = requireNonNull(uri).getScheme();
+    String scheme = requireNonNull(uri, "URI argument missing").getScheme();
```
adutra, line 35: Couldn't we settle on a standard message format? E.g.. `argument %s is null`

api/client/src/main/java/org/projectnessie/client/auth/oauth2/RefreshTokensFlow.java
snazy: Which one?

snazy: Not sure BTW why the arg is `@Nullable` but the impl requires non-null

adutra: > Which one?

```java
Objects.requireNonNull(currentTokens.getRefreshToken(), "currentTokens.getRefreshToken() is null");
```

> Not sure BTW why the arg is @Nullable but the impl requires non-null

The interface is used both for initial token fetches and token refreshes; `currentTokens` is null for initial token fetches and non-null for token refreshes.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
snazy: At least we have something that helps more than just an NPE w/o a reason

adutra (approved):

PR #9100, opened 2024-07-15
Catalog: Don't expose ADLS + GCS credentials

snazy:
ADLS + GCS credentials are no longer sent to the client. It is considered insecure to expose the server's credentials to clients, even if this is likely very convenient. Unless we have a secure mechanism to provide per-client/table credentials, users have to configure object store credentials when using GCS or ADLS via the local Iceberg configuration(s).

### DIFF
--- a/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
+++ b/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
@@ -272,25 +272,6 @@ public class IcebergConfigurer {
     gcsBucketOptions
         .deleteBatchSize()
         .ifPresent(dbs -> configOverrides.put(GCS_DELETE_BATCH_SIZE, Integer.toString(dbs)));
-    // FIXME it is not safe to send de/encryption keys and oauth2 tokens
-    gcsBucketOptions
-        .decryptionKey()
-        .ifPresent(key -> configOverrides.put(GCS_DECRYPTION_KEY, key.key()));
-    gcsBucketOptions
-        .encryptionKey()
-        .ifPresent(key -> configOverrides.put(GCS_ENCRYPTION_KEY, key.key()));
-    gcsBucketOptions
-        .oauth2Token()
-        .ifPresent(
-            token -> {
-              configOverrides.put(GCS_OAUTH2_TOKEN, token.token());
-              token
-                  .expiresAt()
-                  .ifPresent(
-                      e ->
-                          configOverrides.put(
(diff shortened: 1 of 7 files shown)


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -273,24 +273,24 @@ private Map<String, String> gcsConfigOverrides(StorageUri warehouseLocation) {
...
-            token -> {
-              configOverrides.put(GCS_OAUTH2_TOKEN, token.token());
-              token
-                  .expiresAt()
-                  .ifPresent(
-                      e ->
-                          configOverrides.put(
-                              GCS_OAUTH2_TOKEN_EXPIRES_AT, String.valueOf(e.toEpochMilli())));
-            });
+    //    gcsBucketOptions
```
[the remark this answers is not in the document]
snazy: It's big yellow that way :)

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/AbstractIcebergCatalogTests.java
```diff
@@ -74,7 +74,11 @@ protected RESTCatalog catalog() {
   }
 
   protected Map<String, String> catalogOptions() {
-    return singletonMap(CatalogProperties.WAREHOUSE_LOCATION, WAREHOUSE_NAME);
+    return ImmutableMap.of(
+        CatalogProperties.WAREHOUSE_LOCATION,
+        WAREHOUSE_NAME,
+        "adls.sas-token.account.dfs.core.windows.net",
```
adutra, line 80: Do we need this ADLS-specific option for all tests that inherit from this class?

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/WarehouseAccount.java
```diff
@@ -0,0 +1,27 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface WarehouseAccount {}
```
adutra, line 27: This seems specific to Azure, how about `WarehouseAdlsAccount`?

adutra (approved):

PR #9107, opened 2024-07-16
Catalog / GCS: minor enhancements

adutra:


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsBucketOptions.java
```diff
@@ -23,6 +23,9 @@
 
 public interface GcsBucketOptions {
 
+  /** Default value for {@link #authType()}, being {@link GcsAuthType#NONE}. */
+  GcsAuthType DEFAULT_AUTH_TYPE = GcsAuthType.NONE;
```
snazy, line 27: Can you mention `NONE` as the default value for `authType` so it gets into the docs on the web site?

snazy: Guess that's missing for S3 + ADLS as well.

snazy: I meant to add those defaults to `CatalogS3BucketConfig` et all using the `@WithDefault` annotation

adutra: Add `@WithDefault` to an optional property? Isn't that contradictory? I think we should remove the `Optional` in that case. 

Also, I vaguely recall that we did it on purpose to avoid `@WithDefault` in those bucket config classes because they trigger the instantiation of the whole class. But I think this isn't a concern anymore, because there is no more methods like `Optional<BucketConfig> defaultBucket()`.

snazy: No, leave `Optional` - it's needed for the builder approach (bucket -> default).
Nothing to do with default values.

snazy (approved):

PR #9108, opened 2024-07-16
Catalog / Helm: fix wrong paths and invalid env vars

adutra:
Summary of changes:

* Fix config options of type KeySecret
* Remove all quotes from config options, both as env vars and in configmap
* Quote env var names

Note: TLDR is, we don't have the choice but to give up on double-quoting bucket names. This is sad, because bucket names can contain dots; but unfortunately, if that's the case, it will be impossible to provide configuration specific to such buckets.


dimas-b (approved):

adutra:
Closing as I'm preparing another PR with a better handling of invalid bucket names.

PR #9120, opened 2024-07-17
CLI: Add `REVERT CONTENT` command

snazy:
Based on a user's idea, this new `REVERT CONTENT` command allows updating one or more individial content keys to a state as in another commit.

The example at hand is that the user's query engine performed a commit to Nessie, but removed the metadata object after that, leaving a Nessie commit w/ an invalid metadata location.

### DIFF
--- a/cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/RevertContentCommand.java
+++ b/cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/RevertContentCommand.java
@@ -1,0 +1,196 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.nessie.cli.commands;
+
+import static java.util.Objects.requireNonNull;
+import static org.projectnessie.nessie.cli.cli.BaseNessieCli.STYLE_BOLD;
+import static org.projectnessie.nessie.cli.cli.BaseNessieCli.STYLE_ERROR;
+import static org.projectnessie.nessie.cli.cli.BaseNessieCli.STYLE_FAIL;
+import static org.projectnessie.nessie.cli.cli.BaseNessieCli.STYLE_FAINT;
+import static org.projectnessie.nessie.cli.cli.BaseNessieCli.STYLE_INFO;
+
+import jakarta.annotation.Nonnull;
+import java.io.PrintWriter;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.jline.terminal.Terminal;
+import org.jline.utils.AttributedString;
+import org.jline.utils.AttributedStringBuilder;
+import org.projectnessie.client.api.CommitMultipleOperationsBuilder;
+import org.projectnessie.model.Branch;
+import org.projectnessie.model.CommitMeta;
+import org.projectnessie.model.CommitResponse;
+import org.projectnessie.model.Content;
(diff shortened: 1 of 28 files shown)


adutra:
Two general questions as I'm trying to digest the use case:

1. Wouldn't a `REVERT COMMIT` command be easier for users to grasp? Just like git revert, simply pass the commit sha (or named ref) to revert?
2. What if the user makes a mistake when providing the content keys? From what I see, providing a key that cannot be found in the source commit would result in a `DELETE` operation for that key being committed. Isn't that a bit dangerous?

snazy:
>     1. Wouldn't a `REVERT COMMIT` command be easier for users to grasp? Just like git revert, simply pass the commit sha (or named ref) to revert?

I'd think of `REVERT COMMIT` as a revert of a whole commit, not individual keys.

>     2. What if the user makes a mistake when providing the content keys? From what I see, providing a key that cannot be found in the source commit would result in a `DELETE` operation for that key being committed. Isn't that a bit dangerous?

Yes, it would (try to) delete those keys. You think it would be better to just warn and hint users to the "DROP TABLE/VIEW/NAMESPACE" commands?

cli/cli/src/main/java/org/projectnessie/nessie/cli/commands/RevertContentCommand.java
```diff
@@ -0,0 +1,120 @@
...
+    Map<ContentKey, Content> contentsMap = contents.toContentsMap();
+    Reference effectiveReference =
+        requireNonNull(
+            contents.getEffectiveReference(), "No effective reference, Nessie API v2 required");
+    for (ContentKey contentKey : contentKeys) {
+      Content content = contentsMap.get(contentKey);
+      if (content != null) {
+        commit.operation(Operation.Put.of(contentKey, content));
+      } else {
+        commit.operation(Operation.Delete.of(contentKey));
```
dimas-b, line 75: I tend to think that deletes may be unexpected by the user of this command. Could we add a flag like `REVERT CONTENT OF ... [DELETE MISSING]`?

dimas-b (commented):
Nice idea :)

adutra:
> Yes, it would (try to) delete those keys. You think it would be better to just warn and hint users to the "DROP TABLE/VIEW/NAMESPACE" commands?

Maybe implement @dimas-b suggestion of `DELETE MISSING`?

snazy:
Added `DRY` + `ALLOW DELETES` clauses.

cli/grammar/src/main/resources/org/projectnessie/nessie/cli/syntax/RevertContentStatement.help.txt
```diff
@@ -0,0 +1,14 @@
+Creates a new commit in Nessie to set the state of the given content-keys to the same state as those content keys
+existed on the `TO STATE` reference.
+
+This command is for example useful when another system has created a Nessie commit but the referenced Iceberg
+metadata object was not written.
+
+It is recommended to use the `DRY` keyword to investigate the changes before actually committing those to Nessie.
+
+Contents that exist on the `TO STATE` are updated in the new Nessie commit to that state.
+contents that do not exist on the `TO STATE` are deleted only if the `ALLOW DELETES` clause is specified, otherwise
```
adutra, line 10:
```suggestion
Contents that do not exist on the `TO STATE` are deleted only if the `ALLOW DELETES` clause is specified, otherwise
```

snazy: Eagle eye

adutra (approved):

cli/grammar/src/main/congocc/nessie/nessie-cli.ccc
```diff
@@ -191,6 +194,69 @@ CreateNamespaceStatement
       ]
     ;
 
+RevertContentStatement
+    : <REVERT>
+      <CONTENT>
+      { setOptionalNextTokenTypes(TokenType.DRY); }
+      [ /dryRun/=<DRY>
```
dimas-b, line 201: +1 to the dry run option in general, but doesn't `REVERT CONTENT DRY OF myKey ...` look a bit awkward? Why not `REVERT CONTENT OF ... DRY RUN`?

dimas-b:
`ALLOW DELETES` LGTM :+1:

cli/grammar/src/main/congocc/nessie/nessie-cli.ccc
snazy: b/c `MERGE` has `DRY` at the beginning as well

PR #9203, opened 2024-07-26
Helm chart: configurable liveness and readiness probes

adutra:


dimas-b (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+- Helm chart: liveness and readiness probes are now configurable via the `livenessProbe` and 
+  `readinessProbe` Helm values.
```
snazy, line 18: No `{{` - so :+1:

helm/nessie/values.yaml
```diff
@@ -716,3 +716,31 @@ affinity: {}
 #                operator: In
 #                values:
 #                  - nessie
+
+# -- Configures the liveness probe for nessie pods.
+livenessProbe:
+  # -- Number of seconds after the container has started before liveness probes are initiated. Minimum value is 0.
+  initialDelaySeconds: 5
+  # -- How often (in seconds) to perform the probe. Minimum value is 1.
+  periodSeconds: 30
```
snazy, line 725:
```suggestion
  periodSeconds: 10
```

helm/nessie/values.yaml
```diff
@@ -716,3 +716,31 @@ affinity: {}
...
+  failureThreshold: 3
+  # -- Number of seconds after which the probe times out. Minimum value is 1.
+  timeoutSeconds: 10
+  # -- Optional duration in seconds the pod needs to terminate gracefully upon probe failure. Minimum value is 1.
+  terminationGracePeriodSeconds: 30
+
+# -- Configures the readiness probe for nessie pods.
+readinessProbe:
+  # -- Number of seconds after the container has started before readiness probes are initiated. Minimum value is 0.
+  initialDelaySeconds: 10
```
snazy, line 738:
```suggestion
  initialDelaySeconds: 5
```

helm/nessie/values.yaml
```diff
@@ -716,3 +716,31 @@ affinity: {}
...
+  # -- How often (in seconds) to perform the probe. Minimum value is 1.
+  periodSeconds: 45
```
snazy, line 740:
```suggestion
  periodSeconds: 10
```

snazy (commented):
Just suggestion to lower the initial and delay values

adutra:
> Just suggestion to lower the initial and delay values

There is a complex equation to solve here: if the first probe kicks in earlier, that's generally a good thing... unless it fails, in which case, we'd have to pay the price of waiting `periodSeconds` until the next probe.

So if there is little chance for the probe to succeed at 5 seconds, maybe it's better to let it kick in at, say, 10 seconds.

But with all that said, I'm OK to make 5 seconds the default, especially if the period is 10 seconds instead of 30.

dimas-b (approved):

snazy (approved):

PR #9217, opened 2024-07-29
Catalog/Iceberg: return `prefix` in config overrides

snazy:
Set the "default" prefix in the config _overrides_. Iceberg/Java respects the "prefix" in the config defaults, but Iceberg/Python does not, if the user specified "prefix" in the `load_catalog()` call.


snazy:
See [Pyiceberg](https://github.com/apache/iceberg-python/blob/be5c42649914e71e8366c22558f8234ce062b145/pyiceberg/catalog/rest.py#L341-L344) code.

adutra (approved):

snazy:
Cannot move it to the "overrides", because that breaks the Nessie Spark SQL extensions.

PR #9221, opened 2024-07-29
Gradle wrapper - download and verify

snazy:
This change downloads the "right" Gradle wrapper (matching the Gradle version) and verifies its integrity.


gradlew
```diff
@@ -89,6 +89,8 @@ APP_BASE_NAME=${0##*/}
 APP_HOME=$( cd -P "${APP_HOME:-./}" > /dev/null && printf '%s
 ' "$PWD" ) || exit
 
+. ${APP_HOME}/gradle/gradlew-include.sh
```
adutra, line 92:
```suggestion
. "${APP_HOME}"/gradle/gradlew-include.sh
```

adutra (commented):
Why do we need this? Isn't the default wrapper already taking care fo downloading and verifying ?

snazy:
> Why do we need this? Isn't the default wrapper already taking care fo downloading and verifying ?

No, it doesn't update the wrapper jar

gradle/gradlew-include.sh
```diff
@@ -0,0 +1,53 @@
...
+  fi
+fi
+if [ ! -e "${GRADLE_WRAPPER_SHA256}" ]; then
+  curl --location --output "${GRADLE_WRAPPER_SHA256}" https://services.gradle.org/distributions/gradle-${GRADLE_DIST_VERSION}-wrapper.jar.sha256 || exit 1
+fi
+if [ ! -e "${GRADLE_WRAPPER_JAR}" ]; then
+  # The Gradle version extracted from the `distributionUrl` property does not contain ".0" patch
+  # versions. Need to append a ".0" in that case to download the wrapper jar.
+  GRADLE_VERSION="$(echo "$GRADLE_DIST_VERSION" | sed 's/^\([0-9]*[.][0-9]*\)$/\1.0/')"
+  curl --location --output "${GRADLE_WRAPPER_JAR}" https://raw.githubusercontent.com/gradle/gradle/v${GRADLE_VERSION}/gradle/wrapper/gradle-wrapper.jar || exit 1
```
XN137, line 44: justFYI:

the gradle wrapper jar under that URL is never the official one from the release:

see git history:
https://github.com/gradle/gradle/commits/v8.9.0/gradle/wrapper

so if we are unlucky and the jar changes in a new release, the sha256 check will probably be failing.
as of right now, we are in luck though:
```
==> gradle-8.9-rc-1-wrapper.jar.sha256 <==
498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17
==> gradle-8.9-rc-2-wrapper.jar.sha256 <==
498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17
==> gradle-8.9-wrapper.jar.sha256 <==
498495120a03b9a6ab5d155f5de3c8f0d986a449153702fb80fc80e134484f17%             
```

snazy: That behavior is weird, but correct. The jar in there is the one that makes it into the Gradle release (`gradle wrapper` "builds" it from that one).
I have no clue why Gradle doesn't publish `gradle-wrapper.jar`s though. It's disappointing.

snazy: BTW: That approach is used in many Apache projects - nobody complained so far AFAIK

XN137: afaik the wrapper is build from scratch:
https://github.com/gradle/gradle/blob/master/platforms/core-runtime/wrapper-main/build.gradle.kts#L69-L85
the commited `gradle/wrapper/gradle-wrapper.jar` is what bootstraps the build of gradle though (kind of a hen egg problem)

but yeah i think there is a ticket with many interested parties asking for an official download link of the wrapper jar...

snazy: It's _re_-built, yes - but deterministically

adutra (approved):

PR #9239, opened 2024-08-01
Add signer-keys service

snazy:
Adds a service to provide secret keys that can be used to sign and verify signatures. Keys are rotated automatically and shared across all Nessie instances.

This code is going to be used to sign and verify the Nessie S3 sign endpoint URIs.

### DIFF
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.time.Instant;
+import java.util.Arrays;
 import javax.crypto.spec.SecretKeySpec;
 import org.immutables.value.Value;
 import org.projectnessie.nessie.immutables.NessieImmutable;
@@ -32,7 +33,7 @@ import org.projectnessie.nessie.immutables.NessieImmutable;
 public interface SignerKey {
   String name();
 
-  String secretKey();
+  byte[] secretKey();
 
   Instant creationTime();
 
@@ -43,14 +44,17 @@ public interface SignerKey {
   @JsonIgnore
   @Value.Lazy
   default SecretKeySpec secretKeySpec() {
-    byte[] secretKeyBytes = ("NessieSign:" + secretKey()).getBytes(UTF_8);
+    int pre = "NessieSign:".length();
+    int keyLen = secretKey().length;
+    byte[] secretKeyBytes = Arrays.copyOf("NessieSign:".getBytes(UTF_8), pre + keyLen);
+    System.arraycopy(secretKey(), 0, secretKeyBytes, pre, keyLen);
     return new SecretKeySpec(secretKeyBytes, "hmacSha256");
   }
 
   @Value.Check
   default void check() {
     checkState(!name().isEmpty(), "Key name must not be empty");
-    checkState(secretKey().length() >= 8, "Secret key too short");
+    checkState(secretKey().length >= 32, "Secret key too short");
     checkState(
         creationTime().compareTo(rotationTime()) < 0, "creationTime must be before rotationTime");
     checkState(
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
@@ -44,11 +44,7 @@ public interface SignerKey {
   @JsonIgnore
   @Value.Lazy
   default SecretKeySpec secretKeySpec() {
-    int pre = "NessieSign:".length();
-    int keyLen = secretKey().length;
-    byte[] secretKeyBytes = Arrays.copyOf("NessieSign:".getBytes(UTF_8), pre + keyLen);
-    System.arraycopy(secretKey(), 0, secretKeyBytes, pre, keyLen);
-    return new SecretKeySpec(secretKeyBytes, "hmacSha256");
+    return new SecretKeySpec(secretKey(), "hmacSha256");
   }
 
   @Value.Check
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
@@ -16,13 +16,11 @@
 package org.projectnessie.catalog.service.objtypes;
 
 import static com.google.common.base.Preconditions.checkState;
-import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.time.Instant;
-import java.util.Arrays;
 import javax.crypto.spec.SecretKeySpec;
 import org.immutables.value.Value;
 import org.projectnessie.nessie.immutables.NessieImmutable;
--- a/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
+++ b/catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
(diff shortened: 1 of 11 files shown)


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/SignerSignature.java
```diff
@@ -0,0 +1,106 @@
...
+import jakarta.validation.constraints.NotNull;
+import java.time.Clock;
+import java.time.Instant;
+import java.util.List;
+import java.util.Optional;
+import org.projectnessie.catalog.service.objtypes.SignerKey;
+import org.projectnessie.nessie.immutables.NessieImmutable;
+
+@NessieImmutable
+public abstract class SignerSignature {
```
snazy, line 32: This object is going to encapsulate all parameters for S3 signing, respecting writeable + read-only locations, object-storage-friendly write location and the signature.

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKeysObj.java
```diff
@@ -0,0 +1,77 @@
...
+    return OBJ_ID;
+  }
+
+  @Override
+  @Value.Default
+  default ObjType type() {
+    return OBJ_TYPE;
+  }
+
+  List<SignerKey> signerKeys();
```
adutra, line 58: Suggestion: turn this into `Map<String, SignerKey>` for faster key access in the method below.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Signer keys service implementation based on {@link Persist} using a {@link SignerKeysObj} per
+ * Nessie repository. Relies on Nessie's distributed cache invalidation, when used in a horizontally
+ * scaled setup (multiple Nessie processes serving the same repository).
+ */
+@RequestScoped
+public class SignerKeysServiceImpl implements SignerKeysService {
+  public static final long SPIN_LOOP_MIN_SLEEP_MILLIS = 20;
+  public static final long SPIN_LOOP_MAX_SLEEP_MILLIS = 200;
+  public static final Duration NEW_KEY_ROTATE_AFTER = Duration.of(3, DAYS);
```
adutra, line 50: Should this be configurable?

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
```diff
@@ -0,0 +1,64 @@
...
+import org.immutables.value.Value;
+import org.projectnessie.nessie.immutables.NessieImmutable;
+
+@NessieImmutable
+@JsonSerialize(as = ImmutableSignerKey.class)
+@JsonDeserialize(as = ImmutableSignerKey.class)
+public interface SignerKey {
+  String name();
+
+  String secretKey();
```
adutra, line 35:
```suggestion
  /** Base64-encoded secret key. */
  String secretKey();
```

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
```diff
@@ -0,0 +1,64 @@
...
+  Instant creationTime();
+
+  Instant rotationTime();
+
+  Instant expirationTime();
+
+  @JsonIgnore
+  @Value.Lazy
+  default SecretKeySpec secretKeySpec() {
+    byte[] secretKeyBytes = ("NessieSign:" + secretKey()).getBytes(UTF_8);
```
adutra, line 46: This looks suspicious, shouldn't we decode `secretKey()` (it's base64-encoded) first? And why prepend `"NessieSign:"`?

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
```diff
@@ -0,0 +1,64 @@
...
+    return new SecretKeySpec(secretKeyBytes, "hmacSha256");
```
adutra, line 47: Technically, this key is not a "hmacSha256" key. It's more a "custom" algorithm.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/SignerSignature.java
```diff
@@ -0,0 +1,105 @@
...
+
+  public abstract long expirationTimestamp();
+
+  @SuppressWarnings("UnstableApiUsage")
+  public String sign(SignerKey signerKey) {
+    Hasher hasher = hmacSha256(signerKey.secretKeySpec()).newHasher();
+    hasher.putLong(expirationTimestamp());
+    hasher.putString("prefix=" + prefix(), UTF_8);
+    hasher.putString("identifier=" + identifier(), UTF_8);
+    hasher.putString("b=" + warehouseLocation(), UTF_8);
```
adutra, line 51: Why `b`?

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
```diff
@@ -0,0 +1,178 @@
...
+            persist.fetchTypedObj(
+                SignerKeysObj.OBJ_ID, SignerKeysObj.OBJ_TYPE, SignerKeysObj.class);
+        break;
+      } catch (ObjNotFoundException notFound) {
+        signerKeys = addNewKey(null);
+        try {
+          if (storeInitial(signerKeys)) {
+            break;
+          }
+          persistSpinLoop();
```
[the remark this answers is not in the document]
snazy: Race w/ another Nessie instance storing the obj

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKeysObj.java
snazy: The access would be faster - deserialization, which happens every time the object's needed, slower.
The implementation also relies on the order of the elements (oldest first, newest last).

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
snazy: Yea - want to do that as a follow-up after #9170

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
snazy: Actually, this should be a byte[]

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
snazy: damn, right

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/SignerSignature.java
snazy: for "bucket" - "w" is already used for _writeable locations_

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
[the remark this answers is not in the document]
adutra: Yes, but now that this instance lost the insertion battle, why wait to fetch the object that the other instance inserted?

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKeysObj.java
adutra: Good point, maybe make this a derived/lazy attribute?

catalog/service/common/src/main/java/org/projectnessie/catalog/service/objtypes/SignerKey.java
adutra: The key is going to serve as the base key for a hmacSha256 encryption later on, but the key itself was generated by a custom algorithm. But it doesn't matter, it seems the value passed here is not relevant and is never read.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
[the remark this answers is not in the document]
snazy: True, removed

adutra (approved):

PR #9244, opened 2024-08-01
Add ability to generate IAM policies

snazy:
Required by #9170

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
@@ -16,9 +16,16 @@
 package org.projectnessie.catalog.files.s3;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static java.lang.String.format;
 import static java.util.Collections.singletonList;
 import static java.util.concurrent.CompletableFuture.completedFuture;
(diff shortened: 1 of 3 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3SessionsManager.java
```diff
@@ -212,6 +226,118 @@ private Credentials executeAssumeRoleRequest(
...
+
+  static class ParsedIamStatements {
+    static final LoadingCache<String, ObjectNode> STATEMENTS =
+        Caffeine.newBuilder()
+            .maximumSize(2000)
+            .expireAfterAccess(Duration.of(1, ChronoUnit.HOURS))
+            .build(ParsedIamStatements::parseStatement);
+
+    private static ObjectNode parseStatement(String stmt) {
+      try (MappingIterator<Object> values = MAPPER.readerFor(ObjectNode.class).readValues(stmt)) {
```
adutra, line 321: Why `readValues` instead of `readValue`? Do we expect `stmt` to be empty or to contain many JSON objects?

snazy: Necessary for configuration checks/validation.

adutra (approved):

PR #9253, opened 2024-08-02
Drop support for Java 8

snazy:


adutra (dismissed):

catalog/format/iceberg-bench/build.gradle.kts
```diff
@@ -17,7 +17,7 @@
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 
 plugins {
-  id("nessie-conventions-client")
+  id("nessie-conventions-server")
```
XN137, line 20: the change looks a bit odd as in other places we change from `nessie-conventions-server` to `nessie-conventions-client` :thinking: 

are we saying both client and server components for iceberg now require j11 ?
or does this fix a previous misconfiguration of using client for a server component?

tools/content-generator/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 
 plugins {
   alias(libs.plugins.nessie.run)
-  id("nessie-conventions-server")
+  id("nessie-conventions-client")
```
XN137, line 21: here we switch conventions the other way around.
as both are basically the same convention now, it may not matter but it looks confusing in the diff.

XN137 (approved):
for my understanding: this PR assumes that all downstream projects are now building for java11 and also exclusively use the jakarta annotations instead of javax?

generally lgtm, i guess we'll know if we missed anything in the downstream projects later on

catalog/format/iceberg-bench/build.gradle.kts
snazy: Iceberg requires Java >=11 now

snazy:
> for my understanding: this PR assumes that all downstream projects are now building for java11 

yes

> also exclusively use the jakarta annotations instead of javax?

no - jakarta/javax didn't change in this PR

catalog/format/iceberg-bench/build.gradle.kts
XN137: yes but that explains why the convention themselves would change, not why iceberg-bench would switch from client to server

PR #9281, opened 2024-08-06
Build / IntelliJ: include the project root dir in the IDE window name

adutra:
I'd like to propose that we include the project's directory in the window name for IntelliJ (it's included by default, but our custom name doesn't include it).

The main reason for me is that I have several git workspaces for Nessie being worked on in parallel, and knowing which workspace I'm in saves me a few seconds each time I switch workspaces.

(I also think the version in the title is actually not that useful, because it changes frequently, but the title won't change until you do a Gradle build.)

### DIFF
--- a/build-logic/src/main/kotlin/nessie-conventions-root.gradle.kts
+++ b/build-logic/src/main/kotlin/nessie-conventions-root.gradle.kts
@@ -35,7 +35,8 @@ plugins {
 loadNessieProjects(rootProject)
 
 val projectName = rootProject.file("ide-name.txt").readText().trim()
-val ideName = "$projectName ${rootProject.version.toString().replace(Regex("^([0-9.]+).*"), "$1")}"
+val ideName =
+  "$projectName ${rootProject.version.toString().replace(Regex("^([0-9.]+).*"), "$1")} [${rootProject.rootDir}]"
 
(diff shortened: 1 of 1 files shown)


build-logic/src/main/kotlin/nessie-conventions-root.gradle.kts
```diff
@@ -35,7 +35,8 @@ plugins {
 loadNessieProjects(rootProject)
 
 val projectName = rootProject.file("ide-name.txt").readText().trim()
-val ideName = "$projectName ${rootProject.version.toString().replace(Regex("^([0-9.]+).*"), "$1")}"
+val ideName =
+  "$projectName ${rootProject.version.toString().replace(Regex("^([0-9.]+).*"), "$1")} [${rootProject.rootDir}]"
```
snazy, line 39: Would the last path element be sufficient?
```suggestion
  "$projectName } [in ../${rootProject.rootDir.name}]"
```

I'm okay to remove the version

adutra: Let's leave the version for now since it's used for Eclipse as well, and maybe Eclipse has a better synchronization.

snazy (dismissed):

snazy (approved):

dimas-b (approved):

PR #9282, opened 2024-08-06
Prepare for Jakarta Servlet Spec 6

snazy:
Effort to prepare for [Jakarta Servlet Specification 6, URI Path Canonicalization](https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization) and [RFC 3986, Section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3), which disallows control characters (< 0x20) and more. Most importantly, we use `%1D` to "escape" the dot (`.`) character in content key elements - URLs that contain `%1D` will be rejected. The other problematic character is `/`, even in its URL encoded form.

To indicate that a server supports the new "escaping", it announces the specVerion `2.2.0` (bumped from `2.1.0`). The default client spec version has been bumped from `2` to `3`.

See [NESSIE-SPEC-2.0.md updated in this change](https://github.com/projectnessie/nessie/blob/main/api/NESSIE-SPEC-2-0.md).

Via a future PR, the implementations in `org.projectnessie.client.rest.v2` can check the spec version announced by the service and use either the new escaping or the legacy control character. The most problematic part are CEL filters that are constructed from application code.

For posterity, the `Namespace`/`ContentKey`s' `toPathString()` function got two counterparts `toPathStringControlChars()` and `toPathStringEscaped()`, with the intent to eventually deprecate `toPathString()`.

Overall, the behavior on the client side must eventually change but not yet - there's still some time. At some point, we can require services to run with spec version 2.2.0 (services would reject lower versions) and services checking for clients announcing spec version 3 - when there's no way around _Jakarta Servlet Specification 6_.


api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -44,15 +43,70 @@ private Util() {}
...
+   *
+   * <p>Escaping, for compatibility w/ <a
+   * href="https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization">Jakarta
+   * Servlet Specification 6, URI Path Canonicalization</a>. This is available with Nessie Spec
+   * version 2.2.0, as advertised via {@link NessieConfiguration#getSpecVersion()}.
+   *
+   * <ul>
+   *   <li>Fact: two consecutive dots ({@code ..}) would represent an <em>empty</em> namespace
+   *       elements, which is illegal. It is correct to assume that this character sequence does not
+   *       occur in encoded content keys.
```
dimas-b, line 61: Not sure. `..` is removed only if it is a path segment, i.e. it occurs withing `/` separators line `some/../path` -> `some/path`. A path line `some..path` is still legal, I think.

api/client/src/main/java/org/projectnessie/client/api/ns/ClientSideGetNamespace.java
```diff
@@ -58,13 +58,13 @@ public GetNamespaceResult getWithResponse()
       if (!(c instanceof Namespace)) {
         throw new NessieNamespaceNotFoundException(
             contentKeyErrorDetails(key),
-            String.format("Namespace '%s' does not exist", key.toPathString()));
+            String.format("Namespace '%s' does not exist", key.toString()));
```
adutra, line 61: Given the complexity of what we are trying to achieve, I wonder if relying on `toString()` is a good move, even for a mere error message.

snazy: I don't have a good idea what to put into those messages.
`toPathString()` gives the control-character (which is often not really visible)
The encoded form looks weird to humans.
That's why I went with the native `.join(".")`

dimas-b:
> The most problematic part are CEL filters that are constructed from application code.

I suppose we can keep the old encoding for CEL expressions (they are not affected by the Servlet spec) and maybe expose the new encoding as a separate variable.

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -44,15 +43,70 @@ private Util() {}
...
+   *
+   * <p>The {@code encoded} parameter is split at dot ({@code .}) characters.
+   *
+   * <p>Legacy compatibility: a dot ({@code .}) character can be represented using ASCII 31 (0x1F)
+   * or ASCII 0 (0x00).
+   *
+   * <p>Escaping, for compatibility w/ <a
+   * href="https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization">Jakarta
+   * Servlet Specification 6, URI Path Canonicalization</a>. This is available with Nessie Spec
+   * version 2.2.0, as advertised via {@link NessieConfiguration#getSpecVersion()}.
```
dimas-b, line 56: It looks like `/` is not allowed (by default) in the body of path segments: `Rejecting Suspicious Sequences` -> `The encoded "/" character (e.g. /path%2Finfo)` (also `\`).

`ContentKeys` allow those chars, so I think we have to escape them too using custom escaping because that URI-encoding is transparent to the new validation rules.

dimas-b: Perhaps also the `#` (fragment) char

api/model/src/main/java/org/projectnessie/model/Util.java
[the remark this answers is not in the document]
dimas-b: but `..bar` is not a valid `ContentKey`

api/client/src/main/java/org/projectnessie/client/api/ns/ClientSideGetNamespace.java
dimas-b: I'd say we ought to define our own canonical `ContentKey` textual representation (consistent with servlet path rules). It would be useful for CLI parameters too.

As for `.toString()` in error messages, I personally prefer the less visually pleasing but unambiguous encoded form.

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -44,15 +43,70 @@ private Util() {}
...
+   *
+   * <ul>
+   *   <li>Fact: two consecutive dots ({@code ..}) would represent an <em>empty</em> namespace
```
dimas-b, line 59: Re: the specific encoding method... just my 2 cents: I think we can make it visually acceptable by using a rare char for escaping, e.g. `~`. So a literal period would be `~.`. Since `~` used to be a valid char, we can indicate the new encoding scheme by using `.` as the prefix, so examples:

`some.namespace` 
`some%1Dnamespace` - old escaping
`.some~.namespace` - new escaping
`some~namespace` - old literal `~`
`.some~~namespace` - new literal `~`

I hope in this case most textual respresentations are human-readable.

This is just for discussion. WDYT?

snazy: `~.` is ambiguous - `~` could be the last char of the element.

snazy:
I'll pull in some servlet-spec v6 URI validation code

api/model/src/main/java/org/projectnessie/model/Util.java
dimas-b: Leading `.` means all `~` are special.

api/model/src/testUriCompliance/java/org/projectnessie/model/TestContentKeyEncoding.java
```diff
@@ -0,0 +1,184 @@
...
+      List<String> elements,
+      Collection<UriCompliance.Violation> expectedViolations) {
+    List<String> actualElements = Util.fromPathString(encoded);
+    soft.assertThat(actualElements).containsExactlyElementsOf(elements);
+    soft.assertThat(actualElements).containsExactlyElementsOf(Util.fromPathString(encoded));
+
+    HttpURI uri = HttpURI.from(format("http://hostname/%s", encoded));
+    soft.assertThat(uri.getViolations())
+        .describedAs("path-beginning: %s", encoded)
+        .containsExactlyInAnyOrderElementsOf(expectedViolations);
```
dimas-b, line 54: I'm not sure it's a sufficient test.... e.g. when `#` is part of `encoded` it would be treated as a fragment by the URI, thus corrupting key data. I think we have to URI-encode `#` then run it through `HttpURI`, check violations, then get the path and parse the path from `HttpURI`. WDYT?

snazy: I think you're overthinking this. `#` for example is escaped using `UrlEncoder`.

dimas-b: maybe... too much context switching :sweat_smile: I'll try it locally :hourglass:

api/client/src/main/java/org/projectnessie/client/api/ns/ClientSideGetNamespace.java
adutra: `toHumanReadableString()` ?

api/model/src/testUriCompliance/java/org/projectnessie/model/TestContentKeyEncoding.java
dimas-b: reviewing today (hopefully)

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -61,9 +124,39 @@ public static List<String> fromPathString(String encoded) {
...
+    for (String element : elements) {
+      if (sb.length() > 0) {
+        sb.append('.');
+      }
+      int l = element.length();
+      for (int i = 0; i < l; i++) {
+        char c = element.charAt(i);
+        if (c == DOT) {
+          sb.append("..~");
+        } else if (c == '/') {
```
dimas-b, line 152: I believe we also need to handle `\` (example value `hellow\world`) and `%` (ex: `%hellow`). They cause URI violations.

api/model/src/testUriCompliance/java/org/projectnessie/model/TestContentKeyEncoding.java
dimas-b: added a separate comment, resolving this thread.

api/model/src/main/java/org/projectnessie/model/Util.java
dimas-b: suggestion: since we have multiple escape cases, maybe add a generic translation table for the char following `..`?

api/client/src/main/java/org/projectnessie/client/api/ns/ClientSideGetNamespace.java
dimas-b: I think we have two major case:
1. Escaping the `.` separator - usable eveywhere full unicode is supported (e.g. CLI, JSON) - Let's call it `toCanonicalString()`?
2. Path escaping - the `.` plus all "suspicious" chars according to the new servlet spec - old `toPathString()` method.

I'd use (1) for error messages and `.toString()`. WDYT?

adutra: +1 for `toCanonicalString()`

api/model/src/main/java/org/projectnessie/model/Util.java
adutra: The leading dot to indicate the new encoding scheme is an interesting idea imho.

2 suggestions:

1. Can we use an escape char that doesn't get url-encoded? `~` becomes `%5E`. These chars remain the same: `-`, `*`, and `_` 
2. Could we use the leading dot only if there are escaped sequences? This way if there is no leading dot we know that it's either old-style encoding, or a simple namespace that doesn't need escaping.

dimas-b: 1) absolutely... choosing a char may be tricky, `-` and `_` are common in names and `*` is special in shell :)

2) yes, that's the idea. It would make hand-writing keys easy (I hope).... e.g. in CLI args. 

Minor point: `.some.name` and `some.name` would be equivalent (`.` is the separator here). We can choose which one is the "canonical" form.

api/model/src/main/java/org/projectnessie/model/Util.java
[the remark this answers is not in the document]
snazy: Yep - the comment refers to `..` for a content-key (would be an empty namespace element).
`..` alone is invalid as a URI path element - but `..` somewhere in a URI path element is valid.

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -44,15 +53,91 @@ private Util() {}
...
-        .map(x -> x.replace(GROUP_SEPARATOR, DOT).replace(ZERO_BYTE, DOT))
-        .collect(Collectors.toList());
+    List<String> elements = new ArrayList<>();
+    int l = encoded.length();
+    StringBuilder e = new StringBuilder();
+    for (int i = 0; i < l; i++) {
+      char c = encoded.charAt(i);
+      switch (c) {
+        case DOT:
+          if (encoded.charAt(i + 1) == DOT) {
```
dimas-b, line 98: Could this hit a `StringIndexOutOfBoundsException`?

api/model/src/main/resources/org/projectnessie/model/NessieConfiguration.json
```diff
@@ -1,5 +1,5 @@
 {
   "minSupportedApiVersion": 1,
   "maxSupportedApiVersion": 2,
-  "specVersion": "2.1.0"
+  "specVersion": "2.2.0"
```
dimas-b, line 4: Should we have a section for the new version in `NESSIE-SPEC-2-0.md`?

api/model/src/testUriCompliance/java/org/projectnessie/model/TestContentKeyEncoding.java
```diff
@@ -0,0 +1,270 @@
...
+        arguments("he/llo", List.of("he/llo"), Set.of(AMBIGUOUS_PATH_SEPARATOR)),
+        // control-characters cause violations
+        arguments(
+            "he\u001dllo.wo\u001drld",
+            List.of("he.llo", "wo.rld"),
+            Set.of(SUSPICIOUS_PATH_CHARACTERS)),
+        // backslash is a violation
+        arguments("he\\llo.world", List.of("he\\llo", "world"), Set.of(SUSPICIOUS_PATH_CHARACTERS)),
+        // % alone is ambiguous
+        arguments("he%llo.world", List.of("he%llo", "world"), Set.of(AMBIGUOUS_PATH_ENCODING)));
```
dimas-b, line 123: WDYT about adding test cases for escaped `%` and `\` chars (no violations)?

api/client/src/main/java/org/projectnessie/client/api/ns/ClientSideGetNamespace.java
dimas-b: @snazy : WDYT about `toCanonicalString()`?

api/model/src/main/java/org/projectnessie/model/Util.java
snazy: s..t in, s..t out ;)

api/model/src/main/java/org/projectnessie/model/Util.java
snazy: 2 * (512 bytes + array header) for 4 cases feels overkill

api/model/src/testUriCompliance/java/org/projectnessie/model/TestContentKeyEncoding.java
snazy: This test isn't about testing violations - it's about testing `fromPathString` and asserting that the new mechanism doesn't cause violations. There's no URL escaping happening here.

api/model/src/main/java/org/projectnessie/model/Util.java
snazy: Described the new approach in the PR description - hope I got all the dots right.

snazy: also `toCanonicalString` as a "weak" variant of the escaped one

api/model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -198,24 +198,136 @@ protected void validate() {
...
+
   /**
-   * Convert this key to a url encoded path string.
+   * Convert these elements to a URI path/query string, which <em>may violate</em> <a
+   * href="https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#uri-path-canonicalization">Jakarta
+   * Servlet Spec 6, chapter 3.5.2 URI Path Canonicalization</a>, because it uses so-called control
+   * characters and ambiguous path elements.
    *
-   * @return String encoded for path use.
+   * <p>Users must prefer {@link #toPathStringEscaped()}, if the service support Nessie spec version
```
adutra, line 239:
```suggestion
   * <p>Users must prefer {@link #toPathStringEscaped()}, if the service supports Nessie spec version
```

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -25,45 +25,324 @@
...
+   *       elements</b> use the "problematic character escaping":
+   *       <ul>
+   *         <li>a {@code .} is escaped as {@code ._}
+   *         <li>a {@code /} is escaped as <code>.{</code>
+   *         <li>a {@code \} is escaped as <code>.}</code>
+   *         <li>a {@code %} is escaped as {@code .[}
+   *       </ul>
+   * </ul>
+   *
+   * <p>Some examples:
```
adutra, line 189: I think I found a situation where this encoding scheme does not work:

```java
  @Test
  void roundTrip() {
    List<String> expected = List.of("abc.", "{def");
    String escaped = Util.toPathStringEscaped(expected);
    List<String> actual = Util.fromPathString(escaped);
    soft.assertThat(actual).isEqualTo(expected);
  }
```

```
org.opentest4j.AssertionFailedError: 
expected: ["abc.", "{def"]
 but was: ["abc./def"]
```

snazy: Right - seems like it's necessary to use `..` when it's escaping.

snazy:
Updated the algorithm and added a ton more test cases, also cleaning up the existing tests.

Introduced `Elements` as a base interface to removed duplicated code + mostly duplicated javadoc.

api/model/src/main/java/org/projectnessie/model/Elements.java
```diff
@@ -0,0 +1,203 @@
...
+        throw new IllegalArgumentException(
+            format("%s '%s' must not contain a null element", type, elements));
+      }
+      int l = e.length();
+      if (l == 0) {
+        throw new IllegalArgumentException(
+            format("%s '%s' must not contain an empty element", type, elements));
+      }
+      if (".".equals(e)) {
+        throw new IllegalArgumentException(
```
adutra, line 184: This was previously thrown only for the last element of a `Namespace`. It wasn't thrown for `ContentKey` at all, and also not for non-last elements of a `Namespace`.

api/model/src/main/resources/org/projectnessie/model/NessieConfiguration.json
snazy: Added that in #9289

api/model/src/main/java/org/projectnessie/model/Elements.java
adutra: Maybe you could remove this check here and add it in `Namespace.validate()`, then you don't need the `boolean nsSingleDot` parameter, wdyt?

```java
  @Override
  public void validate(String type) {
    Elements.super.validate(type);
    String last = getElements().get(getElements().size() - 1);
    if (DOT_STRING.equals(last)) {
          format("Namespace '%s' must not contain a '.' element", toCanonicalString()));
    }
  }
```

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -25,45 +25,231 @@
...
-import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.stream.Collectors;
 import org.projectnessie.model.types.ContentTypes;
 import org.projectnessie.model.types.RepositoryConfigTypes;
 
 final class Util {
 
+  public static final char DOT = '.';
```
adutra, line 35: All these constants could be private it seems

api/model/src/main/java/org/projectnessie/model/Namespace.java
```diff
@@ -165,36 +168,24 @@ public static Namespace of(Map<String, String> properties, String... elements) {
...
 
+  @Value.Check
+  protected void validate() {
+    Elements.super.validate("Namespace");
+
+    int elementCount = getElementCount();
+    List<String> elements = getElements();
+    if (elementCount > 0 && ".".equals(elements.get(elementCount - 1))) {
+      throw new IllegalArgumentException(
+          format("Namespace '%s' must not contain a '.' element", elements));
```
adutra, line 185: One more nit, sorry: the constant `ERROR_MSG_TEMPLATE` is still in use, and it's used for a similar error message, see `Namespace#parse()`. I'd suggest to remove that usage and delete the constant (since the check will be done later on by  this method).

snazy: done

api/model/src/main/java/org/projectnessie/model/Util.java
```diff
@@ -25,45 +25,231 @@
...
+      char c = element.charAt(i);
+      switch (c) {
+        case DOT:
+        case SLASH:
+        case BACKSLASH:
+        case PERCENT:
+          if (!escaped) {
+            // Not yet "escaped", let the first escaped element start with a '.'
+            if (sb.length() > sbl) {
+              sb.insert(sbl, DOT);
```
dimas-b, line 170: WDYT about putting the escaping mark (`.`) always as the first char of the whole string? I think it would make writing escaped strings easier for humans.

For example: `foo..._bar` becomes `.foo.._bar` - it it easy to see that escaping is "on" and it is easier to interpret the second dot group as a separator followed by an escaped char.

I still feel that using a different rare char for escaping, makes the string even more human-readable IMHO, e.g. if `*` is the escape char, `foo..._bar` becomes `.foo.~.bar`, `.foo._bar` -> `.foo~.bar`, `.foo~bar` -> `.foo~~bar`.

Futhermore, in escaped mode the escape char is always followed by one of the special chars (`~.` - dot `.{` - slash, etc..). `.` is _always_ the elements separator.

This should also help with canonicalization. We could say that canonical strings must be in escape more. Still, simply adding a leading `.` will convert the key to its canonical form, e.g. `foo.bar` -> `.foo.bar`. Of course, `~` will have to be escaped: `foo~bar` -> `.foo~~bar`, `foo/bar` -> `.foo~{bar`. However, my point is that it is fairly intuitive for humans to do.

Also, there is no ambiguity in constructing the canonical string, while currently `foo.._bar` and `foo._bar` both parse to `["foo", "_bar"]`.

Both unescaped and canonical strings will be accepted on input, but we will be able to use only canonical strings in CEL expressions, hashing, and other places that require deterministic behaviour.

While `~` is converted to `%7E` by `URLEncoder`, it is "unreserved" in [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986), so humans will be able to use it with `curt` directly, e.g. `curl http://localhost:19120/api/v2/trees/main/contents/.foo.~.bar`

More  examples: (array, handwritten, canonical, for URI path)

`["a","b","c"]` - `a.b.c` - `.a.b.c` -  (either handwritten or canonical accepted)
`["a","b~c"]` - `a.b~c` - `.a.b~~c` - (either handwritten or canonical accepted)
`["a","b.c"]` - `a.b/c` - `.a.b/c` - `.a.b~{c`
`["a","b.c"]` - `.a.b~.c` - `.a.b~.c` - `.a.b~.c`

WDYT?

dimas-b (commented):
Sorry, lost of comments... I only managed to review end-to-end now.

api/model/src/main/java/org/projectnessie/api/v2/doc/ApiDoc.java
```diff
@@ -132,8 +132,23 @@ public interface ApiDoc {
...
+      "Content key and namespace components are separated by the dot ('.') character.\n"
+          + "The components itself must be escaped using the following rules:\n"
+          + "\n"
+          + "_Legacy behavior_, not compatible with Jakarta Servlet Spec v6 (all Nessie spec versions): "
+          + "dot ('.') characters that are not Nessie namespace separators must be encoded as the 'group separator' ASCII character (0x1D).\n"
+          + "\n"
+          + "_Escaping behavior_, compatible with Jakarta Servlet Spec v6 (since Nessie spec 2.2.0, see `NessieConfiguration.spec-version`): "
+          + "Keys must be escaped using the following rules:\n"
+          + "* A dot `.` in an element serialized as `..~`\n"
+          + "* A slash `/` in an element serialized as `..^`\n"
```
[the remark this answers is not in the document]
snazy: updated

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -530,7 +530,7 @@ private void s3ConfigOverrides(
 
         long expirationTimestamp = systemUTC().instant().plus(3, ChronoUnit.HOURS).getEpochSecond();
 
-        String contentKeyPathString = contentKey.toPathString();
+        String contentKeyPathString = contentKey.toPathStringEscaped();
```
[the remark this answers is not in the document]
snazy: Let's not add too much more to this PR.

api/model/src/main/java/org/projectnessie/model/Elements.java
```diff
@@ -0,0 +1,202 @@
...
+   *
+   * <p>Legacy compatibility: a dot ({@code .}) character can be represented using ASCII 31 (0x1F)
+   * or ASCII 0 (0x00).
+   *
+   * <p>Escaping, for compatibility w/ <a
+   * href="https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0.html#uri-path-canonicalization">Jakarta
+   * Servlet Specification 6, URI Path Canonicalization</a> and <a
+   * href="https://datatracker.ietf.org/doc/html/rfc3986#section-3.3">RFC 3986, section 3.3</a>.
+   * This is available with Nessie Spec version 2.2.0, as advertised via {@link
+   * NessieConfiguration#getSpecVersion()}. If escaping is active, two dots characters {@code ..}
```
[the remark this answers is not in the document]
snazy: TBH, I'm already a bit tired of updating descriptions. Later

api/model/src/main/java/org/projectnessie/model/ContentKey.java
```diff
@@ -198,24 +198,132 @@ protected void validate() {
...
     return Util.toPathString(getElements());
   }
 
+  /**
+   * Convert these elements to a URI path/query string, which <em>may violate</em> <a
+   * href="https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#uri-path-canonicalization">Jakarta
+   * Servlet Spec 6, chapter 3.5.2 URI Path Canonicalization</a>, because it uses so-called control
+   * characters and ambiguous path elements.
+   *
+   * <p>Users must prefer {@link #toPathStringEscaped()}, if the service support Nessie spec version
```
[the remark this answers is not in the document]
snazy: removed, needed it temporarily

snazy:
Updated escaping again.
`*` is the new escape character
Escaped representations start with a `.`

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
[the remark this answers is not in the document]
dimas-b: fair enough.

dimas-b (approved):
LGTM :+1: Thx, @snazy !

PR #9289, opened 2024-08-07
Adopt Nessie REST API + client

snazy:
Follow up of #9282, which introduced content-key escaping that is compatible with Jakarta Servlet Spec v6.

This change tackles all endpoints that have query parameters with "problematic" characters (according to section 3.5.2 of the Jakarta Servlet Spec v6), either a content-key or a CEL filter (the latter is more difficult) and adds `POST` counterparts to the existing Nessie REST API v2 endpoints.

Unlike noted in #9282, it seems safe to add "Nessie spec version gate" at this point.

The Nessie REST API v2 client has been adopted to use the new endpoints, if the Nessie service announces Nessie spec >= 2.2.0.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
@@ -20,6 +20,7 @@ import org.projectnessie.client.api.ns.ClientSideDeleteNamespace;
 import org.projectnessie.client.api.ns.ClientSideGetMultipleNamespaces;
 import org.projectnessie.client.api.ns.ClientSideGetNamespace;
 import org.projectnessie.client.api.ns.ClientSideUpdateNamespace;
+import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.Reference;
 
 /**
@@ -31,6 +32,8 @@ import org.projectnessie.model.Reference;
  * <p>Most changes between v1 and v2 exist at the REST level (HTTP).
  */
 public interface NessieApiV2 extends NessieApiV1 {
+
+  String toPathString(ContentKey key);
 
   GetRepositoryConfigBuilder getRepositoryConfig();
 
--- a/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
@@ -20,7 +20,6 @@ import org.projectnessie.client.api.ns.ClientSideDeleteNamespace;
 import org.projectnessie.client.api.ns.ClientSideGetMultipleNamespaces;
 import org.projectnessie.client.api.ns.ClientSideGetNamespace;
 import org.projectnessie.client.api.ns.ClientSideUpdateNamespace;
-import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.Reference;
 
 /**
@@ -32,8 +31,6 @@ import org.projectnessie.model.Reference;
  * <p>Most changes between v1 and v2 exist at the REST level (HTTP).
  */
 public interface NessieApiV2 extends NessieApiV1 {
-
-  String toPathString(ContentKey key);
 
   GetRepositoryConfigBuilder getRepositoryConfig();
 
--- a/api/client/src/main/java/org/projectnessie/client/rest/v2/HttpApiV2.java
+++ b/api/client/src/main/java/org/projectnessie/client/rest/v2/HttpApiV2.java
@@ -15,6 +15,7 @@
  */
 package org.projectnessie.client.rest.v2;
 
+import java.util.Arrays;
 import java.util.Optional;
 import org.projectnessie.client.api.AssignBranchBuilder;
 import org.projectnessie.client.api.AssignReferenceBuilder;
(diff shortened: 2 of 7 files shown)


api/client/src/main/java/org/projectnessie/client/rest/v2/HttpGetDiff.java
```diff
@@ -48,27 +51,33 @@ protected DiffParams params() {
...
-            .queryParam("max-records", params.maxRecords())
-            .queryParam("page-token", params.pageToken())
-            .queryParam("filter", params.getFilter());
-    params.getRequestedKeys().forEach(k -> req.queryParam("key", k.toPathString()));
+    HttpRequest req = client.newRequest();
+
+    if (api.isNessieSpec220()) {
+      return req.path("trees/.diff")
+          .unwrap(NessieNotFoundException.class)
+          .post(ImmutableDiffJson.builder().from(params).build())
```
dimas-b, line 59: I'm not sure about checking the spec version for choosing request methods is in line with how the spec version was envisioned to be used... The spec version defines server behaviour, so it's file to use it as a way to determine whether the server is able to parse the new `ContentKey` encoding, but as far as REST API goes, I think we have to bump the API version for new methods (even if it's backward compatible).

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
```diff
@@ -110,6 +115,36 @@ public interface HttpTreeApi extends TreeApi {
   @JsonView(Views.V2.class)
   ReferencesResponse getAllReferences(@BeanParam @jakarta.ws.rs.BeanParam ReferencesParams params);
 
+  @Override
+  @POST
+  @jakarta.ws.rs.POST
+  @Produces(MediaType.APPLICATION_JSON)
+  @jakarta.ws.rs.Produces(jakarta.ws.rs.core.MediaType.APPLICATION_JSON)
+  @Path(".all")
+  @jakarta.ws.rs.Path(".all")
```
dimas-b, line 124: `POST /api/v2/trees/.all` to _fetch_ references sounds counter-intuitive to me (based on REST principles).

dimas-b (commented):
> all endpoints that have problematic query parameters [...]

Could you be a bit more specific about this? I'm not sure I understand the problem :sweat_smile:

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
snazy: We have to use POST/PUT in this case, because CEL filters still contain problematic characters.

api/client/src/main/java/org/projectnessie/client/rest/v2/HttpGetAllReferences.java
```diff
@@ -41,9 +45,15 @@ protected ReferencesParams params() {
 
   @Override
   protected ReferencesResponse get(ReferencesParams p) {
-    return client
-        .newRequest()
-        .path("trees")
+    HttpRequest req = client.newRequest();
+
+    if (api.isNessieSpec220()) {
+      return req.path("trees/.all")
```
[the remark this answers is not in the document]
snazy: CEL filters

api/client/src/main/java/org/projectnessie/client/rest/v2/HttpApiV2.java
```diff
@@ -64,7 +81,15 @@ public <C> Optional<C> unwrapClient(Class<C> clientType) {
 
   @Override
   public NessieConfiguration getConfig() {
-    return client.newRequest().path("config").get().readEntity(NessieConfiguration.class);
+    NessieConfiguration config =
+        client.newRequest().path("config").get().readEntity(NessieConfiguration.class);
+    didGetConfig = true;
+    if (config.getSpecVersion() != null) {
```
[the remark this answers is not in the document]
snazy: Let's better do that later.
`NessieApiCompatibilityFilter` is on a completely different code path - and it's not mandatory.

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
dimas-b: but they are not in URI paths :thinking:

api/client/src/main/java/org/projectnessie/client/rest/v2/HttpGetAllReferences.java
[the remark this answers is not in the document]
dimas-b: CEL filters are query parameters - the new Servlet spec does not add any new restrictions there... or did I miss it?

api/model/src/main/java/org/projectnessie/api/v2/http/HttpTreeApi.java
snazy: > but they are not in URI paths 🤔

:facepalm:

api/client/src/main/java/org/projectnessie/client/api/NessieApiV2.java
```diff
@@ -32,6 +33,8 @@
  */
 public interface NessieApiV2 extends NessieApiV1 {
 
+  String toPathString(ContentKey key);
```
dimas-b, line 36: WDYT about moving this declaration to `HttpApiV2`? Technically "path strings" are relevant at the HTTP level.

Callers of this method are also all in the `rest/v2` package.

dimas-b (commented):
LGTM :+1: Just one minor concern about the location of the `toPathString(ContentKey)` method.

api/client/src/main/java/org/projectnessie/client/rest/v2/HttpApiV2.java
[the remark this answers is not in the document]
dimas-b: fair enough.

dimas-b (approved):

PR #9301, opened 2024-08-08
Catalog: Implementation for short-lived user-delegation SAS tokens

### LINKED ISSUE #8909: [Catalog/ADLS] Expose user-delegation SAS
The concept used by `com.azure.storage.file.datalake.DataLakeFileSystemClient.generateUserDelegationSas()` looks similar to what S3 offers.

snazy:
Implements the production code for short-lived user-delegation SAS tokens. Sadly, Azurite doesn't work with this, so the code is effectively untested. In a follow-up it feels appropriate to cache the user-delegation-keys, which have a relatively long lifetime and are needed to generate the actual SAS tokens (only the latter are passed down to clients).

Fixes #8909

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsFileSystemOptions.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsFileSystemOptions.java
@@ -16,11 +16,15 @@
 package org.projectnessie.catalog.files.adls;
 
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Optional;
 import org.projectnessie.catalog.secrets.BasicCredentials;
 import org.projectnessie.catalog.secrets.KeySecret;
 
 public interface AdlsFileSystemOptions {
+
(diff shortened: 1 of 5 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsFileSystemOptions.java
```diff
@@ -35,6 +39,30 @@ public interface AdlsFileSystemOptions {
...
+   * <p>The current default is to not enable short-lived and scoped-down credentials, but the
+   * default may change to enable in the future.
+   */
+  Optional<Boolean> userDelegationEnable();
+
+  /**
+   * Expiration time / validity duration of the user-delegation <em>key</em>, this key is
+   * <em>not</em> passed to the client.
+   *
+   * <p>Defaults to 7 days minus 1 minute (the maximum), must be >= 1 second.
```
dimas-b, line 54: Why do we need a long key expiry default? It looks like we discard the key after obtaining the SAS token? :thinking:

dimas-b (approved):
LGTM overall :+1:

PR #9303, opened 2024-08-08
Doc generator: fix incorrect prefix when property has custom section

adutra:


tools/doc-generator/doclet/src/test/java/tests/smallrye/AllTypes.java
```diff
@@ -106,4 +108,9 @@ public interface AllTypes extends IntOne {
   /** Map of string to {@code MappedA}. */
   @ConfigPropertyName("mappy")
   Map<String, MappedA> mapStringMappedA();
+
+  /** Another map of string to {@code MappedA}, in its own section. */
+  @ConfigItem(section = "Section A")
+  @WithName("map.py")
```
snazy, line 114: Is this scenario causing the `sts.sts` ?

tools/doc-generator/doclet/src/main/java/org/projectnessie/nessie/docgen/SmallryeConfigs.java
```diff
@@ -72,9 +72,12 @@ public List<SmallRyeConfigSection> buildConfigSections(DocletEnvironment env) {
       additionalPrefixes.forEach(
```
adutra, line 72: There is imo a slight risk here that a prefix override (aka section name) could clash with an existing prefix. I didn't fix that.

snazy: Wouldn't that mean that we have duplicate property keys?

tools/doc-generator/doclet/src/test/java/tests/smallrye/AllTypes.java
adutra: The issue was with `@ConfigItem(section = "sts")` : the section name was being appended to the property prefix, which is what we want for the file name, but not for displaying the property names.

the `@WithName` is unrelated, but since we don't have tests with that, I added one.

snazy (approved):
Please verify manually that the docs are fine before merging.

tools/doc-generator/doclet/src/main/java/org/projectnessie/nessie/docgen/SmallryeConfigs.java
adutra: I fear something like this:

```java

@ConfigMapping(prefix = "foo")
public interface Clash {

  String bar();

  @ConfigItem(section = "foo.bar")
  String qix();

}
```

PR #9315, opened 2024-08-09
OAuth2 client: introduce secret suppliers

adutra:
This allows secrets to be dynamically resolved, and only when necessary.

Also deprecates the org.projectnessie.client.auth.oauth2.Secret class, for removal.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
@@ -130,7 +130,7 @@ abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig {
 
   @Value.Derived
   boolean isPublicClient() {
-    return !getClientSecret().isPresent();
+    return !getClientSecretSupplier().isPresent();
   }
 
   @Value.Lazy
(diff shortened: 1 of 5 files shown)


snazy (approved):

sudheeshkatkam:
LGTM!

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -204,8 +204,9 @@ URI getResolvedImpersonationTokenEndpoint() {
    */
   @Value.Lazy
   Optional<HttpAuthentication> getBasicAuthentication() {
-    return getClientSecret()
-        .map(s -> BasicAuthenticationProvider.create(getClientId(), s.getString()));
+    return getClientSecretSupplier()
+        .map(Supplier::get)
+        .map(s -> BasicAuthenticationProvider.create(getClientId(), s));
```
dimas-b, line 209: Sorry for the late review. If the intention of the `getClientSecretSupplier()` to to dynamically resolve the secret (potentially return a different value on every call), should we not call it from inside the `BasicAuthentication` when it is actually needed for HTTP calls?

adutra: True, I will open another PR.

PR #9319, opened 2024-08-09
BasicAuthenticationProvider: ability to dynamically provide a password

adutra:
Also fixes a leftover from #9315 : I forgot to modify `ImpersonationConfig`.

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
+++ b/api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
@@ -205,8 +205,7 @@ abstract class OAuth2ClientConfig implements OAuth2AuthenticatorConfig {
   @Value.Lazy
   Optional<HttpAuthentication> getBasicAuthentication() {
     return getClientSecretSupplier()
-        .map(Supplier::get)
-        .map(s -> BasicAuthenticationProvider.create(getClientId(), s));
+        .map(secretSupplier -> BasicAuthenticationProvider.create(getClientId(), secretSupplier));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
```diff
@@ -205,8 +205,7 @@ URI getResolvedImpersonationTokenEndpoint() {
   @Value.Lazy
   Optional<HttpAuthentication> getBasicAuthentication() {
     return getClientSecretSupplier()
-        .map(Supplier::get)
-        .map(s -> BasicAuthenticationProvider.create(getClientId(), s));
+        .map(secretSupplier -> BasicAuthenticationProvider.create(getClientId(), secretSupplier));
```
adutra, line 208: This has an implication for the OAuth client: now each request to the token endpoint will recompute the header.

adutra: To avoid that, maybe we can keep both `getClientSecret()` and `getClientSecretProvider()` (without deprecating the first one). Then create the basic authentication using the first non-empty optional.

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2ClientConfig.java
dimas-b: Keeping `getClientSecret()` LGTM

adutra: It might have to wait until I'm back :-)

PR #9327, opened 2024-08-12
Move object-storage specific code from `IcebergConfigurer` into `ObjectIO` implementations

snazy:


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -339,17 +245,77 @@ IcebergTableConfig icebergConfigPerTable(
...
+
+          String normalizedWarehouseLocation = normalizeS3Scheme(warehouseLocation);
+
+          // Must use both 's3.signer.uri' and 's3.signer.endpoint', because Iceberg before 1.5.0
+          // does
+          // not handle full URIs passed via 's3.signer.endpoint'. This was changed via
+          // https://github.com/apache/iceberg/pull/8976/files#diff-1f7498b6989fffc169f7791292ed2ccb35b305f6a547fd832f6724057c8aca8bR213-R216,
+          // first released in Iceberg 1.5.0. It's unclear how other language implementations deal
+          // with this.
+          config.put(S3_SIGNER_URI, uriInfo.icebergBaseURI().toString());
```
dimas-b, line 280: Hmmm, changing S3 `config` inside a "predicate" looks pretty obscure to me... Especially considering that some `ObjectIO` implementations do not call it at all.

I'd also move the setting of `S3_REMOTE_SIGNING_ENABLED` here.

Also, the predicate checks `locations`, which we already know... why add a round trip through a `Predicate`?

dimas-b: Maybe add a separate `ObjectIO.configureRemoteSigning()` method?

catalog/files/api/src/main/java/org/projectnessie/catalog/files/api/ObjectIO.java
```diff
@@ -29,4 +32,22 @@ public interface ObjectIO {
   OutputStream writeObject(StorageUri uri) throws IOException;
 
   void deleteObjects(List<StorageUri> uris) throws IOException;
+
+  void icebergWarehouseConfig(
+      StorageUri warehouse,
+      BiConsumer<String, String> defaultConfig,
+      BiConsumer<String, String> configOverride);
+
+  void icebergTableConfig(
```
dimas-b, line 41: `ObjectIO` is kind of Nessie's internal access channel to storage... Why overload it with Iceberg FileIO stuff? How about adding a separate interface (e.g. `IcebergIOConfigurer`) for the new methods?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
snazy: All the URI stuff - that definitely doesn't belong to ObjectIO. Too complex to get it into ObjectIO

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
```diff
@@ -101,4 +113,100 @@ public void deleteObjects(List<StorageUri> uris) throws IOException {
...
+
+    if (forTable) {
+      clientSupplier
+          .generateUserDelegationSas(storageLocations, fileSystemOptions)
+          .ifPresent(sasToken -> config.accept(ADLS_SAS_TOKEN_PREFIX + storageAccount, sasToken));
+    }
+  }
+
+  void icebergConfigDefaults(BiConsumer<String, String> config) {
+    config.accept(FILE_IO_IMPL, "org.apache.iceberg.azure.adlsv2.ADLSFileIO");
```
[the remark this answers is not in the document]
snazy: We usually expose vended credentials, which are object storage dependent anyway.
BTW: That's not new to this change.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
dimas-b: I'm +1 to keep it out of `ObjectIO`. 

Still, WDYT about removing the `Predicate` parameter and instead doing `if(isS3(locations)) {...}` here (including the setting of `S3_REMOTE_SIGNING_ENABLED`)?

snazy: Not really possible - see the comment I've added

dimas-b (approved):

PR #9335, opened 2024-08-13
Allow tweaking `CommitMeta` via HTTP headers

### LINKED ISSUE #9136: [Feature]: Overwrite author with an HTTP header
### Description

Nessie now supports two types of requests: Nessie API v1/2 and Iceberg REST API.
The author of a write operation is inferred from the authentication type. If I remember correctly, with the Nessie APIs, it's possible to define metadata related to the user performing the write operation.

I'm wondering if it would be possible to support, for both APIs, the ability to define and override the message author by setting an HTTP Header (`X-Nessie-Author`?)

### Expected Use Cases

There are cases where there's a generic authentication for an organization, but multiple users within the organization can modify the data catalog.

By standardizing a header compatible with both Nessie API v1/2 and Iceberg REST API, it would become easier to propagate information about the user performing the modification operation in a way that is agnostic to the type of API implementation the system is using.

This solution should also be compatible with `pyiceberg`. At the init time we can specify the headers the library needs to send to the server.

_No response_

snazy:
Fixes #9136

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
+++ b/api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
@@ -19,7 +19,7 @@ import static java.util.Collections.singleton;
 import static org.projectnessie.client.NessieConfigConstants.CONF_CONNECT_TIMEOUT;
 import static org.projectnessie.client.NessieConfigConstants.CONF_ENABLE_API_COMPATIBILITY_CHECK;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_CLIENT_NAME;
-import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_COMMIT_AUTHOR;
+import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_COMMIT_AUTHORS;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_COMMIT_MESSAGE;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_COMMIT_SIGNED_OFF_BY;
 import static org.projectnessie.client.NessieConfigConstants.CONF_NESSIE_DISABLE_COMPRESSION;
@@ -414,7 +414,7 @@ public interface NessieClientBuilder {
       if (s != null) {
         withHttpHeader("Nessie-Commit-Message", s);
       }
-      s = configuration.apply(CONF_NESSIE_COMMIT_AUTHOR);
+      s = configuration.apply(CONF_NESSIE_COMMIT_AUTHORS);
       if (s != null) {
         withHttpHeader("Nessie-Commit-Authors", s);
       }
@@ -422,6 +422,8 @@ public interface NessieClientBuilder {
       if (s != null) {
         withHttpHeader("Nessie-Commit-SignedOffBy", s);
       }
+      // Note: applying arbitrary properties is not possible here with the single-key approach via
+      // `Function<String, String> configuration`.
 
       return this;
     }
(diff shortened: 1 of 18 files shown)


api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -632,6 +632,25 @@ public final class NessieConfigConstants {
...
+   * Override all commit messages with the configured value. The corresponding HTTP header is {@code
+   * Nessie-Commit-Message}.
+   */
+  @ConfigItem public static final String CONF_NESSIE_COMMIT_MESSAGE = "nessie.commit-meta.message";
+
+  /**
+   * Set/override the author of all commits. The corresponding HTTP header is {@code
+   * Nessie-Commit-Authors}.
+   */
+  @ConfigItem public static final String CONF_NESSIE_COMMIT_AUTHOR = "nessie.commit-meta.author";
```
dimas-b, line 645: nit: `authors` (with an s)?.. Config properties are not repeatable (unlike headers)

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestV2TreeResource.java
```diff
@@ -431,7 +438,57 @@ public MergeResponse mergeRefIntoBranch(String branch, Merge merge)
...
+                    break;
+                  case "nessie-commit-signedoffby":
+                    v.stream()
+                        .map(String::trim)
+                        .filter(s -> !s.isEmpty())
+                        .forEach(commitMeta::addAllSignedOffBy);
+                    break;
+                  default:
+                    if (lower.startsWith("nessie-commit-property-")) {
+                      String prop = lower.substring("nessie-commit-property-".length()).trim();
```
dimas-b, line 479: nit: constant?

dimas-b (dismissed):
Nice improvement :+1:

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestV2TreeResource.java
```diff
@@ -431,7 +439,59 @@ public MergeResponse mergeRefIntoBranch(String branch, Merge merge)
...
+                  case "nessie-commit-message":
+                    v.stream()
+                        .map(String::trim)
+                        .filter(s -> !s.isEmpty())
+                        .findFirst()
+                        .ifPresent(commitMeta::message);
+                    break;
+                  case "nessie-commit-authors":
+                    v.stream()
+                        .flatMap(s -> Arrays.stream(s.split(",")))
```
dimas-b, line 468: Sorry, if I caused confusion with my comma-separation comment :sweat_smile: I meant it for client-side config variables only. Headers already support multiple values, so it looks a bit odd to have multiple lists of comma-separated values here. I think we can rely on the standard HTTP way of processing multiple headers here, but split config strings only on the client side. WDYT?

api/client/src/main/java/org/projectnessie/client/NessieClientBuilder.java
```diff
@@ -398,6 +410,21 @@ public NessieClientBuilder fromConfig(Function<String, String> configuration) {
         withApiCompatibilityCheck(Boolean.parseBoolean(s));
       }
 
+      s = configuration.apply(CONF_NESSIE_COMMIT_MESSAGE);
+      if (s != null) {
+        withHttpHeader("Nessie-Commit-Message", s);
+      }
+      s = configuration.apply(CONF_NESSIE_COMMIT_AUTHORS);
+      if (s != null) {
+        withHttpHeader("Nessie-Commit-Authors", s);
```
dimas-b, line 419: to clarify: I meant splitting the config value on `,` here, and setting it as multiple headers.

servers/rest-services/src/main/java/org/projectnessie/services/rest/RestV2TreeResource.java
snazy: This is to have feature-parity when you configre that via iceberg

dimas-b: (sigh)

dimas-b (approved):

PR #9338, opened 2024-08-14
Explicitly prevent empty content keys

### LINKED ISSUE #2648: [CLI/Nessie Server] Prevent creating `ContentKey` from `empty or null` key in both Java API and Python API and investigate empty `ContentKey` life time
Related comment: https://github.com/projectnessie/nessie/pull/2636#pullrequestreview-803612610

As well investigate what would happen if we have an empty `ContentKey` or `ContentKey` with empty elements in both, Java and Python APIs

snazy:
Empty content keys don't make any sense, but are theoretically possible. Empty content keys _could_ cause weird issues, so disallowing them. Also, empty content with an empty key is not be "addressable" at all in any tool/client, so the chance that any content with an empty key exists in the wild is 0.

Fixes #2648

### DIFF
--- a/api/model/src/main/java/org/projectnessie/model/Operation.java
+++ b/api/model/src/main/java/org/projectnessie/model/Operation.java
@@ -156,4 +156,11 @@ public interface Operation {
       return ImmutableUnchanged.builder().key(key).build();
     }
   }
+
+  @Value.Check
+  default void check() {
+    if (getKey().getElementCount() == 0) {
+      throw new IllegalStateException("Content key must not be empty");
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


api/model/src/main/java/org/projectnessie/model/Operation.java
```diff
@@ -156,4 +156,11 @@ static Unchanged of(ContentKey key) {
       return ImmutableUnchanged.builder().key(key).build();
     }
   }
+
+  @Value.Check
+  default void check() {
+    if (getKey().getElementCount() == 0) {
+      throw new IllegalStateException("Content key must not be empty");
```
dimas-b, line 163: nit: add op type and content type to the message for debuggability

dimas-b (approved):

PR #9340, opened 2024-08-14
Disable Swagger-UI + eliminate "duplicate operation ID" warnings in Quarkus tests

snazy:
Disable the Swagger-UI in Quarkus, because the OpenAPI path "mismatches" between the interfaces in the `org.projectnessie.api.v1/2.http` packages and the implementations in the `org.projectnessie.services.rest` package. Those do either produce duplicate entries in the Swagger UI or, when only including the implementations via `mp.openapi.scan.include`, no entries. Also, the Swagger UI is rather a dev-mode thing and is only available on the Quarkus management port and therefore not publicly available. Also disables the Quarkus provided download of the OpenAPI yaml and replaced it with a download of the `openapi.yaml` from `:nessie-model`.


site/docs/develop/rest.md
```diff
@@ -1,9 +1,8 @@
 # Rest API
 
 Nessie's REST APIs are how all applications interact with Nessie. The APIs are specified 
-according to the openapi v3 standard and are available when running the server by going 
-to [localhost:19120/q/openapi](http://localhost:9000/q/openapi). You can also peruse the set of operations our APIs support 
-by going to [SwaggerHub](https://app.swaggerhub.com/apis/projectnessie/nessie).
+according to the openapi v3 standard and are available when running the server by at the path
+`/nessie-openapi/openapi.yaml` (for example via `curl http://127.0.0.1:19120//nessie-openapi/openapi.yaml`)
```
dimas-b, line 5: Are all resource exposed over HTTP?

snazy: It's the publicly available openapi.yaml?

dimas-b (approved):

PR #9341, opened 2024-08-14
Eliminate more Quarkus test warnings

snazy:
* quarkus-rest vs resteasy-classic (in `:nessie-quarkus:test`)
* split-package warnings
* remaining "non-indexed classes" warnings (big hammer: using a Quarkus extension)


servers/quarkus-server/src/main/java/org/projectnessie/server/configchecks/ConfigChecks.java
```diff
@@ -36,6 +37,11 @@ public class ConfigChecks {
   @Inject QuarkusJdbcConfig jdbcConfig;
 
   public void configCheck(@Observes StartupEvent event) {
+    if (System.getProperty("org.gradle.test.worker") != null) {
```
dimas-b, line 40: Isn't it preferable to reduce the log threshold for this class in tests?

snazy: yea

PR #8708, opened 2024-06-03
Wire `SecretsProvider`s up to Quarkus

snazy:

PR #8707, opened 2024-06-03
Add `SecretsProvider` implementations for AWS, GCP, Vault

snazy:

PR #9345, opened 2024-08-14
Reference secrets by name, do not inject

snazy:
Warehouses/object-stores require secrets (access-keys, tokens, etc). These secrets are currently injected into the object-store specific configuration holders, when the effective bucket configuration is calculated.

The configuration of warehouses/object-stores should be distinct from the definition of secrets. This requires to not have the secrets in a bucket configuration at any time, but instead _reference_ secrets by name. These names are then resolved to secrets.

This change replaces the `Secret` types in all bucket configuration types with a `String` as the _reference_ to the secret.


helm/nessie/templates/_helpers.tpl
```diff
@@ -273,44 +273,52 @@ Define environkent variables for catalog storage options.
 */}}
 {{- define "nessie.catalogStorageEnv" -}}
 {{ $global := .}}
-{{- include "nessie.secretToEnv" (list .Values.catalog.storage.s3.defaultOptions.accessKeySecret "awsAccessKeyId" "nessie.catalog.service.s3.default-options.access-key.name" . ) }}
-{{- include "nessie.secretToEnv" (list .Values.catalog.storage.s3.defaultOptions.accessKeySecret "awsSecretAccessKey" "nessie.catalog.service.s3.default-options.access-key.secret" . ) }}
+{{- include "nessie.secretToEnv" (list .Values.catalog.storage.s3.defaultOptions.accessKeySecret "awsAccessKeyId" "nessie-catalog-secrets.s3.default-options.access-key.name" "nessie.catalog.service.s3.default-options.access-key" "nessie-catalog-secrets.s3.default-options.access-key" . ) }}
+{{- include "nessie.secretToEnv" (list .Values.catalog.storage.s3.defaultOptions.accessKeySecret "awsSecretAccessKey" "nessie-catalog-secrets.s3.default-options.access-key.secret" "" "" . ) }}
```
dimas-b, line 277: This seems to produce incorrect k8s yaml... for example:
```
            - name: "nessie-catalog-secrets.s3.default-options.access-key.name"
              valueFrom:
                secretKeyRef:
                  name: "k8sSecABC"
                  key: "key111"
            - name: "nessie.catalog.service.s3.default-options.access-key"
              value: "nessie-catalog-secrets.s3.default-options.access-key"- name: "nessie-catalog-secrets.s3.default-options.access-key.secret"
              valueFrom:
                secretKeyRef:
                  name: "k8sSecABC"
                  key: "access111"
```
The second one (secret) seems to be broken syntax-wise.

Also, `nessie-catalog-secrets.s3.default-options.access-key` is not referenced from the config map.

Test values (diff):
```
diff --git a/helm/nessie/values.yaml b/helm/nessie/values.yaml
index 3e066c45a..ace6e7f7f 100644
--- a/helm/nessie/values.yaml
+++ b/helm/nessie/values.yaml
@@ -138,14 +138,14 @@ bigtable:
 catalog:
 
   # -- Whether to enable the REST catalog service.
-  enabled: false
+  enabled: true
 
   # -- Iceberg catalog settings.
   iceberg:
 
     # -- The default warehouse name. Required. This is just a symbolic name; it must refer to a
     # declared warehouse below.
-    defaultWarehouse: ~  # warehouse1
+    defaultWarehouse: warehouse1
 
     # -- Iceberg config defaults applicable to all clients and warehouses. Any properties that are
     # common to all iceberg clients should be included here. They will be passed to all clients on
@@ -166,7 +166,7 @@ catalog:
     # warehouse must be defined.
     warehouses:
       # -- Symbolic name of the warehouse. Required.
-    - name: ~  # warehouse1
+    - name: warehouse1
       # -- Location of the warehouse. Required. Used to determine the base location of a table.
       # Scheme must be either s3 (Amazon S3), gs (Google GCS) or abfs / abfss (Azure ADLS). Storage
       # properties for each location can be defined below.
@@ -249,7 +249,7 @@ catalog:
         # -- Settings only relevant when clientAuthenticationMode is ASSUME_ROLE.
         serverIam:
           # -- Whether to enable server assume-role functionality.
-          enabled: ~  # false
+          enabled: true  # false
           # -- The ARN of the role to assume for accessing S3 data. This parameter is required for
           # Amazon S3, but may not be required for other storage providers (e.g. Minio does not use it
           # at all).
@@ -271,11 +271,11 @@ catalog:
         # -- AWS credentials. Required when serverAuthenticationMode is STATIC.
         accessKeySecret:
           # -- The secret name to pull AWS credentials from.
-          name: ~
+          name: k8sSecABC
           # -- The secret key storing the AWS secret key id.
-          awsAccessKeyId: ~
+          awsAccessKeyId: key111
           # -- The secret key storing the AWS secret access key.
-          awsSecretAccessKey: ~
+          awsSecretAccessKey: access111
 
       # -- Per-bucket S3 settings. Override the general settings above.
       buckets: []
```
CLI: `helm template --debug --namespace nessie-ns .`

dimas-b (commented):
Java code LGTM... helm is tricky :thinking:

catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/AbstractMapBasedSecretsProvider.java
```diff
@@ -0,0 +1,59 @@
...
+      @Nonnull SecretType secretType,
+      // only used for type-safety at the call site
+      @SuppressWarnings("unused") @Nonnull Class<S> secretJavaType) {
+    Map<String, String> secretData = resolveSecret(name);
+    if (secretData == null) {
+      return Optional.empty();
+    }
+
+    @SuppressWarnings("unchecked")
+    S secret = (S) secretType.fromValueMap(secretData);
```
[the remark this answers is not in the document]
snazy: Why? Each SecretType has only one Java type.

helm/nessie/templates/_helpers.tpl
snazy: Hrm - that stuff won't become my friend.
Anyway, it should work now. Added a new `ci/catalog-secrets-values.yaml` and a new step to CI to run this `helm template` against each of those.

catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/AbstractMapBasedSecretsProvider.java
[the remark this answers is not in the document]
dimas-b: the second comment is totally optional, but the first one makes sense, IMHO, it utilizes the actual runtime type parameter for casting (as opposed to compiled validation at the call site).

helm/nessie/templates/_helpers.tpl
snazy: > should reference nessie-catalog-secrets.s3.default-options.access-key from the config map as nessie.catalog.service.s3.default-options.access-key=nessie-catalog-secrets.s3.default-options.access-key

it is there:
```yaml
            #
            # s3.default-options.access-key
            #
            - name: "nessie.catalog.service.s3.default-options.access-key"
              value: "nessie-catalog-secrets.s3.default-options.access-key"
            - name: "nessie-catalog-secrets.s3.default-options.access-key.name"
              valueFrom:
                secretKeyRef:
                  name: "defaultCreds"
                  key: "defaultAccessKeyId"
            - name: "nessie-catalog-secrets.s3.default-options.access-key.secret"
              valueFrom:
                secretKeyRef:
                  name: "defaultCreds"
                  key: "defaultSecretAccessKey"#
```

via `helm template --debug --namespace nessie-ns --values ci/catalog-secrets-values.yaml .`

snazy: Finally!!!
Got something into `ci.yml` that validates all this using "real" "secrets".

snazy:
Ping @dimas-b

catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/Secret.java
```diff
@@ -15,5 +15,9 @@
  */
 package org.projectnessie.catalog.secrets;
 
+import java.util.Map;
+
 /** Base interface for all secrets. */
-public interface Secret {}
+public interface Secret {
```
adutra, line 21: I am a bit worried by the current definition fo a secret as a collection of key value pairs. I think it would be simpler to think of it as a single key-value _entry_.

For example, `S3BucketOptions` now contain an `Optional<String> accessKey();` but nothing tells the user that the string must point to a secret containing not one, but two entries, one for the access key itself, another for the secret key. From a user perspective, this is imho difficult to get right and I expect users to struggle with this.

Also, thinking ahead of time of how this would map to the Kubernetes world, not all secrets are created with the shape expected by Nessie in this PR. E.g. if Nessie expects a `BasicCredentials` type, then the secret _must_ contain two entries named `name` and `secret`. This will become an operational burden since secrets may have been created by other applications (Helm charts, operators, ArgoCD etc.) with different shapes.

I would like to propose another way of referencing secrets: by URI. For example, `S3BucketOptions` could expose 2 methods: `Optional<URI> accessKey();` and `Optional<URI> secretAccessKey();`. The URI could have the following general form (to be refined):

```
nessie-secret://[secret-provider-id]/provider-specific-part
```

Examples:

```
nessie-secret://k8s/my-namespace/my-secret?key=secret-key
nessie-secret://hashicorp-vault/my-namespace/my-secret?key=secret-key
nessie-secret://file/mnt/config/passwd
nessie-secret://env/MY_PASSWORD
nessie-secret://quarkus-config/my.password
```

etc. Then, resolving a secret would boil down to fetching the provider by ID (must exist via ServiceLoader), then resolving the secret itself. 

Does that make sense?

snazy: The purpose of having a composite is key rotation, which requires that composites like access-key+secret-access-key or username+password are in sync to prevent race conditions when reading (and caching!) secrets. AFAIK you can create secrets that have multiple key-value pairs in helm.

adutra: Could actually use URNs in fact: `urn:nessie-secret:k8s:namespace:secret?key=secret-key` etc.

adutra: > you can create secrets that have multiple key-value pairs in helm

Indeed but that's not what all charts do. Bitnami charts for databases e.g. usually create secrets only for the passwords, not for the users.

snazy: Using `URI` to identify a secret is a good idea, so I've updated the branch.
Not sure about the query part or the URI though. But let's see how that would look once the follow-up #8707 is in. Maybe there's a solution for that - so users _could_ choose.

adutra: Nice! I will have a look.

Re: secrets rotation. Could that be left to the secret provider as an implementation detail? IOW, if a provider is given a set of secrets to resolve, it could first group them by location, to avoid fetching the same resource more than once. And by doing so, the issue with out-of-sync secret pairs would be automatically solved.

snazy: > > you can create secrets that have multiple key-value pairs in helm
> 
> Indeed but that's not what all charts do. Bitnami charts for databases e.g. usually create secrets only for the passwords, not for the users.

Well, the secrets-provider stuff here is (primarily) for object-store credentials. For database connections, we have to leverage what Quarkus, the connector implementation and the secrets-provider implementations have - it's not what we can control.

snazy: > Nice! I will have a look.
> 
> Re: secrets rotation. Could that be left to the secret provider as an implementation detail? IOW, if a provider is given a set of secrets to resolve, it could first group them by location, to avoid fetching the same resource more than once. And by doing so, the issue with out-of-sync secret pairs would be automatically solved.

I haven't updated the branch. What I meant with "_could_ work" is a change on top of the direct follow-up - maybe sth that's inside the caching secrets provider - that _could potentially_ work.

adutra: > Well, the secrets-provider stuff here is (primarily) for object-store credentials. For database connections, we have to leverage what Quarkus, the connector implementation and the secrets-provider implementations have - it's not what we can control.

Agreed. Was just noting. I don't know how object store secrets are usually created, but I think it's quite reasonable to assume that for AWS credentials, the access key and the secret key will be fetched from the same location. It's however less reasonable imho to assume the same for pairs of `(token, expires_at)` secrets. But let's see how it goes and if users complain or not.

snazy: Updated to use the proposed URNs (via `URI`) and added a `ResolvingSecretsProvider`. It's probably never going to happen that someone uses different external secrets providers, but possible that some secrets are retrieved via Quarkus (pod environment) and some via an external secrets provider.

catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/UnsafePlainTextSecretsProvider.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.secrets;
+
+import jakarta.annotation.Nonnull;
+import java.net.URI;
+import java.util.Map;
+
+public class UnsafePlainTextSecretsProvider extends AbstractMapBasedSecretsProvider {
+  private final Map<URI, Map<String, String>> unsafeSecrets;
```
dimas-b, line 23: Is the `Map` necessary with URNs? How about `urn:nessie-secret:plain:key1=value1,key2=value2`?

dimas-b (commented):
The URN idea SGTM, but I've a comment about the specific impl. (below)

helm/nessie/templates/_helpers.tpl
```diff
@@ -291,50 +291,65 @@ Define environkent variables for catalog storage options.
...
 {{- $secretKey := get $secret $key -}}
 {{- with $global -}}
 {{- if (and $secretName $secretKey) -}}
-- name: {{ $envVarName | quote }}
+{{ if $addRef -}}
+#
+# {{ $midfix }}
+#
+- name: {{ (printf "nessie.catalog.service.%s" $midfix) | quote }}
+  value: {{ (printf "urn:nessie-secret:quarkus:nessie-catalog-secrets.%s" $midfix) | quote }}
```
dimas-b, line 350: All non-secret config goes to the `ConfigMap`, but this still produces and env. var for the `nessie` pod:
```
          env:
            #
            # s3.default-options.access-key
            #
            - name: "nessie.catalog.service.s3.default-options.access-key"
              value: "urn:nessie-secret:quarkus:nessie-catalog-secrets.s3.default-options.access-key"
 ```
Why not keep it in the `ConfigMap`?

File: catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/UnsafePlainTextSecretsProvider.java (hunk 4 above)
Comment (snazy) [reply catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/UnsafePlainTextSecretsProvider.java:23]: Not sure what you're asking for. This class is rather just for testing purposes.
Also, there's by design no need to ask for multiple secrets.

File: catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/ResolvingSecretsProvider.java [hunk 6]
```diff
@@ -0,0 +1,73 @@
...
+    String provider = next.substring(idxNessieSecret + 1, idxProvider);
+    checkArgument(
+        !provider.isBlank(),
+        "Invalid secret URI, must be in the form 'urn:nessie-secret:<provider>:<secret-name>'");
+
+    next = next.substring(idxProvider + 1);
+    checkArgument(
+        !next.isBlank() && next.charAt(0) != ':',
+        "Invalid secret URI, must be in the form 'urn:nessie-secret:<provider>:<secret-name>'");
+    name = URI.create(next);
```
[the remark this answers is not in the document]
snazy: `SecretsProvider` _implementations_ don't know their name, and they don't need to know their name - that's the whole idea here.
`ResolvingSecretsProvider` is mandatory - and in #8708 hidden behind a short-time cache.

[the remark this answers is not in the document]
adutra: But shouldn't `ResolvingSecretsProvider` be the only implementation of `SecretsProvider`? Components like `SmallryeConfigSecretsProvider` and `UnsafePlainTextSecretsProvider` look more "low-level" and imho should only be accessed by `ResolvingSecretsProvider` and thus implement a different interface (probably an SPI since we would later accept more secret sources like Vault etc.). That second interface would likely take a String as input, not an URI.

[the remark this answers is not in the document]
snazy: Separated into two interfaces.

catalog/secrets/api/src/main/java/org/projectnessie/catalog/secrets/UnsafePlainTextSecretsProvider.java
dimas-b: I mean putting plain test secrets directly in the URN and avoiding an extra config/map layer.

dimas-b (approved):

PR #9368, opened 2024-08-17
Introduce `CASSANDRA2` version store type, deprecate `CASSANDRA` version store type

snazy:
The current `CASSANDRA` version store type uses a lot of columns in the `objs` table, which cause quite some overhead in storage. The new `CASSANDRA2` version store type works like `CASSANDRA`, but uses a single column for all object values and also serializes object-ids as binary values.


dimas-b:
Is this worth the trouble given that `CASSANDRA` is "experimental" still?

snazy:
> Is this worth the trouble given that `CASSANDRA` is "experimental" still?

It's more about removing all the obj-type specific (de)serializers

servers/quarkus-config/src/main/java/org/projectnessie/quarkus/config/VersionStoreConfig.java
```diff
@@ -30,7 +30,10 @@ enum VersionStoreType {
     ROCKSDB,
     DYNAMODB,
     MONGODB,
+    /** Cassandra variant using many distinct columns. */
     CASSANDRA,
```
dimas-b, line 34: `@Deprecated` (as other old store types)?

versioned/storage/cassandra2/src/main/java/org/projectnessie/versioned/storage/cassandra2/Cassandra2BackendFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.versioned.storage.cassandra2;
+
+import jakarta.annotation.Nonnull;
+import org.projectnessie.versioned.storage.common.persist.BackendFactory;
+
+public class Cassandra2BackendFactory implements BackendFactory<Cassandra2BackendConfig> {
+
+  public static final String NAME = "Cassandra";
```
dimas-b, line 23: `Cassandra2`?

dimas-b (approved):

PR #9371, opened 2024-08-18
Mitigate OOM during `:nessie-server-admin-tool:intTest`

### LINKED ISSUE #8571: Investigate resource leak in Server Admin Tool integration tests
#8548 introduces integration tests for various backends using `MultiEnvTestEngine`. Unfortunately there seems to be a resource leak, which is currently mitigated with `forkEvery = 4` Gradle setting. It could be a consequence of a `@QuarkusMainTest` being executed by `MultiEnvTestEngine`. If so, a simple fix could be to migrate to `QuarkusMainIntegrationTest` – but that would require changes to the tests.

snazy:
Looking into an `.hprof` file, there are a many suspicious `ScheduledThreadPoolExecutor`s active - each with one task in its queue - and multiple instances of these executors for the same task type (but different instances).
* `io.opentelemetry.sdk.metrics.export.PeriodicMetricReader.Scheduled`
* `com.google.api.gax.rpc.Watchdog`
* (lambda to) `com.mongodb.internal.connection.PowerOfTwoBufferPool#prune`

Since those are active/running, they keep references to the class loader, in turn "everything".

Fixes #8571

### DIFF
--- a/tools/server-admin/src/main/resources/application.properties
+++ b/tools/server-admin/src/main/resources/application.properties
@@ -161,4 +161,18 @@ quarkus.arc.ignored-split-packages=\
 
 quarkus.http.test-port=0
 
+# Disable a "lot of things" that are not needed for integration testing. Disabled in an effort to get hold
+# of the class-loader OOM leak described in https://github.com/projectnessie/nessie/issues/8571
+# Many of these settings are probably unnecessary, but otoh don't hurt.
 %test.quarkus.devservices.enabled=false
+%test.quarkus.micrometer.enabled=false
+%test.quarkus.micrometer.registry-enabled-default=false
+%test.quarkus.mongodb.metrics.enabled=false
+%test.quarkus.otel.enabled=false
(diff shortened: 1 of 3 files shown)


snazy:
Side effect: duration of `:nessie-server-admin-tool:intTest` is nearly cut in half

tools/server-admin/src/main/resources/application.properties
```diff
@@ -161,4 +161,18 @@ quarkus.arc.ignored-split-packages=\
 
 quarkus.http.test-port=0
 
+# Disable a "lot of things" that are not needed for integration testing. Disabled in an effort to get hold
```
XN137, line 164: mentions integration testing however the file seems to be a non-test resource?

XN137 (approved):

tools/server-admin/src/main/resources/application.properties
snazy: `%test` profile

PR #9382, opened 2024-08-20
Require Java 21 for Nessie build

snazy:
Also bump Quarkus requirements to Java 21

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -907,6 +907,7 @@ jobs:
           java-version: | 
             11
             17
+            21
 
       - name: Setup Gradle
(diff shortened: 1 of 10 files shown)


XN137 (approved):
so wondering:
which part of the build still compiles for older java versions?
all the artifacts used directly by iceberg?
do we cover this in CI or will `query-engine-integration-tests` tell us in case we broke something?

changes themselves lgtm

snazy:
> so wondering: which part of the build still compiles for older java versions?

Everything that's client-facing or is consumed downstream (#9253 makes it probably clearer)

> all the artifacts used directly by iceberg? 

only client-facing ones

> do we cover this in CI or will `query-engine-integration-tests` tell us in case we broke something?

both

PR #9388, opened 2024-08-21
Migrate to new `shadow` plugin ID

### LINKED ISSUE #7912: Build: Consider replacing shadow plugin
### Issue description

The shadow-plugin (ID `com.github.johnrengelman.shadow`) is currently rather unmaintained and already has a bunch of issues, especially with newer Gradle versions (8.2+) and when the included classes are built for Java 21 or newer (class file version 65+).

There is a fork of the shadow plugin (ID `io.github.goooler.shadow`) that works against newer Gradle versions and Java 21. That one however is not recognized by other plugins like the jmh plugin (ID `me.champeau.jmh`), which only check against the "original" plugin ID.

snazy:
See [Shadow release notes](https://github.com/GradleUp/shadow/releases/tag/8.3.0), the new plugin ID and Maven coordinates. Requires a bump of the `me.champeau.jmh` as well.

Fixes #7912

### DIFF
--- a/build-logic/src/main/kotlin/nessie-shadow-jar.gradle.kts
+++ b/build-logic/src/main/kotlin/nessie-shadow-jar.gradle.kts
@@ -16,7 +16,7 @@
 
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 
-plugins { id("com.github.johnrengelman.shadow") }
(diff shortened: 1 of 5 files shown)


snazy:
Depends on a new release of https://github.com/melix/jmh-gradle-plugin

PR #9399, opened 2024-08-23
Minor cleanup in `:nessie-versioned-storage-store` + commit-logic-impl

snazy:
* Unify two 2x duplicated code fragments
* typo
* unused import

### DIFF
--- a/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
+++ b/versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
@@ -387,6 +387,17 @@ class BaseCommitHelper {
     }
   }
 
+  static ObjId idForExpectedContent(StoreKey key, StoreIndex<CommitOp> headIndex) {
+    StoreIndexElement<CommitOp> expected = headIndex.get(key);
+    if (expected != null) {
+      CommitOp expectedContent = expected.content();
+      if (expectedContent.action().exists()) {
+        return expectedContent.value();
+      }
+    }
+    return null;
+  }
+
   void verifyMergeTransplantCommitPolicies(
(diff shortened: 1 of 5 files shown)


versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
```diff
@@ -387,6 +387,17 @@ private void validateNamespacesExistForContentKeys(
     }
   }
 
+  static ObjId idForExpectedContent(StoreKey key, StoreIndex<CommitOp> headIndex) {
```
XN137, line 390: seems like all callers use `el.key()` for the 1st param, maybe this should happen inside the method?

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
```diff
@@ -387,6 +387,17 @@ private void validateNamespacesExistForContentKeys(
...
+    StoreIndexElement<CommitOp> expected = headIndex.get(key);
+    if (expected != null) {
+      CommitOp expectedContent = expected.content();
+      if (expectedContent.action().exists()) {
+        return expectedContent.value();
+      }
+    }
+    return null;
```
XN137, line 398: do we use `@Nullable` on methods in the codebase? or return `Optional` ?

XN137 (approved):

versioned/storage/store/src/main/java/org/projectnessie/versioned/storage/versionstore/BaseCommitHelper.java
snazy: both :D

PR #9400, opened 2024-08-23
GC: always look into all keys of the last commit

snazy:
There is a chance that GC might not recognize all live contents, if the commit history is shorter than the cutoff policy _and_(!) commits are not accessible (database corruption).

No realistical functional or behavioral change, rather a precaution.

### DIFF
--- a/gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
+++ b/gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
@@ -15,6 +15,8 @@
  */
 package org.projectnessie.gc.identify;
 
+import static java.lang.String.format;
+
 import com.google.common.base.Preconditions;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import jakarta.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
```diff
@@ -241,6 +243,7 @@ private ReferencesWalkResult identifyContentsForReference(
 
       LogEntryHolder holder = new LogEntryHolder();
       String lastCommitId = null;
+      String finalCommitId = null;
```
XN137, line 246: nit: "last" and "final" are confusing prefixes when used together imo

XN137 (approved):
> There is a chance that GC might not recognize all live contents, if the commit history is shorter than the cutoff policy and(!) commits are not accessible (database corruption).

is the scenario hard to reproduce in a test?

generally change lgtm but maybe someone with more knowledge about our gc impl should review as well

gc/gc-base/src/main/java/org/projectnessie/gc/identify/IdentifyLiveContents.java
dimas-b: I tend to agree... How about `firstNonLiveCommitId` (to match the log message)?

dimas-b (approved):

PR #9439, opened 2024-08-29
Choose the Nessie client by name, case-insensitive

snazy:


dimas-b (approved):
I assume we assume client names are ASCII? :)

snazy:
> I assume we assume client names are ASCII? :)

Unless you create one with emojis and put those in environment variables :P

PR #9445, opened 2024-08-29
Use non-blocking random

snazy:


sudheeshkatkam (approved):

adutra (approved):

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/SignerKeysServiceImpl.java
```diff
@@ -92,11 +91,10 @@ private SignerKeysObj addNewKey(SignerKeysObj current) {
     Instant expires = now.plus(NEW_KEY_EXPIRE_AFTER);
 
     byte[] secretBytes = new byte[32]; // 256 bits for SHA256
-    try {
-      SecureRandom.getInstanceStrong().nextBytes(secretBytes);
-    } catch (NoSuchAlgorithmException e) {
-      throw new RuntimeException(e);
-    }
+    // Use `/dev/urandom`, which does not block.
```
dimas-b, line 94: `/dev/urandom` does not block, true, but it generates weaker randomness in cases when `/dev/random` would block. Since it is used for generating a secret I'm not sure the perf. trade-off is justified.

How often do we re-generate keys? Is blocking actually critical here?

snazy: That assumption _was_ a little bit true - but is no longer true.
Fact is, the man pages for (u)random are rather misleading.
There's a [nice article](https://www.2uo.de/myths-about-urandom/) about this.

TL;DR you just need 256 bits of "proper" entropy - and that only once - very early when the machine boots.
Some still say that /dev/random might be needed for very long lived GPG/SSL/SSH keys.

It's easier to steal a signed URL or start Nessie with a debugger or so.

dimas-b (approved):

PR #9447, opened 2024-08-29
Switch to new s3-sign endpoint using an opaque path parameter

### LINKED ISSUE #9444: [Catalog] pass s3-sign parameters via some blob
Replace the s3-sign endpoint query parameters with a path-parameter that is an "opaque blob" for clients, containing the current query parameters.

snazy:
This adds a little bit more obscurity, but also makes S3 signing compatible to some query engines that do not support query parameters for the s3-signing endpoint.

Fixes #9444

### DIFF
--- a/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
+++ b/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
@@ -99,31 +99,9 @@ public interface ExternalBaseUri {
    * deal with this.
    *
    * @param prefix the prefix of the request (warehouse and reference)
-   * @param contentKeyPathString the content key to sign the request for
    * @param signerParam the signer params path parameter (URL safe representation)
    */
   default String icebergS3SignerPathWithPath(String prefix, String signerParam) {
-    return format("v1/%s/s3-sign2/%s", encode(prefix, UTF_8), signerParam);
-  }
-
-  /**
-   * The path, without a leading slash, of the URI to sign a request to S3 for a specific content
-   * key.
-   *
-   * <p>Must use both {@code s3.signer.uri} and {@code s3.signer.endpoint}, because Iceberg before
-   * 1.5.0 does not handle full URIs passed via {@code s3.signer.endpoint}. This was changed via <a
-   * href="https://github.com/apache/iceberg/pull/8976/files#diff-1f7498b6989fffc169f7791292ed2ccb35b305f6a547fd832f6724057c8aca8bR213-R216">this
-   * Iceberg PR</a> first released in Iceberg 1.5.0. It's unclear how other language implementations
-   * deal with this.
-   *
-   * @param prefix the prefix of the request (warehouse and reference)
-   * @param contentKeyPathString the content key to sign the request for
-   * @param uriQuery the query string for the URL
-   */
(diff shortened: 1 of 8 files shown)


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
```diff
@@ -88,6 +88,24 @@ default void icebergConfigOverrides(BiConsumer<String, String> config) {
...
+   * href="https://github.com/apache/iceberg/pull/8976/files#diff-1f7498b6989fffc169f7791292ed2ccb35b305f6a547fd832f6724057c8aca8bR213-R216">this
+   * Iceberg PR</a> first released in Iceberg 1.5.0. It's unclear how other language implementations
+   * deal with this.
+   *
+   * @param prefix the prefix of the request (warehouse and reference)
+   * @param contentKeyPathString the content key to sign the request for
+   * @param signerParam the signer params path parameter (URL safe representation)
+   */
+  default String icebergS3SignerPathWithPath(String prefix, String signerParam) {
+    return format("v1/%s/s3-sign2/%s", encode(prefix, UTF_8), signerParam);
```
dimas-b, line 106: Why do we encode `prefix` but not `signerParam`? It looks odd that encoding assumptions are different for these two input parameters.

IMHO, a "URL-safe representation" is not crystal clear. E.g. when constructing ` java.net.URI` from components, it will encode special chars in the provided path string, and on parsing that path element back it will decode those change and we get the original string. Since we assume `signedParams` in `IcebergApiV1S3SignResource` is URL-decoded (by the Web container), I tend to think we have to apply the encoding here, even  though it will probably not change the string in runtime.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
```diff
@@ -102,7 +120,8 @@ default void icebergConfigOverrides(BiConsumer<String, String> config) {
    * @param contentKeyPathString the content key to sign the request for
    * @param uriQuery the query string for the URL
    */
-  default String icebergS3SignerPath(String prefix, String contentKeyPathString, String uriQuery) {
+  default String icebergS3SignerPathWithQuery(
```
dimas-b, line 123: This method appears to be unused now.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/ExternalBaseUri.java
snazy: Base64 using URL-alphabet is safe

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergApiV1S3SignResource.java
```diff
@@ -68,6 +68,40 @@ public Response mapException(Exception ex) {
     return errorMapper.toResponse(ex, IcebergEntityKind.UNKNOWN);
   }
 
+  @Operation(operationId = "iceberg.v1.s3sign.blob")
+  @POST
+  @Path("/v1/{prefix}/s3-sign2/{signedParams}")
```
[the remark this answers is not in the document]
snazy: done

dimas-b (approved):

PR #9448, opened 2024-08-29
Helm chart: strengthen default security context

adutra:
This PR adds sane defaults for pod and container security contexts.

Since the default filesystem is now read-only, it also adds a new mount point for /tmp with read-write access.

It also adds two more options: `extraVolumes` and `extraVolumeMounts`. These can serve as escape hatches in case users need more locations with read-write access.

### DIFF
--- a/helm/nessie/templates/NOTES.txt
+++ b/helm/nessie/templates/NOTES.txt
@@ -33,6 +33,6 @@ To connect to Nessie, please execute the following commands:
   export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ include "nessie.fullname" . }} -o jsonpath="{ .status.loadBalancer.ingress[0].ip }")
   echo http://$SERVICE_IP:{{ index .Values.service.ports "nessie-server" }}
 {{- else if contains "ClusterIP" .Values.service.type }}
-  nohup kubectl --namespace {{ .Release.Namespace }} port-forward svc/nessie 19120:{{ index .Values.service.ports "nessie-server" }} &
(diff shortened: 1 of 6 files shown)


helm/nessie/templates/NOTES.txt
```diff
@@ -15,6 +15,6 @@ To connect to Nessie, please execute the following commands:
   export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ include "nessie.fullname" . }} -o jsonpath="{ .status.loadBalancer.ingress[0].ip }")
   echo http://$SERVICE_IP:{{ index .Values.service.ports "nessie-server" }}
 {{- else if contains "ClusterIP" .Values.service.type }}
-  nohup kubectl --namespace {{ .Release.Namespace }} port-forward svc/nessie 19120:{{ index .Values.service.ports "nessie-server" }} &
+  nohup kubectl --namespace {{ .Release.Namespace }} port-forward svc/{{ include "nessie.fullname" . }} 19120:{{ index .Values.service.ports "nessie-server" }} &
```
adutra, line 18: Minor, unrelated fix.

dimas-b (dismissed):
GH reports `no new line character at the end of file` in `values.yaml`... otherwise LGTM :+1:

adutra:
Badk to draft, it probably needs to be rebased on top of #9456

adutra:
Ready for  review, but the Helm tests will fail until there is an official image with the new UID and GID.

adutra:
> Ready for review, but the Helm tests will fail until there is an official image with the new UID and GID.

Scratch that, I forgot that CI builds its own images for Helm testing.

dimas-b (approved):

PR #9449, opened 2024-08-29
CLI: make tool runnable without history

adutra:
Setting `--history=false` wasn't working correctly.


adutra:
BTW I thought we could just specify `--no-history` but this doesn't seem to work, we need to specify `--history=false` or `--no-history=true`. Probably a bug in Picocli.

dimas-b (approved):

PR #9453, opened 2024-08-30
"Thread per test class" via `MultiEnvTestEngine`

### LINKED ISSUE #9441: Fix class/resource leak in `:nessie-quarkus:test`
Added a commit to #9433 to increase the max test-worker heap size to 3g. This makes the `:nessie-quarkus:test` task finish.

There is a resource leak, but that's already present with earlier Quarkus versions, so I suspect something that's present in all our tests is causing the leak, #9440 is one cause, but not the only one.

This issue is to investigate, find the root cause, fix it and bring the max test-worker heap size back to 2g.

snazy:
See #9441 for a complete description of the underlying problem. TL;DR is: `ThreadLocal`s from various 3rd party libraries leak into the single `Test worker` thread that runs all the tests, resulting in TL objects/suppliers from the various Quarkus test class loaders, eventually leading to nasty OOMs.

This change updates the `MultiEnvTestEngine` by using the new `ThreadPerTestClassExecutionExecutorService` and also "assimilate" really all tests, even the non-multi-env tests, so that those also run on a thread per test-class. The logic to distinguish multi-env from non-multi-env tests via `MultiEnvExtensionRegistry.registerExtension()` via test discovery is not perfect (but good enough), it can add multi-env tests to the non-multi-env tests, so an additional check is needed there.

Since each test class runs on "its own thread", the `ThreadLocal`s are registered on that thread. Once the test class finishes, the thread is disposed and its thread locals become eligible for garbage collection, which is what is needed.

The bump of the max-heap size for test workers is also reduced back to 2g (was changed in #9433).

Fixes #9441

### DIFF
--- a/testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
+++ b/testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
@@ -26,6 +26,7 @@ import org.junit.jupiter.engine.config.JupiterConfiguration;
 import org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor;
 import org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor;
 import org.junit.jupiter.engine.discovery.DiscoverySelectorResolver;
+import org.junit.platform.engine.ConfigurationParameters;
 import org.junit.platform.engine.EngineDiscoveryRequest;
 import org.junit.platform.engine.ExecutionRequest;
 import org.junit.platform.engine.TestDescriptor;
@@ -51,7 +52,8 @@ public class MultiEnvTestEngine implements TestEngine {
 
   private static final MultiEnvExtensionRegistry registry = new MultiEnvExtensionRegistry();
 
-  private final JupiterTestEngine delegate = new JupiterTestEngine();
+  private final ThreadPerTestClassExecutionTestEngine delegate =
+      new ThreadPerTestClassExecutionTestEngine();
 
   static MultiEnvExtensionRegistry registry() {
     return registry;
@@ -76,36 +78,36 @@ public class MultiEnvTestEngine implements TestEngine {
       preliminaryResult.accept(
           descriptor -> {
             if (descriptor instanceof ClassBasedTestDescriptor) {
-              Class<?> testClass = ((ClassBasedTestDescriptor) descriptor).getTestClass();
-              registry.registerExtensions(testClass);
+              registry.registerExtensions(((ClassBasedTestDescriptor) descriptor));
             }
           });
 
+      ConfigurationParameters configurationParameters =
+          discoveryRequest.getConfigurationParameters();
+
       // JupiterEngineDescriptor must be the root, that's what the JUnit Jupiter engine
       // implementation expects.
-      JupiterEngineDescriptor multiEnvDescriptor =
+      JupiterEngineDescriptor multiEnvRootDescriptor =
           new JupiterEngineDescriptor(
-              uniqueId,
-              new DefaultJupiterConfiguration(discoveryRequest.getConfigurationParameters()));
+              uniqueId, new DefaultJupiterConfiguration(configurationParameters));
 
+      // Handle the "multi-env" tests.
       List<String> extensions = new ArrayList<>();
-      AtomicBoolean envDiscovered = new AtomicBoolean();
+      AtomicBoolean multiEnvDiscovered = new AtomicBoolean();
       registry.stream()
           .forEach(
               ext -> {
                 extensions.add(ext.getClass().getSimpleName());
-                for (String envId :
-                    ext.allEnvironmentIds(discoveryRequest.getConfigurationParameters())) {
-                  envDiscovered.set(true);
+                for (String envId : ext.allEnvironmentIds(configurationParameters)) {
+                  multiEnvDiscovered.set(true);
                   UniqueId segment = uniqueId.append(ext.segmentType(), envId);
 
                   MultiEnvTestDescriptor envRoot = new MultiEnvTestDescriptor(segment, envId);
-                  multiEnvDescriptor.addChild(envRoot);
+                  multiEnvRootDescriptor.addChild(envRoot);
 
                   JupiterConfiguration envRootConfiguration =
                       new CachingJupiterConfiguration(
-                          new MultiEnvJupiterConfiguration(
-                              discoveryRequest.getConfigurationParameters(), envId));
+                          new MultiEnvJupiterConfiguration(configurationParameters, envId));
                   JupiterEngineDescriptor discoverResult =
                       new JupiterEngineDescriptor(segment, envRootConfiguration);
                   new DiscoverySelectorResolver()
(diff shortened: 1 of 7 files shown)


testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
```diff
@@ -116,18 +118,53 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
...
+                    ClassBasedTestDescriptor classBased = (ClassBasedTestDescriptor) child;
+                    boolean multi = registry().isMultiEnvClass(classBased);
+                    if (multi) {
+                      envRoot.addChild(child);
+                    }
+                  }
+                }
+              });
+
+      // Also execute all other tests via the MultiEnv test engine to get the "thread per
```
adutra, line 133: That's actually probably better anyways, since the filtering done in `MultiEnvTestFilter` becomes straightforward and we know that _all_ tests are going to be executed by the multi-env engine. I like it 👍

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.jupiter.engine.support.JupiterThrowableCollectorFactory;
+import org.junit.platform.engine.EngineDiscoveryRequest;
+import org.junit.platform.engine.ExecutionRequest;
+import org.junit.platform.engine.TestDescriptor;
+import org.junit.platform.engine.UniqueId;
+import org.junit.platform.engine.support.hierarchical.HierarchicalTestEngine;
+import org.junit.platform.engine.support.hierarchical.HierarchicalTestExecutorService;
+import org.junit.platform.engine.support.hierarchical.ThrowableCollector;
+
+public class ThreadPerTestClassExecutionTestEngine
```
adutra, line 34: One minor drawback here is that this new engine is only used as a delegate to `MultiEnvTestEngine`, never as a "top-level" engine. IOW, only modules that include `:nessie-multi-env-test-engine` will benefit from the improvement.

Granted, most "problematic" modules have a dependency on the multi-env engine anyways. But i was thinking of the kubernetes operator PR: this module does not have multi-env tests, so it wouldn't benefit from these improvements as it is today. But maybe it's just a matter of adding a dependency to `:nessie-multi-env-test-engine` there.

adutra (commented):
LGTM but I think we should wait for @dimas-b review as well.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
snazy: Or use JUnit, if https://github.com/junit-team/junit5/issues/3939 gets merged and released.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
```diff
@@ -0,0 +1,85 @@
...
+  /** Returns {@code org.junit.jupiter} as the group ID. */
+  @Override
+  public Optional<String> getGroupId() {
+    return Optional.of("org.junit.jupiter");
+  }
+
+  /** Returns {@code junit-jupiter-engine} as the artifact ID. */
+  @Override
+  public Optional<String> getArtifactId() {
+    return Optional.of("junit-jupiter-engine");
```
dimas-b, line 51: I'm not sure this is correct. Do we _have to_ use this ID to make stuff work?

testing/multi-env-test-engine/src/test/java/org/projectnessie/tools/compatibility/engine/TestMultiEnvTestEngine.java
```diff
@@ -46,8 +46,7 @@ void junitJupiter() {
                 .list()
                 .stream()
                 .map(e -> e.getTestDescriptor().getUniqueId()))
-        .isNotEmpty()
-        .allSatisfy(id -> assertThat(id.getEngineId()).hasValue(JupiterEngineDescriptor.ENGINE_ID));
```
dimas-b, line 50: I think this test is not meaningful as a whole without this assertion. The handling of `PlainTest` is covered by the test below.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionExecutorService.java
```diff
@@ -0,0 +1,117 @@
...
+      Thread threadPerClass =
+          new Thread(
+              () -> {
+                try {
+                  testTask.execute();
+                } catch (Exception e) {
+                  failure.set(e);
+                }
+              },
+              "TEST THREAD FOR " + lastSegment.getValue());
```
dimas-b, line 93: I think it's preferable to use the whole test's `uniqueId` for traceability. Last segments may have the same value for multiple tests under the multi-env engine.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestFilter.java
```diff
@@ -46,39 +44,27 @@ private Optional<Class<?>> classFor(TestDescriptor object) {
...
   }
 
-  private FilterResult filter(Class<?> testClass, UniqueId id) {
-    // Use the static extension data collected during the discovery phase.
-    // It is possible to reload extensions based of class objects from test descriptors,
-    // however that would add unnecessary overhead.
-    MultiEnvExtensionRegistry registry = registry();
-
+  /**
+   * This filter effectively routes all tests via the {@link MultiEnvTestEngine}, both actual
```
dimas-b, line 48: In that case could we simply disable the default "jupiter" engine somehow?

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
dimas-b: I'm not sure this is ideal for downstream projects. Not all tests may want to get the "thread per class" model by default.

Could we add an annotation to control this behaviour?

snazy: I don't see any reason to add more complexity here.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestFilter.java
snazy: No

testing/multi-env-test-engine/src/test/java/org/projectnessie/tools/compatibility/engine/TestMultiEnvTestEngine.java
snazy: There's a `.isEmpty()` below

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionExecutorService.java
dimas-b: @snazy WDYT?

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
dimas-b: I mean: would a Nessie artifact ID work here?

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionExecutorService.java
snazy: I'd prefer to defer that to the JUnit-PR. Also not sure how that would help in real life.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
```diff
@@ -116,18 +118,53 @@ public TestDescriptor discover(EngineDiscoveryRequest discoveryRequest, UniqueId
...
+      // test-class" behavior.
+      registry()
+          .probablyNotMultiEnv()
```
dimas-b, line 136: Why collect test descriptors in an "extension" registry? Why not reuse `preliminaryResult` here and filter out all multi-env tests?

We already rediscover tests for each env. ID, so we're probably not saving much time by collecting test descriptors during the extension registration process... On the other hand using a discovery result directly would be much more clear... WDYT?

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
snazy: The ID doesn't really matter. And with the Junit-PR this would go away entirely.

dimas-b: what's the Junit PR?

snazy: https://github.com/projectnessie/nessie/pull/9453#discussion_r1739658367

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/MultiEnvTestEngine.java
snazy: The result is sadly not correct that way. We can make it "perfect" in a follow-up.

testing/multi-env-test-engine/src/main/java/org/projectnessie/junit/engine/ThreadPerTestClassExecutionTestEngine.java
dimas-b: It's an issue ;)

adutra: https://github.com/junit-team/junit5/pull/3941

dimas-b (approved):

PR #9456, opened 2024-08-30
Switch Nessie Docker images to use UID 10000

adutra:
Many companies have strict requirements on container security and a common convention is that containers should be run by non-root, non-system users with UID >= 1000.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ as necessary. Empty sections will not end in the release notes.
 
 - Support for Java 8 has been removed, even for Nessie clients. Minimum runtime requirement for clients
   is Java 11.
+- Nessie Docker images now all execute as user `nessie` (UID 10000 and GID 10001). They would
+  previously execute as user `default` (UID 185 and GID 0). This is a security improvement, as the
+  Nessie images no longer run with a UID within the privileged range, and the GID is no longer 0
+  (root). If you have any custom configurations, especially Kubernetes manifests containing security
+  contexts, that rely on the previous user `default` (UID 185 and GID 0), you will need to adjust
+  them to reference the new user `nessie` (UID 10000 and GID 10001) from now on.
 
 ### Breaking changes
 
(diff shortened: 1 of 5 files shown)


snazy:
Not really a fan of 1001 actually. I can see conflicts between the UID/GID of the container and the (same) UID/GID on the host. Why not randomly picked very high IDs?

adutra:
Don't we already have the same problem with user 185?

I think UID 1001 is so commonly used in the k8s that I think it's safer to go with that than any other number.

snazy:
`commonly used` is the culprit. 1001 would clash with those and existing UIDs/GIDs on the host ("normal users").

There is some [chatter](https://github.com/hexops-graveyard/dockerfile?tab=readme-ov-file#do-not-use-a-uid-below-10000) that one shouldn't use UID/GID below 10000 ([more](https://news.ycombinator.com/item?id=25621610)) to avoid some privilege escalation from a container.

adutra:
I get the cons, but I think they only apply to users running the container in their laptops. No cloud provider in their right mind would ever grant any privilege to user 1001, because they would be basically granting privileges on their machines to half of the containers in the world :-)

But OK, let's change to 10000. Is that OK with you?

dimas-b (approved):
UID `10000` LGTM... However, I wonder if this is becoming more like a well-known port situation... If many containers use `10000` (good round number) for their UIDs, would it make sense for Nessie to use something like `29120` (related to the 19120 default port ;) )?

adutra:
> UID `10000` LGTM... However, I wonder if this is becoming more like a well-known port situation... If many containers use `10000` (good round number) for their UIDs, would it make sense for Nessie to use something like `29120` (related to the 19120 default port ;) )?

I had the same idea, but not sure what's best honestly. It's also becoming a quest for the perfect UID which I think is a never-ending one... let's go with 10000, imho that's more than good enough!

dimas-b:
>  let's go with 10000, imho that's more than good enough!

SGTM :+1:

PR #9462, opened 2024-09-03
Add ability to change pathType of ingress

iglooom:
My hoster interprets ImplementationSpecific as Exact instead of Prefix, so I need to change it to work properly

### DIFF
--- a/helm/nessie/templates/ingress.yaml
+++ b/helm/nessie/templates/ingress.yaml
@@ -39,7 +39,7 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ . }}
-            pathType: ImplementationSpecific
+            pathType: {{ default "ImplementationSpecific" $.Values.ingress.pathType }}
             backend:
               {{- if semverCompare ">=1.22-0" $kubeVersion }}
               service:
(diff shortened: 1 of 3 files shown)


helm/nessie/templates/ingress.yaml
```diff
@@ -39,7 +39,7 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ . }}
-            pathType: ImplementationSpecific
+            pathType: {{ default "ImplementationSpecific" $.Values.ingress.pathType }}
```
adutra, line 42: Could you add the new option `ingress.pathType` to values.yaml and document it please? Otherwise it looks like a "hidden feature" 🙂

adutra (commented):
Thanks for the PR @iglooom, I'm sure it will be useful to many users!

helm/nessie/templates/ingress.yaml
iglooom: You're right, done this.
Actually, I'd prefer to Prefix by default, but for backward compatibility, I left it as it was.

adutra (approved):
Thanks @iglooom !

dimas-b (approved):
LGTM. Thanks for the contribution @iglooom !

PR #9466, opened 2024-09-04
Helm chart: add license headers and LICENSE file

adutra:


dimas-b (approved):

snazy:
LGTM - just wonder why we need `##` (double hash mark)?

adutra:
> LGTM - just wonder why we need `##` (double hash mark)?

You need some char sequence that only appears in the header, otherwise the `delimiter` regex would risk matching portions of the header instead of matching the beginning of the actual file. With YAML, it's hard because the syntax is quite flexible.

snazy:
> > LGTM - just wonder why we need `##` (double hash mark)?
> 
> You need some char sequence that only appears in the header, otherwise the `delimiter` regex would risk matching portions of the header instead of matching the beginning of the actual file. With YAML, it's hard because the syntax is quite flexible.

Ah, right - the delimiter gift that keeps on giving

snazy (approved):

PR #9511, opened 2024-09-11
Catalog/Iceberg: Add support for `remove-partition-specs` update + new `endpoints` field in config response

snazy:


snazy:
Closing in favor of #9906

PR #9517, opened 2024-09-12
[Bug]: Helm Chart does not render in version 0.96 because of secretToEnv

Milias:
Hello, thanks for working on this project!

We found today that after release `0.96.0` the helm chart does not deploy with the following error:

```
error deploying nessie: unable to deploy helm chart: 
error executing 'helm upgrade nessie --values /tmp/3296770843 --install /nessie/chart --kube-context pod-context': 
Error: UPGRADE FAILED: 
template: nessie/charts/nessie/templates/deployment.yaml:100:16: executing "nessie/charts/nessie/templates/deployment.yaml" at <include "nessie.secretToEnv" (list $secret "username" (printf "quarkus.datasource.%s.username" $dbKind) false .)>:
error calling include: template: nessie/charts/nessie/templates/_helpers.tpl:339:15: executing "nessie.secretToEnv" at <index . 5>: error calling index: reflect: slice index out of range
```

I think the update to `secretToEnv` missed also updating those secrets that are not related to the catalog, specifically those `quarkus.datasource` environment variables in the deployment. I have fixed it as in this PR, by restoring the previous behavior of `secretToEnv` by adding a new template `secretKeyToEnv`, and using it in `deployment.yaml`.

I don't know if this is the best approach to fix this, but it solves the most immediate issue.

Could you please take a look if possible?

Thanks a lot!

### DIFF
--- a/helm/nessie/templates/_helpers.tpl
+++ b/helm/nessie/templates/_helpers.tpl
@@ -358,3 +358,27 @@ config types know about that symbolic name and resolve it via a SecretsProvider,
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+
+{{/*
+Define an env var from secret key.
+*/}}
+{{- define "nessie.secretKeyToEnv" -}}
+{{- $secret := index . 0 -}}
+{{- $key := index . 1 -}}
+{{- $envVarName := index . 2 -}}
+{{- $global := index . 3 -}}
+{{- if $secret -}}
+{{- $secretName := get $secret "name" -}}
+{{- $secretKey := get $secret $key -}}
+{{- with $global -}}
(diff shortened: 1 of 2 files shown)


adutra:
Hi @Milias thanks for reporting this, and I agree with your proposed fixed.

Could you just do some renaming? I think it would make more sense to rename templates as follows:

* `nessie.secretToEnv` => `nessie.catalogSecretToEnv`
* `nessie.secretKeyToEnv` => `nessie.secretToEnv`

Thanks!

Milias:
Thank you @adutra for the quick review!

I changed the names as requested and verified that I didn't accidentally break it.

adutra (approved):

PR #9550, opened 2024-09-17
Nit: fix JSON alias typo in IcebergMetadataUpdate interface

adutra:


catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
```diff
@@ -514,7 +514,7 @@ static SetTrustedLocation setTrustedLocation(String location) {
   @JsonSerialize(as = ImmutableSetProperties.class)
   @JsonDeserialize(as = ImmutableSetProperties.class)
   interface SetProperties extends IcebergMetadataUpdate {
-    @JsonAlias({"updated", "updated"})
+    @JsonAlias({"updates", "updated"})
```
adutra, line 517: tbh I don't see where `updated` is being used. But there is a similar occurrence of `@JsonAlias` in `RemoveProperties`.

snazy: Legacy Iceberg

PR #9551, opened 2024-09-17
Replace versioned-spi operation types with model operation types

snazy:


versioned/spi/src/main/java/org/projectnessie/versioned/Put.java
```diff
@@ -1,111 +0,0 @@
...
-   * @return a put operation for the key and value
-   */
-  @Nonnull
-  public static Put of(@Nonnull ContentKey key, @Nonnull Content value) {
-    return ImmutablePut.builder().key(key).valueSupplier(() -> value).build();
-  }
-
-  /** Creates a lazily-evaluated put-operation for the given key, payload and ByteString value. */
-  @Nonnull
-  public static Put ofLazy(ContentKey key, int payload, ByteString value) {
```
adutra, line 73: Don't we need the lazy deserialization anymore?

snazy: Don't seems so :shrug:

adutra (approved):

PR #9553, opened 2024-09-17
Richer access checks

### LINKED ISSUE #9559: [Catalog] Allow authz-restricted ability to configure a namespace's `location` property
This change introduces the ability to distinguish individual checks by the (external) API being used (Nessie, Iceberg) and for Nessie Catalog (Iceberg REST) to information about the kind(s) of changes being applied.

The individual changes that can be distinguished are:
* Catalog API operation
* Metadata update actions, with special actions wrt to the `location` property
* Snapshot operation
* Snapshot summary extracts (for example whether a snapshot added or removed data/delete files)

All new attributes can be retrieved from the existing `Check` type via new attributes exposed via `AccessCheckMeta` holding the source API, "for write flag" and per-content-key flags. The flags represent the mentioned "individual changes".

snazy:
This change introduces the ability to distinguish individual checks by the (external) API being used (Nessie, Iceberg) and for Nessie Catalog (Iceberg REST) to information about the kind(s) of changes being applied.

The individual changes that can be distinguished are:
* Catalog API operation
* Metadata update actions, with special actions wrt to the `location` property
* Snapshot operation
* Snapshot summary extracts (for example whether a snapshot added or removed data/delete files)

All new attributes can be retrieved from the existing `Check` type via new attributes.

Fixes #9559 (and more)

### DIFF
--- a/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
+++ b/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
@@ -33,6 +33,7 @@ import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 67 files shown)


catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
```diff
@@ -420,6 +446,89 @@ interface AddSnapshot extends IcebergMetadataUpdate {
...
+            state.addCatalogOp(CatalogOps.SNAP_OP_REPLACE);
+            break;
+          case "overwrite":
+            state.addCatalogOp(CatalogOps.SNAP_OP_OVERWRITE);
+            break;
+          case "delete":
+            state.addCatalogOp(CatalogOps.SNAP_OP_DELETE);
+            break;
+          default:
+            break;
```
adutra, line 528: Maybe add `SNAP_OP_OTHER`?

servers/services/src/main/java/org/projectnessie/services/authz/ApiContext.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package org.projectnessie.services.authz;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+@Value.Style(allParameters = true)
+public interface ApiContext {
+  String getApiName();
```
adutra, line 23: Should this be an `enum { NESSIE, ICEBERG }`?

adutra (approved):

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/rest/IcebergMetadataUpdate.java
snazy: Maybe not yet 🤷

PR #9564, opened 2024-09-18
MySQL/MariaDB: change from `BLOB` type to `LONGBLOB`

snazy:
Unfortunately, the type for binary columns was set to `BLOB`, which is way too small (max 2^16 bytes = 64k). This change updates the type to `LONGBLOB`, which is max 4G, way more than we'll ever need.


versioned/storage/jdbc/src/main/java/org/projectnessie/versioned/storage/jdbc/DatabaseSpecifics.java
```diff
@@ -192,7 +192,7 @@ static class MariaDBDatabaseSpecific implements DatabaseSpecific {
       typeIdMap.put(JdbcColumnType.OBJ_ID_LIST, Types.VARCHAR);
       typeMap.put(JdbcColumnType.BOOL, "BIT(1)");
       typeIdMap.put(JdbcColumnType.BOOL, Types.BIT);
-      typeMap.put(JdbcColumnType.VARBINARY, "BLOB");
+      typeMap.put(JdbcColumnType.VARBINARY, "LONGBLOB");
```
dimas-b, line 195: Do we need to fix this for serializers that produce columns like `t_headers`?

dimas-b (approved):

PR #9577, opened 2024-09-19
Construct `*ApiImpl` instead of injecting the V1 rest instances

snazy:
The change #9553 has to pass an additional parameter depending on the "receiving API" to the `*ApiImpl` types, which is quite complex to implement with CDI. Therefore this change removes the old `Rest*Service` types and constructs the `*ApiImpl` types directly.


snazy:
This is a prerequisite for #9553

dimas-b (approved):

PR #9578, opened 2024-09-19
Nit: Make function to build a Basic Authorization header publicly available (for internal use)

snazy:


snazy:
This is a prerequisite for #9553

api/client/src/main/java/org/projectnessie/client/auth/BasicAuthenticationProvider.java
```diff
@@ -69,25 +69,13 @@ public HttpAuthentication build(Function<String, String> configSupplier) {
...
-
-      String userPass = username + ':' + password;
-      byte[] encoded = Base64.getEncoder().encode(userPass.getBytes(StandardCharsets.UTF_8));
-      String encodedString = new String(encoded, StandardCharsets.UTF_8);
-      return "Basic " + encodedString;
-    }
-
     private BasicAuthentication(String username, String password) {
-      String header = basicHeader(username, password);
+      String header = basicAuthorizationHeader(username, password);
```
dimas-b, line 73: Why expose an auth scheme that is deprecated in Nessie in public API? Maybe add an internal utility class for it?

PR #9588, opened 2024-09-20
Events API: use Nessie model API directly and remove DTOs

adutra:


events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
```diff
@@ -193,13 +199,24 @@ protected Event newContentRemovedEvent(
...
+        .reference(makeReference(branch))
         .hash(hash.asString())
         .contentKey(contentKey)
         .commitCreationTimestamp(commitTimestamp)
         .build();
   }
 
+  private static Reference makeReference(NamedRef refName) {
+    if (refName instanceof TagName) {
+      return Tag.of(refName.getName(), null);
```
adutra, line 211: The `Branch` class from Nessie model is "annoying" because it conveys a hash; I chose to not include any hash though, because hashes are already set elsewhere and with events, we must be careful not to confuse hashes before the event with hashes after it. Putting hashes here would be ambiguous imho.

dimas-b: I see the point, but IMHO having a `Tag` of `Branch` as a "result" object without a hash on the client side is strange. 

Could we instead add tests that the hashes in different attributes agree?

dimas-b (commented):
I like the idea of using model classes for events :)

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
adutra: We could, but there will be some hashes missing: we don't have the hashes from the source ref for merges and transplants, simply because the results from the version store do not include these. We'd need a small change to the version store API, I can look into it in a separate PR.

adutra: FYI #9592

adutra: Updated.

events/api/src/main/java/org/projectnessie/events/api/TransplantEvent.java
```diff
@@ -30,4 +31,10 @@ public interface TransplantEvent
   default EventType getType() {
     return EventType.TRANSPLANT;
   }
+
+  /**
+   * The sequence of hashes from the source reference that were transplanted into the target
+   * reference.
+   */
+  List<String> getSourceHashes();
```
adutra, line 39: Wanted to note: the initial design of this API was very cautious when adding lists of elements. E.g. the commit event does not contain a list of operations, but instead, each operation is emitted as a standalone event. This was donne to prevent giant commit events with thousands of operations.

Do we need to take this aspect into consideration here as well? Can there be a transplant operation with thousands of transplanted hashes?

dimas-b: Nothing prevents huge transplants :thinking: 

On the other hand, do we really need to expose source hashes in events? If we do not record transplant sources in commit metadata, I'm not sure publishing them in events is very useful due to lack of permanent "source of truth". In the end a transplant is just a regular commit. Why would it matter to event consumers what the source commit was? The event does not provide information on conflict resolutions that happened during the transplanting anyway, does it?

adutra: > On the other hand, do we really need to expose source hashes in events?

We do not, but I need at least one hash (the last one?) so that I can create a source `Branch` object with a non-null hash, remember? ;-)

How about we expose only the last transplanted hash?

dimas-b: I believe it's source "reference" (may be a tag).

TBH, I'm not sure having a "source reference" is very meaningful for transplant results.

I think it is used in the API only to allow for reference-based access checks, but once applied, it does not look very meaningful to me. The same (transplanted) hash may be reachable from multiple references, so what's the point of reporting one reference for it in events? Event consumers probably should be attach any logic to the reference name that was used to trigger a transplant, only to the outcome of such a transplant.

Anyway, I'm fine with reporting only one source hash for transplants.

adutra: That's a very good point. I will remove the source reference from `TransplantEvent`, that will fix the issue with source hashes as well.

> I believe it's source "reference" (may be a tag).

Yes.

adutra: Added `TransplantEvent.getSize()` though, could be interesting to inform subscribers about how big was the transplant.

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
```diff
@@ -66,50 +71,56 @@ protected Event newCommitEvent(
...
-        .hashAfter(Objects.requireNonNull(result.getResultantTargetHash()).asString())
+        .hashAfter(hashAfter.asString())
         .commonAncestorHash(commonAncestorHash)
         .build();
   }
 
   protected Event newTransplantEvent(
       TransplantResult result, String repositoryId, @Nullable Principal user) {
+    Hash hashAfter = Objects.requireNonNull(result.getResultantTargetHash());
+    Hash sourceHash = result.getSourceHashes().get(result.getSourceHashes().size() - 1);
```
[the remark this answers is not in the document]
adutra: Hmm it seems your review just came in, but targets a previous state of the PR. We agreed with @dimas-b to change this and remove source hashes from transplant events.

events/api/src/main/java/org/projectnessie/events/api/MultiReferenceEvent.java
```diff
@@ -27,12 +29,6 @@
  */
 public interface MultiReferenceEvent extends Event {
 
-  /**
-   * The source reference where the committed operations come from. This is usually a branch, but
-   * not always (e.g. it could be a tag or a detached reference).
-   */
-  Reference getSourceReference();
```
dimas-b, line 34: Hmmm, this contradicts the javadoc now?

events/api/src/main/java/org/projectnessie/events/api/TransplantEvent.java
```diff
@@ -30,4 +30,7 @@ public interface TransplantEvent
   default EventType getType() {
     return EventType.TRANSPLANT;
   }
+
+  /** The number of commits that were transplanted into the target reference. */
+  int getSize();
```
dimas-b, line 35: nit: maybe `getCommitCount()`?.. "Size" invokes thoughts about data :)

events/api/src/main/java/org/projectnessie/events/api/MultiReferenceEvent.java
adutra: Updated the class javadoc.

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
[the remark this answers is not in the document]
snazy: Yea - I reviewed yesterday - but didn't submit 🤦

events/api/src/main/java/org/projectnessie/events/api/TransplantEvent.java
```diff
@@ -31,10 +30,4 @@ public interface TransplantEvent
   default EventType getType() {
     return EventType.TRANSPLANT;
   }
-
-  /**
-   * The sequence of hashes from the source reference that were transplanted into the target
-   * reference.
-   */
-  List<String> getSourceHashes();
```
snazy: WFM!

events/service/src/main/java/org/projectnessie/events/service/EventFactory.java
```diff
@@ -66,50 +70,53 @@ protected Event newCommitEvent(
...
         .properties(config.getStaticProperties())
-        .reference(ReferenceMapping.map(targetBranch))
+        .reference(makeReference(targetBranch, commit.getHash()))
         .hashBefore(Objects.requireNonNull(commit.getParentHash()).asString())
         .hashAfter(commit.getHash().asString())
         .commitMeta(
             ImmutableCommitMeta.builder()
                 .committer(Objects.requireNonNull(commitMeta.getCommitter()))
-                .authors(commitMeta.getAllAuthors())
+                .addAllAllAuthors(commitMeta.getAllAuthors())
```
snazy, line 79: immutable's funny function names 🤣

snazy (approved):

PR #9614, opened 2024-09-24
Persistable `LakehouseConfig`

snazy:
This change allows to use the `LakehouseConfig` from the Quarkus configuration, which is the current behavior, and with change to alternatively persist the `LakehouseConfig`.

For this to work, the `LakehouseConfig` object tree needs to be injectable via the CDI request-scope, so some refactoring to make this work is included in this change.

Some options, which were still in the `LakehouseConfig`, had been moved into two object types `S3StsCache` (the old `S3Sts`, but moved out of the LC tree) and `GcsConfig`.

Retrieving and updating the `LakehouseConfig` object is implemented via `LakehouseConfigManagement` (plus its implementation).

### DIFF
--- a/servers/quarkus-catalog/src/main/java/org/projectnessie/server/catalog/LakehouseConfigManagementImpl.java
+++ b/servers/quarkus-catalog/src/main/java/org/projectnessie/server/catalog/LakehouseConfigManagementImpl.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.server.catalog;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
+import java.util.UUID;
+import org.projectnessie.catalog.files.config.ImmutableAdlsOptions;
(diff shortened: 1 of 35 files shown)


servers/quarkus-catalog/src/main/java/org/projectnessie/server/catalog/LakehouseConfigManagementImpl.java
```diff
@@ -0,0 +1,110 @@
...
+public class LakehouseConfigManagementImpl implements LakehouseConfigManagement {
+  private final Persist persist;
+  private final boolean dynamic;
+  private final LakehouseConfig staticLakehouseConfig;
+
+  public LakehouseConfigManagementImpl(
+      Persist persist, boolean dynamic, LakehouseConfig staticLakehouseConfig) {
+    this.persist = persist;
+    this.dynamic = dynamic;
+    this.staticLakehouseConfig = staticLakehouseConfig;
```
dimas-b, line 44: nit: `staticLakehouseConfig` is not need when `dynamic == true`... I think we can simplify the provider that creates this object by not constructing the static config in that case.

servers/quarkus-catalog/src/main/java/org/projectnessie/server/catalog/LakehouseConfigManagementImpl.java
```diff
@@ -0,0 +1,110 @@
...
+            .versionToken(UUID.randomUUID().toString())
+            .lakehouseConfig(config)
+            .build();
+
+    try {
+      while (true) {
+        try {
+          LakehouseConfigObj existing = loadCurrent();
+          checkArgument(
+              existing.lakehouseConfig().equals(expected),
```
dimas-b, line 91: I'm not sure I completely follow the logic of this loop... `expected` is a loop invariant, so `existing.lakehouseConfig()` must be an invariant too. What does the loop achieve by `loadCurrent()`? Do we expect other processes to change the version of the `LakehouseConfigObj` without changing its `lakehouseConfig()`?

dimas-b (approved):

servers/quarkus-catalog/src/main/java/org/projectnessie/server/catalog/LakehouseConfigManagementImpl.java
snazy: Yea, it would likely fail in this assertion in case of two concurrent requests here.

PR #9617, opened 2024-09-24
[Catalog] More flexible named buckets

snazy:
This (big) change addresses a couple of things:

* Fix an issue for ADLS to consider the whole URI-authority to select a named ADLS "bucket" - this is the storage-account-name, optionally narrowed with a container/file-system name. This has been wrong in Nessie.
* Added new attributes `authority` and `pathPrefix` to the named buckets (S3/GCS/ADLS). `authority` is the bucket name (S3/GCS) or the storage-account-name + container name (ADLS). The optional `pathPrefix` can be used to restrict selecting a named bucket configuration.
* The `authority` attribute defaults to the `name` property, which in turn defaults to the named-buckets map key. In other words, both the `name` and the `authority` attributes are optional, if the map-key is the bucket name. Recommendation is to configure the `authority` attribute explicitly.
* The named bucket configuration with the longest storage location URI match will be selected. The `authority` must match, the longest matching `pathPrefix` will be used.
* `S3Sts` (renamed to `S3StsCache`) has been moved out of `S3Options`, because it is statically configured (per server).
* Technical GCS config options have been moved from `GcsOptions` into `GcsConfig`, because those are statically configured (per server).
* Named buckets normalization became unnecessary and has been removed.

### DIFF
--- a/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/PerBucket.java
+++ b/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/PerBucket.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (C) 2024 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.projectnessie.catalog.files.config;
+
+import java.util.Optional;
+
+public interface PerBucket {
+  /**
+   * The human consumable name of the bucket. If unset, the name of the bucket will be extracted
+   * from the configuration option name, e.g. if {@code
+   * nessie.catalog.service.s3.bucket1.name=my-bucket} is set, the bucket name will be {@code
+   * my-bucket}; otherwise, it will be {@code bucket1}.
+   *
+   * <p>This should only be defined if the bucket name contains non-alphanumeric characters, such as
+   * dots or dashes.
+   */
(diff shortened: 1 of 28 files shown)


catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/PerBucket.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Optional;
+
+public interface PerBucket {
+  /**
+   * The human consumable name of the bucket. If unset, the name of the bucket will be extracted
+   * from the configuration option name, e.g. if {@code
+   * nessie.catalog.service.s3.bucket1.name=my-bucket} is set, the bucket name will be {@code
+   * my-bucket}; otherwise, it will be {@code bucket1}.
+   *
+   * <p>This should only be defined if the bucket name contains non-alphanumeric characters, such as
```
dimas-b, line 27: nit: I think "should only" is too strong for this recommendation. We respect `name` regardless of the value it contains. I'd say `This can be used if the bucket name contains non-alphanumeric characters ...`

catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/PerBucket.java
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * The authority part in a storage location URI. This is the bucket name for S3 and GCS, for ADLS
+   * this is the storage account name (optionally prefixed with the container/file-system name).
+   * Defaults to {@link #name()}.
+   *
+   * <p>For S3 and GCS this option should mention the name of the bucket.
+   *
+   * <p>For ADLS: The value of this option is using the {@code container@storageAccount} syntax. It
+   * is mentioned as {@code <file_system>@<account_name>} in the <a
```
dimas-b, line 40: Do you mean that `.dfs.core.windows.net` (or similar) is part of `authority` or not?

dimas-b (dismissed):

catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/PerBucket.java
snazy: Yep, that includes `.dfs.core.windows.net`

dimas-b (approved):

PR #9621, opened 2024-09-25
[BUG-9620] Adding metadata.namespace to all helm template files

ev1lm0nk3y:
This should resolve https://github.com/projectnessie/nessie/issues/9620

Issue #9620, opened 2024-09-25
[Bug]: Helm chart does not specify metadata.namespace which breaks CD tool ArgoCD


ev1lm0nk3y:
### What happened

The helm chart for nessie does not specify namespace in any resource, so
1. you could accidentally install it in a namespace you didn't want to
2. CD tools, like ArgoCD, don't know where to put the objects

### How to reproduce it

1. `helm template testing nessie/nessie`
2. Check manifests for the lack of `metadata.namespace` entries

### Nessie server type (docker/uber-jar/built from source) and version

0.98.0

### Client type (Ex: UI/Spark/pynessie ...) and version

helm

adutra:
Hi @ev1lm0nk3y thanks for reporting this!

The absence of namespace fields in our chart is actually by design. See this issue for context: https://github.com/helm/helm/issues/5465. Helm authors generally consider a best practice to _not_ include any namespace field in any resource, so that Helm can honor the `--namespace` flag passed via the command line.

I'm not an expert in ArgoCD, but isn't ArgoCD capable of installing the chart in a given namespace? Looking at [this page](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/), it seems this should be possible.

ev1lm0nk3y:
I was getting all kinds of errors while trying to deploy it so that is why
I filed the bug. I understand honoring the namespace flag but the link you
provided kinda said exactly what I'm proposing. What happens with helm, as
far as i understand, the namespace flag doesn't add that value to the
generated manifest, it can't because helm is effectively a templating
engine. When that flag is used, helm will set that kube context to that
namespace for the deployment, thus allowing all the resources to be
installed without needing to explicitly define it.

Without `namespace: {{ .Release.Namespace }}` only the basic invocation of
helm would work.

ev1lm0nk3y:
Also, while Argo can install helm charts in various namespaces, it is run
from the argocd namespace usually and I haven't seen it do any trickery
with the context before applying. ArgoCD will effectively run `helm
template --namespace foo myrelease repo/chart` and then apply the output.

PR #9625, opened 2024-09-25
Helm chart: more flexible services configuration

adutra:


helm/nessie/templates/servicemonitor.yaml
```diff
@@ -21,15 +21,14 @@ apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
   name: {{ template "nessie.fullname" . }}
-  namespace: {{ .Release.Namespace }}
```
adutra, line 24: Unrelated, but see #9620.

dimas-b (dismissed):

helm/nessie/templates/servicemonitor.yaml
snazy: So this is rather the "opposite" of #9621, but addresses the same issue, correct?

adutra: Yes, I wanted to make things consistent, even if I actually agree with the fix proposed in #9621.

dimas-b (approved):

PR #9628, opened 2024-09-25
Remove unused `sourceHashes` from `TransplantResult`

snazy:
This attribute is not returned to clients and never read/used elsewhere.


adutra:
It's used here: #9588.

Please do not delete these attributes until we decide how to proceed on that PR.

adutra (approved):

PR #9642, opened 2024-09-26
[Catalog] Do not return `*FileIO` options from the Iceberg REST config endpoint

snazy:
Nessie returns all necessary `*FileIO` configurations via the `loadTable` (et al) responses, so it is not necessary to return that information from the Iceberg REST config endpoint and is therefore removed by this change.


dimas-b (commented):
I can only see CLI and test changes in this PR... What makes the config endpoint stop sending FileIO properties? :thinking:

snazy:
> I can only see CLI and test changes in this PR... What makes the config endpoint stop sending FileIO properties? 🤔

Ugh! Looks like some stuff was lost after some rebases (and my reflog doesn't help :( )

dimas-b (approved):

PR #9648, opened 2024-09-26
Cache: keep (deserialized) object around

snazy:
Refactor the cached-key-value to keep the deserialized object around leveraging a `java.lang.ref.SoftReference` to it. While this increases the overall heap pressure for each cached entity, the referenced objects are eligible for garbage collection, if Java GC decides to do so. This change continues to keep the serialized representation around (using a strong reference) - a "simple" `byte[]` is still way less individual Java objects compared to an "exploded" `Obj`, so less effort for GC.

This change should help to reducing the deserializing effort especially for very frequently accessed objects, despite the semantics of `SoftReference`.

### DIFF
--- a/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
+++ b/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
@@ -31,6 +31,7 @@ import com.github.benmanes.caffeine.cache.Expiry;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
 import jakarta.annotation.Nonnull;
+import java.lang.ref.SoftReference;
 import java.time.Duration;
 import org.checkerframework.checker.index.qual.NonNegative;
 import org.projectnessie.versioned.storage.common.exceptions.ObjTooLargeException;
@@ -44,10 +45,11 @@ import org.projectnessie.versioned.storage.serialize.ProtoSerialization;
 class CaffeineCacheBackend implements CacheBackend {
 
   public static final String CACHE_NAME = "nessie-objects";
-  private static final byte[] NON_EXISTING_SENTINEL = "NON_EXISTING".getBytes(UTF_8);
+  private static final CacheKeyValue NON_EXISTING_SENTINEL =
+      new CacheKeyValue("x", ObjId.EMPTY_OBJ_ID, 0L, new byte[0], null);
 
   private final CacheConfig config;
-  final Cache<CacheKeyValue, byte[]> cache;
+  final Cache<CacheKeyValue, CacheKeyValue> cache;
 
   private final long refCacheTtlNanos;
   private final long refCacheNegativeTtlNanos;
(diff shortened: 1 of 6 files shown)


versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -58,15 +60,15 @@ class CaffeineCacheBackend implements CacheBackend {
     refCacheTtlNanos = config.referenceTtl().orElse(Duration.ZERO).toNanos();
     refCacheNegativeTtlNanos = config.referenceNegativeTtl().orElse(Duration.ZERO).toNanos();
 
-    Caffeine<CacheKeyValue, byte[]> cacheBuilder =
+    Caffeine<CacheKeyValue, CacheKeyValue> cacheBuilder =
```
dimas-b, line 63: nit: why not use two fixed `String` objects for values: one for "live" entries, another for `NON_EXISTING_SENTINEL`?.. probably not a lot of difference memory-wise, but it could be more readable, I hope.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -311,6 +338,26 @@ public int hashCode() {
     public String toString() {
       return "{" + repositoryId + ", " + id + '}';
     }
+
+    Obj getObj() {
+      Obj obj = (Obj) this.object.get();
+      if (obj == null) {
+        obj = ProtoSerialization.deserializeObj(id, 0L, this.serialized, null);
+        // re-create the soft reference - but don't care about JMM side effects
+        this.object = new SoftReference<>(obj);
```
dimas-b, line 347: Should we perhaps adjust defaults in `CacheSizing` to allow for more "free" heap for the soft-referenced objects?

dimas-b (approved):

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: From 70% to 60%?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: Would need a composite for the serialized and soft-reference anyway. Sure, it's a bit more readable, but I'm a bit more concerned about heap pressure here.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
dimas-b: SGTM

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
dimas-b: Values could we static constants. We only need them to distinguish the sentinel from real values, but the actually value is always in the key... anyway this is totally optional.

dimas-b: I think I was confused. Please disregard.

dimas-b (dismissed):

dimas-b (approved):

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheSizing.java
```diff
@@ -82,14 +82,14 @@ void tinyHeapNoCache() {
...
     // Assuming a 4G max heap, requesting 70% (358MB), sizing must yield 2867MB.
     soft.assertThat(CacheSizing.builder().build().calculateEffectiveSizeInMB(BYTES_4G))
-        .isEqualTo(2867);
+        .isEqualTo(2457);
   }
 
   @Test
   void defaultSettings1G() {
     soft.assertThat(CacheSizing.builder().build().calculateEffectiveSizeInMB(BYTES_1G))
         // 70 % of 1024 MB
```
dimas-b, line 91: nit: 60%

PR #9662, opened 2024-09-29
Update Docker Compose instructions in Getting Started guide

edgarrmondragon:
Going through the steps, I noticed that the Keycloak container wouldn't start because it's referencing a file at `../authn-keycloak/config/iceberg-realm.json`, which does not exist if one just downloads the `docker-compose.yml` file in a random directory.

These updated instructions are more accurate and do allow me to start all the containers, so let me know what you think.

Thanks!


snazy (approved):
Thanks @edgarrmondragon !

Issue #9656, opened 2024-09-27
Setting up development environment fails with `gradle quarkusDev`


Koenkk:
### Issue description

I'm trying to setup a local development environment for Nessie.

The first problem I ran into is after executing `./gradlew quarkusDev` the UI is not accessible on `http://localhost:19120`. It looks to me that `:nessie-quarkus` does not start when running `./gradlew quarkusDev`. This is solved by executing `./gradlew :nessie-quarkus:quarkusDev`, am I doing something wrong or is the information in the `README.md` outdated?

Now that the UI is accessible, I run into a second problem. Whenever I open the UI I get the following error (in the terminal running `./gradlew :nessie-quarkus:quarkusDev`):

```
     2024-09-27 16:08:36,125 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-4) HTTP Request to /api/v1/trees/tree failed, error id: c9627606-91f0-47d4-976a-6af0365bcab2-6: java.lang.NoClassDefFoundError: org/projectnessie/nessie/relocated/protobuf/UnsafeByteOperations
        at org.projectnessie.versioned.storage.common.indexes.StoreIndexImpl.serialize(StoreIndexImpl.java:522)
        at org.projectnessie.versioned.storage.common.logic.CommitLogicImpl.buildCommitObj(CommitLogicImpl.java:718)
        at org.projectnessie.versioned.storage.common.logic.CommitLogic.buildCommitObj(CommitLogic.java:186)
        at org.projectnessie.versioned.storage.common.logic.CommitLogicImpl.doCommit(CommitLogicImpl.java:277)
        at org.projectnessie.versioned.storage.common.logic.RepositoryLogicImpl.initializeInternalRef(RepositoryLogicImpl.java:261)
        at org.projectnessie.versioned.storage.common.logic.RepositoryLogicImpl.initialize(RepositoryLogicImpl.java:81)
        at org.projectnessie.versioned.storage.common.logic.RepositoryLogicImpl.initialize(RepositoryLogicImpl.java:73)
        at org.projectnessie.quarkus.providers.storage.PersistProvider.produceWithInitializedRepository(PersistProvider.java:104)
        at org.projectnessie.quarkus.providers.storage.PersistProvider_ProducerMethod_produceWithInitializedRepository_sL325iqwi3y6FdhYHbfY
```

[...]

PR #9668, opened 2024-09-30
Fix running `nessie-quarkus` instructions

Koenkk:
See https://github.com/projectnessie/nessie/issues/9656


snazy (approved):
Thanks @Koenkk !

PR #9686, opened 2024-10-02
Events RI: use Quarkus Messaging extension

adutra:
This is a complete rewrite of the RI module, focusing on:

* Using Quarkus + Messaging extension
* Also showcasing an example with NATS


events/ri/src/main/java/org/projectnessie/events/ri/messaging/kafka/json/KafkaJsonEventSubscriber.java
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * ServiceLoader shim. This is the class that will be instantiated by the ServiceLoader and
+   * registered with the Events service.
+   *
+   * <p>It delegates to the actual implementation, which is a CDI bean.
+   */
+  public static class ServiceLoaderShim extends DelegatingEventSubscriber {
+
+    public ServiceLoaderShim() {
+      super(CDI.current().select(KafkaJsonEventSubscriber.class).get());
```
dimas-b, line 47: I like the idea of leveraging Quarkus CDI for subscribers. However, do we really need to go through the `ServiceLoader` in this case? Why not have a `@Default` no-op `EventSubscriber`, which a real one will override using `@Priority`?

adutra: I thought about that too, but with the caveat that we need to support the case where there are more than one subscriber registered (unlikely, but supported). But it's still doable by injecting `Instance<EventSubscriber>` and iterating over all the discovered beans.

Let's do that in a follow-up PR, does that sound OK to you?

dimas-b: +1 for follow-up

dimas-b (approved):

PR #9696, opened 2024-10-03
Events SPI: load implementations via CDI

adutra:
...instead of ServiceLoader.


events/ri/README.md
```diff
@@ -161,10 +161,9 @@ messages without waiting for the broker's acknowledgement. If your implementatio
 the broker's acknowledgement synchronously, then you should report to Nessie that this subscriber is
 going to be blocking by returning `true` from the `isBlocking()` method.
 
-Finally, the messaging subscribers use a "shim" to initialize the subscriber. This is because
-subscribers are loaded by ServiceLoader, and the loaded instance is not managed by Quarkus. The shim
-provides a bridge between the Quarkus-managed instance and the ServiceLoader-loaded instance.
-See `KafkaJsonEventSubscriber.ServiceLoaderShim` for example.
+Finally, `EventSubscriber` implementations are discovered automatically by Quarkus, and therefore
+must be declared as application-scoped  CDI beans. This is done by annotating the implementation
+class with `@ApplicationScoped`.
```
dimas-b, line 166: nit: I think `@Singleton` is ok too.

events/service/src/main/java/org/projectnessie/events/service/EventSubscribers.java
```diff
@@ -119,21 +107,23 @@ public synchronized void start(Function<EventSubscriber, EventSubscription> subs
...
+          } catch (Exception e) {
+            errors.add(e);
+          }
         }
+        if (!errors.isEmpty()) {
+          RuntimeException e = new RuntimeException("Error closing at least one subscriber");
+          errors.forEach(e::addSuppressed);
+          throw e;
+        }
+        LOGGER.info("Done closing subscribers.");
```
dimas-b, line 125: This is unreachable in case of closing exceptions... I'm not sure where the `RuntimeException` goes... would it make sense to log it just in case?

I know this behaviour is not changed by this PR, so take this comment as a nit :)

dimas-b (dismissed):

events/service/src/main/java/org/projectnessie/events/service/EventSubscribers.java
adutra: I fear that it would be logged twice, but why not – better twice than never ;-)

servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/events/fixtures/MockEventSubscriber.java
```diff
@@ -15,103 +15,19 @@
...
-  @Override
-  public void onReferenceDeleted(ReferenceDeletedEvent event) {
-    recordEvent(event);
-  }
 
-  @Override
-  public void onCommit(CommitEvent event) {
-    recordEvent(event);
-  }
+public interface MockEventSubscriber extends EventSubscriber {
```
adutra, line 22: Turns out under `testFixtures` a bean is not detected by Quarkus, so I had to move it to `test` instead.

dimas-b (approved):

PR #9711, opened 2024-10-07
Helm chart: explicitly include namespace in created resources

### LINKED ISSUE #9620: [Bug]: Helm chart does not specify metadata.namespace which breaks CD tool ArgoCD

adutra:
Fixes #9620. Obsoletes #9621.

### DIFF
--- a/helm/nessie/templates/servicemonitor.yaml
+++ b/helm/nessie/templates/servicemonitor.yaml
@@ -20,7 +20,8 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: {{ template "nessie.fullname" . }}
+  name: {{ include "nessie.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "nessie.labels" . | nindent 4 }}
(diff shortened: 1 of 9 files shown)


helm/nessie/templates/servicemonitor.yaml
```diff
@@ -20,7 +20,8 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  name: {{ template "nessie.fullname" . }}
+  name: {{ include "nessie.fullname" . }}
```
snazy, line 23: Intentional change?

adutra: Yes, to align with other occurrences of this template inclusion. (In this context, using `template` or `include` yields the same result.)

dimas-b (approved):

PR #9731, opened 2024-10-10
Persist: introduce `deleteWithReferenced(Obj)`

snazy:
... to delete an `Obj` only if its `referenced()` timestamp has the expected value.


snazy:
This is a prerequisite for #9688

adutra (approved):

PR #9762, opened 2024-10-15
server-admin-tool intTest: Re-add `forkEvery`

snazy:
Reverts #9371

### DIFF
--- a/tools/server-admin/build.gradle.kts
+++ b/tools/server-admin/build.gradle.kts
@@ -185,6 +185,9 @@ if (Os.isFamily(Os.FAMILY_MAC) && System.getenv("CI") != null) {
 }
 
 tasks.named<Test>("intTest").configure {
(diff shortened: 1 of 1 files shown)


dimas-b (approved):

tools/server-admin/build.gradle.kts
```diff
@@ -185,6 +185,9 @@ if (Os.isFamily(Os.FAMILY_MAC) && System.getenv("CI") != null) {
 }
 
 tasks.named<Test>("intTest").configure {
+  // Reduce likelihood of OOM due to too many Quarkus apps in memory;
+  // Ideally, set this number to the number of IT classes to run for each backend.
+  forkEvery = 5
```
adutra, line 190: In fact, I think it's not a good idea to use MultiEnv with Quarkus. We should probably look into removing MultiEnv from this module's tests.

PR #9863, opened 2024-11-04
Refactor Nessie's HTTP authentication (Quarkus 3.16 prep)

snazy:
Previously we had `NessieHttpAuthenticator`, which extended `io.quarkus.vertx.http.runtime.security.HttpAuthenticator`. Sadly, `HttpAuthenticator`'s constructor has been made package-private so the approach using `NessieHttpAuthenticator` no longer works.

This change keeps Nessie's configuration approach (simplifies it a bit) but uses public API extension points by implementing a Nessie `HttpAuthenticationMechanism` implementation. It works a bit differently, but the result is the same.


servers/quarkus-auth/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
```diff
@@ -1,146 +0,0 @@
...
-          if (anonymousPaths.contains(path)
-              || anonymousPathPrefixes.stream().anyMatch(path::startsWith)) {
-            return anonymous();
-          }
-        }
-
-        // Disallow unauthenticated requests when requested by configuration.
-        // Note: Quarkus by default permits unauthenticated requests unless there are
-        // specific authorization rules that validate the security identity.
-        throw new AuthenticationFailedException("Missing or unrecognized credentials");
```
dimas-b, line 140: Aren't we losing this use case with the new auth approach? The idea here was to deny access when the Nessie auth property is set in the config but a `SecurityIdentity` was not established. I'm not sure it works the same way with this PR.

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
```diff
@@ -0,0 +1,116 @@
...
+    return PRIORITY;
+  }
+
+  @Override
+  public Uni<SecurityIdentity> authenticate(
+      RoutingContext context, IdentityProviderManager identityProviderManager) {
+    if (enabled) {
+      var path = context.request().path();
+      if (checkPath(path)) {
+        return Uni.createFrom().nullItem();
```
dimas-b, line 87: Should we use an `AnonymousAuthenticationRequest` is this case rather than return null?

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      if (!path.contains("..")) {
+        if (getClass().getClassLoader().getResource("META-INF/resources" + path) != null) {
```
dimas-b, line 91: Sorry, I'm very confused why we need this special case here :thinking:

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
snazy: See https://github.com/projectnessie/nessie/pull/9845#discussion_r1827412980

servers/quarkus-auth/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
snazy: Unauthenticated requests fail with the changes in this PR

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
dimas-b: I'm not sure that comment applies. Returning `null` means this 'mechanism" did not handle the request (so the next mechanism will have a try), but the intention appears to be that we want the request to be handles specifically as "anonymous".

dimas-b: I think the new Nessie code relies on Quarkus to attempt creating an anonymous identity automatically if none of the authentication mechanisms returned an identity... IMHO, this is not intuitive. Would you mind adding a comment about that? Otherwise, the connection between `NessieAuthMechanism` and `NessieAnonymousIdentityProvider` is rather obscure.

dimas-b: Also, Using `0` priority does not guarantee that there's no other auth mechanism after Nessie :shrug:

servers/quarkus-auth/src/main/java/org/projectnessie/server/authn/NessieHttpAuthenticator.java
dimas-b: right, my bad, I missed that, sorry.

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
dimas-b: Why not get a `IdentityProviderManager` injected and do an explicit `.authenticate(AnonymousAuthenticationRequest.INSTANCE)` request?

snazy: > Why not do an explicit `.authenticate(AnonymousAuthenticationRequest.INSTANCE)` request?

https://github.com/projectnessie/nessie/pull/9845#discussion_r1827412980 : "...because `IdentityProvider#authenticate` requires a 2nd `AuthenticationRequestContext`, which we don't have there."

snazy: > Also, Using `0` priority does not guarantee that there's no other auth mechanism after Nessie 🤷

"Multiple mechanisms are sorted in descending order..."

dimas-b: Priority `-1` is allowed :shrug:

dimas-b: Re: anonymous identity, I mean this:
```
diff --git a/servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java b/servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
index b4a726c9d..9c3a7e9ae 100644
--- a/servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
+++ b/servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
@@ -84,7 +84,7 @@ public class NessieAuthMechanism implements HttpAuthenticationMechanism {
     if (enabled) {
       var path = context.request().path();
       if (checkPath(path)) {
-        return Uni.createFrom().nullItem();
+        return identityProviderManager.authenticate(AnonymousAuthenticationRequest.INSTANCE);
       }
 
       if (!path.contains("..")) {
```

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAnonymousIdentityProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public Class<AnonymousAuthenticationRequest> getRequestType() {
+    return AnonymousAuthenticationRequest.class;
+  }
+
+  @Override
+  public Uni<SecurityIdentity> authenticate(
+      AnonymousAuthenticationRequest request, AuthenticationRequestContext context) {
+    return anonymousIdentityProvider.authenticate(request, context);
```
dimas-b, line 44: nit: would it work if we just had a "provider" for this `AnonymousIdentityProvider` (as opposed to delegating to it)?

servers/quarkus-server/src/test/java/org/projectnessie/server/TestAnonPathsWithAuthDisabled.java
```diff
@@ -0,0 +1,87 @@
...
+        "GET,/api/v1/config",
+        "GET,/api/v1/config",
+        "GET,/iceberg/v1/config"
+      })
+  public void anonPath(String method, String path) throws Exception {
+    var urlConnection =
+        (HttpURLConnection) getQuarkusBaseTestUri().resolve(path).toURL().openConnection();
+    urlConnection.setRequestMethod(method);
+    try (var in = urlConnection.getInputStream()) {
+      System.out.println(new String(in.readAllBytes()));
```
dimas-b, line 64: Could we collect the output in memory and add to assertions instead?

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
dimas-b: @snazy : Could you comment with some examples for these resources and how accessing them worked before this PR?

dimas-b (commented):
LGTM overall :+1: just a couple more comments

servers/quarkus-authn/src/main/java/org/projectnessie/server/authn/NessieAuthMechanism.java
dimas-b: TBH, I'm skeptical about opening anonymous access to all jar-based resources.

snazy: @dimas-b - this is how it worked before and how it's standard in Quarkus, right?

snazy: @dimas-b what's the issue with allowing _static_ resources?
It does _not_ allow access to _any_ resource - only _static web resources_.

dimas-b: If opens them by default, without a user action... which is not ideal from my POV. Previously we listed all anonymous resources explicitly. I'd prefer to keep the explicit whitelist.

However, if you have some new use cases for static resources, I do not have a strong objection... Maybe we can list that as a changelog item. WDYT?

dimas-b (approved):

PR #9900, opened 2024-11-11
Testing: disable looking up GCP credentials

snazy:
This is a follow-up of #9866 (and it's revert #9899). The stack traces indicate that the global-tracer field init race happens when `BigTableBackendBuilder` tries to fetch the GCP credentials via `Instance<CredentialsProvider>`, which eventually lets the Google client try to find the default app-credentials, which then implicitly initializes OpenCensus->OpenTelemetry.

The workaround here is to add a new undocumented configuration option to disable GCP credentials lookup, which is also not necessary when using the BT emulator.

### DIFF
--- a/servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/BigTableBackendBuilder.java
+++ b/servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/BigTableBackendBuilder.java
@@ -50,10 +50,14 @@ public class BigTableBackendBuilder implements BackendBuilder {
   public Backend buildBackend() {
     CredentialsProvider credentialsProvider = null;
     Exception googleCredentialsException = null;
-    try {
-      credentialsProvider = credentialsProviderInstance.get();
-    } catch (Exception e) {
-      googleCredentialsException = e;
+    // The disableCredentialsLookupForTests() is only used for tests and only to prevent the
+    // OpenTelemetry/global-tracer static-field initialization race described in #9866
+    if (!bigTableConfig.disableCredentialsLookupForTests()) {
+      try {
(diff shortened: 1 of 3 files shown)


servers/quarkus-common/src/main/java/org/projectnessie/quarkus/providers/storage/BigTableBackendBuilder.java
```diff
@@ -50,10 +50,14 @@ public class BigTableBackendBuilder implements BackendBuilder {
   public Backend buildBackend() {
     CredentialsProvider credentialsProvider = null;
     Exception googleCredentialsException = null;
-    try {
-      credentialsProvider = credentialsProviderInstance.get();
-    } catch (Exception e) {
-      googleCredentialsException = e;
+    // The disableCredentialsLookupForTests() is only used for tests and only to prevent the
+    // OpenTelemetry/global-tracer static-field initialization race described in #9866
+    if (!bigTableConfig.disableCredentialsLookupForTests()) {
```
dimas-b, line 55: I'm not sure "ForTests" actually adds any guardrails, but it makes the java code look odd. Could you name the java method simply as `disableCredentialsLookup()`? I suppose end users will get enough warning from having "credentials" in the name.

dimas-b: ... and of course, we'll keep it undocumented (other than javadoc)

dimas-b: IMHO, the code reads awkwardly, but up to you :)

dimas-b (approved):

PR #9906, opened 2024-11-12
Catalog/Iceberg: support new `remove-partition-specs` metadata-update

snazy:
The functionality is untested, because there's no API in Iceberg/Java that uses that metadata-update.


catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -1352,7 +1353,6 @@ public static void setCurrentSchema(
 
   public static void addPartitionSpec(AddPartitionSpec u, IcebergTableMetadataUpdateState state) {
     IcebergPartitionSpec spec = u.spec();
-    checkArgument(spec.specId() >= 0, "Invalid spec-ID %s", spec.specId());
```
adutra, line 1355: Why this change?

snazy: Good Q - don't recall - reverted.

adutra (approved):

adutra (approved):

PR #9965, opened 2024-11-21
Build: tackle Gradle deprecations marked as for-removal in Gradle 9

snazy:


build-logic/build.gradle.kts
```diff
@@ -45,3 +45,7 @@ tasks.register("codeChecks").configure {
   description = "Runs code style and license checks"
   dependsOn("spotlessCheck")
 }
+
+kotlin {
```
snazy, line 49: Gradle complains about Java 23 vs Kotlin targeting 22 max (and going to fail in Gradle 9 in such a case).

build-logic/src/main/kotlin/Utilities.kt
```diff
@@ -214,9 +215,9 @@ fun DependencyHandlerScope.nessieProject(
   if (artifactId.startsWith(":")) {
     throw IllegalArgumentException("artifactId for nessieProject() must not start with ':'")
   }
-  return if (!isIncludedInNesQuEIT(project(":").dependencyProject.gradle)) {
```
snazy, line 217: That `dependencyProject` getter is deprecated for removal in Gradle 9

build-logic/src/main/kotlin/nessie-common-src.gradle.kts
```diff
@@ -178,7 +178,7 @@ class MemoizedGitInfo {
   companion object {
     private fun execProc(rootProject: Project, cmd: String, vararg args: Any): String {
       val buf = ByteArrayOutputStream()
-      rootProject.exec {
+      rootProject.providers.exec {
```
snazy, line 181: Not sure why, but `Project.exec` got deprecated for removal in 9, but doing that via `ProviderFactory.exec` is not :shrug:

tools/doc-generator/site-gen/build.gradle.kts
```diff
@@ -166,99 +166,87 @@ val cliHelp by tasks.registering(JavaExec::class) {
...
-  mainClass = gcMainClass
-  args("--help")
-
-  doFirst {
-    delete(gcHelpDir)
-  }
-
-  standardInput = InputStream.nullInputStream()
-  standardOutput = ByteArrayOutputStream()
+val gcHelp by tasks.registering(Sync::class) {
```
snazy, line 169: The change in this project is to generate each help output using one Gradle task per invocation into distinct directories and then collect those using a `Sync` task

... `javaexec` is deprecated for removal in 9

snazy:
CI failure is unrelated, should be fixed via #9970

dimas-b (approved):

PR #10011, opened 2024-11-29
Add clusterIP and traffic policy to helm service config

### LINKED ISSUE #9998: [Feature]: Helm — specify `externalTrafficPolicy` for service
### Description

I'd like my external load balancer to handle the load balancing.

### Expected Use Cases

Override the value for `externalTrafficPolicy` for the main and extra service definitions

N/A

EternalDeiwos:
Closes #9998


helm/nessie/templates/service.yaml
```diff
@@ -39,6 +39,13 @@ spec:
       protocol: TCP
     {{- end }}
   sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- if $svc.clusterIP }}
```
adutra, line 42:
```suggestion
  {{- if .Values.service }}
```

adutra (changes requested):

helm/nessie/templates/service.yaml
EternalDeiwos: Good catch; I corrected to `.Values.service.clusterIP` though because I don't want the field to appear in the spec at all if it is not specified.

helm/nessie/templates/service.yaml
```diff
@@ -93,4 +100,13 @@ spec:
...
   {{- end }}
-{{- end }}
\ No newline at end of file
+  {{- if $svc.clusterIP }}
+  clusterIP: {{ $svc.clusterIP }}
+  {{- end }}
+  {{- if $svc.externalTrafficPolicy }}
+  externalTrafficPolicy: {{ $svc.externalTrafficPolicy }}
+  {{- end }}
+  {{- if and (ge (int .Capabilities.KubeVersion.Minor) 31) ($svc.trafficDistribution) }}
```
adutra, line 109:
```suggestion
  {{- if and (ge (int $.Capabilities.KubeVersion.Minor) 31) ($svc.trafficDistribution) }}
```

adutra (changes requested):

adutra:
@EternalDeiwos I think we should also add `internalTrafficPolicy` for completeness, wdyt?

EternalDeiwos:
@adutra added `internalTrafficPolicy` as well. Originally I had it and removed it before I opened the PR since considered it more application configuration rather than environment configuration, but I guess there are cases where one might want to override it.

Happy to add it in.

helm/nessie/templates/service.yaml
```diff
@@ -39,6 +39,13 @@ spec:
       protocol: TCP
     {{- end }}
   sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- if .Values.service.clusterIP }}
+  clusterIP: {{ .Values.service.clusterIP }}
+  {{- end -}}
```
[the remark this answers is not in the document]
EternalDeiwos: Fixed. Thanks.

helm/nessie/templates/service.yaml
EternalDeiwos: Fixed. Thanks.

helm/nessie/templates/service.yaml
```diff
@@ -39,6 +39,14 @@ spec:
       protocol: TCP
     {{- end }}
   sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- if .Values.service.clusterIP }}
+  clusterIP: {{ .Values.service.clusterIP }}
+  {{- end }}
+  externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy }}
```
adutra, line 45: I think we still have a problem here: the tests are failing because this value can only be set if the service is of type `LoadBalancer`.

So, I think we need to wrap this line in an if/else template block.

EternalDeiwos: That's a good catch. It will also work for `NodePort` as well.

Would something like this work?

```
{{- if and $svc.externalTrafficPolicy (or (eq $svc.type "LoadBalancer") (eq $svc.type "NodePort")) }}
```

Or would you prefer this separated out with multiple if-statements wrapping each other? Not sure what would go in the _else_ block.

adutra: I think your suggestion is fine 👍

adutra (approved):
LGTM, thank you for your contribution (and patience)!

snazy (approved):
Thanks!

PR #10033, opened 2024-12-03
Catalog: return Iceberg snapshot log based on Nessie commit history

### LINKED ISSUE #9969: [Bug]: Snapshot validation fails when running table compaction
### What happened

I am running table compaction using Spark Actions. my spark action code is:
```scala
sparkActions
            .rewriteDataFiles(table)
            .option(RewriteDataFiles.PARTIAL_PROGRESS_ENABLED, "true")
            .option(RewriteDataFiles.PARTIAL_PROGRESS_MAX_COMMITS, "200")
            .option(RewriteDataFiles.USE_STARTING_SEQUENCE_NUMBER, "false")
            .option(RewriteDataFiles.REWRITE_JOB_ORDER, RewriteJobOrder.FILES_DESC.orderName())
            .option(RewriteDataFiles.TARGET_FILE_SIZE_BYTES, (1024 * 1024 * 1024).toString) // 1 GB
            .filter(Expressions.lessThan("watermark", "2024-11-17T00:00:00.000+00:00"))
            .filter(Expressions.greaterThanOrEqual("watermark", "2024-11-16T00:00:00.000+00:00"))
            .zOrder("watermark", "user_id")
            .execute()

```

The job runs fine, however none of the file groups are able to commit. the commits fail with the following error:

```2024-11-21T15:02:04,127 ERROR [Rewrite-Service-4] actions.BaseCommitService: Failure during rewrite commit process, partial progress enabled. Ignoring
org.apache.iceberg.exceptions.ValidationException: Cannot determine history between starting snapshot 8855497291943185445 and the last known ancestor 1637454130654249874
	at org.apache.iceberg.exceptions.ValidationException.check(ValidationException.java:49) ~[__app__.jar:?]
	at org.apache.iceberg.MergingSnapshotProducer.validationHistory(MergingSnapshotProducer.java:805) ~[__app__.jar:?]
	at org.apache.iceberg.MergingSnapshotProducer.addedDeleteFiles(MergingSnapshotProducer.java:563) ~[__app__.jar:?]
	at org.apache.iceberg.MergingSnapshotProducer.validateNoNewDeletesForDataFiles(MergingSnapshotProducer.java:479) ~[__app__.jar:?]
	at org.apache.iceberg.MergingSnapshotProducer.validateNoNewDeletesForDataFiles(MergingSnapshotProducer.java:425) ~[__app__.jar:?]
	at org.apache.iceberg.BaseRewriteFiles.validate(BaseRewriteFiles.java:140) ~[__app__.jar:?]
	at org.apache.iceberg.SnapshotProducer.apply(SnapshotProducer.java:252) ~[__app__.jar:?]
	at org.apache.iceberg.SnapshotProducer.lambda$commit$2(SnapshotProducer.java:403) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder.runSingleThreaded(Tasks.java:219) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:203) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder.run(Tasks.java:196) ~[__app__.jar:?]
	at org.apache.iceberg.SnapshotProducer.commit(SnapshotProducer.java:401) ~[__app__.jar:?]
	at org.apache.iceberg.actions.RewriteDataFilesCommitManager.commitFileGroups(RewriteDataFilesCommitManager.java:92) ~[__app__.jar:?]
	at org.apache.iceberg.actions.RewriteDataFilesCommitManager.commitOrClean(RewriteDataFilesCommitManager.java:114) ~[__app__.jar:?]
	at org.apache.iceberg.actions.RewriteDataFilesCommitManager$CommitService.commitOrClean(RewriteDataFilesCommitManager.java:152) ~[__app__.jar:?]
	at org.apache.iceberg.actions.BaseCommitService.commitReadyCommitGroups(BaseCommitService.java:229) ~[__app__.jar:?]
	at org.apache.iceberg.actions.BaseCommitService.offer(BaseCommitService.java:151) ~[__app__.jar:?]
	at org.apache.iceberg.actions.RewriteDataFilesCommitManager$CommitService.offer(RewriteDataFilesCommitManager.java:144) ~[__app__.jar:?]
	at org.apache.iceberg.spark.actions.RewriteDataFilesSparkAction.lambda$doExecuteWithPartialProgress$4(RewriteDataFilesSparkAction.java:347) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413) ~[__app__.jar:?]
	at org.apache.iceberg.util.Tasks$Builder$1.run(Tasks.java:315) ~[__app__.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) ~[?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) ~[?:?]
	at java.lang.Thread.run(Thread.java:829) ~[?:?]
2024-11-21T15:02:04,134 INFO [Driver] actions.BaseCommitService: Closing commit service for copilot.copilot.messages waiting for all commits to finish
2024-11-21T15:02:04,228 ERROR [Driver] actions.RewriteDataFilesSparkAction: partial-progress.enabled is true but no rewrite commits succeeded. Check the logs to determine why the individual commits failed. If this is persistent it may help to increase partial-progress.max-commits which will break the rewrite operation into smaller commits.```



### How to reproduce it

Run compaction against an iceberg table. using spark actions. 
iceberg version: 1.6.0
catalog: nessie: 0.99.0

docker

Spark

### Additional information

The nessie catalog had 3 branches. I dropped 2 of the dev branches to ensure there is only 1 branch during comapction.

### LINKED ISSUE #10013: [Bug]: Snapshot 123 is not an ancestor of 456 when running DELETE query on Trino (parallel writes)
### What happened

Every so many seconds we append new data to a Table using pyIceberg (using Arrow tables).

On the background we run some Table maintenance operations; specifically removing old data is important (since we have limited space available). We do this using Trino:

```
DELETE FROM pack WHERE epoch_timestamp_tz <= timestamp '2024-11-29 13:15' AND timestampns < 1.7328861183132278e+18
```

Now this query takes a few minutes to complete but in the end it fails to create a new snapshot:

```
java.lang.IllegalArgumentException: Snapshot 3398968223100354057 is not an ancestor of 5174469298448934675
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkArgument(Preconditions.java:430)
	at org.apache.iceberg.BaseRowDelta.validate(BaseRowDelta.java:113)
	at org.apache.iceberg.SnapshotProducer.apply(SnapshotProducer.java:235)
	at org.apache.iceberg.SnapshotProducer.lambda$commit$2(SnapshotProducer.java:386)
```

I'm sure it's related to the parallel write operations, because when I stop the ingest activities (Trino is then the only remaining writer); the query succeeds.

I know with use of Nessie you do not have table history (can't see older snapshots) since Nessie maintains these are catalog level instead; i'm wondering if this has anything to do with my finding?

Or are there additional iceberg configuration to set?
Can I extend the query to claim exclusive write access to the Table?

### How to reproduce it

1. Table is created in pyIceberg:
```
catalog.create_table(
                identifier=f"{self.namespace}.{table_name}",
                schema=pt.schema,
                properties={
                    "gc.enabled": True,
})
Also partition specs are added:

update.add_field(
                    source_column_name="epoch_timestamp_tz",
                    transform=HourTransform(),
                    partition_field_name="epoch_hours",
                )
```
2. append Arrow tables every second
3. run trino query which take a few minutes to complete.
4. 

docker: quay.io/projectnessie/nessie:0.100.1

PyIceberg + Trino

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by snazy**
The current behavior of Nessie's Iceberg REST is to return only the most recent Iceberg snapshot. However, this seems to conflict with some Iceberg operations, which are not only maintenance operations, but rather related to "merge on read" / (equality) deletes.

This change changes Nessie's behavior by returning older snapshots from a load-table and update-table operations.

Register-table operation however do not change, because only the latest snapshot is actually imported. The behavior does change by returning an error if the table to be registered has more than 1 snapshots.

Fixes #10013
Fixes #9969

### CODE REVIEW

File: catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java [hunk 1]
```diff
@@ -303,42 +313,133 @@ public CompletionStage<SnapshotResponse> retrieveSnapshot(
...
+                        "Failed to retrieve table-metadata {}",
+                        ((IcebergContent) nextContent).getMetadataLocation());
+                    return null;
+                  });
+      collectorStage =
+          collectorStage.thenCombine(
+              olderSnapStage,
+              (collector, olderSnap) -> {
+                if (olderSnap != null) {
+                  collector.snapshotHistory.put(nextSnapshotId, olderSnap);
```
dimas-b, line 405: I wonder whether we need the `collector` at all. The retrieval stages already know their position in the list and snapshot IDs. I wonder if it might be simpler to just put those stages into an array, do `CompletableFuture.allOf()` and then get all non-null results into the final list.

dimas-b: Alternatively, given that `thenCombine` stages complete in order, we can probably just keep adding `olderSnap` directly to the list of results and pass it down to the next combined stage.

dimas-b (commented):
LGTM :+1: A couple of minor comments below

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -303,42 +313,133 @@ public CompletionStage<SnapshotResponse> retrieveSnapshot(
...
 
-    ContentResponse contentResponse =
-        contentService(apiContext)
-            .getContent(
-                key, reference.name(), reference.hashWithRelativeSpec(), false, requestMeta);
-    Content content = contentResponse.getContent();
+    var contentHistory =
+        treeService(apiContext)
+            .getContentHistory(
+                key, reference.name(), reference.hashWithRelativeSpec(), requestMeta);
```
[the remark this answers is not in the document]
snazy: Sadly, we have to return all the previous snapshots, because we cannot know which ones are needed.

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
snazy: SGTM, updated

snazy:
So, I hope that all tests are happy again.

The changes became more intrusive, but it should all work now.

Had to let the register-table endpoint not accept Nessie catalog URIs anymore, because that would lead to two Nessie tables pointing to the same snapshot (NessieTableSnapshot) - and it's actually not good in general to have different tables work in the same (storage) location. OTOH externally registered tables now get all snapshots imported via register-table. The check is a bit racy, but I suspect, it's okay-ish / better-than-nothing :tm: .

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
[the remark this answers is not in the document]
dimas-b: How will it work with Nessie GC? We never remove snapshots from iceberg metadata, IIRC :thinking:

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -654,10 +652,22 @@ public static NessieTableSnapshot icebergTableSnapshotToNessie(
...
-    for (IcebergSnapshotLogEntry logEntry : iceberg.snapshotLog()) {
-      // TODO ??
-      logEntry.snapshotId();
-      logEntry.timestampMs();
+    for (var icebergSnapshotLogEntry : iceberg.snapshotLog()) {
+      var snapId = icebergSnapshotLogEntry.snapshotId();
+      if (snapId == iceberg.currentSnapshotId()) {
+        continue;
+      }
+      if (previous == null) {
```
dimas-b, line 660: Is `previous` ever non-null? IJ shows all non-test call paths leading to `null` :thinking:

dimas-b: Those calls come from `importIcebergTable`... So, essentially if someone commits via Nessie API, each `NessieTableSnapshot` (after importing) will have duplicate "implicit" snapshot lists. WDYT?

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
```diff
@@ -1599,6 +1626,23 @@ public static void addSnapshot(AddSnapshot u, IcebergTableMetadataUpdateState st
         .icebergManifestListLocation(icebergSnapshot.manifestList())
         .icebergManifestFileLocations(icebergSnapshot.manifests());
 
+    // If another Iceberg snapshot was add in this same update-transaction, memoize the previous
```
dimas-b, line 1629: nit: `add` -> `added`

catalog/service/impl/src/main/java/org/projectnessie/catalog/service/impl/CatalogServiceImpl.java
```diff
@@ -303,42 +328,136 @@ public CompletionStage<SnapshotResponse> retrieveSnapshot(
...
+    return olderSnapsFuture.thenApply(
+        x -> {
+          var history = new ArrayList<ImplicitIcebergSnapshot>();
+          for (Long previousSnapshotId : previousSnapshotIds) {
+            try {
+              var snapFuture = olderSnaps.get(previousSnapshotId);
+              var implicitIcebergSnapshot =
+                  snapFuture != null
+                      ? implicitIcebergSnapshot(
+                          (NessieTableSnapshot) snapFuture.get(0, TimeUnit.SECONDS))
```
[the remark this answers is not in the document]
snazy: Why? It's already finished

catalog/service/common/src/main/java/org/projectnessie/catalog/service/api/CatalogService.java
```diff
@@ -78,10 +76,7 @@ CompletionStage<Stream<SnapshotResponse>> commit(
       ApiContext apiContext)
       throws BaseNessieClientServerException;
 
-  interface CatalogUriResolver {
-    URI icebergSnapshot(
-        Reference effectiveReference, ContentKey key, NessieEntitySnapshot<?> snapshot);
-  }
+  boolean checkIcebergSnapshotPresent(String metadataLocation, long versionId);
```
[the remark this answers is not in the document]
snazy: The property in `IcebergContent` is `versionId`

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
snazy: Yes - unfortunate.

catalog/service/common/src/main/java/org/projectnessie/catalog/service/api/CatalogService.java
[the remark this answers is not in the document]
dimas-b: But the method checks for snapshots and it's not apparent that the `versionId` parameter here should be the snapshot ID.

catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
dimas-b: why keep `previous` if it is always `null` in production call paths?

alekhed:
Is there any chance to release this one soon?

gc/gc-iceberg-inttest/src/intTest/java/org/projectnessie/gc/iceberg/inttest/ITSparkIcebergNessieLocal.java
```diff
@@ -164,12 +164,12 @@ public void roundTripLocalWithDeleteFiles() throws Exception {
...
       DeleteSummary deleteSummary = expire(icebergFiles, liveContentSet, maxFileModificationTime);
-      soft.assertThat(deleteSummary.deleted()).isEqualTo(13L);
+      soft.assertThat(deleteSummary.deleted()).isEqualTo(7L);
       Set<StorageUri> filesAfter = allFiles(icebergFiles);
       Set<StorageUri> removedFiles = new HashSet<>(filesBefore);
       removedFiles.removeAll(filesAfter);
-      // make sure expired delete file (parquet) is also removed from GC.
-      soft.assertThat(removedFiles).anyMatch(u -> u.location().endsWith("-deletes.parquet"));
+      // make sure delete file (parquet) although it has expired, is not removed by GC.
+      soft.assertThat(removedFiles).noneMatch(u -> u.location().endsWith("-deletes.parquet"));
```
dimas-b, line 172: Is this behaviour change related to the purpose of this PR (from the description)?

PR #10142, opened 2024-12-26
Enable Azure Key Vault IT

nagyesta:
- Add Lowkey Vault Client dependency to be able to use custom HTTP client during tests
- Reconfigure SecretClient to use custom HTTP client and accept self-signed cert
- Enable previously disabled test

### Note
Hi,
I have noticed that you wanted to use Lowkey Vault in your ITs and they we disabled due to the SSL issues. I thought I would help with them... 
Please feel free to use this PR if it makes sense or let me know in case you disagree with the approach!
Either way, thank you very much for using Lowkey Vault and feel free to let me know in case further assistance is needed!


dimas-b (approved):
Thanks for your contribution, @nagyesta !

nagyesta:
Thank you both @adutra and @dimas-b !

PR #10154, opened 2024-12-30
OAuth client: add support for custom request parameters

adutra:
For context, this change is motivated by the fact that Auth0 requires an `audience` field to be included in `client_credentials` requests to the token endpoint:

https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow/call-your-api-using-the-client-credentials-flow#request-tokens

The audience value must correspond to a declared "API identifier", and is mandatory.

This deviates from standard OAuth 2.0 specs and have annoyed other client implementors, e.g. [this one](https://github.com/usebruno/bruno/issues/2002). 

Rather than creating a specific `audience` field just for Auth0, I chose instead to open up for arbitrary custom parameters to be passed.

Ironically, Iceberg REST client is not affected by this issue, because if the `audience` field is present, it is (mistakenly) included in all requests (and ignored by the auth server).

### DIFF
--- a/api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
+++ b/api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
@@ -161,6 +161,13 @@ public final class NessieConfigConstants {
    * <pre>{@code
    * nessie.authentication.oauth2.extra-params = "custom_param1=custom_value1,custom_param2=custom_value2"
    * }</pre>
+   *
+   * For example, Auth0 requires the {@code audience} parameter to be set to the API identifier.
+   * This can be done by setting the following configuration:
+   *
+   * <pre>{@code
+   * nessie.authentication.oauth2.extra-params = "audience=https://nessie-catalog/api"
+   * }</pre>
    */
   @ConfigItem(section = "OAuth2 Authentication")
   public static final String CONF_NESSIE_OAUTH2_EXTRA_PARAMS =
--- a/api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
+++ b/api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
@@ -151,6 +151,22 @@ public final class NessieConfigConstants {
       "nessie.authentication.oauth2.client-secret";
 
   /**
+   * Extra parameters to include in each request to the token endpoint. This is useful for custom
+   * parameters that are not covered by the standard OAuth2.0 specification. Optional, defaults to
+   * empty.
+   *
+   * <p>The format of this field is a comma-separated list of key-value pairs, separated by an equal
+   * sign. The values must NOT be URL-encoded. Example:
+   *
+   * <pre>{@code
+   * nessie.authentication.oauth2.extra-params = "custom_param1=custom_value1,custom_param2=custom_value2"
+   * }</pre>
+   */
+  @ConfigItem(section = "OAuth2 Authentication")
+  public static final String CONF_NESSIE_OAUTH2_EXTRA_PARAMS =
+      "nessie.authentication.oauth2.extra-params";
+
+  /**
    * Username to use when authenticating against the OAuth2 server. Required if using OAuth2
    * authentication and "password" grant type, ignored otherwise.
    */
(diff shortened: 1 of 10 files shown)


api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticatorConfig.java
```diff
@@ -143,11 +146,26 @@ static OAuth2AuthenticatorConfig fromConfigSupplier(Function<String, String> con
...
   }
 
+  static Map<String, String> parseExtraParams(String text) {
+    if (text == null || text.isBlank()) {
+      return Map.of();
+    }
+    return Arrays.stream(text.split(","))
+        .map(s -> s.split("=", 2))
+        .collect(
+            Collectors.toMap(a -> a[0].trim(), a -> a.length > 1 ? a[1].trim() : "", (a, b) -> a));
```
dimas-b, line 166: Would it make sense to throw on duplicate keys (default merge function)? The order or application of the merge function is not well-defined anyway, so it will be hard to predict the resultant config map values in case of key collisions... WDYT?

dimas-b (commented):
Nice improvement! :+1: Just one minor comment.

api/client/src/main/java/org/projectnessie/client/auth/oauth2/OAuth2AuthenticatorConfig.java
adutra: Good idea but I had to revert to a classical for loop because the `BinaryOperator` that merges duplicate entries does not have access to the offending key, only to the offending values.

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2ClientConfig.java
```diff
@@ -312,12 +313,12 @@ static Stream<Arguments> testFromConfig() {
             ImmutableMap.of(
                 CONF_NESSIE_OAUTH2_TOKEN_ENDPOINT, "https://example.com/token",
                 CONF_NESSIE_OAUTH2_CLIENT_SECRET, "s3cr3t",
-                CONF_NESSIE_OAUTH2_REFRESH_SAFETY_WINDOW, "PT10S",
```
adutra, line 315: The previous test for empty client ID was actually a test for the `@Check` method and is already covered above (cf. line125), so I removed it.

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
```diff
@@ -150,6 +150,22 @@ public final class NessieConfigConstants {
...
+  /**
+   * Extra parameters to include in each request to the token endpoint. This is useful for custom
+   * parameters that are not covered by the standard OAuth2.0 specification. Optional, defaults to
+   * empty.
+   *
+   * <p>The format of this field is a comma-separated list of key-value pairs, separated by an equal
+   * sign. The values must NOT be URL-encoded. Example:
+   *
+   * <pre>{@code
+   * nessie.authentication.oauth2.extra-params = "custom_param1=custom_value1,custom_param2=custom_value2"
```
snazy, line 162: Should there be an example for the `audience` field, mentioning Auth0?

api/client/src/test/java/org/projectnessie/client/auth/oauth2/TestOAuth2Client.java
```diff
@@ -578,6 +578,24 @@ void testBadRequest() throws Exception {
     }
   }
 
+  @Test
+  void testExtraRequestParameters() throws Exception {
+
+    try (HttpTestServer server = new HttpTestServer(handler(), true)) {
+
```
snazy, line 585: Nit-troll: superfluous new-line :disappear: ;) ;)

snazy (approved):
LGTM!
Maybe just add an example specific for Auth0

dimas-b (approved):

api/client/src/main/java/org/projectnessie/client/NessieConfigConstants.java
dimas-b: Maybe an Auth0 guide in the site docs (not in javadoc)?

adutra: In this case I think we need to extract the authentication stuff out of the "Client Configuration" page and create a dedicated "Client Authentication" page, wdyt? Then we could add instructions for Keycloak, Auth0, etc.

dimas-b: I think having a dedicated "Client Authentication" is great. I think it would make it easier for end users to find relevant examples. From Zulip feedback it feels like existing docs are a bit too "technical" :)

adutra: Okay, I'm adding that to my todo list.

PR #10264, opened 2025-01-23
Catalog: Allow passing -1 for new schema/partition-spec/sort-order

snazy:
Nessie re-assigns IDs for new schemas/partition-specs/sort-orders. The check that the provided ID for those must be valid (>= 0) is therefore superfluous, it can actually unnecessarily lead to problems.

This change also fixes an issue that the last-added schema/spec/sort ID is set to -1, if the schema/spec/sort already existed. This lets the set-current-schema/set-default-partition-spec/set-default-sort-order updates with `-1` for the last-added one fail, but it should return the ID of the schema/spec/sort ID that already existed.

Also fixing the related checks in IcebergMetadataUpdate.

Added tests for the relevant add*-updates to TestNessieModelIceberg.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ as necessary. Empty sections will not end in the release notes.
 
 ### Fixes
 
+- Nessie re-assigns IDs for new schemas/partition-specs/sort-orders. The check that the provided ID for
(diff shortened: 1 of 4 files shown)


dimas-b (approved):

snazy:
Thanks @dimas-b !

PR #10292, opened 2025-01-29
REST Catalog: Apply warehouse property overrides to tables

dimas-b:
* Add two new warehouse config flags to control the propagation of config overrides from the warehouse level to table level.

* By default old logic applies, which means warehouse config overrides do not affect table properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+- Catalog: Warehouse configuration overrides can now be applied to tables.
+
 ### Changes
 
 ### Deprecations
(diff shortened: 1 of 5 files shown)


bigluck:
Thanks @dimas-b 
For context: https://project-nessie.zulipchat.com/#narrow/channel/371187-general/topic/How.20to.20use.20PyarrowFileIO.20on.20pyiceberg.3F

snazy (commented):
Thanks for the PR, I see the issue here.

I guess it's better to instead have such properties on AdlsFileSystemOptions/GcsBucketOptions/S3BucketOptions - i suspect, having those for just the config-defaults (and not overrides) should be sufficient. Aka via the `ObjectIO.configureIcebergTable` implementations.

dimas-b:
Closing in favour of #10296

PR #10296, opened 2025-01-29
Catalog: Add table config overrides to bucket configuration

dimas-b:
* Add a new property map to `BucketOptions` to allow users to configure Iceberg table config overrides or provide custom table-level properties to Catalog clients.

### DIFF
--- a/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/BucketOptions.java
+++ b/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/BucketOptions.java
@@ -15,4 +15,10 @@
  */
 package org.projectnessie.catalog.files.config;
 
-public interface BucketOptions {}
+import java.util.Map;
+
+public interface BucketOptions {
+
+  /** Iceberg table configuration overrides for all tables stored in this bucket. */
+  Map<String, String> tableConfigOverrides();
+}
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
@@ -144,9 +144,7 @@ public class AdlsObjectIO implements ObjectIO {
     }
 
     icebergConfigDefaults(config);
-    AdlsNamedFileSystemOptions fileSystemOptions = icebergConfigOverrides(storageLocations, config);
-
-    fileSystemOptions.tableConfigOverrides().forEach(config);
+    icebergConfigOverrides(storageLocations, config);
   }
 
   @Override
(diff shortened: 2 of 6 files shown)


catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/BucketOptions.java
```diff
@@ -15,4 +15,10 @@
  */
 package org.projectnessie.catalog.files.config;
 
-public interface BucketOptions {}
+import java.util.Map;
+
+public interface BucketOptions {
```
snazy, line 20: I wanted to propose to add this base interface - but seems it already there ;)

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
```diff
@@ -144,7 +144,9 @@ public void configureIcebergTable(
     }
 
     icebergConfigDefaults(config);
-    icebergConfigOverrides(storageLocations, config);
+    AdlsNamedFileSystemOptions fileSystemOptions = icebergConfigOverrides(storageLocations, config);
+
+    fileSystemOptions.tableConfigOverrides().forEach(config);
```
snazy, line 149: Could be done in `icebergConfigOverrides`

catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/BucketOptions.java
```diff
@@ -15,4 +15,10 @@
...
+
+  /** Iceberg table configuration overrides for all tables stored in this bucket. */
+  Map<String, String> tableConfigOverrides();
```
snazy, line 23: Can you mention this new property in the CHANGELOG.md?

snazy (commented):
Overall LGTM - minor comments

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/adls/AdlsObjectIO.java
dimas-b: good point

dimas-b: fixed

catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/BucketOptions.java
snazy: At least the ones for for the default ones - i suspect not many users have bucket-specific configs.

snazy (approved):

PR #10304, opened 2025-01-30
Site: Improve the `Nessie Spark SQL Extensions` page

snazy:
We get a bunch of user questions about the use of Nessie Spark SQL extensions. The version of the Nessie Spark SQL extensions _must_ match the Nessie version used in Iceberg.

This change updates the `Nessie Spark SQL Extensions` page to give users a choice of the Spark version, the Iceberg version and the Scala version. The change builds upon the already existing macros, but now considers the actual versions. Those versions are pulled in from various sources: Iceberg releases via GitHub API and then the Nessie version grabbed from Iceberg's `gradle/libs.versions.toml` file, the content of the file `integrations/spark-scala.properties`.


snazy:
This is how that page looks with this change:
image

dimas-b (approved):

adutra (approved):
LGTM. I'm just a little worried that the large number of tabs may not render well on small devices, but nothing that should prevent us from merging this.

snazy:
> LGTM. I'm just a little worried that the large number of tabs may not render well on small devices, but nothing that should prevent us from merging this.

Yea - I just couldn't come up with a better way w/o doing our own HTML/CSS/JS magic (which nobody should let me do)

PR #10331, opened 2025-02-03
Include NOTICE+LICENSE in every jar

snazy:
Adds the contents of `LICENSE` and `NOTICE` to every jar using the following patterns:
* `META-INF/licenses/<maven-group-id>/<maven-artifact-id>-<version>/LICENSE`
* `META-INF/licenses/<maven-group-id>/<maven-artifact-id>-<version>/NOTICE`

Also explicitly clarifies where dependencies only licensed under "GNU General Public License, Version 2 with the GNU Classpath Exception" are used (included in binary executable and containers).


snazy:
Review note: run `./gradlew pTML -PjarWithGitInfo`

NOTICE
```diff
@@ -34,15 +34,15 @@ Copyright The Apache Software Foundation
 Netty
 Copyright The Netty Project
 
-This product includes additional software licensed under the terms
-of the following licenses:
+This product may include or have transitive dependencies to additional software
```
dimas-b, line 37: IMHO `may include` is too weak. Downstream projects are required to copy this NOTICE. I think it has to be exact.

snazy: It's a warning to look at the transitive dependencies as well.

LICENSE-BINARY-DIST
```diff
@@ -1,5 +1,5 @@
-LICENSE applicable to Nessie binary distributions
--------------------------------------------------
+LICENSE applicable to Nessie executables and containers
```
jbonofre, line 1: Does it apply to `nessie-client` and related (integration tests) ?

NOTICE
jbonofre: As this `NOTICE` is about source distribution (not binary distribution), we should only consider code copied from other Apache licensed projects.
So, it means the `NOTICE` should contains `NOTICE` from:
* Netty
* Apache Polaris
* Apache Iceberg
NB: Gradle doesn't provide any `NOTICE`.

It's not required to mention something.

LICENSE-BINARY-DIST
jbonofre: Imho, it's pretty hard to have `LICENSE` and `NOTICE` that can apply to all distributed binary artifacts. I would propose to have `LICENSE` and `NOTICE` for each distributed binary artifacts, specific for each artifact.

NOTICE-BINARY-DIST
```diff
@@ -1,5 +1,5 @@
-NOTICE applicable to Nessie binary distributions
-------------------------------------------------
+NOTICE applicable to Nessie executables and containers
```
jbonofre, line 1: Same comment here as `LICENSE-BINARY-DIST`.

LICENSE-BINARY-DIST
jbonofre: It seems that we already ship correct `LICENSE`/`NOTICE`. I'm doing a full pass.

jbonofre (commented):
It's a way better change. For clarity, I would rename the `LICENSE`/`NOTICE` specific to docker (and runner fat jar). However:
1. in `NOTICE`, I would remove the list of license for transitive deps. This `NOTICE` file is for source distribution, so dependencies don't matter here. Basically, I would remove this part:
```
This product has transitive dependencies to additional software licensed
under the terms of the following licenses:
* Apache Software License, Version 2.0
* Creative Commons 1.0 Universal
* Eclipse Distribution License, Version 1.0
* Eclipse Public License, Version 1.0
* Eclipse Public License, Version 2.0

See also the LICENSE file.

The Nessie web site provides aggregated license reports for recent Nessie
binary distributions, available at https://projectnessie.org/docs/ and
https://projectnessie.org/downloads/.

This product includes software developed at
The Apache Software Foundation (http://www.apache.org/).
```
2. The `LICENSE-BINARY-DIST`/`NOTICE-BINARY-DIST` is only valid for container images (not sure about the runner fat jar). It's not correct for jar files. I would rename to be explicit like `LICENSE-CONTAINER-DIST`.
3. The distributed nessie jar files should include a clean `META-INF/LICENSE`/`META-INF/NOTICE`. Currently, the nessie jar files don't include `LICENSE`/`NOTICE`.

CHANGELOG.md
```diff
@@ -22,6 +22,11 @@ as necessary. Empty sections will not end in the release notes.
 
 ### Commits
 
+### Note
+
+- This release has no code changes.
+- NOTICE and LICENSE files clarifications, included in jars published to Maven Central.
```
dimas-b, line 28: nit: `included in published jars`... technically, they are included inside the jars, inside the docker images too.

gradle/built-uber-dists/NOTICE-BINARY-DIST
```diff
@@ -0,0 +1,446 @@
+
+Dremio
+Copyright 2015-2025 Dremio Corporation
```
dimas-b, line 3: Should the header be 1:1 with the root NOTICE?

nit: extra black line at the top.

gradle/built-uber-dists/NOTICE-BINARY-DIST
```diff
@@ -0,0 +1,446 @@
...
+The Nessie web site provides aggregated license reports for recent Nessie
+binary distributions, available at https://projectnessie.org/docs/ and
+https://projectnessie.org/downloads/.
+
+
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+---
```
dimas-b, line 40: Use the same separator and general text style as the root NOTICE?

gradle/built-uber-dists/LICENSE-BINARY-DIST
dimas-b, line 1: `Nessie distributions contain some or all of the following dependencies` (in the unchanged section) -- I believe the language has to be exact. We have to be sure what we include, otherwise how can claim that our license is correct? @jbonofre WDYT?

If we want to use the same text for all binary artifacts (each having a different sub-set of bundled dependendencies) how about this text:

```
Nessie distributions contain only dependencies licensed under the Apache License.
The following is the list of all redistributed dependencies unified across all published
Nessie artifacts. Each Nessie artifact bundles a sub-set of those dependencies.
```

WDYT?

tools/notice/build.gradle.kts
```diff
@@ -30,16 +30,20 @@ val noticeDir = project.layout.buildDirectory.dir("notice")
...
+      rootProject.layout.files(
+        "gradle/built-uber-dists/NOTICE-BINARY-DIST",
+        "gradle/built-uber-dists/LICENSE-BINARY-DIST",
+      )
+    )
     inputs.property("version", project.version)
-    from(rootProject.projectDir) {
-      include("NOTICE-BINARY-DIST", "NOTICE", "LICENSE-BINARY-DIST", "LICENSE")
+    from(rootProject.projectDir.resolve("gradle/built-uber-dists")) {
+      include("NOTICE-BINARY-DIST", "LICENSE-BINARY-DIST")
```
dimas-b, line 42: `nessie-model-0.102.5-SNAPSHOT.jar` does not have a NOTICE. I think we should put it in all jars, right?

integrations/spark-extensions/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ dependencies {
   compileOnly("org.apache.spark:spark-core_${sparkScala.scalaMajorVersion}") { forSpark(sparkScala.sparkVersion) }
   compileOnly("org.apache.spark:spark-hive_${sparkScala.scalaMajorVersion}") { forSpark(sparkScala.sparkVersion) }
   implementation(nessieClientForIceberg())
+  implementation(nessieProject("nessie-notice"))
```
dimas-b, line 39: LICENSE and NOTICE are included in the jar under `/META-INF/resources/`... why not directly under `META-INF` (e.g. Guava has it there)?

dimas-b (commented):
LGTM overall :+1:

integrations/spark-extensions/build.gradle.kts
jbonofre: The "regular" place is `META-INF/LICENSE` and `META-INF/NOTICE` (at least it's what some maven plugins like shade plugin are looking for).

jbonofre:
Which profile should be used to see `META-INF/LICENSE` and `META-INF/NOTICE` included in Nessie jar files ?

tools/notice/build.gradle.kts
jbonofre: I think we have to build with a specific profile. I think we should do that by default.

snazy:
> Which profile should be used to see `META-INF/LICENSE` and `META-INF/NOTICE` included in Nessie jar files ?

https://github.com/projectnessie/nessie/pull/10331#issuecomment-2631899118

integrations/spark-extensions/build.gradle.kts
snazy: > META-INF/LICENSE and META-INF/NOTICE

Pretty bad for uber-jars - everything overwrites everything

tools/notice/build.gradle.kts
snazy: Nope, let's not do unnecessary stuff like that. It's done for releases.

gradle/built-uber-dists/LICENSE-BINARY-DIST
snazy: We need the list for the automatic license check

integrations/spark-extensions/build.gradle.kts
dimas-b: true, but Nessie has a specific binary NOTICE for uber jars, right?

gradle/built-uber-dists/LICENSE-BINARY-DIST
dimas-b: Yes, I'm suggesting clarifying the text as it relates to what is possibly bundled.

tools/notice/build.gradle.kts
dimas-b: It was my mistake with the gradle command. After rebuilding I see LICENSE and NOTICE are included under `META-INF/licenses/org.projectnessie.nessie/nessie-model-0.102.5-SNAPSHOT/`

integrations/spark-extensions/build.gradle.kts
snazy: That's not what LICENSE + NOTICE are for.

gradle/built-uber-dists/LICENSE-BINARY-DIST
snazy: in 9bc6280fc0c73da526832e6c88819d0d94ffd240

CHANGELOG.md
snazy: This mention is (only) about /NOTICE + /LICENSE

gradle/built-uber-dists/NOTICE-BINARY-DIST
snazy: i'll leave something to follow-up :P

integrations/spark-extensions/build.gradle.kts
snazy: The "binary uber/container/..." distributables get theirs in /META-INF/resources/LICENSE.txt (and NOTICE.txt)

dimas-b (approved):

jbonofre (approved):
LGTM, thanks !

PR #10443, opened 2025-02-21
Add `copy` command to the ContentGenerator tool

dimas-b:
This command is intended for replicating a "real" content object under different content keys for testing purposes.

### DIFF
--- a/tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/CopyContent.java
+++ b/tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/CopyContent.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright (C) 2020 Dremio
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/CopyContent.java
```diff
@@ -0,0 +1,94 @@
...
+  @Option(
+      names = {"-r", "--ref", "--branch"},
+      description = "Branch name for making changes (defaults to the default branch if not set).")
+  private String ref;
+
+  @Min(value = 1, message = "At least one copy is to be made.")
+  @Option(
+      names = {"-n", "--num-copies"},
+      required = true,
+      defaultValue = "10000",
```
adutra, line 42: isn't this a bit extreme for a default value?

adutra (approved):

tools/content-generator/src/main/java/org/projectnessie/tools/contentgenerator/cli/CopyContent.java
dimas-b: This tools is not really meaningful aside from stress tests, so I'd say it's right on target ;)

PR #10464, opened 2025-02-25
nessie-gc truncate table to avoid increasing storage of DB

BerkanDemirtas:
In case of using relational database for nessie-gc, nessie-gc keeps data of older runs inside tables and after some runs that cause storage problem(which is caused by data and index storage). 
That truncate step will help us to delete these data from these tables:

-          "gc_live_sets"
-          "gc_live_set_contents"
-         "gc_live_set_content_locations"
-         "gc_file_deletions"


dimas-b:
@BerkanDemirtas Thanks for your contribution! Unfortunately it looks like the author of the commit is not known to GH, which causes the CLA issue noted above. Would you mind amending your commit to set the author according to you GH profile and then sign the CLA? Thx!

PR #10510, opened 2025-03-05
Helm: Allow annotations on the configmap. 

BenCoughlan15:
This pull request introduces enhancements to the Helm chart for Nessie by adding support for custom annotations in the ConfigMap. The changes ensure that users can define additional annotations through the `values.yaml` file, which will be applied to the ConfigMap.

This would allow us to set the argocd sync-wave, which ensures the CM syncs first, leading to a more stable Nessie deployment. 
i.e. 

```
  annotations:
    argocd.argoproj.io/sync-wave: "-1"
```

There may of course, be other use cases. 

Key changes include:

* [`helm/nessie/templates/configmap.yaml`](diffhunk://#diff-4dd7692ce974f0d9968339f65a8aad2db5dd16da0012dee981c6222566ca4191R29-R32): Added support for custom annotations by including a conditional block to insert annotations if they are defined in the `values.yaml` file.
* [`helm/nessie/values.yaml`](diffhunk://#diff-2ec11c0c74822e15c1da4fdd05297e86c7ebecfe975c5c1173454565571b9534R743-R745): Introduced a new field `configMapAnnotations` to allow users to specify additional annotations for the ConfigMap.

### DIFF
--- a/helm/nessie/templates/configmap.yaml
+++ b/helm/nessie/templates/configmap.yaml
@@ -26,6 +26,10 @@ metadata:
     {{- if .Values.configMapLabels }}
     {{- toYaml .Values.configMapLabels | nindent 4 }}
     {{- end }}
+  annotations:
+    {{- if .Values.configMapAnnotations }}
+    {{- toYaml .Values.configMapAnnotations | nindent 4 }}
+    {{- end }}
 data:
   application.properties: |-
     {{- $map := dict -}}
--- a/helm/nessie/templates/configmap.yaml
+++ b/helm/nessie/templates/configmap.yaml
@@ -26,10 +26,10 @@ metadata:
     {{- if .Values.configMapLabels }}
     {{- toYaml .Values.configMapLabels | nindent 4 }}
     {{- end }}
+  {{- if .Values.configMapAnnotations }}
   annotations:
-    {{- if .Values.configMapAnnotations }}
-    {{- toYaml .Values.configMapAnnotations | nindent 4 }}
-    {{- end }}
+    {{- tpl (toYaml .Values.configMapAnnotations) . | nindent 4 }}
+  {{- end }}
 data:
   application.properties: |-
     {{- $map := dict -}}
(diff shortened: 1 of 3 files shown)


dimas-b (approved):
Thanks for your contribution, @BenCoughlan15 !

helm/nessie/templates/configmap.yaml
```diff
@@ -26,6 +26,10 @@ metadata:
     {{- if .Values.configMapLabels }}
     {{- toYaml .Values.configMapLabels | nindent 4 }}
     {{- end }}
+  annotations:
+    {{- if .Values.configMapAnnotations }}
+    {{- toYaml .Values.configMapAnnotations | nindent 4 }}
+    {{- end }}
```
adutra, line 32: Two minor suggestions:

1. Put the if block start before `annotations:`
2. Use `tpl` to templetize the input (I know we don't do this for `configMapLabels` line 27 – but maybe we should, in fact)

```suggestion
  {{- if .Values.configMapAnnotations }}
  annotations:
    {{- tpl (toYaml .Values.configMapAnnotations) . | nindent 4 }}
  {{- end }}
```

adutra (commented):
LGTM with some minor remarks.

Also could you please create a new test file `configmap_test.yaml` and add tests like the ones below:

```yaml
release:
  name: nessie
  namespace: nessie-ns

templates:
  - configmap.yaml

tests:
  - it: should not create annotations by default
    asserts:
      - notExists:
          path: metadata.annotations
  - it: should create annotations if defined
    set:
      configMapAnnotations:
        foo: bar
    asserts:
    - equal:
        path: metadata.annotations
        value:
          foo: bar
  - it: should create annotations with templates
    set:
      configMapAnnotations:
        foo: configmap-{{ .Release.Name }}-{{ .Release.Namespace }}
    asserts:
      - equal:
          path: metadata.annotations
          value:
            foo: configmap-nessie-nessie-ns
```

helm/nessie/templates/configmap.yaml
snazy: @adutra wanna just commit the suggestion and then merge?

adutra (approved):

PR #10526, opened 2025-03-10
Make `cache-enable-soft-references` default to `false`

dimas-b:
Following up on #10217.

Using soft references in the Nessie cache may cause heavy GC overhead in some environments. While it may be used for fine-tuning Nessie caches to improve performance, enabling it by default seems risky for less advanced users and smaller environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ as necessary. Empty sections will not end in the release notes.
 
 ### Highlights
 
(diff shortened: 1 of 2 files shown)


bigluck:
x issue message: I had to disable this configuration too, our server went crazy.

Screenshot 2025-03-13 at 18 14 48

dimas-b:
> Should we revert to the pre-soft-ref defaults for the cache-sizing as well?

I do not mind, but let's merge this one first :)

snazy (approved):

PR #10582, opened 2025-03-21
Make S3 signing URL validity configurable

snazy:
S3 signing URLs communicated to Iceberg REST clients currently have a hard coded validity of 3 hours. This can be too long for certain use cases. Since there's no way to instruct Iceberg clients to refresh the signing-URL, this change introduces a configuration (for all or just some buckets) to change that duration.


adutra:
FYI https://github.com/apache/iceberg/issues/12602

PR #10591, opened 2025-03-22
Simplify Lowkey Vault configuration

nagyesta:
- Use client factory to create SecretClient


dimas-b:
@nagyesta : Thanks for your contribution! However, I'm not sure I understand the purpose of this change... Would you mind explaining it a bit?

nagyesta:
Hi @dimas-b ,
sure! Ultimately this is not changing anything, just a simplification using the new `getClientFactory()` convenience method. The new code does almost exactly the same thing as the many lines you had there before but you don't need to worry about the internal details anymore and it will keep working just fine even in case of future changes.

snazy (approved):
Thanks @nagyesta!
The changes look reasonable and are a nice simplification.

nagyesta:
Thank you @snazy !

PR #10629, opened 2025-04-01
Introduce a hard capacity limit for the objects cache

snazy:
Implements a hard limit to prevent the cache's retained heap size to grow "infinitely" in case cache eviction cannot keep up with newly added entries. The default is to allow 10% on top of the cache-capacity.

The "overshoot" is necessary to eventually trigger the cache eviction.

Also adds the recommended `Scheduler.systemScheduler`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,7 @@ as necessary. Empty sections will not end in the release notes.
 
 * Introduces a hard objects-cache capacity limit to ensure that the cache does never consume more than
   the configured cache-capacity plus a configurable "overshoot" (defaults to 10%). New cache entries are
-  admitted as long as the current cache size is less than the "cache-capacity + overshoot". The
-  "overshoot" is necessary to eventually trigger the cache eviction.
+  admitted as long as the current cache size is less than the "cache-capacity + overshoot".
 
 ### Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+* Introduces a hard objects-cache capacity limit to ensure that the cache does never consume more than
+  the configured cache-capacity plus a configurable "overshoot" (defaults to 10%). New cache entries are
+  admitted as long as the current cache size is less than the "cache-capacity + overshoot". The
+  "overshoot" is necessary to eventually trigger the cache eviction.
+
 ### Changes
 
 * Nessie's REST API endpoints now accept "truncated timestamps" in relative-commit-specs, aka without the
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3IamPolicies.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3IamPolicies.java
@@ -25,6 +25,7 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
+import com.github.benmanes.caffeine.cache.Scheduler;
 import java.io.IOException;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
@@ -139,6 +140,7 @@ final class S3IamPolicies {
         Caffeine.newBuilder()
             .maximumSize(2000)
             .expireAfterAccess(Duration.of(1, ChronoUnit.HOURS))
+            .scheduler(Scheduler.systemScheduler())
             .build(ParsedIamStatements::parseStatement);
 
     private static ObjectNode parseStatement(String stmt) {
--- a/catalog/secrets/cache/src/main/java/org/projectnessie/catalog/secrets/cache/CachingSecretsBackend.java
+++ b/catalog/secrets/cache/src/main/java/org/projectnessie/catalog/secrets/cache/CachingSecretsBackend.java
@@ -16,19 +16,21 @@
 package org.projectnessie.catalog.secrets.cache;
 
 import static java.util.Collections.singletonList;
+import static java.util.Objects.requireNonNull;
 
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.Expiry;
+import com.github.benmanes.caffeine.cache.Scheduler;
 import com.google.common.annotations.VisibleForTesting;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
 import java.net.URI;
+import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.concurrent.TimeUnit;
-import java.util.function.LongSupplier;
 import org.projectnessie.catalog.secrets.Secret;
 import org.projectnessie.catalog.secrets.SecretType;
 import org.projectnessie.catalog.secrets.SecretsProvider;
@@ -38,51 +40,21 @@ public class CachingSecretsBackend {
   private static final Secret CACHE_NEGATIVE_SENTINEL = Map::of;
 
   public static final String CACHE_NAME = "nessie-secrets";
-  private static final long NOT_CACHED = 0L;
 
   @VisibleForTesting final Cache<CacheKeyValue, Secret> cache;
   private final long ttlNanos;
-  private final LongSupplier clock;
 
   public CachingSecretsBackend(SecretsCacheConfig config) {
     OptionalLong ttl = config.ttlMillis();
     this.ttlNanos = ttl.isPresent() ? TimeUnit.MILLISECONDS.toNanos(ttl.getAsLong()) : 0L;
-    this.clock = config.clockNanos();
 
     Caffeine<CacheKeyValue, Secret> cacheBuilder =
         Caffeine.newBuilder()
             .expireAfter(
-                new Expiry<CacheKeyValue, Secret>() {
-                  @Override
-                  public long expireAfterCreate(
-                      CacheKeyValue key, Secret value, long currentTimeNanos) {
-                    long expire = key.expiresAtNanosEpoch;
-                    if (expire == NOT_CACHED) {
-                      return 0L;
-                    }
-                    long remaining = expire - currentTimeNanos;
-                    return Math.max(0L, remaining);
-                  }
-
-                  @Override
-                  public long expireAfterUpdate(
-                      CacheKeyValue key,
-                      Secret value,
-                      long currentTimeNanos,
-                      long currentDurationNanos) {
-                    return currentDurationNanos;
-                  }
-
-                  @Override
-                  public long expireAfterRead(
-                      CacheKeyValue key,
-                      Secret value,
-                      long currentTimeNanos,
-                      long currentDurationNanos) {
-                    return currentDurationNanos;
-                  }
-                })
-            .ticker(clock::getAsLong)
+                Expiry.<CacheKeyValue, Secret>creating(
+                    (key, value) -> Duration.ofNanos(key.expireAfterNanos)))
+            .scheduler(Scheduler.systemScheduler())
+            .ticker(config.clockNanos()::getAsLong)
             .maximumSize(config.maxElements());
     config
         .meterRegistry()
@@ -110,10 +82,7 @@ public class CachingSecretsBackend {
       URI name,
       SecretType secretType,
       Class<S> secretJavaType) {
-    long ttl = ttlNanos;
-    long expires = ttl != 0L ? clock.getAsLong() + ttl : 0L;
-
-    CacheKeyValue key = new CacheKeyValue(repositoryId, name, expires);
+    CacheKeyValue key = new CacheKeyValue(repositoryId, name, ttlNanos);
 
     Secret fromCache =
         cache.get(
@@ -124,6 +93,9 @@ public class CachingSecretsBackend {
                   (Optional<Secret>) backend.getSecret(name, secretType, secretJavaType);
               return loaded.orElse(CACHE_NEGATIVE_SENTINEL);
             });
+    // cannot be null (just make IDE happy)
+    requireNonNull(fromCache);
+
     if (fromCache == CACHE_NEGATIVE_SENTINEL) {
       return Optional.empty();
     }
@@ -136,13 +108,12 @@ public class CachingSecretsBackend {
     final String repositoryId;
     final String name;
 
-    // Revisit this field before 2262-04-11T23:47:16.854Z (64-bit signed long overflow) ;) ;)
-    final long expiresAtNanosEpoch;
+    final long expireAfterNanos;
 
-    CacheKeyValue(String repositoryId, URI name, long expiresAtNanosEpoch) {
+    CacheKeyValue(String repositoryId, URI name, long expireAfterNanos) {
       this.repositoryId = repositoryId;
       this.name = name.toString();
-      this.expiresAtNanosEpoch = expiresAtNanosEpoch;
+      this.expireAfterNanos = expireAfterNanos;
     }
 
     @Override
--- a/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
+++ b/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
@@ -56,8 +56,8 @@ class CaffeineCacheBackend implements CacheBackend {
   static final long ONE_MB = 1024L * 1024L;
   public static final String METER_CACHE_CAPACITY_MB = "cache_capacity_mb";
   public static final String METER_CACHE_CAPACITY = "cache.capacity";
+  public static final String METER_CACHE_ADMIT_CAPACITY = "cache.capacity.admitted";
   public static final String METER_CACHE_WEIGHT = "cache.weight";
-  public static final String METER_CACHE_REJECTIONS = "cache.rejections";
   public static final String METER_CACHE_REJECTED_WEIGHT = "cache.rejected-weight";
 
   private final CacheConfig config;
@@ -145,13 +145,13 @@ class CaffeineCacheBackend implements CacheBackend {
                       .tag("cache", CACHE_NAME)
                       .baseUnit(BaseUnits.BYTES)
                       .register(reg);
-                  Gauge.builder(METER_CACHE_WEIGHT, "", x -> (double) currentWeightReported())
-                      .description("Current reported weight of the objects cache in bytes.")
+                  Gauge.builder(METER_CACHE_ADMIT_CAPACITY, "", x -> admitWeight)
+                      .description("Admitted capacity of the objects cache in bytes.")
                       .tag("cache", CACHE_NAME)
                       .baseUnit(BaseUnits.BYTES)
                       .register(reg);
-                  Gauge.builder(METER_CACHE_REJECTIONS, "", x -> rejections.get())
-                      .description("Number of cache-puts that have been rejected.")
+                  Gauge.builder(METER_CACHE_WEIGHT, "", x -> (double) currentWeightReported())
+                      .description("Current reported weight of the objects cache in bytes.")
                       .tag("cache", CACHE_NAME)
                       .baseUnit(BaseUnits.BYTES)
                       .register(reg);
--- a/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
+++ b/versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
@@ -16,7 +16,6 @@
 package org.projectnessie.versioned.storage.cache;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.Collections.singletonList;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static org.projectnessie.versioned.storage.common.persist.ObjType.CACHE_UNLIMITED;
@@ -28,11 +27,18 @@ import static org.projectnessie.versioned.storage.serialize.ProtoSerialization.s
 import com.github.benmanes.caffeine.cache.Cache;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.Expiry;
-import io.micrometer.core.instrument.Tag;
+import com.github.benmanes.caffeine.cache.Scheduler;
+import com.google.common.annotations.VisibleForTesting;
+import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.binder.BaseUnits;
 import io.micrometer.core.instrument.binder.cache.CaffeineStatsCounter;
 import jakarta.annotation.Nonnull;
 import java.lang.ref.SoftReference;
 import java.time.Duration;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.IntConsumer;
+import java.util.function.LongSupplier;
 import org.projectnessie.versioned.storage.common.exceptions.ObjTooLargeException;
 import org.projectnessie.versioned.storage.common.persist.Obj;
 import org.projectnessie.versioned.storage.common.persist.ObjId;
@@ -47,12 +53,23 @@ class CaffeineCacheBackend implements CacheBackend {
   private static final CacheKeyValue NON_EXISTING_SENTINEL =
       new CacheKeyValue("x", ObjId.EMPTY_OBJ_ID, 0L, new byte[0], null, false);
 
+  static final long ONE_MB = 1024L * 1024L;
+  public static final String METER_CACHE_CAPACITY_MB = "cache_capacity_mb";
+  public static final String METER_CACHE_CAPACITY = "cache.capacity";
+  public static final String METER_CACHE_WEIGHT = "cache.weight";
+  public static final String METER_CACHE_REJECTIONS = "cache.rejections";
+  public static final String METER_CACHE_REJECTED_WEIGHT = "cache.rejected-weight";
+
   private final CacheConfig config;
   final Cache<CacheKeyValue, CacheKeyValue> cache;
 
   private final long refCacheTtlNanos;
   private final long refCacheNegativeTtlNanos;
   private final boolean enableSoftReferences;
+  private final long admitWeight;
+  private final AtomicLong rejections = new AtomicLong();
+  private final IntConsumer rejectionsWeight;
+  private final LongSupplier weightSupplier;
 
   CaffeineCacheBackend(CacheConfig config) {
     this.config = config;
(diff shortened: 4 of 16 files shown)


catalog/secrets/cache/src/main/java/org/projectnessie/catalog/secrets/cache/CachingSecretsBackend.java
```diff
@@ -51,6 +52,7 @@ public CachingSecretsBackend(SecretsCacheConfig config) {
 
     Caffeine<CacheKeyValue, Secret> cacheBuilder =
         Caffeine.newBuilder()
+            .scheduler(Scheduler.systemScheduler())
             .expireAfter(
                 new Expiry<CacheKeyValue, Secret>() {
```
ben-manes, line 57: maybe switch to `Expiry.creating(...)`?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -63,6 +64,7 @@ class CaffeineCacheBackend implements CacheBackend {
 
     Caffeine<CacheKeyValue, CacheKeyValue> cacheBuilder =
         Caffeine.newBuilder()
+            .scheduler(Scheduler.systemScheduler())
             .maximumWeight(config.capacityMb() * 1024L * 1024L)
             .weigher(this::weigher)
             .expireAfter(
```
ben-manes, line 70: ditto?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/StsClientsPool.java
```diff
@@ -58,6 +59,7 @@ public StsClientsPool(
     this.clientBuilder = clientBuilder;
     this.clients =
         Caffeine.newBuilder()
+            .scheduler(Scheduler.systemScheduler())
```
ben-manes, line 62: (ignored as no expiration, harmless)

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: We need `expireAfterCreate` and ``expireAfterUpdate` here.
And would require a bigger change, as all supplying function are defined to provide the expiration timestamp.

catalog/secrets/cache/src/main/java/org/projectnessie/catalog/secrets/cache/CachingSecretsBackend.java
snazy: That's a nice proposal!
It simplifies the code.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/StsClientsPool.java
dimas-b: @snazy : shall we revert this change then (for simplicity)?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
ben-manes: Oops, so `Expiry.writing`?

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3IamPolicies.java
```diff
@@ -139,6 +140,7 @@ private static class ParsedIamStatements {
         Caffeine.newBuilder()
             .maximumSize(2000)
             .expireAfterAccess(Duration.of(1, ChronoUnit.HOURS))
+            .scheduler(Scheduler.systemScheduler())
```
dimas-b, line 143: What benefit are we expecting from this change?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -63,13 +64,16 @@ class CaffeineCacheBackend implements CacheBackend {
 
     Caffeine<CacheKeyValue, CacheKeyValue> cacheBuilder =
         Caffeine.newBuilder()
+            .scheduler(Scheduler.systemScheduler())
```
dimas-b, line 67: What benefit are we expecting from this change?

dimas-b: I'd like to formally test the effects of this change :)

snazy:
@ben-manes Some problem is that the cache's retained total heap size can grow up to OOMs in case eviction cannot keep up. WDYT of having an optional way to not admit any new cache entries, when a bounded cache's capacity would be exceeded?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: Test included

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -110,18 +127,35 @@ public long expireAfterRead(
                   singletonList(Tag.of("cache", CACHE_NAME)),
                   "",
                   x -> config.capacityMb());
+              meterRegistry.gauge(
+                  "current_weight",
```
dimas-b, line 131: Could you make this `current_weight_mb` to be comparable to `cache_capacity_mb`?

Also, how about renaming to `cache_weight_mb`?.. "current" is only at the time of measurement :)

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -110,18 +127,35 @@ public long expireAfterRead(
...
+                  singletonList(Tag.of("cache", CACHE_NAME)),
+                  "",
+                  x -> currentWeight());
+              meterRegistry.gauge(
+                  "rejections", singletonList(Tag.of("cache", CACHE_NAME)), "", x -> rejections());
```
dimas-b, line 136: Could we rename to `cache_rejections` to simplify finding it in dashboards?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -145,6 +179,17 @@ public void put(@Nonnull String repositoryId, @Nonnull Obj obj) {
     putLocal(repositoryId, obj);
   }
 
+  @VisibleForTesting
+  void cachePut(CacheKeyValue key, CacheKeyValue value) {
+    var w = weigher(key, value);
+    if (currentWeight.get() + w < admitWeight) {
+      currentWeight.addAndGet(w);
+      cache.put(key, value);
```
dimas-b, line 187: Maybe move the add after `put` (in case `put` happens to fail)?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -145,6 +179,17 @@ public void put(@Nonnull String repositoryId, @Nonnull Obj obj) {
...
+    } else {
+      rejections.incrementAndGet();
```
dimas-b, line 189: How about also tracking the weight of rejected objects? Cf. `CaffeineStatsCounter.recordEviction(...)`

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -145,6 +179,17 @@ public void put(@Nonnull String repositoryId, @Nonnull Obj obj) {
     putLocal(repositoryId, obj);
   }
 
+  @VisibleForTesting
+  void cachePut(CacheKeyValue key, CacheKeyValue value) {
+    var w = weigher(key, value);
+    if (currentWeight.get() + w < admitWeight) {
+      currentWeight.addAndGet(w);
```
dimas-b, line 186: I wonder if we could leverage the cache's [Policy.Eviction.weightedSize()](https://github.com/projectnessie/nessie/pull/10640/files#diff-89933869cb93756b3343911334c3f3455d7a608f052be1997cbcc42c24b13972R119) :thinking: 

That method could cause inline eviction, so more work on `put`, but we'd get the benefit of actually evicting stale objects rather that rejecting fresh objects... WDYT?

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: Can rename to _mb - but i don't think that renaming an existing metric's a good idea ;)

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: Hm - not sure that's really interesting though.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: It's already above the "max weight" bound in such cases.

snazy:
> > Some problem is that the cache's retained total heap size can grow up to OOMs in case eviction cannot keep up. WDYT of having an optional way to not admit any new cache entries, when a bounded cache's capacity would be exceeded?
> 
> To avoid serializing all writes against an single lock for eviction, we decouple the hashmap and eviction policies using a ring buffer. That means most of the time the writer can update the hash table, queue up its work to replay on the policy, schedule the policy to be run, and return to the caller. If the write rate exceeds the eviction rate then the buffer fills up, causing back pressure by writers falling back to acquiring the eviction lock and performing the maintenance work themselves. In those cases the writer threads are descheduled on the lock and the eviction is able to do a batch of work when draining the buffer. The write buffer is sized to a max of `128 * ceilingPowerOfTwo(NCPU)`.
> 
> What often happens is that eviction is stalled not by the write rate but due to the a stalled executor (defaults to `ForkJoinPool.commonPool()`). The eviction itself is fast but deferred to the executor since the entry's removal might be blocked by an in-flight compute in the ConcurrentHashMap (e.g. hashbin collision). If the computes are very slow causing delays then one can use async cache to decouple the hash map from the loading work. More often though stall is due to the commonPool being cpu optimized but swamped by blocking I/O due to being the default in `CompletableFuture`. If that is the case then `Caffeine.executor(Runnable::run)` is often ideal by doing the work on the caller. I think someday we'll switch to virtual threads by default but its only robust in jdk24 which is brand new. Usually the system also needs to switch elsewhere because its abusing common pool which is lowering its overall performance.
> 
> I think we can rely on the write buffer causing back pressure because if the write rate truly exceeds the eviction rate then entry loading is extremely cheap (miss penalty), which makes the cache kind of pointless. Usually the problem is something stalling eviction so it builds up until forced through, so I'd look to see if adjusting your FJP usage or swapping the executor resolves it for you. You might see in your logs the cache's warning when it detects problems: _"The cache is experiencing excessive wait times for acquiring the eviction lock. This may indicate that a long-running computation has halted eviction when trying to remove the victim entry. Consider using AsyncCache to decouple the computation from the map operation."_

I'm a bit blind on the actual root cause, but what might happen in some cases is that the CPUs are all totally busy and eviction then cannot keep up. We don't use async-loading not use loader-function - it's all "dumb" `put(k,v)` and `invalidate(k)` and `getIfPresent`.

I wrote a dumb isolated test with multiple threads (not using FJP) putting objects into a cache. Each thread runs in a tight loop w/ a `Thread.yield` between each `.put()`. Eviction wasn't able to keep up in this case with 8 or more threads running the tight loop (total 16 cores/32 threads). Although the test itself is in no way realistic, I think that's what we've seen happening in a few cases, just due to bursts of excessive load. In reality however, I guess it's a combination of an excessive put-rate _and_ CPUs on the servers being quite busy.

Not sure whether the write-buffer size is a problem there, because from a "PI times thumb" estimation, I doubt that could exceed the total heap size. But I'm not 100% sure on that. Is there a way to tweak the write-buffer size?

From some experiments locally, it seems that '.executor(Runnable::run)` is quite helpful.

I'll dig more into this. Not sure whether we can get rid of the "overshoot thingy", but maybe we can.

snazy:
(I'm now into the "fun" situation to assert behaviors of concurrent things in a unit test... :facepalm:)

ben-manes:
I have a similar tool called [Stresser](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/test/java/com/github/benmanes/caffeine/cache/Stresser.java) ([CI run](https://github.com/ben-manes/caffeine/actions/runs/14190156490/job/39752810956#step:7:287)) which shows some of the internal metrics. CI only runs on a few cores though.

It is really easy to starve FJP since it is the hidden default most everywhere, but it only NUM_CPU threads which makes it pretty awful for I/O work where it is most often used due to being CompletableFuture's default, e.g. `supplyAsync`. It feels like that should be a static analysis warning to nudge on...

snazy:
Oh yea, `FJP.commonPool()` can easily introduce some unexpected gifts - with the usual "works on my machine" non-repro.

From some more testing:
* `Runnable::run` is the better `Executor` - it causes way less "bad" situations - but still the chance isn't 0
* For testing, I'm using `.executor(t -> CompletableFuture.delayedExecutor(2, TimeUnit.MILLISECONDS).execute(t))` to let a test intentionally into the "ugly" situations
* Feels like the combination of `Runnable::run` and the "overshoot protection" together are a good combo.


This is the "tool" I've been initially running to simulate these "ugly" situation (not in Git - just a one-off thingy). Without the protection on `limitCapacity`, it OOMs relatively quickly w/ 7-8 threads (on my machine).
```java
package org.projectnessie.versioned.storage.cache;

import static org.junit.jupiter.params.provider.Arguments.arguments;
import static org.projectnessie.versioned.storage.cache.CaffeineCacheBackend.ARRAY_OVERHEAD;

import com.github.benmanes.caffeine.cache.Caffeine;
import com.github.benmanes.caffeine.cache.Expiry;
import com.github.benmanes.caffeine.cache.Scheduler;
import com.github.benmanes.caffeine.cache.Ticker;
import java.time.Duration;
import java.time.temporal.ChronoUnit;
import java.util.concurrent.Executors;
import java.util.concurrent.ThreadLocalRandom;
import java.util.concurrent.TimeUnit;
import java.util.concurrent.atomic.AtomicBoolean;
import java.util.concurrent.atomic.AtomicLong;
import java.util.stream.Stream;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

public class TestPushHard {
  private static final long ONE_MB = 1024 * 1024;

  @ParameterizedTest
  @MethodSource
  void testPushHard(int numAdders, boolean withScheduler) throws Exception {
    var rt = Runtime.getRuntime();
    var maxMemory = rt.maxMemory();
    var capacity = maxMemory / 2;
    var limitCapacity = capacity * 4 / 3;
    var currentWeight = new AtomicLong();
    var success = new AtomicLong();
    var reject = new AtomicLong();

    var builder =
        Caffeine.newBuilder()
            .maximumWeight(capacity)
            .weigher(TestPushHard::weigh)
            .expireAfter(Expiry.creating((k, v) -> Duration.of(1, ChronoUnit.HOURS)))
            .removalListener(
                (k, v, cause) -> {
                  currentWeight.addAndGet(-weigh(k, v));
                })
            .ticker(Ticker.systemTicker());
    if (withScheduler) {
      builder.scheduler(Scheduler.systemScheduler());
    }

    var cache = builder.build();
    var idGen = new AtomicLong();

    var executor = Executors.newFixedThreadPool(numAdders);
    try {
      var stop = new AtomicBoolean();
      for (int i = 0; i < numAdders; i++) {
        executor.execute(
            () -> {
              while (!stop.get()) {
                byte[] v = new byte[ThreadLocalRandom.current().nextInt(4096, 65536)];
                int weight = weigh(0L, v);
                if (currentWeight.get() + weight < limitCapacity) {
                  cache.put(idGen.incrementAndGet(), v);
                  currentWeight.addAndGet(weight);
                  Thread.yield();
                  success.incrementAndGet();
                } else {
                  reject.incrementAndGet();
                }
              }
            });
      }

      for (long sec = 0; sec < 20; sec++) {
        System.out.printf(
            "weight: %dm ok:%d reject:%d - max: %dm free %dm total %dm%n",
            currentWeight.get() / ONE_MB,
            success.get(),
            reject.get(),
            rt.maxMemory() / ONE_MB,
            rt.freeMemory() / ONE_MB,
            rt.totalMemory() / ONE_MB);
        Thread.sleep(1000L);
      }
      stop.set(true);
    } finally {
      executor.shutdown();
      executor.awaitTermination(1, TimeUnit.MINUTES);
    }
  }

  static int weigh(Long key, byte[] value) {
    return 24 // j.l.Long
        + ARRAY_OVERHEAD
        + value.length;
  }

  static Stream<Arguments> testPushHard() {
    return Stream.of(2, 4, 6, 7, 8, 9, 10)
        .flatMap(num -> Stream.of(arguments(num, true), arguments(num, false)));
  }
}
```

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
```diff
@@ -0,0 +1,114 @@
...
+
+  @RepeatedTest(5)
+  public void testCacheOvershoot() throws Exception {
+    var config =
+        CacheConfig.builder()
+            .capacityMb(4)
+            .cacheCapacityOvershoot(0.1d)
+            // Production uses Runnable::run, but that lets this test sometimes run way too
+            // long, so we introduce some delay to simulate the case that eviction cannot keep up.
+            .executor(t -> delayedExecutor(1, TimeUnit.MILLISECONDS).execute(t))
```
snazy, line 45: Note: this delay let's the test hit the relevant `seen*` assertions pretty quickly.

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
```diff
@@ -0,0 +1,114 @@
...
+import org.assertj.core.api.junit.jupiter.SoftAssertionsExtension;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.projectnessie.versioned.storage.commontests.objtypes.SimpleTestObj;
+
+@ExtendWith(SoftAssertionsExtension.class)
+public class TestCacheOvershoot {
+  @InjectSoftAssertions protected SoftAssertions soft;
+
+  @RepeatedTest(5)
```
snazy, line 37: Ran this one couple 1000 times locally - no failures.

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestDistributedInvalidations.java
```diff
@@ -43,10 +44,10 @@ public class TestDistributedInvalidations {
...
   protected AtomicLong clockNanos;
 
-  protected CacheBackend backend1noSpy;
-  protected CacheBackend backend2noSpy;
-  protected CacheBackend backend1;
-  protected CacheBackend backend2;
+  CaffeineCacheBackend backend1noSpy;
+  CaffeineCacheBackend backend2noSpy;
+  CaffeineCacheBackend backend1;
+  CaffeineCacheBackend backend2;
```
snazy, line 50: Need the concrete type here, otherwise Mockito complains about the new `cachePut()` being seen as an unwanted interaction, so had to add some `.verify()` for those.

ben-manes:
yeah, large memory allocations makes it harder to juggle. Usually its low load times for small results like a cache in front of a db. Regardless the GC has its own required slack for concurrency, so its pretty easy to exhaust all of that with large objects. Sometimes soft references are okay as a safety net to let the GC know it can discard, but those rightfully have a bad reputation. The limit protection seems like a good approach for your usage.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
dimas-b: Caffeine's eviction metrics trace the weight of evicted objects. It's be nice to trace the weight of rejected objects here for symmetry (using the same `DistributionSummary` meter).

It is certainly an interesting metrics because the weight of entries values a lot in runtime.

ben-manes:
I suppose its worth mentioning that striping is how many caches try to solve this since eviction requires a global lock making it single threaded. Then you have N caches able to evict in parallel to better handle concurrent write throughput. This has the negative effect of reducing the max entry size to 1/Nth so larger entries can't be retained. Memcached is smarter by segmenting by object size, rather than hashing the LRU space, so they obtain a nicer balance by using dedicated regions (slab allocator). That makes a ton of sense if one is size aware like your off-heap cache, but lots of oddities if trying to do that for on-heap Java as you know far better than me. Otherwise caches often forgo trying to optimize eviction as reads are the common case, but this can also become unusably slow if not handled thoughtfully (e.g. ehcache's [crazy behavior](https://gist.github.com/ben-manes/6312727adfa2235cb7c5e25cae523ad0)).

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -110,18 +131,68 @@ public long expireAfterRead(
...
+                  singletonList(Tag.of("cache", CACHE_NAME)),
+                  "",
+                  x -> (double) currentWeightTracked() / ONE_MB);
+              meterRegistry.gauge(
+                  "cache_weight_reported_mb",
+                  singletonList(Tag.of("cache", CACHE_NAME)),
+                  "",
+                  x -> (double) currentWeightReported() / ONE_MB);
+              meterRegistry.gauge(
+                  "cache_rejected_weight_mb",
```
dimas-b, line 145: Why not a `DistributionSummary` similar to caffeine's evicted weight metrics?

I'm not sure a "guage" is correct here, because it's a cumulative value.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -145,6 +216,18 @@ public void put(@Nonnull String repositoryId, @Nonnull Obj obj) {
     putLocal(repositoryId, obj);
   }
 
+  @VisibleForTesting
+  void cachePut(CacheKeyValue key, CacheKeyValue value) {
+    var w = weigher(key, value);
+    if (currentWeight.get() + w < admitWeight) {
```
dimas-b, line 222: Suggestion: how about calling `cache.policy().eviction().ifPresent(Policy.Eviction::weightedSize)` when we're about to reject?

This call will perform any delayed evictions, if I'm not mistaken. It will cause overhead on `put`, but only in case of overflow. The benefit would be that we'll get rid of stale entries rather than fresh entries.

My local testing shows that with this approach the cache size is contained with the "overshoot" cap. I can show more details tomorrow, but below is a quick summary. WDYT?

* A few thread put 256K objects into cache
* The limit is 1GB, with 0.1 overshoot
* X is the number of objects inserted
* Y is the increase in heap usage (after `System.gc()`)
* The lower line is the weight limit
* The higher line is the "admit" limit
* The circles represent `main` code
* The dots represent this PR with the proposed change.
* There are no rejections.

image

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: True, a DS seems better here.

However, I also updates the other metrics to report the values as bytes, but also specify the base-unit `bytes` in the metric. AFAIK the base-unit's exposed so the metrics should also "magically" appear correctly.

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
snazy: > Suggestion: how about calling `cache.policy().eviction().ifPresent(Policy.Eviction::weightedSize)` when we're about to reject?

No need to, all cache operations trigger eviction when necessary - and we always perform a getIfPresent before a put.

snazy: So the put-function now looks like this:
```java

  void cachePut(CacheKeyValue key, CacheKeyValue value) {
    var w = weigher(key, value);
    if (weightSupplier.getAsLong() + w < admitWeight) {
      cache.put(key, value);
    } else {
      rejections.incrementAndGet();
      rejectionsWeight.accept(w);
    }
  }
```
where `weightSupplier` yields `policy().eviction().weightedSize()`.

The interesting part now is that I didn't manage to run into the `else` branch, which means that the cache weight stays below the permitted "overshoot" of 10%. It does exceed the configured capacity though.

I think that it's still possible that shredding happens, but it seems very very unlikely. So I'd like to keep the shredding-safety-net.

snazy: Okay, and now I already revert my statement above: I can manage to run into the `else` branch. With 16 threads (no `yield()` in the loop) putting objects into the cache it shreds. That's however a super-stress situation that's IMHO in no way realistic. With the `yield()` it doesn't shred - even with 16 threads (== number of cores of my CPU).

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -83,36 +108,82 @@ public long expireAfterCreate(
...
-                  singletonList(Tag.of("cache", CACHE_NAME)),
-                  "",
-                  x -> config.capacityMb());
-            });
+            .orElse(x -> {});
 
     this.cache = cacheBuilder.build();
+
+    var eviction = cache.policy().eviction().orElseThrow();
+    weightSupplier = () -> eviction.weightedSize().orElse(0L);
```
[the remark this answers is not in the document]
snazy: > This this particular state of the code, we're calling eviction.weightedSize() on every put (and so force eviction on every put).

That's not true. It only evicts when it's required to evict. See `com.github.benmanes.caffeine.cache.BoundedLocalCache.BoundedPolicy.BoundedEviction#weightedSize`

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ as necessary. Empty sections will not end in the release notes.
 
 ### New Features
 
+* Introduces a hard objects-cache capacity limit to ensure that the cache does never consume more than
+  the configured cache-capacity plus a configurable "overshoot" (defaults to 10%). New cache entries are
+  admitted as long as the current cache size is less than the "cache-capacity + overshoot". The
+  "overshoot" is necessary to eventually trigger the cache eviction.
```
dimas-b, line 20: I'm not sure the last sentence is 100% correct. From my POV the overshoot is needed for Nessie code to admit new entries when the cache is full (because eviction will not reduce the cache much lower than the limit). Eviction will happen, of course, but it's not the reason for having the overshoot parameter.

versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+    } finally {
+      stop.set(true);
+
+      executor.shutdown();
+      executor.awaitTermination(10, TimeUnit.MINUTES);
+    }
+
+    soft.assertThat(cache.currentWeightReported()).isLessThanOrEqualTo(admitWeight);
```
dimas-b, line 126: Minor: I believe after all insertions are done, current weight should not exceed the lower limit (no overshoot), because the executor performs eviction tasks immediately.

dimas-b (approved):
LGTM :+1: Thanks @snazy !

versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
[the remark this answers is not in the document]
dimas-b: When the cache is full, it'll always be required to evict, right? It will evict either inside `put` or here... as far as I can tell.

PR #10640, opened 2025-04-03
Add `cache_weight_mb` gauge

dimas-b:
Expose the weighted size of the objects cache as a metric.

Since weigher is designed to approximate actual size of objects on the heap, the metric is defined in MB and can be compared to `cache_capacity_mb` exactly and to heap usage approximately (but with good accuracy).


versioned/storage/cache/src/main/java/org/projectnessie/versioned/storage/cache/CaffeineCacheBackend.java
```diff
@@ -105,16 +108,26 @@ public long expireAfterRead(
...
+              List<Tag> tags = singletonList(Tag.of("cache", CACHE_NAME));
+              meterRegistry.gauge("cache_capacity_mb", tags, "", x -> config.capacityMb());
+              meterRegistry.gauge("cache_weight_mb", tags, "", x -> weightedSizeMb());
             });
 
     this.cache = cacheBuilder.build();
   }
 
+  private double weightedSizeMb() {
+    // Note: calling Policy.Eviction::weightedSize may actually perform evictions.
```
dimas-b, line 120: I suppose evictions in this call may only happen in practice under load, which may

PR #10668, opened 2025-04-09
Cache-config: fix default for "soft references" to `false`

snazy:
This change also adds some missing smallrye-config defaults, no change behavior for the other config options, which are missing [in the docs](https://projectnessie.org/nessie-0-103-3/configuration/#version-store-advanced-settings).

### DIFF
--- a/servers/quarkus-config/src/main/java/org/projectnessie/quarkus/config/QuarkusStoreConfig.java
+++ b/servers/quarkus-config/src/main/java/org/projectnessie/quarkus/config/QuarkusStoreConfig.java
@@ -114,9 +114,6 @@ public interface QuarkusStoreConfig extends StoreConfig {
 
   String CONFIG_CACHE_CAPACITY_MB = "cache-capacity-mb";
 
-  boolean DEFAULT_CONFIG_CACHE_ENABLE_SOFT_REFERENCES = true;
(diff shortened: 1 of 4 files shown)


servers/quarkus-config/src/main/java/org/projectnessie/quarkus/config/QuarkusStoreConfig.java
```diff
@@ -114,9 +114,6 @@ public interface QuarkusStoreConfig extends StoreConfig {
 
   String CONFIG_CACHE_CAPACITY_MB = "cache-capacity-mb";
 
-  boolean DEFAULT_CONFIG_CACHE_ENABLE_SOFT_REFERENCES = true;
```
snazy, line 117: Defaults moved to `StoreConfig` to unify the defaults-constants.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/config/StoreConfig.java
```diff
@@ -73,6 +73,12 @@ public interface StoreConfig {
 
   String CONFIG_REFERENCE_NEGATIVE_CACHE_TTL = "reference-cache-negative-ttl";
 
+  boolean DEFAULT_CONFIG_CACHE_ENABLE_SOFT_REFERENCES = false;
```
dimas-b, line 76: Is it used?

PR #10784, opened 2025-05-07
Migrate to a different Maven publishing plugin

snazy:
Nessie and CEL-java are using the [Gradle Nexus publish plugin](https://github.com/gradle-nexus/publish-plugin), which works great for publishing to Nexus and Sonatype OSSRH. Sonatype OSSRH reaches EOL on June 30, 2025 - a migration to "Maven Central Repository" is necessary. [Sonatype mentions](https://central.sonatype.org/publish/publish-portal-ossrh-staging-api/#configuration) that is's enough to update the configuration of the `gradle-nexus/publish-plugin`. A "canary" release of CEL-Java worked fine, but a Nessie release ran into quite some errors (empty deployment, missing mandatory files). This might be because of the _Portal OSSRH Staging API_, which _is a partial reimplementation of the OSSRH / Nexus Repository Manager 2 Staging APIs_. The noteable differences between CEL-Java and Nessie is that Nessie has way more artifacts and the publishing/deployment takes much longer (< 5 minutes vs 25-30 minutes).

In the mid/long term it is likely anyway the better alternative to use the [Portal Publisher API](https://central.sonatype.org/publish/publish-portal-api/). Althought there are many Gradle plugins around that support that API, none seems to fully fit the needs of multi-project deployments - to publish an aggregated file containing the contents for all modules to be released. Two Gradle plugins looked promising though:

* [GradleUp/nmcp](https://github.com/GradleUp/nmcp) (author also took over the shadow plugin) has support for "aggregated deployments", but in the current form the plugin fails for projects that change the build-directory, which is what the Nessie Spark extensions do.
* [zenhelix/maven-central-publish](https://github.com/zenhelix/maven-central-publish/tree/main) however was chosen as it was easier to integrate, but required some work in our root `build.gradle.kts` integrating the support for aggregated multi-project deployments.

### DIFF
--- a/build-logic/src/main/kotlin/PublishingHelperPlugin.kt
+++ b/build-logic/src/main/kotlin/PublishingHelperPlugin.kt
@@ -78,26 +78,11 @@ constructor(private val softwareComponentFactory: SoftwareComponentFactory) : Pl
               }
 
               tasks.named("generatePomFileForMavenPublication").configure {
-                val e = project.extensions.getByType(PublishingHelperExtension::class.java)
-
                 pom {
-                  name.set(
-                    if (e.mavenName.isPresent) {
-                      e.mavenName.get()
-                    } else {
-                      project.name
-                    }
-                  )
+                  val ep = project.extensions.getByType(PublishingHelperExtension::class.java)
+                  name.set(ep.mavenName.orElse(project.name))
                   description.set(project.description)
                   if (project != rootProject) {
-                    // Add the license to every pom to make it easier for downstream project to
-                    // retrieve the license.
-                    licenses {
-                      license {
-                        name.set("The Apache License, Version 2.0")
-                        url.set("https://www.apache.org/licenses/LICENSE-2.0.txt")
-                      }
-                    }
                     withXml {
                       val projectNode = asNode()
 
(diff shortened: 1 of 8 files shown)


build-logic/src/main/kotlin/PublishingHelperPlugin.kt
```diff
@@ -78,26 +78,11 @@ constructor(private val softwareComponentFactory: SoftwareComponentFactory) : Pl
               }
 
               tasks.named("generatePomFileForMavenPublication").configure {
-                val e = project.extensions.getByType(PublishingHelperExtension::class.java)
```
snazy, line 81: This change adds "all" information to every pom - just to be on the safe side.

gc/gc-iceberg-inttest/build.gradle.kts
```diff
@@ -15,7 +15,7 @@
  */
 
 plugins {
-  id("nessie-conventions-client")
```
snazy, line 18: This module has no production code - so nothing to publish.

gc/gc-tool-inttest/build.gradle.kts
```diff
@@ -15,7 +15,7 @@
  */
 
 plugins {
-  id("nessie-conventions-client")
```
snazy, line 18: This module has no production code - so nothing to publish.

build.gradle.kts
```diff
@@ -53,24 +61,140 @@ tasks.named<Wrapper>("wrapper").configure { distributionType = Wrapper.Distribut
...
-      nexusUrl.set(uri("https://ossrh-staging-api.central.sonatype.com/service/local/"))
-      snapshotRepositoryUrl.set(uri("https://central.sonatype.com/repository/maven-snapshots/"))
+
+  deploymentName = "${project.name}-$version"
+
+  // publishingType
+  //   AUTOMATIC = fully automatic release
+  //   USER_MANAGED = user has to manually publish/drop
+  publishingType =
+    if (System.getenv("CI") != null) PublishingType.AUTOMATIC else PublishingType.USER_MANAGED
```
snazy, line 81: Aka: in CI, release if successful. When run manually, do not publish automatically.

build.gradle.kts
```diff
@@ -53,24 +61,143 @@ tasks.named<Wrapper>("wrapper").configure { distributionType = Wrapper.Distribut
...
+            )
+          }
+        }
+      )
+
+      task.publishingType.set(
+        ext.publishingType.map {
+          when (it) {
+            PublishingType.AUTOMATIC ->
+              io.github.zenhelix.gradle.plugin.client.model.PublishingType.AUTOMATIC
```
dimas-b, line 145: Is this not a no-op?

dimas-b (dismissed):

PR #10836, opened 2025-05-16
Add `public` workaround to some tests

snazy:
See https://github.com/quarkusio/quarkus/issues/47760#issuecomment-2881416779


servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/AbstractQuarkusRest.java
```diff
@@ -28,7 +28,7 @@
 @ExtendWith(QuarkusNessieClientResolver.class)
 public abstract class AbstractQuarkusRest extends BaseTestNessieRest {
 
-  @Inject Instance<Persist> persist;
+  @Inject public Instance<Persist> persist;
```
dimas-b, line 31: Do we have to make fields `public` too? The Quarkus issue appears to be related to class loading :thinking:

snazy: Only in this case. And it's rather a workaround for now. But it's test code - so it doesn't really hurt.

dimas-b (approved):

PR #10859, opened 2025-05-23
fix: migrate aws.region → client.region (Fixes #8262)

### LINKED ISSUE #8262: Cleanup: replace system-property `aws.region` with Iceberg property `client.region`
### Issue description

_No response_

nmshr:
Replaces all internal uses of the system property aws.region with Iceberg’s client.region


snazy:
Hm - this looks like an AI generated PR and the approach does not do what's requested by the issue.

Issue #9991, opened 2024-11-25
[ Docs ]: Missing arguments in Nessie GC documentation for "gc/expire" phase, causes "UnknownHostException" when using Minio as Data Lake.


kishlay-kr:
## Description
There is one important field missing in the documentation for Nessie GC "gc" arguments. This omission causes the expire phase of nessie-gc to fail with the error "**Received an UnknownHostException when attempting to interact with a service**".

## Setup:
1. Nessie version: 0.99.0
2. Trino version: 459
3. Minio version: RELEASE.2024-08-29T01-40-52Z

I am using Minio as the data lake for my local setup. 

## Steps to reproduce:
1. Run expire phase of the nessie-gc tool using minio as the data-lake. ```java -jar nessie-gc.jar expire -<other args>``` 
2. Use the s3 arguments provided in the documentation.

## Proposed fix:
Nessie-gc has some command arguments for expire phase specific to the data lake ->
```    
For S3:
    - s3.access-key-id
    - s3.secret-access-key
    - s3.endpoint, if you use an S3 compatible object store like MinIO
```
Here, 1 more argument is needed 
```s3.path-style-access=true```

Without this argument the s3 data-lake endpoint generated by nessie-gc is aws compatible which in our setup gives "UnknownHostException".

snazy:
Hi @kishlay-kr,
do you want to provide a PR to add some more context information to the help message?

kishlay-kr:
Sure, will raise a PR for the same.

PR #11169, opened 2025-08-15
Add missing s3.path-style-access to GC Tool help

huw0:
I found #9991 while using the Nessie GC Tool for the first time.

This fixes the help output.

Before:

```
For S3:
    - s3.access-key-id
    - s3.secret-access-key
    - s3.endpoint, if you use an S3 compatible object store like MinIO
```

After:

```
For S3:
    - s3.access-key-id
    - s3.secret-access-key
For non-AWS S3 you need to specify the endpoint and possibly enable
    path-style-access:
    - s3.endpoint
    - s3.path-style-access=true
```

_Phrasing taken from the docs_.


dimas-b (approved):
Thanks for the fix, @huw0 ! LGTM :+1:

PR #11276, opened 2025-09-10
Build/local-testing: Allow using custom Quarkus versions and not enforce Quarkus platform versions

snazy:
Nessie releases use the Quarkus version defined in the `libs` version catalog using Gradle's `enforcedPlatform`
dependency constraint mechanism. This means that the Quarkus version used by default is fixed and the dependencies
defined by the Quarkus platform are enforced.

For testing purposes, it is possible to use a different Quarkus version by setting the `quarkus.custom.version`
system property.
It is also possible to not enforce the Quarkus platform versions, i.e. using Gradle's `platform` dependency
constraint mechanism, by setting the `quarkus.custom.noEnforcePlatform` system property to `true`.

Example, to test against a pre-release Quarkus 3.28.0.CR1 version:
```bash
./gradlew -Dquarkus.custom.version=3.28.0.CR1 -Dquarkus.custom.noEnforcedPlatform=true test
./gradlew -Dquarkus.custom.version=3.28.0.CR1 -Dquarkus.custom.noEnforcedPlatform=true intTest
```

### DIFF
--- a/events/ri/build.gradle.kts
+++ b/events/ri/build.gradle.kts
@@ -18,6 +18,11 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 plugins {
   alias(libs.plugins.quarkus)
+    .version(
+      libs.plugins.quarkus.asProvider().map {
+        System.getProperty("quarkus.custom.version", it.version.requiredVersion)
+      }
+    )
   id("nessie-conventions-quarkus")
 }
 
--- a/events/ri/build.gradle.kts
+++ b/events/ri/build.gradle.kts
@@ -31,7 +31,7 @@ dependencies {
   implementation(project(":nessie-events-spi"))
 
   // Quarkus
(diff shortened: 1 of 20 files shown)


events/ri/build.gradle.kts
```diff
@@ -18,6 +18,11 @@ import org.apache.tools.ant.taskdefs.condition.Os
 
 plugins {
   alias(libs.plugins.quarkus)
+    .version(
+      libs.plugins.quarkus.asProvider().map {
+        System.getProperty("quarkus.custom.version", it.version.requiredVersion)
+      }
+    )
```
snazy, line 25: This mechanism is sadly the only option I could come up with to align the Quarkus plugin version with a requested custom Quarkus version.

Within a `plugins {}`, it's impossible to use `project` and it's impossible to access anything in `build-logic` (Gradle just complains with a `ClassNotFoundException).

PR #11303, opened 2025-09-16
Iceberg 1.10: S3 remote signing can no longer be "reconfigured"

snazy:
Since Iceberg 1.10 it is no longer possible to use different auth configurations with S3 remote signing. This is caused by https://github.com/apache/iceberg/pull/13215 which keeps auth related configuration in static fields in `o.a.i.aws.s3.signer.S3V4RestSignerClient`.

### DIFF
--- a/servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/IcebergResourceLifecycleManager.java
+++ b/servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/IcebergResourceLifecycleManager.java
@@ -15,20 +15,26 @@
  */
 package org.projectnessie.server.catalog;
 
+import com.github.benmanes.caffeine.cache.Cache;
 import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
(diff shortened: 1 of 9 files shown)


servers/quarkus-server/src/testFixtures/java/org/projectnessie/server/catalog/IcebergResourceLifecycleManager.java
```diff
@@ -38,28 +44,69 @@ public Map<String, String> start() {
...
+    @Override
+    public void stop() {
+      cleanupS3V4RestSignerClient();
+    }
+  }
+
+  /** Stops the delete and "common" worker pools in {@link ThreadPools}. */
+  public static void stopIcebergWorkerPools() {
     ThreadPools.getDeleteWorkerPool().shutdown();
     ThreadPools.getWorkerPool().shutdown();
```
adutra, line 82: You may want to stop the newly-introduced `ThreadPools.authRefreshPool()`.

adutra (approved):

PR #11329, opened 2025-09-19
Fix maxOS CI

snazy:


snazy:
Ref https://scans.gradle.com/s/ii2y7eyykr5to/tests/overview?outcome=FAILED

PR #11493, opened 2025-10-21
Bugfix: S3 remote signing for batch deletions does not work

snazy:
The HTTP uri in the S3 sign request contains URIs with a path that only contains the bucket, but no path (like `http://localhost:12345/bucket?delete`). This fails the path-like check in `org.projectnessie.catalog.files.s3.S3Utils#asS3Location`.

This change updates the regex to fix the issue.

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
@@ -41,7 +41,7 @@ public final class S3Utils {
   //   /bucket/path
   // but not
   //   /bucket/
-  private static final Pattern S3_PATH_PATTERN = Pattern.compile("^/([^/]+)(/.+|)$");
+  private static final Pattern S3_PATH_PATTERN = Pattern.compile("^/([^/]+)(/.+|$)");
 
   private S3Utils() {}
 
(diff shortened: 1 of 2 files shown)


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
```diff
@@ -36,7 +36,12 @@ public final class S3Utils {
 
   private static final Pattern S3_HOST_PATTERN =
       Pattern.compile("^((.+)\\.)?s3[.-]([a-z0-9-]+)\\..*");
-  private static final Pattern S3_PATH_PATTERN = Pattern.compile("^/([^/]+)/.*");
+  // This matches any of
+  //   /bucket
+  //   /bucket/path
+  // but not
+  //   /bucket/
+  private static final Pattern S3_PATH_PATTERN = Pattern.compile("^/([^/]+)(/.+|)$");
```
adutra: This also matches `/bucket/path/` – is that OK?

dimas-b: Does this change deserve a dedicated test case?

dimas-b: nit: `"^/([^/]+)(/.*|$)"` - is it more readable?

PR #11553, opened 2025-10-30
fix(deps): Update com.github.ben-manes.caffeine:caffeine to v3.2.4 + adopt tests

snazy:
The read of `Eviction` properties is "just" a volatile read since Caffeine 3.2.3 and trigger cleanups asynchronously. Before 3.2.3, cleanups happened synchronously.  This change breaks the initially present assertions of this test, but not the functionality of the production code. See https://github.com/ben-manes/caffeine/issues/1897

### DIFF
--- a/versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
+++ b/versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
@@ -16,6 +16,7 @@
 package org.projectnessie.versioned.storage.cache;
 
 import static java.util.concurrent.CompletableFuture.delayedExecutor;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


versioned/storage/cache/src/test/java/org/projectnessie/versioned/storage/cache/TestCacheOvershoot.java
```diff
@@ -123,9 +123,14 @@ private void testCacheOvershoot(Executor evictionExecutor) throws Exception {
       executor.awaitTermination(10, TimeUnit.MINUTES);
     }
 
+    // The read of `Eviction` properties is "just" a volatile read since Caffeine 3.2.3
+    // and trigger cleanups asynchronously. Before 3.2.3, cleanups happened synchronously.
+    // This change breaks the initially present assertions of this test, but not the
```
dimas-b: "This change" is good in a commit message, but we want to keep it in the comment (forever)?

snazy:
Um - so `@RetryingTest` doesn't like softly

PR #11777, opened 2025-12-21
Helm: Add chart test configuration

sourajitsaha17:
## Description
Add `tests.disabled` parameter to allow users to disable Helm chart tests.

## Motivation
Currently, test resources are always created with the chart and it is not configurable. In our use case, with OIDC authentication enabled the chart test fails with 401 leading to failed deployment. We would like to use our custom chart tests and disable the default one.

## Changes
- ✅ Wrap test templates with `{{- if not (and .Values.tests .Values.tests.disabled) }}`
- ✅ Update documentation in `README.md.gotmpl`
- ✅ Generate updated `README.md` with helm-docs

## Testing
- [x] Tested with `helm template` (both enabled and disabled)
- [x] Tested with local k8s deployment
- [x] Tested with `helm lint`
- [x] Verified backward compatibility (default: enabled)


dimas-b (approved):
LGTM, pending @adutra 's review :)

PR #11879, opened 2026-01-12
Errorprone: JavaUtilDate

snazy:
Preparation for #11829


catalog/files/impl/src/main/java/org/projectnessie/catalog/files/gcs/GcsClients.java
```diff
@@ -181,7 +181,15 @@ static Credentials buildCredentials(
           AccessToken accessToken =
               new AccessToken(
                   oauth2token.token(),
-                  oauth2token.expiresAt().map(i -> new Date(i.toEpochMilli())).orElse(null));
+                  oauth2token
+                      .expiresAt()
+                      .map(
+                          i -> {
+                            @SuppressWarnings("JavaUtilDate")
```
dimas-b: Do we want to enforce `JavaUtilDate` at all?

dimas-b (approved):

PR #11885, opened 2026-01-12
Require Java 17 as the minimum Java version for clients

snazy:
This follows Iceberg's recent change to drop Java 11.


snazy:
As a side effect, this enables JUnit v6 #11852 and Agrona v2 #10105.

PR #11891, opened 2026-01-12
Errorprone: ConstantPatternCompile

snazy:
Preparation for #11829


versioned/storage/bigtable-tests/src/main/java/org/projectnessie/versioned/storage/bigtabletests/BigTableBackendProcessTestFactory.java
```diff
@@ -92,6 +92,7 @@ public void start() {
               .start();
       System.out.println("Cloud Bigtable emulator started with PID " + process.pid());
       if (emulatorPort == 0) {
+        @SuppressWarnings("ConstantPatternCompile")
```
dimas-b: why suppress and not extract a constant?

snazy: Just because it's a one-off thing (test stuff)

PR #11995, opened 2026-01-29
Update to Quarkus 3.31

snazy:
Contains all the changes that are necessary for Nessie due to the Quarkus 3.31 version bump.
None of these changes can be applied to earlier Quarkus versions.

Noteworthy:
* `io.quarkus.vertx.http.HttpServer` is a interface for Quarkus _tests_ to retrieve information about the HTTP related stuff. It does not yet work for all cases, which use the "legacy" way and extract the information via `static`ally callable code. It's definitely not always the "best designed code", but it works (aka good enough).
* The Mongo-DB changes are necessary to account for the new way to define (multiple) Mongo drivers and that have an `active` flag. The internal-ish Mongo related classes in the Quarkus extension have changed. The added config-source makes the whole change transparent to users.


snazy:
Heads up: I've added another commit to replace the relocating `quarkus-junit5` with `quarkus-junit` (new in 3.31).

PR #12008, opened 2026-01-31
Azurite: use old service version in tests

snazy:
Azurite does not work with the Azure services version used by recent Azure SDK (#12003). The last Azurite release was quite a while ago.


snazy:
This change enables #12003, see CI results on https://github.com/projectnessie/nessie/commit/a0587426c75ffef999423b9e4bfc9d3c82f00b53 which has this change and the version bump.

PR #12046, opened 2026-02-07
Site: add a script to remove versioned-docs with identical content

snazy:
This script removes versions from the versioned docs that have identical content.

This effectively removes 30 of the exiting 68 versions.
The reduction of the time for mkdocs site generation is significant: from 500s to 180s (on my local machine, not in CI).
In CI, the site-job currently runs for nearly 30 minutes.

PR #12047, opened 2026-02-07
site-docs: Apply cleanup-versions script from #12046

snazy:


snazy:
> I very much doubt anyone reads a specific historical version of Nessie docs 😉

Maybe - nobody knows

PR #12308, opened 2026-04-13
Fix IntelliJ import

snazy:
After #12126, Scala 2.13 became the default version used during IntelliJ/Gradle syncs. The corresponding config file however mentioned 2.12 as the first and therefore "only" Scala version to sync, which breaks the IDE with "no such project" errors.

### DIFF
--- a/integrations/spark-scala.properties
+++ b/integrations/spark-scala.properties
@@ -24,5 +24,5 @@ sparkVersions=4.0,3.5,3.4
 
 sparkVersion-3.4-scalaVersions=2.13,2.12
 sparkVersion-3.5-scalaVersions=2.13,2.12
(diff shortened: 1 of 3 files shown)


dimas-b (approved):

integrations/spark-scala.properties
```diff
@@ -24,5 +24,5 @@ sparkVersions=4.0,3.5,3.4
 
 sparkVersion-3.4-scalaVersions=2.13,2.12
 sparkVersion-3.5-scalaVersions=2.13,2.12
-sparkVersion-3.5-scalaVersions=2.12,2.13
+sparkVersion-3.5-scalaVersions=2.13,2.12
```
dimas-b, line 27: is it duplicated on line 26?

PR #12335, opened 2026-04-16
fix: prune unreferenced schemas from table metadata

jshmchenxi:
Nessie maintains a single snapshot per table per branch, making historical schemas unreachable. Tables with frequent schema evolution can accumulate hundreds of schema versions in the metadata, causing significant metadata bloat.

This change ensures only referenced schemas are included:
- Write path (icebergTableSnapshotToNessie): only store the current schema in NessieTableSnapshot
- Read path (nessieTableSnapshotToIceberg): only include the current schema in metadata output; fall back to including all schemas when no current schema is set for compatibility
- Handle edge case where snapshot references a different schema than the metadata-level current schema

Field mappings from all schemas are still collected internally for partition spec and sort order resolution.

### DIFF
--- a/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
+++ b/catalog/format/iceberg/src/main/java/org/projectnessie/catalog/formats/iceberg/nessie/NessieModelIceberg.java
@@ -601,6 +601,8 @@ public class NessieModelIceberg {
 
     Map<Integer, NessieSchema> icebergSchemaIdToSchema = new HashMap<>();
 
+    // Populate the schema lookup map from all schemas (needed for snapshot schema resolution),
+    // but only store the current schema in the snapshot to prevent unbounded schema accumulation.
     partsWithV1DefaultPart(iceberg.schemas(), iceberg.schema(), IcebergSchema::schemaId)
         .forEach(
             schema -> {
@@ -610,8 +612,8 @@ public class NessieModelIceberg {
 
               icebergSchemaIdToSchema.put(schema.schemaId(), nessieSchema);
 
-              snapshot.addSchemas(nessieSchema);
               if (isCurrent) {
+                snapshot.addSchemas(nessieSchema);
                 snapshot.currentSchemaId(nessieSchema.id());
               }
             });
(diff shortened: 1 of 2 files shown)


dimas-b:
@jshmchenxi : Please fix conflicts... I believe CI will not run otherwise :sweat_smile:

snazy (commented):
Don't Iceberg manifests reference older schemas or partition-specs?

snazy:
I think there is now a table-update operations to remove schemas, which could be implemented via `org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate`.

jshmchenxi:
> Don't Iceberg manifests reference older schemas or partition-specs?

Good point, yes older schemas are referenced by manifests via `schema-id` per Iceberg spec. However the manifest also includes a `schema` field for the JSON representation of the table schema, so manifests are self-contained — `ManifestReader` parses the inline `schema` JSON without looking up `schema-id` from `TableMetadata.schemas()`. Engines (Trino/Spark) only call `SnapshotUtil.schemaFor(table, snapshotId)` to look up the schema for the *current* snapshot, which we always preserve.

> I think there is now a table-update operations to remove schemas, which could be implemented via `org.projectnessie.catalog.formats.iceberg.rest.IcebergMetadataUpdate`.

Yes, [#12022](https://github.com/apache/iceberg/pull/12022) added the `RemoveSchemas` REST update type and [#12089](https://github.com/apache/iceberg/pull/12089) wired it into `RemoveSnapshots` (i.e. `ExpireSnapshots(cleanExpiredMetadata=true)`).

However, per the [Nessie spec](https://projectnessie.org/develop/spec/#iceberg-table) — *"none of the built-in iceberg maintenance procedures can be used"* against Nessie tables, since Nessie tracks references across branches/commits separately from Iceberg snapshots. So implementing `RemoveSchemas` as an `IcebergMetadataUpdate` wouldn't help in practice — no client would call it (the only producer is `ExpireSnapshots(cleanExpiredMetadata=true)` which isn't recommended for Nessie).

What I was trying to address is an open Iceberg issue — https://github.com/apache/iceberg/issues/5219. Since Nessie only keeps the latest snapshot, server-side pruning seems like a natural fit. The current PR does it transparently in `nessieTableSnapshotToIceberg` / `icebergTableSnapshotToNessie`. Would that approach work, or would you prefer a different mechanism (e.g., a config flag, or a `cleanMetadata` operation)?

snazy:
I'm less concerned, actually not at all, about clients accessing the latest snapshot with only the current schema.

I'm more concerned about Nessie GC running into issues around missing schemas. IIRC there was a bug in PyIceberg quite a while ago, where the "embedded" schema or partition-spec (don't recall which one) was not included in the way as Iceberg/Java does. The fix (or rather workaround) was to fallback from the schema/spec included in the manifest to the information from the metadata.

PR #12397, opened 2026-05-04
Quarkus 3.35 upgrade preparation

snazy:
Need to update a few places where the Quarkus http server port is required.

While at it, this change also uses the officially supported path via `io.quarkus.vertx.http.HttpServer` instead of `ConfigProvider`, because the latter will eventually only yield the _statically_ configured values in all cases, not the effective runtime values.


snazy:
This change is required for #12395

dimas-b (approved):

Issue #10865, opened 2025-06-03
[Bug]: Nessie CLI does not properly redirect STDOUT


dimas-b:
### What happened

```
$ java -jar nessie-cli-0.103.3-SNAPSHOT.jar \
  -u http://localhost:19120/api/v2 \
  -c "create namespace ns2" \
  -c "SHOW LOG" >/tmp/log.txt
```
Note: output goes to the terminal (console) but not to `/tmp/log.txt`
```
$ cat /tmp/log.txt 
```
^ empty

Nessie CLI should allow redirecting its output when `--command` or `--run-script` options are used.

### How to reproduce it

See above

### Nessie server type (docker/uber-jar/built from source) and version

0.103.*

### Client type (Ex: UI/Spark/pynessie ...) and version

Nessie CLI

mj006648:
Hi @dimas-b, I'd like to take a look at this as my first contribution to Nessie. From a quick read of `NessieCliImpl`, my working hypothesis is that the jline `Terminal` built in `call()` writes directly to the console and bypasses shell redirection (the existing inline comment already hints at this). For non-interactive runs (`-c` / `-s`), it should be safe to fall back to a plain `System.out` PrintWriter so that `>` and `|` work as expected.

Before I open a PR, does this direction sound reasonable, or is there a preferred approach (e.g. routing through `--non-ansi`, or a dumb-terminal builder)? Happy to follow whichever path you prefer.

dimas-b:
@mj006648 : you assessment looks correct to me 🙂 

Instead of `--non-ansi` I'd personally use `--plain` / `-P`, but the former works too.

PR #12425, opened 2026-05-15
Add --plain / -P alias for the --non-ansi CLI option

mj006648:
## What

Add `--plain` (long) and `-P` (short) as aliases for the existing
`--non-ansi` option of `nessie-cli`, as discussed in #10865.

## Why

`--non-ansi` already produces output that's safe to redirect or pipe,
but the name doesn't surface that as a workaround. Anyone hitting the
redirect bug is more likely to search for "plain output" than for
"ANSI". The new aliases make the existing behavior discoverable
without breaking anything.

## How

- New constant `OPTION_PLAIN = "--plain"` in `NessieCliImpl`.
- `@Option` `names` extended to `{"-P", "--plain", "--non-ansi"}` on
  the same `dumbTerminal` field.
- Description updated to spell out the redirection / pipe use case.

Same field, same code path. 5 lines changed.

## Verification

Built locally with `:nessie-cli:shadowJar` (GraalVM 21.0.11) and ran
the resulting jar against a few redirect / pipe scenarios:

| Scenario | Exit | Size | ANSI escapes |
|---|---|---|---|
| `--plain -c "EXIT" > file` | 0 | 2731 B | 0 |
| `-P -c "EXIT" > file` | 0 | 2731 B | 0 |
| `--non-ansi -c "EXIT" > file` (regression) | 0 | 2731 B | 0 |
| `--plain -c "EXIT" \| head` | 0 | 38 lines | 0 |
| `--plain -s -` (stdin script) | 0 | 2731 B | 0 |
| `--plain --quiet -c "EXIT" > file` | 0 | 0 B | 0 |

Identical output across all three names — the only diff is the
`startup took N ms` line in the banner.

## One thing I noticed

Without any of these flags, redirecting `-c` / `-s` actually crashes
rather than silently producing an empty file:

```
java.lang.IllegalStateException: Unable to create a terminal
  at org.jline.terminal.TerminalBuilder.doBuild(...)
  at org.projectnessie.nessie.cli.cli.NessieCliImpl.call(NessieCliImpl.java:189)
```

jline can't load any terminal provider (FFM / jansi / JNA) when stdout
is not a TTY. This PR doesn't change that path — it just makes the
workaround discoverable. Let me know if it's worth a follow-up to fall
back to a plain `PrintWriter` automatically in non-interactive mode.

## Tests

No new tests — the change is a picocli alias on an existing field, and
the option-parsing path is already covered. Happy to add one if you'd
like.

Relates to #10865

### DIFF
--- a/cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
+++ b/cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
@@ -91,6 +91,7 @@ public class NessieCliImpl extends BaseNessieCli implements Callable<Integer> {
   public static final String OPTION_KEEP_RUNNING = "--keep-running";
   public static final String OPTION_CONTINUE_ON_ERROR = "--continue-on-error";
   public static final String OPTION_NON_ANSI = "--non-ansi";
+  public static final String OPTION_PLAIN = "--plain";
 
   public static final String HISTORY_FILE_DEFAULT = "~/.nessie/nessie-cli.history";
   public static final AttributedStyle STYLE_ERROR_HIGHLIGHT = STYLE_ERROR.italic().bold();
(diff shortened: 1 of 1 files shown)


dimas-b (commented):
Thanks for your contribution, @mj006648 !

This change makes sense, but it's not quite what I meant in #10865 😅 

I guess it might be worth trying to force a `PosixSysTerminal` (my use case is/was on Linux). It should use normal STDIN/OUT streams. Colours (ANSI) is a slightly different concern.

If the `PosixSysTerminal` approach works we can still keep this change, but maybe add another option like `--stdout` / `-S` for using stream instead of PTY.

If you prefer we can still merge this change without closing #10865 and then have another PR to improve more. WDYT?

mj006648:
Thanks for the clarification — I'd misread your comment and ended up fixing only the discoverability of the workaround, not the underlying PTY-vs-stream issue. The redirect behavior is in fact unchanged here.

Happy to go with merging this one as a small discoverability improvement and keep #10865 open. I'll open a follow-up that forces a PosixSysTerminal (with a separate --stdout / -S option as you suggested) for the real fix once this lands.

dimas-b (approved):

dimas-b:
@mj006648 : sounds good - merging

PR #12431, opened 2026-05-23
nessie-cli: add --stdout / -S for stream-backed terminal (#10865)

mj006648:
Follow-up to #12425, addressing the underlying PTY-vs-stream issue from #10865 that the `--plain` alias didn't fix. As discussed in https://github.com/projectnessie/nessie/pull/12425#issuecomment-4466091419.

Adds a new `--stdout` / `-S` option. When set, builds the jline terminal as:

```java
TerminalBuilder.builder()
    .jansi(false)
    .system(false)
    .streams(System.in, System.out)
    .type("dumb")
    .build();
```

`system(false).streams(...)` gives a stream-backed Terminal that follows redirected streams without opening `/dev/tty`. `type("dumb")` suppresses ANSI without going through `.dumb(true).provider(DUMB)`, which re-triggers a TTY lookup and fails the same way on a redirected stdout.

Kept separate from `--plain` since the concerns are orthogonal:
- `--plain` / `--non-ansi` — ANSI on/off (existing)
- `--stdout` — stream routing (new, implies plain)

Verified locally against a real Nessie server (the exact `-c "SHOW LOG" > /tmp/log.txt` scenario from the issue body now works). `:nessie-cli:test` and `:spotlessCheck` pass. Tested on Linux; should be OS-agnostic but unverified on Windows / macOS.

Did not change the default (no-flag) behaviour, which still crashes on redirected stdout — that seemed like a separate discussion. Happy to fold it in if you'd rather.

### DIFF
--- a/cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
+++ b/cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
@@ -128,7 +128,7 @@ public class NessieCliImpl extends BaseNessieCli implements Callable<Integer> {
   @Option(
       names = {"-S", OPTION_STDOUT},
       description = {
-        "Use standard System.in/System.out streams for the terminal instead of opening",
+        "Use standard in/out streams for communicating with the operator instead of opening",
         "the controlling PTY. This makes shell redirection (>) and pipes (|) work even",
         "in environments where jline can still detect a TTY (e.g. interactive shells).",
         "Implies " + OPTION_PLAIN + " (no ANSI control sequences are emitted)."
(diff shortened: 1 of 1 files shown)


cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
```diff
@@ -124,6 +125,17 @@ public class NessieCliImpl extends BaseNessieCli implements Callable<Integer> {
       defaultValue = "false")
   private boolean dumbTerminal;
 
+  @Option(
+      names = {"-S", OPTION_STDOUT},
+      description = {
+        "Use standard System.in/System.out streams for the terminal instead of opening",
```
dimas-b: nit: just `standard in/out` is probably sufficient.

dimas-b: `for the terminal ...` -> `for communicating with the operator`?

dimas-b (dismissed):
Thanks again for your contribution, @mj006648 ! This PR LGTM with minor comments.

cli/cli/src/main/java/org/projectnessie/nessie/cli/cli/NessieCliImpl.java
mj006648: Addressed both nits in e38ddd3, thanks!

dimas-b (approved):

PR #12432, opened 2026-05-28
Normalize S3 scheme in IcebergConfigurer writeable derivation (fixes #12426)

### LINKED ISSUE #12426: IcebergConfigurer.writeable check fails for HadoopFileIO tables (scheme s3a:// vs warehouse s3://) — signing rejects all commits
## Version

- Nessie server: **0.107.5** (Quarkus image `ghcr.io/projectnessie/nessie:0.107.5`)
- Iceberg client: 1.7.0
- Spark: 3.5.x
- pyiceberg: 0.7.x (also affected when using HadoopFileIO via RestCatalog Python client)

## Description

When a Spark client writes to a Nessie REST catalog using `HadoopFileIO` (the default classical Iceberg FileIO via `hadoop-aws`), the resulting `metadata.location` is prefixed with scheme `s3a://`. The Nessie server registers the warehouse with scheme `s3://`. The internal `IcebergConfigurer.icebergConfigPerTable()` writeable derivation appears to apply `S3Utils.normalizeS3Scheme` only to the warehouse path, not to the `metadata.location` it compares against — so the resulting array `writeable[]` is empty and **all PUT/DELETE signing requests are subsequently rejected** with `HTTP 403 unauthorized signing request`.

End result: tables created via `HadoopFileIO` against a Nessie REST catalog backed by S3-compatible storage (MinIO, AWS S3) become silently unwriteable. Reads work fine.

### Minimal Nessie config

```yaml
nessie:
  image: ghcr.io/projectnessie/nessie:0.107.5
  environment:
    nessie.catalog.default-warehouse: dev
    nessie.catalog.warehouses.dev.location: s3://my-bucket/    # scheme s3://
    nessie.catalog.service.s3.default-options.endpoint: http://minio:9000
    nessie.catalog.service.s3.default-options.path-style-access: "true"
    nessie.catalog.service.s3.default-options.region: us-east-1
```

### Spark client config that triggers the bug

```bash
spark-submit \
  --conf spark.sql.catalog.demo=org.apache.iceberg.spark.SparkCatalog \
  --conf spark.sql.catalog.demo.catalog-impl=org.apache.iceberg.rest.RESTCatalog \
  --conf spark.sql.catalog.demo.uri=http://nessie:19120/iceberg \
  --conf spark.sql.catalog.demo.warehouse=dev \
  # io-impl OMITTED -> defaults to HadoopFileIO which triggers the bug.
  --conf spark.hadoop.fs.s3a.endpoint=http://minio:9000 \
  ...
```

### Operation that fails

```python
df = spark.createDataFrame([(1, "x")], ["id", "name"])
df.writeTo("demo.test_ns.test_table").create()
```

### Observed error

```
org.apache.iceberg.exceptions.CommitFailedException:
  Unauthorized signing request
  caused by: org.apache.iceberg.aws.s3.signer.S3V4RestSignerClient:
    HTTP 403 — signing rejected, writeable=[]
```

Inspecting `GET /iceberg/v1/{prefix}/config?warehouse=dev`:

```json
{
  "overrides": {
    "s3.signer.uri": "http://nessie:19120/iceberg/v1/aws/s3/sign",
    "writeable": []
  }
}
```

The Spark `HadoopFileIO` writes `metadata.location: s3a://my-bucket/...` to the commit. The warehouse is registered as `s3://my-bucket/`. Tables written with `HadoopFileIO` have `metadata.location: s3a://...`; tables written with `io-impl=S3FileIO` (AWS SDK v2 direct) have `metadata.location: s3://...` and work correctly.

## Root cause hypothesis

In `IcebergConfigurer.icebergConfigPerTable()` (`servers/quarkus-server/src/main/java/.../catalog/IcebergConfigurer.java` or equivalent; approximate lines 225-320 of 0.107.5), the writeable derivation appears to be roughly:

```java
String warehouseNormalized = S3Utils.normalizeS3Scheme(warehouse.getLocation());
// "s3://my-bucket/"

String metadataLoc = table.metadataLocation();
// "s3a://my-bucket/.../00000-...metadata.json"

if (metadataLoc.startsWith(warehouseNormalized)) {
  // "s3a://...".startsWith("s3://...") -> FALSE
  writeable.add(metadataLoc);
}
// -> writeable[] stays empty
```

`S3Utils.normalizeS3Scheme()` is called on the warehouse but **not** on the `metadataLoc` before the `startsWith` check. The scheme mismatch (`s3://` vs `s3a://`) makes the predicate always false, so the writeable array stays empty and the signing endpoint then rejects every PUT/DELETE.

## Suggested fix

Normalize both sides before the comparison:

```java
String warehouseNormalized = S3Utils.normalizeS3Scheme(warehouse.getLocation());
String metadataLocNormalized = S3Utils.normalizeS3Scheme(metadataLoc);

if (metadataLocNormalized.startsWith(warehouseNormalized)) {
  writeable.add(metadataLoc);
}
```

A more defensive variant would normalize the warehouse scheme at storage time and normalize all metadata locations once at startup.

## Workaround

Force `io-impl=org.apache.iceberg.aws.s3.S3FileIO` on every Spark client. S3FileIO writes `s3://` natively, eliminating the mismatch.

## Impact

- **Severity:** high for any user mixing Spark + Nessie REST + S3-compatible storage (MinIO, AWS S3) with default `HadoopFileIO`.
- **Workaround known:** yes (`io-impl=S3FileIO`) but requires Iceberg 1.7+ and `iceberg-aws-bundle`.
- **Silenced confusion:** the failure surfaces in the signing endpoint, not the scheme check itself — users typically spend hours debugging credentials before reaching the actual cause.

Happy to help test a fix against this scenario if useful.

mj006648:
## Summary

With the default `HadoopFileIO`, Spark writes `metadata.location` prefixed
with `s3a://`, while the warehouse is registered as `s3://`. The
`startsWith` check in `IcebergConfigurer.icebergConfigPerTable()` is done
on the raw strings, so the scheme mismatch makes the predicate false and
`writeable[]` ends up empty. The S3 signer then rejects every PUT / DELETE
with `HTTP 403 unauthorized signing request`. Reads don't go through the
signer, which is what makes this confusing in practice.

Running both sides of the prefix check through `S3Utils.normalizeS3Scheme`
fixes it. The `StorageUri` values kept in `writeable` / `readOnly` are
left at their original scheme, and the existing normalization in
`configureS3RequestSigningForTable` still produces the canonical form for
the signer payload, so the wire format is unchanged.

Kept this to the call-site fix; the storage-time variant from the issue
is a wider change and probably belongs in a separate PR.

## Related Issue

Fixes #12426

## Changes

- `IcebergConfigurer.java`
  - Normalize `warehouseLocation` once at the top of `icebergConfigPerTable`.
  - Normalize `tableMetadata.location()` and the `icebergWriteLocation`
    path before the `startsWith` checks.
- `TestIcebergConfigurer.java`
  - Add `s3a://` and `s3n://` rows to the existing parameterized test.
  - Compare `signerParams.writeLocations()` against the normalized form
    (what the signer actually receives).

## How to Verify

```bash
./gradlew :nessie-catalog-service-rest:test \
  --tests "org.projectnessie.catalog.service.rest.TestIcebergConfigurer"
```

The new `s3a://` / `s3n://` rows fail on `main` (empty `writeLocations`)
and pass with this change.

## Checklist

- [x] DCO sign-off
- [x] Unit test covers `s3://`, `s3a://`, `s3n://`
- [x] No behavior change for the `s3://`-only path
- [x] No changes to the signer wire format

### DIFF
--- a/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
+++ b/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
@@ -227,13 +227,24 @@ public class IcebergConfigurer {
     Set<StorageUri> readOnly = new HashSet<>();
     Set<StorageUri> maybeWriteable = writeAccessGranted ? writeable : readOnly;
     StorageUri locationUri = StorageUri.of(tableMetadata.location());
-    (tableMetadata.location().startsWith(warehouseLocation) ? maybeWriteable : readOnly)
+    // Normalize the S3 scheme on both sides of the prefix check so that tables written by
+    // HadoopFileIO (metadata.location prefixed with s3a://) are still classified as writeable
+    // against a warehouse registered with the canonical s3:// scheme. Without this, the
+    // resulting writeable[] is empty and the S3 request signer rejects every PUT/DELETE.
+    String normalizedWarehouseForPrefix = normalizeS3Scheme(warehouseLocation);
+    String normalizedTableLocationForPrefix = normalizeS3Scheme(tableMetadata.location());
+    (normalizedTableLocationForPrefix.startsWith(normalizedWarehouseForPrefix)
+            ? maybeWriteable
+            : readOnly)
         .add(locationUri);
 
     if (!icebergWriteObjectStorage(tableConfig, tableMetadata.properties(), warehouseLocation)) {
       String writeLocation = icebergWriteLocation(tableMetadata.properties());
-      if (writeLocation != null && !writeLocation.startsWith(tableMetadata.location())) {
-        (writeLocation.startsWith(warehouseLocation) ? maybeWriteable : readOnly)
+      if (writeLocation != null
+          && !normalizeS3Scheme(writeLocation).startsWith(normalizedTableLocationForPrefix)) {
+        (normalizeS3Scheme(writeLocation).startsWith(normalizedWarehouseForPrefix)
+                ? maybeWriteable
+                : readOnly)
             .add(StorageUri.of(writeLocation));
       }
     }
--- a/catalog/service/rest/src/test/java/org/projectnessie/catalog/service/rest/TestIcebergConfigurer.java
+++ b/catalog/service/rest/src/test/java/org/projectnessie/catalog/service/rest/TestIcebergConfigurer.java
@@ -21,6 +21,7 @@ import static java.time.temporal.ChronoUnit.DAYS;
 import static org.junit.jupiter.params.provider.Arguments.arguments;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.projectnessie.catalog.files.s3.S3Utils.normalizeS3Scheme;
 import static org.projectnessie.catalog.secrets.UnsafePlainTextSecretsManager.unsafePlainTextSecretsProvider;
 import static org.projectnessie.catalog.service.rest.IcebergConfigurer.S3_SIGNER_ENDPOINT;
 import static org.projectnessie.catalog.service.rest.IcebergConfigurer.S3_SIGNER_URI;
@@ -196,7 +197,10 @@ public class TestIcebergConfigurer {
               p -> p.signerSignature().warehouseLocation(),
               p -> p.signerSignature().identifier())
           .containsExactly(
-              signerKey.name(), List.of(loc), warehouseLocation, key.toPathStringEscaped());
+              signerKey.name(),
+              List.of(normalizeS3Scheme(loc)),
+              warehouseLocation,
+              key.toPathStringEscaped());
       soft.assertThat(signPath).doesNotStartWith("/");
       soft.assertThat(signUri).isNotNull();
     } else {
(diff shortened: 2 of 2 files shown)


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
```diff
@@ -227,13 +227,24 @@ IcebergTableConfig icebergConfigPerTable(
...
+            : readOnly)
         .add(locationUri);
 
     if (!icebergWriteObjectStorage(tableConfig, tableMetadata.properties(), warehouseLocation)) {
       String writeLocation = icebergWriteLocation(tableMetadata.properties());
-      if (writeLocation != null && !writeLocation.startsWith(tableMetadata.location())) {
-        (writeLocation.startsWith(warehouseLocation) ? maybeWriteable : readOnly)
+      if (writeLocation != null
+          && !normalizeS3Scheme(writeLocation)
+              .startsWith(normalizeS3Scheme(tableMetadata.location()))) {
```
dimas-b: nit: it looks like the value of `normalizeS3Scheme(tableMetadata.location()` could be stored in a local var too... in order to avoid multiple `normalizeS3Scheme()` calls.

dimas-b (dismissed):
Thanks, @mj006648 !

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergConfigurer.java
mj006648: Thanks, addressed in the latest push by storing the normalized table metadata location in a local variable and reusing it for the prefix checks.

dimas-b (approved):

mj006648:
This has been approved with all checks passing for two weeks. Anything still pending, or is it ready to merge?

dimas-b:
@snazy : WDYT?

snazy (approved):
LGTM

PR #12500, opened 2026-06-02
Remove resteasy-classic from keycloak-testcontainer

snazy:
... so it doesn't conflict with resteasy-reactive

### DIFF
--- a/testing/keycloak-container/src/main/java/org/projectnessie/testing/keycloak/CustomKeycloakContainer.java
+++ b/testing/keycloak-container/src/main/java/org/projectnessie/testing/keycloak/CustomKeycloakContainer.java
@@ -15,11 +15,23 @@
  */
 package org.projectnessie.testing.keycloak;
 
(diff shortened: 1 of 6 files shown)


testing/keycloak-container/src/main/java/org/projectnessie/testing/keycloak/CustomKeycloakContainer.java
```diff
@@ -316,15 +320,138 @@ private RealmRepresentation createRealm() {
...
+      if (response.statusCode() == status) {
+        return;
+      }
+    }
+
+    throw new IllegalStateException(
+        "Keycloak request failed with HTTP " + response.statusCode() + ": " + response.body());
+  }
+
+  private static void retry(CheckedRunnable runnable) {
```
adutra, line 420: nit: `org.keycloak.common.util.Retry` is apparently available in `keycloak-client-common-synced`.

adutra (approved):

PR #12620, opened 2026-06-24
Catalog: fail closed on S3 signing snapshot lookup errors

snazy:


catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -189,8 +189,10 @@ private Uni<SnapshotResponse> fetchSnapshot() {
                   requestMeta.build(),
                   ICEBERG_V1);
       // consider an import failure as a non-existing content:
-      // signing will be authorized for the future location only.
```
dimas-b, line 192: I think this comment still makes sense. Why remove it?

dimas-b (dismissed):

dimas-b (approved):

PR #12643, opened 2026-06-27
feat(helm): add optional Nessie GC CronJob

### LINKED ISSUE #12429: [Feature]: add nessie GC to the helm chart
### Description

As an operator i would like to be able to configure a nessie GC job as part of the helm chart.

I would prefer a cronjob that runs the gc docker image based on a schedule.

Currently i would have to wrap the nessie helm chart or use kustomize to install the GC cronjob

### Expected Use Cases

_No response_

_No response_

mvanhorn:
## Summary

Adds an optional, opt-in Nessie Garbage Collection (GC) CronJob to the Helm chart, addressing #12429. When `gc.enabled=true`, the chart renders a `batch/v1` `CronJob` that runs the published Nessie GC tool image (`ghcr.io/projectnessie/nessie-gc`) on a configurable schedule. It is disabled by default, so existing installs are unaffected.

The chart provides only the schedulable scaffolding; GC behavior (version store connection, subcommand, options) stays operator-driven via `gc.args` / `gc.extraArgs` / `gc.env`, which keeps the change bounded and avoids baking opinionated version-store config into the chart.

New `gc` values cover: `enabled`, `schedule`, `image` (repository/tag/pullPolicy), `command`/`args`/`extraArgs`, `env`, `resources`, `serviceAccountName`, `podAnnotations`/`podLabels`, `nodeSelector`/`tolerations`/`affinity`, and CronJob policy fields (`concurrencyPolicy`, `suspend`, `startingDeadlineSeconds`, `successfulJobsHistoryLimit`, `failedJobsHistoryLimit`, `restartPolicy`, `backoffLimit`). The template reuses the chart's existing `_helpers.tpl` name/label helpers and inherits the chart's `podSecurityContext`/`securityContext` so the job runs in restricted Pod Security namespaces.

## Why this matters

Operators currently have to wrap the chart or bolt on their own CronJob (via Kustomize or a sidecar chart) to schedule GC. Maintainer @adutra confirmed on the issue that this "makes sense" and that external contributions are welcome. Shipping a first-class, opt-in GC CronJob removes that friction while leaving the default install behavior unchanged.

Notable correctness details:
- GC pods use a GC-specific `app.kubernetes.io/name` so they do **not** match the Nessie `Service` selector (`nessie.selectorLabels`), preventing GC pods from being added as Service endpoints for ports they don't serve.
- Rendering fails fast with a clear message if `gc.schedule` is missing, or if no `gc.command`/`gc.args`/`gc.extraArgs` is supplied (so an enabled CronJob can't be scheduled with a usage-error entrypoint).
- The CronJob name is truncated to stay within Kubernetes name-length limits for long release names.

## Testing

- `helm template nessie helm/nessie -n nessie-ns` -> no CronJob rendered (GC disabled by default).
- `helm template nessie helm/nessie -n nessie-ns --set gc.enabled=true --set 'gc.args={gc}'` -> one CronJob rendered with chart-standard metadata, hardened security contexts, and the GC image.
- `helm unittest helm/nessie` -> 79 tests pass, including the new `tests/gc_cronjob_test.yaml` suite (default-disabled, happy path, custom image, value flow-through, GC-specific pod labels, security context, and the missing-schedule / missing-args failure paths).
- `helm lint helm/nessie` -> clean.
- `helm/nessie/README.md` regenerated with `helm-docs --chart-search-root=helm`.

Fixes #12429


dimas-b:
@mvanhorn : Could you sign the CLA, please?

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+{{- if not .Values.gc.schedule -}}
+{{- fail "gc.schedule is required when gc.enabled is true" -}}
+{{- end -}}
+{{- if and (not .Values.gc.command) (not .Values.gc.args) (not .Values.gc.extraArgs) -}}
+{{- fail "gc.args (or gc.command/gc.extraArgs) must specify the GC subcommand to run when gc.enabled is true" -}}
+{{- end -}}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ printf "%s-gc" (include "nessie.fullname" .) | trunc 52 | trimSuffix "-" }}
```
adutra, line 29: Shouldn't you truncate at 60 instead of 52? Since the suffix is only 3 chars?

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+      labels:
+        {{- include "nessie.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: gc
+    spec:
+      backoffLimit: {{ .Values.gc.backoffLimit }}
+      template:
+        metadata:
+          labels:
+            {{- /*
+              Use a GC-specific app name so these pods do NOT match the Nessie
```
adutra, line 54: 1. Shouldn't we also use the GC specific labels in `jobTemplate.metadata.labels`?
2. How about moving this to a template declared next to `nessie.selectorLabels`? I think this would improve maintainability.

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+          {{- end }}
+          restartPolicy: {{ .Values.gc.restartPolicy }}
+          securityContext:
+            {{- tpl (toYaml .Values.podSecurityContext) . | nindent 12 }}
+          containers:
+            - name: {{ .Chart.Name }}-gc
+              securityContext:
+                {{- tpl (toYaml .Values.securityContext) . | nindent 16 }}
+              image: "{{ tpl .Values.gc.image.repository . }}:{{ tpl .Values.gc.image.tag . | default .Chart.Version }}"
+              imagePullPolicy: {{ tpl .Values.gc.image.pullPolicy . }}
```
adutra, line 88: Should we reuse the global `imagePullSecrets` here? Or open up for GC-specific pull secrets?

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+                {{- end }}
+                {{- with .Values.gc.extraArgs }}
+                {{- tpl (toYaml .) $ | nindent 16 }}
+                {{- end }}
+              {{- end }}
+              {{- with .Values.gc.env }}
+              env:
+                {{- tpl (toYaml .) $ | nindent 16 }}
+              {{- end }}
+              resources:
```
adutra, line 106: nit: use `{{- with .Values.gc.resources }}` to avoid rendering an empty `resources` block.

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
+{{/**
+
+  Copyright (C) 2025 Dremio
```
adutra, line 3:
```suggestion
  Copyright (C) 2026 Dremio
```

helm/nessie/tests/gc_cronjob_test.yaml
```diff
@@ -0,0 +1,131 @@
+##
+## Copyright (C) 2024 Dremio
```
adutra, line 2:
```suggestion
## Copyright (C) 2026 Dremio
```

helm/nessie/templates/gc-cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+              {{- with .Values.gc.command }}
+              command:
+                {{- tpl (toYaml .) $ | nindent 16 }}
+              {{- end }}
+              {{- if or .Values.gc.args .Values.gc.extraArgs }}
+              args:
+                {{- with .Values.gc.args }}
+                {{- tpl (toYaml .) $ | nindent 16 }}
+                {{- end }}
+                {{- with .Values.gc.extraArgs }}
```
adutra, line 98: 1. What is the difference between `args` and `extraArgs`? It's not clear to me.
2. This theoretically allows someone to input only `extraArgs` and no `args`, which is weird.

mvanhorn:
@dimas-b sorry for the slow reply. The CLA is signed - `license/cla` reports "Contributor License Agreement is signed" on the current head (62c160b9), and CLAassistant posted the signed badge on 27 June, so I think it was already green when you asked and the state you saw was stale.

Nothing else is outstanding on my side: the branch is mergeable and `license/cla` is the only check on it. Happy to rebase or split anything up if that helps review.

dimas-b:
@mvanhorn : sorry about the CLA confusion :sweat_smile: It looks good now :sweat_smile: 

Could you respond to @adutra 's review?

mvanhorn:
Thanks @dimas-b, and thanks @adutra for the careful read. Taking the three questions here; the concrete ones (truncation length, `{{- with }}` on resources, the two copyright lines) I am fixing now.

**imagePullSecrets**: it already reuses the global one. `gc-cronjob.yaml` lines 73-76 use the same `{{- if .Values.imagePullSecrets }}` block as `deployment.yaml` lines 51-54, so a chart-wide pull secret already covers the GC pods. I did not add a `gc.imagePullSecrets` override because nothing has needed a GC image from a different registry yet, and it is easy to add if that comes up.

**Labels**: point 1 is deliberate, and I should have made that more visible than a template comment. The GC pod template intentionally does not carry `nessie.selectorLabels`, because the Nessie Service selects on exactly those labels, so GC pods would be registered as Service endpoints despite not serving the API or management ports. That is why they get `app.kubernetes.io/name: <name>-gc` instead.

Point 2 I agree with. I will pull the GC labels into `nessie.gcSelectorLabels` and `nessie.gcLabels` helpers declared directly after `nessie.selectorLabels` in `_helpers.tpl`, so the two sit together and the reason they differ is visible in one place.

**args vs extraArgs**: `args` replaces the image's default arguments and `extraArgs` is appended after it, so the intent was `args` for the subcommand and its core options, `extraArgs` for one-off flags without restating the list. Your second point is fair though. Setting only `extraArgs` still renders a complete `args:` array, so it replaces the image defaults exactly as `args` would and the distinction stops meaning anything. Options:

1. drop `extraArgs` and have people write the full `args` list
2. keep both, but fail rendering if `extraArgs` is set without `args`
3. leave it and document the edge explicitly

I lean towards 1, since `gc.command` plus `gc.args` already covers every case and one fewer knob is easier to explain. Happy to go whichever way you prefer.

mvanhorn:
Pushed the resources guard and the two copyright lines.

On the truncation, I think 52 has to stay. Kubernetes caps CronJob names at 52 characters rather than 63 because the CronJob controller appends 11 characters when it names the Jobs it creates, and Job names are themselves capped at 63. So the 3-character `-gc` suffix is not what 52 is budgeting for; the controller's generated suffix is. At 60 a long release name would produce Job names that fail validation once the CronJob actually fires, which is a failure you would only see on the first scheduled run rather than at install time.

The `trunc 63` on line 60 is a different case and is correct, since that is a label value with no generated suffix.

adutra:
> I lean towards 1, since `gc.command` plus `gc.args` already covers every case and one fewer knob is easier to explain.

Agreed, let's go with option 1 👍

PR #12657, opened 2026-06-29
WIP/Stack: Floci

snazy:
Bunch of commits to migrate away from Minio, Dynamodb-container, fake-gcs-server and Azurite to Floci.


snazy:
Test failures would be fixed via https://github.com/floci-io/floci/pull/1636

PR #12741, opened 2026-07-14
test: run dynamodb storage tests on floci

snazy:
Switch both DynamoDB backend test factories from amazon/dynamodb-local to the Floci AWS emulator on port 4566.

This keeps the existing endpoint and shared-network contract used by Quarkus test resources while replacing the DynamoDB Local image marker and startup command with the Floci image.

### DIFF
--- a/versioned/storage/dynamodb-tests/src/main/resources/org/projectnessie/versioned/storage/dynamodbtests/Dockerfile-floci-version
+++ b/versioned/storage/dynamodb-tests/src/main/resources/org/projectnessie/versioned/storage/dynamodbtests/Dockerfile-floci-version
@@ -1,3 +1,3 @@
 # Dockerfile to provide the image name and tag to a test.
 # Version is managed by Renovate - do not edit.
-FROM docker.io/amazon/dynamodb-local:3.3.0
+FROM docker.io/floci/floci:1.5.33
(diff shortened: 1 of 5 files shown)


versioned/storage/dynamodb-tests/src/main/resources/org/projectnessie/versioned/storage/dynamodbtests/Dockerfile-floci-version
```diff
@@ -1,3 +1,3 @@
 # Dockerfile to provide the image name and tag to a test.
 # Version is managed by Renovate - do not edit.
-FROM docker.io/amazon/dynamodb-local:3.3.0
+FROM docker.io/floci/floci:1.5.33
```
adutra, line 3: What is the advantage of switching to Floci here? SInce`dynamodb-local` is maintained by Amazon, I'd suppose that's a better choice?

snazy: One less image and a faster image startup.
dynamodb-local had some incompatibility issues in the past, so I'm not sure either is any better. Just that the Floci guys are quite quick to release fixes.

adutra (approved):

PR #12743, opened 2026-07-15
test: replace minio helper with floci s3

snazy:
Rename the former MinIO testcontainer module and public JUnit extension API to Floci S3 while keeping the same S3-facing test utility behavior.

The helper now uses the Floci AWS emulator through Testcontainers, updates Quarkus and GC S3 tests to inject FlociS3Access/FlociS3Container, and removes the MinIO image marker from the test infrastructure.


gc/gc-iceberg-files/src/test/java/org/projectnessie/gc/iceberg/files/AbstractFiles.java
```diff
@@ -93,7 +93,7 @@ public void iceberg() throws Exception {
   /**
    * Creates many files, lists the files, deletes 10% of the created files, lists again.
    *
-   * <p>Minio in the used configuration is not particularly fast - creating 100000 objects with 4
+   * <p>FlociS3 in the used configuration is not particularly fast - creating 100000 objects with 4
```
adutra, line 96: Is this paragraph still valid given the change from MinIO to Floci?

snazy: Yea, creating 100000 objects is 100000 put-object requests.

adutra: Sure but I was referring to the benchmark results: are they valid for Floci?

adutra (approved):

gc/gc-iceberg-files/src/test/java/org/projectnessie/gc/iceberg/files/AbstractFiles.java
snazy: Ah, probably not.

PR #12759, opened 2026-07-17
Jackson 3 slice - object-storage-mock

snazy:


testing/object-storage-mock/src/main/java/org/projectnessie/objectstoragemock/s3/BatchDeleteRequest.java
```diff
@@ -38,5 +45,78 @@ default boolean quiet() {
 
   @JsonProperty("Object")
   @JacksonXmlElementWrapper(useWrapping = false)
+  @JacksonXmlProperty(localName = "Object")
   List<S3ObjectIdentifier> objectsToDelete();
+
+  final class Deserializer extends ValueDeserializer<BatchDeleteRequest> {
```
adutra, line 51: What was the issue here that made you create a custom deserializer?

snazy: TL;DR It's a specialty in the AWS SDK XML, where "things" can be represented as scalars or lists/arrays. (Also something that Minio used to trip over.)

Jackson 3 XML can expose AWS SDK XML like `<Object><Key>...</Key></Object>` as a scalar text item instead of the object-shaped DTO Jackson/Immutables expected.

The custom deserializer handles the XML token shapes directly:
* Object may be a repeated unwrapped XML element, not a normal JSON array.
* A single Object may appear as one value.
* The object identifier may arrive either as an object with Key / VersionId, or as a scalar string key.

Similar for `S3ObjectIdentifier`, accepts scalar and object forms.

adutra (approved):

PR #12865, opened 2026-08-08
fix: Support custom S3 virtual-host request signing

### LINKED ISSUE #12854: [Bug]: S3 request signing rejects virtual-host-style URIs, making remote signing unusable with storages that require them (OBS)
### What happened

Nessie tells the client that path-style access is disabled (`s3.path-style-access: false`), so the Iceberg client correctly addresses the object store in virtual-host style. However, the S3 signing endpoint rejects exactly that URI form with `URI not allowed for signing`, and only accepts path-style URIs.
 
For object stores that accept both addressing styles this is invisible. For stores that **only** accept virtual-host style — such as SberCloud OBS — the two requirements have no intersection, and remote signing cannot work at all:
 
- the client must send virtual-host style (the store rejects path-style with `VirtualHostDomainRequired`);
- Nessie only signs path-style.
Since SberCloud OBS also has no STS, vended credentials aren't an alternative either, so there is no way to keep storage credentials off the client.
 
We may well have misconfigured something — if so, we'd be grateful for a pointer, since we couldn't find a combination of settings that works.

## Environment
 
- Nessie `0.107.9` and `0.108.3` (identical behaviour on both)
- Version store: JDBC2 / PostgreSQL 16
- Object store: SberCloud OBS, endpoint `https://obs.ru-moscow-1.hc.sbercloud.ru`, region `ru-moscow-1`
- Client: Spark 3.5 + Iceberg 1.10.1 (`iceberg-spark-runtime-3.5_2.12`, `iceberg-aws-bundle`)
- Auth: OIDC via Keycloak 26, CEL authorization enabled
- Deployment: Docker Compose
Hostnames and bucket name below are sanitized; the OBS endpoint is the real public one.
 
## Configuration
 
Nessie S3 settings:
 
```properties
nessie.catalog.default-warehouse=warehouse
nessie.catalog.warehouses.warehouse.location=s3://example-bucket/nessie-authz
 
nessie.catalog.service.s3.default-options.region=ru-moscow-1
nessie.catalog.service.s3.default-options.endpoint=https://obs.ru-moscow-1.hc.sbercloud.ru
nessie.catalog.service.s3.default-options.path-style-access=false
nessie.catalog.service.s3.default-options.request-signing-enabled=true
nessie.catalog.service.s3.default-options.chunked-encoding-enabled=false
# OBS has no STS
nessie.catalog.service.s3.default-options.assume-role-enabled=false
 
nessie.catalog.service.s3.buckets.obs.name=example-bucket
nessie.catalog.service.s3.buckets.obs.endpoint=https://obs.ru-moscow-1.hc.sbercloud.ru
nessie.catalog.service.s3.buckets.obs.path-style-access=false
nessie.catalog.service.s3.buckets.obs.region=ru-moscow-1
```
 
Spark session — deliberately no S3 credentials, remote signing requested:
 
```python
.config("spark.sql.catalog.nessie", "org.apache.iceberg.spark.SparkCatalog")
.config("spark.sql.catalog.nessie.type", "rest")
.config("spark.sql.catalog.nessie.uri", "http://nessie.internal:29120/iceberg/")
.config("spark.sql.catalog.nessie.warehouse", "warehouse")
.config("spark.sql.catalog.nessie.token", token)  # bearer token from Keycloak
.config("spark.sql.catalog.nessie.header.X-Iceberg-Access-Delegation", "remote-signing")
.config("spark.sql.catalog.nessie.s3.remote-signing-enabled", "true")
.config("spark.sql.catalog.nessie.io-impl", "org.apache.iceberg.aws.s3.S3FileIO")
```
 
## Steps to reproduce
 
1. Configure Nessie against an S3-compatible store with `path-style-access=false`.
2. Create a namespace and an Iceberg table (this succeeds — Nessie writes metadata itself with its own credentials, no signing involved).
3. Write data to the table from a client using `X-Iceberg-Access-Delegation: remote-signing`.

## Isolated reproduction without Spark
 
`loadTable` with `X-Iceberg-Access-Delegation: remote-signing` returns:
 
```json
{
  "location": "s3://example-bucket/nessie-authz/sales/events_<uuid>",
  "config": {
    "s3.path-style-access": "false",
    "s3.endpoint": "https://obs.ru-moscow-1.hc.sbercloud.ru",
    "s3.remote-signing-enabled": "true",
    "s3.signer": "S3V4RestSigner",
    "s3.signer.endpoint": "v1/main%7Cwarehouse/s3sign/<token>",
    "client.region": "ru-moscow-1"
  }
}
```
 
Calling the signing endpoint directly with the same object in both addressing styles:
 
```bash
# 1. virtual-host style — what the client actually sends, per s3.path-style-access=false
curl -s -X POST "$SIGN_URL" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"region":"ru-moscow-1","method":"PUT",
       "uri":"https://example-bucket.obs.ru-moscow-1.hc.sbercloud.ru/nessie-authz/sales/events_<uuid>/data/probe.txt",
       "headers":{"Host":["example-bucket.obs.ru-moscow-1.hc.sbercloud.ru"]}}'
```
 
```json
{
  "error": {
    "message": "URI not allowed for signing: https://example-bucket.obs.ru-moscow-1.hc.sbercloud.ru/nessie-authz/sales/events_<uuid>/data/probe.txt",
    "type": "NotAuthorizedException",
    "code": 403
  }
}
```
 
```bash
# 2. path-style — signed successfully
curl -s -X POST "$SIGN_URL" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"region":"ru-moscow-1","method":"PUT",
       "uri":"https://obs.ru-moscow-1.hc.sbercloud.ru/example-bucket/nessie-authz/sales/events_<uuid>/data/probe.txt",
       "headers":{"Host":["obs.ru-moscow-1.hc.sbercloud.ru"]}}'
```
 
```json
{
  "uri": "https://obs.ru-moscow-1.hc.sbercloud.ru/example-bucket/nessie-authz/sales/events_<uuid>/data/probe.txt",
  "headers": { "Authorization": ["AWS4-HMAC-SHA256 Credential=.../ru-moscow-1/s3/aws4_request, ..."] }
}
```
 
Sending that correctly signed path-style request to OBS:
 
```xml
<Error>
  <Code>VirtualHostDomainRequired</Code>
  <Message>Virtual host domain is required while accessing a specific bucket.</Message>
  <VirtualHost>example-bucket.obs.ru-moscow-1.hc.sbercloud.ru</VirtualHost>
</Error>
HTTP 403
```
 
So the only URI form Nessie will sign is the one OBS refuses, and the only form OBS accepts is the one Nessie refuses to sign.

docker 0.107.9 and 0.108.3 (identical behaviour on both)

Client: Spark 3.5 + Iceberg 1.10.1 (iceberg-spark-runtime-3.5_2.12, iceberg-aws-bundle)

## Expected
 
Nessie signs the virtual-host-style URI that it instructed the client to use, and the write succeeds.
 
## Actual
 
Every data file write fails:
 
```
org.apache.iceberg.exceptions.ForbiddenException: Forbidden:
URI not allowed for signing:
https://example-bucket.obs.ru-moscow-1.hc.sbercloud.ru/nessie-authz/sales/events_<uuid>/data/event_date%3D2026-07-02/00000-12-<uuid>-0-00005.parquet
  at org.apache.iceberg.aws.s3.signer.S3V4RestSignerClient.sign(S3V4RestSignerClient.java:262)
  at software.amazon.awssdk.services.s3.DefaultS3Client.putObject(DefaultS3Client.java:11883)
  at org.apache.iceberg.aws.s3.S3OutputStream.completeUploads(S3OutputStream.java:443)
```
 
The rejection comes from Nessie — the request never reaches the object store.

## Workarounds tried, none successful
 
| Attempt | Result |
|---|---|
| `path-style-access=true` on Nessie and client | Signing works, OBS returns `VirtualHostDomainRequired` |
| `buckets.obs.external-endpoint` = endpoint without bucket | No change |
| `buckets.obs.external-endpoint` = endpoint with bucket in host (`https://example-bucket.obs...`) | No change |
| Explicit bucket declaration via `nessie.catalog.service.s3.buckets.*` | Applied (visible in `loadTable` response), no effect on signing |
| Upgrade 0.107.9 → 0.108.3 | Identical behaviour |
 
A rewriting proxy in front of OBS is not viable either: AWS SigV4 covers the `Host` header, so rewriting path-style into virtual-host invalidates the signature.
 
## Suggestion
 
When a bucket is configured with `path-style-access=false`, the signing endpoint could accept virtual-host-style URIs by resolving `<bucket>.<endpoint-host>` back to `s3://<bucket>/<key>` before checking it against the table location — mirroring the addressing style Nessie itself advertises to clients in the `loadTable` response.
 
Happy to test a snapshot build or provide any further detail.

mvanhorn:
Extend `S3Utils` with a bucket-aware HTTP-to-S3 conversion path: when a caller supplies a bucket already derived from an authorized S3 location, recognize `<bucket>.<endpoint-host>` as virtual-host style without requiring an `s3` hostname marker; otherwise preserve the existing AWS virtual-host and path-style parsing behavior. In `IcebergS3SignParams`, derive candidate bucket identities only from the signed warehouse/write/read location set, use a matching candidate to normalize the requested URI before the existing current/historical location checks, and pass the verified bucket through `SigningRequest` rather than re-extracting it from the custom HTTP hostname. Update `S3Signer` to use that bucket hint when reconstructing the storage URI used for per-bucket option and credential resolution, while retaining the legacy conversion fallback for callers without a hint. Nessie's HTTP-to-S3 location conversion recognizes virtual-host-style requests only when the hostname matches its AWS-shaped `s3` pattern. A request such as `https://example-bucket.obs.example.com/key` is therefore interpreted as path style, producing the wrong bucket and failing the table-location authorization check even when Nessie advertised `s3.path-style-access=false`. The same conversion is used later by `S3Signer` to resolve per-bucket credentials, so changing only the authorization comparison would still select storage options from the wrong location. This makes remote signing unusable for S3-compatible services such as SberCloud OBS that require virtual-host addressing and do not offer STS credentials vending.

Testing: Convert AWS virtual-host URLs and custom-endpoint path-style URLs exactly as before, and convert an OBS-shaped `https://example-bucket.obs.example.com/key` URL to `s3://example-bucket/key` only when `example-bucket` is supplied as the bucket hint; Preserve URI validation failures for unsupported schemes, missing hosts, and malformed path-style requests; do not treat a hostname that lacks the exact `<authorized-bucket>.` prefix as the hinted bucket.

Fixes #12854

### DIFF
--- a/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
+++ b/catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
@@ -101,6 +101,14 @@ public final class S3Utils {
   }
 
   public static String asS3Location(String uri) {
+    return asS3Location(uri, Optional.empty());
+  }
+
+  public static String asS3Location(String uri, String bucketHint) {
+    return asS3Location(uri, Optional.of(requireNonNull(bucketHint, "Bucket hint missing")));
+  }
+
+  private static String asS3Location(String uri, Optional<String> bucketHint) {
     URI httpUri = URI.create(uri);
     checkArgument(httpUri.getScheme() != null, "No scheme in URI: '%s'", httpUri);
     checkArgument(httpUri.getHost() != null, "No host in URI: '%s'", httpUri);
@@ -111,9 +119,15 @@ public final class S3Utils {
     String bucket;
     String key;
     Matcher matcher = S3_HOST_PATTERN.matcher(httpUri.getHost());
+    Optional<String> customVirtualHostedBucket =
+        bucketHint.filter(hint -> httpUri.getHost().startsWith(hint + "."));
     if (matcher.matches() && matcher.group(2) != null) {
       // virtual-hosted style
       bucket = matcher.group(2);
+      key = httpUri.getPath();
+    } else if (customVirtualHostedBucket.isPresent()) {
+      // virtual-hosted style with a caller-provided bucket for custom S3 endpoints
+      bucket = customVirtualHostedBucket.get();
       key = httpUri.getPath();
     } else {
       // path-style access style
--- a/catalog/files/impl/src/test/java/org/projectnessie/catalog/files/s3/TestS3Utils.java
+++ b/catalog/files/impl/src/test/java/org/projectnessie/catalog/files/s3/TestS3Utils.java
@@ -111,6 +111,22 @@ class TestS3Utils {
     assertThat(S3Utils.asS3Location(location)).isEqualTo(expected);
   }
 
+  @ParameterizedTest
+  @CsvSource({
+    "https://example-bucket.obs.example.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://example-bucket.s3.amazonaws.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://example-bucket.other.s3.amazonaws.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket.other/mydir/myfile",
+    "https://obs.example.com/example-bucket/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://not-example-bucket.obs.example.com/mydir/myfile, example-bucket, s3://mydir/myfile",
+  })
+  void asS3LocationWithBucketHint(String location, String bucket, String expected) {
+    assertThat(S3Utils.asS3Location(location, bucket)).isEqualTo(expected);
+  }
+
   @Test
   void asS3LocationErrors() {
     assertThatThrownBy(() -> S3Utils.asS3Location("ftp://localhost:9000/mybucket/mydir/myfile"))
@@ -125,5 +141,17 @@ class TestS3Utils {
     assertThatThrownBy(() -> S3Utils.asS3Location("https://s3.us-east-1.amazonaws.com"))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessage("Invalid S3 URI: 'https://s3.us-east-1.amazonaws.com'");
+
+    assertThatThrownBy(
+            () ->
+                S3Utils.asS3Location("ftp://example-bucket.obs.example.com/key", "example-bucket"))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Unsupported URI scheme: 'ftp'");
+    assertThatThrownBy(() -> S3Utils.asS3Location("http:///key", "example-bucket"))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("No host in URI: 'http:///key'");
+    assertThatThrownBy(() -> S3Utils.asS3Location("https://obs.example.com/", "example-bucket"))
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Invalid S3 URI: 'https://obs.example.com/'");
   }
 }
--- a/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
+++ b/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
@@ -17,7 +17,6 @@ package org.projectnessie.catalog.service.rest;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
-import static org.projectnessie.catalog.files.s3.S3Utils.extractBucketName;
 import static org.projectnessie.catalog.files.s3.S3Utils.normalizeS3Scheme;
 import static org.projectnessie.catalog.formats.iceberg.nessie.CatalogOps.CATALOG_S3_SIGN;
 import static org.projectnessie.catalog.formats.iceberg.rest.IcebergError.icebergError;
@@ -102,10 +101,37 @@ abstract class IcebergS3SignParams {
         "locations must be S3 URIs");
   }
 
-  @Value.Lazy
   String requestedS3Uri() {
-    return S3Utils.asS3Location(request().uri());
+    return requestedLocation().uri();
   }
+
+  Optional<String> requestedBucket() {
+    return requestedLocation().bucket();
+  }
+
+  @Value.Lazy
+  RequestedLocation requestedLocation() {
+    return Stream.concat(
+            Stream.of(warehouseLocation()),
+            Stream.concat(writeLocations().stream(), readLocations().stream()))
+        .map(URI::create)
+        .filter(uri -> S3Utils.isS3scheme(uri.getScheme()))
+        .map(S3Utils::extractBucketName)
+        .flatMap(Optional::stream)
+        .distinct()
+        .map(
+            bucket ->
+                new RequestedLocation(
+                    S3Utils.asS3Location(request().uri(), bucket), Optional.of(bucket)))
+        .filter(
+            location ->
+                location.bucket().orElseThrow().equals(URI.create(location.uri()).getAuthority()))
+        .findFirst()
+        .orElseGet(
+            () -> new RequestedLocation(S3Utils.asS3Location(request().uri()), Optional.empty()));
+  }
+
+  record RequestedLocation(String uri, Optional<String> bucket) {}
 
   @Value.Lazy
   boolean write() {
@@ -275,12 +301,16 @@ abstract class IcebergS3SignParams {
 
   private IcebergS3SignResponse sign(String uriToSign) {
     URI uri = URI.create(uriToSign);
-    Optional<String> bucket = extractBucketName(uri);
     Optional<String> body = Optional.ofNullable(request().body());
 
     SigningRequest signingRequest =
         SigningRequest.signingRequest(
-            uri, request().method(), request().region(), bucket, body, request().headers());
+            uri,
+            request().method(),
+            request().region(),
+            requestedBucket(),
+            body,
+            request().headers());
 
     SigningResponse signed = signer().sign(signingRequest);
 
@@ -297,7 +327,8 @@ abstract class IcebergS3SignParams {
                 List.of()));
     String s3Uri = requestedS3Uri();
     LOGGER.warn(
-        "Unauthorized signing request: key: {}, ref: {}, s3-uri: {}, request uri: {}, request method: {}, warehouse: {}, writeable locations: {}, readable locations: {}",
+        "Unauthorized signing request: key: {}, ref: {}, s3-uri: {}, request uri: {}, request"
+            + " method: {}, warehouse: {}, writeable locations: {}, readable locations: {}",
         key(),
         ref(),
         s3Uri,
--- a/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
+++ b/catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
@@ -58,6 +58,7 @@ import org.projectnessie.error.NessieNotFoundException;
 import org.projectnessie.model.Content;
 import org.projectnessie.model.ContentKey;
 import org.projectnessie.model.IcebergContent;
+import org.projectnessie.storage.uri.StorageUri;
 import org.projectnessie.versioned.RequestMeta.RequestMetaBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -114,10 +115,10 @@ abstract class IcebergS3SignParams {
     return Stream.concat(
             Stream.of(warehouseLocation()),
             Stream.concat(writeLocations().stream(), readLocations().stream()))
-        .map(URI::create)
-        .filter(uri -> S3Utils.isS3scheme(uri.getScheme()))
-        .map(S3Utils::extractBucketName)
-        .flatMap(Optional::stream)
+        .map(StorageUri::of)
+        .filter(uri -> S3Utils.isS3scheme(uri.scheme()))
+        .map(StorageUri::authority)
+        .flatMap(Stream::ofNullable)
         .distinct()
         .map(
             bucket ->
@@ -125,7 +126,7 @@ abstract class IcebergS3SignParams {
                     S3Utils.asS3Location(request().uri(), bucket), Optional.of(bucket)))
         .filter(
             location ->
-                location.bucket().orElseThrow().equals(URI.create(location.uri()).getAuthority()))
+                location.bucket().orElseThrow().equals(StorageUri.of(location.uri()).authority()))
         .findFirst()
         .orElseGet(
             () -> new RequestedLocation(S3Utils.asS3Location(request().uri()), Optional.empty()));
--- a/SECURITY_THREAT_MODEL.md
+++ b/SECURITY_THREAT_MODEL.md
@@ -224,6 +224,11 @@ a missing-table or future-location fallback. Missing content may be handled as a
 case where the signer endpoint was already minted for that location; authorization and unexpected
 lookup failures are fail-closed conditions. `(documented)`
 
+S3 request signing may recognize custom virtual-host endpoints that do not use AWS-style `s3`
+hostnames. This recognition is anchored to a bucket derived from the signer-authorized warehouse,
+write, or read locations; it does not widen the table paths permitted by current-location,
+historical-location, metadata-file, or object-storage-layout checks. `(documented)`
+
 Object-store credentials or temporary access material exposed through catalog features are in scope only
 for configured credential-vending or request-signing modes. Operators are responsible for cloud IAM
 policy scope, role trust policies, external IDs, token/session duration, and bucket or filesystem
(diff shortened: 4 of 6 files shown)


dimas-b:
@mvanhorn : Please sign the CLA.

mvanhorn:
The CLA is already signed on this one. CLAassistant's check on this PR reports "All committers have signed the CLA" (signed badge, https://cla-assistant.io/projectnessie/nessie?pullRequest=12865), and no checks are currently failing. Let me know if something else is blocking it.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
```diff
@@ -111,10 +119,16 @@ public static String asS3Location(String uri) {
...
     Matcher matcher = S3_HOST_PATTERN.matcher(httpUri.getHost());
+    Optional<String> customVirtualHostedBucket =
+        bucketHint.filter(hint -> httpUri.getHost().startsWith(hint + "."));
     if (matcher.matches() && matcher.group(2) != null) {
       // virtual-hosted style
       bucket = matcher.group(2);
       key = httpUri.getPath();
+    } else if (customVirtualHostedBucket.isPresent()) {
+      // virtual-hosted style with a caller-provided bucket for custom S3 endpoints
+      bucket = customVirtualHostedBucket.get();
```
dimas-b, line 130: How about we configure the Nessie Server as a whole to recognize custom virtual host domains?

This will make `S3_HOST_PATTERN` a default sub-case for virtual-host addressing.

dimas-b (commented):
Thanks for you contribution, @mvanhorn ! Please expect a slow review as it's Summer time 😉 

If you prefer, Nessie's Zulip space may be used for more interactive discussions about this change. The join link is on https://projectnessie.org/

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
```diff
@@ -101,6 +101,14 @@ private static Optional<String> extractBucketFromS3Uri(URI uri) {
   }
 
   public static String asS3Location(String uri) {
+    return asS3Location(uri, Optional.empty());
+  }
+
+  public static String asS3Location(String uri, String bucketHint) {
+    return asS3Location(uri, Optional.of(requireNonNull(bucketHint, "Bucket hint missing")));
```
adutra, line 108: nit: `Optional.of()` already invokes `Objects.requireNonNull()`.

catalog/files/impl/src/test/java/org/projectnessie/catalog/files/s3/TestS3Utils.java
```diff
@@ -111,6 +111,22 @@ void asS3Location(String location, String expected) {
...
+  @CsvSource({
+    "https://example-bucket.obs.example.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://example-bucket.s3.amazonaws.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://example-bucket.other.s3.amazonaws.com/mydir/myfile, example-bucket,"
+        + " s3://example-bucket.other/mydir/myfile",
+    "https://obs.example.com/example-bucket/mydir/myfile, example-bucket,"
+        + " s3://example-bucket/mydir/myfile",
+    "https://not-example-bucket.obs.example.com/mydir/myfile, example-bucket, s3://mydir/myfile",
```
adutra, line 124: Hmm this example makes me pause. It's clearly a virtual-hosted URI, but since the buckets don't match, it silently falls back to path-style and considers `mydir` as the bucket, which is most certainly wrong. I wonder if we shouldn't fail here?

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -102,11 +101,38 @@ && readLocations().stream().allMatch(l -> l.startsWith("s3:")),
         "locations must be S3 URIs");
   }
 
-  @Value.Lazy
   String requestedS3Uri() {
```
adutra:
```suggestion
  @Value.Lazy
  String requestedS3Uri() {
```

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -102,11 +101,38 @@ && readLocations().stream().allMatch(l -> l.startsWith("s3:")),
...
+  Optional<String> requestedBucket() {
+    return requestedLocation().bucket();
   }
 
+  @Value.Lazy
+  RequestedLocation requestedLocation() {
+    return Stream.concat(
+            Stream.of(warehouseLocation()),
+            Stream.concat(writeLocations().stream(), readLocations().stream()))
+        .map(URI::create)
```
adutra, line 117: `URI.create()` may throw on certain valid S3 locations. Use `org.projectnessie.storage.uri.StorageUri` instead.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -102,11 +101,38 @@ && readLocations().stream().allMatch(l -> l.startsWith("s3:")),
...
+        .map(S3Utils::extractBucketName)
+        .flatMap(Optional::stream)
+        .distinct()
+        .map(
+            bucket ->
+                new RequestedLocation(
+                    S3Utils.asS3Location(request().uri(), bucket), Optional.of(bucket)))
+        .filter(
+            location ->
+                location.bucket().orElseThrow().equals(URI.create(location.uri()).getAuthority()))
```
adutra, line 128: Same here, let's use `StorageUri`.

adutra:
@mvanhorn @snazy @dimas-b 

I think this PR makes a good step towards a better recognition of non-S3 virtual-host locations. But I'd like to propose an alternate design:

Virtual-host style by definition puts the bucket as the first subdomain. This is valid for all cases, S3 and non-S3 alike. 

Therefore, if we know that `pathStyleAccess=false`, then e.g. for a URI like `https://anything.arbitrary.endpoint.com/key` we know the bucket is `anything`: no hostname heuristic is needed. This would let us drop the `S3_HOST_PATTERN` detection entirely.

But, the `pathStyleAccess=false` information isn't available at signing time without a schema change.

Looking at the flow: `IcebergConfigurer.configureS3RequestSigningForTable` is where the signing token is minted. It has access to `StorageLocations` but not to `S3BucketOptions`. The `S3BucketOptions.pathStyleAccess()` method lives in `S3ObjectIO.icebergConfigOverrides()`, one level away. Getting it plumbed through to the token requires:

1. Adding `pathStyleAccess` to `SignerSignature`
2. Including it in the HMAC signature, which makes old in-flight tokens invalid after upgrade (breaking change)

If we are concerned about the breaking change, we could try to handle old tokens gracefully by falling back to the current heuristic when the boolean is absent.

WDYT?

snazy:
Agree, the PR makes sense. Once the `StorageUri` change is in, it looks good to merge.

mvanhorn:
Agreed on the principle: when the request really is virtual-hosted, the first subdomain is the bucket, and that holds for non-S3 endpoints too.

Where I'd push back is on needing `pathStyleAccess` to get there. The PR as it stands doesn't infer the bucket from the hostname at all. `requestedLocation()` takes the bucket candidates from `warehouseLocation()`, `writeLocations()` and `readLocations()`, and those are already covered by the token: `SignerSignature.sign()` hashes them as `b=`, `w=` and `r=`. So the bucket set is signed input we already have at signing time. The host check is then just `host.startsWith(bucket + ".")`, and the candidate is kept only if it round-trips (`bucket.equals(URI.create(uri).getAuthority())`).

That makes it a verification against a known value rather than a heuristic, which I think is actually stronger than deriving it from `pathStyleAccess`. `pathStyleAccess=false` tells you how the bucket is configured, not that this particular request URI arrived virtual-hosted. If a client sends a path-style URI for a bucket configured with `pathStyleAccess=false`, taking the first subdomain gives you the endpoint's first label as the bucket and we sign the wrong thing. The `startsWith` check can't fail that way.

The honest cost is that this doesn't let us delete `S3_HOST_PATTERN`. It stays as the fast path for real AWS hosts. I'd take that trade over putting `pathStyleAccess` in the HMAC: the graceful-fallback-when-absent option keeps the heuristic alive anyway, so we'd end up maintaining both paths permanently plus a token compatibility window.

If you want `pathStyleAccess` plumbed through for other reasons, happy to do it as a follow-up. It composes fine with this: when the flag is present and false, skip the candidate loop entirely.

@snazy - noted on `StorageUri`, I'll rebase once that lands.

snazy:
@mvanhorn mind adding the `StorageUri` change to your PR?

mvanhorn:
Pushed in 28e3a13. Both `URI.create` call sites in `requestedLocation()` now use `StorageUri.of` with the bucket taken from `authority()`, plus a test covering an S3 location with a space and square brackets in the key that `URI.create` rejects.

I left the `URI.create` in `sign()` alone: `SigningRequest.signingRequest` takes a `java.net.URI`, so switching it would change the signing API rather than the location parsing. Happy to do that separately if you want it.

`TestIcebergS3SignParams` is green locally, 55 tests.

On @adutra's other point, whether a bucket mismatch should fail instead of silently falling back to path-style: I've left the current fallback as-is, since that reads like a decision for you two rather than something to slip into this change.

catalog/service/rest/src/main/java/org/projectnessie/catalog/service/rest/IcebergS3SignParams.java
```diff
@@ -102,11 +102,38 @@ && readLocations().stream().allMatch(l -> l.startsWith("s3:")),
...
+        .flatMap(Stream::ofNullable)
+        .distinct()
+        .map(
+            bucket ->
+                new RequestedLocation(
+                    S3Utils.asS3Location(request().uri(), bucket), Optional.of(bucket)))
+        .filter(
+            location ->
+                location.bucket().orElseThrow().equals(StorageUri.of(location.uri()).authority()))
+        .findFirst()
```
snazy, line 130: (blocking) `findFirst()` makes this ambiguous when authorized buckets overlap. For example, with warehouse bucket `foo` and a historical read bucket `foo.bar`, `foo.bar.obs.example.com` matches both candidates but selects `foo` because the warehouse is visited first. We then authorize the path and select credentials as `foo` while signing a request addressed to `foo.bar`. Can we either disambiguate using the configured endpoint/addressing style, or reject multiple matching candidates? Please add a current/historical-location test with `foo` and `foo.bar`.

mvanhorn:
Good catch, and it was worse than ambiguous ordering: we authorized the path and picked credentials as `foo` while signing a request addressed to `foo.bar`.

93242d9 fails closed. An exact authority match is required first, endpoint/path-style disambiguation only applies after that, and more than one surviving candidate is rejected rather than guessed. I went with rejection over disambiguation because guessing wrong here is a cross-bucket authorization error, and a clear failure is easier to operate than a silent mismatch.

Added the current/historical test with `foo` and `foo.bar`, and noted the stricter rejection in SECURITY_THREAT_MODEL.md since it is a behavior change for anyone relying on the old fallback. Gradle spotless, checkstyle and TestIcebergS3SignParams are green.

catalog/files/impl/src/main/java/org/projectnessie/catalog/files/s3/S3Utils.java
dimas-b: Any feedback on this proposal? 😉 

I mean that we could configure specific virtual host domains in the Nessie Server and use that configuration to decide whether the `httpUri` is a virtual host or not.

AWS becomes a config default as opposed to a code-level hard-coded pattern.

WDYT?

dimas-b: The proposal from @adutra [here](https://github.com/projectnessie/nessie/pull/12865#issuecomment-5354116201) also looks reasonable to me, if we do not want to bother with new config.

mvanhorn:
Sorry for the slow turnaround, and thanks for keeping this alive.

I like @adutra's design better than mine. Deriving the bucket from the first subdomain when `pathStyleAccess=false` is the actual definition of virtual-host style, so it is right by construction everywhere, S3 and non-S3, and it lets `S3_HOST_PATTERN` go away instead of growing another special case. My pattern approach only ever approximates that.

The cost he identified is real though: `pathStyleAccess` has to reach `SignerSignature`, and folding it into the HMAC invalidates in-flight tokens across an upgrade. If that break is acceptable on a minor, I would rather pay it once than carry the heuristic.

On configurable virtual-host domains: I would take that as the fallback if the schema change is off the table. It moves AWS from hard-coded to a config default, which is strictly better than today, but it still asks operators to describe something the addressing style already tells us.

@snazy on the `findFirst()` ambiguity: agreed it is a real bug, and note it mostly disappears under adutra's design, since `foo.bar.obs.example.com` yields bucket `foo.bar` directly rather than matching candidates. Either way I will reject multiple matching candidates rather than silently taking the first, and add the current/historical-location test with `foo` and `foo.bar` that you asked for.

Happy to redo this on adutra's design if you both agree that is the direction.

PR #12874, opened 2026-08-14
Fix invalid git clone URL in try-nessie guide

daviddallakyan2005:
## Summary
- `site/docs/guides/try-nessie.md` started with `git clone projectnessie/nessie`, which is not a valid Git URL, so the first command on the getting-started page fails.
- Changed it to `git clone https://github.com/projectnessie/nessie.git`.

## Test plan
- [x] Docs-only change to an MkDocs page; no build affected


dimas-b (approved):
Nice catch! Thanks, @daviddallakyan2005 !

PR #12884, opened 2026-08-15
Prefetch commits when identifying merge bases

### LINKED ISSUE #12843: [Bug]:  Merge times out on branches with deep commit history
### What happened

Merging any branch into a long-lived branch that has a deep commit history takes
minutes and never returns within normal client/proxy timeouts. This happens even
when the merge is effectively a **no-op** (the source head is already an ancestor
of the target), so it is not related to the number of changed keys.

We have a long-lived branch that receives frequent fan-in merges from many
short-lived working branches (thousands of merge commits accumulated over months).
Merging a small working branch (~45 content keys) into it:

- Ref resolution is instant (`GET /trees/{ref}` ~10–40 ms).
- `diff` between the two branches (~10k keys) returns in ~5 s.
- **The merge itself does not complete**: >280 s as a `dryRun`, and a 600 s client
  timeout is exceeded. The client sees `HttpTimeoutException: request timed out`,
  though the server keeps processing and eventually commits (logs `POST .../merge 200`).

A thread dump taken while the merge is running shows the time is spent in
`MergeBase.identifyAllMergeBases` walking the commit graph and fetching commits
**one at a time** from the JDBC2 store. The worker thread burns almost no CPU
(~130 ms across 2 dumps) — it is almost entirely blocked on sequential DB round-trips.

Relevant frames:

org.postgresql.jdbc.PgStatement.executeInternal
org.projectnessie.versioned.storage.jdbc2.Jdbc2Persist.fetchTypedObj (Jdbc2Persist.java:164)
org.projectnessie.versioned.storage.cache.CachingPersistImpl.fetchTypedObj
org.projectnessie.versioned.storage.common.logic.CommitLogicImpl.fetchCommit (CommitLogicImpl.java:865)
org.projectnessie.versioned.storage.common.logic.CommitLogicImpl.lambda$identifyMergeBase$0 (CommitLogicImpl.java:847)
org.projectnessie.versioned.storage.common.logic.MergeBase.lambda$shallowCommit$1 (MergeBase.java:313)
org.projectnessie.versioned.storage.common.logic.MergeBase.shallowCommit (MergeBase.java:310)
org.projectnessie.versioned.storage.common.logic.MergeBase.flagReachableCommits (MergeBase.java:148)
org.projectnessie.versioned.storage.common.logic.MergeBase.findMergeBases (MergeBase.java:103)
org.projectnessie.versioned.storage.common.logic.MergeBase.identifyAllMergeBases (MergeBase.java:92)

The cost appears to scale with the **commit-history depth of the target branch**,
not with the size of the merge. Because the working branches are never deleted and
the target is never re-based/compacted, history grows without bound and merges get
progressively slower until they no longer complete. Server-side auto-merges are
affected too (they silently stall).

### How to reproduce it

1. Deploy Nessie with the JDBC2 version store (PostgreSQL).
2. Create a long-lived branch `main-ish` and, in a loop, create a short working
   branch, commit to it, and merge it into `main-ish` — repeat enough times to build
   a deep linear/fan-in history (tens of thousands of commits).
3. Attempt `MERGE BRANCH <any-working-branch> INTO main-ish` (or `POST /trees/{ref}/history/merge`),
   including a case where the working branch head is already merged (no-op).
4. Observe the merge does not return within client timeouts; a thread dump shows the
   worker in `MergeBase.identifyAllMergeBases` doing per-commit `fetchTypedObj` calls
   against the JDBC2 store.

`JDBC2` version store, backend = PostgreSQL (Amazon Aurora PostgreSQL). `nessie.server.authentication.enabled=true`, authorization rules enabled (CEL). Deployed in Eks cluster with aws cognito for auth.   - Nessie server: **0.104.1** (`ghcr.io/projectnessie/nessie:0.104.1`), Quarkus 3.21.3, Java 21 - Nessie CLI: **0.106.1** - REST API: v2 (spec 2.2.0) - Deployment: Kubernetes (official Docker image)

_No response_

### Additional information

_No response_

daviddallakyan2005:
## Summary
- Fixes #12843: `MergeBase` identified merge bases by loading each parent with `persist.fetchTypedObj`. On JDBC2 that is one borrowed connection per commit, so merges into a deep history time out — including no-op merges.
- Prefetch via `fetchTypedObjsIfExist` in chunks of 100, using `CommitObj.tail()` the same way `CommitLogIter` already bulk-fetches the commit log. No algorithm change (a later PR could stop the walk once the nearest base is known).

@dimas-b invited a PR on the issue. Happy to adjust the chunk size.

## Test plan
- [x] `./gradlew :nessie-versioned-storage-common:test --tests '*MergeBase*'`
- [x] `./gradlew :nessie-versioned-storage-inmemory:test --tests '*CommitLogicTests'`
- [x] `./gradlew :nessie-versioned-storage-jdbc2:test --tests '*CommitLogicTests.mergeBasePrefetchesDeepHistory'`
- [x] `./gradlew sAp compileAll jar codeChecks`
- [ ] CI `CI intTest versioned/stores` (full JDBC2 Testcontainers suite)

The new `mergeBasePrefetchesDeepHistory` test counts Persist fetches on a 120-commit chain and asserts bulk fetches (not per-id `fetchTypedObj`), chunk size ≤ 100, and that overlapping `tail()` windows are not fetched twice.


versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -841,21 +844,76 @@ public ObjId findMergeBase(@Nonnull ObjId targetId, @Nonnull ObjId sourceId)
...
+      ObjId[] ids = new ObjId[n];
+      Iterator<ObjId> it = pending.iterator();
+      for (int i = 0; i < n; i++) {
+        ids[i] = it.next();
+        it.remove();
+      }
+      // perform a bulk-load against the database, populates the cache
+      CommitObj[] objs = persist.fetchTypedObjsIfExist(ids, COMMIT, CommitObj.class);
+      for (int i = 0; i < n; i++) {
+        cache.put(ids[i], objs[i]);
```
dimas-b: The `cache` will keep growing for the whole duration of the `identifyMergeBase()` operation... and I suppose it can grow very large 🤔 

I think we should at least remove the returned entries and probably cap the size of what remains.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
```diff
@@ -841,21 +844,76 @@ public ObjId findMergeBase(@Nonnull ObjId targetId, @Nonnull ObjId sourceId)
...
+
+    @Override
+    public CommitObj apply(ObjId commitId) {
+      if (EMPTY_OBJ_ID.equals(commitId)) {
+        return null;
+      }
+      if (cache.containsKey(commitId)) {
+        return cache.get(commitId);
+      }
+      enqueue(commitId);
```
dimas-b: `pending` may be non-empty and `commitId` will en up at the end. I'm not sure it is an optimal loading strategy. Side branches (`secondaryParents()` from previously loaded commits) may not need to be preloaded if the caller asks for some other commit.

Rather than trying to predict what will be loaded, should we rather implement batch load inside `MergeBase`? It has an exact processing `queue` 🤔

dimas-b:
Thanks for your continued contributions, @daviddallakyan2005 !

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
daviddallakyan2005: No cache now. `MergeBase` stages loaded commits and drops each one as it becomes a `ShallowCommit`, bounded at `2 * DEFAULT_PARENTS_PER_COMMIT`, one `tail()` window per merged branch. Nothing is evicted: a batch only takes as much look-ahead as there is room for. Happy to change the bound.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/CommitLogicImpl.java
daviddallakyan2005: Moved into `MergeBase`. `CommitLogicImpl` is a one-line bulk fetch again, and `secondaryParents()` are no longer prefetched.

I didn't drive it off the queue though: the queue holds already-loaded commits, so its unloaded frontier is 1–2 ids on a linear history, which is #12843, and that still costs a round trip per commit. Look-ahead comes from `tail()` of commits the walk has reached instead, the same paging `CommitLogIter` uses. Bounded, worst case 39 commits loaded and unused.

Two 60-commit branches: 121 × `fetchTypedObj` → 5 bulk calls, 121 ids, none twice. Merge base next to both heads: 100+ ids → 22. Traversal unchanged.

dimas-b: @daviddallakyan2005 : heads up: I can probably review again only in Sept :sweat_smile:

daviddallakyan2005: @dimas-b No rush. Sept is fine.

If you have higher-priority Nessie problems to solve while this waits, I would be happy to take some of them on. I enjoyed working with Nessie and would like to contribute more. Point me at whatever is actually in the way.

snazy (commented):
Thanks for the contribution.
I don't see a blocker or anything really concerning here - will leave it up to @dimas-b as he already took a look at this change.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
```diff
@@ -307,28 +318,108 @@ private ShallowCommit shallowCommit(ObjId objId) {
...
+    return result;
+  }
+
+  /**
+   * Remembers {@link CommitObj#tail()}, the first-parent chain ahead of the given commit, so that
+   * the next batch continues along the chain that is currently being walked. Merge parents are not
+   * added here: those are only needed once the commit carrying them is processed.
+   */
+  private void rememberChainAhead(CommitObj commit) {
+    for (ObjId id : commit.tail()) {
```
dimas-b, line 401: The iteration pattern in `flagReachableCommits()` is breadth-first (by commit parents). This method makes pre-fetching work in a depth-first way.

As far as I can tell, this strategy will work fine for long segments of unbranched commits. However, if the commit tree is heavyly branched, the `prefetched` buffer will quickly fill with `tails` of a couple of merge commits, which will stifle the batching when the main loop moves to the next merge parent. WDYT?

Would it be possible to make some unit tests that show the effectiveness of the pre-fetching algorithm both in mostly linear and heavily branched commit histories? The test histories do not have to be long as long as the batching effect is observable. I suppose the top-level criterion is the number of Persistence API hits vs. the number of commits traversed with and without batching (i.e. one batch load call counts as `1` regardless of the size of its ID list argument).

My personal gut feel is that `rememberChainAhead` should probably go breadth-first to match the main iteration loop.

dimas-b (commented):
@daviddallakyan2005 : the change generally LGTM, but I'd like to see some numbers that show its effectiveness, if you do not mind 😉 

Feel free to ping me in Nessie's Zulip channel for faster turnaround if you have questions about the kind of test I mean.

versioned/storage/common/src/main/java/org/projectnessie/versioned/storage/common/logic/MergeBase.java
dimas-b: Also: another efficiency metric should probably be the number of unnecessary loaded commits (pre-loaded, but never used). Those consume the Persistence cache space, IIRC.

snazy: A potential mitigation (not a full breadth-first approach) could be to use half of the buffer. That leaves capacity for the next independently reached path, instead of letting the first tail monopolize all 40 slots.

Maybe something along this:

```
private void rememberChainAhead(CommitObj commit) {
  int added = 0;
  for (ObjId id : commit.tail()) {
    if (EMPTY_OBJ_ID.equals(id)
        || chainAhead.size() >= PREFETCH_LIMIT
        || added >= DEFAULT_PARENTS_PER_COMMIT) {
      break;
    }
    if (!commits.containsKey(id) && !prefetched.containsKey(id) && chainAhead.add(id)) {
      added++;
    }
  }
}
```

PR #9302, opened 2024-08-08
Catalog: Implementation for down-scoped GCP/GCS access tokens

snazy:
Implements the production code for down-scoped access tokens. Currently untested.

Fixes #9298

PR #12901, opened 2026-08-18
Catalog/GCS: fix down-scoped credentials failing on every request

amdubovoi:
## Summary

Down-scoped GCS credentials (`downscoped-credentials.enable=true`) never worked. Two
independent bugs make `GcsStorageSupplier.generateDelegationToken()` fail for every
credential-vending request:

1. `GcsClients.buildCredentials()` returns unscoped credentials. Without an OAuth2 scope,
   google-auth-library signs a self-signed JWT instead of fetching an access token, and
   Google's token exchange rejects it with `invalid_scope`.

2. The generated Credential Access Boundary conditions use CEL `matches()`:

   ```java
   return "resource.name.matches('"
       + quoteForCelString(format("projects/_/buckets/%s/objects/", bucket))
       + RANDOMIZED_PART
       + quoteForCelString(path)
       + ".*')";
   ```

   Google's IAM does not support `matches()` in these conditions, only `startsWith()`,
   `endsWith()` and equality. Every request is rejected with `invalid_request`, whatever the
   pattern contains. `Pattern.quote()` also produces `\Q...\E`, which is Java-specific and
   would not work even if regular expressions were supported.

The conditions are also too broad. A prefix without a trailing `/` matches sibling locations
that start with the same characters, so a credential for `warehouse/orders` also grants access
to `warehouse/orders2`. S3 avoids this because its resource ARN ends in `/*`, which stops the
prefix at a path boundary.

## Impact

Every Iceberg REST client loading a table from a warehouse with GCS down-scoped credentials
enabled gets an HTTP 500. The option is off by default and documented as experimental, so
deployments that never enabled it are unaffected.

Broken since the feature was added in 2b98754 (Aug 2024, #9302).

## Fix

Scope the source credentials to `https://www.googleapis.com/auth/cloud-platform`, and build the
conditions from `startsWith()` with a trailing `/`.

`RANDOMIZED_PART` goes away with the regex. It only existed to match the randomized
object-storage layout, and it did not match that layout anyway: the hash is inserted after
`write.data.path`, which Nessie sets to the warehouse, so those files land under
`<warehouse>/<hash>/<namespace>/<table>/` rather than between the bucket and the table path.

Verified against a real GCS bucket: a vended credential reads its own table (200), is denied a
sibling table (403), and lists the table prefix both with and without a trailing slash (200).

Credentials are scoped to a table's location, so they do not cover tables using
`write.object-storage.enabled=true`, whose data files live under a randomized prefix ahead of
the table path. That shape cannot be expressed with prefix conditions. This is documented on
the `enable()` option and in `SECURITY_THREAT_MODEL.md`.

## Tests

The existing tests asserted the `\Q...\E` expressions and passed, because the CEL evaluator
used in tests accepts `matches()` while Google's does not. They now assert the `startsWith()`
form, and the `validateCelSyntax` test fails if `matches()` reappears.

### DIFF
--- a/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/GcsDownscopedCredentials.java
+++ b/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/GcsDownscopedCredentials.java
@@ -33,6 +33,11 @@ public interface GcsDownscopedCredentials {
    *
    * <p>The current default is to not enable short-lived and scoped-down credentials, but the
    * default may change to enable in the future.
+   *
+   * <p>Credentials are scoped to a table's location. Tables using {@code
+   * write.object-storage.enabled=true} write their data files under a randomized prefix that
+   * precedes the table path, which cannot be expressed as a scope, so clients are denied access to
+   * those data files.
    */
   Optional<Boolean> enable();
 
--- a/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/GcsDownscopedCredentials.java
+++ b/catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/GcsDownscopedCredentials.java
@@ -38,6 +38,11 @@ public interface GcsDownscopedCredentials {
    * write.object-storage.enabled=true} write their data files under a randomized prefix that
    * precedes the table path, which cannot be expressed as a scope, so clients are denied access to
    * those data files.
+   *
+   * <p>With {@code auth-type=ACCESS_TOKEN} or {@code auth-type=USER} the configured credential is
+   * used with its existing OAuth scopes and must already have the {@code
+   * https://www.googleapis.com/auth/cloud-platform} scope. Nessie requests that scope for
+   * service-account credentials and for Application Default Credentials that support scoping.
    */
   Optional<Boolean> enable();
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ as necessary. Empty sections will not end in the release notes.
 
 ### Upgrade notes
 
+- Catalog/GCS: Down-scoped credentials, enabled via
+  `nessie.catalog.service.gcs.default-options.downscoped-credentials.enable`, were not functional and
+  failed every credential-vending request. They work now. Vended credentials are scoped to a table's
+  location, so they do not cover tables using `write.object-storage.enabled=true`, which writes data
+  files under a randomized prefix that Credential Access Boundary conditions cannot express. Do not
+  enable down-scoped credentials for warehouses whose tables use that layout.
+
 ### Breaking changes
 
 ### New Features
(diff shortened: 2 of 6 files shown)


amdubovoi:
Left for a follow-up: for tables with `write.object-storage.enabled=true`, `IcebergConfigurer` sets `write.data.path` to the warehouse but does not add that location to `StorageLocations`, so vended credentials do not cover the data files. This affects S3 as well. The fix widens the credential scope for those tables, so it seemed better to keep it out of this PR

snazy:
Nice catch!

This explains a particularly annoying “enabled but fails every time” path.

Switching from `matches()` to the IAM-supported `startsWith()` form, and scoping the refreshable source credentials to cloud-platform, looks right. I also appreciate the trailing-slash handling: it prevents a credential for one table location from granting object access to a sibling prefix with the same start.

The documentation around `write.object-storage.enabled=true` makes the remaining limitation clear. One small non-blocking follow-up: document that `auth-type=ACCESS_TOKEN` still requires the operator to provide a token minted with the cloud-platform scope, since Nessie cannot add scopes to an already-minted token.

Otherwise, LGTM.

amdubovoi:
Thanks for the review. Documented the `ACCESS_TOKEN` scope requirement and rebased on latest main

catalog/files/api/src/main/java/org/projectnessie/catalog/files/config/GcsDownscopedCredentials.java
```diff
@@ -33,6 +33,15 @@ public interface GcsDownscopedCredentials {
...
    * default may change to enable in the future.
+   *
+   * <p>Credentials are scoped to a table's location. Tables using {@code
+   * write.object-storage.enabled=true} write their data files under a randomized prefix that
+   * precedes the table path, which cannot be expressed as a scope, so clients are denied access to
+   * those data files.
+   *
+   * <p>With {@code auth-type=ACCESS_TOKEN} the configured token is used as-is, so it must already
+   * have the {@code https://www.googleapis.com/auth/cloud-platform} scope that token exchange
+   * requires. Nessie adds that scope to the other auth types itself.
```
snazy: Minor, but worthwhile correction:

With `auth-type=ACCESS_TOKEN` or `auth-type=USER`, the configured credential is used with its existing OAuth scopes and must already have `https://www.googleapis.com/auth/cloud-platform`, which token exchange requires. Nessie requests that scope for service-account credentials. Application Default Credentials must likewise support or already have that scope.

amdubovoi: Okay, that sent me for another round of Google STS rabbit-hole exploration, and here's what I found:

You're right about USER: `createScoped` is `return this` there, so it does nothing

But token exchange doesn't need `cloud-platform`. STS took a `devstorage.read_only` token fine, and even a `userinfo.email` one (which then 403s on the object). The real failure was a step earlier: an unscoped `ServiceAccountCredentials` can't mint a token at all (400)

Reworded the javadoc. Can drop the no-op `createScoped` on the USER branch too if you want

snazy:
Thanks! This is close. Could we make the requirement explicit as `https://www.googleapis.com/auth/cloud-platform`, rather than “grant access to the bucket’s objects”? A Storage-scoped token can satisfy the latter but still be rejected by the downscoping token exchange.

Otherwise, the USER and scope-capable ADC distinction looks right.

And the term "rabbit-hole" totally spot-on ;)

amdubovoi:
Makes sense, `cloud-platform` it is. I couldn't get a Storage-scoped token rejected in my testing, but that path isn't documented anywhere I could find, so pinning it to `cloud-platform` is the safer thing to promise

Updated

snazy (approved):
Thanks for the contribution!

